PR #784, opened 2023-02-08
[oneDPL] radix sort: moved some type aliases to place of using ones

MikeDvorskiy:
[oneDPL] radix sort: moved some type aliases to place of using ones.


SergeyKopienko (approved):
LGTM

PR #785, opened 2023-02-08
Fix error in iterator comparison

SergeyKopienko:
In this PR we fix an error in the comparison of input and output iterators which was introduced in https://github.com/oneapi-src/oneDPL/pull/761 :
in this comparison we receive compile error in DPC++ compiler (win) ```Error GE509BDA3 invalid operands to binary expression```

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
@@ -183,8 +183,8 @@ __pattern_transform_scan_base(_ExecutionPolicy&& __exec, _Iterator1 __first, _It
     const auto __n = __last - __first;
 
     // This is a temporary workaround for an in-place exclusive scan while the SYCL backend scan pattern is not fixed.
-    const bool bInplaceExclusiveScan = __n > 1 && !_Inclusive{} && __is_equal_iterators(__first, __result);
-    if (!bInplaceExclusiveScan)
+    const bool __is_scan_inplace_exclusive = __n > 1 && !_Inclusive{} && __is_equal_iterators(__first, __result);
+    if (!__is_scan_inplace_exclusive)
     {
         __pattern_transform_scan_base_impl(__exec, __first, __last, __result, __unary_op, __init, __binary_op,
                                            _Inclusive{});
--- a/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
@@ -166,7 +166,7 @@ template <typename Iterator1, typename Iterator2, std::enable_if_t<!std::is_same
 constexpr bool
 __is_equal_iterators(Iterator1 it1, Iterator2 it2)
 {
-    static_assert("In-place exclusive scan works correctly only if an input and an output iterators are the same type.");
+    // In-place exclusive scan works correctly only if an input and an output iterators are the same type.
     return false;
 }
 
--- a/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
@@ -155,17 +155,18 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
         .wait();
 }
 
-template <typename Iterator1, typename Iterator2, std::enable_if_t<std::is_same_v<Iterator1, Iterator2>, int> = 0>
+template <typename Iterator1, typename Iterator2>
 constexpr bool
 __is_equal_iterators(Iterator1 it1, Iterator2 it2)
 {
-    return it1 == it2;
-}
+    using DecayIt1 = std::decay_t<Iterator1>;
+    using DecayIt2 = std::decay_t<Iterator2>;
 
-template <typename Iterator1, typename Iterator2, std::enable_if_t<!std::is_same_v<Iterator1, Iterator2>, int> = 0>
-constexpr bool
-__is_equal_iterators(Iterator1 it1, Iterator2 it2)
-{
+    if constexpr (std::is_same_v<DecayIt1, DecayIt2>)
+    {
+        return it1 == it2;
+    }
+
     // In-place exclusive scan works correctly only if an input and an output iterators are the same type.
     return false;
 }
--- a/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
@@ -159,12 +159,9 @@ template <typename _Iterator1, typename _Iterator2>
 constexpr bool
 __check_equal_iterators(_Iterator1 __it1, _Iterator2 __it2)
 {
-    using _Decay_Iterator1 = std::decay_t<_Iterator1>;
-    using _Decay_Iterator2 = std::decay_t<_Iterator2>;
-
     // In-place exclusive scan works correctly only if an input and an output iterators are the same type.
     // Otherwise, there is no way to check an in-place case and a workaround below is not applied.
-    if constexpr (std::is_same_v<_Decay_Iterator1, _Decay_Iterator2>)
+    if constexpr (::std::is_same_v<::std::decay_t<_Iterator1>, ::std::decay_t<_Iterator2>>)
     {
         return __it1 == __it2;
     }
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -155,6 +155,28 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+template <typename Iterator>
+bool
+__is_equal_iterators_impl(Iterator it1, Iterator it2)
+{
+    return it1 == it2;
+}
+
+template <typename Iterator1, typename Iterator2>
+bool
+__is_equal_iterators_impl(Iterator1 it1, Iterator2 it2)
```
MikeDvorskiy, line 167: I think it is a redefinition of `__is_equal_iterators_impl` in case of Iterator1 and Iterator2 are the same type..
Please use SFINAE approach, like this
https://godbolt.org/z/rGfz7j8nh

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -155,6 +155,28 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+{
+    static_assert("Unable to compare the iterators of different types");
```
MikeDvorskiy, line 169: Should be another text, like this:
"In-place exclusive scan works correctly only if an input and an output iterators are the same type."

MikeDvorskiy (changes requested):

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -155,6 +155,21 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+__is_equal_iterators(Iterator1 it1, Iterator2 it2)
+{
+    return it1 == it2;
+}
+
+template <typename Iterator1, typename Iterator2, std::enable_if_t<!std::is_same_v<Iterator1, Iterator2>, int> = 0>
+constexpr bool
+__is_equal_iterators(Iterator1 it1, Iterator2 it2)
+{
+    static_assert("In-place exclusive scan works correctly only if an input and an output iterators are the same type.");
```
MikeDvorskiy, line 169: And should not be any assert here - just a comment.
Otherwise it "kills" the regular implementation  - "else" code branch (191)

SergeyKopienko: FYI: in our tests, there are no cases when it's happened now in our tests.
If we will have a simple comment, depending on iterator types we have correct/incorrect working code without compile errors.
But up to you.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -155,6 +155,21 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
         .wait();
 }
 
+template <typename Iterator1, typename Iterator2, std::enable_if_t<std::is_same_v<Iterator1, Iterator2>, int> = 0>
```
MikeDvorskiy: I would like to notice that for example a case `const Iterator1 `and `Iterator2` should work as well here.
Because it is a typical case when input iterator is const, output is non-const (of course).

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
MikeDvorskiy: > in our tests

"Unfortunately", a user runs not only our tests)...

SergeyKopienko: Yes, sure, and in this case better to have some info is our algorithm works correctly or not.
How are you going to give this use to oneDPL users? Simple by comment?

MikeDvorskiy: > we have correct/incorrect working code without compile errors.

Unfortunately, there is no way to check it whether compile time or run time in a general case.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -167,7 +183,7 @@ __pattern_transform_scan_base(_ExecutionPolicy&& __exec, _Iterator1 __first, _It
     const auto __n = __last - __first;
 
     // This is a temporary workaround for an in-place exclusive scan while the SYCL backend scan pattern is not fixed.
-    const bool bInplaceExclusiveScan = __n > 1 && !_Inclusive{} && __first == __result;
+    const bool bInplaceExclusiveScan = __n > 1 && !_Inclusive{} && __is_equal_iterators(__first, __result);
```
akukanov, line 186: We do not use the Hungarian notation, and we (sort of have to) use uglification. This variable is better named `__is_scan_inplace_exclusive`

Also please add all needed underscores to the `__is_equal_iterators` helper.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
SergeyKopienko: Already fixed.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
SergeyKopienko: I believe we may check that ```std::address()``` is applicable for current iterator type and in this case use it too before ```return false``` in ```__is_equal_iterators``` function.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -155,6 +155,22 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
         .wait();
 }
 
+template <typename _Iterator1, typename _Iterator2>
+constexpr bool
+__is_equal_iterators(_Iterator1 it1, _Iterator2 it2)
```
MikeDvorskiy, line 160: As far as we have a plural form  - "iterators" we have to use the other form of "to be" verb  or don't use "to be" at all. So, my suggestion
`__are_equal_iterators` or `__check_equal_iterators`

SergeyKopienko: OK, let's use your variant ``` __check_equal_iterators```

SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -155,6 +155,22 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+{
+    using _Decay_Iterator1 = std::decay_t<_Iterator1>;
+    using _Decay_Iterator2 = std::decay_t<_Iterator2>;
+
+    if constexpr (std::is_same_v<_Decay_Iterator1, _Decay_Iterator2>)
+    {
+        return it1 == it2;
+    }
+
+    // In-place exclusive scan works correctly only if an input and an output iterators are the same type.
```
MikeDvorskiy, line 170: I think the comment describes the situation non quite right. 
Lets add the following text:
"Otherwise, there is no way to check an in-place case and a workaround below is not applied."

SergeyKopienko: Fixed,

MikeDvorskiy (approved):

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -155,6 +155,23 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
         .wait();
 }
 
+template <typename _Iterator1, typename _Iterator2>
+constexpr bool
+__check_equal_iterators(_Iterator1 it1, _Iterator2 it2)
```
akukanov, line 160: `it1` and `it2` also need underscores.

akukanov (approved):

Issue #788, opened 2023-02-09
Build error: functional-style cast, buffer_wrapper.pass


abagusetty:
```[bash]
...../oneDPL/test/general/buffer_wrapper.pass.cpp:37:26: error: no matching conversion for functional-style cast from 'sycl::buffer<unsigned int, dim, aligned_allocator<unsigned int>>' to 'sycl::host_accessor<const unsigned int, 1, sycl::access_mode::read>'
        T* actual_data = sycl::host_accessor<T, 1, sycl::access_mode::read>(buf).get_pointer();
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

SergeyKopienko:
@abagusetty, could you please specify on which compiler you received this compile error?

abagusetty:
@SergeyKopienko Thanks for looking into it. I have just tried with this commit [hash](fdf63c45d86062665193c9935155d6688e73ba7c).

abagusetty:
@SergeyKopienko Thanks for such a prompt fix #793

Compiler was built with this hash from source: https://github.com/intel/llvm/commit/1ed2378b6072a0f70c55f4105339b527c7015a75
With C++ 17.

SergeyKopienko:
@abagusetty, thank you for this update.
May you check our fix from https://github.com/oneapi-src/oneDPL/pull/793 and send me info that this issue is resolved?

abagusetty:
Just checked. This issue is resolved. Thanks again!

SergeyKopienko:
Thank you, @abagusetty !

PR #793, opened 2023-02-13
Fix errors in test buffer_wrapper.pass

SergeyKopienko:
In this PR we fix compile error from the issue https://github.com/oneapi-src/oneDPL/issues/788 :
```
test/general/buffer_wrapper.pass.cpp:37:26: error: no matching conversion for functional-style cast from 'sycl::buffer<unsigned int, dim, aligned_allocator<unsigned int>>' to 'sycl::host_accessor<const unsigned int, 1, sycl::access_mode::read>'
        T* actual_data = sycl::host_accessor<T, 1, sycl::access_mode::read>(buf).get_pointer();
                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This error were introduced in https://github.com/oneapi-src/oneDPL/pull/281

### DIFF
--- a/test/general/buffer_wrapper.pass.cpp
+++ b/test/general/buffer_wrapper.pass.cpp
@@ -34,7 +34,7 @@ struct test_buffer_wrapper
         EXPECT_TRUE(end - begin == size, "wrong effect of iterator's operator - iterator");
 
         auto buf = begin.get_buffer();
-        sycl::host_accessor<T, 1, sycl::access_mode::read> buf_accessor(buf);
+        sycl::host_accessor buf_accessor(buf, sycl::read_only);
         T* actual_data = buf_accessor.get_pointer();
 
         EXPECT_TRUE(actual_data == expected_data, "wrong effect of iterator's method get_buffer");
@@ -50,7 +50,7 @@ main()
     std::size_t size = 1000;
     sycl::buffer<std::uint32_t> buf{size};
     test_buffer_wrapper test{};
-    sycl::host_accessor<DataType, 1, sycl::access_mode::read> buf_accessor(buf);
+    sycl::host_accessor buf_accessor(buf, sycl::read_only);
     auto data_ptr = buf_accessor.get_pointer();
 
     test(oneapi::dpl::begin(buf), oneapi::dpl::end(buf), data_ptr, size);
--- a/test/general/buffer_wrapper.pass.cpp
+++ b/test/general/buffer_wrapper.pass.cpp
@@ -34,7 +34,9 @@ struct test_buffer_wrapper
         EXPECT_TRUE(end - begin == size, "wrong effect of iterator's operator - iterator");
 
         auto buf = begin.get_buffer();
-        T* actual_data = sycl::host_accessor<T, 1, sycl::access_mode::read>(buf).get_pointer();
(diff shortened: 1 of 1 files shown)


test/general/buffer_wrapper.pass.cpp
```diff
@@ -48,7 +50,8 @@ main()
     std::size_t size = 1000;
     sycl::buffer<std::uint32_t> buf{size};
     test_buffer_wrapper test{};
-    auto data_ptr = sycl::host_accessor<std::uint32_t, 1, sycl::access_mode::read>(buf).get_pointer();
+    sycl::host_accessor buf_accessor(buf, sycl::read_only);
```
SergeyKopienko, line 53: In this line we use constructor as described at https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_interface_for_buffer_host_accessors :
```
  /* Available only when: (Dimensions > 0) */
  template <typename AllocatorT, typename TagT>
  host_accessor(buffer<DataT, Dimensions, AllocatorT>& bufferRef, TagT tag,
                const property_list& propList = {});
```

In the SYCL source code from DPCPP compiler it constructor from the file: https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/accessor.hpp, line 3019 :
```
  template <typename T = DataT, int Dims = Dimensions, typename AllocatorT,
            typename = detail::enable_if_t<IsSameAsBuffer<T, Dims>::value>>
  host_accessor(
      buffer<T, Dims, AllocatorT> &BufferRef, mode_tag_t<AccessMode>,
      const property_list &PropertyList = {},
      const detail::code_location CodeLoc = detail::code_location::current())
      : host_accessor(BufferRef, PropertyList, CodeLoc) {}
```

test/general/buffer_wrapper.pass.cpp
```diff
@@ -34,7 +34,9 @@ struct test_buffer_wrapper
         EXPECT_TRUE(end - begin == size, "wrong effect of iterator's operator - iterator");
 
         auto buf = begin.get_buffer();
-        T* actual_data = sycl::host_accessor<T, 1, sycl::access_mode::read>(buf).get_pointer();
```
MikeDvorskiy, line 37: 1) By following up to the issue #788 where we have  got `'no matching conversion for functional-style cast from 'sycl::buffer<unsigned int, dim, aligned_allocator<unsigned int>>' to 'sycl::host_accessor<const unsigned int, 1, sycl::access_mode::read>'`
It tells that we have a buffer parametrized by `unsigned int` and `sycl::host_accessor` parametrized  by `const unsigned int`. 
2) The proposed changes probably doesn't resolve the issue because the SYCL spec tells " using value_type = // const DataT for read-only accessors, DataT otherwise" ,  so we have got the same error...

3) Furthermore, looking into the SYCL spec I cannot see such constructor with a buffer and access mode, how it is proposed in the changes.

MikeDvorskiy (changes requested):
It looks the problem requires the further investigation....

test/general/buffer_wrapper.pass.cpp
SergeyKopienko: About (3) - please see my comments at https://github.com/oneapi-src/oneDPL/pull/793#discussion_r1104332954

PR #715, opened 2022-12-06
Changing clang-format check to only run on git diff delta

danhoeflinger:
This PR does the following: 
1) Change the clang-format github runner to check **only the delta changed lines** made by the current PR.  In the current main, this github runner checks the **entire file for any file which differs between the PR and the main branch**. 
2) Update our clang format to the newest version, and updates the style file to a modern group of settings which results in a very similar result to the existing combination of version /settings.
3) Change the clang-format github runner to compare the current branch head to the "merge base", or common ancestor with main head.  In the current main, the github runner checks the "merge commit", which flags any divergent files between the current PR's branch and main.  This means any PRs merged since the last rebase of the current branch will be included in the files reported on.  
4) Update github runner to newest version of ubuntu from 20.04.

Main goal and purpose:
Targets the github runner check to be focussed only on changes for which the PR's author is responsible.  Accomplish fewer false positive red X in our PRs for clang-format which need to be explained as unrelated, and make that check more meaningful when it does fail.

PR #796, opened 2023-02-14
Make CI job names more readable

dmitriy-sobolev:
Change mixed approach for naming jobs to "name=value,name_1=value_1,...,name_n=value_n" for values which need keys for clarity. 

Name before the changes:
`ubuntu-20.04_g++_backend=tbb_device=HOST_cxx17_release`
Name after the changes:
`ubuntu-20.04,cmplr=g++,backend=tbb,device=HOST,std=17,build_type=release`

Upd. The most important properties are shown first due to 28-symbol limit of the view panel:
`HOST,bknd=tbb,cmplr=g++,ubuntu-20.04,std=c++17,cfg=release`

Previous '_' separator might have confused c++ developers. For example, ones may get surprised to find 'tbb_device'.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -251,7 +251,7 @@ jobs:
           ctest --timeout %TEST_TIMEOUT% -C ${{ matrix.build_type }} --output-on-failure %ctest_flags%
 
   macos-testing:
-    name: be=${{ matrix.backend }},cmplr=${{ matrix.cxx_compiler }},os=${{ matrix.os }},std=${{ matrix.std }},cfg=${{ matrix.build_type }}
+    name: be=${{ matrix.backend }},device=HOST,cmplr=${{ matrix.cxx_compiler }},os=${{ matrix.os }},std=${{ matrix.std }},cfg=${{ matrix.build_type }}
     runs-on: ['${{ matrix.os }}']
     timeout-minutes: 60
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
             path: html
 
   linux-testing:
-    name: ${{ matrix.os }}_${{ matrix.cxx_compiler }}_backend=${{ matrix.backend }}_device=${{ matrix.device_type }}_cxx${{ matrix.std }}_${{ matrix.build_type }}
+    name: os=${{ matrix.os }},compiler=${{ matrix.cxx_compiler }},backend=${{ matrix.backend }},device=${{ matrix.device_type }},cxx=${{ matrix.std }},build_type=${{ matrix.build_type }}
     runs-on: ['${{ matrix.os }}']
     timeout-minutes: 60
     strategy:
@@ -180,7 +180,7 @@ jobs:
           ctest --timeout ${TEST_TIMEOUT} --output-on-failure ${ctest_flags}
 
   windows-testing:
-    name: ${{ matrix.os }}_${{ matrix.cxx_compiler }}_backend=${{ matrix.backend }}_device=${{ matrix.device_type }}_cxx${{ matrix.std }}_${{ matrix.build_type }}
+    name: os=${{ matrix.os }},compiler=${{ matrix.cxx_compiler }},backend=${{ matrix.backend }},device=${{ matrix.device_type }},cxx=${{ matrix.std }},build_type=${{ matrix.build_type }}
     runs-on: ['${{ matrix.os }}']
     timeout-minutes: 60
     strategy:
@@ -251,7 +251,7 @@ jobs:
           ctest --timeout %TEST_TIMEOUT% -C ${{ matrix.build_type }} --output-on-failure %ctest_flags%
 
   macos-testing:
-    name: ${{ matrix.os }}_${{ matrix.cxx_compiler }}_backend=${{ matrix.backend }}_cxx${{ matrix.std }}_${{ matrix.build_type }}
+    name: os=${{ matrix.os }},compiler=${{ matrix.cxx_compiler }},backend=${{ matrix.backend }},cxx=${{ matrix.std }},build_type=${{ matrix.build_type }}
     runs-on: ['${{ matrix.os }}']
     timeout-minutes: 60
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,7 +74,7 @@ jobs:
             path: html
 
   linux-testing:
-    name: ${{ matrix.backend }},device=${{ matrix.device_type }},cmplr=${{ matrix.cxx_compiler }},os=${{ matrix.os }},std=${{ matrix.std }},cfg=${{ matrix.build_type }}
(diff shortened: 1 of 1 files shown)


rarutyun (commented):
I like that change, but I would like to see what others are thinking.

This is a huge improvement to me. Thanks.

danhoeflinger (commented):
I also like this a lot.
The only downside I notice is that the names start to be clipped by the space allowed in github for checks.  
If we could shave a few characters without confusing matters that may be for the better.  Possible candidates which I think wouldn't cause too much confusion:
drop `os=` 
`compiler` -> `cmplr`

That would save us 6 characters and remove all of the relevant clipping of names for the current set.

danhoeflinger (approved):
LGTM, but you may want to get others' opinions still.

akukanov:
Is this the order we prefer to see? Given that the view panel may truncate a lot (for example, I see only `windows-2019_dpcpp_backend=...`), what would be the most important information to show? I would say it's the backend, the device, and the compiler (in that order, but any other order would still be better than using most of the space for OS name & version)

dmitriy-sobolev:
> Is this the order we prefer to see? Given that the view panel may truncate a lot (for example, I see only `windows-2019_dpcpp_backend=...`), what would be the most important information to show? I would say it's the backend, the device, and the compiler (in that order, but any other order would still be better than using most of the space for OS name & version)

Done. I also removed `backend=` therefore backend, device and compiler can fit 28 symbol limit of the view panel.

akukanov:
This seems good to me, but I guess more opinions need to be collected.

SergeyKopienko (dismissed):
LGTM

danhoeflinger (dismissed):
Yep, I like it.  LGTM

mmichel11 (dismissed):
LGTM.

These changes make the CI job names a lot more readable and easier to search for certain configurations.

danhoeflinger:
FYI, I believe clang-format is complaining because the branch is not rebased with current head of main, files which are different from this branch to the current head of main have suggested changes.  No action needs to be done here before merging.

Clang-format PR #715 is still awaiting approval.  However, I believe this check would still appear red even after that is merged if any of the suggestions are in the branch to branch diff delta lines.  It should be possible to change #715 to compare to the  nearest ancestor or merge base.  We can discuss that question there...

.github/workflows/ci.yml
```diff
@@ -72,7 +72,7 @@ jobs:
             path: html
 
   linux-testing:
-    name: ${{ matrix.os }}_${{ matrix.cxx_compiler }}_backend=${{ matrix.backend }}_device=${{ matrix.device_type }}_cxx${{ matrix.std }}_${{ matrix.build_type }}
+    name: be=${{ matrix.backend }},device=${{ matrix.device_type }},cmplr=${{ matrix.cxx_compiler }},os=${{ matrix.os }},std=${{ matrix.std }},cfg=${{ matrix.build_type }}
```
rarutyun, line 75: I slightly prefer to have `backend` instead of `be`. If people think otherwise, I am ready to approve it as well. Overall - big improvement. Please also rebase it on the last state of the main branch as Dan suggested

dmitriy-sobolev: I do not have strong opinion on that as well.
@danhoeflinger what do you think?

danhoeflinger: I think `be` is clear enough, especially after people get used to it.  I think characters come at a high cost here, so I'd keep it as it is if it were up to me.

julianmi (commented):
I like these changes a lot. In case we need to reduce the number of characters, I think we could drop `os=` since the OS strings are descriptive in my opinion and the `=` in `std`.

Why don't we report the device for macOS for consistency?

dmitriy-sobolev:
> I like these changes a lot. In case we need to reduce the number of characters, I think we could drop `os=` since the OS strings are descriptive in my opinion and the `=` in `std`.
> 
> Why don't we report the device for macOS for consistency?

I think the further reduction is no more relevant because: 
 - Everything already fits in "Show all checks panel" in the "Conversation" panel. 
 - Job names are limited to 28 symbols in the "Checks" tab and it is possible to show up to the first 3 arguments. It is possible to fit 4th argument sacrificing keys. I think it does not worth so. 

MacOS does not support CPU and GPU offloading so far, so the implied device is always HOST.

.github/workflows/ci.yml
rarutyun: @danhoeflinger It took me several to figure it out what be means (I had several variants). But I don't want to insist

rarutyun (approved):

julianmi:
> MacOS does not support CPU and GPU offloading so far, so the implied device is always HOST.

So, should we include the device anyways to make it more explicit?

dmitriy-sobolev:
> So, should we include the device anyways to make it more explicit?

It's done.

.github/workflows/ci.yml
danhoeflinger: Is there any *visible* place we can place a legend for the checks?  I'm not seeing a great place, but I don't know much about options available to admins for the repo.
I don't have a strong opinion here, `bknd` is a more explicit middleground but still kicks out the compiler from the "checks" view.

dmitriy-sobolev: I checked available tools which I have for my own forks. Looks like there are no such options.

I've seen another approach which solves that problem: create multiple workflow files, thus moving one parameter to a workflow name. This could be a backend for oneDPL. This will result in the following:

-  "Checks" tab will have multiple workflows
- "Show all checks" panel in the "Conversation" tab will list jobs like

```
dpcpp backend / device=CPU,cmplr=icpx,os=ubuntu-20.04,std=17,cfg=release
...
tbb backend / cmplr=g++,os=ubuntu-18.04,std=17,cfg=release
...
```
I think the current readability is not that bad for that change...

danhoeflinger: Thanks for looking at that.  I would vote to merge as it is and if it creates confusion, we could explore something like that in the future.

akukanov:
One more suggestion to try: how about moving device to the first place, and dropping the prefix? It will also give some space to change backend prefix to `bknd`.
It would look like this: `HOST,bknd=serial,cmplr=g++,os=ubuntu-20.04,std=17,cfg=release` and when truncated `HOST,bknd=serial,cmplr=g++,o...`

.github/workflows/ci.yml
```diff
@@ -74,7 +74,7 @@ jobs:
             path: html
 
   linux-testing:
-    name: ${{ matrix.os }}_${{ matrix.cxx_compiler }}_backend=${{ matrix.backend }}_device=${{ matrix.device_type }}_cxx${{ matrix.std }}_${{ matrix.build_type }}
+    name: ${{ matrix.device_type }},bknd=${{ matrix.backend }},cmplr=${{ matrix.cxx_compiler }},${{ matrix.os }},std${{ matrix.std }},cfg=${{ matrix.build_type }}
```
akukanov, line 77: If the overall length is not a limit, I suggest to change to `std=c++${{ matrix.std }}` (e.g. `std=c++17`) so that it is similar to the well-known compiler option.

Otherwise, the improvements look very good to me.

akukanov (approved):

PR #801, opened 2023-02-15
Improvements in utils_ranges_sycl.h

SergeyKopienko:


include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -517,9 +517,8 @@ struct __get_sycl_range
         auto buf = copy_direct(__first, __last, copy_direct_tag());
         buf = copy_back(__first, buf, copy_back_tag());
 
-        auto __p_buf = ::std::shared_ptr<oneapi::dpl::__internal::__lifetime_keeper<decltype(buf)>>(
-            new oneapi::dpl::__internal::__lifetime_keeper<decltype(buf)>(buf));
```
SergeyKopienko, line 521: First memory allocation at this ```new``` operator call

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -517,9 +517,8 @@ struct __get_sycl_range
         auto buf = copy_direct(__first, __last, copy_direct_tag());
         buf = copy_back(__first, buf, copy_back_tag());
 
-        auto __p_buf = ::std::shared_ptr<oneapi::dpl::__internal::__lifetime_keeper<decltype(buf)>>(
```
SergeyKopienko, line 520: Second memory allocation at this ```::shared_ptr``` constructor call.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -517,9 +517,8 @@ struct __get_sycl_range
...
-            new oneapi::dpl::__internal::__lifetime_keeper<decltype(buf)>(buf));
-        m_buffers.push_back(__p_buf);
```
SergeyKopienko, line 522: Extra addRef / releaseRef at this push_call.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -517,9 +517,8 @@ struct __get_sycl_range
...
+        using BufferType = oneapi::dpl::__internal::__lifetime_keeper<decltype(buf)>;
+        m_buffers.push_back(::std::make_shared<BufferType>(buf));
```
SergeyKopienko, line 521: Here we have only one memory allocation at whole ```::std::make_shared<BufferType>(buf)```
 - for details please see https://en.cppreference.com/w/cpp/memory/shared_ptr/make_shared
And call ```void push_back( T&& value )``` without extra ```addRef / releaseRef``` calls

test/support/utils.h
```diff
@@ -539,8 +539,7 @@ class Wrapper
     }
     Wrapper(Wrapper&& input)
     {
-        my_field = input.my_field;
-        input.my_field = nullptr;
```
MikeDvorskiy, line 543: I'm not sure that it (checking NULL) is  not used in the tests for some checks for a moved instance.
it should be double checked.

test/support/utils.h
```diff
@@ -524,12 +524,12 @@ class Wrapper
   public:
     Wrapper()
     {
-        my_field = ::std::shared_ptr<T>(new T());
+        my_field = ::std::make_shared<T>();
```
rarutyun, line 527: If we are doing some changes there let's use member-init-list for all constructors then.

SergeyKopienko: Fixed.

test/support/utils.h
SergeyKopienko: There are no changes in run-time:
https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr
10): ```shared_ptr( shared_ptr&& r ) noexcept;```

"10) Move-constructs a shared_ptr from r. After the construction, *this contains a copy of the previous state of r, r is empty and its stored pointer is null. The template overload doesn't participate in overload resolution if Y* is not implicitly convertible to (until C++17)compatible with (since C++17) T*."

SergeyKopienko: Also checked in debugger under MSVC compiler.

test/support/utils.h
```diff
@@ -552,8 +551,7 @@ class Wrapper
     Wrapper&
     operator=(Wrapper&& input)
     {
-        my_field = input.my_field;
-        input.my_field = nullptr;
+        my_field = ::std::move(input.my_field);
```
SergeyKopienko, line 554: https://en.cppreference.com/w/cpp/memory/shared_ptr/operator%3D
(2) ```shared_ptr& operator=( shared_ptr&& r ) noexcept;```
"Move-assigns a shared_ptr from r. After the assignment, *this contains a copy of the previous state of r, and r is empty. Equivalent to shared_ptr<T>(std::move(r)).swap(*this)."

test/support/utils.h
```diff
@@ -523,24 +523,23 @@ class Wrapper
 {
   public:
     Wrapper()
+        : my_field(::std::make_shared<T>())
```
SergeyKopienko, line 526: This constructor is required at least for test uninitialized_construct.pass.cpp

rarutyun (approved):

PR #802, opened 2023-02-15
Fix broken table markup

dcbenito:
Fixed table.


ValentinaKats (approved):
LGTM. It fixes the issue in documentation.

Issue #787, opened 2023-02-09
ONEAPI_DEVICE_SELECTOR inplace of SYCL_DEVICE_FILTER


abagusetty:
I was trying to use the compiler version and get into runtime issues with tests since this environment variable is set into the cmake-setup and prevents being over-ridden/unset in a bash environment

```[bash]
1/1 Test #14: sycl_iterator.pass ...............Subprocess aborted***Exception:   0.77 sec

WARNING: The enviroment variable SYCL_DEVICE_FILTER is deprecated. Please use ONEAPI_DEVICE_SELECTOR instead.
For more details, please refer to:
https://github.com/intel/llvm/blob/sycl/sycl/doc/EnvironmentVariables.md#oneapi_device_selector

terminate called after throwing an instance of 'sycl::_V1::exception'
  what():  ONEAPI_DEVICE_SELECTOR cannot be used in conjunction with SYCL_DEVICE_FILTER
```
Occurances:
```[bash]
./make/Makefile.common:58:        export SYCL_DEVICE_FILTER=$(device_type)
./test/CMakeLists.txt:74:            set(SYCL_DEVICE_FILTER *:acc)
./test/CMakeLists.txt:77:            set(SYCL_DEVICE_FILTER ${ONEDPL_DEVICE_BACKEND}:${ONEDPL_DEVICE_TYPE})
./test/CMakeLists.txt:79:        set_tests_properties(${_test_name} PROPERTIES ENVIRONMENT SYCL_DEVICE_FILTER=${SYCL_DEVICE_FILTER})
```

abagusetty:
Additional Context: I was using the latest compiler built with the hash https://github.com/intel/llvm/commit/1ed2378b6072a0f70c55f4105339b527c7015a75

dmitriy-sobolev:
Hello, @abagusetty. Thanks for highlighting that issue. 

`SYCL_DEVICE_FILTER` variable is deprecated and I am going to replace it with `ONEAPI_DEVICE_SELECTOR` (only for recent releases of `Intel® oneAPI DPC++/C++ Compiler` supporting it). I am investigating how to override an environment variable set via `set_test_properties` after configuring cmake and before running tests.
 
Meanwhile, could you describe your case in more details? What is the value of `ONEAPI_DEVICE_SELECTOR` you set? What is the cmake command you ran?

[...]

PR #803, opened 2023-02-15
Use new device selection variable

dmitriy-sobolev:
This PR addresses #787 with the following changes:
- Use `ONEAPI_DEVICE_SELECTOR` instead of the deprecated `SYCL_DEVICE_FILTER` with newer compiler versions
- Allow oneDPL testing with `cuda` and `hip` device backends
- Clarify readme for `ONEDPL_DEVICE_TYPE` and `ONEDPL_DEVICE_BACKEND` variables

The issue with overriding device selection variable set in cmake still is not solved. I believe it does not have big value, because basic cases (setting a device type and backend) can be configured with `ONEDPL_DEVICE_TYPE` and `ONEDPL_DEVICE_BACKEND`.

Additional change:

- Changed messages: "add" to "adding", "use" to "using", therefore the messages are aligned.


CMakeLists.txt
```diff
@@ -72,12 +72,16 @@ string(TOUPPER ${ONEDPL_BACKEND} ONEDPL_BACKEND)
 message(STATUS "Using parallel policies with ${ONEDPL_BACKEND} backend")
 string(TOLOWER ${ONEDPL_BACKEND} ONEDPL_BACKEND)
 
+# CMAKE_CXX_COMPILER and CMAKE_CXX_COMPILER_VERSION cannot be used because
+# CMake 3.19 and older will detect IntelLLVM compiler as CLang with CLang-specific version, see https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_COMPILER_ID.html
+if (CMAKE_CXX_COMPILER MATCHES ".*(dpcpp-cl|dpcpp|icx-cl|icpx)(.exe)?$")
```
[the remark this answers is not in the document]
danhoeflinger: Nope, I'm wrong,  nevermind.

CMakeLists.txt
```diff
@@ -72,12 +72,16 @@ string(TOUPPER ${ONEDPL_BACKEND} ONEDPL_BACKEND)
...
+    set(DPCPP_COMPLER TRUE)
+    execute_process(COMMAND ${CMAKE_CXX_COMPILER} --version OUTPUT_VARIABLE DPCPP_COMPILER_VERSION_RAW)
```
danhoeflinger, line 79: For these variable names `DPCPP_COMPILER*`.  Since DPCPP is deprecated in favor of `icpx -fsycl`, would it be better to name these differently?  Possibly `ICPX_FSYCL_COMPILER*`.

dmitriy-sobolev: I am hesitant about `ICPX_FSYCL_COMPILER`, because `fsycl` is just an option, it does not determine a compiler and there is another relevant driver: `icx-cl`.
What do you think about `INTEL_LLVM_COMPILER` compiler instead? It will be aligned with the compiler id provided by new cmake versions.

Meanwhile, there is also a `clang++` driver. It is available when the compiler is built from [sources](https://github.com/intel/llvm). But cmake detects it as clang (according to my experiments).

danhoeflinger: Yes, `INTEL_LLVM_COMPILER*` is a better name.

danhoeflinger (approved):
LGTM, the PR description no longer matches the code, (`ONEAPI_DEVICE_SELECTOR` vs `ONEAPI_DEVICE_FILTER`), but the result is the correct code change.

dmitriy-sobolev:
I'll take a closer look at what happens in https://github.com/oneapi-src/oneDPL/blob/main/cmake/windows-dpcpp-toolchain.cmake. This may require changes as well.

dmitriy-sobolev:
> I'll take a closer look at what happens in https://github.com/oneapi-src/oneDPL/blob/main/cmake/windows-dpcpp-toolchain.cmake. This may require changes as well.

It was handled in #827, where `icpx`, `icx` and `icx-cl` drivers support was added.

dmitriy-sobolev:
@danhoeflinger, could you take a look at this PR once again? I rebased it after the changes done in #827.

PR #805, opened 2023-02-21
Split test sycl_iterator.pass into multiple tests

SergeyKopienko:
In this PR we split test ```sycl_iterator.pass.cpp``` into multiple tests:
- ```sycl_iterator_buf1.pass.cpp``` : all tests which starts through ```test1buffer``` call;
- ```sycl_iterator_buf2.pass.cpp``` : all tests which starts through ```test2buffers``` call;
- ```sycl_iterator_buf3.pass.cpp``` : all tests which starts through ```test3buffers``` call.

Also renamed some test variables to avoid spell errors in these tests.


SergeyKopienko:
@dmitriy-sobolev thank you for your commit https://github.com/oneapi-src/oneDPL/pull/805/commits/b0796f8e0883e9e93f7674145d41a7a5cd6ba51b

PR #807, opened 2023-02-21
Adding missing include for fwd declaration of transform_*_scan

danhoeflinger:
Adding missing include to give `glue_numeric_impl.h` access to forward declaration of `transform_inclusive_scan()` and `transform_exclusive_scan()`  found in `glue_numeric_defs.h` because these functions are used before they are defined.

Found this attempting to build a usage of `exclusive_scan` when compiling with `tbb` or `omp` backend with a parallel execution policy.  It resulted in a compiler error until this was added.  With `dpcpp` backend this is not an issue because when `_ONEDPL_HETERO_BACKEND` is defined, it is already included through this chain:

`glue_numeric_impl.h` -> `hetero/algorithm_impl_hetero.h` -> `../parallel_backend.h` -> 
`hetero/dpcpp/parallel_backend_sycl.h` -> `utils_ranges_sycl.h` -> `../../glue_numeric_defs.h`

### DIFF
--- a/include/oneapi/dpl/pstl/glue_numeric_impl.h
+++ b/include/oneapi/dpl/pstl/glue_numeric_impl.h
@@ -27,6 +27,7 @@
 
 #include "numeric_fwd.h"
 #include "execution_impl.h"
+#include "glue_numeric_defs.h"
 
 namespace oneapi
 {


timmiesmith (approved):
This looks ok to me.

akukanov:
I think that the issue might be in some other place.
How does it happen that `glue_numeric_impl.h` is included before `glue_numeric_defs.h`?

danhoeflinger:
> I think that the issue might be in some other place. How does it happen that `glue_numeric_impl.h` is included before `glue_numeric_defs.h`?

Using a TBB backend, my cpp file includes: 
```
```

Which results in this path to glue_numeric_impl.h:
`oneDPL/include/oneapi/dpl/algorithm` -> `oneDPL/include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h` ->
`oneDPL/include/oneapi/dpl/internal/../pstl/glue_numeric_impl.h`

and `glue_numeric_defs.h` is never included without this change


In general, don't we try to make these headers self sufficient and not rely upon the context from which they were included?  It seems that even if `glue_numeric_defs.h` should be included elsewhere, it is needed for this header alone because of the usage of `transform_*_scan` prior to its definition.

akukanov:
Then I think the bug is in `<oneapi/dpl/algorithm>` as it includes implementation header for scan without first including the forward declarations. Why, I wonder, by-segment scan and reduce are included in `<oneapi/dpl/algorithm>` instead of `<oneapi/dpl/numeric>`?

danhoeflinger:
> Then I think the bug is in `<oneapi/dpl/algorithm>` as it includes implementation headers (for scan particularly) without first including the forward declarations. Why, I wonder, by-segment scan and reduce are included in `<oneapi/dpl/algorithm>` instead of `<oneapi/dpl/numeric>`?

That is a good question.  It seems like it fits far better in `<oneapi/dpl/numeric>`, perhaps it was decided at some point to place all "extensions" in `<oneapi/dpl/algorithm>`.  Sadly, the git history does not go back far enough to provide insight.  

As it stands, including `<oneapi/dpl/algorithm>` effectively gets you a partial `<oneapi/dpl/numeric>`, which is complete enough to use when using `dpcpp` backend for `oneapi::dpl::exclusive_scan` (and probably other APIs).

Do you think it is worth potentially breaking some existing user code to move the inclusion of the by-segment implementation to `<oneapi/dpl/numeric>` where they seem to belong?

In any case, I would still be in favor of adding the include currently in this PR to make this header more self sufficient.

akukanov:
> Using a TBB backend, my cpp file includes:
> 
> ```
> #include <oneapi/dpl/execution>
> #include <oneapi/dpl/algorithm>
> #include <oneapi/dpl/iterator> 
> ```

As we have discussed in person, this code is not quite correct, as the use of `exclusive_scan` requires `<oneapi/dpl/numeric>`.

The question of the right header for scan-by-segment and reduce-by-segment algorithms remains; we need to discuss it in the team and decide what's the best way to go.

SergeyKopienko (changes requested):
I think that better at first to rebase this branch from current state of main branch.
There are a lot of changes.
And also in this case we sill not have CI errors in checks of this PR because sycl_iterator.pass.cpp already split in main branch into 5 separate files.

danhoeflinger:
As discussed, closing this as this is a result of incorrect usage (using `oneapi::dpl::exclusive_scan` without including `<oneapi/dpl/numeric>`) .

PR #809, opened 2023-02-22
Fix conflicting argument names

rarutyun:
The guess is `__in` is some conflicting name with macro that defined in Microsoft header. We don't have `__in` as the name in other places

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -414,7 +414,7 @@ struct __parallel_transform_scan_single_group_submitter
     template <typename _KernelName, typename _Policy, typename _InRng, typename _OutRng, typename _InitType,
               typename _BinaryOperation, typename _UnaryOp>
     static auto
-    __launch_dynamic_bounds_scan(const _Policy& __policy, _InRng __in, _OutRng __out, ::std::size_t __n,
+    __launch_dynamic_bounds_scan(const _Policy& __policy, _InRng __in_rng, _OutRng __out_rng, ::std::size_t __n,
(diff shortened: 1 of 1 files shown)


SergeyKopienko (approved):
LGTM

danhoeflinger (approved):
LGTM,  
I attempted to build on main to confirm and see the compiler errors windows, then checked this branch on windows and this does in fact resolve the compiler errors for `sort_ranges_sycl.pass`.  From a code perspective it looks harmless.

danhoeflinger:
For more information on what we were conflicting with:  
In `MSVC\14.34.31933\include\sal.h` line 2647:
 
`#define __in                                                     _SAL1_Source_(__in, (), _In_)`

It looks like this is translating to "The Microsoft source-code annotation language (SAL)" annotations signifying that a parameter is read only input data to a function.  They have some code analysis tools which utilize these annotations.

https://learn.microsoft.com/en-us/cpp/code-quality/understanding-sal?view=msvc-170

timmiesmith:
https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/ms235402(v=vs.100)?redirectedfrom=MSDN documents that '__in', '__out' and other keywords are reserved. Hat tip to @adamfidel for finding the link.

adamfidel:
> https://learn.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/ms235402(v=vs.100)?redirectedfrom=MSDN documents that `__in`, `__out` and other keywords are reserved. Hat tip to @adamfidel for finding the link.

That link describes the newer SAL syntax of `_In_` and `_Out_`. The old syntax uses `__in` and `__out` which is apparently still reserved in some versions of MSVC. Here is an older document describing it: https://learn.microsoft.com/en-us/archive/blogs/michael_howard/a-brief-introduction-to-the-standard-annotation-language-sal

PR #810, opened 2023-02-22
Fix forwarding references in single-group scan

adamfidel:
This PR makes the following changes:
- Pass ranges as forwarding references and use `std::forward` where necessary
- Create `all_view`s in the outermost function call and pass them by forwarding reference instead of passing the `__range_holder`


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -598,15 +598,15 @@ __pattern_transform_scan_single_group(_ExecutionPolicy&& __exec, _InRng&& __in_r
...
+                                                                           std::forward<_OutRng>(__out_rng), __n,
+                                                                           __init, __binary_op, __unary_op);
             else
                 return __parallel_transform_scan_single_group_submitter<
                     _Inclusive::value, __num_elems_per_item, __wg_size,
-                    /* _IsFullGroup= */ false>::__launch_static_bounds_scan(__exec, __in_rng.all_view(),
-                                                                            __out_rng.all_view(), __n, __init,
-                                                                            __binary_op, __unary_op);
+                    /* _IsFullGroup= */ false>::__launch_static_bounds_scan(__exec, std::forward<_InRng>(__in_rng),
+                                                                            std::forward<_OutRng>(__out_rng), __n,
```
danhoeflinger, line 608: `std` -> `::std` for these and later on as well.

adamfidel: Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -414,7 +414,7 @@ struct __parallel_transform_scan_single_group_submitter
     template <typename _KernelName, typename _Policy, typename _InRng, typename _OutRng, typename _InitType,
               typename _BinaryOperation, typename _UnaryOp>
     static auto
-    __launch_dynamic_bounds_scan(const _Policy& __policy, _InRng __in_rng, _OutRng __out_rng, ::std::size_t __n,
+    __launch_dynamic_bounds_scan(const _Policy& __policy, _InRng&& __in_rng, _OutRng&& __out_rng, ::std::size_t __n,
```
SergeyKopienko, line 417: Up to me - I don't understand why you are use now universal reference here, because there are no std::forward operations inside this function with params ```__in_rng``` and ```__out_rng```

adamfidel: The forwarding references are here for two reasons:
1. Avoiding a copy of the range object if the calling side passes it as a glvalue
2. Consistency with the rest of the structs in this file (e.g., `__parallel_for_submitter`, `__parallel_partial_sort_submitter`, `__parallel_scan_submitter`)

SergeyKopienko: OK

danhoeflinger (approved):
LGTM

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -131,21 +131,23 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
         }
     }
     else
     {
         oneapi::dpl::__par_backend_hetero::__pattern_transform_scan_multi_group(
-            std::forward<_ExecutionPolicy>(__exec), __buf1, __buf2, __unary_op, __init, __binary_op, _Inclusive{})
+            ::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(), __buf2.all_view(), __unary_op, __init,
+            __binary_op, _Inclusive{})
             .wait();
     }
```
akukanov, line 152: As far as I can see, there is no difference between lines 140-143 and 148-151. Can the conditions be reorganized to avoid duplication? For example, like in this pseudocode:
```
if constexpr (can use group scan) {
    if (runtime conditions to use a single group) {
        use a single group;
        return;
    }
}
use all groups;
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
rarutyun, line 417: We are not forwarding in this function and as far as I can see, we cannot.. Does it make sense to accept the function argument by const lvalue ref? and by non-const lvalue ref for output range? This is also relevant for 468 line

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
adamfidel: Thanks! I couldn't figure out how to combine the compile-time and runtime conditions in a way that avoids code duplication. But using an early return as you suggest makes this a lot cleaner.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: It's possible to change the input and output ranges to const refs and I don't believe it would make a semantic difference. But we cannot change the output range to a non-const lvalue ref because the ranges are being created and passed as an rvalue here: https://github.com/oneapi-src/oneDPL/blob/1854cb66f4bd141c664f9198ce526205d59cd8eb/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h#L134

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -125,29 +125,24 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
-        }
-        else
-        {
-            oneapi::dpl::__par_backend_hetero::__pattern_transform_scan_multi_group(
-                std::forward<_ExecutionPolicy>(__exec), __buf1, __buf2, __unary_op, __init, __binary_op, _Inclusive{})
-                .wait();
+                oneapi::dpl::__par_backend_hetero::__pattern_transform_scan_single_group(
+                    ::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(), __buf2.all_view(), __n, __unary_op, __init,
+                    __binary_op, _Inclusive{})
+                    .wait();
```
akukanov, line 136: Please check clang-format recommendations for this block.

adamfidel: Fixed.

julianmi (commented):
LGTM.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
rarutyun: How output range can be `const`?

akukanov: I agree that we likely overuse forwarding references, and not only in this place. However, I'm afraid if we take the output by lvalue reference here, we would need to do the same in the caller, and then possibly everywhere up the call stack. Do we want to? 
Or maybe the right questions would be: is there any significant downside of using forwarding references even in case the parameters are not forwarded down the stack? And if yes, what needs to be done to make it right?

rarutyun: > However, I'm afraid if we take the output by lvalue reference here, we would need to do the same in the caller, and then possibly everywhere up the call stack. Do we want to?

It's not necessarily so. When we work with forwarding references the caller should forward that to the function it calls (if it's further unused on the caller side, of course). The callee is allowed to choose how they take the parameter so, it is not necessarily forwarding reference by any other (or better to say, "right") way. The same situation with execution policy. We forward it down to the stack but the backend takes it how it wills.

> Or maybe the right questions are: is there any significant downside of using forwarding references even in case the parameters are not forwarded down the stack? And if yes, what needs to be done to make it right?

I don't see any downsides except the code looks "suspicious". I would try to do forwarding as much as we can (of course, in correct way) and decide how to accept that on the level where this very argument is used. It should allow us to do not modify the upper-level (stack-wise) code when we want to change something in the code that uses ranges and/or execution policies

adamfidel: @rarutyun Do you have any objections with keeping the input and output ranges as forwarding references?

adamfidel:
Closing this PR, since all of its changes have been incorporated into a new PR https://github.com/oneapi-src/oneDPL/pull/853.

PR #811, opened 2023-02-22
Improve utility functions in the SYCL backend

akukanov:
This patch reorganizes and improves various utility functions spread over multiple places in the SYCL backend.

* `__max_local_allocation_size` is renamed to `__adjust_to_local_mem_size`, with improved semantics: the second argument is the size of local memory used by a work item, the third argument is the work group size
* `__max_compute_units` is slightly improved (based on SYCL specification for `sycl::info::device::max_compute_units`), its usage is made consistent across the code (via renaming local variables)
* In `execution_sycl_defs.h`, SFINAE helpers are simplified via the use of C++17 features
* SYCL runtime information helpers are moved from `execution_sycl_defs.h` to `parallel_backend_sycl_utils.h`
* `__dpl_bit_cast`, `__dpl_bit_flor`, `__dpl_ceiling_div` are moved from the SYCL backend to generic utilities in `pstl/utils.h"


include/oneapi/dpl/pstl/utils.h
```diff
@@ -567,6 +576,62 @@ struct __lifetime_keeper : public __lifetime_keeper_base
...
+#else
+    _Dst __result;
+    ::std::memcpy(&__result, &__src, sizeof(_Dst));
+    return __result;
+#endif
+}
+
+// The max power of 2 not exceeding the given value, same as C++20 std::bit_floor
+template <typename _T>
+::std::enable_if_t<::std::is_integral<_T>::value && ::std::is_unsigned<_T>::value, _T>
```
rarutyun, line 604: I believe `is_unsigned` implies `is_integral`. And also we could use `_v` API.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -567,6 +576,62 @@ struct __lifetime_keeper : public __lifetime_keeper_base
...
+__dpl_bit_floor(_T __x) noexcept
+{
+    if (__x == 0)
+        return 0;
+#if __cpp_lib_int_pow2 >= 202002L || _ONEDPL___cplusplus >= 202002L && __has_include(<bit>)
+    return ::std::bit_floor(__x);
+#elif _ONEDPL_BACKEND_SYCL
+    // Use the count-leading-zeros function
+    return 1 << (sycl::clz(_T{0}) - sycl::clz(__x) - 1);
```
rarutyun, line 613: is not is better to use:

```suggestion
    return 1 << (sizeof(_T)) - sycl::clz(__x) - 1);
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -567,6 +576,62 @@ struct __lifetime_keeper : public __lifetime_keeper_base
...
+    if constexpr (sizeof(_T) > 4) __x |= (__x >> 32);
+    __x += 1; // Now it equals to the next greater power of 2, or 0 in case of wraparound
+    return (__x == 0) ? 1 << (sizeof(_T) * 8 - 1) : __x >> 1;
+#endif
+}
+
+// rounded up result of (__number / __divisor)
+template <typename _T1, typename _T2>
+constexpr auto
+__dpl_ceiling_div(_T1 __number, _T2 __divisor) -> decltype((__number - 1) / __divisor + 1)
```
rarutyun, line 630: do we really need trailing `decltype`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -20,15 +20,10 @@
...
-#if _ONEDPL___cplusplus >= 202002L && __has_include(<bit>)
-#    include <bit>
-#else
-#    include <cstring> // memcpy
-#endif
-
 #include "../../iterator_impl.h"
 
 #include "sycl_defs.h"
+#include "execution_sycl_defs.h"
```
rarutyun, line 26: missing includes: `<algorithm>` `<cstddef>`, `<string>`, `<cstdint>`

include/oneapi/dpl/pstl/utils.h
akukanov: `sizeof` returns the number of bytes, while we need the number of bits. I'd need to include `<numeric_limits>` etc. to make it right. Is it worth it? I guess `clz` maps efficiently to the hardware.

include/oneapi/dpl/pstl/utils.h
akukanov: From https://en.cppreference.com/w/cpp/types/is_unsigned:
> **If T is an arithmetic type**, provides the member constant value equal to true if T(0) < T(-1): this results in true for the unsigned integer types and the type bool and in false for the signed integer types and **the floating-point types**.

include/oneapi/dpl/pstl/utils.h
akukanov: Seemingly not; thanks.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
akukanov: Well, except for `<algorithm>` for `std::min`, all the rest of standard types are used in consistence with the SYCL spec - so it should be safe to assume that the necessary headers are included through the SYCL headers. Is it necessary to duplicate? In the lines 77-80, `std::vector` is used; should we also explicitly include `<vector>`?

I can change most or even all these std types to `auto` ("almost always auto", yeah) to formally have no need to include extra headers. Do you think it would be better?

rarutyun: I think so. The potential problem could be after we refactor the code we might end up depending on the include order. I hope that sycl includes everything but nobody prevents them to refactor their code (unless headers are explicitly listed in the SYCL spec). I prefer to do headers self-contained because it's just safer.

include/oneapi/dpl/pstl/utils.h
rarutyun: > I guess clz(0) can be optimized by compilers and/or maps efficiently to the hardware.

Probably but it's still runtime. Why should we rely on the optimization and "constructing" initialized object when all we need is know at compile-time.

Why `<limits>`? Probably just `sizeof(T) * 8`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
akukanov: You understand that technically `auto` changes nothing with regard to headers being self-contained, right? The functionality does not change, no new headers are included, and we still rely on SYCL implementation to do the proper job (while "nobody prevents them to refactor their code"). The only thing is that we **formally** do not use standard types. meanwhile losing the information on the return types of the routines.

julianmi (commented):
I think these changes are very useful and improve the project's structure.

I only have one general comment: Maybe it's just me but the "adjust" in `__adjust_to_local_mem_size` sounds to me like we would change `__wg_size` and not return it. I think we typically use it to adjust it so why don't we update `__wg_size` directly? And should we rename it to `__adjust_work_group_to_local_mem_size` or something similar to make explicit what this routine adjusts?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
rarutyun: Probably with auto you are right. If they don't include proper file, the code with auto doesn't work at all. I also understand what you mean: when SYCL returns some type it should have it's definition, on other words, include a proper header. Otherwise, it cannot be used. (Let us not to go in forward declaration details).

What I am struggling with is how to review the code? we have explicit usage of types in that. What should we do? Look it that's what we are using and distinguish it from types that other framework returns?

akukanov: Yes, I understand this concern. I think a careful reviewer can recognize where the use of certain types (standard or not) is associated with some API (and so implicitly covered by an already included header) versus where the types are used "on their own" and so a header is needed. But definitely it's extra effort, and some potential issues might be missed.

My viewpoint is, though, that a probability of something going wrong some time in the future is always non-zero, and trying to make the code resilient to future changes and possible mistakes is not always worth it. We need to make sure the code is free of actual errors, while robustness against potential future errors is a second-order concern at best.

In this very case, absence of standard headers is not an error because the standard types are used in the context of SYCL API.

akukanov:
@julianmi, thanks; I tried to further improve the semantics of the function (and renamed it again along the way). So now the WG size parameter is optional, and if not set (0 by default), it is obtained via `__max_work_group_size`.

julianmi (commented):
I think the naming and interface are much clearer now. The other changes look good to me. There are some formatting issues, but I think some of them are debatable.

rarutyun (dismissed):

akukanov:
Rebased to the top of main, resolved the conflict in `parallel_backend_sycl_reduce.h`. No other changes made.

rarutyun (approved):

PR #813, opened 2023-02-22
Updated table type and updated 'overview' to 'introduction'.

dcbenito:
Updated the table type in tested_standard_cpp_api. 
Updated the 'overview' to 'introduction' and updated the index/conf files.

### DIFF
--- a/documentation/library_guide/api_for_sycl_kernels/tested_standard_cpp_api.rst
+++ b/documentation/library_guide/api_for_sycl_kernels/tested_standard_cpp_api.rst
@@ -62,397 +62,802 @@ The printed result is:
 Tested Standard C++ API Reference
 =================================
 
-===================================== ========== ========== ==========
-C++ Standard API                      libstdc++  libc++     MSVC
-===================================== ========== ========== ==========
-``std::swap``                         Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
-``std::lower_bound``                  Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
-``std::upper_bound``                  Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
-``std::binary_search``                Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
-``std::equal_range``                  Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
-``std::tuple``                        Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
-``std::pair``                         Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
-``std::reference_wrapper``            Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
-``std::ref/cref``                     Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
-``std::divides``                      Tested     Tested     Tested
-------------------------------------- ---------- ---------- ----------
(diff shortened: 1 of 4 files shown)


documentation/library_guide/conf.py
```diff
@@ -37,7 +37,7 @@
 author = 'Intel'
 
 # The full version, including alpha/beta/rc tags
-release = '2022.0.0'
+release = '2023.0'
```
dcbenito, line 40: This was a mistake, it should be 2022.1 (I think?) for our next release.

akukanov:
Can I ask what's the motivation for table style change?

One of the great benefits of RST is that the file is well readable as-is, as a text, and so is easier to maintain. The current table style is aligned with that, the content is easily recognizable as a table. The proposed text style is not.

dcbenito:
The motivation was that the original table is very prone to errors due to whitespace, whereas the new style is not as human-readable, but less prone to mistakes. If you all dislike the new version, I don't mind switching back to the old style.

akukanov:
I see. You are right that the table formatting in the text requires extra care - for example, when a column becomes not wide enough. Personally I still prefer it more, but I rarely edit it myself, so I would ask others for their preferences.

documentation/library_guide/api_for_sycl_kernels/tested_standard_cpp_api.rst
```diff
@@ -62,397 +62,802 @@ The printed result is:
 Tested Standard C++ API Reference
```
SergeyKopienko, line 62: Up to me - the old style of this table is more readable, so I prefer the old style.
I think that ability to read "as table" is more important in this file than then difficulties to edit it.

dcbenito: @ValentinaKats what do you think? It sounds like we have more votes for the old style.

ValentinaKats: I prefer new style more because it's less error-prone. With old style any extra "*" symbol leads to the table not being generated and usually it takes us time to notice it. 
Is there a way to enable some kind of syntax checks for rst pages?

akukanov: In this very table, I do not see a single "*" symbol :)

The very first column can be extended for more space, it that helps. But overall, thanks to the aligned dashed lines between the table rows, it's easy to observe that a new or edited row does not fit some of the column size.

Do not get me wrong, I am not saying that each table should be reformatted to use the simple style. But my preference is to keep it looking like a table, if it is simple enough for that. This exact table is simple.

Also note that it can be prepended with the `.. table::` directive that allows setting e.g. relative column width.

akukanov: >... leads to the table not being generated and usually it takes us time to notice it.

I thought our CI has a job do generate the documentation from RSTs, which should fail if there are errors. Do I miss something?

dcbenito: When I build my docs, I check for errors. There were none when I published to IDZ (the intel website), so a change was introduced after I published for the 2022.0 release. I don't mind reverting back to the old table, the new version was suggested by a coworker after I spent a bunch of time trying to find/fix/rebuild the docs. 

Let's close this PR with no changes and revisit changing the tables later if we need to.

PR #815, opened 2023-02-24
Disable subgroup load/stores in single-group scan

adamfidel:
Validation of [oneDPL-2022.1.0-rc1](https://github.com/oneapi-src/oneDPL/releases/tag/oneDPL-2022.1.0-rc1) failed because subgroup load/store is not available in the tested configuration. It is not clear when it is possible to use subgroup load/store, so this PR disables that pathway for now.


julianmi (commented):
Disabling these sub-group operations for now makes sense to me.

timmiesmith (approved):
LGTM

PR #821, opened 2023-02-28
Catch incorrect Kernel name generation in tests

SergeyKopienko:
The main goal of this PR - to detect errors by tests like fixed in PRs (Kernels names generation in algorithms):
- https://github.com/oneapi-src/oneDPL/pull/819
- https://github.com/oneapi-src/oneDPL/pull/818
- https://github.com/oneapi-src/oneDPL/pull/823
- https://github.com/oneapi-src/oneDPL/pull/820
- https://github.com/oneapi-src/oneDPL/pull/816

For this purpose we added new CMake option ```ONEDPL_TEST_EXPLICIT_KERNEL_NAMES``` with the next available values:
- ```AUTO``` (default state)- usage of Kernel names in tests depends on ```-fsycl-unnamed-lambda``` / ```-fno-sycl-unnamed-lambda``` compiler options : if unnamed lambda is used then Kernel's names is not used and vice versa;
- ```ALWAYS``` - always use Kernel names in tests.

Examples of usage ```ONEDPL_TEST_EXPLICIT_KERNEL_NAMES``` in CMake command line :
- ```-DONEDPL_TEST_EXPLICIT_KERNEL_NAMES=AUTO```
- ```-DONEDPL_TEST_EXPLICIT_KERNEL_NAMES=ALWAYS```

**By default this mean if we start tests with unnamed lambda support, then Kernel's names isn't really used in tests code.
If we start tests without lambda support, then Kernel's names is used in tests code as before.**

In C++ code we added the definition of the new macro ```TEST_EXPLICIT_KERNEL_NAMES``` :
```
#ifndef TEST_EXPLICIT_KERNEL_NAMES
#    if __SYCL_UNNAMED_LAMBDA__
#        define TEST_EXPLICIT_KERNEL_NAMES 0
#    else
#        define TEST_EXPLICIT_KERNEL_NAMES 1
#    endif // __SYCL_UNNAMED_LAMBDA__
#endif // !TEST_EXPLICIT_KERNEL_NAMES
```

Now usage of Kernel names in tests depends on ```TEST_EXPLICIT_KERNEL_NAMES``` macro state:
```
namespace TestUtils
{
	template <typename KernelName = oneapi::dpl::execution::DefaultKernelName, typename Arg>
	inline auto
	make_device_policy(Arg&& arg)
	{
	#if TEST_EXPLICIT_KERNEL_NAMES
		return oneapi::dpl::execution::make_device_policy<KernelName>(::std::forward<Arg>(arg));
	#else
		return oneapi::dpl::execution::make_device_policy(::std::forward<Arg>(arg));
	#endif // TEST_EXPLICIT_KERNEL_NAMES
	}
	
	#if _ONEDPL_FPGA_DEVICE
	template <unsigned int unroll_factor = 1, typename KernelName = oneapi::dpl::execution::DefaultKernelNameFPGA, typename Arg>
	inline auto
	make_fpga_policy(Arg&& arg)
	{
	#if TEST_EXPLICIT_KERNEL_NAMES
		return oneapi::dpl::execution::make_fpga_policy<unroll_factor, KernelName>(::std::forward<Arg>(arg));
	#else
		return oneapi::dpl::execution::make_fpga_policy<unroll_factor>(::std::forward<Arg>(arg));
	#endif // TEST_EXPLICIT_KERNEL_NAMES
	#endif // _ONEDPL_FPGA_DEVICE
}; // namespace TestUtils
```

Now in tests we should always call ```make_device_policy``` and ```make_fpga_policy``` functions from namespace ```TestUtils```.


dmitriy-sobolev:
It seems the PR changes the logic of naming kernels in the tests, so:
- Kernel names are automatically generated in "unnamed lambda" mode.
- The kernel names are manually set in "no unnamed lambda" mode. 

This case becomes not covered:
- The kernel names are manually set in "unnamed lambda" mode.

@SergeyKopienko, what do you think about preserving the third case? Could you add a switch to the testing harness which forces setting kernel names regardless of the lambda naming mode?

What do you think about necessity of `lambda_naming.pass` test after the changes?

SergeyKopienko:
> It seems the PR changes the logic of naming kernels in the tests, so:
> 
> * Kernel names are automatically generated in "unnamed lambda" mode.
> * Kernel names are manually set in "no unnamed lambda" mode.
> 
> This case becomes not covered:
> 
> * Kernel names are manually set in "unnamed lambda" mode.
> 
> @SergeyKopienko, what do you think about preserving the third case? Could you add a switch to the testing harness which forces setting kernel names regardless of the lambda naming mode?
> 
> What do you think about necessity of `lambda_naming.pass` test after the changes?

@dmitriy-sobolev good point, thanks!
Added macro ```TEST_USE_KERNEL_NAMES_IN_TESTS``` which can be setup from command line

SergeyKopienko:
@dmitriy-sobolev, about file ```lambda_naming.pass``` - I prefer to save it.

test/CMakeLists.txt
```diff
@@ -32,6 +32,22 @@ else()
     message(STATUS "oneDPL: -fno-fast-math is not supported by current compiler")
 endif()
 
+# usage of Kernel names in tests
+set(TestExplicitKernelNameVariants AUTO ALWAYS)
```
dmitriy-sobolev, line 36: What do you think about having "ON" and "OFF" options as some other existing configuration variables have? You can try https://cmake.org/cmake/help/v3.4/module/CMakeDependentOption.html#module:CMakeDependentOption.

test/CMakeLists.txt
```diff
@@ -32,6 +32,22 @@ else()
...
+set(TEST_EXPLICIT_KERNEL_NAMES AUTO CACHE STRING "Use Kernel names in tests")
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed: TEST_EXPLICIT_KERNEL_NAMES were renamed to ONEDPL_EXPLICIT_KERNEL_NAMES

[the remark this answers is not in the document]
SergeyKopienko: Fixed: ONEDPL_EXPLICIT_KERNEL_NAMES CMake option described in https://github.com/oneapi-src/oneDPL/blob/main/cmake/README.md

test/CMakeLists.txt
dmitriy-sobolev: You can try:
```
include(CMakeDependentOption)
cmake_dependent_option(TEST_EXPLICIT_KERNEL_NAMES "Explicitly providing kernel names in the tests" ON "ONEDPL_USE_UNNAMED_LAMBDA" OFF)
```

SergeyKopienko: Are we able in this case setup this property value in CMake params too ?

test/CMakeLists.txt
```diff
@@ -32,6 +32,22 @@ else()
...
+set_property(CACHE TEST_EXPLICIT_KERNEL_NAMES PROPERTY STRINGS ${TestExplicitKernelNameVariants})
+message(STATUS "Usage of Kernel names in tests: TEST_EXPLICIT_KERNEL_NAMES=${TEST_EXPLICIT_KERNEL_NAMES}")
```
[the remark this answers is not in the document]

test/support/test_config.h
```diff
@@ -70,6 +70,16 @@
...
 
+// The TEST_EXPLICIT_KERNEL_NAMES macro may be defined on CMake level in CMakeLists.txt
+// so we should check here is it defined or no
+#ifndef TEST_EXPLICIT_KERNEL_NAMES
+#    if __SYCL_UNNAMED_LAMBDA__
+#        define TEST_EXPLICIT_KERNEL_NAMES 0
+#    else
+#        define TEST_EXPLICIT_KERNEL_NAMES 1
+#    endif // __SYCL_UNNAMED_LAMBDA__
+#endif // !TEST_USE_KERNEL_NAMES
```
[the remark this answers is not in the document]
SergeyKopienko: The CMake param ```TEST_USE_KERNEL_NAMES``` and defile were renamed to ```ONEDPL_EXPLICIT_KERNEL_NAMES```

test/CMakeLists.txt
dmitriy-sobolev: I tested this command and it looks it does not support overriding. So a user should solely rely on the dependecy('ONEDPL_USE_UNNAMED_LAMBDA), which does not suit us.

test/CMakeLists.txt
```diff
@@ -32,6 +32,22 @@ else()
...
+if (${TEST_EXPLICIT_KERNEL_NAMES} STREQUAL AUTO)
+    message(STATUS "    Usage of Kernel names in tests depends on unnamed lambda option")
+    # State of TEST_EXPLICIT_KERNEL_NAMES will be definel later in C++ code
+elseif(${TEST_EXPLICIT_KERNEL_NAMES} STREQUAL ALWAYS)
+    add_compile_definitions(TEST_EXPLICIT_KERNEL_NAMES=1)
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed: now we ```use add_definitions```

test/CMakeLists.txt
```diff
@@ -32,6 +32,23 @@ else()
     message(STATUS "oneDPL: -fno-fast-math is not supported by current compiler")
 endif()
 
+# usage of Kernel names in tests
+set(TestExplicitKernelNameVariants AUTO ALWAYS)
+set(ONEDPL_EXPLICIT_KERNEL_NAMES AUTO CACHE STRING "Use Kernel names in tests")
```
SergeyKopienko, line 37: Please see the example at https://stackoverflow.com/questions/39670879/cmake-variable-with-limited-valid-values

cmake/README.md
```diff
@@ -16,6 +16,7 @@ The following variables are provided for oneDPL configuration:
 | ONEDPL_USE_AOT_COMPILATION   | BOOL   | Enable ahead-of-time compilation for the GPU or CPU device types                              | OFF           |
 | ONEDPL_ENABLE_SIMD           | BOOL   | Enable SIMD vectorization by passing an OpenMP SIMD flag to the compiler if supported; the flag is passed to user project compilation string if enabled | ON           |
 | ONEDPL_AOT_ARCH              | STRING | Architecture options for ahead-of-time compilation, supported values can be found [here](https://software.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/compilation/ahead-of-time-compilation.html)                                                                                            | "*" for GPU device and "avx" for CPU device |
+| ONEDPL_EXPLICIT_KERNEL_NAMES | STRING | Control usage of Kernel names in tests [AUTO, ALWAYS]: if AUTO then Kernel names are not used in tests when unnamed lambda option is switched on; if ALWAYS then Kernel names are used in tests explicitly | AUTO          |
```
akukanov, line 19: As Dmitry said in his comment above, it makes sense to add "TEST" to the names of CMake variables that are supposed to only affect testing. Let's do that at least for this new variable.

@dmitriy-sobolev, for the existing variables, could renaming result in any issues for users (in other words, should it be considered as a potentially breaking change)?

test/support/utils_invoke.h
```diff
@@ -37,6 +37,107 @@ struct unique_kernel_name;
 template <typename Policy, int idx>
 using new_kernel_name = unique_kernel_name<typename ::std::decay<Policy>::type, idx>;
 
+constexpr bool
+explicit_use_kernel_names()
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed.

test/support/test_config.h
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/parallel_api/numeric/numeric.ops/reduce_by_segment_zip.pass.cpp
```diff
@@ -66,10 +66,8 @@ test_with_usm()
 
     //run reduce_by_segment algorithm 
     auto new_last = oneapi::dpl::reduce_by_segment(
-        oneapi::dpl::execution::make_device_policy(q), begin_keys_in,
-        end_keys_in, d_values, begin_keys_out, d_output_values);
-
-    q.wait();
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed.

test/support/utils_invoke.h
```diff
@@ -37,6 +37,107 @@ struct unique_kernel_name;
...
+/**
+ * make_policy functions test wrappers
+ * The main purpose of this function wrapper in TestUtils namespace - to cut template params from
+ * oneapi::dpl::execution::device_policy function calls depends on TEST_EXPLICIT_KERNEL_NAMES macro state.
+ * 
+ * ATTENTION: Please avoid to use oneapi::dpl::execution::device_policy directly in tests.
+ */
+template <typename... Types>
+inline auto
+make_device_policy(sycl::device d)
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/support/test_config.h
```diff
@@ -70,6 +70,16 @@
 #define TEST_UNNAMED_LAMBDAS 0
 #endif
 
+// The ONEDPL_EXPLICIT_KERNEL_NAMES macro may be defined on CMake level in CMakeLists.txt
+// so we should check here is it defined or no
+#ifndef ONEDPL_EXPLICIT_KERNEL_NAMES
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/general/test_policies.pass.cpp
```diff
@@ -67,18 +67,19 @@ main()
...
 #endif
-    test_policy_instance(make_device_policy<class Kernel_13>(sycl::device{TestUtils::default_selector}));
-    test_policy_instance(make_device_policy<class Kernel_14>(sycl::queue{TestUtils::default_selector, sycl::property::queue::in_order()}));
-    test_policy_instance(make_device_policy<class Kernel_15>(dpcpp_default));
-    test_policy_instance(make_device_policy<class Kernel_16>());
+    test_policy_instance(TestUtils::make_device_policy<class Kernel_13>(sycl::device{TestUtils::default_selector}));
+    test_policy_instance(TestUtils::make_device_policy<class Kernel_14>(sycl::queue{TestUtils::default_selector, sycl::property::queue::in_order()}));
+    test_policy_instance(TestUtils::make_device_policy<class Kernel_15>(dpcpp_default));
+    // Special case: required to call make_device_policy directly from oneapi::dpl::execution namespace
+    test_policy_instance(oneapi::dpl::execution::make_device_policy<class Kernel_16>());
```
SergeyKopienko, line 82: Special case: required to call make_device_policy directly from oneapi::dpl::execution namespace

test/support/utils_invoke.h
```diff
@@ -37,6 +37,107 @@ struct unique_kernel_name;
...
+/**
+ * make_policy functions test wrappers
+ * The main purpose of this function wrapper in TestUtils namespace - to cut template params from
+ * oneapi::dpl::execution::device_policy function calls depends on ONEDPL_EXPLICIT_KERNEL_NAMES macro state.
+ * 
+ * ATTENTION: Please avoid to use oneapi::dpl::execution::device_policy directly in tests.
+ */
+template <typename NewKernelName, typename OldKernelName = oneapi::dpl::execution::DefaultKernelName>
+inline auto
+make_device_policy(const oneapi::dpl::execution::device_policy<OldKernelName>& policy
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed.
But we had compile error here: https://github.com/oneapi-src/oneDPL/pull/821/files#r1136780099. I have fixed it by call function directly from oneapi::dpl::execution namespace and added comment.
Another variant - don't use our new ```TestUtils::make_device_policy``` wrappers in this file at all and make calls of ```oneapi::dpl::execution::make_device_policy```

cmake/README.md
SergeyKopienko: Fixed: ```ONEDPL_EXPLICIT_KERNEL_NAMES``` renamed to ```TEST_EXPLICIT_KERNEL_NAMES```
- in C++ code
- in CMake

test/CMakeLists.txt
```diff
@@ -32,6 +32,22 @@ else()
     message(STATUS "oneDPL: -fno-fast-math is not supported by current compiler")
 endif()
 
+# usage of Kernel names in tests
+set(TEST_EXPLICIT_KERNEL_NAMES AUTO CACHE STRING "Use explicit Kernel names in tests")
```
SergeyKopienko, line 36: This line strongly required.
Otherwise we have CMake errors like "set_property could not find CACHE variable TEST_EXPLICIT_KERNEL_NAMES"

cmake/README.md
dmitriy-sobolev: @akukanov, I do not consider it is a breaking change, because those cmake options (`ONEDPL_DEVICE_TYPE` and `ONEDPL_DEVICE_BACKEND`) affect only oneDPL tests.

test/CMakeLists.txt
dmitriy-sobolev: Will we be able to override it by calling `cmake -DTEST_EXPLICIT_KERNEL_NAMES=OFF ...`?

cmake/README.md
```diff
@@ -16,6 +16,7 @@ The following variables are provided for oneDPL configuration:
 | ONEDPL_USE_AOT_COMPILATION   | BOOL   | Enable ahead-of-time compilation for the GPU or CPU device types                              | OFF           |
 | ONEDPL_ENABLE_SIMD           | BOOL   | Enable SIMD vectorization by passing an OpenMP SIMD flag to the compiler if supported; the flag is passed to user project compilation string if enabled | ON           |
 | ONEDPL_AOT_ARCH              | STRING | Architecture options for ahead-of-time compilation, supported values can be found [here](https://software.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/compilation/ahead-of-time-compilation.html)                                                                                            | "*" for GPU device and "avx" for CPU device |
+| TEST_EXPLICIT_KERNEL_NAMES   | STRING | Control usage of Kernel names in tests [AUTO, ALWAYS]: if AUTO then Kernel names are not used in tests when unnamed lambda option is switched on; if ALWAYS then Kernel names are used in tests explicitly | AUTO          |
```
dmitriy-sobolev, line 19: This may be a bit cleaner:
```suggestion
| TEST_EXPLICIT_KERNEL_NAMES   | STRING | Control kernel naming. Affects only oneDPL test targets. Supported values: AUTO, ALWAYS. AUTO: rely on the compiler if "Unnamed SYCL lambda kernels" feature is on, otherwise provide kernel names explicitly; ALWAYS: provide kernel names explicitly | AUTO          |
```

SergeyKopienko: This change were applied (committed)

cmake/README.md
akukanov: @dmitriy-sobolev so let's create another patch to change the name prefix for the three CMake variables to `ONEDPL_TEST_` to clearly indicate both the project and the purpose.
This proposal does not extend to the corresponding macros in the code - these should be similar but exact match is not the goal. In particular, the `ONEDPL_` prefix should certainly **not** be used for test-only macros,

akukanov:
I am fine with this patch.

test/support/utils_invoke.h
```diff
@@ -37,6 +37,52 @@ struct unique_kernel_name;
 template <typename Policy, int idx>
 using new_kernel_name = unique_kernel_name<typename ::std::decay<Policy>::type, idx>;
 
+constexpr bool
+use_explicit_kernel_names()
+{
+#ifndef TEST_EXPLICIT_KERNEL_NAMES
+    static_assert(false, "The state of TEST_EXPLICIT_KERNEL_NAMES is not defined");
```
MikeDvorskiy, line 44: In case such preprocessor logic, in particular, some macro is not defined - usually a diagnostic is also on preprocessor level, like this:
https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/pstl/parallel_backend.h#L53

In general, we dont need such function `use_explicit_kernel_names()` at all...
It is simpler (and more clear) to use macro `TEST_EXPLICIT_KERNEL_NAMES` directly in the places of using (lines  67 and 79)

test/support/test_config.h
```diff
@@ -70,6 +70,16 @@
 #define TEST_UNNAMED_LAMBDAS 0
 #endif
 
+// The TEST_EXPLICIT_KERNEL_NAMES macro may be defined on CMake level in CMakeLists.txt
+// so we should check here is it defined or no
```
MikeDvorskiy, line 74: "is it" => "it is"

test/support/utils_invoke.h
```diff
@@ -37,6 +37,52 @@ struct unique_kernel_name;
...
+
+#if _ONEDPL_FPGA_DEVICE
+/**
+ * make_fpga_policy functions test wrappers
+ * The main purpose of this function wrapper in TestUtils namespace - to cut template params from
+ * oneapi::dpl::execution::device_policy function calls depends on TEST_EXPLICIT_KERNEL_NAMES macro state.
+ * 
+ * ATTENTION: Please avoid to use oneapi::dpl::execution::make_fpga_policy directly in tests.
+ */
+template <unsigned int new_unroll_factor, typename NewKernelName, typename Arg>
```
dmitriy-sobolev, line 75: Should it be aligned with the interface above?
```suggestion
template <unsigned int new_unroll_factor = 1, typename NewKernelName = DefaultKernelNameFPGA, typename Arg>
```

SergeyKopienko: Fixed (applied)

test/support/test_config.h
SergeyKopienko: Fixed: "if it is defined or not"

dmitriy-sobolev (approved):

PR #823, opened 2023-02-28
Correct usage of _HasDefaultName utility

adamfidel:
The `_HasDefaultName` utility does not match kernel names that have multiple template parameters. This PR fixes the utility to check if any of the template parameters is the default name.

Co-authored with @rarutyun.


SergeyKopienko (approved):
Up to me it seems OK.
But please get additional approves.

julianmi (approved):
LGTM.

PR #826, opened 2023-03-01
Explicit naming of workaround for GNU-like compiler drivers on windows

danhoeflinger:
For the workaround for GNU-Like front ends on windows, this PR explicitly names the CMAKE_CXX_COMPILER_ID to "IntelLLVM Workaround", and CMAKE_CXX_COMPILER_VERSION to the accurate version number.  It previously was CMAKE_CXX_COMPILER_ID="Clang " (with a space) and did not set the version.
 
The workaround is designed to get around incorrect cmake handling of GNU-like intel compilers on windows and functions by changing CMAKE_CXX_COMPILER_ID to something unrecognized by cmake.  If we do instead set CMAKE_CXX_COMPILER_ID to "Clang" and attempt to masquerade as clang, other problems occur.  Therefore, it is less confusing to label it as an "IntelLLVM Workaround".

This PR changes the condition for avoiding `target_compile_features(oneDPL INTERFACE cxx_std_17)` to when "IntelLLVM Workaround" compiler is the CMAKE_CXX_COMPILER_ID to explicitly tie it to the workaround which is preventing usage.  The `target_compile_features(oneDPL INTERFACE cxx_std_17)` check relies upon a "known" CMAKE_CXX_COMPILER_ID to function properly, and we have deliberately modified it away from that knowledge.  

This PR re-enables the check when using `dpcpp-cl`, as it is fine to check `target_compile_features(oneDPL INTERFACE cxx_std_17)`, because the workaround should not be applied for that MSVC-like front end. 

The PR also adds a #TODO  for enabling icpx on windows, which currently does not work as it does not support some flags which are supplied on windows: `/bigobj /Zc:__cplusplus /EHsc`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,13 +227,13 @@ jobs:
           ) else (
             set ninja_targets=build-onedpl-tests
           )
-          if "${{ matrix.cxx_compiler }}" == "dpcpp"(
+          if "${{ matrix.cxx_compiler }}" == "dpcpp" (
             set TOOLCHAIN_FILE_ENABLED=TRUE
           )
-          if "${{ matrix.cxx_compiler }}" == "icpx"(
+          if "${{ matrix.cxx_compiler }}" == "icpx" (
             set TOOLCHAIN_FILE_ENABLED=TRUE
           )
-          if defined TOOLCHAIN_FILE_ENABLED(
+          if %TOOLCHAIN_FILE_ENABLED% == TRUE (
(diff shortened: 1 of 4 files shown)


danhoeflinger:
In light of this conversation, 
https://github.com/oneapi-src/oneDPL/pull/827#discussion_r1123206091

I think I will close this PR in favor or incorporating the changes into #827.  They are intertwined with each other, and will be easiest to address together.

PR #828, opened 2023-03-01
Checking fno-fast-math alternative compiler flags

danhoeflinger:
Checking fno-fast-math alternative compiler flags (for icx[-cl], dpcpp-cl).


test/CMakeLists.txt
```diff
@@ -16,7 +16,15 @@ set (ranlux_24_48_test.pass_timeout_debug "900") # 15min
 set (ranlux_24_48_test.pass_timeout_release "720") # 12min
 
 # disable fast math for ALL TESTS
-check_cxx_compiler_flag("-fno-fast-math" _fno_fast_math_option)
+foreach(_fno_fast_math_flag -fno-fast-math /clang:-fno-fast-math)
```
adamfidel, line 19: I'm curious which compiler has its no-fast-math flag as `/clang:-fno-fast-math`. Could you write a comment to say which compiler this is trying to target?

test/CMakeLists.txt
```diff
@@ -16,7 +16,15 @@ set (ranlux_24_48_test.pass_timeout_debug "900") # 15min
...
+    string(MAKE_C_IDENTIFIER ${_fno_fast_math_flag} FLAG_DISPLAY_NAME)
+    check_cxx_compiler_flag(${_fno_fast_math_flag} ${FLAG_DISPLAY_NAME}_option)
+    if (${FLAG_DISPLAY_NAME}_option)
+        target_compile_options(oneDPL INTERFACE ${_fno_fast_math_flag})
```
adamfidel, line 23: I'm a little confused about the logic here. From what I can tell, this is the effect based on which flag is present or not present in the compiler:

|  `-fno-fast-math` | `/clang:-fno-fast-math` | Behavior |
|---|---|---|
| Present | Not Present | `target_compile_options(oneDPL INTERFACE -fno-fast-math)` and `add_compile_options("-fno-fast-math")`  |
|  Not Present | Present | `target_compile_options(oneDPL INTERFACE /clang:-fno-fast-math)` and `add_compile_options("-fno-fast-math")`  |
| Present | Present | `target_compile_options(oneDPL INTERFACE -fno-fast-math)` and `add_compile_options("-fno-fast-math")`  |
| Not Present | Not Present | -  |

For the case where `/clang:-fno-fast-math` is available but `-fno-fast-math` is not, it seems like it will still try to set `-fno-fast-math` in line 21. Is that correct?

adamfidel (changes requested):

test/CMakeLists.txt
SergeyKopienko: Initially this were made by me at https://github.com/oneapi-src/oneDPL/pull/729
I think that I investigated this issue on dpcpp compiler.

danhoeflinger: Interestingly, icx-cl and dpcpp-cl do not accept `-fno-fast-math`, but do accept `/clang:-fno-fast-math`, passing the option to the underlying clang driver.  I'm not aware of another version that they accept.  This is what I am targetting.

(They both do accept `-ffast-math` though)

test/CMakeLists.txt
danhoeflinger: Sorry this is a bug.  I'll update the code to properly use the caught option.

adamfidel (approved):
LGTM.

Issue #829, opened 2023-03-02
Unable to use -Wall -Werror in 2022.1.0-rc2


rscohn2:
I compile my code with -Wall -Werror. I am trying the oneDPL-2022.1.0-rc2 and am getting unused variable warnings.
```
/nfs/site/home/rscohn1/local/projects/dds/dr/build/icpx/_deps/dpl-src/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h:5\
20:43: error:
      unused variable '__id_in_subgroup' [-Werror,-Wunused-variable]
                    const ::std::uint16_t __id_in_subgroup = __subgroup.get_local_id();
```

adamfidel:
Thanks for the report. This is fixed in #833.

adamfidel:
Fixed in https://github.com/oneapi-src/oneDPL/pull/833.

PR #830, opened 2023-03-02
remove unused variable to avoid warning

rscohn2:
Addresses #829


adamfidel:
Thanks for looking into this issue. We included this change in https://github.com/oneapi-src/oneDPL/pull/833.

PR #831, opened 2023-03-02
Update release notes for oneDPL 2022.1.0

ValentinaKats:


documentation/release_notes.rst
```diff
@@ -8,6 +8,51 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
...
+- Improved performance of ``sort`` and ``stable_sort`` APIs for arithmetic data types
+  used with DPC++ execution policies.
+
+Fixed Issues
+------------
+- Fixed ``TBB headers not found`` issue occurring with libstdc++ version 9 when
+  oneTBB headers are not present in the environment. The fix requires inclusion of
+  the oneDPL headers before the libstdc++ headers.
+- С++ 17 is now required as a minimal standard version when oneDPL is integrated through CMake
+  on Linux and Windows with some exceptions, see `New in This Release`_
```
julianmi, line 28:
```suggestion
  on Linux and Windows with some exceptions, see `New in This Release`_.
```

akukanov, line 28: Should it be something like "oneDPL CMake scripts now enforce C++17 as the minimally required language version"? Is it important to mention Windows and Linux? If yes, asterisks (*) might be needed. And I would remove the "some exceptions" disclaimer.

documentation/release_notes.rst
```diff
@@ -8,6 +8,51 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
...
+
+Known Issues and Limitations
+----------------------------
+.. _New in This Release:
+
+New in This Release:
+^^^^^^^^^^^^^^^^^^^
+- Intel® oneAPI DPC++/C++ Compiler drivers with Linux-style options (``dpcpp`` and ``icpx``) may not work properly on
+  Windows due to some limitations of CMake 3.20 and newer. It may result, for example, in not recognized compiler features.
+  We recommend using ``icx`` or ``icx-cl`` on Windows while integrating oneDPL through CMake.
```
akukanov, line 40: It does not sound clear to me. How could limitations of CMake cause compiler drivers not work properly? Let's discuss over email with @dmitriy-sobolev

documentation/release_notes.rst
```diff
@@ -8,6 +8,51 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
...
+-	Input and output iterators should have the same type for in-place ``exclusive_scan`` algorithm. 
+- Undefined behaviour in in-place ``exclusive_scan`` algorithm when output iterator is not equal to input iterator.
```
akukanov, line 42: Is it really an issue? I do not quite understand how input and output iterator types could be different for in-place operations. @SergeyKopienko, please help me understand what problem is described here.

SergeyKopienko: I mean the next thing.
In this release we fixed error in in-place exclusive scan: we fixed the case when input and output iterator has the same type and they are equal:
```
constexpr bool
__check_equal_iterators(_Iterator1 __it1, _Iterator2 __it2)
{
    // In-place exclusive scan works correctly only if an input and an output iterators are the same type.
    // Otherwise, there is no way to check an in-place case and a workaround below is not applied.
    if constexpr (::std::is_same_v<::std::decay_t<_Iterator1>, ::std::decay_t<_Iterator2>>)
    {
        return __it1 == __it2;
    }

    return false;
}
```

```
template <typename _ExecutionPolicy, typename _Iterator1, typename _Iterator2, typename _UnaryOperation,
          typename _InitType, typename _BinaryOperation, typename _Inclusive>
oneapi::dpl::__internal::__enable_if_hetero_execution_policy<_ExecutionPolicy, _Iterator2>
__pattern_transform_scan_base(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __last, _Iterator2 __result,
                              _UnaryOperation __unary_op, _InitType __init, _BinaryOperation __binary_op, _Inclusive)
{
    if (__first == __last)
        return __result;

    const auto __n = __last - __first;

    const bool __is_scan_inplace_exclusive = __n > 1 && !_Inclusive{} && __check_equal_iterators(__first, __result);
    if (!__is_scan_inplace_exclusive)
    {
        __pattern_transform_scan_base_impl(__exec, __first, __last, __result, __unary_op, __init, __binary_op,
                                           _Inclusive{});
    }
    else
    {
        // ...
    }

    return __result + __n;
}
```

But we not fixed the cases when, for example, ```__result``` is ```__first + 1```.
Also we are not goes to fix the case when they are some different types but they describes the same data: for example, I think, ```sycl_iterator``` and ```raw pointer``` when we working on host.

documentation/release_notes.rst
ValentinaKats: Thanks, fixed

documentation/release_notes.rst
```diff
@@ -8,6 +8,51 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
...
+- Improved performance of the ``inclusive_scan``, ``exclusive_scan``, ``reduce`` and
+  ``max_element`` algorithms on heterogeneous devices.
+- Improved performance of ``sort`` and ``stable_sort`` APIs for arithmetic data types
+  used with DPC++ execution policies.
+
+Fixed Issues
+------------
+- Fixed ``TBB headers not found`` issue occurring with libstdc++ version 9 when
+  oneTBB headers are not present in the environment. The fix requires inclusion of
+  the oneDPL headers before the libstdc++ headers.
```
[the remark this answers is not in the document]
ValentinaKats: Thanks, I changed the note accordingly

documentation/release_notes.rst
akukanov: Thanks.
In the C++ standard, the remark for the in-place scan only says that "`result` may be equal to `first`". The case of overlapping ranges (e.g. `result == first+1`) is not required to support, we do not plan to "fix" this. I think the case of different iterator types is also not required to support; there is no even a standard way to test if two types can be compared for equality which we could use instead of `is_same_v` in the above code. The case of `cbegin` and `begin` we should support, but don't we already, thanks to `decay_t`?

Let's drop this note.

akukanov (approved):

PR #832, opened 2023-03-05
Fix Hanging buffer_wrapper.pass Test

mmichel11:
An issue on certain systems is that the `buffer_wrapper.pass` test case hangs indefinitely. The cause of this issue has been identified as the presence of multiple host accessors with write access to the same underlying buffer. When the second host accessor is created, its constructor may hang waiting for the first to destruct which will never happen. Since we only perform read operations with these accessors, we can explicitly mark them as `sycl::read_only`. This change resolves the issue as the accessors no longer have the ability to modify the underlying buffer and the construction of subsequent accessors will no longer block.


SergeyKopienko (approved):
LGTM

danhoeflinger (approved):
LGTM

PR #833, opened 2023-03-06
Adjust local memory requirement for single-group scan, sort and reduce

adamfidel:
This PR changes the single-group scan, sort and reduce algorithm to account for local device memory that may be used by the compiled kernel.

For scan and sort, we ensure that we only attempt to allocate up to half of the local memory provided by the device. For reduce, we double the local memory requirement.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
@@ -130,7 +130,7 @@ struct __subgroup_radix_sort
 
         // Pessimistically only use half of the memory to take into account memory used by compiled kernel
         const ::std::size_t __max_slm_size =
-            __q.get_device().template get_info<sycl::info::device::local_mem_size>() * 0.5;
+            __q.get_device().template get_info<sycl::info::device::local_mem_size>() / 2;
 
         const auto __n_uniform = 1 << (::std::uint32_t(log2(__n - 1)) + 1);
         const auto __req_slm_size_val = sizeof(_T) * __n_uniform;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
@@ -136,7 +136,7 @@ struct __subgroup_radix_sort
         const auto __req_slm_size_val = sizeof(_T) * __n_uniform;
         const auto __req_slm_size_counters = __counter_buf_sz * sizeof(uint32_t);
 
-        return __req_slm_size_val <= __max_slm_size - __req_slm_size_counters; //counters should be placed in SLM
+        return __req_slm_size_val + __req_slm_size_counters <= __max_slm_size; //counters should be placed in SLM
     }
 
     template <typename _KernelName>
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
@@ -128,7 +128,9 @@ struct __subgroup_radix_sort
     {
         assert(__n <= 1 << (sizeof(uint16_t) * 8)); //the kernel is designed for data size <= 64K
 
-        const auto __max_slm_size = __q.get_device().template get_info<sycl::info::device::local_mem_size>();
+        // Pessimistically only use half of the memory to take into account memory used by compiled kernel
+        const ::std::size_t __max_slm_size =
+            __q.get_device().template get_info<sycl::info::device::local_mem_size>() * 0.5;
 
         const auto __n_uniform = 1 << (::std::uint32_t(log2(__n - 1)) + 1);
         const auto __req_slm_size_val = sizeof(_T) * __n_uniform;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -314,8 +314,9 @@ __parallel_transform_reduce(_ExecutionPolicy&& __exec, _ReduceOp __reduce_op, _T
     // TODO: find a way to generalize getting of reliable work-group size
     ::std::size_t __work_group_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
     // change __work_group_size according to local memory limit
+    // Pessimistically double the memory requirement to take into account memory used by compiled kernel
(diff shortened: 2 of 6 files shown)


SergeyKopienko:
I think required to check ```__max_local_allocation_size``` and it's usage too.
Seems that it's not fixed - at least at ```include\oneapi\dpl\pstl\hetero\algorithm_ranges_impl_hetero.h```

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -120,7 +120,9 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
     if ((__n_uniform & (__n_uniform - 1)) != 0)
         __n_uniform = __par_backend_hetero::__dpl_bit_floor(__n) << 1;
 
-    const auto __max_slm_size = __exec.queue().get_device().template get_info<sycl::info::device::local_mem_size>();
+    // Pessimistically only use half of the memory to take into account memory used by compiled kernel
+    const auto __max_slm_size =
+        __exec.queue().get_device().template get_info<sycl::info::device::local_mem_size>() * 0.5;
```
danhoeflinger, line 125: Similarly to sort, should this be explicitly `::std::size_t` instead of auto?

adamfidel:
> I think required to check `__max_local_allocation_size` and it's usage too. Seems that it's not fixed - at least at `include\oneapi\dpl\pstl\hetero\algorithm_ranges_impl_hetero.h`

That call is inside of the reduce by segment implementation. It doesn't seem to allocate local memory in that algorithm so it might not be necessary to adjust this call.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
adamfidel: Yes, just changed.

SergeyKopienko:
And what about
```
        // change __wgroup_size according to local memory limit
        __wgroup_size = oneapi::dpl::__internal::__max_local_allocation_size(::std::forward<_ExecutionPolicy>(__exec),
                                                                             sizeof(_Type), __wgroup_size);
```
from include\oneapi\dpl\pstl\hetero\dpcpp\parallel_backend_sycl.h ?

adamfidel:
> And what about
> 
> ```
>         // change __wgroup_size according to local memory limit
>         __wgroup_size = oneapi::dpl::__internal::__max_local_allocation_size(::std::forward<_ExecutionPolicy>(__exec),
>                                                                              sizeof(_Type), __wgroup_size);
> ```
> 
> from include\oneapi\dpl\pstl\hetero\dpcpp\parallel_backend_sycl.h ?

This check is from the original scan algorithm. It compiles the kernel and adjusts the work group size in the next few lines, so we do not need to adjust the memory requirement outside of what it is already doing.

danhoeflinger:
> > I think required to check `__max_local_allocation_size` and it's usage too. Seems that it's not fixed - at least at `include\oneapi\dpl\pstl\hetero\algorithm_ranges_impl_hetero.h`
> 
> That call is inside of the reduce by segment implementation. It doesn't seem to allocate local memory in that algorithm so it might not be necessary to adjust this call.

Isn't this workaround specifically meant for algorithms which use sycl group operations?   I don't think this `reduce_by_segment` implementation ends up calling any of those.

Radix sort has one joint_exclusive_scan call, but has a comment:

https://github.com/oneapi-src/oneDPL/blob/dev/adamfidel/single_group_mem_req/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h#L268-L269

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -128,7 +128,9 @@ struct __subgroup_radix_sort
     {
         assert(__n <= 1 << (sizeof(uint16_t) * 8)); //the kernel is designed for data size <= 64K
 
-        const auto __max_slm_size = __q.get_device().template get_info<sycl::info::device::local_mem_size>();
+        // Pessimistically only use half of the memory to take into account memory used by compiled kernel
+        const ::std::size_t __max_slm_size =
+            __q.get_device().template get_info<sycl::info::device::local_mem_size>() * 0.5;
```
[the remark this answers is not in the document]
adamfidel: Good point. I changed it to an integer division by 2.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -314,8 +314,9 @@ __parallel_transform_reduce(_ExecutionPolicy&& __exec, _ReduceOp __reduce_op, _T
     // TODO: find a way to generalize getting of reliable work-group size
     ::std::size_t __work_group_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
     // change __work_group_size according to local memory limit
+    // Pessimistically double the memory requirement to take into account memory used by compiled kernel
     __work_group_size = oneapi::dpl::__internal::__max_local_allocation_size(::std::forward<_ExecutionPolicy>(__exec),
-                                                                             sizeof(_Tp), __work_group_size);
+                                                                             sizeof(_Tp) * 2, __work_group_size);
```
SergeyKopienko, line 319: May be better to use in this place ```/ 2``` as in other places ?

adamfidel: Reduce is a little different here because instead of using `get_info<sycl::info::device::local_mem_size>()` directly like sort and scan do, it calls `__max_local_allocation_size` which internally uses  `get_info<sycl::info::device::local_mem_size>()`. So instead, we double the memory requirement as opposed to halving the available memory.

danhoeflinger: I think we achieve the desired effect here, but are abusing the arguments to the function. 
The argument is `__type_size`, and we are doubling the true type size to end up with something which then doesn't exceed half the memory size.  
We could consider adding an additional optional float parameter defaulted to 1 which scales the memory returned from the device query prior to using it.  I **would** use a float here and then cast back to integer because otherwise you severely limit the expressiveness of that option.

akukanov: I have a patch already that refactors this function, with the semantics of this parameter to be "the amount of SLM needed per work item". This change seems to be OK with that new semantics. As this case requires O(workgroup size) of local memory and computes the group size based on local memory size, doubling the requirement essentially reduces the effective WG size (and the real memory needed) twice. If that size is still at least 512, then the real kernels will use at most half of the available memory.

The bottom line is that I believe this change is good and achieves the goal we want.

danhoeflinger: Works for me.  In any case, the comment above the line of code should resolve any question of the intention of the call.

akukanov (dismissed):

timmiesmith (dismissed):
LGTM

timmiesmith (approved):

PR #834, opened 2023-03-07
Fix header path in pkg-config files generated without OS subdir

dmitriy-sobolev:
This generates the right header path for the package config files when they are generated with `SKIP_HEADERS_SUBDIR` set to `TRUE`.
This is useful for creation of e.g., conda packages which do not have `windows` or `linux` subdirectories.

### DIFF
--- a/cmake/README.md
+++ b/cmake/README.md
@@ -108,4 +108,8 @@ This section is applicable for oneDPL packaging creation process, but not for us
 
 How to use:
 
-`cmake [-DOUTPUT_DIR=<output_dir>] -P cmake/scripts/generate_config.cmake`
+`cmake [-DSKIP_HEADERS_SUBDIR=<TRUE|FALSE>] [-DOUTPUT_DIR=<output_dir>] -P cmake/scripts/generate_config.cmake`
+
+When `SKIP_HEADERS_SUBDIR` is set to false or not defined (by default), the script adds the subdirectories:
+- `windows` and `linux` for headers in `<prefix></subdirectory>/include` pattern.
+- `pkgconfig-win` and `pkgconfig-lin` for pkg-config files in `<prefix></subdirectory>/dpl.pc` pattern.
--- a/cmake/scripts/generate_config.cmake
+++ b/cmake/scripts/generate_config.cmake
@@ -48,19 +48,14 @@ configure_file("${ONEDPL_ROOT}/cmake/templates/oneDPLConfigVersion.cmake.in"
                "${OUTPUT_DIR}/oneDPLConfigVersion.cmake"
                @ONLY)
 
-set(_onedpl_pkgconfig_header_suffix include)
-if (NOT SKIP_HEADERS_SUBDIR)
+if (SKIP_HEADERS_SUBDIR)
+    set(_onedpl_pkgconfig_header_suffix include)
+    configure_file("${ONEDPL_ROOT}/integration/pkgconfig/dpl.pc.in" "${OUTPUT_DIR}/dpl.pc" @ONLY)
+else()
     set(_onedpl_pkgconfig_header_suffix windows/include)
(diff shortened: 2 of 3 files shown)


adamfidel:
Could you please update `cmake/README.md` and add the option for `SKIP_HEADERS_SUBDIR` to the documentation here?
https://github.com/oneapi-src/oneDPL/blob/06f7b0a50768e573a9e9ae8ba3be7160dd84696e/cmake/README.md?plain=1#L111


I did some testing and I can confirm that running `cmake -DOUTPUT_DIR=/tmp/install -DSKIP_HEADERS_SUBDIR=true -P cmake/scripts/generate_config.cmake` does modify the pkg-config files as expected. So I believe these changes work as expected.

Strangely, I cannot run the CMake install target as a whole. I get the following error:
```
$ cmake --install . --prefix /tmp/install
CMake Error: Error processing file: [...snip...]/cmake/cmake_install.cmake
```
I get this same error on main however, so it might be something weird with my setup. It is not something related to your changes.

adamfidel:
I'm not familiar with how the installed package is intended to be used with respect to pkg-config, but does it make sense to create a single `pkgconfig/dpl.pc` file if `SKIP_HEADERS_SUBDIR` is set to true? With these changes, it will create both `pkgconfig-lin/dpl.pc` and `pkgconfig-win/dpl.pc` which have identical contents.

It might be okay, but I'm just wondering if it would be cleaner to create a single pkg-config file instead of two identical ones here.

adamfidel (approved):
LGTM.

dmitriy-sobolev:
> Could you please update `cmake/README.md` and add the option for `SKIP_HEADERS_SUBDIR` to the documentation here?
> 
> https://github.com/oneapi-src/oneDPL/blob/06f7b0a50768e573a9e9ae8ba3be7160dd84696e/cmake/README.md?plain=1#L111
> 
> I did some testing and I can confirm that running `cmake -DOUTPUT_DIR=/tmp/install -DSKIP_HEADERS_SUBDIR=true -P cmake/scripts/generate_config.cmake` does modify the pkg-config files as expected. So I believe these changes work as expected.
> 
> Strangely, I cannot run the CMake install target as a whole. I get the following error:
> 
> ```
> $ cmake --install . --prefix /tmp/install
> CMake Error: Error processing file: [...snip...]/cmake/cmake_install.cmake
> ```
> 
> I get this same error on main however, so it might be something weird with my setup. It is not something related to your changes.

Thank you for the review. I updated the cmake documentation. 

It seems cmake files should be generated before using `cmake --install` command according to a note from note from https://cmake.org/cmake/help/latest/manual/cmake.1.html#install-a-project:
> CMake provides a command-line signature to install an already-generated project binary tree:

You can try these commands:
```
cmake --configure ..
cmake --install . --prefix /path/to/dpl
```

dmitriy-sobolev:
> I'm not familiar with how the installed package is intended to be used with respect to pkg-config, but does it make sense to create a single `pkgconfig/dpl.pc` file if `SKIP_HEADERS_SUBDIR` is set to true? With these changes, it will create both `pkgconfig-lin/dpl.pc` and `pkgconfig-win/dpl.pc` which have identical contents.
> 
> It might be okay, but I'm just wondering if it would be cleaner to create a single pkg-config file instead of two identical ones here.

It definitely makes sense. I've done that. 

I expect the following use case of the pkg-config files. A user sources oneDPL, the source script sets path `PKG_CONFIG_PATH`, what allows compiling a third-party apps as:
```
# `pkg-config --cflags dpl` is evaluated to -I/path/to/oneDPL/include
icpx -fsycl my_app.cpp `pkg-config --cflags dpl` 
```

adamfidel (approved):
Great, thanks for making those changes!

PR #835, opened 2023-03-08
Remove deprecated `target::host_buffer`

BenBrock:
Replacing a use of deprecated `target::host_buffer` with `target::host_task`.  There is [one other use of `host_buffer` here](https://github.com/BenBrock/oneDPL/blob/remove-deprecated-host-buffer/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h#L131), but it is using the internal namespace, so not sure if it should be replaced as well.


include/oneapi/dpl/internal/iterator_impl.h
```diff
@@ -45,7 +45,7 @@ struct extract_accessor<oneapi::dpl::__internal::sycl_iterator<Mode, T, Allocato
     static constexpr sycl::access::mode mode = Mode;
     static constexpr int dim = 1;
     using buffer_type = sycl::buffer<T, dim, Allocator>;
-    using accessor_type = sycl::accessor<T, dim, mode, sycl::access::target::host_buffer>;
+    using accessor_type = sycl::accessor<T, dim, mode, sycl::access::target::host_task>;
```
akukanov, line 48: I think we should instead use `host_accessor` here. @MikeDvorskiy please take a look.

danhoeflinger: I agree that `sycl::host_accessor` should be used as the return type.

Also, the sycl buffer member function used to get the return value,  `get_access<mode>()`, is also deprecated in SYCL 2020 and `get_host_access()` should be used instead.  These should change together.

BenBrock:
I've updated the PR to use `sycl::host_accessor` instead.  I quickly searched through the codebase for any usage of `get_mode<..., target...>`, but didn't find any.

danhoeflinger:
> I've updated the PR to use `sycl::host_accessor` instead. I quickly searched through the codebase for any usage of `get_mode<..., target...>`, but didn't find any.

To be more specific because there are many `get_access` overloads...
From the SYCL spec for the buffer class:
```
template <access_mode Mode>
accessor<T, Dimensions, Mode, target::host_buffer> get_access()
```
"Deprecated in SYCL 2020. Use get_host_access() instead.
Returns a valid host accessor to the buffer with the specified access mode and target."


---
The instance I was referring to is right after your change,  `include/oneapi/dpl/internal/iterator_impl.h`: line 53.

The function which makes use of this `accessor_type` should be changed.
```
    static accessor_type
    get(oneapi::dpl::__internal::sycl_iterator<Mode, T, Allocator>& iter)
    {
        return iter.get_buffer().template get_access<mode>();  // <-- here
    }
```

I think there are some instances elsewhere as well in `include/oneapi/dpl/internal/function.h` (a helper function which may not be used anymore) and in a test `test/xpu_api/functional/test_functional.cpp`.

akukanov:
The possible problem with trying to change `get_access` to `get_host_access` now is that the latter cannot be parameterized by the mode, instead it uses deduction tags as function arguments. So more changes will likely be needed in the class, and etc. And I am not even sure that this class is needed in the long term, as we now re-think the approach for buffer access modes in our pseudo-iterator class.

So I suggest to either accept the current one-line change in this PR, or postpone it completely until we figure out how we deal with access modes in general.

danhoeflinger:
> So I suggest to either accept the current one-line change in this PR, or postpone it completely until we figure out how we deal with access modes in general.

Its a fair point, I hadn't fully considered the implications of how this will work in context.  However, I believe these changes must be done together.  To return a `sycl::host_accessor<T,dim,mode>` from this function, we need to use `get_host_access()`, as there is no conversion from the result of `get_access<mode>()` (the old return type) to a `sycl::host_accessor<T,dim,mode>`.  Testing this briefly as a one-line change results in an error.

Therefore, I think its best to postpone this.  As you mention, this struct may not be needed in general.  As far as I can find, I don't believe it is used anywhere in the repository (and hasn't been in some time).

BenBrock:
Closing this PR.

PR #838, opened 2023-03-10
Fix a compilation error in async algorithms introduced by #811

akukanov:
My bad, forgot to change the use of `__is_convertible_to_event` after changing its definition from `bool_constant` to `bool`.


SergeyKopienko (approved):
LGTM, compile error fixed.

PR #842, opened 2023-03-13
CI: fix "Node.js 12 actions are deprecated" warnings

dmitriy-sobolev:
Using the latest actions versions is needed to fix deprecations, e.g.:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/setup-python@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

Specifying the latest version for python resulted in another warning:
> The `python-version` input is not set. The version of Python currently in `PATH` will be used.

That is why the latest stable (3.x) version of python was explicitly specified according to https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#using-the-python-version-input

The latest versions of the actions are found in https://github.com/actions (search for repositories in this channel and their latest major versions).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,7 +56,9 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v4
+        with:
+          python-version: '3.x'
       - name: Install prerequisites
         run: |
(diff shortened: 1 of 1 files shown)


timmiesmith (approved):
LGTM

PR #843, opened 2023-03-13
Fix Integer Wraparound in Single Group Reduce

mmichel11:
One of the issues that has been identified with the new reduce implementation are segfaulting tests on certain operating systems when compiled with the latest DPC++ compiler in release mode with C++17 and ran on the CPU. This issue has been identified as an illegal memory access that stems from the incorrect result of [this](https://github.com/oneapi-src/oneDPL/blob/661cc91707e39a4efde47c7a573eb40c71cc971f/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L219) assignment.

Within `const ::std::int64_t __items_to_process = __n - (__iters_per_work_item * __global_id);`, `__n` and `__iters_per_work_item * __global_id` are both `std::size_t` datatypes. Despite assigning the result to a `std::int64_t` type, the result may wrap around to a large integer when `__iters_per_work_item * __global_id > __n` causing the incorrect branch to be taken and a segfault when memory is accessed with that offset. In practice, this behavior is uncommon but is the cause of these segfaults discussed above.

To remedy this issue and prevent this wraparound behavior, we explicitly cast the two terms to `std::int64_t` before the subtraction takes place.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
@@ -216,7 +216,7 @@ struct transform_reduce_known
                const ::std::size_t __global_offset, _AccLocal& __local_mem, const _Acc&... __acc) const
     {
         const ::std::size_t __adjusted_global_id = __global_offset + __iters_per_work_item * __global_id;
-        const ::std::int64_t __items_to_process = 
+        const ::std::int64_t __items_to_process =
             static_cast<std::int64_t>(__n) - static_cast<std::int64_t>(__iters_per_work_item * __global_id);
         // Add neighbour to the current __local_mem
         if (__items_to_process >= __iters_per_work_item)
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
@@ -229,7 +229,7 @@ struct transform_reduce_known
         }
         else if (__adjusted_global_id < __adjusted_n)
         {
-            const _Size __items_to_process =__adjusted_n - __adjusted_global_id;
+            const _Size __items_to_process = __adjusted_n - __adjusted_global_id;
             // Keep these statements in the same scope to allow for better memory alignment
             typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
             for (_Size __i = 1; __i < __items_to_process; ++__i)
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
@@ -215,11 +215,10 @@ struct transform_reduce_known
                const ::std::size_t /* unused __iters_per_work_item */, const ::std::size_t __global_id,
                const ::std::size_t __global_offset, _AccLocal& __local_mem, const _Acc&... __acc) const
     {
-        const ::std::size_t __adjusted_global_id = __global_offset + __iters_per_work_item * __global_id;
-        const ::std::int64_t __items_to_process =
-            static_cast<std::int64_t>(__n) - static_cast<std::int64_t>(__iters_per_work_item * __global_id);
+        const _Size __adjusted_global_id = __global_offset + __iters_per_work_item * __global_id;
+        const _Size __adjusted_n = __global_offset + __n;
         // Add neighbour to the current __local_mem
-        if (__items_to_process >= __iters_per_work_item)
+        if (__adjusted_global_id + __iters_per_work_item < __adjusted_n)
         {
             // Keep these statements in the same scope to allow for better memory alignment
             typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
@@ -228,8 +227,9 @@ struct transform_reduce_known
                 __res = __binary_op(__res, __unary_op(__adjusted_global_id + __i, __acc...));
             __local_mem[__local_id] = __res;
         }
-        else if (__items_to_process > 0)
+        else if (__adjusted_global_id < __adjusted_n)
         {
+            const _Size __items_to_process =__adjusted_n - __adjusted_global_id;
             // Keep these statements in the same scope to allow for better memory alignment
             typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
             for (_Size __i = 1; __i < __items_to_process; ++__i)
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
@@ -216,7 +216,8 @@ struct transform_reduce_known
                const ::std::size_t __global_offset, _AccLocal& __local_mem, const _Acc&... __acc) const
     {
         const ::std::size_t __adjusted_global_id = __global_offset + __iters_per_work_item * __global_id;
-        const ::std::int64_t __items_to_process = __n - (__iters_per_work_item * __global_id);
+        const ::std::int64_t __items_to_process = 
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -216,7 +216,8 @@ struct transform_reduce_known
                const ::std::size_t __global_offset, _AccLocal& __local_mem, const _Acc&... __acc) const
     {
         const ::std::size_t __adjusted_global_id = __global_offset + __iters_per_work_item * __global_id;
-        const ::std::int64_t __items_to_process = __n - (__iters_per_work_item * __global_id);
+        const ::std::int64_t __items_to_process =
```
rarutyun, line 219: I don't think we want negative values here either, do we?

I would focus on understanding (logically) why this underflow of unsigned happens here. Having this information we can decide if it's the fix we want

danhoeflinger: @julianmi would be better to weigh in on this, but I do believe that zero and negative values here are expected in normal operation.  This occurs because we are defining kernels for *only* each power of 2 up to 64K, with `__iters_per_work_item` anywhere from 1 to 128 depending on the kernel size.    However, we support all `n` values between powers of two, so those get sent to the next power of 2 up.  It results in a negative value here anytime we have a `_global_id` for our work item which would map fully to data larger than _n.   

(Currently work is not evenly distributed to the work_items in strided access, but rather chunked up in contiguous blocks.)

I believe the intention is that the `__items_to_process` is `::std::int64_t` to be able to handle such cases, but the problem is that the underflow occurs in the unsigned subtraction expression prior to being assigned to the signed type.  Casting before the subtraction brings the code in line with the intention.

Often, I think we get away with this because the very large (modulo wrapped) underflowed unsigned number usually results in the "correct" negative number when converted to a signed 64 bit integer.  However, this is taking advantage of behavior that is not explicitly defined.

mmichel11: I agree that understanding why the underflow is happening is most important for deciding on the fix.

With regards to negative values of `__items_to_process`, I believe in this scenario a negative value (or 0) is purposefully used to determine if the current work item is assigned any work. My understanding aligns with Dan's comment and I believe the logic of the code below this line is:

1. `__items_to_process >= __iters_per_work_item`: We have a full set of data to process for this work item. A full set of data is exactly `__iters_per_work_item` elements. We use a fully unrolled loop to transform and reduce the assigned data.
2. `__items_to_process > 0`: We have an incomplete set of data to process consisting of less than `__iters_per_work_item` elements. This always corresponds to the work item that processes the last few elements in the underlying data. In this scenario, we use a normal for loop to transform and reduce the assigned data since we do not know the exact amount of work until runtime.
3. `__items_to_process <= 0`: The start index our work item is assigned is beyond the range of elements in the underlying data. All work will be done by work items with a lower ID. In this case we do not do anything and just exit the loop.

@julianmi Is my understanding of this piece of code correct?

julianmi: Thanks for catching this. Both you statements, @danhoeflinger and @mmichel11, are correct.

julianmi (approved):
LGTM.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
akukanov: While I understand the reasoning above, I do not like it. Essentially, if the result is negative, it means that some work items have no work at all and stay idle. This sounds bad from the performance viewpoint.

Instead, why do not we treat the `__iters_per_work_item` template parameter as the upper bound (if it is even needed), and start using the currently unused function parameter with the same name which would contain the actual number of iterations per work item, computed by the caller?

akukanov (changes requested):
Please do not commit before responding to my comment in the discussion above

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: > While I understand the reasoning above, I do not like it. Essentially, if the result is negative, it means that some work items have no work at all and stay idle. This sounds bad from the performance viewpoint.

I agree. The current design uses a fixed grid of `__work_group_size` and  `__iters_per_work_item` throughout the iterative tree reductions. This leads to many work items idling depending on the number of items to reduce, especially if the number of elements is small compared to the grid size. I am working on a grid-strided loop to balance the work across the work items. This will make the fixed `__iters_per_work_item` parameter obsolete.

> Instead, why do not we treat the `__iters_per_work_item` template parameter as the upper bound (if it is even needed), and start using the currently unused function parameter with the same name which would contain the actual number of iterations per work item, computed by the caller?

While I have tested this approach, it prevents the unrolling of the main loop over `__iters_per_work_item` elements since `__items_to_process` would become dynamic. The unrolling is performance critical.

akukanov: I see.
Can we use the static template parameter for unrolling while safe, then process the remaining iterations in a dynamic loop? For example, using the previous power of 2 should be a safe bet. Would that still lose much performance, especially if measured with data sizes between powers of two?
Or would this be just a throw-out work because of the further changes @julianmi outlined?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -216,7 +216,8 @@ struct transform_reduce_known
...
+            static_cast<std::int64_t>(__n) - static_cast<std::int64_t>(__iters_per_work_item * __global_id);
```
akukanov, line 220: Technically, casting `size_t` to `int64_t` on 64-bit paltforms is a downcast and may lose significant digits :)

I suggest to take a slightly different approach: compute the number of work items that will have work  (`_n / __iters_per_work_item`) , and then compare it to the ID of the item.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: My experiments showed significant performance degradations when using shorted loops (e.g., 16 or 8 elements). I would argue to fix this as proposed by @mmichel11 and focus on the future approach with a grid-strided loop.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: Good point. I think we can work around the negative numbers. I will send out a patch shortly.

danhoeflinger: While this is technically true, if it is any performance cost to support sizes of n greater than 9,223,372,036,854,775,807 (maximum signed 64 bit integer), I don't think it is worth it.

julianmi: I have just pushed a change to first check if we can do the full `__iters_per_work_item` loop and if not if the work item has work at all. If so, we calculate `__items_to_process`. Please have another look.

danhoeflinger: Yes, this is an improvement and should be at least as good performance wise.

mmichel11: This change looks good to me as well.

rarutyun: What is under review right now looks much better comparing with the previous version. Need to go though it logically one more time (for me)

akukanov (approved):

rarutyun (approved):

PR #845, opened 2023-03-17
Add implementation of oneapi::dpl::cbegin() and oneapi::dpl::cend()

SergeyKopienko:
In this implementation we add the next new functional:
- hetero const iterator - class ```sycl_const_iterator```;
- implementation of ```oneapi::dpl::cbegin()```;
- implementation of ```oneapi::dpl::cend()```.

In tests we add tests for the next functional:
- we may check equality of ```sycl_iterator``` and ```sycl_const_iterator```;
- for check ```oneapi::dpl::cbegin()``` and ```oneapi::dpl::cend()``` work;
- in the file ```scan2.pass.cpp``` we add test for in-place exclusive scan when the same```sycl::buffer``` described through the input iterators ```sycl_const_iterator``` and through the output iterators ```sycl_iterator```;
- modified tests system for algorithm tests when used three pairs of hetero iterators: keys (from, to), vals (from, to) and results (from, to), In this case tests system now using ```sycl_const_iterator``` iterators for keys.


include/oneapi/dpl/internal/function.h
```diff
@@ -63,12 +64,25 @@ using oneapi::dpl::__par_backend_hetero::__internal::__buffer;
 // algorithm from either a SYCL iterator or a USM pointer.
 template <sycl::access::mode Mode, typename Policy, typename Iterator>
 auto
-get_access(Policy, Iterator i, typename ::std::enable_if<is_hetero_iterator<Iterator>::value, void>::type* = nullptr)
```
MikeDvorskiy, line 66: Is there code which uses that function?
I guess that function is not used any more or probably I'm wrong?

MikeDvorskiy:
Looks "overcomplicated"  redundant... lets discuss it offline

include/oneapi/dpl/internal/function.h
SergeyKopienko: You are right but I hadn't goal to remove any not used functions in this PR.

SergeyKopienko:
> Looks "overcomplicated" and redundant... lets discuss it offline.

Please see https://github.com/oneapi-src/oneDPL/pull/850

PR #849, opened 2023-03-21
[onedpl][radix] re-enabling one group radix sort

MikeDvorskiy:
[onedpl][radix] re-enabling one group radix sort

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
@@ -62,8 +62,9 @@
 #    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(SIZE) intel::reqd_sub_group_size(SIZE)
 #endif
 
-// TODO: re-enable when sort test passes for all devices of interest
-#define _ONEDPL_USE_SINGLE_GROUP_RADIX_SORT 0
+// There is oneAPI compiler (version 2023.1) issue with NDEBUG flag for ATSM device.
+// So, for  oneAPI compiler version 2023.1 it is recommended to define _ONEDPL_USE_SINGLE_GROUP_RADIX_SORT into 0
+#define _ONEDPL_USE_SINGLE_GROUP_RADIX_SORT 1
 
 namespace __dpl_sycl
 {


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -62,8 +62,9 @@
 #    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(SIZE) intel::reqd_sub_group_size(SIZE)
 #endif
 
-// TODO: re-enable when sort test passes for all devices of interest
-#define _ONEDPL_USE_SINGLE_GROUP_RADIX_SORT 0
+// There is oneAPI compiler (version 2023.1) issue with NDEBUG flag for ATSM device.
+// So, for  oneAPI compiler version 2023.1 it is recommended to define _ONEDPL_USE_SINGLE_GROUP_RADIX_SORT into 0
```
rarutyun, line 66: If that something you want to recommend that you should probably check `#ifnded this_macro_name` and only then define it.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -62,8 +62,12 @@
...
 #endif
 
-// TODO: re-enable when sort test passes for all devices of interest
-#define _ONEDPL_USE_SINGLE_GROUP_RADIX_SORT 0
+//It is recommended to update a graphic driver to one of the last versions, 23.05.25593.11 at least,
+//otherwise in case of Intel® Data Center GPU Flex Series and compiled code with NDEBUG flag
+//it is recommended to define _ONEDPL_USE_SINGLE_GROUP_RADIX_SORT into 0.
+#if !defined(_ONEDPL_USE_SINGLE_GROUP_RADIX_SORT)
+#    define _ONEDPL_USE_SINGLE_GROUP_RADIX_SORT 1
+#endif
```
akukanov, line 70: I see two issues here.

First, this file is clearly not the right place for the note; nobody will look for it there. Let's document it as a known issue in release notes.

Second, I do not like the suggestion to redefine an internal macro for a workaround. Do we really want it? If this is a driver issue and is already addressed in new versions, updating the driver is the proper solution.

We do not want the code to contain a not-well-documented macro that allows disabling a part of the implementation, do we? What are the reasons for this to be considered at all, why not just use the implementation as intended?

MikeDvorskiy: yes, comment was updated and the main recommendation  - to update a driver.

MikeDvorskiy: if we document it as a known issue in release notes we have to have a public macro, at least.
I deliberately did not introduce a public version, because  a user should not know about that possibility to off one group implementation. It is rather for a "quick fix" possibility for advanced user who has the broken driver and request support here.

akukanov: In the public release note we should not even mention the existence of this macro, just recommend updating the driver, period.

Unfortunately, we have little idea of how wide the impact could be, as we only saw a failure with one version of the driver on one system. Assuming the impact is limited and the solution is already there, I think it makes no sense to keep any macro, even the internal one. To those who are affected and cannot update the driver we can recommend to use oneDPL 2022.1.0 where the code is disabled.

MikeDvorskiy: Agree. I will update the changes.

akukanov (approved):

PR #852, opened 2023-03-22
Fix an error in __check_equal_iterators

SergeyKopienko:


SergeyKopienko:
@MikeDvorskiy what do you think about this approach?

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,6 +152,24 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
     }
 }
 
+template <sycl::access::mode _Mode1, sycl::access::mode _Mode2, typename _T, typename _Allocator>
+bool
+__check_equal_iterators(sycl_iterator<_Mode1, _T, _Allocator> __it1, sycl_iterator<_Mode2, _T, _Allocator> __it2)
+{
+    if (__it1 != __it2)
+        return false;
```
danhoeflinger, line 160: As @akukanov  mentioned offline, it isn't valid to compare these unless we know these are iterators on the same buffer.  At the very least, they should be moved after the check for the buffers.  

However, with the knowledge that a `(sub_buffer == parent_buffer)` is `true`, I don't believe this check is even known to be valid then.  The `idx` within `sycl_iterator` that is checked here is layered on top of the inherent offset for a sub_buffer.  Without a way to check the sub_buffer's offset, if either iterator's buffer is a sub_buffer I think we have to merely return the result of the buffer equality check (and take the performance hit).

If both buffers are not a sub_buffer, and the buffers are equal, then we can check the equality of the iterators like this.

It also may be worth renaming this function, as we are warping the meaning to basically be:

`__check_if_iterator_equality_is_possible()`

True: these iterators may be the same
False: we know these iterators are different

akukanov: > it isn't valid to compare these unless we know these are iterators on the same buffer

Note also that the same concern applies to the original implementation of this function. It is in general not valid, I think, to compare two iterators just on the base that they are of the same type.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,77 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+};
+
+// Check the ability to compare two sycl_iterator's
+/*
+ * @return bool - true, if these iterators may be the same;
+ *         false - we know these iterators are different
+ */
+template <sycl::access::mode _Mode1, sycl::access::mode _Mode2, typename _T, typename _Allocator>
+bool
+__check_if_iterator_equality_is_possible(sycl_iterator<_Mode1, _T, _Allocator> __it1,
```
MikeDvorskiy, line 183: Why just `operator==` has not been added to sycl_iterator?

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,77 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+{
+    const auto buf1 = __it1.get_buffer();
+    const auto buf2 = __it2.get_buffer();
+
+    // We are unable to compare two sycl_iterator's from different buffers
+    if (buf1 != buf2)
+        return false;
+
+    // We are unable to compare two sycl_iterator's if one of them is sub_buffer
+    if (buf1.is_sub_buffer() || buf2.is_sub_buffer())
```
MikeDvorskiy, line 194: SYCL specific code os not possible here, on `dpl/pstl/hetero` level, just on `dpl/pstl/hetero/dpcpp` level.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
SergeyKopienko: Because as said @akukanov, the implementation of sycl_iterator is technically correct. And I agree with this.

MikeDvorskiy (dismissed):
I suggest adding a proper `operator==` to sycl_iterator class and use it instead of two-step comparison via `__check_if_iterator_equality_is_possible` and existing `operator==`.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,77 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+    else
+    {
+        // 2. we should be able to compare two iterators in run-time:
+        //    for example they may have different types or iterate different contaqiners and so on.
+        if (!__check_if_iterator_equality_is_possible(__it1, __it2))
+            return false;
 
-    return false;
+        // 3. Now we are ready to compare two iterators
+        return __it1 == __it2;
```
danhoeflinger, line 224: I don't think splitting `__check_if_iterator_equality_is_possible()` out from `__check_equal_iterators()` is the best approach.  

I think we must assume that iterators are equal whenever they might be for the purposes of deciding to use a more conservative algorithm at a performance penalty.  I would embed the iterator comparison from L224 in with `__check_if_iterator_equality_is_possible()`, but only call it after we know that the buffers are equal and neither is a sub_buffer.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,77 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
 __check_equal_iterators(_Iterator1 __it1, _Iterator2 __it2)
 {
-    // In-place exclusive scan works correctly only if an input and an output iterators are the same type.
-    // Otherwise, there is no way to check an in-place case and a workaround below is not applied.
-    if constexpr (::std::is_same_v<::std::decay_t<_Iterator1>, ::std::decay_t<_Iterator2>>)
+    // 1. operator==(_Iterator1, _Iterator2) should be available in compile-time
+    if constexpr (!is_equality_comparable<_Iterator1, _Iterator2>::value)
     {
-        return __it1 == __it2;
+        return false;
```
danhoeflinger, line 214: Just because iterators are not equality comparable does not mean they not equal, it means "we don't know".  It seems like this may lead to being too aggressive with the faster algorithm when we are not sure if the algorithm is in-place or not.

I think when they are not comparable, it means they may be equal and we must pick the conservative algorithm.

SergeyKopienko:
> I suggest adding a proper `operator==` to sycl_iterator class and use it instead of two-step comparison via `__check_if_iterator_equality_is_possible` and existing `operator==`.

I think that we should not change the sycl_iterator's behavior: we can get undesirable effects.
@akukanov, @danhoeflinger, what do you think about this ?

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
akukanov: I rather disagree. I think the iterator implementation can, and should, guarantee that any two instances created for the same sequence are comparable. Thus, if instances are not comparable, we consider them being created for different sequences (buffers) and therefore we do know that the algorithm does not run in place.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
akukanov: True, the current implementation of iterators is technically correct. But if we extend it to also check the equality of the buffers, it will also be technically correct, just somewhat slower.

The problem, however, is that such implementation would provide stronger comparison guarantees than those of a typical C++ iterator (and likely also stronger than what we want to document in the specification). Therefore algorithms that work with generic iterators would still not be able to use these guarantees, and would have to first ensure that performing the comparison would be valid.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,77 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+
+    return true;
+}
+
+template <typename _Iterator1, typename _Iterator2>
+bool
+__check_if_iterator_equality_is_possible(_Iterator1, _Iterator2)
+{
+    return true;
+}
```
akukanov, line 205: I think for arbitrary iterators the check should return false, as we cannot assume two instances can be compared.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
danhoeflinger: Yes, I understand your point on iterators in general.  I think I was wrong for the general case here.

However, on that note, I think that our `sycl_iterator` implementation breaks this rule when we take sub_buffers into account.

Psuedo-code example:

```
buffer A(n);
buffer B(A, m); // sub_buffer with offset m

sycl_iterator iter1 = begin(A) + m;
sycl_iterator iter2 = begin(B);
sycl_iterator iter3 = begin(B) + m;

iter1 == iter2; // this is will return false, but these point to the same data 
//We only use comparison of idx and don't take a sub_buffer's offset into account

iter1 == iter3; // this is will return true, but these point to different data in the same parent buffer
```
I would argue that neither of these comparisons are valid and these iterators are not comparable because they are based on different buffers.  However, they **can** point to the same underlying data.  I think we can live with this, but we have to be aware of it in this context.

I don't think there is a way to extract the starting offset from a sub_buffer, but I'd be happy to be wrong about that.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,77 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+};
+
+// All with implemented operator ==
 template <typename _Iterator1, typename _Iterator2>
-constexpr bool
+struct is_equality_comparable<_Iterator1, _Iterator2,
+                              std::void_t<decltype(::std::decay_t<_Iterator1>() == ::std::decay_t<_Iterator2>())>>
+    : std::true_type
+{
+};
```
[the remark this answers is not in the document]
akukanov: And even simpler, you can define the main template for two iterators to be equivalent to (or inherit from) `std::is_same`

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
akukanov: You are right about the sub-buffers, but the same considerations equally apply to many other algorithms, which might have an incorrect behavior if they are used with overlapping sub-buffers. And I believe such usage violates the requirements of the C++ standard (or at least the expectations) that algorithms apply to non-overlapping sequences or, in case of in-place usage, to exactly the same sequence passed more than one time to the call.

In other words, I believe applying an algorithm to an overlapping sequence, no matter how it is created, results in undefined behavior (i.e. we can assume it does not happen in a correct program). Perhaps it deserves being documented somewhere. 

So, this covers the case of `iter1 != iter2` you described. And for the case of `iter1 == iter3` while they are actually not equal - this is no different than comparing iterators over different buffers, is not it?

akukanov: In short, I think what I said above means that we consider the sequences for sub-buffers over the same buffer being different, and if they actually overlap, passing those to an algorithm would result in undefined behavior.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
akukanov: And actually, why cannot we compare two iterators if one of them is over a sub-buffer?

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
danhoeflinger: This gets a little tricky, because the `operator==` we are using to compare buffers compares the root buffers.
Therefore, its not easy to determine if we have the same sub-buffer or different sub-buffers.

akukanov: I see.

OK, then what we get in practice? The actual data represented by sub-buffers may match, overlap, or not overlap within the "root" buffer. If they match, iterators obtained with begin() will compare equal, and in-place is correctly detected. If they overlap, it's undefined behavior, so we simply ignore this case. If they do not overlap, iterators will still compare equal, and we use the more pessimistic in-place algorithm - but still everything works correctly. I'd say it is acceptable. In any case, it seems better than refusing to compare iterators if any of them is over a sub-buffer (I'll leave the detailed analysis of that case as an exercise :)). Do we have other ways to deal with sub-buffers?

danhoeflinger: I think we have two options for `sycl_iterators` and `__check_if_iterator_equality_is_possible()`:

1) Only check if the buffers are `operator==`, FALSE: out-of-place, TRUE: in-place
or 
2) Check if the buffers are `operator==` FALSE: out-of-place True: continue
    Check if either buffer is a sub-buffer TRUE: in-place  False: continue
    Make (now valid) check of `sycl_iterator` `operator==`  TRUE: in-place, FALSE: out-of-place

(2) Is more complicated but allows us to catch the one (likely rare) extra case where two `sycl_iterators` are non-overlapping, but both from the same "root" buffer.

I'd be fine with either choice.  However, I don't believe we can know `sycl_iterator` `operator==` is valid unless we confirm that sub-buffers are not involved.  Without validity we can't trust the result in either direction.

akukanov: OK. I think I am fine with the option (2).

test/parallel_api/numeric/numeric.ops/scan_2.pass.cpp
```diff
@@ -321,6 +321,42 @@ run_test()
         test1buffer<alloc_type, test_scan_inplace<ValueType, TestingAlgoritmExclusiveScanExt<2, BinaryOperation>>>();
     }
 }
+
+void
+test_equality_of_iterators()
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed (removed)

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
SergeyKopienko: We have 3-level check of two iterators:

1 level - we check in compile time that we are able to compare two iterators:
```
    // 1. operator==(_Iterator1, _Iterator2) should be available in compile-time
    if constexpr (!is_equality_comparable<_Iterator1, _Iterator2>::value)
    {
        return false;
    }
```
2 level - we check some additional requirements for compare two iterators:
 - for sycl_iterator - that they describes the same buffer and so on:
```
template <sycl::access::mode _Mode1, sycl::access::mode _Mode2, typename _T, typename _Allocator>
bool
__check_if_iterator_equality_is_possible(sycl_iterator<_Mode1, _T, _Allocator> __it1,
                                         sycl_iterator<_Mode2, _T, _Allocator> __it2)
{
    // ...
}
```
 - for iterators of all other types there are no requirements:
```
bool
__check_if_iterator_equality_is_possible(_Iterator1, _Iterator2)
{
    return true;
}
```

3 level - we compare two iterators and return result:
```
        // 3. Now we are ready to compare two iterators
        return __it1 == __it2;
```

So up to me the function seems correct:
```
bool
__check_if_iterator_equality_is_possible(_Iterator1, _Iterator2)
{
    return true;
}
```

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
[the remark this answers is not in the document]
SergeyKopienko: You mean that we can compare two iterators of the same type.
But I mean that we should also check the case, for example, for types ```const int*``` and ```int *```
The types are different but they able to describe the same source data and they are comparable.

My implementation covered this case, your proposal (as far as I understood) - not covered this case.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
SergeyKopienko: Fixed: option (2) implemented.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
SergeyKopienko: Fixed: if one is sub_buffer, continue like in case when iterators are equal.

danhoeflinger (commented):
Functionally, I think this is correct as it is.  If others are OK with it as is, I could live with it.  
I've pointed out two issues I have above, and here is an explanation of a refactor which should solve both.

---

For the purposes of this explanation, I will call call the existing implementation of `__check_if_iterator_equality_is_possible`, `__check_if_iterator_equality_is_possible_OLD`.

`__check_if_iterator_equality_is_possible` should be the name of the top level function which is called by `__pattern_transform_scan_base` (replacing `__check_equal_iterators`). This solves the semantics issue when `sycl_iterators` are involved, but still selects the appropriate algorithm.

`__check_if_iterator_equality_is_possible` should have two overloads, one for `sycl_iterators` and another for the general iterator case.

The `sycl_iterator` overload of the new `__check_if_iterator_equality_is_possible` can skip the `is_equality_comparable` section (always true), and implement the contents of `__check_if_iterator_equality_is_possible_OLD` directly, then add the `sycl_iterator` `operator==` call at the end once we have determined it is a valid comparison.  This resolves the current issues with `bIsEqual`, since it will now be encapsulated in its proper scope and is only active for `sycl_iterators` where it is relevant.

The general iterator overload of `__check_if_iterator_equality_is_possible` should use the current `is_equality_comparable` functionality, but skip the current `__check_if_iterator_equality_is_possible_OLD` (always true), then merely compare the iterators with an `operator==`.  One could add a comment here mentioning that we know exactly if these iterators are equal or not in this case.

The `is_equality_comparable<sycl_iterator>` overload can be deleted, as can the `__check_if_iterator_equality_is_possible_OLD(_Iterator)` overload as they are no longer needed.

One thing we lose with this is that previously we were reusing the expression  `__it1 == __it2`, for both sets of inputs and now we are not.  However, in the process we get to delete a pair of functions where each was really to serve only one of the two sets of inputs.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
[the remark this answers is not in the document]
akukanov: As I said, decay the types before the use of the structure.

```
if constexpr (!is_equality_comparable<std::decay_t<_Iterator1>, std::decay_t<_Iterator2>>::value) {...}
```

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,59 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+
+// Two sycl_iterator's (with different access modes)
+template <sycl::access::mode _Mode1, sycl::access::mode _Mode2, typename _T, typename _Allocator>
+struct is_equality_comparable<sycl_iterator<_Mode1, _T, _Allocator>, sycl_iterator<_Mode2, _T, _Allocator>>
+    : std::true_type
+{
+};
+
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed: unrequired overload was removed

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,85 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+    else
+    {
+        // 2. we should be able to compare two iterators in run-time:
+        //    for example they may have different types or iterate different contaqiners and so on.
+        bool bIsEqual = false;
+        if (!__check_if_iterator_equality_is_possible(__it1, __it2, bIsEqual))
+            return false;
+
+        // 3. Now we are ready to compare two iterators
+        return bIsEqual || __it1 == __it2;
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,77 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+    const auto buf1 = __it1.get_buffer();
+    const auto buf2 = __it2.get_buffer();
+
+    // We are unable to compare two sycl_iterator's from different buffers
+    if (buf1 != buf2)
+        return false;
+
+    // We are unable to compare two sycl_iterator's if one of them is sub_buffer
+    if (buf1.is_sub_buffer() || buf2.is_sub_buffer())
+        return false;
```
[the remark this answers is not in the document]

danhoeflinger (approved):
LGTM

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -152,18 +152,58 @@ __pattern_transform_scan_base_impl(_ExecutionPolicy&& __exec, _Iterator1 __first
...
+// TODO In C++20 we may try to use std::equality_comparable
+template <typename _Iterator1, typename _Iterator2, typename = void>
+struct __is_equality_comparable : std::false_type
+{
+};
+
+// All with implemented operator ==
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable<_Iterator1, _Iterator2,
+                                std::void_t<decltype(::std::decay_t<_Iterator1>() == ::std::decay_t<_Iterator2>())>>
```
akukanov, line 164:
```suggestion
                                std::void_t<decltype(::std::decay_t<_Iterator1>{} == ::std::decay_t<_Iterator2>{})>>
```

Ideally, we should use `declval` here, but I think it is safe to expect iterator types being default-constructible.

akukanov:
I have not checked the new test. The implementation is OK now.

danhoeflinger (commented):
I took a closer look at the testing and found a few things to address.

test/general/sycl_iterator/sycl_iterator_operations.pass.cpp
```diff
@@ -0,0 +1,110 @@
...
+        EXPECT_FALSE(__iterators_possibly_equal(intData, floatData), "wrong __iterators_possibly_equal result");
+    }
+
+    {
+        int srcIntData = 0;
+        const auto& intConstData = srcIntData;
+        auto& intData = srcIntData;
+        float floatData = .0;
+
+        //intConstData = 1; // OK: error: cannot assign to variable 'intConstData' with const-qualified type 'const int &'
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed (removed)

test/general/sycl_iterator/sycl_iterator_operations.pass.cpp
```diff
@@ -0,0 +1,110 @@
...
+                 "wrong __iterators_possibly_equal result");
+    EXPECT_FALSE(__iterators_possibly_equal(oneapi::dpl::begin(buf1), oneapi::dpl::begin(buf2)),
+                 "wrong __iterators_possibly_equal result");
+    EXPECT_FALSE(__iterators_possibly_equal(it1, oneapi::dpl::begin(buf2)), "wrong __iterators_possibly_equal result");
+
+    EXPECT_FALSE(__iterators_possibly_equal(oneapi::dpl::begin(buf1), nullptr),
+                 "wrong __iterators_possibly_equal result");
+    EXPECT_FALSE(__iterators_possibly_equal(nullptr, oneapi::dpl::begin(buf2)),
+                 "wrong __iterators_possibly_equal result");
+
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

danhoeflinger (approved):
LGTM now with the testing changes.  Thanks

PR #853, opened 2023-03-22
Restructure transform_scan to fix FPGA compile error

adamfidel:
This PR restructures the internal transform_scan code towards the following goals:
- Consistency with the other single-group algorithms (e.g., reduce) with respect to what functionality is in the numeric header and what functionality is in the SYCL implementation header
- Fixing the compile errors with scan in FPGA emulation mode

### Current Design

The following diagram shows the current design for the internal functions of scan:

scan_code_layout-current drawio

In the above design, there is inconsistency with some `__pattern_*` functions appearing in the numeric header and some appearing in the SYCL backend header. Additionally, there is not a clear distinction between the scan pattern code and the backend code, whether it is the SYCL device backend or FPGA backend.

### Proposed Design

The new design places the SYCL device implementation in the SYCL backend header, which allows the FPGA implementation to easily defer to the device implementation. This creates a cleaner separation between the pattern and the implementations themselves.

scan_code_layout-proposed

### Other Changes

This PR also incorporates the changes from https://github.com/oneapi-src/oneDPL/pull/810, which include minor changes to forwarding references and some cleanup.


adamfidel:
All tests from `build-onedpl-tests` target now compile with both FPGA and device backend.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
```diff
@@ -110,20 +110,35 @@ __parallel_transform_reduce(_ExecutionPolicy&& __exec, _ReduceOp __reduce_op, _T
...
+auto
+__parallel_transform_scan(_ExecutionPolicy&& __exec, _Range1&& __in_rng, _Range2&& __out_rng, ::std::size_t __n,
+                          _UnaryOperation __unary_op, _InitType __init, _BinaryOperation __binary_op, _Inclusive)
+{
+    // workaround until we implement more performant version for patterns
+    using _Policy = typename ::std::decay<_ExecutionPolicy>::type;
+    using __kernel_name = typename _Policy::kernel_name;
+    auto __device_policy = oneapi::dpl::execution::make_device_policy<__kernel_name>(__exec.queue());
+    return oneapi::dpl::__par_backend_hetero::__parallel_transform_scan(
+        __device_policy, ::std::forward<_Range1>(__in_rng), ::std::forward<_Range2>(__out_rng), __n, __unary_op, __init,
```
timmiesmith, line 125: Should `__device_policy` be forwarded as well?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -645,18 +626,63 @@ __pattern_transform_scan_single_group(_ExecutionPolicy&& __exec, _InRng&& __in_r
...
+        if (__n <= __single_group_upper_limit && __max_slm_size >= __req_slm_size)
+        {
+            return __parallel_transform_scan_single_group(
+                std::forward<_ExecutionPolicy>(__exec), ::std::forward<_Range1>(__in_rng), ::std::forward<_Range2>(__out_rng), __n, __unary_op, __init, __binary_op,
+                _Inclusive{});
+        }
+    }
+
+    // Either we can't use group scan or this input is too big for one workgroup
+    using _Type = typename _InitType::__value_type;
```
timmiesmith, line 685: `_Type` appears to be unused.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -241,14 +193,16 @@ __pattern_transform_scan_base(_ExecutionPolicy&& __exec, _Iterator1 __first, _It
...
         // Run main algorithm and save data into temporary buffer
-        __pattern_transform_scan_base_impl(__policy, __first, __last, __first_tmp, __unary_op, __init, __binary_op,
-                                           _Inclusive{});
+        oneapi::dpl::__par_backend_hetero::__parallel_transform_scan(__policy, __buf1.all_view(), __buf2.all_view(), __n, __unary_op, __init, __binary_op,
+                                           _Inclusive{}).wait();
 
         // Move data from temporary buffer into results
         oneapi::dpl::__internal::__pattern_walk2_brick(
-            ::std::forward<_NewExecutionPolicy>(__policy), __first_tmp, __last_tmp, __result,
+            ::std::move(__policy), __first_tmp, __last_tmp, __result,
```
MikeDvorskiy, line 205: Why do we do such change?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -666,8 +692,8 @@ __pattern_transform_scan_multi_group(_ExecutionPolicy&& __exec, _InRng&& __in_rn
     _NoAssign __no_assign_op;
     _NoOpFunctor __get_data_op;
 
-    return oneapi::dpl::__par_backend_hetero::__parallel_transform_scan(
-        ::std::forward<_ExecutionPolicy>(__exec), __in_rng.all_view(), __out_rng.all_view(), __binary_op, __init,
+    return __future(__parallel_transform_scan_multi_group(
```
MikeDvorskiy, line 695: Is `__future ` created here just by `event`? I think it is not quite correct.
Why do we no return a future returned by `__parallel_transform_scan_multi_group`?

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
adamfidel: `__policy` is created a few lines above, so forwarding it does not make sense because it is not passed in as a parameter. It is not used after is line, so a `move` here should be fine.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -835,7 +835,7 @@ __pattern_scan_copy(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __
         oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _IteratorOrTuple>();
     auto __buf2 = __keep2(__output_first, __output_first + __n);
 
-    auto __res = __par_backend_hetero::__parallel_transform_scan(
+    auto __res = __par_backend_hetero::__parallel_transform_scan_multi_group(
```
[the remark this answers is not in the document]
adamfidel: What used to be called `__parallel_transform_scan` did much more than just a generic transform scan: it received higher-order functions for a work-group level scan, a scan between work-groups and a global scan on the work-group scan's results. This by definition can not use a single-work group optimization because it by design is for multiple work groups. Most of the algorithms that called this function (such as `copy_if`) relied on overloading the higher-order functions and thus cannot easily be optimized for single work groups.

Because of this `__parallel_transform_scan` has been renamed to `__parallel_transform_scan_multi_group` to make it apparent that this is by design for multiple work groups. It also make it clearer that calling algorithms such as `copy_if` are explicitly relying on multiple work group behavior.

The `__parallel_transform_scan` algorithm now just does a

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: There are two branches in this function that both return futures. The first branch for the single-work-group version of the algorithm returns a future with no value, since a value is not needed. This is the second branch which calls the multiple work-group version of the algorithm that returns a future with a value. Since the value is not needed in the general case for transform scan, we drop the value here and return a future with just the event.

We cannot modify the multiple work group version's future to not return a value because it used by other algorithms that directly call that version, such as `copy_if`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
```diff
@@ -110,20 +110,35 @@ __parallel_transform_reduce(_ExecutionPolicy&& __exec, _ReduceOp __reduce_op, _T
...
+        __binary_op, _Inclusive{});
+}
+
 template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _BinaryOperation, typename _InitType,
           typename _LocalScan, typename _GroupScan, typename _GlobalScan,
           oneapi::dpl::__internal::__enable_if_fpga_execution_policy<_ExecutionPolicy, int> = 0>
 auto
-__parallel_transform_scan(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2, _BinaryOperation __binary_op,
+__parallel_transform_scan_multi_group(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2, _BinaryOperation __binary_op,
```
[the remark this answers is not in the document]
adamfidel: I believe I answered this question in my response to your previous question, but the `__parallel_transform_scan_multi_group` function is here because some algorithms (e.g., `copy_if`) rely on semantics provided solely by the the multiple work group version of the algorithm and cannot easily be modified to call the single-work-group version.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
```diff
@@ -110,20 +110,35 @@ __parallel_transform_reduce(_ExecutionPolicy&& __exec, _ReduceOp __reduce_op, _T
...
 {
     // workaround until we implement more performant version for patterns
     using _Policy = typename ::std::decay<_ExecutionPolicy>::type;
     using __kernel_name = typename _Policy::kernel_name;
     auto __device_policy = oneapi::dpl::execution::make_device_policy<__kernel_name>(__exec.queue());
-    return oneapi::dpl::__par_backend_hetero::__parallel_transform_scan(
-        __device_policy, ::std::forward<_Range1>(__rng1), ::std::forward<_Range2>(__rng2), __binary_op, __init,
-        __local_scan, __group_scan, __global_scan);
+    return oneapi::dpl::__par_backend_hetero::__parallel_transform_scan_multi_group(
+        __device_policy, ::std::forward<_Range1>(__rng1), ::std::forward<_Range2>(__rng2), __binary_op, __init, __local_scan, __group_scan, __global_scan);
```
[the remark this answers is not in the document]
adamfidel: Yes, I will change to forward the policy.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
adamfidel: Yes, will change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: Will remove.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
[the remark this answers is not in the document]
julianmi: I see. Thanks for clarifying. Should we comment this behavior in code then? I think this is quite different to the other algorithms. Especially since `copy_if` doesn't follow the sketch of the proposed design above.

Further, I think the naming of `__parallel_transform_scan`, `__parallel_transform_scan_single_group`, `__parallel_transform_scan_multi_group` could be improved to better distinguish them. Especially the naming of the latter two suggests that the only difference is how many work groups are used which doesn't seem to be the case.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -173,6 +173,13 @@ make_wrapped_policy(_Policy&& __policy)
         ::std::forward<_Policy>(__policy));
 }
 
+template <template <typename> class _NewKernelName, typename _Policy,
+          oneapi::dpl::__internal::__enable_if_host_execution_policy<_Policy, int> = 0>
+_Policy make_wrapped_policy(_Policy&& __policy)
+{
+    return __policy;
+}
+
```
danhoeflinger, line 182: It seems like this was added to support some usages that have since been removed.  It seems like this is now dead code.

Is there any functionality for host policies that needs to be replaced in the `rebind_policy` calls?

adamfidel: You are correct that this is just dead code now. I have removed it.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -97,60 +97,6 @@ __pattern_transform_reduce(_ExecutionPolicy&& __exec, _ForwardIterator __first,
...
-                .wait();
-        }
-    }
-    else
-    {
-        oneapi::dpl::__par_backend_hetero::__pattern_transform_scan_multi_group(
-            std::forward<_ExecutionPolicy>(__exec), __buf1, __buf2, __unary_op, __init, __binary_op, _Inclusive{})
-            .wait();
-    }
-}
```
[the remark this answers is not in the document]
adamfidel: Good catch. I updated the diagram to reflect this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: It seems that the general confusion here and [above](https://github.com/oneapi-src/oneDPL/pull/853/files#r1206662794) both stem from the fact that `__pattern_transform_scan_multi_group` is called in two ways:  

1. The simple case which does not need a return value called here
2. The general case where more is embedded into the provided functors, and a value is expected in the `__future` return (like `copy_if`).  

Perhaps it makes sense to have a two functions to match these two use cases:  (1) which which serves the simple case called here, drops the value from the future with a comment, sets up the local, between group and global scans, and otherwise just calls the (2) which returns a future with a value and has a different name.  

I don't have good names for these functions, brainstorming some:
1. `__parallel_transform_scan_multi_group`
2. `__parallel_transform_scan_`  +  `general_purpose`, `builder`, or `_custom`

danhoeflinger: To be clear, this would add a layer to the new diagram between `__parallel_transform_scan_multi_group` and `__parallel_scan_submitter` which would be (2) above.  That function would be the entry point for `copy_if` and other patterns which need the custom scan functionality.

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
```diff
@@ -74,21 +74,17 @@ exclusive_scan_by_segment_impl(Policy&& policy, InputIterator1 first1, InputIter
...
               internal::replace_if_fun<OutputType, ::std::negate<FlagType>>(::std::negate<FlagType>(), init));
 #else
     replace_copy_if(policy1, first2, last2 - 1, _flags.get() + 1, _temp.get() + 1, ::std::negate<FlagType>(), init);
 #endif
 
-    typename internal::rebind_policy<policy_type, ExclusiveScan2<policy_type>>::type policy2(policy);
-
     // scan key-flag tuples
-    inclusive_scan(policy2, make_zip_iterator(_temp.get(), _flags.get()),
+    inclusive_scan(policy, make_zip_iterator(_temp.get(), _flags.get()),
```
danhoeflinger, line 87: Should be able to forward `policy`, right?

include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
```diff
@@ -64,8 +64,8 @@ inclusive_scan_by_segment_impl(Policy&& policy, InputIterator1 first1, InputIter
     transform(policy, first1, last1 - 1, first1 + 1, _mask.get() + 1,
               oneapi::dpl::__internal::__not_pred<BinaryPredicate>(binary_pred));
 
-    typename internal::rebind_policy<policy_type, InclusiveScan1<policy_type>>::type policy1(policy);
-    inclusive_scan(policy1, make_zip_iterator(first2, _mask.get()), make_zip_iterator(first2, _mask.get()) + n,
+
+    inclusive_scan(policy, make_zip_iterator(first2, _mask.get()), make_zip_iterator(first2, _mask.get()) + n,
```
danhoeflinger, line 68: forward `policy` ?

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
```diff
@@ -145,7 +141,7 @@ exclusive_scan_by_segment_impl(Policy&& policy, InputIterator1 first1, InputIter
     replace_copy_if(policy1, first2, last2 - 1, _flags.get() + 1, _temp.get() + 1, ::std::negate<FlagType>(), init);
```
danhoeflinger, line 141: move `policy1`?

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
```diff
@@ -145,7 +141,7 @@ exclusive_scan_by_segment_impl(Policy&& policy, InputIterator1 first1, InputIter
...
 #    endif
 
-    typename internal::rebind_policy<policy_type, ExclusiveScan2<policy_type>>::type policy2(policy);
+    auto policy2 = oneapi::dpl::__par_backend_hetero::make_wrapped_policy<ExclusiveScan2>(policy);
 
     // scan key-flag tuples
     transform_inclusive_scan(policy2, make_zip_iterator(_temp.get(), _flags.get()),
```
danhoeflinger, line 147: move `policy2`?

include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
```diff
@@ -102,7 +102,7 @@ inclusive_scan_by_segment_impl(Policy&& policy, InputIterator1 first1, InputIter
     transform(::std::forward<Policy>(policy), first1, last1 - 1, first1 + 1, _mask.get() + 1,
               oneapi::dpl::__internal::__not_pred<BinaryPredicate>(binary_pred));
 
-    typename internal::rebind_policy<policy_type, InclusiveScan1<policy_type>>::type policy1(policy);
+    auto policy1 = oneapi::dpl::__par_backend_hetero::make_wrapped_policy<InclusiveScan1>(policy);
     transform_inclusive_scan(policy1, make_zip_iterator(first2, _mask.get()),
```
danhoeflinger, line 106: move `policy1`?

include/oneapi/dpl/internal/reduce_by_segment_impl.h
```diff
@@ -126,8 +124,7 @@ reduce_by_segment_impl(Policy&& policy, InputIterator1 first1, InputIterator1 la
     CountType N = scanned_tail_flags[n - 1] + 1;
 
     // scatter the keys and accumulated values
-    typename internal::rebind_policy<policy_type, Reduce3<policy_type>>::type policy3(policy);
-    oneapi::dpl::for_each(policy3, make_zip_iterator(first1, scanned_tail_flags, mask, scanned_values, mask + 1),
+    oneapi::dpl::for_each(policy, make_zip_iterator(first1, scanned_tail_flags, mask, scanned_values, mask + 1),
```
danhoeflinger, line 127: move `policy`?

danhoeflinger (commented):
Some more policy forwarding / moving changes.

I think in general it looks pretty good. I think it would benefit from the small restructure and/or renaming for `__parallel_transform_scan_multi_group`, but it seems close to me.

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
adamfidel: Done.

include/oneapi/dpl/internal/reduce_by_segment_impl.h
adamfidel: Done.

include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
adamfidel: Done.

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
adamfidel: Done.

include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
adamfidel: Done.

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -605,7 +586,7 @@ __pattern_transform_scan_single_group(_ExecutionPolicy&& __exec, _InRng&& __in_r
                         __scan_single_wg_kernel<::std::integral_constant<::std::uint16_t, __wg_size>,
                                                 ::std::integral_constant<::std::uint16_t, __num_elems_per_item>,
                                                 /* _IsFullGroup= */ std::true_type, _Inclusive, _CustomName>>>()(
-                    __exec, __in_rng.all_view(), __out_rng.all_view(), __n, __init, __binary_op, __unary_op);
+                    __exec, std::forward<_InRng>(__in_rng), std::forward<_OutRng>(__out_rng), __n, __init, __binary_op, __unary_op);
```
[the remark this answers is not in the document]
adamfidel: Done.

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
danhoeflinger: This is the fault of the `#if 1`, but I think you should move the one in the  `#else` as well.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
danhoeflinger: I think you missed the third one on Line 629 calling `__parallel_transform_scan_dynamic_single_group_submitter`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: After discussing with @rarutyun, we decided to rename `__parallel_transform_scan_multi_group` to `__parallel_transform_scan_base` to make it clearer that this function is a generic version that supports multiple algorithms.

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
adamfidel: Done.

rarutyun (approved):

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -218,12 +164,18 @@ __pattern_transform_scan_base(_ExecutionPolicy&& __exec, _Iterator1 __first, _It
...
     // This is a temporary workaround for an in-place exclusive scan while the SYCL backend scan pattern is not fixed.
     const bool __is_scan_inplace_exclusive = __n > 1 && !_Inclusive{} && __iterators_possibly_equal(__first, __result);
     if (!__is_scan_inplace_exclusive)
     {
-        __pattern_transform_scan_base_impl(__exec, __first, __last, __result, __unary_op, __init, __binary_op,
-                                           _Inclusive{});
+        auto __keep2 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _Iterator2>();
+        auto __buf2 = __keep2(__result, __result + __n);
+
+        oneapi::dpl::__par_backend_hetero::__parallel_transform_scan(__exec, __buf1.all_view(), __buf2.all_view(), __n, __unary_op, __init, __binary_op,
```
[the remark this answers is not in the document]
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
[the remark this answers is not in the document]
adamfidel: Quoting my answer to a previous question here, since it's relevant:

> After discussing with @rarutyun, we decided to rename `__parallel_transform_scan_multi_group` to `__parallel_transform_scan_base` to make it clearer that this function is a generic version that supports multiple algorithms.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Sounds fine to me.  I still think that separating into two functions (one which wraps the other) may be useful in the future, as it can clarify these two uses and their associated differences.

PR #860, opened 2023-03-24
Make some template parameters always required

akukanov:
Let's for now require the WG size and the number of iterations per work item to be explicitly set.
Maybe we will decide on some good default values later.


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_esimd_radix_sort.h
```diff
@@ -90,30 +90,28 @@ radix_sort(_ExecutionPolicy&& __exec, _Range&& __rng)
...
-    oneapi::dpl::experimental::esimd::impl::radix_sort<
-        _ExecutionPolicy, _Range, IsAscending, WorkGroupSize, ItemsPerWorkItem, RadixBits>(
-            ::std::forward<_ExecutionPolicy>(__exec), __rng);
+    oneapi::dpl::experimental::esimd::impl::radix_sort<WorkGroupSize, DataPerWorkItem, IsAscending, RadixBits>
+        (::std::forward<_ExecutionPolicy>(__exec), __rng);
 }
 
-template <typename _ExecutionPolicy, typename _Iterator, bool IsAscending = true, std::uint16_t WorkGroupSize = 256,
-          std::uint16_t ItemsPerWorkItem = 16, std::uint32_t RadixBits = 8>
+template <std::uint16_t WorkGroupSize, std::uint16_t DataPerWorkItem, bool IsAscending = true,
```
dmitriy-sobolev, line 104: What do you think about swapping `bool IsAscending` and `std::uint32_t RadixBits`? All the numeric parameters will be grouped.

dmitriy-sobolev: Nevermind, the users will more likely change soring order rather than the number of bits.

dmitriy-sobolev (approved):

Issue #855, opened 2023-03-24
Cannot use device iterators in oneDPL algorithms


BenBrock:
I'd like to pass device iterators---by which I mean random access iterators that work in device kernels---into oneDPL algorithms.  Currently this doesn't work.

Here's a minimal example of what I'd like to do:

```cpp
// Allocate USM device buffer
int* x_d = sycl::malloc_device<int>(100, q);

// Fill buffer pointed to by `x_d` with data.
. . .

// Create span from buffer
std::span<int> x(x_d, 100);

// Pass iterators from `std::span` into oneDPL reduce.
auto sum = oneapi::dpl::reduce(oneapi::dpl::execution::make_device_policy(q),
                               x.begin(), x.end(), 0, std::plus());
```

[[full code tarball](https://benjaminbrock.net/downloads/device_iterators.tar.gz)]

Here, instead of passing an `int*` to a USM device buffer into oneDPL reduce, I'm passing in the iterator type of `std::span`, which happens to be GCC's `__normal_iterator`.  Currently, this results in a seg fault, I believe because oneDPL is creating a CPU-side copy of the buffer before launching the algorithm.  (And the CPU-side access of a USM device allocation causes a seg fault.)  Looking through some of the oneDPL code, it seems like this is what happens with most iterators, except for raw pointers and some special iterator types.

In this specific example, I could of course call `.data()` instead of `.begin()` to get raw pointers, which would have the desired behavior.  However, I'm interested in using more complicated device iterator types that can't be represented by raw pointers.

Is there any way to have oneDPL directly launch the kernel with my iterators, instead of copying the data CPU-side?

MikeDvorskiy:
Talking about any containers based on USM  - oneDPL supports just std::vector with USM allocator.

https://oneapi-src.github.io/oneDPL/parallel_api/pass_data_algorithms.html

```
Pass Data to Algorithms
You can use one of the following ways to pass data to an algorithm executed with a device policy:
- oneapi:dpl::begin and oneapi::dpl::end functions
- Unified shared memory (USM) pointers and std::vector with USM allocators
- Iterators of host-side std::vector
```

MikeDvorskiy:
> Is there any way to have oneDPL directly launch the kernel with my iterators, instead of copying the data CPU-side?
Currently, just USM pointers or begin/end over a sycl::buffer.

BenBrock:
[...]

Issue #854, opened 2023-03-22
Wrong executing algorithms with cross-device memory


BenBrock:
If I execute a oneDPL algorithm on one device, while one of the outputs is located on another device, the algorithm does not execute successfully.  I've reproduced this on 2- and 4-GPU machines with Xe Link on ORTCE with `inclusive_scan` and `for_each`.

I have written a [minimal example with `inclusive_scan`](https://benjaminbrock.net/downloads/cross_gpu_algoritihms.tar.gz), summarized below.

- `v` points to a USM device memory allocation on GPU 0.

- `ptrs[1]` points to a USM device memory allocation on GPU `1`.

- Execute `inclusive_scan` on the input `[v, v + n)`, writing results to `[ptrs[1], ptrs[1] + n)`

- The output still contains zeros after the algorithm is executed, instead of the correct result.

```cpp
// Allocate USM device buffers on each device, to be used as outputs
std::vector<int*> ptrs;
for (auto&& device : devices) {
  int* ptr = sycl::malloc_device<int>(n, device, context);
  ptrs.push_back(ptr);
}

// Allocate USM device buffer on device 0, to be used as input
int* v = sycl::malloc_device<int>(n, devices[0], context);

// Copy data to input buffer
sycl::queue q(context, devices[0]);
q.memcpy(v, lv.data(), lv.size()*sizeof(int));

for (auto&& ptr : ptrs) {
  sycl::queue q(context, devices[0]);
  oneapi::dpl::execution::device_policy policy(q);

  // Perform `inclusive_scan` on device 0
  // Input buffer is on device 0, output buffer on every device, round robin
  oneapi::dpl::inclusive_scan(policy, v, v + n, ptr);
}
. . .

// Memcpy buffers pointed to by `ptrs` and print them out.
```

Output:
```
(base) bbrock@sdp125072:~/src/distributed-ranges/examples/shp$ ./write_test-dosxx 
Reference: [0, 1, 3, 6, 10, 15, 21, 28, 36, 45]
Result for GPU 0: [0, 1, 3, 6, 10, 15, 21, 28, 36, 45]
Result for GPU 1: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
Result for GPU 2: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
Result for GPU 3: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
```

[...]

PR #861, opened 2023-03-25
Implement `direct_iterator` and `make_direct_iterator`

BenBrock:
Implement `direct_iterator` and `make_direct_iterator`, which allow users to wrap device iterators that should be used directly inside SYCL kernels by oneDPL.  This PR addresses #855 and #854.

This will likely require some work before being accepted, but I just wanted articulate my proposed fix for these issues.

### DIFF
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -887,6 +887,140 @@ class discard_iterator
     difference_type __my_position_;
 };
 
+template <typename _Iter>
+class direct_iterator
+{
+  public:
+    typedef typename ::std::iterator_traits<_Iter>::value_type value_type;
+    typedef typename ::std::iterator_traits<_Iter>::difference_type difference_type;
+    typedef typename ::std::iterator_traits<_Iter>::reference reference;
+    using pointer = iterator;
+
+    using iterator_category = std::random_access_iterator_tag;
+
+    using iterator = direct_iterator<Iter>;
+
+    using is_passed_directly = ::std::true_type;
+
+    direct_iterator(_Iter iter) noexcept : __iter(iter) {}
+    direct_iterator() noexcept = default;
+    ~direct_iterator() noexcept = default;
+    direct_iterator(const direct_iterator&) noexcept = default;
+    direct_iterator&
+    operator=(const direct_iterator&) noexcept = default;
+
+    bool
+    operator==(const direct_iterator&) const noexcept = default;
+    bool
+    operator!=(const direct_iterator&) const noexcept = default;
+
+    iterator
+    operator+(difference_type offset) const noexcept
+    {
+        return iterator(__iter + offset);
+    }
+
+    iterator
+    operator-(difference_type offset) const noexcept
+    {
+        return iterator(__iter - offset);
+    }
+
+    difference_type
+    operator-(iterator other) const noexcept
+    {
+        return __iter - other.__iter;
+    }
+
+    bool
+    operator<(iterator other) const noexcept
+    {
+        return __iter < other.__iter;
+    }
+
+    bool
+    operator>(iterator other) const noexcept
+    {
+        return __iter > __iter;
+    }
+
+    bool
+    operator<=(iterator other) const noexcept
+    {
+        return __iter <= other.__iter;
+    }
+
+    bool
+    operator>=(iterator other) const noexcept
+    {
+        return __iter >= other.__iter;
+    }
+
+    iterator&
+    operator++() noexcept
+    {
+        ++__iter;
+        return *this;
+    }
+
+    iterator
+    operator++(int) noexcept
+    {
+        iterator other = *this;
+        ++(*this);
+        return other;
+    }
+
+    iterator&
+    operator--() noexcept
+    {
+        --__iter;
+        return *this;
+    }
+
+    iterator
+    operator--(int) noexcept
+    {
+        iterator other = *this;
+        --(*this);
+        return other;
+    }
+
+    iterator&
+    operator+=(difference_type offset) noexcept
+    {
+        __iter += offset;
+        return *this;
+    }
+
+    iterator&
+    operator-=(difference_type offset) noexcept
+    {
+        __iter -= offset;
+        return *this;
+    }
+
+    reference operator*() const noexcept { return *__iter; }
+
+    reference operator[](difference_type offset) const noexcept { return reference(*(*this + offset)); }
+
+    friend iterator
+    operator+(difference_type n, iterator iter)
+    {
+        return iter.__iter + n;
+    }
+
+  private:
+    _Iter __iter;
+};
+
+template <typename Iterator>
+direct_iterator<Iterator>
+make_direct_iterator(Iterator iterator)
+{
+    return direct_iterator<Iterator>(iterator);
+}
+
 } // namespace dpl
 } // namespace oneapi
 
--- a/test/general/direct_iterator.pass.cpp
+++ b/test/general/direct_iterator.pass.cpp
@@ -1,0 +1,71 @@
+// -*- C++ -*-
+//===-- direct_iterator.pass.cpp ------------------------------------------===//
+//
+// Copyright (C) Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// This file incorporates work covered by the following copyright and permission
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#include <oneapi/dpl/execution>
+#include <oneapi/dpl/algorithm>
+#include <oneapi/dpl/iterator>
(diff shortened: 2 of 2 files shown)


test/general/direct_iterator.pass.cpp
```diff
@@ -0,0 +1,71 @@
...
+    dpl::make_direct_iterator d_first(p);
+    dpl::make_direct_iterator d_last(p + n);
+
+    auto v_dev = dpl::reduce(d_first, d_last, 0);
+
+    EXPECT_EQ(v_ref, v_dev);
+
+#if __cpp_lib_span >= 202002L
+
+    std::span<T> x(p, n);
```
MikeDvorskiy, line 59: It seems span `x` is not used...

test/general/direct_iterator.pass.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    sycl::queue q(sycl::default_selector_v);
+
+    T* p = sycl::malloc_device<T>(n, q);
+
+    q.memcpy(p, v.data(), n * sizeof(T)).wait();
+
+    auto v_ref = std::reduce(v.begin(), v.end(), 0);
+
+    dpl::make_direct_iterator d_first(p);
```
MikeDvorskiy, line 50: `p` is a pointer here.. A pointer is passed directly by oneDPL design. A pointer doesn't  require a wrapper...

test/general/direct_iterator.pass.cpp
BenBrock: Oops, just fixed that typo to actually initialize `s_first` and `s_last` using `x.begin()` and `x.end()`.

test/general/direct_iterator.pass.cpp
BenBrock: Happy to remove this part of the test if you prefer.  My idea was that this could serve as a temporary workaround for #854.

(Although as I mention in the issue, there is unfortunately a bug in the level zero that keeps this workaround from working on Intel multi-GPU systems.)

test/general/direct_iterator.pass.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+    auto v_dev = dpl::reduce(d_first, d_last, 0);
+
+    EXPECT_EQ(v_ref, v_dev);
+
+#if __cpp_lib_span >= 202002L
+
+    std::span<T> x(p, n);
+
+    dpl::make_direct_iterator s_first(x.begin());
```
MikeDvorskiy, line 61: Probably we can avoid this "identical" iterator-wrapper, by introducing just specialization for the trait `oneapi::dpl::__internal::is_passed_directly<_Iter>` ?

like `oneapi::dpl::__internal::is_passed_directly<[_Iter](std::span::iterator)>`, with returns `std::true_type` ?

akukanov:
I wonder if providing a wrapper iterator for the purpose of only passing something as-is to oneDPL, is the right approach. Instead, should we maybe follow the approach we use for SYCL buffers, i.e. provide "wrapper" functions that return some object suitable to pass the original class to oneDPL algorithms, without attempting to make it a correct functioning iterator? Or in other words, should we extend the applicability of `dpl::begin()/end()` beyond SYCL buffers to other containers of interest?

test/general/direct_iterator.pass.cpp
BenBrock: You don't always want to pass in `std::span` iterators directly, since they might not be accessible on the device.  Suppose a user wrote the following:

```cpp
std::vector<int> v(...);
std::span s(v);

// Runtime error, since `s.begin()` is a host iterator and cannot
// be used directly on the device.
dpl::reduce(policy, s.begin(), s.end());
```

You can't in general know whether a span is accessible on the device, and this holds for most ranges you might encounter.  There are a lot of iterator types that users might want to pass into oneDPL directly, and I don't think we can automatically most of them.  I will add a better motivating example below.

BenBrock:
Here's a better example illustrating why I think this is needed.  There are potentially many complicated iterator types users will want to pass directly into oneDPL algorithms, and it's not always possible to identify which ones can and can't be passed in directly.

Suppose you wanted to implement a ranges-style dot product using oneDPL, like below.

```cpp
template <std::ranges::forward_range X, std::ranges::forward_range Y>
auto dot_product_onedpl(sycl::queue q, X &&x, Y &&y) {
  auto z = std::ranges::views::zip(x, y)
         | std::ranges::views::transform(
           [](auto &&elem) {
             auto &&[a, b] = elem;
             return a * b;
           });

  oneapi::dpl::execution::device_policy policy(q);

  shp::__detail::direct_iterator d_first(z.begin());
  shp::__detail::direct_iterator d_last(z.end());
  return oneapi::dpl::experimental::reduce_async(
             policy, d_first, d_last, std::ranges::range_value_t<X>(0), std::plus())
      .get();
}
```

The iterator type passed into oneDPL is rather complicated.  It's not in general possible to know whether a transform view or a zip view is directly accessible on the device.  As a user, I happen to know that `X` and `Y` are device ranges, and so I know that the resulting view `z` can safely be used on the device.

This gets more complicated when you also have user-defined data structures and views.

We will definitely need to keep something like this in our own codebase for distributed ranges.  I'll leave it up to you guys to decide whether something like this is more broadly applicable to users.  My intention is basically to give users the option of forcing oneDPL algorithms to use device iterators directly on the device.  This can be used in cases where either it's not possible to determine whether an iterator can be passed directly (`span` and a few other views) as well as when complex iterator types make using `is_passed_directly` a bit unwieldy (most views and some user-defined types).

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -887,6 +887,146 @@ class discard_iterator
...
+    }
+
+    iterator&
+    operator-=(difference_type offset) noexcept
+    {
+        __iter -= offset;
+        return *this;
+    }
+
+    reference operator*() const noexcept { return *__iter; }
```
MikeDvorskiy, line 1009: To tell the truth, I really don't  understand an essence of that wrapper over `_Iter`.
That wrapper repeats the all standard RA iterator functionality, including dereferencing. If `_Iter` is not accessible on a device, `direct_iterator` also is not accessible on a device... So, what's an essence here?

BenBrock: @MikeDvorskiy Sorry for being late getting back to you; this slipped past my inbox.

The idea here is that you have a range/iterator accessible on the device.  Let's say a `std::span<int>` to device memory.  Then, you create a view based on that range.  For example:

```cpp
template <typename T>
auto sum_times_two(std::span<T> x) {
  auto z = x
         | std::ranges::views::transform(
           [](auto &&elem) {
             return elem*2;
           });

  oneapi::dpl::execution::device_policy policy(q);

  return oneapi::dpl::experimental::reduce_async(
             policy, z.begin(), z.end(), T(0), std::plus())
      .get();
}
```

This code works, but has terrible performance.  The reason is that oneDPL does not know `transform_view<...>::iterator` is device accessible, so it copies all the elements one-by-one from the device to the host, then uses a buffer to copy it back to the device.  We use `direct_iterator` to force oneDPL to use the iterator directly, since we know that it can be used directly on the device.

```cpp
template <typename T>
auto sum_times_two(std::span<T> x) {
  auto z = x
         | std::ranges::views::transform(
           [](auto &&elem) {
             return elem*2;
           });

  oneapi::dpl::execution::device_policy policy(q);

  shp::__detail::direct_iterator d_first(z.begin());
  shp::__detail::direct_iterator d_last(z.end());

  return oneapi::dpl::experimental::reduce_async(
             policy, d_first, d_last, T(0), std::plus())
      .get();
}
```

This example is a bit simplified.  In the use case in distributed ranges, we have an actual `device_ptr` as the underlying iterator type, so we do know that the data lives on the device.  It might be worth thinking about how we could integrate distributed range's concepts of device vs. host memory with distributed ranges, but I think there will always be some cases where a user wants to explicitly "promote" a range to being directly accessible on the device.  Using a standard library view is a prime example of this, as we're unlikely to be able to hardwire locality information into a view without modifying the standard.  (Or providing our own implementation of all views.)

masterleinad:
What's left for driving this pull request to completion? Not being able to use custom device iterators is one of the limitations for us in https://github.com/kokkos/kokkos compared with thrust.

BenBrock:
I think the primary blocker is just resources on the oneDPL team.  This was going to be merged as part of #1479, but that's been delayed.

Maybe @akukanov, @rarutyun, or @MikeDvorskiy can comment on the possibility of accepting this PR individually to enable libraries like Kokkos?

akukanov:
Making `is_passed_directly` a public customization point in the oneDPL namespace sounds good to me; better than the proposed iterator wrapper.

I suggest to open a RFC discussion at https://github.com/oneapi-src/oneDPL/discussions and/or a design proposal following the process here https://github.com/oneapi-src/oneDPL/tree/main/rfcs. The goal is to have a dedicated design discussion of this idea. The only thing really needed is motivating use cases, but if you have ideas/preferences for how you would customize this trait, that would be useful for the design. I hope it is not too much to ask you for :)

The eventual outcome should be a patch to the oneDPL specification that describes the new functionality, and a patch to this repo that implements it. But after the design is accepted in principle, we will take care of these unless you will want to stay involved.

PR #863, opened 2023-03-27
Modify to only test feature macro for `<bit>`

BenBrock:
Modify `utils.h` to only check the feature test macro for `<bit>`, instead of also looking at the C++ version.

This was causing a compilation error in the most recent version of intel/llvm compiling with libstdc++ 9.

### DIFF
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -30,9 +30,11 @@
 #    include "hetero/dpcpp/sycl_iterator.h"
 #endif
 
-#if __cpp_lib_bit_cast >= 201806L && __has_include(<bit>)
+#if __has_include(<bit>)
 #    include <bit>
-#else
+#endif
+
+#if !(__cpp_lib_bit_cast >= 201806L)
 #    ifndef __has_builtin
 #        define __has_builtin(__x) 0
 #    endif
@@ -587,7 +589,7 @@ template <typename _Dst, typename _Src>
     sizeof(_Dst) == sizeof(_Src) && ::std::is_trivially_copyable_v<_Dst> && ::std::is_trivially_copyable_v<_Src>, _Dst>
 __dpl_bit_cast(const _Src& __src) noexcept
 {
-#if __cpp_lib_bit_cast >= 201806L && __has_include(<bit>)
+#if __cpp_lib_bit_cast >= 201806L
     return ::std::bit_cast<_Dst>(__src);
 #elif _ONEDPL_BACKEND_SYCL && _ONEDPL_LIBSYCL_VERSION >= 50300
     return sycl::bit_cast<_Dst>(__src);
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/utils.h
```diff
@@ -584,10 +584,10 @@ struct __lifetime_keeper : public __lifetime_keeper_base
 // Bitwise type casting, same as C++20 std::bit_cast
 template <typename _Dst, typename _Src>
 ::std::enable_if_t<
-    sizeof(_Dst) == sizeof(_Src) && ::std::is_trivially_copyable_v<_Dst> && ::std::is_trivially_copyable_v<_Src>, _Dst>
+    sizeof(_Dst) == sizeof(_Src) && ::std::is_trivially_copyable_v<_Dst>&& ::std::is_trivially_copyable_v<_Src>, _Dst>
```
akukanov, line 587: Let's revert these clang-format inspired changes; these are totally unrelated to the patch, and we recently agreed that it's OK for clang-format checks to fail if the suggested changes are useless or bad for code readability.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -30,7 +30,7 @@
 #    include "hetero/dpcpp/sycl_iterator.h"
 #endif
 
-#if _ONEDPL___cplusplus >= 202002L && __has_include(<bit>)
+#if __cpp_lib_bit_cast >= 201806L && __has_include(<bit>)
```
akukanov, line 33: There are two issues with this. 

First, the feature macro is defined, according to the standard, in `<version>` and in `<bit>`. At this point, neither is guaranteed to be included, and each is only available since C++20. That's why the code currently checks for the standard version.

Second, `__cpp_lib_bit_cast` is not the only feature from `<bit>` we need here; see the line 610.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -584,10 +584,10 @@ struct __lifetime_keeper : public __lifetime_keeper_base
 // Bitwise type casting, same as C++20 std::bit_cast
 template <typename _Dst, typename _Src>
 ::std::enable_if_t<
-    sizeof(_Dst) == sizeof(_Src) && ::std::is_trivially_copyable_v<_Dst> && ::std::is_trivially_copyable_v<_Src>, _Dst>
+    sizeof(_Dst) == sizeof(_Src) && ::std::is_trivially_copyable_v<_Dst>&& ::std::is_trivially_copyable_v<_Src>, _Dst>
 __dpl_bit_cast(const _Src& __src) noexcept
 {
-#if __cpp_lib_bit_cast >= 201806L || _ONEDPL___cplusplus >= 202002L && __has_include(<bit>)
+#if __cpp_lib_bit_cast >= 201806L && __has_include(<bit>)
```
akukanov, line 590: In combination with the feature macro (instead of the C++ version check, `__has_include(<bit>)` makes no sense I think, because the feature macro implies that the header should be available.

akukanov (commented):
See comments below, but first of all - could you please add more details about the error you faced and its root case analysis?

BenBrock:
@akukanov Updated the `ifdef` checks as you suggested.  I believe this is ready to merge (and it fixes the issue in the problematic environment).

The CI is failing, but only for clang-format issues unrelated to this PR, so I have left them as is.

akukanov (approved):

PR #864, opened 2023-03-30
Fix equal.pass on platforms non-supporting double type

dmitriy-sobolev:
Some platforms do not support the double type. The testing harness cannot recognize it as a part of a complex type. And thus, the related test cases are not skipped. 

This PR replaces a double type in a complex type. It allows fixing the error:
```
Build program log for 'Intel(R) Iris(R) Xe Graphics
...
error: Double type is not supported on this platform.
in file: equal.pass.cpp:88

```


SergeyKopienko (approved):
LGTM

PR #866, opened 2023-03-30
Rename the implementation directory and the main file for esimd radix sort

akukanov:
Two main goals:
* clearly distinguish from the oneDPL parallel backend implementation
* set the directory as the place for more "kernel templates"

Guard macros are also changed.

Alternative variants are welcome.

### DIFF
--- a/include/oneapi/dpl/experimental/kernel_templates
+++ b/include/oneapi/dpl/experimental/kernel_templates
@@ -12,6 +12,6 @@
 #include "../internal/common_config.h"
 #include "../pstl/onedpl_config.h"
 
-#include "../pstl/hetero/dpcpp/parallel_backend_esimd_radix_sort.h"
+#include "../pstl/hetero/dpcpp/kt_esimd_radix_sort.h"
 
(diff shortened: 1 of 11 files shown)


dmitriy-sobolev (approved):
LGTM

dmitriy-sobolev:
Another option is to place `kt` folder inside `include/oneapi/dpl/experimental` folder so that `intel/oneapi/dpl/pstl/hetero` folder contains files for dpc++ backend only. However, I do not have strong opinion on that.

akukanov:
> Another option is to place kt folder inside include/oneapi/dpl/experimental folder 

Actually that is my preferred option for long term, but I wanted to minimize disruptions for you at this time, and also I do not know how much the new code depends on the core oneDPL implementation. If you prefer to do it now rather than later, I will update the patch.

dmitriy-sobolev:
> If you prefer to do it now rather than later, I will update the patch.

Let's do it now.

dmitriy-sobolev (approved):

PR #871, opened 2023-04-02
Improve utility functions

akukanov:
The patch refactors the SIMD utility functions for load, store, gather, and scatter:

* Removes `enable_if` SFINAE, instead using specific parameter types
* Aligns the API semantics (parameter order, types, and names)
* For `gather`, also fixes the result type to differ from offsets, and moves items-to-bytes conversion to the caller (consistent to `scatter`)

Also it removes unneeded `inline` specifiers and adds missed standard headers.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
@@ -20,28 +20,28 @@ using is_sycl_accessor_v = is_sycl_accessor<T>::value;
 
 template <typename SIMD, typename Input>
 typename ::std::enable_if_t<::std::is_pointer_v<Input>, void>
-simd_load(SIMD& simd, const Input& input, uint32_t offset)
+load_simd(SIMD& simd, const Input& input, uint32_t offset)
 {
     simd.copy_from(input + offset);
 }
 
 template <typename SIMD, typename Input>
 typename ::std::enable_if_t<is_sycl_accessor_v<Input>, void>
-simd_load(SIMD& simd, const Input& input, uint32_t offset)
+load_simd(SIMD& simd, const Input& input, uint32_t offset)
 {
     simd.copy_from(input, offset);
 }
 
 template <typename SIMD, typename Output>
 typename ::std::enable_if_t<::std::is_pointer_v<Output>, void>
-simd_store(const SIMD& simd, Output& output, uint32_t offset)
+store_simd(const SIMD& simd, Output& output, uint32_t offset)
 {
     simd.copy_to(output + offset);
 }
 
 template <typename SIMD, typename Output>
 typename ::std::enable_if_t<is_sycl_accessor_v<Output>, void>
-simd_store(const SIMD& simd, Output& output, uint32_t offset)
+store_simd(const SIMD& simd, Output& output, uint32_t offset)
 {
     simd.copy_to(output, offset);
 }
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
@@ -63,7 +63,7 @@ void
 scatter(T* output, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets,
         sycl::ext::intel::esimd::simd<T, N> values, sycl::ext::intel::esimd::simd_mask<N> mask = 1)
 {
-    return sycl::ext::intel::esimd::scatter(output, offsets * size32<T>, values, mask);
+    sycl::ext::intel::esimd::scatter(output, offsets * size32<T>, values, mask);
 }
 
 template<typename T, int N, sycl::access_mode Mode, sycl::access::placeholder P>
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
@@ -18,7 +18,7 @@ copy_from(T* input, ::std::uint32_t base_offset, sycl::ext::intel::esimd::simd<T
 
 template <typename T, int N, typename... Args>
 void
-copy_from(const sycl::accessor<Args...>& input, ::std::uint32_t base_offset, sycl::ext::intel::esimd::simd<T, N>& values)
+copy_from(const sycl::accessor<T, 1, Args...>& input, ::std::uint32_t base_offset, sycl::ext::intel::esimd::simd<T, N>& values)
 {
     values.copy_from(input, base_offset);
 }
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
```diff
@@ -3,79 +3,73 @@
...
-template<typename T, int N, typename InputT>
-inline
-typename ::std::enable_if_t<::std::is_pointer_v<InputT>, sycl::ext::intel::esimd::simd<T, N>>
-gather(const InputT& input, sycl::ext::intel::esimd::simd<T, N> offsets, uint32_t base_offset)
+template <typename T, int N>
+sycl::ext::intel::esimd::simd<T, N>
+gather(T* input, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets, ::std::uint32_t base_offset)
 {
-    return sycl::ext::intel::esimd::gather(input + base_offset, offsets*static_cast<uint32_t>(sizeof(T)));
+    return sycl::ext::intel::esimd::gather(input + base_offset, offsets);
```
akukanov, line 44: Computing the base address is incorrect if the offset is in bytes (which it is supposed to be).

Perhaps it makes sense to move to the element-based offset in all functions, converting to bytes when required by ESIMD API.

SergeyKopienko (changes requested):
copy_from(const T* input, ::std::uint32_t base_offset, sycl::ext::intel::esimd::simd<T, N>& values)

SergeyKopienko:
@akukanov unfortunately there are many compile errors now

include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
akukanov: Changed to use element-based offsets in the utilities. Note that the accessor variants of copy_to/from now also convert it to bytes - I believe this matches the ESIMD API, essentially fixing a bug.

akukanov:
> @akukanov unfortunately there are many compile errors now

Fixed.

> copy_from(const T* input, ::std::uint32_t base_offset, sycl::ext::intel::esimd::simd<T, N>& values)
> gather(const T* input, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets, ::std::uint32_t base_offset)

Fixed as well.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
```diff
@@ -3,79 +3,73 @@
...
-typename ::std::enable_if_t<std::is_pointer_v<InputT>, void>
-scatter(InputT& input, sycl::ext::intel::esimd::simd<uint32_t, N> offsets,
-        sycl::ext::intel::esimd::simd<T, N> vals, sycl::ext::intel::esimd::simd_mask<N> mask = 1)
+template <typename T, int N>
+void
+scatter(T* output, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets,
+        sycl::ext::intel::esimd::simd<T, N> values, sycl::ext::intel::esimd::simd_mask<N> mask = 1)
 {
-    return sycl::ext::intel::esimd::scatter(input, offsets, vals, mask);
+    return sycl::ext::intel::esimd::scatter(output, offsets, values, mask);
```
SergeyKopienko: ```return``` not required here

include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
```diff
@@ -3,79 +3,84 @@
...
+template <typename T, int N, typename AccessorT>
+typename ::std::enable_if<!::std::is_pointer<AccessorT>::value, sycl::ext::intel::esimd::simd<T, N>>::type
+gather(AccessorT input, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets,::std::uint32_t base_offset)
 {
-    return sycl::ext::intel::esimd::gather<T>(input, offsets*static_cast<uint32_t>(sizeof(T)),
-                                              base_offset*static_cast<uint32_t>(sizeof(T)));
+    //static_assert(::std::is_same<decltype(offsets), decltype(offsets * sizeof(T))>::value, "'offsets' and 'offsets * sizeof(T)' are not the same");
+    //static_assert(::std::is_same<decltype(base_offset), decltype(base_offset * sizeof(T))>::value, "'base_offset' and 'base_offset * sizeof(T)' are not the same");
+
+    return sycl::ext::intel::esimd::gather<T>(input, offsets /** sizeof(T)*/, base_offset /** sizeof(T)*/);
```
dmitriy-sobolev, line 59: The comment is the same as in https://github.com/oneapi-src/oneDPL/pull/871/files#r1156252968, but the imitated API is [that](https://intel.github.io/llvm-docs/doxygen/group__sycl__esimd__memory.html#gab8e33bce282903e5c4e74b009ed0d0c9). Both offsets should be multiplied by `sizeof(T)` because `gather` expect offsets expressed in bytes. Static cast is needed to explicitly pass arguments with types which ESIMD expects.
```suggestion
    return sycl::ext::intel::esimd::gather<T>(input, offsets * static_cast<::std::uint32_t>(sizeof(T)), base_offset * static_cast<::std::uint32_t>(sizeof(T)));
```

SergeyKopienko: Fixed

include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
```diff
@@ -3,79 +3,84 @@
...
+sycl::ext::intel::esimd::simd<T, N>
+gather(const T* input, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets, ::std::uint32_t base_offset)
 {
-    return sycl::ext::intel::esimd::gather(input + base_offset, offsets*static_cast<uint32_t>(sizeof(T)));
+    //static_assert(::std::is_same<decltype(input), decltype(input + base_offset)>::value, "'input' and 'input + base_offset' are not the same");
+    //static_assert(::std::is_same<decltype(offsets), decltype(offsets * sizeof(T))>::value, "'offsets' and 'offsets * sizeof(T)' are not the same");
+    //decltype(offsets)::dummy;
+    //decltype(sizeof(T))::dummy;
+    //decltype(offsets * sizeof(T))::dummy;
+    return sycl::ext::intel::esimd::gather(input + base_offset, offsets/* * sizeof(T)*/);
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

SergeyKopienko (approved):
Now it's compiled and LGTM

akukanov:
Thank you both for the help with fixing the issues. The changes were fine, but I still wanted to completely avoid SFINAE to only allow the API to be used with accessors and pointers, and also I simplified the way to deal with `sizeof` causing SIMD type change.

PR #872, opened 2023-04-03
Improve Performance of Group Reductions with Unknown Identities

julianmi:
This patch improves the throughput of the group reductions with unknown identities. These are used in many algorithms with specialized or used-defined operators or data types including the `min_element()`, `max_element()`, `minmax_element()`, `is_partitioned()`, and `lexicographical_compare()` algorithms.

The main changes are the switch from a while loop to a for loop and the removal of the `__global_idx < __n` check which is covered by the `__global_idx + __k < __n`.

This change preserves the same memory ordering and allows for non-commutative algorithms. While the memory access patterns could be improved by using a decreasing stride instead of the used increasing stride, this would require commutative algorithms.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
@@ -298,7 +298,7 @@ struct reduce_over_group
         auto __local_idx = __item_id.get_local_id(0);
         auto __group_size = __item_id.get_local_range().size();
 
-        for (_Size __k = 1; __k < __group_size; __k <<= 1)
+        for (::std::uint32_t __k = 1; __k < __group_size; __k <<= 1)
         {
             __dpl_sycl::__group_barrier(__item_id);
             if (__local_idx % (2 * __k) == 0 && __local_idx + __k < __group_size && __global_idx + __k < __n)
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
@@ -301,7 +301,7 @@ struct reduce_over_group
         for (::std::uint32_t __k = 1; __k < __group_size; __k <<= 1)
         {
             __dpl_sycl::__group_barrier(__item_id);
-            if (__local_idx % (2 * __k) == 0 && __local_idx + __k < __group_size && __global_idx + __k < __n)
+            if ((__local_idx & (2 * __k - 1)) == 0 && __local_idx + __k < __group_size && __global_idx + __k < __n)
             {
                 __local_mem[__local_idx] = __bin_op1(__local_mem[__local_idx], __local_mem[__local_idx + __k]);
             }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
@@ -298,18 +298,14 @@ struct reduce_over_group
         auto __local_idx = __item_id.get_local_id(0);
         auto __group_size = __item_id.get_local_range().size();
 
-        auto __k = 1;
-
-        do
+        for (_Size __k = 1; __k < __group_size; __k <<= 1)
         {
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -298,18 +298,14 @@ struct reduce_over_group
         auto __local_idx = __item_id.get_local_id(0);
         auto __group_size = __item_id.get_local_range().size();
 
-        auto __k = 1;
-
-        do
+        for (_Size __k = 1; __k < __group_size; __k <<= 1)
```
danhoeflinger, line 301: We are changing the type of `__k` from auto (with `int` literal) to the same type as `__n`, which is likely a `::std::size_t`.  

The type has to fit the next power of 2 higher than the maximum workgroup size.  Does this mean we can fit in `::std::uint16_t`?  `::std::uint32_t` would be safe certainly.

julianmi: I agree that we could fix `__k` to an unsigned integer. I will change it to `::std::uint32_t` to be safe. There seems to be no performance differences between 16 or 32-bit unsigned integers but using an integer with `auto` seems to be slower.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -298,18 +298,14 @@ struct reduce_over_group
...
         {
             __dpl_sycl::__group_barrier(__item_id);
-            if (__local_idx % (2 * __k) == 0 && __local_idx + __k < __group_size && __global_idx < __n &&
-                __global_idx + __k < __n)
+            if (__local_idx % (2 * __k) == 0 && __local_idx + __k < __group_size && __global_idx + __k < __n)
```
mmichel11, line 304: Since `__k` is a power of two, `__local_idx % (2 * __k)` can be computed as `__local_idx & (2 * __k - 1)`. I am not sure if this will have any noticeable benefit but it may be worth trying.

danhoeflinger (dismissed):
LGTM

mmichel11 (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -298,18 +298,14 @@ struct reduce_over_group
         auto __local_idx = __item_id.get_local_id(0);
         auto __group_size = __item_id.get_local_range().size();
 
-        auto __k = 1;
-
-        do
+        for (::std::uint32_t __k = 1; __k < __group_size; __k <<= 1)
```
akukanov, line 301: Let's use `__k *= 2` here. Compilers know how to optimize it, and the code would be more self-descriptive. We can also use a more descriptive name, e.g. `__power_2`

julianmi: I will change this.

danhoeflinger (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -298,18 +298,15 @@ struct reduce_over_group
...
 
-        auto __k = 1;
-
-        do
+        for (::std::uint32_t __power_2 = 1; __power_2 < __group_size; __power_2 *= 2)
         {
             __dpl_sycl::__group_barrier(__item_id);
-            if (__local_idx % (2 * __k) == 0 && __local_idx + __k < __group_size && __global_idx < __n &&
-                __global_idx + __k < __n)
+            if ((__local_idx & (2 * __power_2 - 1)) == 0 && __local_idx + __power_2 < __group_size &&
```
SergeyKopienko, line 304: In this line you changing previous logic : is it ok ?

SergeyKopienko: OK

SergeyKopienko (approved):
LGTM

PR #874, opened 2023-04-04
Switch to USM Host Memory for Reduction Result

julianmi:
This patch aims at reducing the required transfer times in the family of reduce algorithms for copying the result back to the host.

The current mainline version uses a single value buffer that is wrapped in a future-like object. This patch introduces a future-like class for reduction results based on USM host memory. This eliminates the L0 call to `zeCommandListAppendMemoryCopy` and directly transfers the result instead. This improves the transfer times significantly with great runtime benefits for smaller input arrays.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -112,8 +112,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
 
         const bool __has_usm_host_allocations = has_usm_host_allocations(__exec.queue());
-
-        sycl::buffer<_Tp> __res_buf(sycl::range<1>(__has_usm_host_allocations ? 0 : 1));
+            __res_buf = ::std::make_unique<sycl::buffer<_Tp>>(sycl::range<1>(1));
         _Tp* __res_host_ptr = __has_usm_host_allocations ? sycl::malloc_host<_Tp>(1, __exec.queue()) : nullptr;
 
         sycl::event __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
@@ -124,12 +123,14 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
                 [=](sycl::nd_item<1> __item_id) {
                     __work_group_reduce_kernel<_Tp>(__item_id, __n, __n_items, __transform_pattern, __reduce_pattern,
                                                     __init, __temp_local, __res_acc, __rngs...);
-                    __res_acc[0] = __result;
+                    if (__has_usm_host_allocations)
+                        *__res_host_ptr = __result;
+                        (*__res_acc)[0] = __result;
                 });
         });
 
-        return __reduce_future<_ExecutionPolicy, sycl::event, _Tp>(::std::forward<_ExecutionPolicy>(__exec),
-                                                                   ::std::move(__reduce_event), __res);
+        return __reduce_future<_ExecutionPolicy, sycl::event, _Tp>(
+                                                                   ::std::move(__reduce_event), std::move(__res_buf),
     }
 }; // struct __parallel_transform_reduce_small_submitter
 
@@ -182,6 +183,7 @@ struct __parallel_transform_reduce_device_kernel_submitter<_Tp, __work_group_siz
         _Tp* __res = sycl::malloc_host<_Tp>(1, __exec.queue());
         return __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
+                __res_acc = new acc_type(__res_buf->template get_access<access_mode::write>(__cgh));
             __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__work_group_size), __cgh);
             __cgh.parallel_for<_KernelName...>(
                 sycl::nd_range<1>(sycl::range<1>(__n_groups * __work_group_size), sycl::range<1>(__work_group_size)),
@@ -341,7 +343,8 @@ struct __parallel_transform_reduce_impl
                 __cgh.depends_on(__reduce_event);
 
                 oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
-                auto __temp_acc = sycl::accessor(__temp, __cgh, sycl::read_write, __dpl_sycl::__no_init{});
+                using acc_type = decltype(__res_buf->template get_access<access_mode::write>(__cgh));
+                    __res_acc = new acc_type(__res_buf->template get_access<access_mode::write>(__cgh));
                 __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__work_group_size), __cgh);
 #if _ONEDPL_COMPILE_KERNEL && _ONEDPL_KERNEL_BUNDLE_PRESENT
                 __cgh.use_kernel_bundle(__kernel.get_kernel_bundle());
@@ -370,6 +373,7 @@ struct __parallel_transform_reduce_impl
                             {
                                 __reduce_pattern.apply_init(__init, __result);
                                 *__res = __result;
+                                    (*__res_acc)[0] = __result;
                             }
 
                             __temp_acc[__offset_1 + __group_idx] = __result;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -343,7 +343,7 @@ struct __parallel_transform_reduce_impl
                 __cgh.depends_on(__reduce_event);
 
                 // get an access to data under SYCL buffer
-                oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); 
+                oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
                 sycl::accessor __temp_acc{__temp, __cgh, sycl::read_write};
                 auto __res_acc = __res_container.get_acc(__cgh);
                 __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__work_group_size), __cgh);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -112,6 +112,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
 
         const bool __has_usm_host_allocations = has_usm_host_allocations(__exec.queue());
+        SyclBufferUniqPtr<_Tp> __res_buf;
             __res_buf = ::std::make_unique<sycl::buffer<_Tp>>(sycl::range<1>(1));
         _Tp* __res_host_ptr = __has_usm_host_allocations ? sycl::malloc_host<_Tp>(1, __exec.queue()) : nullptr;
 
@@ -181,6 +182,7 @@ struct __parallel_transform_reduce_device_kernel_submitter<_Tp, __work_group_siz
         const _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
         _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
 
+        SyclBufferUniqPtr<_Tp> __res_buf;
         sycl::buffer<_Tp> __res_buf(sycl::range<1>(1));
         _Tp* __res = sycl::malloc_host<_Tp>(1, __exec.queue());
         return __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
@@ -338,6 +340,7 @@ struct __parallel_transform_reduce_impl
         _Size __offset_1 = 0;
         _Size __offset_2 = __n_groups;
 
+        SyclBufferUniqPtr<_Tp> __res_buf;
         sycl::buffer<_Tp> __res_buf(sycl::range<1>(1));
         sycl::event __reduce_event;
         do
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -111,19 +111,20 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
                 __reduce_op, __transform_op};
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
 
-        const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
+        const bool __has_usm_host_allocations = has_usm_host_allocations(__exec.queue());
 
-        sycl::buffer<_Tp> __res(sycl::range<1>(1));
+        sycl::buffer<_Tp> __res_buf(sycl::range<1>(__has_usm_host_allocations ? 0 : 1));
+        _Tp* __res_host_ptr = __has_usm_host_allocations ? sycl::malloc_host<_Tp>(1, __exec.queue()) : nullptr;
 
         sycl::event __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
-            sycl::accessor __res_acc{__res, __cgh, sycl::write_only, __dpl_sycl::__no_init{}};
+            auto __res_acc = __res_buf.template get_access<access_mode::write>(__cgh);
             __cgh.parallel_for<_Name...>(
                 sycl::nd_range<1>(sycl::range<1>(__work_group_size), sycl::range<1>(__work_group_size)),
                 [=](sycl::nd_item<1> __item_id) {
                     __work_group_reduce_kernel<_Tp>(__item_id, __n, __n_items, __transform_pattern, __reduce_pattern,
                                                     __init, __temp_local, __res_acc, __rngs...);
-                __res[0] = __result;
+                    __res_acc[0] = __result;
                 });
         });
 
@@ -207,20 +208,19 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<_Tp, __work_group
     auto
     operator()(_ExecutionPolicy&& __exec, sycl::event& __reduce_event, _Size __n, _ReduceOp __reduce_op,
                _TransformOp __transform_op, _InitType __init, sycl::buffer<_Tp>& __temp) const
-    {
+                        if (__has_usm_host_allocations)
         using _NoOpFunctor = unseq_backend::walk_n<_ExecutionPolicy, oneapi::dpl::__internal::__no_op>;
-        auto __transform_pattern =
+                            *__res_host_ptr = __result;
             unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _NoOpFunctor>{
                 __reduce_op, _NoOpFunctor{}};
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
-
         // Lower the work group size of the second kernel to the next power of 2 if __n < __work_group_size.
         auto __work_group_size2 = __work_group_size;
         if constexpr (__iters_per_work_item == 1)
         {
             if (__n < __work_group_size)
             {
-                        __res[0] = __result;
+                            __res_acc[0] = __result;
                 if ((__work_group_size2 & (__work_group_size2 - 1)) != 0)
                     __work_group_size2 = oneapi::dpl::__internal::__dpl_bit_floor(__work_group_size2) << 1;
             }
@@ -322,7 +322,6 @@ struct __parallel_transform_reduce_impl
 
         // Create temporary global buffers to store temporary values
         sycl::buffer<_Tp> __temp(sycl::range<1>(2 * __n_groups));
-        sycl::buffer<_Tp> __res(sycl::range<1>(1));
         // __is_first == true. Reduce over each work_group
         // __is_first == false. Reduce between work groups
         bool __is_first = true;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -45,6 +45,11 @@ class __reduce_mid_work_group_kernel;
 template <typename... _Name>
 class __reduce_kernel;
 
+//------------------------------------------------------------------------
+// parallel_transform_reduce - async patterns
+// Please see the comment for __parallel_for_submitter for optional kernel name explanation
+//------------------------------------------------------------------------
+
 // Single work group kernel that transforms and reduces __n elements to the single result.
 template <typename _Tp, typename _NDItemId, typename _Size, typename _TransformPattern, typename _ReducePattern,
           typename _InitType, typename _AccLocal, typename _Res, typename... _Acc>
@@ -85,10 +90,8 @@ __device_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Size _
         __temp_acc[__group_idx] = __result;
 }
 
-//------------------------------------------------------------------------
-// parallel_transform_reduce - async patterns
-// Please see the comment for __parallel_for_submitter for optional kernel name explanation
-//------------------------------------------------------------------------
+template <typename T>
+using SyclBufferUniqPtr = ::std::unique_ptr<sycl::buffer<T>>;
 
 // Parallel_transform_reduce for a small arrays using a single work group.
 // Transforms and reduces __work_group_size * __iters_per_work_item elements.
@@ -113,10 +116,11 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
 
         const bool __has_usm_host_allocations = has_usm_host_allocations(__exec.queue());
         SyclBufferUniqPtr<_Tp> __res_buf;
+        if (!__has_usm_host_allocations)
             __res_buf = ::std::make_unique<sycl::buffer<_Tp>>(sycl::range<1>(1));
         _Tp* __res_host_ptr = __has_usm_host_allocations ? sycl::malloc_host<_Tp>(1, __exec.queue()) : nullptr;
 
-        sycl::event __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
+        return __exec.queue().submit([&, __n](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
             auto __res_acc = __res_buf.template get_access<access_mode::write>(__cgh);
             __cgh.parallel_for<_Name...>(
@@ -125,6 +129,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
                     __work_group_reduce_kernel<_Tp>(__item_id, __n, __n_items, __transform_pattern, __reduce_pattern,
                                                     __init, __temp_local, __res_acc, __rngs...);
                     if (__has_usm_host_allocations)
+                    if (__use_usm)
                         *__res_host_ptr = __result;
                         (*__res_acc)[0] = __result;
                 });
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +573,45 @@ class __future : private std::tuple<_Args...>
...
+        __my_event.wait_and_throw();
+#endif
+    }
+
+    auto
+    get()
+    {
+        using _Tp = typename ::std::remove_pointer<_Res>::type;
+        __my_event.wait();
+        _Tp __local_val = __my_res[0];
```
SergeyKopienko, line 609: Why we can't use ```auto``` here instead of ```_Tp``` type ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +573,45 @@ class __future : private std::tuple<_Args...>
...
+        sycl::free(__my_res, __my_exec.queue());
```
SergeyKopienko, line 610: If you call ```free``` function not in destructor, I think possible cases when the instance of this class will be destroyed without ```free``` call.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: I think we can. I will change it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: Good point, thanks. I will change that.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -604,10 +606,8 @@ class __reduce_future
     auto
     get()
     {
-        using _Tp = typename ::std::remove_pointer<_Res>::type;
         __my_event.wait();
-        _Tp __local_val = __my_res[0];
-        sycl::free(__my_res, __my_exec.queue());
+        auto __local_val = __my_res[0];
```
SergeyKopienko, line 610: or may be more simple: ```return __my_res[0];```

julianmi: This is what I have just realized and changed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +573,44 @@ class __future : private std::tuple<_Args...>
...
+    void
+    wait()
+    {
+#if !ONEDPL_ALLOW_DEFERRED_WAITING
+        __my_event.wait_and_throw();
+#endif
+    }
+
+    auto
+    get()
```
SergeyKopienko, line 607: May we declare this function as ```const``` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +573,44 @@ class __future : private std::tuple<_Args...>
...
+// A contract for a future class for reduce: <execution policy, sycl::event, USM host memory for the reduced value>
+template <typename _ExecutionPolicy, typename _Event, typename _Res>
+class __reduce_future
+{
+    _ExecutionPolicy __my_exec;
+    _Event __my_event;
+    _Res __my_res;
+
+  public:
+    __reduce_future(_ExecutionPolicy __exec, _Event __e, _Res __res)
```
SergeyKopienko, line 585: Usually you construct the instance of this class like
```__reduce_future(::std::forward<_ExecutionPolicy>(__exec), ...```
but this declaration mean that move semantic here really isn't possible, I think.

File: include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h [hunk 6]
```diff
@@ -573,6 +574,48 @@ class __future : private std::tuple<_Args...>
...
+template <typename _ExecutionPolicy, typename _Event, typename _Res>
+class __reduce_future
+{
+    _ExecutionPolicy __my_exec;
+    _Event __my_event;
+    ::std::shared_ptr<_Res> __my_res;
+
+  public:
+    __reduce_future(const _ExecutionPolicy& __exec, _Event __e, _Res __res)
+        : __my_exec(__exec), __my_event(__e), __my_res(::std::make_shared<_Res>(__res))
```
mmichel11, line 587: Since `_Event __e` is passed by value in the constructor, in the initializer list it may be better to have `__my_event(::std::move(__e))` to avoid an unnecessary copy.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +574,48 @@ class __future : private std::tuple<_Args...>
     }
 };
 
+// A contract for a future class for reduce: <execution policy, sycl::event, USM host memory for the reduced value>
+template <typename _ExecutionPolicy, typename _Event, typename _Res>
+class __reduce_future
+{
+    _ExecutionPolicy __my_exec;
+    _Event __my_event;
+    ::std::shared_ptr<_Res> __my_res;
```
mmichel11, line 583: Should this class be copy constructible or should we allow shared ownership? If we do not need it to be copy constructible, then I think the best option is to delete the copy constructor / assignment operator and make this `::std::unique_ptr<_Res>`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -606,8 +611,8 @@ class __reduce_future
     auto
     get()
     {
-        __my_event.wait();
-        return __my_res[0];
+        __my_event.wait_and_throw();
+        return *__my_res.get()[0];
```
SergeyKopienko, line 615: return (*__my_res)[0];

SergeyKopienko:
@julianmi please review and merge the PR https://github.com/oneapi-src/oneDPL/pull/875 into this branch

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: Agreed. I will change that.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: I tend to agree that our future-like objects should not be copied similar to `std::future`. I will change that if there are no other opinions on when copying a future could be useful.

julianmi:
> @julianmi please review and merge the PR #875 into this branch

Those are great improvements, thanks @SergeyKopienko.

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: Fixed: now we using ::std::unique_ptr<_Res> with custom deleter.

akukanov (dismissed):
I would like @MikeDvorskiy to also take a look before this patch is committed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -140,12 +139,13 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
                     if (__local_idx == 0)
                     {
                         __reduce_pattern.apply_init(__init, __result);
-                        __res_acc[0] = __result;
+                        __res[0] = __result;
```
SergeyKopienko, line 142: *__res = __result;

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -269,7 +268,7 @@ struct __parallel_transform_reduce_impl
                             if (__n_groups == 1)
                             {
                                 __reduce_op.apply_init(__init, __result);
-                                __res_acc[0] = __result;
+                                __res[0] = __result;
```
SergeyKopienko, line 271: *__res = __result;

SergeyKopienko (changes requested):

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +575,45 @@ class __future : private std::tuple<_Args...>
     }
 };
 
+// A contract for a future class for reduce: <execution policy, sycl::event, USM host memory for the reduced value>
+template <typename _ExecutionPolicy, typename _Event, typename _Res>
+class __reduce_future
```
MikeDvorskiy, line 580: I think we don't need the same addition feature type.
We may generalize existing one  - class __future, by  passing instead of buffer some RAII type based on ::std::unique_ptr. f.e.
```
template<typename Res>
using usm_value_buf = ::std::unique_ptr<_Res, ::std::function<void(_Res*)>>;
auto make_usm_value_buf(queue q, Res* res) { return usm_value_buf(res, [q](_Res* __res) { ::sycl::free(res, queue); }); }
```

MikeDvorskiy (changes requested):

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -120,11 +120,10 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
 
         const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
 
-        sycl::buffer<_Tp> __res(sycl::range<1>(1));
+        _Tp* __res = sycl::malloc_host<_Tp>(1, __exec.queue());
```
MikeDvorskiy, line 123: Regarding the host memory - what of we have async call and further async calls on the device and we don't need access to reduction result on the host? We should bot get performance degradation in that case.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -65,19 +65,19 @@ struct __parallel_transform_reduce_seq_submitter<_Tp, __internal::__optional_ker
...
 
         sycl::event __reduce_event = __exec.queue().submit([&, __n](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
-            auto __res_acc = __res.template get_access<access_mode::write>(__cgh);
             __cgh.single_task<_Name...>([=] {
                 _Tp __result = __transform_pattern(__n, __rngs...);
                 __reduce_pattern.apply_init(__init, __result);
-                __res_acc[0] = __result;
+                __res[0] = __result;
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: How would you access the result on the device then? Calling `get()` on the returned future would return a host accessor to the data.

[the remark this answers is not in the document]
julianmi: Good point. I think this would require a separate buffer and USM kernel which would add quite a bit of code. Is that worth it for the discussed performance improvements?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
rarutyun: I don't suggest to have type-erased utility (like `std::function`). Let's avoid allocation we don't control. You can always write deleter as a named class (not as a lambda) so I don't see why this overhead is necessary.

In general I am in favor of genericity idea. We always use either buffer of pointer, right? Can we just have compile-time check and rely that pointer always was allocated via USM?

SergeyKopienko: Usage of ```std::function``` were removed from this PR.

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +652,20 @@ class __future : private std::tuple<_Args...>
...
+inline bool
+__use_USM_host_allocations(sycl::queue __queue)
+{
+    auto __device = __queue.get_device();
+    if (!__device.is_gpu())
+        return false;
+    if (!__device.has(sycl::aspect::usm_host_allocations))
+        return false;
+    if (__device.get_info<sycl::info::device::vendor_id>() != 32902)
+        return false;
```
akukanov, line 665: Filtering out some code paths based on vendor id instead of HW capabilities might cause bad perception. What alternative checks might be used here?

julianmi: I think the key characteristic is the backend. L0 provides a low-overhead implementation of USM host memory whereas other backends use slower pinned host memory (mlock on Linux).

SergeyKopienko:
There are 3 places in code like
```__storage __res_container = __storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);```

but in all these places you pass __res_container  into your Kernel code by reference: ```[&]``` in lambda.
For example, will take a look into
```struct __parallel_transform_reduce_small_submitter```
passing instance of ```__res_container``` by reference, from my point of view this mean that that there are no shared_ptr pointers incrementation doing.

And if I am right, then you return result
```return __future(__reduce_event, __res_container);```
and on caller side we will forget about in, this mean that you container(s) will be delated but still used inside the Kernel code.

It's my opinion.
May be I am wrong but better to check.

File: include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h [hunk 15]
```diff
@@ -508,8 +509,76 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
...
 
-//A contract for future class: <sycl::event or other event, a value or sycl::buffers...>
-//Impl details: inheretance (private) instead of aggregation for enabling the empty base optimization.
+template <typename _T>
+struct __accessor
+{
+    using __accessor_t = sycl::accessor<_T, 1, sycl::access::mode::read_write, __dpl_sycl::__target_device,
+                                        sycl::access::placeholder::false_t>;
+    __accessor_t m_acc;
+    _T* m_ptr;
```
SergeyKopienko, line 518: Better to initialize this field on declaration: ```_T* m_ptr = null_ptr;```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -508,8 +509,76 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
...
+};
+
+template <typename _ExecutionPolicy, typename _T>
+struct __storage
+{
+  private:
+    using __sycl_buffer_t = sycl::buffer<_T, 1>;
+    ::std::shared_ptr<__sycl_buffer_t> m_sycl_buf;
+    ::std::shared_ptr<_T> m_usm_buf;
+    bool m_usm;
```
SergeyKopienko, line 535: Better to initialize this field on declaration: ```bool m_usm = false;```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -508,8 +509,76 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
...
+            m_sycl_buf = ::std::make_shared<__sycl_buffer_t>(__sycl_buffer_t(__n));
+    }
+
+    auto
+    get_acc(sycl::handler& __cgh)
+    {
+        __accessor<_T> acc;
+        if (m_usm)
+        {
+            acc.m_usm = true;
```
SergeyKopienko, line 556: We can move this duplicated line outside of condition: ```acc.m_usm = m_usm;```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -508,8 +509,76 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
...
+    }
+
+    auto
+    get_value(size_t idx = 0)
+    {
+        return m_usm ? *(m_usm_buf.get() + idx) : m_sycl_buf->get_host_access(sycl::read_only)[idx];
+    }
+
+    bool
+    get_usm()
```
SergeyKopienko, line 574: I believe we can declare this method as ```const```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -508,8 +509,76 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
...
+        return m_usm ? m_ptr : &m_acc[0];
+    }
+};
+
+template <typename _ExecutionPolicy, typename _T>
+struct __storage
+{
+  private:
+    using __sycl_buffer_t = sycl::buffer<_T, 1>;
+    ::std::shared_ptr<__sycl_buffer_t> m_sycl_buf;
```
mmichel11, line 533: Since we are exclusively using `m_sycl_buf` or `m_usm_buf`, it may be worth it to move these two variables into a union or `std::variant` so we only ever store one pointer in this class.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -508,8 +509,76 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
...
-//A contract for future class: <sycl::event or other event, a value or sycl::buffers...>
-//Impl details: inheretance (private) instead of aggregation for enabling the empty base optimization.
+template <typename _T>
+struct __accessor
+{
+    using __accessor_t = sycl::accessor<_T, 1, sycl::access::mode::read_write, __dpl_sycl::__target_device,
+                                        sycl::access::placeholder::false_t>;
+    __accessor_t m_acc;
+    _T* m_ptr;
+    bool m_usm;
```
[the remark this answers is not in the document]
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: This was tried before but rendered tricky when passing them to a kernel. Thus, the current solution uses smart pointers which should incur only minor overheads. The runtime performance is very close to the explicit mode proposed in #1106.

julianmi:
I've rebased this PR and introduced checks to support multiple compiler versions. Currently, DPC++ 2023.1 and 20230822 and above support the USM host approach proposed here. Other compilers fall back to the existing buffer implementation. I've verified this approach with oneAPI releases back to 2021.4. The patch shows similar performance to the mainline when using buffers and significant speedups when using USM host memory on L0 GPUs, especially for small arrays.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: Ah, I see. This makes sense, then.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: I mean
```bool
get_usm() const
{
    return m_usm;
}
```

julianmi: Of course, my bad. I fixed it.

julianmi:
> There are 3 places in code like
> 
> ```c++
> __storage __res_container = __storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);
> ```
> 
> but in all these places you pass __res_container into your Kernel code by reference: `[&]` in lambda. For example, will take a look into
> 
> ```c++
> struct __parallel_transform_reduce_small_submitter
> ```
> 
> passing instance of `__res_container` by reference, from my point of view this mean that that there are no shared_ptr pointers incrementation doing.
> 
> And if I am right, then you return result
> 
> ```c++
> return __future(__reduce_event, __res_container);
> ```
> 
> and on caller side we will forget about in, this mean that you container(s) will be delated but still used inside the Kernel code.
> 
> It's my opinion. May be I am wrong but better to check.

Thanks for raising this, @SergeyKopienko. I think this issue could occur if someone calls `reduce_async` without getting the result. I don't think this is a realistic case though. And even if someone would do this, the runtime manages the lifetime of buffers and the USM version is only used with L0, which pools memory operations. I couldn't reproduce this issue with a long-running kernel and USM or buffer memory.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -523,6 +602,16 @@ class __future : private std::tuple<_Args...>
...
     }
 
+    template <typename _ExecutionPolicy, typename _T>
+    constexpr auto
+    __wait_and_get_value(__storage<_ExecutionPolicy, _T>& __buf)
+    {
+        // Explicit wait in case of USM memory. Buffer accessors are synchronous.
+        if (__buf.get_usm())
+            wait();
+        return __buf.get_value();
```
MikeDvorskiy, line 612: SYCL buffer guarantees just data synchronization... But a kernel might still work...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +662,26 @@ class __future : private std::tuple<_Args...>
     }
 };
 
+// Only use USM host allocations on L0 GPUs. Other devices show significant slowdowns and will use a buffer instead.
+inline bool
+__use_USM_host_allocations(sycl::queue __queue)
+{
+// The unified future above is currently only supported by DPCPP 2023.1 and compilers starting from 20230820.
```
MikeDvorskiy, line 669: It looks very weird...
Why would not just call `__device.has(sycl::aspect::usm_host_allocations`? or why such call doesn't  make the all necessary checks?
A question rather is to a runtime team...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: I think the data dependence should be on the buffer that is used in the kernel (read/write access). Thus, the host accessor called here should wait for the kernel completion. This follows the existing `__wait_and_get_value(sycl::buffer<_T>& __buf)` implementation above.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: We can query the support for USM host memory with `sycl::aspect::usm_host_allocations`. However, its use has shown to be beneficial on L0 GPUs only. This routine checks for all three conditions and uses guards to fall back to USM for compilers not supporting the unified future implementation.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
MikeDvorskiy: "Thus, the host accessor called here should wait for the kernel completion"

I don't think so. The host accessor is waiting for the data will be copied back to the host, nothing more.

MikeDvorskiy: Actually, looking at the spec:
```
As said in [Section 3.9.8](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:synchronization), the only blocking operations in SYCL (apart from explicit wait operations) are:

host accessor constructor, which waits for any kernels enqueued before its creation that write to the corresponding object to finish and be copied back to host memory before it starts processing. The host accessor does not necessarily copy back to the same host memory as initially given by the user;
```

So,

MikeDvorskiy (dismissed):
LGTM

mmichel11 (dismissed):
LGTM. It's probably best to check with @SergeyKopienko as well if he's ready to approve.

SergeyKopienko (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -508,8 +509,86 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
...
+    {
+        if (m_usm)
+            m_ptr = m_usm_buf.get();
+        else
+            m_acc = sycl::accessor(*m_sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{});
+    }
+#else
+    __accessor(sycl::handler& __cgh, bool __usm, ::std::shared_ptr<sycl::buffer<_T, 1>> m_sycl_buf,
+               ::std::shared_ptr<_T> m_usm_buf)
+        : m_usm(__usm), m_acc(sycl::accessor(*m_sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{}))
```
akukanov, line 537: Should not `m_usm` be set to `false` there, as de facto the buffer is used?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -508,8 +509,86 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
...
+{
+  private:
+    using __accessor_t = sycl::accessor<_T, 1, sycl::access::mode::read_write, __dpl_sycl::__target_device,
+                                        sycl::access::placeholder::false_t>;
+    __accessor_t m_acc;
+    _T* m_ptr = nullptr;
+    bool m_usm = false;
+
+  public:
+// Compilers before 2023.1 don't provide a default constructor for sycl::accessor. Fallback to a buffer instead.
```
akukanov, line 523: The comment talks about a fallback, yet precedes the "main" branch. It should better be moved or rephrased.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -508,8 +509,86 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
...
+#if __SYCL_COMPILER_VERSION == 20230320 || __SYCL_COMPILER_VERSION >= 20230820
```
akukanov, line 524: Does it make sense to introduce a "feature presence" macro in the config header?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: I've reworded the comment.

akukanov (dismissed):

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -78,6 +78,14 @@
 #    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE_IF_SUPPORTED(SIZE)
 #endif
 
+// The unified future supporting USM host memory and buffers is only supported by DPCPP 2023.1
+// and compilers starting from 20230820.
+#if (__SYCL_COMPILER_VERSION == 20230320 || __SYCL_COMPILER_VERSION >= 20230820)
+#    define _ONEDPL_SYCL_USM_HOST_PRESENT 1
```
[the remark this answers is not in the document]
julianmi: I've switched to using `_ONEDPL_LIBSYCL_VERSION` to exclude the problematic compiler 2023.1 which has `_ONEDPL_LIBSYCL_VERSION == 60200`.

akukanov (approved):

SergeyKopienko (approved):
LGTM

PR #876, opened 2023-04-05
Add the feature macro for dynamic selection

akukanov:
For this branch, I prefer a "minimally viable" implementation that does not introduce new files. Later we should add a special header to the main branch.


include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -27,6 +27,9 @@
 #define ONEDPL_VERSION_MINOR 1
 #define ONEDPL_VERSION_PATCH 1
 
+// Feature macros
+#define ONEDPL_HAS_DYNAMIC_SELECTION 202304L
```
rarutyun, line 31: I thought the idea was to introduce `<version>` header.

In this case user should include the feature header he/she is going to use to get the correct value, which sort of reminds me egg and chicken problem. I am specifically saying "sort of" because yeah, users also have `__has_include` but logically it looks similar.

akukanov: As I mentioned in the description, the version header will be introduced into the main branch first.

There is no chicken and egg problem, because virtually any oneDPL header adds the config header. We should document it better, but anyway the new macro is as accessible as the documented version macros, no more, no less.

And by the way, adding the `<version>` header does not eliminate the need for `__has_include`, because that header did not exist in all the previous library versions.

timmiesmith (approved):
Looks good to me.

vossmjp (approved):
Looks ok to me.

PR #885, opened 2023-04-13
Import tests from dpstd library - part 1/16

SergeyKopienko:
By this PR we open series of PR's in which we import dpstd  library tests into `test/xpu_api` tests.

All ```.cpp``` test files were imported into our repository with minimal changes from source.
To see the difference for all test file you can compare it's state between the first and the last it's commits in the file history.

List of all linked PR's :
- part  2 / 16 - https://github.com/oneapi-src/oneDPL/pull/887
- part  3 / 16 - https://github.com/oneapi-src/oneDPL/pull/888
- part  4 / 16 - https://github.com/oneapi-src/oneDPL/pull/889
- part  5 / 16 - https://github.com/oneapi-src/oneDPL/pull/890
- part  6 / 16 - https://github.com/oneapi-src/oneDPL/pull/891
- part  7 / 16 - https://github.com/oneapi-src/oneDPL/pull/892
- part  8 / 16 - https://github.com/oneapi-src/oneDPL/pull/893
- part  9 / 16 - https://github.com/oneapi-src/oneDPL/pull/894
- part 10 / 16 - https://github.com/oneapi-src/oneDPL/pull/895
- part 11 / 16 - https://github.com/oneapi-src/oneDPL/pull/896
- part 12 / 16 - https://github.com/oneapi-src/oneDPL/pull/897
- part 13 / 16 - https://github.com/oneapi-src/oneDPL/pull/898
- part 14 / 16 - https://github.com/oneapi-src/oneDPL/pull/899
- part 15 / 16 - https://github.com/oneapi-src/oneDPL/pull/900
- part 16 / 16 - https://github.com/oneapi-src/oneDPL/pull/901


test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
```diff
@@ -0,0 +1,103 @@
...
+}
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+int
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    kernel_test<forward_iterator<const int*>, KernelBSearchTest1>();
+    kernel_test<bidirectional_iterator<const int*>, KernelBSearchTest2>();
+    // Enable the test when GPU runtime fix issue
```
MikeDvorskiy, line 97: Usually we use a kind of "broken" macro to temporary exclude a "broken" test case.
A "broken" macro is defined depending on OS/Compiler version/Library version etc

test/CMakeLists.txt
```diff
@@ -145,24 +169,37 @@ endmacro()
...
-            onedpl_add_test(${_file} false)
+            onedpl_add_test(${_file} false false)
         endif()
     endif()
 
 endforeach()
+
+file(GLOB_RECURSE UNIT_TESTS "std/*.pass.cpp")
+foreach (_file IN LISTS UNIT_TESTS)
+
```
adamfidel, line 201: Extra newlines at the beginning and end of this `foreach` block.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
```diff
@@ -0,0 +1,103 @@
...
+    sycl::range<1> host_buffer_sz{N};
+    sycl::buffer<sycl::cl_int, 1> host_data_buffer(host_vbuf, host_buffer_sz);
+    deviceQueue.submit([&](sycl::handler& cgh) {
+        auto ret_access = buffer1.get_access<sycl_write>(cgh);
+        auto data_access = host_data_buffer.get_access<sycl_read>(cgh);
+        cgh.single_task<KC>([=]() {
+            int device_vbuf[N];
+            for (size_t i = 0; i < N; i++)
+                device_vbuf[i] = data_access[i];
+            ret_access[0] = test(Iter(device_vbuf), Iter(device_vbuf + N), 0, true);
```
MikeDvorskiy, line 68: Why do we need a type conversion here?
`Iter(...)`

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
```diff
@@ -0,0 +1,103 @@
...
+    sycl::cl_bool ret = false;
+    sycl::range<1> numOfItems{1};
+    sycl::buffer<sycl::cl_bool, 1> buffer1(&ret, numOfItems);
+    const unsigned N = 1000;
+    const int M = 10;
+    int host_vbuf[N];
+    int x = 0;
+    for (size_t i = 0; i < N; ++i)
+    {
+        host_vbuf[i] = x;
```
adamfidel, line 53: This loop looks like it's setting values from 0 to `(M-1)` by resetting the `x` variable once it gets to M. Could you instead change this line to 
```suggestion
        host_vbuf[i] = i % M;
```

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
```diff
@@ -0,0 +1,103 @@
...
+    }
+
+    std::sort(host_vbuf, host_vbuf + N);
+    sycl::range<1> host_buffer_sz{N};
+    sycl::buffer<sycl::cl_int, 1> host_data_buffer(host_vbuf, host_buffer_sz);
+    deviceQueue.submit([&](sycl::handler& cgh) {
+        auto ret_access = buffer1.get_access<sycl_write>(cgh);
+        auto data_access = host_data_buffer.get_access<sycl_read>(cgh);
+        cgh.single_task<KC>([=]() {
+            int device_vbuf[N];
```
MikeDvorskiy, line 65: Why do we need to create a private array (in the device registers)? I think we may call test with the device data, which provided by `data_access`.

test/support/utils.h
```diff
@@ -636,6 +636,17 @@ transform_reduce_serial(InputIterator first, InputIterator last, T init, BinaryO
     return init;
 }
 
+template <typename TState>
+void
+exitOnError(TState state, const char* sFailMsg = "Fail")
```
adamfidel, line 641: Should this name be in snake case (i.e., `exit_on_error`) to match the rest of the functions in this file?

test/support_stdlib/oneapi_std_test_config.h
```diff
@@ -0,0 +1,33 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _ONEAPI_STD_TEST_config_H
+#define _ONEAPI_STD_TEST_config_H
+
+#include "../support/test_config.h"
+#include "../support/utils.h"
+
+#define _ONEAPI_STD_TEST_STRING(X) _ONEAPI_STD_TEST_STRING_AUX(oneapi/dpl/X)
```
adamfidel, line 22: Should this macro be moved after the definition of `_ONEAPI_STD_TEST_STRING_AUX` on the following line?

test/support_stdlib/oneapi_std_test_config.h
```diff
@@ -0,0 +1,33 @@
...
+#define _ONEAPI_STD_TEST_STRING_AUX(X) #X
+//to support the optional including: <algorithm>, <iterator>, <numeric>, <array> 
+#define _ONEAPI_STD_TEST_HEADER(HEADER_ID) _ONEAPI_STD_TEST_STRING(HEADER_ID)
```
adamfidel, line 25: Could you expand on the benefit of using this macro? From what I can tell, it is used to do something like
```c++
#include _ONEAPI_STD_TEST_HEADER(algorithm)
```
instead of
```c++
```
Is there something else?

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
```diff
@@ -0,0 +1,103 @@
...
+{
+    sycl::queue deviceQueue = TestUtils::get_test_queue();
+    sycl::cl_bool ret = false;
+    sycl::range<1> numOfItems{1};
+    sycl::buffer<sycl::cl_bool, 1> buffer1(&ret, numOfItems);
+    const unsigned N = 1000;
+    const int M = 10;
+    int host_vbuf[N];
+    int x = 0;
+    for (size_t i = 0; i < N; ++i)
```
MikeDvorskiy, line 51: Probably, it makes sense to make a sequence generator more advanced? 
F.e.
```
std::default_random_engine gen{std::random_device{}()};
std::uniform_real_distribution<float> dist(0, 100.0);
std::generate(in_host.begin(), in_host.end(), [&]{ return dist(gen); });
```

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
```diff
@@ -0,0 +1,103 @@
...
+template <typename Iter, typename KC>
+void
+kernel_test()
+{
+    sycl::queue deviceQueue = TestUtils::get_test_queue();
+    sycl::cl_bool ret = false;
+    sycl::range<1> numOfItems{1};
+    sycl::buffer<sycl::cl_bool, 1> buffer1(&ret, numOfItems);
+    const unsigned N = 1000;
+    const int M = 10;
```
MikeDvorskiy, line 48: Why `M` is signed integer type while `N` - unsigned integer type?

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+            /* int d[] = {0, 2, 4, 6};
+	    for (int *e = d; e <= d+4; ++e)
+	        for (int y = -1; y <= 7; ++y)
+	           ret_access[0] &= test(d, e, y, (y % 2 ==0) && ((e - d)*2 > y));
+	   */
+        });
+    });
+
+    auto ret_access_host = buffer1.get_access<sycl_read>();
```
MikeDvorskiy, line 86: Whether do we need `queue::wait` before?

MikeDvorskiy (commented):
To continue review it would be greater to answer(and/or make changes) the first portion of questions.

test/CMakeLists.txt
SergeyKopienko: Fixed.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
```diff
@@ -0,0 +1,103 @@
...
+            x = 0;
+    }
+
+    std::sort(host_vbuf, host_vbuf + N);
+    sycl::range<1> host_buffer_sz{N};
+    sycl::buffer<sycl::cl_int, 1> host_data_buffer(host_vbuf, host_buffer_sz);
+    deviceQueue.submit([&](sycl::handler& cgh) {
+        auto ret_access = buffer1.get_access<sycl_write>(cgh);
+        auto data_access = host_data_buffer.get_access<sycl_read>(cgh);
+        cgh.single_task<KC>([=]() {
```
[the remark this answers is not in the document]
SergeyKopienko: This is also from source test state.
I think out goal is simple to import tests into our repository.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
SergeyKopienko: the same.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
SergeyKopienko: the same.

test/support/utils.h
SergeyKopienko: Fixed: renamed to ``exit_on_error```

test/support_stdlib/oneapi_std_test_config.h
SergeyKopienko: Fixed,

test/support_stdlib/oneapi_std_test_config.h
SergeyKopienko: Fixed: this macro and it's usage was removed.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
SergeyKopienko: Yes, sure. Fixed.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
SergeyKopienko: @MikeDvorskiy it's from source code of importing test. We have the goal to import all these tests into our code but no goal to improve these tests and so on.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
akukanov: Let's check if it works now, and then decide. Random access iterator is the one people will use most, I believe.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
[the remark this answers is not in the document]
akukanov: Mikhail, what complication do you see?
We test binary_search, yes - but over the data passed into the kernel as a buffer. That makes sense to me.
Or are you concerned about ret_access etc.? It is just the way to combine results of multiple checks with boolean AND, and return the value to the host.

Anyway, as Sergey mentioned, the primary goal for now is to integrate these tests into our repo. It's OK to point out some obvious and simple improvements, but making these tests "perfect" is not a goal.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
akukanov: To test the algorithm with different iterator types.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
akukanov: On the one hand, it's more practical that a search is done over the data in global / local memory than in private arrays. On the other hand, I do not know if an iterator obtained by accessor.begin() is convertible to different types.

Let's add an issue/task to address this question later.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
akukanov: I would go even further, and changed this loop to generate sorted data, instead of applying std::sort later.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
akukanov: If the test is only intended to check that a standard algorithm can be used on the device, any data are suitable I believe, and we can just generate a sorted sequence of integers as I suggested in another comment. This approach assumes that the std library implementation is well tested on its own, and we do not change it.

A different approach is to assume less - for example, expect that the implementation has different code paths, and not all might work on the device. In this case, we'd need comprehensive testing, randomization, checking special cases, etc. 

I tend to prefer the first approach, but we can discuss.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
SergeyKopienko: Checked: it's working OK. So code has been uncommented.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
SergeyKopienko: Fixed: conversion removed.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
SergeyKopienko: Fixed.

SergeyKopienko:
> To continue review it would be greater to answer(and/or make changes) the first portion of questions.
@MikeDvorskiy comments were fixed, please continue.

MikeDvorskiy (commented):
A test design might be better... But, as far as I understand it is because 
we take these tests AS IS from "compiler std test".
And "refactoring because refactoring" doesn't make sense, I guess.
So, I have to be agreed with minor comments in the some places.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include "oneapi_std_test_config.h"
+
+#include _ONEAPI_STD_TEST_HEADER(algorithm)
+#include _ONEAPI_STD_TEST_HEADER(iterator)
+
+#include <iostream>
+
+#include "test_iterators.h"
+#include "checkData.h"
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search_comp.pass.cpp
```diff
@@ -0,0 +1,100 @@
...
+#include "oneapi_std_test_config.h"
+
+#include _ONEAPI_STD_TEST_HEADER(algorithm)
+#include _ONEAPI_STD_TEST_HEADER(iterator)
+#include _ONEAPI_STD_TEST_HEADER(functional)
+
+#include <iostream>
+
+#include "test_iterators.h"
+#include "checkData.h"
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search_comp.pass.cpp
```diff
@@ -0,0 +1,100 @@
...
+            for (int x = 1; x < M; ++x)
+                ret_access[0] &= test(device_vbuf, device_vbuf + N, x, true);
+
+            ret_access[0] &= test(device_vbuf, device_vbuf + N, -1, false);
+            ret_access[0] &= test(device_vbuf, device_vbuf + N, M, false);
+        });
+    }).wait();
+
+    auto ret_access_host = buffer1.get_host_access(sycl::read_only);
+    EXPECT_TRUE(ret_access_host[0], "");
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/CMakeLists.txt
```diff
@@ -57,15 +57,30 @@ add_custom_target(run-onedpl-tests
...
+        list(GET _test_name_list -2 parent_folder_name)
+        list(GET _test_name_list -1 test_name)
+
+        set(result_test_name "std-${parent_folder_name}-${test_name}")
+        string(REPLACE "-" "-" _test_name ${result_test_name})
+    else()
+        #process oneDPL tests: target name equal to cpp file name without extension
+        get_filename_component(_test_name ${test_source_file} NAME)
+        string(REPLACE "\.cpp" "" _test_name ${_test_name})
+    endif()
```
dmitriy-sobolev, line 83: Does this code construct unique target names from files with the same names using their path? If, so Is it possible to name these files uniquely to simplify the logic?

SergeyKopienko: Yes, exactly, But I preferer to generate unique name to avoid test names like test1.pass.cpp, ..., testN.pass.cpp and so on.
Additionally this test are copied from another repository and better to save their names I believe.

test/support_stdlib/testsuite_common_types.h
```diff
@@ -0,0 +1,98 @@
+// -*- C++ -*-
+// typelist for the C++ library testsuite.
+//
+// Copyright (C) 2005-2019 Free Software Foundation, Inc.
+//
+// This file is part of the GNU ISO C++ Library.  This library is free
+// software; you can redistribute it and/or modify it under the
+// terms of the GNU General Public License as published by the
+// Free Software Foundation; either version 3, or (at your option)
+// any later version.
```
dmitriy-sobolev, line 10: Is this the right license? I guess it should be `Apache License v2.0 with LLVM Exceptions` as in the other tests. 

Also, there are some files without license e.g. `test/support_stdlib/testsuite_iterators.h`. Could you specify a license there?

SergeyKopienko:
@MikeDvorskiy could you please continue your review? As far as I see all your comments fixed or answered.

test/support_stdlib/testsuite_common_types.h
MikeDvorskiy: @SergeyKopienko, what's your answers here?

SergeyKopienko: My answer: it's license from the source test file so I prefer to keep it.

SergeyKopienko: File deleted as unused.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search1_g.pass.cpp
```diff
@@ -0,0 +1,128 @@
...
+}
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+int
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    auto ret = kernel_test1();
+    ret &= kernel_test2();
+    EXPECT_TRUE(ret, "");
```
MikeDvorskiy, line 124: "Wrong result of binary_search" ?

test/std/algorithms/alg.sorting/alg.binary.search/equal.range/equal_range1_g.pass.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+int
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    auto ret = kernel_test1();
+    ret &= kernel_test2();
+    EXPECT_TRUE(ret, "Wrong result of equal_range");
```
MikeDvorskiy, line 129: What's difference "equal_range1" from "equal_range"?
I think this difference should be reflected in an error message as well.

test/std/algorithms/alg.sorting/alg.binary.search/equal.range/equal_range2_g.pass.cpp
```diff
@@ -0,0 +1,136 @@
...
+    return ret;
+}
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+int
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    auto ret = kernel_test();
+    EXPECT_TRUE(ret, "Wrong result of equal_range");
```
MikeDvorskiy, line 132: Please see the previous comment about wording of a error message.

test/std/algorithms/alg.sorting/alg.binary.search/equal.range/equal_range_comp.pass.cpp
```diff
@@ -0,0 +1,103 @@
...
+        auto ret_access = buffer1.get_access<sycl_write>(cgh);
+        cgh.single_task<KC>([=]() {
+            ret_access[0] = test(device_vbuf, device_vbuf + N, 0);
+            for (int x = 1; x <= M; ++x)
+                ret_access[0] &= test(device_vbuf, device_vbuf + N, x);
+        });
+    }).wait();
+
+    auto ret_access_host = buffer1.get_host_access(sycl::read_only);
+    EXPECT_TRUE(ret_access_host[0], "Wrong result of equal_range");
```
MikeDvorskiy, line 88: "Wrong result of equal_range with comparator" ?

test/std/algorithms/alg.sorting/alg.binary.search/lower.bound/lower_bound1_g.pass.cpp
```diff
@@ -0,0 +1,92 @@
...
+    return ret;
+}
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+int
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    auto ret = kernel_test();
+    EXPECT_TRUE(ret, "Wrong result of lower_bound");
```
MikeDvorskiy, line 88: what's difference "lower_bound_1" from "lower_bound"?
That difference should be reflected in this error message.

test/std/algorithms/alg.sorting/alg.binary.search/lower.bound/lower_bound_comp.pass.cpp
```diff
@@ -0,0 +1,108 @@
...
+        auto ret_access = buffer1.get_access<sycl_write>(cgh);
+        cgh.single_task<KC>([=]() {
+            ret_access[0] = test(device_vbuf, device_vbuf + N, 0);
+            for (int x = 1; x <= M; ++x)
+                ret_access[0] &= test(device_vbuf, device_vbuf + N, x);
+        });
+    }).wait();
+
+    auto ret_access_host = buffer1.get_host_access(sycl::read_only);
+    EXPECT_TRUE(ret_access_host[0], "Wrong result of lower_bound");
```
MikeDvorskiy, line 93: "Wrong result of lower_bound with comparator" ?

test/std/algorithms/alg.sorting/alg.binary.search/lower.bound/partitioned_g.pass.cpp
```diff
@@ -0,0 +1,187 @@
...
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    sycl::queue deviceQueue = TestUtils::get_test_queue();
+    auto ret = kernel_test1(deviceQueue);
+    if (TestUtils::has_type_support<double>(deviceQueue.get_device()))
+    {
+        ret &= kernel_test2(deviceQueue);
+    }
+    EXPECT_TRUE(ret, "Wrong result of lower_bound");
```
MikeDvorskiy, line 183: What's a difference form others "lower_bound" tests?
I think a diff should be mentioned in this error message as well (to quick get info and find place).

test/std/algorithms/alg.sorting/alg.binary.search/upper.bound/partitioned_g.pass.cpp
```diff
@@ -0,0 +1,186 @@
...
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    sycl::queue deviceQueue = TestUtils::get_test_queue();
+    auto ret = kernel_test1(deviceQueue);
+    if (TestUtils::has_type_support<double>(deviceQueue.get_device()))
+    {
+        ret &= kernel_test2(deviceQueue);
+    }
+    EXPECT_TRUE(ret, "Wrong result of upper_bound");
```
MikeDvorskiy, line 182: What's a key difference form others "upper_bound" tests?
I think a diff should be mentioned in this error message as well.

test/std/algorithms/alg.sorting/alg.binary.search/upper.bound/upper_bound.pass.cpp
```diff
@@ -0,0 +1,109 @@
...
+        cgh.single_task<KC>([=]() {
+            ret_access[0] = test(device_vbuf, device_vbuf + N, 0);
+
+            for (int x = 1; x <= M; ++x)
+                ret_access[0] &= test(device_vbuf, device_vbuf + N, x);
+        });
+    }).wait();
+
+    auto ret_access_host = buffer1.get_host_access(sycl::read_only);
+    EXPECT_TRUE(ret_access_host[0], "Wrong result of upper_bound");
```
MikeDvorskiy, line 94: What's a key difference form others "upper_bound" tests?
I think a diff should be mentioned in this error message as well.

test/std/algorithms/alg.sorting/alg.binary.search/upper.bound/upper_bound_comp.pass.cpp
```diff
@@ -0,0 +1,107 @@
...
+        auto ret_access = buffer1.get_access<sycl_write>(cgh);
+        cgh.single_task<KC>([=]() {
+            ret_access[0] = test(device_vbuf, device_vbuf + N, 0);
+            for (int x = 1; x <= M; ++x)
+                ret_access[0] &= test(device_vbuf, device_vbuf + N, x);
+        });
+    }).wait();
+
+    auto ret_access_host = buffer1.get_host_access(sycl::read_only);
+    EXPECT_TRUE(ret_access_host[0], "Wrong result of upper_bound");
```
MikeDvorskiy, line 92: What's a key difference form others "upper_bound" tests?
I think a diff should be mentioned in this error message as well.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search_comp.pass.cpp
```diff
@@ -0,0 +1,99 @@
...
+            for (int x = 1; x < M; ++x)
+                ret_access[0] &= test(device_vbuf, device_vbuf + N, x, true);
+
+            ret_access[0] &= test(device_vbuf, device_vbuf + N, -1, false);
+            ret_access[0] &= test(device_vbuf, device_vbuf + N, M, false);
+        });
+    }).wait();
+
+    auto ret_access_host = buffer1.get_host_access(sycl::read_only);
+    EXPECT_TRUE(ret_access_host[0], "Wrong result of binary_search");
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search1_g.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/lower.bound/lower_bound_comp.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search2_g.pass.cpp
```diff
@@ -0,0 +1,109 @@
...
+    return ret;
+}
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+int
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    auto ret = kernel_test();
+    EXPECT_TRUE(ret, "");
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

SergeyKopienko:
@ValentinaKats could you please clarify, should we really have unique error text in each test? Or may we have the same error text in different test files and it's not the problem for our nightly tests and so on?

test/std/algorithms/alg.sorting/alg.binary.search/upper.bound/upper_bound_comp.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/equal.range/equal_range1_g.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/equal.range/equal_range2_g.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/upper.bound/upper_bound1_g.pass.cpp
```diff
@@ -0,0 +1,92 @@
...
+    return ret;
+}
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+int
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    auto ret = kernel_test();
+    EXPECT_TRUE(ret, "Wrong result of upper_bound");
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/upper.bound/partitioned_g.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/upper.bound/upper_bound.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/lower.bound/partitioned_g.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/lower.bound/lower_bound1_g.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/equal.range/equal_range_comp.pass.cpp
SergeyKopienko: Fixed

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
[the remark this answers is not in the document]
MikeDvorskiy: "the primary goal for now is to integrate these tests into our repo. "is to integrate these tests into our repo."
(from https://github.com/intel-innersource/applications.compilers.source.dpcpplib)

I would ask "why do we need that?" Why do we need such duplication of huge amount (hundreds files) of the tests (especially in "oneIntel" context). Why we cannot rely on "dpcpplib" test reports? What we should do if that huge amount test will have divergence between dpcpplib and oneDPL repo?

Of source, in case if dpcpplib repo "is died" - situation more or less clear.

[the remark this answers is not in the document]
SergeyKopienko: @akukanov could you please take a look at https://github.com/oneapi-src/oneDPL/pull/885#discussion_r1358096648 ?

ValentinaKats:
> @ValentinaKats could you please clarify, should we really have unique error text in each test? Or may we have the same error text in different test files and it's not the problem for our nightly tests and so on?

As far as I know, there is no strict requirement to provide unique error text in each test. In testing results, there is the name of a failed test and error text. For me, a combination of test name and error text looks unique enough to identify the issue.

SergeyKopienko:
> > @ValentinaKats could you please clarify, should we really have unique error text in each test? Or may we have the same error text in different test files and it's not the problem for our nightly tests and so on?
> 
> As far as I know, there is no strict requirement to provide unique error text in each test. In testing results, there is the name of a failed test and error text. For me, a combination of test name and error text looks unique enough to identify the issue.

Thanks, @ValentinaKats !

@MikeDvorskiy  - so I think that current error messages state in this PR is good enough.

test/std/algorithms/alg.sorting/alg.binary.search/binary.search/binary_search.pass.cpp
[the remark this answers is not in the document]
akukanov: applications.compilers.source.dpcpplib is an old, de-facto abandoned internal repository for oneDPL, Except for these tests, all its content was already moved to our main repository. I prefer to have everything related to oneDPL project in a single place, instead of two.

MikeDvorskiy (dismissed):
Basing on na assumption that  "dpcpplib repo is "not active" any more" and these tests are moved (not duplicated ) from  the mentioned "dpcpplib" repo into oneDPL repo and ones will be tested at a  oneDPL test cycle (no tests run duplications as well), I am OK with that changes.

test/CMakeLists.txt
```diff
@@ -90,6 +105,12 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
     endif()
 
     target_include_directories(${_test_name} PRIVATE "${CMAKE_CURRENT_LIST_DIR}")
+    if (${is_std_test})
+        target_include_directories(${_test_name} PRIVATE "${CMAKE_CURRENT_LIST_DIR}/support_stdlib")
+    else()
+        target_include_directories(${_test_name} PRIVATE "${CMAKE_CURRENT_LIST_DIR}/support")
+    endif()
```
dmitriy-sobolev, line 112: Adding paths to  `${CMAKE_CURRENT_LIST_DIR}/support_stdlib` and `${CMAKE_CURRENT_LIST_DIR}/support` is not necessary because `${CMAKE_CURRENT_LIST_DIR}` is already there.
The current tests include test utilities as `"support/some_utility_header"`. You can remove this section and include the newly created headers as `"support_stdlib/some_utility_header"`. I have no strong objections against the current approach, though.

test/CMakeLists.txt
```diff
@@ -57,15 +57,30 @@ add_custom_target(run-onedpl-tests
...
         string(TOLOWER "${CMAKE_BUILD_TYPE}" _build_type_in_lower)
     endif()
 
-    get_filename_component(_test_name ${test_source_file} NAME)
-    string(REPLACE "\.cpp" "" _test_name ${_test_name})
+    if (${is_std_test})
+        #process std tests: target name equal to parent folders names plus cpp file name without extension
+        file(RELATIVE_PATH _test_name "${CMAKE_CURRENT_SOURCE_DIR}" "${test_source_file}")
+        string(REPLACE "\.cpp" "" _test_name ${_test_name})
+        string(REPLACE "/" ";" _test_name_list ${_test_name})
```
dmitriy-sobolev, line 71: To make sure that the received path uses "/" as a separator on both Linux and Windows, you can use `TO_CMAKE_PATH` option while receiving the path through `file` command: https://cmake.org/cmake/help/v3.15/command/file.html#path-conversion.

SergeyKopienko:
@MikeDvorskiy, @akukanov, @dmitriy-sobolev FYI: all new tests has been moved into `test/xpu_api` folder.
Please review again.
Thanks!

test/CMakeLists.txt
SergeyKopienko: This change has been rolled back.

test/CMakeLists.txt
SergeyKopienko: This change has been rolled back.

dmitriy-sobolev (approved):
LGTM. I have not checked each test case in detail, though.

MikeDvorskiy (approved):
Re-approved after moving the files in another directory.
(I "trust and commit')

PR #886, opened 2023-04-13
Fixed the exit code issue

wills-int:
It seems that the jobs will show passed, although it didn't. For example, one of the jobs will still showed that it passed successfully when it states in the workflow " The system cannot find the path specified." I have fixed the issue and now it will show a RED(X) instead of success, if the job actually fails.


dmitriy-sobolev:
Let me provide an example of the problem being solved for other reviewers:
image

Some CI Job provided "green" status, while there were issues during its execution. We need to force "red" status in such cases in order not to skip issues unintentionally.

.github/workflows/ci.yml
```diff
@@ -178,6 +178,7 @@ jobs:
             -DCMAKE_CXX_COMPILER=${{ matrix.cxx_compiler }} -DONEDPL_BACKEND=${{ matrix.backend }} -DONEDPL_DEVICE_TYPE=${{ matrix.device_type }} ..
           make VERBOSE=1 -j${BUILD_CONCURRENCY} ${make_targets}
           ctest --timeout ${TEST_TIMEOUT} --output-on-failure ${ctest_flags}
+          exit 1
```
dmitriy-sobolev, line 181: Let's remove it, because it does not solve the initial problem. It forces "red" status for all the runs, even if they finish without issues.

PR #902, opened 2023-04-17
Remove unnecessary comparisons in minmax_element

### LINKED ISSUE #873: minmax may be faster than min and max
I tried to evaluate onedpl and thrust min/max functions. Please see the experimental results on an Nvidia V100 GPU.

./main 100000 1000
Total number of points: 264708400
Average execution time of thrust:min() and thrust:max(): 5681.332520 (us)
Average execution time of thrust:min_max(): 2873.853516 (us)
PASS

./main 100000 1000
Total number of points: 264708400
Average execution time of onedpl:min() and onedpl:max(): 8479.713867 (us)
Average execution time of onedpl:min_max(): 21022.216797 (us)
PASS

Reference
https://github.com/zjin-lcf/HeCBench/tree/master/minmax-cuda
https://github.com/zjin-lcf/HeCBench/tree/master/minmax-sycl

joeatodd:
This PR removes array index comparisons from the `minmax_element` algorithm which I believe are unnecessary. As with the `min_element` algorithm, I think we assume that the array position of `__a` is always lower than that of `__b`.

I believe this fixes #873. I ran the benchmark mentioned in that issue & recovered the performance:

```
Total number of points: 264708400
Average execution time of onedpl:min() and onedpl:max(): 21960.263672 (us)
Average execution time of onedpl:min_max(): 16694.007812 (us)
PASS
```
running on RTX 3060.

Note: I tried swapping the initial value & logic of `__chosen_for_max` as I thought it might improve performance but actually it got slightly slower.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/utils_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/utils_hetero.h
@@ -116,13 +116,11 @@ struct __identity_reduce_fn
     {
         using ::std::get;
         auto __chosen_for_min = __a;
-        auto __chosen_for_max = __b;
-        // if b "<" a or if b "==" a and b_index < a_index
-        if (__comp(get<2>(__b), get<2>(__a)) || (!__comp(get<2>(__a), get<2>(__b)) && get<0>(__b) < get<0>(__a)))
+        auto __chosen_for_max = __a;
+        if (__comp(get<2>(__b), get<2>(__a)))
             __chosen_for_min = __b;
-        // if a ">" b or if a "==" b and a_index > b_index
-        if (__comp(get<3>(__b), get<3>(__a)) || (!__comp(get<3>(__a), get<3>(__b)) && get<1>(__b) < get<1>(__a)))
-            __chosen_for_max = __a;
+        if (!__comp(get<3>(__b), get<3>(__a)))
+            __chosen_for_max = __b;
         auto __result = _ReduceValueType{get<0>(__chosen_for_min), get<1>(__chosen_for_max), get<2>(__chosen_for_min),
                                          get<3>(__chosen_for_max)};
 


julianmi:
Thank you, @joeatodd. This is perfect timing. I came to the same conclusion about the unnecessary checks and have a branch up for testing here: https://github.com/oneapi-src/oneDPL/tree/dev/julianmi/mimmax_performance. The changes are very close so I think we can go ahead with reviewing your changes.

include/oneapi/dpl/pstl/hetero/utils_hetero.h
```diff
@@ -116,13 +116,11 @@ struct __identity_reduce_fn
     {
         using ::std::get;
         auto __chosen_for_min = __a;
-        auto __chosen_for_max = __b;
```
julianmi, line 119: Why was this logic changed? I think the original version was easier to read and would omit the if not check.

joeatodd: I tried it both ways round, and found that this was slightly faster. I'd be interested to know if you find the same?

julianmi:
> This PR removes array index comparisons from the `minmax_element` algorithm which I believe are unnecessary. As with the `min_element` algorithm, I think we assume that the array position of `__a` is always lower than that of `__b`.

I agree that the array positions are as you stated. However, this check was used for a general implementation of the reduce algorithm that doesn't guarantee a processing order. If two elements share the same value, the first element for min and the last element for max should be returned. oneDPL currently uses a general implementation of the reduce algorithm that processes the reductions in order by work items. Thus, the proposed changes should be fine and are aligned with the `min_element` approach.

include/oneapi/dpl/pstl/hetero/utils_hetero.h
julianmi: I don't see runtime changes on A100 or PVC. How much difference did you observe? I would suggest keeping the original logic if it is not significantly faster. Also, the runtime might depend on the structure of the input data and which paths are taken in this comparison.

timmiesmith:
Approving to allow per-commit testing. A second approve from @julianmi will be needed before merging.

julianmi (approved):
Looks good and now matches the version I have tested and profiled.

MikeDvorskiy:
I think,  the statement "array position of __a is always lower than that of __b" is true just for serial algorithm and just for comparison call with the same arguments order..
For parallel implementation, based on reduce, for example, the mentioned above statement my be no true.

julianmi:
> I think, the statement "array position of __a is always lower than that of __b" is true just for serial algorithm and just for comparison call with the same arguments order.. For parallel implementation, based on reduce, for example, the mentioned above statement my be not true. (At least bey reason that reduce implementation usually doesn't rely on an argument order for a binary operation, other words, reduction usually requires a commutative binary operation)

I agree. I think we will need additional comments and refactoring for clarity. I will do this after this PR is merged.

danhoeflinger (commented):
I believe that the assumption "the array position of __a is always lower than that of __b" is true for the current state of the repository for this `minmax_element` implementation, which is the only user of this functor.  

`sycl::reduce_over_group()` would break this assumption, but the implementation avoids `sycl::reduce_over_group()` because the custom types used are not supported by that API.  The alternate function provided and the current overall algorithm structure should preserve the assumption otherwise.

I think this looks good.  However, the price for this performance gain is a new assumption for this algorithm.  I think we should document it with a comment in the functor.  Something like:

```
// Assumes (a_index < b_index), Reduction algorithm must preserve relative ordering of arguments to the binary operation.
```

We should also look at a better way to explicitly state and enforce these assumptions, but that belongs in a separate PR.  An issue was made to document this: https://github.com/oneapi-src/oneDPL/issues/904

include/oneapi/dpl/pstl/hetero/utils_hetero.h
```diff
@@ -117,11 +117,9 @@ struct __identity_reduce_fn
         using ::std::get;
         auto __chosen_for_min = __a;
         auto __chosen_for_max = __b;
-        // if b "<" a or if b "==" a and b_index < a_index
-        if (__comp(get<2>(__b), get<2>(__a)) || (!__comp(get<2>(__a), get<2>(__b)) && get<0>(__b) < get<0>(__a)))
+        if (__comp(get<2>(__b), get<2>(__a)))
```
MikeDvorskiy, line 120: Probably does it make sense to add a couple of  asserts like "asset(a_index < b_index)" here  and prior the second if?

MikeDvorskiy:
> // Assumes (a_index < b_index), Reduction algorithm must preserve relative ordering of arguments to the binary operation.

I suggest adding corresponding RT asserts in the code.

include/oneapi/dpl/pstl/hetero/utils_hetero.h
joeatodd: Thanks - good suggestion :+1:

joeatodd:
Thanks for useful suggestions. I've added a comment describing the assumption that `a_index < b_index` and a couple of asserts for same.

joeatodd (commented):
Thanks folks for useful suggestions. I've added a comment describing the index ordering assumption and a couple of asserts for same.

julianmi (approved):
Thanks for the changes. The clang-format suggestions seem unrelated to the changed code.

danhoeflinger:
> Thanks for the changes. The clang-format suggestions seem unrelated to the changed code.

I'll take this opportunity to mention https://github.com/oneapi-src/oneDPL/pull/715.  This could clean up the clang-formatting checks to be more relevant (especially with the proposed modification I mention most recently in the comments but haven't yet done to the PR).  I'd love discussion / feedback there.

include/oneapi/dpl/pstl/hetero/utils_hetero.h
akukanov: The assertions do not use the comparator but `operator<`, which seems a mistake.

julianmi: I think the comparator should be used on the values only. The assertions check if the indices are ordered such that we can leave out the special case if both values are the same.

PR #903, opened 2023-04-17
Fix compile errors if <oneapi/dpl/async> included before <oneapi/dpl/execution>

SergeyKopienko:
In this PR we fix two errors:
1. Compile error described in issue https://github.com/oneapi-src/oneDPL/issues/881 - if <oneapi/dpl/async> included before <oneapi/dpl/execution>
2. Compile error if we include <oneapi/dpl/async> and DONEDPL_BACKEND != dpcpp

Also in this PR we added two new tests with different orders of headers inclusion:
- [test/general/header_inclusion_order_async_0.pass.cpp](https://github.com/oneapi-src/oneDPL/blob/26c01a7c9090e67c9cf0c74f5d83cbe01773149e/test/general/header_inclusion_order_async_0.pass.cpp)
- [test/general/header_inclusion_order_async_1.pass.cpp](https://github.com/oneapi-src/oneDPL/blob/26c01a7c9090e67c9cf0c74f5d83cbe01773149e/test/general/header_inclusion_order_async_1.pass.cpp)


include/oneapi/dpl/async
```diff
@@ -19,13 +19,26 @@
...
-#include "oneapi/dpl/internal/async_extension_defs.h"
+#if _ONEDPL_BACKEND_SYCL
+#    if !_ONEDPL_ASYNC_FORWARD_DECLARED
+#        include "oneapi/dpl/internal/async_extension_defs.h"
+#        define _ONEDPL_ASYNC_FORWARD_DECLARED 1
+#    endif // !_ONEDPL_ASYNC_FORWARD_DECLARED
 
-#if _ONEDPL_EXECUTION_POLICIES_DEFINED
+#    if _ONEDPL_EXECUTION_POLICIES_DEFINED
 // If <execution> has already been included, pull in implementations
```
akukanov, line 29: I'd recommend to indent the comment same as `include` (ignoring the # sign).

include/oneapi/dpl/internal/async_impl/async_utils.h
akukanov, line 1: Can we remove this header completely, moving its content into another suitable utility header?

include/oneapi/dpl/async
```diff
@@ -19,13 +19,26 @@
...
+#endif     // _ONEDPL_BACKEND_SYCL
+
+namespace oneapi
+{
+namespace dpl
+{
+}; // namespace dpl
+}; // namespace oneapi
 
 namespace dpl = oneapi::dpl;
```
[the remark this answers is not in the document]
SergeyKopienko: From my point of view - we are unable to move this empty namespace definitions under ```#if _ONEDPL_BACKEND_SYCL``` - as I understand we should have declared ```dpl``` namespace after include of this header regardless of ```_ONEDPL_BACKEND_SYC``` state.

include/oneapi/dpl/async
SergeyKopienko: Fixed.

include/oneapi/dpl/execution
```diff
@@ -55,6 +55,12 @@
 #    include "oneapi/dpl/pstl/glue_numeric_ranges_impl.h"
 #endif // _ONEDPL_NUMERIC_RANGES_FORWARD_DECLARED
 
+#if _ONEDPL_ASYNC_FORWARD_DECLARED
+#    ifdef _ONEDPL_BACKEND_SYCL
+#        include "oneapi/dpl/internal/async_impl/glue_async_impl.h"
+#    endif // _ONEDPL_BACKEND_SYCL
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed.

include/oneapi/dpl/async
SergeyKopienko: But now we have clang format error here:

```
--- a/include/oneapi/dpl/async
+++ b/include/oneapi/dpl/async
@@ -26,7 +26,7 @@
 #    endif // !_ONEDPL_ASYNC_FORWARD_DECLARED
 
 #    if _ONEDPL_EXECUTION_POLICIES_DEFINED
-         // If <execution> has already been included, pull in implementations
+// If <execution> has already been included, pull in implementations
 #        include "oneapi/dpl/internal/async_impl/async_utils.h"
 #        include "oneapi/dpl/internal/async_impl/async_impl.h"
 #        include "oneapi/dpl/internal/async_impl/glue_async_impl.h"
Error: Process completed with exit code 1.
```

include/oneapi/dpl/internal/async_impl/async_utils.h
SergeyKopienko: Fixed: it's context was moved into ```include/oneapi/dpl/internal/async_extension_defs.h```

include/oneapi/dpl/async
```diff
@@ -19,13 +19,26 @@
...
+#        define _ONEDPL_ASYNC_FORWARD_DECLARED 1
+#    endif // !_ONEDPL_ASYNC_FORWARD_DECLARED
 
-#if _ONEDPL_EXECUTION_POLICIES_DEFINED
+#    if _ONEDPL_EXECUTION_POLICIES_DEFINED
 // If <execution> has already been included, pull in implementations
-#    include "oneapi/dpl/internal/async_impl/async_impl.h"
-#    include "oneapi/dpl/internal/async_impl/glue_async_impl.h"
-#endif // _ONEDPL_EXECUTION_POLICIES_DEFINED
+#        include "oneapi/dpl/internal/async_impl/async_utils.h"
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed: this include was removed.

include/oneapi/dpl/async
akukanov: I would say we should ignore this, as fixing it would make readability worse

include/oneapi/dpl/async
```diff
@@ -19,13 +19,25 @@
...
+#    endif // !_ONEDPL_ASYNC_FORWARD_DECLARED
 
-#if _ONEDPL_EXECUTION_POLICIES_DEFINED
-// If <execution> has already been included, pull in implementations
-#    include "oneapi/dpl/internal/async_impl/async_impl.h"
-#    include "oneapi/dpl/internal/async_impl/glue_async_impl.h"
-#endif // _ONEDPL_EXECUTION_POLICIES_DEFINED
+#    if _ONEDPL_EXECUTION_POLICIES_DEFINED
+         // If <execution> has already been included, pull in implementations
+#        include "oneapi/dpl/internal/async_impl/async_impl.h"
```
akukanov, line 30: Actually, should `async_impl.h` be included directly into the public header? If it only contains internal implementation details and no public API definitions, then it should be included into `glue_async_impl.h` but not here.

include/oneapi/dpl/internal/async_impl/async_utils.h
akukanov: Sorry Sergey, but this does not seem to be the right place. `async_extension_defs.h` is not a utility header but forward declarations header.

I think the best place is `pstl/hetero/dpcpp/execution_sycl_defs.h`

test/parallel_api/experimental/asynch_include_order_error.pass.cpp
```diff
@@ -0,0 +1,55 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#include "support/test_config.h"
+
+// Including `async` after `execution` and `numeric` compiles successfully.
+#include <oneapi/dpl/async>
+#include <oneapi/dpl/execution>
+#include <oneapi/dpl/numeric>
```
akukanov, line 21: The implementation of the test does not match the comment, and also the order of includes is the same as in the other test.

include/oneapi/dpl/async
akukanov: I see that it does include some public API. But then, why do we have two distinct implementation headers? Should we move everything into `glue_async_impl.h` (and remove any possible duplication)?

@MikeDvorskiy do you know what's the difference between the definitions in `async_impl.h` and `glue_async_impl.h`?

test/parallel_api/experimental/asynch_include_order_error.pass.cpp
SergeyKopienko: It was the reproducer of compile error from customer.
Removed.

akukanov: Just to note, I have not suggested that you should remove it. I only pointed out that there is a mismatch of the description (as I read it) and the behavior, and also that another test has exactly the same order of includes while we perhaps want to test different orders.

It's up to you to decide if a special regression test is needed or not. Actually, we have a number of header inclusion order tests in `test/general`, adding one more for `<async>` might be completely reasonable.

SergeyKopienko: I have removed this new test (with reproducer from the user) because I also changed the order of includes in the already existing test.
That is the reason why you see the same order of includes in the tests.

Two new tests were added too.

SergeyKopienko: But I should note that our approach in these tests of inclusion only partially generates and catching compile time error.

For example, by this new added tests on `main` branch I found only one error of two - when I have compiled this two new tests with `-DONEDPL_BACKEND=dpcpp`.

As in user example from https://github.com/oneapi-src/oneDPL/issues/881 the compile time error not on inclusion of `async` header but on later call of something from `async` header algorithms:
```
auto f = oneapi::dpl::experimental::reduce_async(policy, v, v + n, T(0), std::plus());
```

It's just for you info.

include/oneapi/dpl/async
SergeyKopienko: Fixed: inclusion of `async_impl.h` removed from public header `async'.

include/oneapi/dpl/async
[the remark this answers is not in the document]
SergeyKopienko: From my point of view - we should still have this declaration to avoid the compile error in our new tests
in case when we use ```-DONEDPL_BACKEND=tbb``` :
```
FAILED: test/CMakeFiles/header_inclusion_order_async_1.pass.dir/general/header_inclusion_order_async_1.pass.cpp.o 
/usr/bin/oneapi_compiler/linux_prod/compiler/linux/bin/icpx -DONEDPL_USE_DPCPP_BACKEND=0 -DPSTL_USE_DEBUG -DTBB_USE_DEBUG=1 -D_PSTL_TEST_SUCCESSFUL_KEYWORD=1 -I/oneDPL/test -I/oneDPL/include -isystem /usr/lib/oneapi-tbb-2021.5.0/include -fsycl -O0 -g -fno-fast-math -fopenmp-simd -MD -MT test/CMakeFiles/header_inclusion_order_async_1.pass.dir/general/header_inclusion_order_async_1.pass.cpp.o -MF test/CMakeFiles/header_inclusion_order_async_1.pass.dir/general/header_inclusion_order_async_1.pass.cpp.o.d -o test/CMakeFiles/header_inclusion_order_async_1.pass.dir/general/header_inclusion_order_async_1.pass.cpp.o -c /oneDPL/test/general/header_inclusion_order_async_1.pass.cpp
In file included from /oneDPL/test/general/header_inclusion_order_async_1.pass.cpp:18:
/oneDPL/include/oneapi/dpl/async:34:17: error: use of undeclared identifier 'oneapi'
namespace dpl = oneapi::dpl;
                ^
1 error generated.
```

include/oneapi/dpl/internal/async_impl/async_utils.h
SergeyKopienko: Fixed.

include/oneapi/dpl/async
SergeyKopienko: OK.

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
```diff
@@ -16,19 +16,23 @@
...
+#endif // _ONEDPL_BACKEND_SYCL
 
 namespace oneapi
 {
 namespace dpl
 {
 namespace experimental
 {
-
+#ifdef _ONEDPL_BACKEND_SYCL
```
MikeDvorskiy, line 35: I think we should not to wrap the functions below in `_ONEDPL_BACKEND_SYCL` macro, because the functions are a kind of "general" for host and hetero backends.

include/oneapi/dpl/async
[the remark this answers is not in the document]
akukanov: There should not be such an issue with what I have proposed before:

> Can't we just move `namespace dpl = oneapi::dpl;` to the scope of `#if _ONEDPL_BACKEND_SYCL` ... ?

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
MikeDvorskiy, line 35: I think we should not to wrap the functions below in _ONEDPL_BACKEND_SYCL macro, because the functions are a kind of "general" - for host and hetero policies.
But I see that `__enable_if_device_execution_policy` is used here... 
Essentially, this file should be "universal" for host and device,, and should use just `__enable_if_execution_policy` like in `glue_algorithm_impl.h`
And should be `async_impl_hetero.h` for hetero async patterns and `async_impl.h` for host async patterns.

test/parallel_api/experimental/asynch_include_order_error.pass.cpp
akukanov: Adding more checks, including API calls, to the body of the `header_*` tests is completely fine.

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
akukanov: Is it possible to do now, though? Ideally, yes, the files should be universal, but to the best of my knowledge we only have an implementation for device execution policies.

akukanov: As far as I understand, Sergey tries to address this issue:
> Compile error if we include <oneapi/dpl/async> and DONEDPL_BACKEND != dpcpp

The question is, what do we want to happen in this combination, which is essentially not supported? Perhaps the best thing is to add a `static_assert` with a meaningful error message.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -279,17 +279,32 @@ using __enable_if_convertible_to_events = ::std::enable_if_t<__is_convertible_to
...
+        !::std::is_convertible<_Op1, sycl::event>::value &&
+        oneapi::dpl::__internal::__is_convertible_to_event<_Events...>,
+    _T>::type;
+
+template <typename _ExecPolicy, typename _T, typename _Op1, typename _Op2, typename... _Events>
+using __enable_if_device_execution_policy_double_no_default = typename ::std::enable_if<
+    oneapi::dpl::__internal::__is_device_execution_policy<typename ::std::decay<_ExecPolicy>::type>::value &&
+        !::std::is_convertible<_Op1, sycl::event>::value && !::std::is_convertible<_Op2, sycl::event>::value &&
+        oneapi::dpl::__internal::__is_convertible_to_event<_Events...>,
+    _T>::type;
```
akukanov, line 307: Please change to use `enable_if_t`, `decay_t`

test/parallel_api/experimental/asynch_include_order_error.pass.cpp
SergeyKopienko: Fixed: some small test functional inserted into
[test/general/header_inclusion_order_async_0.pass.cpp](https://github.com/oneapi-src/oneDPL/blob/26c01a7c9090e67c9cf0c74f5d83cbe01773149e/test/general/header_inclusion_order_async_0.pass.cpp)
[test/general/header_inclusion_order_async_1.pass.cpp](https://github.com/oneapi-src/oneDPL/blob/26c01a7c9090e67c9cf0c74f5d83cbe01773149e/test/general/header_inclusion_order_async_1.pass.cpp)

include/oneapi/dpl/async
SergeyKopienko: Fixed: ```async_impl.h``` was removed.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: Fixed.

include/oneapi/dpl/async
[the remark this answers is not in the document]
SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -279,17 +279,32 @@ using __enable_if_convertible_to_events = ::std::enable_if_t<__is_convertible_to
...
+template <typename _ExecPolicy, typename _T, typename _Op1, typename... _Events>
+using __enable_if_device_execution_policy_single_no_default = typename ::std::enable_if_t<
+    oneapi::dpl::__internal::__is_device_execution_policy<typename ::std::decay_t<_ExecPolicy>>::value &&
+        !::std::is_convertible<_Op1, sycl::event>::value &&
+        oneapi::dpl::__internal::__is_convertible_to_event<_Events...>,
+    _T>;
+
+template <typename _ExecPolicy, typename _T, typename _Op1, typename _Op2, typename... _Events>
+using __enable_if_device_execution_policy_double_no_default = typename ::std::enable_if_t<
+    oneapi::dpl::__internal::__is_device_execution_policy<typename ::std::decay_t<_ExecPolicy>>::value &&
```
akukanov, line 304: `typename` is not needed with `enable_if_t` etc. because it is already a part of the alias definition

SergeyKopienko: Fixed.

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
akukanov: > Up to me in this case we should have ability to include headers in any order with any backend without errors.

I am not sure if it makes any sense, though, to include `<async>` if de-facto it cannot be used. Having such ability does not help code portability at all, in my opinion. 

For an example of that. look at `asynch.pass.cpp`. Yes you can include the header with no preprocessor conditions, but all the code in the test still has to be covered by the same conditions as before. I even think that having `#if...#endif` around the include makes it more clear that this functionality is not always available, and helps to understand the use of the same preprocessor checks later in the code.

In other words, I think the problem is not worth fixing at the moment, other than improving the diagnostics by adding a static assert directly into the header. Adding static asserts only when async functions are used seems to me too much (of work and extra code) for no practical value.

include/oneapi/dpl/execution
```diff
@@ -55,6 +55,10 @@
 #    include "oneapi/dpl/pstl/glue_numeric_ranges_impl.h"
 #endif // _ONEDPL_NUMERIC_RANGES_FORWARD_DECLARED
 
+#if _ONEDPL_ASYNC_FORWARD_DECLARED
```
MikeDvorskiy, line 58: I think execution should not contain async API related stuff, because async API is a kind of a option and is enabled by explicit including <async> header. (the <range> has the same "option" logic)

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
```diff
@@ -16,18 +16,30 @@
...
 namespace dpl
 {
 namespace experimental
 {
+#if _ONEDPL_BACKEND_SYCL
+
+// [wait_for_all]
+template <typename... _Ts>
+oneapi::dpl::__internal::__enable_if_convertible_to_events<void, _Ts...>
+wait_for_all(_Ts&&... __events)
```
MikeDvorskiy, line 38: The function has sycl specific code.... I would suggest moving it to `async_impl_hetero.h` f.e.

test/general/header_inclusion_order_async_1.pass.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    using T = float;
+    T* v = sycl::malloc_device<T>(n, q);
+
+    q.fill<T>(v, 1, n).wait();
+
+    auto f = oneapi::dpl::experimental::reduce_async(my_policy, v, v + n, T(0), std::plus());
+
+    T value = f.get();
+    value = value;
```
MikeDvorskiy, line 42: forgot to remove?

SergeyKopienko: It's to avoid unused value warning

SergeyKopienko: Fixed.

test/general/header_inclusion_order_async_0.pass.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    using T = float;
+    T* v = sycl::malloc_device<T>(n, q);
+
+    q.fill<T>(v, 1, n).wait();
+
+    auto f = oneapi::dpl::experimental::reduce_async(my_policy, v, v + n, T(0), std::plus());
+
+    T value = f.get();
+    value = value;
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed.

include/oneapi/dpl/execution
akukanov: As I can see, the logic here is exactly the same as for other headers with the algorithms - inclusion of `<execution>` adds implementation if the declarations are already set by including the main header.

Do you mean that for `<async>` we should not at all use this two-stage logic with separate declarations and definitions, or what?

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
akukanov: `wait_for_all` is a public API, and so it equally makes sense to have it in this header.

Perhaps, designing and proposing some "backend API" for waiting makes sense, but for this patch I would just add a TODO comment or a task to address this later.

SergeyKopienko: ```TODO``` comment inserted into wait_for_all(_Ts&&... __events) implementation

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
```diff
@@ -16,18 +16,31 @@
...
+#if _ONEDPL_BACKEND_SYCL
+#    include "../async_extension_defs.h"
 
-#if _ONEDPL_HETERO_BACKEND
-#    include "async_impl_hetero.h"
-#endif
+#    if _ONEDPL_HETERO_BACKEND
+#        include "async_impl_hetero.h"
+#    endif
+#endif // _ONEDPL_BACKEND_SYCL
```
akukanov, line 25: Let's remove `#if` and `#endif` in the middle.
The whole idea of distinguishing SYCL backends and "generic hetero" layer seems questionable to me at this point; will we ever have a non-SYCL hetero backend?
Also from the practical standpoint `_ONEDPL_HETERO_BACKEND` is always defined to 1 if `_ONEDPL_BACKEND_SYCL` is defined, so the check in the middle is really useless.

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
SergeyKopienko: Fixed:
- #if _ONEDPL_BACKEND_SYCL removed from
- - include/oneapi/dpl/internal/async_extension_defs.h
- - include/oneapi/dpl/internal/async_impl/glue_async_impl.h
- added check into include/oneapi/dpl/async :
```
#if !_ONEDPL_BACKEND_SYCL
#    error "Unable to include oneapi/dpl/async header because _ONEDPL_BACKEND_SYCL is not defined"
#endif // !_ONEDPL_BACKEND_SYCL
```
In tests now we include async header under define check:
```
#   include "oneapi/dpl/async"
```

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
SergeyKopienko: Fixed.

include/oneapi/dpl/async
```diff
@@ -19,11 +19,17 @@
 #include "oneapi/dpl/internal/common_config.h"
 #include "oneapi/dpl/pstl/onedpl_config.h"
 
-#include "oneapi/dpl/internal/async_extension_defs.h"
+#if !_ONEDPL_BACKEND_SYCL
+#    error "Unable to include oneapi/dpl/async header because _ONEDPL_BACKEND_SYCL is not defined"
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed.

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
```diff
@@ -17,17 +17,23 @@
...
 namespace dpl
 {
 namespace experimental
 {
+// [wait_for_all]
+template <typename... _Ts>
+oneapi::dpl::__internal::__enable_if_convertible_to_events<void, _Ts...>
+wait_for_all(_Ts&&... __events)
+{
+    // TODO required to move this implementation into another sycl-specific header file, for example - into async_impl_hetero.h
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

include/oneapi/dpl/async
```diff
@@ -19,11 +19,17 @@
...
+
+#if !_ONEDPL_ASYNC_FORWARD_DECLARED
+#    include "oneapi/dpl/internal/async_extension_defs.h"
+#    define _ONEDPL_ASYNC_FORWARD_DECLARED 1
+#endif // !_ONEDPL_ASYNC_FORWARD_DECLARED
 
 #if _ONEDPL_EXECUTION_POLICIES_DEFINED
-// If <execution> has already been included, pull in implementations
-#    include "oneapi/dpl/internal/async_impl/async_impl.h"
+     // If <execution> has already been included, pull in implementations
```
SergeyKopienko, line 32: // If <execution> has already been included, pull in implementations

SergeyKopienko: @akukanov up to you: my proposal to remove empty space in this line:
- like in other places with this comment;
- to fix clang format error.

If you are agree - please apply this change.
Thanks.

akukanov: I do not want to argue about formatting; it is not that important, and it does not look that bad with only one level of indentation. Do it as you like.

include/oneapi/dpl/execution
MikeDvorskiy: Sorry, I missed that we already have the same logic with "optional" functionality "ranges".
And yes, while async is not included explicitly, `if _ONEDPL_ASYNC_FORWARD_DECLARED ` is not defined and we don't "pull" async stuff.
OK.

include/oneapi/dpl/async
SergeyKopienko: Agree, thanks.
Fixed.

test/general/header_inclusion_order_async_0.pass.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    auto q = sycl::queue{TestUtils::default_selector};
+
+    using T = float;
+    T* v = sycl::malloc_device<T>(n, q);
+
+    q.fill<T>(v, 1, n).wait();
+
+    auto f = oneapi::dpl::experimental::reduce_async(my_policy, v, v + n, T(0), std::plus());
+    f.get();
```
MikeDvorskiy, line 42: To avoid additional questions... Probably, we can call just wait() ?

SergeyKopienko: Fixed.

test/general/header_inclusion_order_async_0.pass.cpp
reble, line 1: The inclusion order used by this test should be already covered by: 
https://github.com/oneapi-src/oneDPL/blob/main/test/parallel_api/experimental/asynch.pass.cpp 
we might not necessarily need it (only for consistency)

test/general/header_inclusion_order_async_1.pass.cpp
reble, line 1: Maybe moving this test to the experimental folder?

test/general/header_inclusion_order_async_1.pass.cpp
```diff
@@ -0,0 +1,49 @@
...
+
+    auto q = sycl::queue{TestUtils::default_selector};
+
+    using T = float;
+    T* v = sycl::malloc_device<T>(n, q);
+
+    q.fill<T>(v, 1, n).wait();
+
+    auto f = oneapi::dpl::experimental::reduce_async(my_policy, v, v + n, T(0), std::plus());
+    f.wait();
```
reble, line 42:
```suggestion
    auto result = f.get();
    EXPECT_TRUE(result == n, "wrong effect from async test inclusion order (I)");
```

reble (approved):
LGTM, some minor comments

test/general/header_inclusion_order_async_0.pass.cpp
SergeyKopienko: The idea was to make some new tests like already exists for other our public headers to check inclusion order.
And these tests doesn't depends on of other tests that's checks correctness of our algorithms and etc..
Without any big checks inside.

test/general/header_inclusion_order_async_1.pass.cpp
SergeyKopienko: I am not sure that this check is required here.
Really in tests like this and other same tests we don't checks runtime-errors and the correctness of our algorithms but compile errors due to different inclusion order or compile errors on algorithm call.
Not more.

test/general/header_inclusion_order_async_1.pass.cpp
SergeyKopienko: Up to me - I prefer to place these tests here.
For example - tests with checks of headers inclusion for our ```oneapi::dpl::experimental::ranges``` are placed here too.

test/general/header_inclusion_order_async_0.pass.cpp
```diff
@@ -0,0 +1,42 @@
+// -*- C++ -*-
+//===----------------------------------------------------------------------===//
```
MikeDvorskiy, line 2: It seems that file (and the second new test) header doesn't correspond "oneDPL standard file header"...
Please have a look https://github.com/oneapi-src/oneDPL/blob/main/test/general/header_inclusion_order_algorithm_0.pass.cpp#L4

SergeyKopienko: For our new tests, which are not imported from LLVM but created by us, the next part in the headers of test files are not  required:
```
//
// This file incorporates work covered by the following copyright and permission
// notice:
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
```

MikeDvorskiy: I see...

test/general/header_inclusion_order_async_0.pass.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    allocator alloc(q);
+    std::vector<T, allocator> data(n, 1, alloc);
+
+    auto f = oneapi::dpl::experimental::reduce_async(TestUtils::make_device_policy(q), data.begin(), data.end());
+    f.wait();
+
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+    return TestUtils::done();
```
MikeDvorskiy, line 41: I think we should pass `TEST_DPCPP_BACKEND_PRESENT` to indicate the test system that a test is skipped in case of DPCPP is not present.

test/general/header_inclusion_order_async_1.pass.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+    allocator alloc(q);
+    std::vector<T, allocator> data(n, 1, alloc);
+
+    auto f = oneapi::dpl::experimental::reduce_async(TestUtils::make_device_policy(q), data.begin(), data.end());
+    f.wait();
+
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+    return TestUtils::done();
```
MikeDvorskiy, line 41: I think we should pass TEST_DPCPP_BACKEND_PRESENT to indicate the test system that a test is skipped in case of DPCPP is not present.

SergeyKopienko: Fixed.

MikeDvorskiy (approved):
LGTM

PR #906, opened 2023-04-20
Add check for sycl header before enabling sycl backend

danhoeflinger:
Adding checks for the presence of sycl headers prior to enabling sycl backend.  
This is required when compilers support `-fsycl` option and define `SYCL_LANGUAGE_VERSION`, without making SYCL implementation available in practice.


timmiesmith (approved):
This looks good to me.

PR #907, opened 2023-04-24
Improve reduce pattern performance with mid-sized inputs

julianmi:
This patch refactors the reduce pattern by:
- Fixed numbers of `__iters_per_work_item` are used independent of the device. This unifies the similar `transform_reduce_known` and `transform_reduce_unknown` structs.
- The sequential kernel was removed since it does not provide significant performance improvements.
- The performance of mid-sized arrays is improved by reducing the `__iters_per_work_item` for the second kernel to the lowest value needed. This better balances the work across work items. If each work item has less than one element to work on, the work group size is further reduced if possible. This lowers the complexity of the work group reduction.
- Improved documentation and naming.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -153,7 +153,7 @@ struct __parallel_transform_reduce_mid_submitter<__work_group_size, __iters_per_
         _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
         _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item1);
 
-        sycl::buffer<_Tp> __temp(sycl::range<1>(1 * __n_groups));
+        sycl::buffer<_Tp> __temp((sycl::range<1>(__n_groups)));
 
         sycl::event __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -359,7 +359,8 @@ struct __parallel_transform_reduce_impl
     }
 }; // struct __parallel_transform_reduce_impl
 
-// General version of parallel_transform_reduce. Processes elements in order to enable non-commutative algorithms.
+// General version of parallel_transform_reduce.
+// The binary operator must be associative but commutativity is not required since the elements are processed in order.
 // Each work item transforms and reduces __iters_per_work_item elements from global memory and stores the result in SLM.
 // 32 __iters_per_work_item was empirically found best for typical devices.
 // Each work group of size __work_group_size reduces the preliminary results of each work item in a group reduction
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -385,90 +385,89 @@ __parallel_transform_reduce(_ExecutionPolicy&& __exec, _ReduceOp __reduce_op, _T
     ::std::size_t __work_group_size = oneapi::dpl::__internal::__slm_adjusted_work_group_size(__exec, sizeof(_Tp) * 2);
 
     // Use single work group implementation if array < __work_group_size * __iters_per_work_item.
-    if (__n <= 256 && __work_group_size >= 256)
+    if (__work_group_size >= 256)
     {
-        return __parallel_transform_reduce_small_impl<256, 1, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                   __reduce_op, __transform_op, __init,
-                                                                   ::std::forward<_Ranges>(__rngs)...);
-    }
-    else if (__n <= 512 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_small_impl<256, 2, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                   __reduce_op, __transform_op, __init,
-                                                                   ::std::forward<_Ranges>(__rngs)...);
-    }
-    else if (__n <= 1024 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_small_impl<256, 4, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                   __reduce_op, __transform_op, __init,
-                                                                   ::std::forward<_Ranges>(__rngs)...);
-    }
-    else if (__n <= 2048 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_small_impl<256, 8, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                   __reduce_op, __transform_op, __init,
-                                                                   ::std::forward<_Ranges>(__rngs)...);
-    }
-    else if (__n <= 4096 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_small_impl<256, 16, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                    __reduce_op, __transform_op, __init,
-                                                                    ::std::forward<_Ranges>(__rngs)...);
-    }
-    else if (__n <= 8192 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_small_impl<256, 32, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                    __reduce_op, __transform_op, __init,
-                                                                    ::std::forward<_Ranges>(__rngs)...);
-    }
+        if (__n <= 256)
+        {
+            return __parallel_transform_reduce_small_impl<256, 1, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                       __reduce_op, __transform_op, __init,
+                                                                       ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 512)
+        {
+            return __parallel_transform_reduce_small_impl<256, 2, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                       __reduce_op, __transform_op, __init,
+                                                                       ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 1024)
+        {
+            return __parallel_transform_reduce_small_impl<256, 4, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                       __reduce_op, __transform_op, __init,
+                                                                       ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 2048)
+        {
+            return __parallel_transform_reduce_small_impl<256, 8, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                       __reduce_op, __transform_op, __init,
+                                                                       ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 4096)
+        {
+            return __parallel_transform_reduce_small_impl<256, 16, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                        __reduce_op, __transform_op, __init,
+                                                                        ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 8192)
+        {
+            return __parallel_transform_reduce_small_impl<256, 32, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                        __reduce_op, __transform_op, __init,
+                                                                        ::std::forward<_Ranges>(__rngs)...);
+        }
 
-    // Use two-step tree reduction.
-    // First step reduces __work_group_size * __iters_per_work_item1 elements.
-    // Second step reduces __work_group_size * __iters_per_work_item2 elements.
-    else if (__n <= 2097152 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_mid_impl<256, 32, 1, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                     __reduce_op, __transform_op, __init,
-                                                                     ::std::forward<_Ranges>(__rngs)...);
+        // Use two-step tree reduction.
+        // First step reduces __work_group_size * __iters_per_work_item1 elements.
+        // Second step reduces __work_group_size * __iters_per_work_item2 elements.
+        else if (__n <= 2097152)
+        {
+            return __parallel_transform_reduce_mid_impl<256, 32, 1, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                         __reduce_op, __transform_op, __init,
+                                                                         ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 4194304)
+        {
+            return __parallel_transform_reduce_mid_impl<256, 32, 2, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                         __reduce_op, __transform_op, __init,
+                                                                         ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 8388608)
+        {
+            return __parallel_transform_reduce_mid_impl<256, 32, 4, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                         __reduce_op, __transform_op, __init,
+                                                                         ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 16777216)
+        {
+            return __parallel_transform_reduce_mid_impl<256, 32, 8, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                         __reduce_op, __transform_op, __init,
+                                                                         ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 33554432)
+        {
+            return __parallel_transform_reduce_mid_impl<256, 32, 16, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                          __reduce_op, __transform_op, __init,
+                                                                          ::std::forward<_Ranges>(__rngs)...);
+        }
+        else if (__n <= 67108864)
+        {
+            return __parallel_transform_reduce_mid_impl<256, 32, 32, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
+                                                                          __reduce_op, __transform_op, __init,
+                                                                          ::std::forward<_Ranges>(__rngs)...);
+        }
     }
-    else if (__n <= 4194304 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_mid_impl<256, 32, 2, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                     __reduce_op, __transform_op, __init,
-                                                                     ::std::forward<_Ranges>(__rngs)...);
-    }
-    else if (__n <= 8388608 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_mid_impl<256, 32, 4, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                     __reduce_op, __transform_op, __init,
-                                                                     ::std::forward<_Ranges>(__rngs)...);
-    }
-    else if (__n <= 16777216 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_mid_impl<256, 32, 8, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                     __reduce_op, __transform_op, __init,
-                                                                     ::std::forward<_Ranges>(__rngs)...);
-    }
-    else if (__n <= 33554432 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_mid_impl<256, 32, 16, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
-                                                                      __reduce_op, __transform_op, __init,
-                                                                      ::std::forward<_Ranges>(__rngs)...);
-    }
-    else if (__n <= 67108864 && __work_group_size >= 256)
-    {
-        return __parallel_transform_reduce_mid_impl<256, 32, 32, _Tp>(::std::forward<_ExecutionPolicy>(__exec), __n,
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -113,30 +139,74 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
...
+        __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
+            __cgh.depends_on(__reduce_event);
+
+            auto __temp_acc = __temp.template get_access<sycl::access_mode::read>(__cgh);
+            auto __res_acc = __res.template get_access<sycl::access_mode::write>(__cgh);
+            __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__work_group_size2), __cgh);
+
+            __cgh.parallel_for<_LeafName...>(
+                sycl::nd_range<1>(sycl::range<1>(__work_group_size2), sycl::range<1>(__work_group_size2)),
+                [=](sycl::nd_item<1> __item_id) {
```
mmichel11, line 203: This kernel is identical to the kernel used in `__parallel_transform_reduce_small_submitter`. Would it make sense to unify this lambda body in a separate functor so that this code is not duplicated?

julianmi: Thanks, I will change that.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -298,114 +364,93 @@ __parallel_transform_reduce(_ExecutionPolicy&& __exec, _ReduceOp __reduce_op, _T
...
-    // Pessimistically double the memory requirement to take into account memory used by compiled kernel
-    // TODO: find a way to generalize getting of reliable work-group size
+    // Get the work group size adjusted to the local memory limit.
+    // Pessimistically double the memory requirement to take into account memory used by compiled kernel.
+    // TODO: find a way to generalize getting of reliable work-group size.
     ::std::size_t __work_group_size = oneapi::dpl::__internal::__slm_adjusted_work_group_size(__exec, sizeof(_Tp) * 2);
 
-    if (__n <= 65536 && __work_group_size >= 512)
+    // Use single work group implementation if array < __work_group_size * __iters_per_work_item.
+    if (__n <= 256 && __work_group_size >= 256)
```
mmichel11, line 373: Would it be a better practice to turn all subsequent `if` statements into `else if` statements?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -113,30 +139,74 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
...
+        __n = __n_groups;
+
+        // Lower the work group size of the second kernel to the next power of 2 if __n < __work_group_size.
+        auto __work_group_size2 = __work_group_size;
+        if constexpr (__iters_per_work_item2 == 1)
+        {
+            if (__n < __work_group_size)
+            {
+                __work_group_size2 = __n;
+                if ((__work_group_size2 & (__work_group_size2 - 1)) != 0)
```
mmichel11, line 186: I see in our codebase that we have `__dpl_bit_floor` but no equivalent `__dpl_bit_ceil`. Am I correct that this if statement, bit floor, and bit shift could be replaced with a single bit ceil operation?

Since `std::bit_ceil` is in the C++20 standard (https://en.cppreference.com/w/cpp/numeric/bit_ceil), it may be useful to add an equivalent `__dpl_bit_ceil` into our utilities.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: I don't have a strong opinion on this but `else if` might be easier to read.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: This is correct. I will change that to `__dpl_bit_ceil ` once #909 is merged.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
danhoeflinger: Doesn't `else if` avoid a lot of unnecessary branching for a single control flow (once you hit your `true` case, you skip all the rest of the branching)?
Unless we have a explicit need to have consecutive `if` statements, I think `else if` is correct.

julianmi: While I agree in general, this routine uses return statements in the if statements.

danhoeflinger: Ah yes, my mistake.  I'll still put my vote in for `else if` because that is the norm here I think.
However, I'll retract my previous statement.

julianmi: I switched to `else if` statements.

danhoeflinger: Also for this set of `if`/`else if` statements:  They are all checking `__work_group_size >=256`, does it make sense to pull this out and above?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -113,102 +170,124 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
...
+        auto __transform_pattern1 =
+            unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item1, _ReduceOp, _TransformOp>{
                 __reduce_op, _TransformOp{__transform_op}};
+        auto __transform_pattern2 =
+            unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item2, _ReduceOp, _NoOpFunctor>{
+                __reduce_op, _NoOpFunctor{}};
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
 
-        const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
+        _Size __size_per_work_group =
```
danhoeflinger, line 182: Can this be `constexpr`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -113,102 +170,124 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
...
+        sycl::buffer<_Tp> __res(sycl::range<1>(1));
+
+        __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
+            __cgh.depends_on(__reduce_event);
+
+            auto __temp_acc = __temp.template get_access<sycl::access_mode::read>(__cgh);
+            auto __res_acc = __res.template get_access<sycl::access_mode::write>(__cgh);
+            __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__work_group_size2), __cgh);
+
+            __cgh.parallel_for<_LeafName...>(
```
danhoeflinger, line 225: We are using `_LeafName...` as the name for this kernel which includes a `work_group_size` which may deviate from the actually used one:  `work_group_size2` which can depend on `__n`.  Can this get us into trouble?  
If not, should `work_group_size` be removed from the name?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -113,102 +170,124 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
...
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
 
-        const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
+        _Size __size_per_work_group =
+            __iters_per_work_item1 * __work_group_size; // number of buffer elements processed within workgroup
+        _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
+        _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item1);
 
-        sycl::buffer<_Tp> __res(sycl::range<1>(1));
+        sycl::buffer<_Tp> __temp((sycl::range<1>(__n_groups)));
```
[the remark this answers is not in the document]
julianmi: This solves the ambiguity of the targeted variable definition vs. a potential function declaration.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: Good point. I think we should remove `work_group_size` from the generated leaf kernel name since it has no meaning. I think we can even remove `work_group_size` from all kernel names since we only use 256.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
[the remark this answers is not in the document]
danhoeflinger: Ah yes, thanks for clarification.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -34,25 +34,70 @@ namespace __par_backend_hetero
...
-template <typename _Tp, typename _KernelName>
-struct __parallel_transform_reduce_seq_submitter;
+// Parallel_transform_reduce for a small arrays using a single work group.
+// Transforms and reduces __work_group_size * __iters_per_work_item elements.
+template <::std::uint16_t __work_group_size, ::std::size_t __iters_per_work_item, typename _Tp, typename _KernelName>
+struct __parallel_transform_reduce_small_submitter;
 
-template <typename _Tp, typename... _Name>
-struct __parallel_transform_reduce_seq_submitter<_Tp, __internal::__optional_kernel_name<_Name...>>
+template <::std::uint16_t __work_group_size, ::std::size_t __iters_per_work_item, typename _Tp, typename... _Name>
```
mmichel11, line 98: Since the input size is small here, can `__iters_per_work_item` be a type smaller than `std::size_t`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -288,7 +358,15 @@ struct __parallel_transform_reduce_impl
     }
 }; // struct __parallel_transform_reduce_impl
 
-// General version of parallel_transform_reduce. Calls optimized kernels.
+// General version of parallel_transform_reduce. Processes elements in order to enable non-commutative algorithms.
```
dmitriy-sobolev, line 361: I do not quite understand the meaning of "Processes elements in order to enable non-commutative algorithms". Does it mean that this implementation provides a deterministic result when `_ReduceOp` predicate is non-commutative?

It may use `sycl::reduce_over_group` which has the following note:
> The result of a call to these functions is non-deterministic if the binary operator is not commutative and associative

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -34,25 +34,70 @@ namespace __par_backend_hetero
...
+    }
+}
+
+// Device kernel that transforms and reduces __n elements to the number of work groups preliminary results.
+template <typename _Tp, typename _NDItemId, typename _Size, typename _TransformPattern, typename _ReducePattern,
+          typename _InitType, typename _AccLocal, typename _Tmp, typename... _Acc>
+void
+__device_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Size __n_items,
+                       const _TransformPattern __transform_pattern, const _ReducePattern __reduce_pattern,
+                       const _InitType __init, const _AccLocal& __local_mem, _Tmp& __temp_acc, const _Acc&... __acc)
```
dmitriy-sobolev, line 75: Const is missing for `_Tmp& __temp_acc`, `__init` is not used.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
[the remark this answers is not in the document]
julianmi: I switched to curly braces for clarity.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: Exactly. This version does not require commutative operators. I tried to clarify the comment.

`sycl::reduce_over_group` is used for the group reduction if the type is arithmetic and the operator is commutative (cf. `__has_known_identity`). We process elements in order otherwise.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -61,50 +106,61 @@ struct __parallel_transform_reduce_seq_submitter<_Tp, __internal::__optional_ker
...
 }
 
-// Parallel_transform_reduce for a single work group
-template <::std::uint16_t __work_group_size, ::std::size_t __iters_per_work_item, typename _Tp, typename _KernelName>
-struct __parallel_transform_reduce_small_submitter;
-
-template <::std::uint16_t __work_group_size, ::std::size_t __iters_per_work_item, typename _Tp, typename... _Name>
-struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_per_work_item, _Tp,
-                                                   __internal::__optional_kernel_name<_Name...>>
+// Parallel_transform_reduce for a mid-sized arrays using two reduction steps.
```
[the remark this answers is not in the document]
julianmi: Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -61,50 +106,61 @@ struct __parallel_transform_reduce_seq_submitter<_Tp, __internal::__optional_ker
     operator()(_ExecutionPolicy&& __exec, _Size __n, _ReduceOp __reduce_op, _TransformOp __transform_op,
                _InitType __init, _Ranges&&... __rngs) const
     {
-        auto __transform_pattern = unseq_backend::transform_reduce_seq<_ExecutionPolicy, _ReduceOp, _TransformOp, _Tp>{
-            __reduce_op, _TransformOp{__transform_op}};
+        auto __transform_pattern =
+            unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _TransformOp>{
+                __reduce_op, _TransformOp{__transform_op}};
```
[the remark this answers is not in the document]
julianmi: Applied.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -113,102 +169,122 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
...
+        _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
+        _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item1);
 
-        sycl::buffer<_Tp> __res(sycl::range<1>(1));
+        sycl::buffer<_Tp> __temp((sycl::range<1>(__n_groups)));
 
         sycl::event __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
-            auto __res_acc = __res.template get_access<access_mode::write>(__cgh);
+            auto __temp_acc = __temp.template get_access<sycl::access_mode::write>(__cgh);
```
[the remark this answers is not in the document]
julianmi: I did not in my tests, but I changed it to provided additional information to the compiler. It might be useful on different platforms or in the future.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -113,102 +169,122 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
...
+            if (__n < __work_group_size)
+            {
+                __work_group_size2 = __n;
+                if ((__work_group_size2 & (__work_group_size2 - 1)) != 0)
+                    __work_group_size2 = oneapi::dpl::__internal::__dpl_bit_floor(__work_group_size2) << 1;
+            }
+        }
+        __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item2);
+
+        sycl::buffer<_Tp> __res(sycl::range<1>(1));
```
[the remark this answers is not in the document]
julianmi: Allocating a buffer of size `__tmp` + 1 for the result is faster than allocation two separate buffers. However, the current implementation transfers whole buffers instead of sub-buffers. The additional transfer costs are higher than the allocation costs when using an extended `__tmp` buffer.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: Applied.

julianmi:
I have addressed the comments so far except for using `__dpl_bit_ceil ` which requires #909 to be merged.

I have further verified that this patch improves the performance on CPU and GPU devices. Thus, the previous CPU-specific code path with distributing one work group per compute unit is not required anymore. This patch will use more work groups which might even be beneficial to hide latencies.

The clang-format hints are unrelated to the code changed in this PR.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
dmitriy-sobolev: Thanks for the clarification! 

The following is not related to the PR and thus does not block it. I was not able to find a statement that the predicates with available known identities are both associative and commutative. `sycl::minimum`, one of these predicates, returns
the first argument when the arguments are equivalent, so the result for `sycl::minimum(a, b)` and `sycl::minimum(b, a)` are different when `a == b`. Could you open an issue if you think it is a valid case?

julianmi: I agree with @danhoeflinger. The spec for `sycl::reduce_over_group` (https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_reduce) specifies the binary operators in https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:function-objects which include `sycl::minimum` and `sycl::maximum`. The only problem I see is with algorithms that use minimum or maximum but the order matters such as for `min_element()` where the first index with that value needs to be returned. Our implementation uses a user-defined type then and we fall back to an in-order group reduction.

danhoeflinger: Let me make a minor correction to what I said:  for `sycl::minimum` and `sycl::maximum` it `has_known_identity` specifically for integral types, floating point types and sycl::half rather than all arithmetic types.  This leaves out complex types (which is good).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -34,181 +34,258 @@ namespace __par_backend_hetero
...
+
+        // Lower the work group size of the second kernel to the next power of 2 if __n < __work_group_size.
+        auto __work_group_size2 = __work_group_size;
+        if constexpr (__iters_per_work_item2 == 1)
+        {
+            if (__n < __work_group_size)
+            {
+                __work_group_size2 = __n;
+                if ((__work_group_size2 & (__work_group_size2 - 1)) != 0)
+                    __work_group_size2 = oneapi::dpl::__internal::__dpl_bit_floor(__work_group_size2) << 1;
```
dmitriy-sobolev, line 211: `__work_group_size2` may become larger than `__work_group_size`. It may result in allocation more SLM than available on the device.

julianmi: I don't think it will. `__work_group_size` is set to a power of two (currently 256). `__work_group_size2` uses the bit ceiling of `__n` only if `__n` is smaller than `__work_group_size`. The maximum of `__work_group_size2` should thus be `__work_group_size`.

The goal of this code is to reduce the complexity of the work group reduction if a smaller number of preliminary results from the first kernel than the `__work_group_size` need to be reduced in the second kernel.

dmitriy-sobolev (approved):
LGTM. 

I think, a better title for the PR is "Improve reduce pattern performance with mid-sized inputs".

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -34,181 +34,258 @@ namespace __par_backend_hetero
...
-          typename _TransformOp, typename _ExecutionPolicy, typename _Size, typename _InitType,
+template <::std::uint16_t __work_group_size, ::std::uint8_t __iters_per_work_item1,
+          ::std::uint8_t __iters_per_work_item2, typename _Tp, typename _ReduceOp, typename _TransformOp,
+          typename _ExecutionPolicy, typename _Size, typename _InitType,
           oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0, typename... _Ranges>
 auto
-__parallel_transform_reduce_small_impl(_ExecutionPolicy&& __exec, _Size __n, _ReduceOp __reduce_op,
-                                       _TransformOp __transform_op, _InitType __init, _Ranges&&... __rngs)
+__parallel_transform_reduce_mid_impl(_ExecutionPolicy&& __exec, _Size __n, const _ReduceOp& __reduce_op,
+                                     const _TransformOp& __transform_op, const _InitType& __init, _Ranges&&... __rngs)
```
rarutyun, line 243: Just curious why Callables are now passed by `const` reference. Why is it better (keeping in mind the in the APIs all user callables come by value)?

PR #909, opened 2023-04-24
Specialize copy_if for small input sizes

adamfidel:
This PR provides an optimization for the `copy_if` algorithm for sizes that are small enough to fit into a single work-group. This should improve the performance of `copy_if`, `remove`, `remove_if`, `remove_copy`,  and `remove_copy_if`.


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -577,6 +580,103 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
+                    constexpr bool __can_use_subgroup_load_store = false;
+#endif
+
+                    if constexpr (__can_use_subgroup_load_store)
+                    {
+                        _ONEDPL_PRAGMA_UNROLL
+                        for (::std::uint16_t __i = 0; __i < _ElemsPerItem; ++__i)
+                        {
+                            auto __idx = __i * _WGSize + __subgroup_id * __subgroup_size;
+                            uint16_t __val = __unary_op(__subgroup.load(__in_rng.begin() + __idx));
```
MikeDvorskiy, line 636: I think `in_rng.begin()` doesn't work in case of pipelined range with our "simple" views like
`***_view_simple`, because ones don't have begin/end operator, just `operator[]`.
We didnot need begin/end before and added simplest implementation for the "views" which provide the necessary semantic.

MikeDvorskiy:
> This PR provides an optimization for the `copy_if` algorithm for sizes that are small enough to fit into a single work-group. This should improve the performance of `copy_if`, `remove`, `remove_if`, `remove_copy`, and `remove_copy_if`.

Am I right that we are introducing two new patterns to our SYCL backend - `parallel_scan_copy `and `parallel_copy_if` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -683,6 +783,123 @@ __pattern_transform_scan_multi_group(_ExecutionPolicy&& __exec, _InRng&& __in_rn
...
+    ::std::size_t __max_wg_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
+
+    if (__n <= __single_group_upper_limit && __max_slm_size >= __req_slm_size &&
+        __max_wg_size >= _SingleGroupInvoker::__targeted_wg_size)
+    {
+        using _SizeBreakpoints =
+            ::std::integer_sequence<::std::uint16_t, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384>;
+
+        return __par_backend_hetero::__static_monotonic_dispatcher<_SizeBreakpoints>::__dispatch(
+            _SingleGroupInvoker{}, __n, ::std::forward<_ExecutionPolicy>(__exec), __n, ::std::forward<_InRng>(__in_rng),
```
MikeDvorskiy, line 886: I guess should be  commented what is `monotonic_dispatcher` and why we use it

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -683,6 +783,123 @@ __pattern_transform_scan_multi_group(_ExecutionPolicy&& __exec, _InRng&& __in_rn
...
+    // Next power of 2 greater than or equal to __n
+    auto __n_uniform = ::oneapi::dpl::__internal::__dpl_bit_ceil(static_cast<::std::make_unsigned_t<_Size>>(__n));
+
+    // The kernel stores n integers for the predicate and another n integers for the offsets
+    // Pessimistically only use half of the memory to take into account memory used by compiled kernel
+    const ::std::size_t __max_slm_size =
+        __exec.queue().get_device().template get_info<sycl::info::device::local_mem_size>() / 4;
+    const auto __req_slm_size = sizeof(::std::uint16_t) * __n_uniform;
+
+    constexpr int __single_group_upper_limit = 16384;
```
MikeDvorskiy, line 875: probably, `::std::uint16_t` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -577,6 +580,103 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
+                sycl::nd_range<1>(_WGSize, _WGSize), [=](sycl::nd_item<1> __self_item) {
+                    const auto& __group = __self_item.get_group();
+                    const auto& __subgroup = __self_item.get_sub_group();
+                    // This kernel is only launched for sizes less than 2^16
+                    const ::std::uint16_t __item_id = __self_item.get_local_linear_id();
+                    const ::std::uint16_t __subgroup_id = __subgroup.get_group_id();
+                    const ::std::uint16_t __subgroup_size = __subgroup.get_local_linear_range();
+
+#if _ONEDPL_SYCL_SUB_GROUP_LOAD_STORE_PRESENT
+                    constexpr bool __can_use_subgroup_load_store = _IsFullGroup;
```
MikeDvorskiy, line 625: As far as I understand usage of "subgroup load/store" is enabling just for such rare case when "n==wg_size", right?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -577,6 +580,103 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
+struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _WGSize, _IsFullGroup,
+                                                        __internal::__optional_kernel_name<_ScanKernelName...>>
+{
+    template <typename _Policy, typename _InRng, typename _OutRng, typename _InitType, typename _BinaryOperation,
+              typename _UnaryOp>
+    auto
+    operator()(const _Policy& __policy, _InRng&& __in_rng, _OutRng&& __out_rng, ::std::size_t __n, _InitType __init,
+               _BinaryOperation __bin_op, _UnaryOp __unary_op)
+    {
+        using _ValueType = uint16_t;
```
julianmi, line 598: `::std::uint16_t`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -577,6 +580,103 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
+
+        // This type is used as a workaround for when an internal tuple is assigned to ::std::tuple, such as
+        // with zip_iterator
+        using __tuple_type = typename ::oneapi::dpl::__internal::__get_tuple_type<
+            typename ::std::decay_t<decltype(__in_rng[0])>, typename ::std::decay_t<decltype(__out_rng[0])>>::__type;
+
+        constexpr ::uint32_t __elems_per_wg = _ElemsPerItem * _WGSize;
```
julianmi, line 605: `::std::uint32_t`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -683,6 +783,123 @@ __pattern_transform_scan_multi_group(_ExecutionPolicy&& __exec, _InRng&& __in_rn
...
+        return __par_backend_hetero::__static_monotonic_dispatcher<_SizeBreakpoints>::__dispatch(
+            _SingleGroupInvoker{}, __n, ::std::forward<_ExecutionPolicy>(__exec), __n, ::std::forward<_InRng>(__in_rng),
+            ::std::forward<_OutRng>(__out_rng), __pred);
+    }
+    else
+    {
+        using _ReduceOp = ::std::plus<_Size>;
+
+        unseq_backend::__create_mask<_Pred, _Size> __create_mask_op{__pred};
+        unseq_backend::__copy_by_mask<_ReduceOp, oneapi::dpl::__internal::__pstl_assign, /*inclusive*/ ::std::true_type,
```
julianmi, line 894: Formatting looks strange to me.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1548,9 +1765,11 @@ struct __is_radix_sort_usable_for_type
 };
 
 #if _USE_RADIX_SORT
```
julianmi, line 1767: I'm not sure if this and the following changes improve the readability.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -683,6 +783,123 @@ __pattern_transform_scan_multi_group(_ExecutionPolicy&& __exec, _InRng&& __in_rn
...
+__parallel_copy_if(_ExecutionPolicy&& __exec, _InRng&& __in_rng, _OutRng&& __out_rng, _Size __n, _Pred __pred)
+{
+    using _SingleGroupInvoker = __invoke_single_group_copy_if<_Size>;
+
+    // Next power of 2 greater than or equal to __n
+    auto __n_uniform = ::oneapi::dpl::__internal::__dpl_bit_ceil(static_cast<::std::make_unsigned_t<_Size>>(__n));
+
+    // The kernel stores n integers for the predicate and another n integers for the offsets
+    // Pessimistically only use half of the memory to take into account memory used by compiled kernel
+    const ::std::size_t __max_slm_size =
```
julianmi, line 871: Would it make this code clearer to specify `__max_slm_size` as half `local_mem_size` and multiply `__req_slm_size` by two?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -683,6 +783,123 @@ __pattern_transform_scan_multi_group(_ExecutionPolicy&& __exec, _InRng&& __in_rn
         unseq_backend::__global_scan_functor<_Inclusive, _BinaryOperation, _InitType>{__binary_op, __init});
 }
 
+template <typename _SizeType>
+struct __invoke_single_group_copy_if
+{
+    // Specialization for devices that have a max work-group size of at least 1024
+    static constexpr ::std::uint16_t __targeted_wg_size = 1024;
```
julianmi, line 790: Did you experiment with the work group size? A smaller size could suffice to saturate the memory bandwidth but might reduce the complexity of the group scan.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -683,6 +783,123 @@ __pattern_transform_scan_multi_group(_ExecutionPolicy&& __exec, _InRng&& __in_rn
...
+
+    template <::std::uint16_t _Size, typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Pred>
+    auto
+    operator()(_ExecutionPolicy&& __exec, ::std::size_t __n, _InRng&& __in_rng, _OutRng&& __out_rng, _Pred&& __pred)
+    {
+        constexpr ::std::uint16_t __wg_size = ::std::min(_Size, __targeted_wg_size);
+        constexpr ::std::uint16_t __num_elems_per_item = ::oneapi::dpl::__internal::__dpl_ceiling_div(_Size, __wg_size);
```
julianmi, line 797: Similar comment as above about tuning this parameter. Other algorithms have shown best performance with 32 elements per work item while this uses 16 at most, I think.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +573,39 @@ class __future : private std::tuple<_Args...>
     }
 };
 
+template <typename>
```
julianmi, line 576: I think a comment about what this dispatcher does would help here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -577,6 +580,103 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
+        using _ValueType = uint16_t;
+
+        // This type is used as a workaround for when an internal tuple is assigned to ::std::tuple, such as
+        // with zip_iterator
+        using __tuple_type = typename ::oneapi::dpl::__internal::__get_tuple_type<
+            typename ::std::decay_t<decltype(__in_rng[0])>, typename ::std::decay_t<decltype(__out_rng[0])>>::__type;
+
+        constexpr ::uint32_t __elems_per_wg = _ElemsPerItem * _WGSize;
+
+        sycl::buffer<_Size> __res(sycl::range<1>(1));
```
julianmi, line 607: This algorithm might benefit from using USM host memory for the result similar to the discussions for reduce in #874.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: Will add documentation.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: Good catch, done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: Yes, that would be better. Will change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: This size was chosen for scan, but I did not do experiments specifically for `copy_if`. I will experiment a bit with these parameters and get back to you.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: I'll test out different parameters.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: This is from `clang-format`. Perhaps I should manually adjust it and ignore its suggestions for this line.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: This is from `clang-format`. I will revert the changes for lines that I did not touch.

adamfidel:
> > This PR provides an optimization for the `copy_if` algorithm for sizes that are small enough to fit into a single work-group. This should improve the performance of `copy_if`, `remove`, `remove_if`, `remove_copy`, and `remove_copy_if`.
> 
> Am I right that we are introducing two new patterns to our SYCL backend - `parallel_scan_copy `and `parallel_copy_if` ?

Yes, although  `parallel_scan_copy` is just moving code from `__pattern_scan_copy` in `algorithm_impl_hetero.h`. This is needed to provide a fallback path for when we cannot call the single-work-group copy_if specialization.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -577,6 +580,103 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
+                    const ::std::uint16_t __subgroup_id = __subgroup.get_group_id();
+                    const ::std::uint16_t __subgroup_size = __subgroup.get_local_linear_range();
+
+#if _ONEDPL_SYCL_SUB_GROUP_LOAD_STORE_PRESENT
+                    constexpr bool __can_use_subgroup_load_store = _IsFullGroup;
+#else
+                    constexpr bool __can_use_subgroup_load_store = false;
+#endif
+
+                    if constexpr (__can_use_subgroup_load_store)
```
mmichel11, line 630: Could we benefit from adding this group load into SLM with the strided access pattern as a utility function somewhere? I could see this being used quite a bit throughout different kernels. The same thing could be done for storing results in global memory from SLM.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: I tested out a few different parameters and they were within 1-2% of each other with no clear winner. Based on those experiments, I think I will just keep it to 1024 for now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: > guarded only for ranges that support `.begin()`.

I guess it is not enough here... `subgroup.load` does requires a pointer and probably doesn't  work with a general iterator.
Also, can we use `sycl::vec` instead of "subgroup" (intel extension) ?
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:vector.type

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: hm... Should we whether  a special code branch for so rare case, just a "bench" case ( "n==wg_size") ? 
What if we keep the code which processes "residual"? and If residual is zero, does a compiler get rid of the "residual code"?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: 1. I replaced the condition with one that checks if `begin()` returns a pointer.
2. I don't see how we can use `sycl::vec` here since this code is mainly just trying to pull items from global memory in an efficient way.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: I think the use of the unary operator here makes this code a little specific to this algorithm. The store is even more algorithm-specific as it first checks a predicate first before copying. In the general case, we should be able to use bulk stores.

I think the introduction of group-level loads and stores to SYCL may help us standardize this a bit when those are introduced.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -577,6 +580,103 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
+    auto
+    operator()(const _Policy& __policy, _InRng&& __in_rng, _OutRng&& __out_rng, ::std::size_t __n, _InitType __init,
+               _BinaryOperation __bin_op, _UnaryOp __unary_op)
+    {
+        using _ValueType = ::std::uint16_t;
+
+        // This type is used as a workaround for when an internal tuple is assigned to ::std::tuple, such as
+        // with zip_iterator
+        using __tuple_type = typename ::oneapi::dpl::__internal::__get_tuple_type<
+            typename ::std::decay_t<decltype(__in_rng[0])>, typename ::std::decay_t<decltype(__out_rng[0])>>::__type;
```
danhoeflinger, line 603: Despite the comment, I'm confused as to why this and the `static_cast` later on are necessary in this pattern.  

I think I understand `__get_tuple_type<>` generally, its so that we can convert our internal tuple to a `std::tuple`.  It seems to be used for the other users of `__parallel_scan_copy`, but I'm wondering if that is because of their internal use of zip iterator.  Is it still necessary here?  

If so, should we be using this elsewhere?

adamfidel: This issue is exposed in the `zip_iterator.pass` test, which calls `copy_if` with a `zip_iterator` input iterator and a `transform_iterator` output iterator. Without the cast, we are met with the following error:
```
[...]/include/oneapi/dpl/pstl/tuple_impl.h:261:31: error: reference to type 'const std::_Swallow_assign' could not bind to an lvalue of type 'const int'
    __value_holder(_Up&& t) : value(::std::forward<_Up>(t))
```
This is the same workaround used in the `__copy_by_mask` functor for the original `copy_if` algorithm. See the comment at https://github.com/oneapi-src/oneDPL/blob/6bb45531b7bd6908a16897ebd8fedbb18d365430/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L517-L535

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -577,6 +580,103 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
+                        _ONEDPL_PRAGMA_UNROLL
+                        for (::std::uint16_t __idx = __item_id; __idx < __n; __idx += _WGSize)
+                        {
+                            __lacc[__idx] = __unary_op(__in_rng[__idx]);
+                        }
+                    }
+
+                    __scan_work_group<_ValueType, /* _Inclusive */ false>(
+                        __group, __lacc.get_pointer(), __lacc.get_pointer() + __elems_per_wg,
+                        __lacc.get_pointer() + __elems_per_wg, __bin_op, __init);
```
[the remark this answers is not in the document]
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -573,6 +573,52 @@ class __future : private std::tuple<_Args...>
...
+// Note that the integers provided in the integer_sequence must be monotonically increasing
+template <typename>
+class __static_monotonic_dispatcher;
+
+template <::std::uint16_t _X, ::std::uint16_t... _Xs>
+class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X, _Xs...>>
+{
+    template <::std::uint16_t... _Vals>
+    using _Head = typename std::conditional_t<
+        sizeof...(_Vals) != 0, std::tuple_element<0, std::tuple<::std::integral_constant<::std::uint32_t, _Vals>...>>,
```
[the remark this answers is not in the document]
adamfidel: Thanks, done.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -625,6 +625,14 @@ __dpl_bit_floor(_T __x) noexcept
 #endif
 }
 
+// The max power of 2 not smaller than the given value, same as C++20 std::bit_ceil
+template <typename _T>
+::std::enable_if_t<::std::is_integral_v<_T>&& ::std::is_unsigned_v<_T>, _T>
```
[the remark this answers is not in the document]
adamfidel: Good catch. I fixed it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Okay, thanks. I think we can leave it as is then. This is probably something more broad to evaluate on a wider scale in the future.

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -56,6 +56,14 @@ using __value_t = decltype(oneapi::dpl::__internal::get_value_type<_R>(0));
 template <typename _Proj, typename _R>
 using __key_t = ::std::remove_cv_t<::std::remove_reference_t<::std::invoke_result_t<_Proj&, __value_t<_R>>>>;
 
+template <class T, class = decltype(std::declval<T&>().begin())>
+std::true_type has_begin_impl(int);
+template <class T>
+std::false_type has_begin_impl(...);
```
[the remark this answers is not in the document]
danhoeflinger: I see this has been refactored and is no longer relevant.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Thanks for the additional information.  I think it is fine here in the context of this PR.

I'll make an issue to discuss this `__get_tuple_type<>` workaround in general.  I think it is there to enable a usage pattern which should be discouraged (and can be rewritten to not require the workaround). Also, there are other algorithm(s?) for which a similar usage pattern will cause the same compilation errors that don't currently use this workaround.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
timmiesmith: Waiting on #874 before merging this shouldn't be done. We may need to revisit this to update this algorithm though.

julianmi: I agree. #874 is currently blocked by a compiler issue. We will need to evaluate its outcome and investigate where and how USM host memory is beneficial.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I believe this was discussed in the original single-work-group scan PR:
https://github.com/oneapi-src/oneDPL/pull/783#discussion_r1113201765

I would assume the same applies here in the `copy_if` case.

danhoeflinger (approved):
This LGTM.  It seems like @MikeDvorskiy may still have some open unresolved questions though, so it would be good to get his OK on those topics before merging.

adamfidel:
> @MikeDvorskiy Am I right that we are introducing two new patterns to our SYCL backend - parallel_scan_copy and parallel_copy_if ?

Yes, this introduces a specific pattern for `__parallel_copy_if`, which is distinct from `__parallel_scan_copy`. With scan_copy, there are higher-order functions that are passed around to implement algorithms such as `unique_copy`. Ideally, we could have added the single-work-group specialization to `__parallel_scan_copy` but it turned out to be difficult to support all of the algorithms that relied on the higher-order function capabilities, so we instead split the pattern into two: one specifically for `copy_if` and the original scan-copy pattern.

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -56,6 +56,17 @@ using __value_t = decltype(oneapi::dpl::__internal::get_value_type<_R>(0));
 template <typename _Proj, typename _R>
 using __key_t = ::std::remove_cv_t<::std::remove_reference_t<::std::invoke_result_t<_Proj&, __value_t<_R>>>>;
 
+template <class T, class = void>
+struct range_has_raw_ptr_iterator : ::std::false_type
```
MikeDvorskiy, line 60: uglification

MikeDvorskiy (approved):
LGTM with a minor comment

include/oneapi/dpl/pstl/utils_ranges.h
adamfidel: Could you please clarify how you would like this line to be formatted? Should I move the inheritance to a new line?

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -56,6 +56,17 @@ using __value_t = decltype(oneapi::dpl::__internal::get_value_type<_R>(0));
...
+{
+};
+
+template <class T>
+struct range_has_raw_ptr_iterator<T, ::std::void_t<decltype(::std::declval<T&>().begin())>>
```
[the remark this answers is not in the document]
adamfidel: I just applied the suggested change from clang-format. Does that match what you were expecting to see?

include/oneapi/dpl/pstl/utils_ranges.h
akukanov: @adamfidel, uglification means adding undescores in front of all the names, like this:
```
template <class _T, class = void>
struct __range_has_raw_ptr_iterator : ::std::false_type
```

Note also that we typically use `typename` instead of `class` for template parameters.

adamfidel: Thanks for the clarification @akukanov. I updated the names to have double underscores and fixed the class/typename issue.

@MikeDvorskiy would you mind re-approving?

danhoeflinger (approved):
LGTM,  I checked over the uglification changes, and tried to confirm that the merge didn't spoil anything (as well as my previous approval).

PR #915, opened 2023-04-28
Disable checked iterators on Windows for tests which required onedpl backend

SergeyKopienko:
The goal of this PR is to avoid compile errors when we use checked iterators from MS STL library inside Kernel's code.
Previously we have compile errors in this situation like this:
```
In file included from test\xpu_api\numerics\numeric.ops\reduce\xpu_reduce.pass.cpp:17:
In file included from include\oneapi/dpl/numeric:21:
In file included from C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.34.31933\include\numeric:11:
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.34.31933\include\xutility(868,5): error: SYCL kernel cannot call a variadic function
    _STL_VERIFY(_First <= _Last, "transposed pointer range");
    ^
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.34.31933\include\yvals.h(190,13): note: expanded from macro '_STL_VERIFY'
            _STL_REPORT_ERROR(mesg);                                                       \
            ^
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.34.31933\include\yvals.h(181,9): note: expanded from macro '_STL_REPORT_ERROR'
        _RPTF0(_CRT_ASSERT, mesg);               \
        ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.22000.0\ucrt\crtdbg.h(764,37): note: expanded from macro '_RPTF0'
    #define _RPTF0(rptno, msg)      _RPT_BASE(rptno, __FILE__, __LINE__, NULL, "%s", msg)
                                    ^
C:\Program Files (x86)\Windows Kits\10\include\10.0.22000.0\ucrt\crtdbg.h(751,48): note: expanded from macro '_RPT_BASE'
        (void) ((1 != _CrtDbgReport(__VA_ARGS__)) || \
                                               ^
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.34.31933\include\xutility(891,9): note: called by '_Adl_verify_range<const int *, const int *>'
        _Verify_range(_First, _Last);
        ^
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.34.31933\include\numeric(73,5): note: called by 'reduce<const int *, int, std::plus<>>'
    _Adl_verify_range(_First, _Last);
    ^
C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.34.31933\include\numeric(102,17): note: called by 'reduce<const int *>'
    return _STD reduce(_First, _Last, _Iter_value_t<_InIt>{}, plus{});
                ^
test\xpu_api\numerics\numeric.ops\reduce\xpu_reduce.pass.cpp(54,39): note: called by 'operator()'
                        out[0] = dpl::reduce(Iter(&in[0]), Iter(&in[0]));
```

How we are doing this:
- define ```_ITERATOR_DEBUG_LEVEL``` as ```0```;
- use the multithread-specific and DLL-specific version of the run-time library: ```/MD```

When:
- on Windows for debug mode builds of out tests.

For which tests:
- for all tests which required **dpcpp** backend.

At this moment it's all tests from the next folders:
- test/xpu_api/algorithms;
- test/xpu_api/functional;
- test/xpu_api/numerics.

All details about ```_ITERATOR_DEBUG_LEVEL``` described at https://learn.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170

We already tried to solve this task in https://github.com/oneapi-src/oneDPL/pull/910, but as result - we had crashes on Windows in debug mode builds of tests.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -59,7 +59,10 @@ add_custom_target(run-onedpl-tests
 
 macro(onedpl_add_test test_source_file switch_off_checked_iterators)
 
-    string(TOLOWER "${CMAKE_BUILD_TYPE}" _build_type_in_lower)
+    set(_build_type_in_lower "unknown")
+    if (CMAKE_BUILD_TYPE)
+        string(TOLOWER "${CMAKE_BUILD_TYPE}" _build_type_in_lower)
+    endif()
 
     get_filename_component(_test_name ${test_source_file} NAME)
     string(REPLACE "\.cpp" "" _test_name ${_test_name})
@@ -75,11 +78,9 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
     #    https://learn.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170
     #    https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170
     #    https://stackoverflow.com/questions/51494506/replacing-md-with-mt-in-cmake-is-not-possible-in-release-mode
-    if (WIN32 AND DEFINED _build_type_in_lower AND ${_build_type_in_lower} STREQUAL "debug")
-        if (${switch_off_checked_iterators})
-            target_compile_definitions(${_test_name} PRIVATE _ITERATOR_DEBUG_LEVEL=0)
-            target_compile_options(${_test_name} PRIVATE "/MD$<$<CONFIG:Debug>:>")
-        endif()
+    if (WIN32 AND ${_build_type_in_lower} STREQUAL "debug" AND ${switch_off_checked_iterators})
+        target_compile_definitions(${_test_name} PRIVATE _ITERATOR_DEBUG_LEVEL=0)
+        target_compile_options(${_test_name} PRIVATE "/MD$<$<CONFIG:Debug>:>")
     endif()
 
     # oneDPL test harness may initialize a C++ iterator using iterator with different type
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -57,7 +57,10 @@ add_custom_target(run-onedpl-tests
     DEPENDS build-onedpl-tests
     COMMENT "Build and run all oneDPL tests")
 
-macro(onedpl_add_test test_source_file)
+macro(onedpl_add_test test_source_file switch_off_checked_iterators)
+
+    string(TOLOWER "${CMAKE_BUILD_TYPE}" _build_type_in_lower)
+
     get_filename_component(_test_name ${test_source_file} NAME)
     string(REPLACE "\.cpp" "" _test_name ${_test_name})
 
@@ -65,6 +68,17 @@ macro(onedpl_add_test test_source_file)
     target_compile_definitions(${_test_name} PRIVATE _PSTL_TEST_SUCCESSFUL_KEYWORD=1)
     if (MSVC)
         target_compile_options(${_test_name} PRIVATE /bigobj)
+    endif()
+
+    # Disable checked iterators on Windows
+    # For details please see https://learn.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170
+    #                        https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170
+    #                        https://stackoverflow.com/questions/51494506/replacing-md-with-mt-in-cmake-is-not-possible-in-release-mode
+    if (WIN32 AND ${_build_type_in_lower} STREQUAL "debug")
+        if (${switch_off_checked_iterators})
+            target_compile_definitions(${_test_name} PRIVATE _ITERATOR_DEBUG_LEVEL=0)
+            target_compile_options(${_test_name} PRIVATE "/MD$<$<CONFIG:Debug>:>")
+        endif()
     endif()
 
     # oneDPL test harness may initialize a C++ iterator using iterator with different type
@@ -79,7 +93,6 @@ macro(onedpl_add_test test_source_file)
 
     add_test(NAME ${_test_name} COMMAND ${CMAKE_CURRENT_BINARY_DIR}/${_test_name})
 
-    string(TOLOWER "${CMAKE_BUILD_TYPE}" _build_type_in_lower)
     if (DEFINED ${_test_name}_timeout_${_build_type_in_lower})
         message(STATUS
         "Timeout for ${_test_name} is set to ${${_test_name}_timeout_${_build_type_in_lower}}, "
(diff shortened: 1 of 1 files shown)


akukanov:
I think this patch is correct, but let somebody with more knowledge of CMake to take a look.

test/CMakeLists.txt
```diff
@@ -67,6 +70,18 @@ macro(onedpl_add_test test_source_file)
         target_compile_options(${_test_name} PRIVATE /bigobj)
     endif()
 
+    # Disable checked iterators on Windows for debug mode
+    # For details please see
+    #    https://learn.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170
+    #    https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170
+    #    https://stackoverflow.com/questions/51494506/replacing-md-with-mt-in-cmake-is-not-possible-in-release-mode
+    if (WIN32 AND ${_build_type_in_lower} STREQUAL "debug")
```
adamfidel, line 78: This breaks support for calling CMake without a build type specified, such as:
```bash
$ cmake -G "Ninja" -DCMAKE_CXX_COMPILER=icx -DONEDPL_BACKEND=dpcpp ..
```
This behavior is currently support, but with this patch now results in the following error:
```
CMake Error at test/CMakeLists.txt:78 (if):
  if given arguments:

    "WIN32" "AND" "STREQUAL" "debug"

  Unknown arguments specified
```

SergeyKopienko: How to reproduce this error?

test/CMakeLists.txt
```diff
@@ -67,6 +70,18 @@ macro(onedpl_add_test test_source_file)
...
+        if (${switch_off_checked_iterators})
+            target_compile_definitions(${_test_name} PRIVATE _ITERATOR_DEBUG_LEVEL=0)
+            target_compile_options(${_test_name} PRIVATE "/MD$<$<CONFIG:Debug>:>")
```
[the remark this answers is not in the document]
SergeyKopienko: I didn't find any other working ways to set this up for a specific application.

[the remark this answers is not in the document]
danhoeflinger: Somewhat relevant to this discussion, but not a solution:
https://github.com/oneapi-src/oneDPL/pull/851#issuecomment-1536369021
in #851, we encountered a separate issue with icpx (GNU-style intel driver) using CMake on Windows which we cannot work around.  
If this is a blocker on icpx, this is more reason to push people to use icx[-cl] on Windows.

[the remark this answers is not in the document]
danhoeflinger: I'm not saying that we shouldn't try to support GNU-style compilers (and it may be relevant for non-intel compilers also), but yes, this is perhaps less of an issue assuming that proposal goes through.

test/CMakeLists.txt
```diff
@@ -67,6 +70,18 @@ macro(onedpl_add_test test_source_file)
         target_compile_options(${_test_name} PRIVATE /bigobj)
     endif()
 
+    # Disable checked iterators on Windows for debug mode
+    # For details please see
+    #    https://learn.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170
+    #    https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170
+    #    https://stackoverflow.com/questions/51494506/replacing-md-with-mt-in-cmake-is-not-possible-in-release-mode
+    if (WIN32 AND DEFINED ${_build_type_in_lower} AND ${_build_type_in_lower} STREQUAL "debug")
+        if (${switch_off_checked_iterators})
```
adamfidel, line 79: Just a small suggestion: the condition for `${switch_off_checked_iterators}` can be included in the first `if`. That way, you can have a single `if` statement instead of nested statements.

test/CMakeLists.txt
```diff
@@ -67,6 +70,18 @@ macro(onedpl_add_test test_source_file)
         target_compile_options(${_test_name} PRIVATE /bigobj)
     endif()
 
+    # Disable checked iterators on Windows for debug mode
+    # For details please see
+    #    https://learn.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170
+    #    https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170
+    #    https://stackoverflow.com/questions/51494506/replacing-md-with-mt-in-cmake-is-not-possible-in-release-mode
+    if (WIN32 AND DEFINED ${_build_type_in_lower} AND ${_build_type_in_lower} STREQUAL "debug")
```
adamfidel, line 78: I tried testing this change and I believe this condition is not correct. As written, you are not checking if the variable `_build_type_in_lower` exists, but rather if there exists a variable whose name is provided by the value of `${_build_type_in_lower}` (e.g., a variable with the name `debug` or `release`).

This condition will always be false and so the workaround is never actually applied.

I considered changing it to the following, but this doesn't make much sense as the variable `_build_type_in_lower ` is always defined since it was explicitly defined at the beginning of the macro:

```suggestion
    if (WIN32 AND DEFINED _build_type_in_lower AND ${_build_type_in_lower} STREQUAL "debug")
```

Instead, I would suggest to define `_build_type_in_lower` as a known string that is not debug, and only redefine it if `CMAKE_BUILD_TYPE` is not falsey. Such as:

```diff
diff --git a/test/CMakeLists.txt b/test/CMakeLists.txt
index 5b266617..bfb4cfb9 100644
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -59,7 +59,10 @@ add_custom_target(run-onedpl-tests

 macro(onedpl_add_test test_source_file switch_off_checked_iterators)

-    string(TOLOWER "${CMAKE_BUILD_TYPE}" _build_type_in_lower)
+    set(_build_type_in_lower "unknown")
+    if (CMAKE_BUILD_TYPE)
+        string(TOLOWER "${CMAKE_BUILD_TYPE}" _build_type_in_lower)
+    endif()

     get_filename_component(_test_name ${test_source_file} NAME)
     string(REPLACE "\.cpp" "" _test_name ${_test_name})
@@ -75,11 +78,9 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
     #    https://learn.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-170
     #    https://learn.microsoft.com/en-us/cpp/build/reference/md-mt-ld-use-run-time-library?view=msvc-170
     #    https://stackoverflow.com/questions/51494506/replacing-md-with-mt-in-cmake-is-not-possible-in-release-mode
-    if (WIN32 AND DEFINED ${_build_type_in_lower} AND ${_build_type_in_lower} STREQUAL "debug")
-        if (${switch_off_checked_iterators})
-            target_compile_definitions(${_test_name} PRIVATE _ITERATOR_DEBUG_LEVEL=0)
-            target_compile_options(${_test_name} PRIVATE "/MD$<$<CONFIG:Debug>:>")
-        endif()
+    if (WIN32 AND ${_build_type_in_lower} STREQUAL "debug" AND ${switch_off_checked_iterators})
+        target_compile_definitions(${_test_name} PRIVATE _ITERATOR_DEBUG_LEVEL=0)
+        target_compile_options(${_test_name} PRIVATE "/MD$<$<CONFIG:Debug>:>")
     endif()

     # oneDPL test harness may initialize a C++ iterator using iterator with different type
```

I have tested the following configurations and I'm able to compile the `xpu_accumulate.pass.exe` target for all of them on Windows:
```
cmake -G "Ninja" -DCMAKE_CXX_COMPILER=icx -DONEDPL_BACKEND=dpcpp -DCMAKE_BUILD_TYPE=Debug ..
cmake -G "Ninja" -DCMAKE_CXX_COMPILER=icx -DONEDPL_BACKEND=dpcpp -DCMAKE_BUILD_TYPE="" ..
cmake -G "Ninja" -DCMAKE_CXX_COMPILER=icx -DONEDPL_BACKEND=dpcpp ..
```

SergeyKopienko: Thank you, @adamfidel. Good point. Fixed.

test/CMakeLists.txt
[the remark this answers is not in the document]
adamfidel: I just tested this patch with icpx.exe, which is a GNU-style compiler and it works fine. It seems like it will work with and without ` -D_ITERATOR_DEBUG_LEVEL` defined and the compiler just ignores the `/MD` flag. So it's okay from that perspective.

adamfidel (approved):
I did some testing and I believe this PR is working as expected. LGTM.

SergeyKopienko:
Thank you, @adamfidel !

PR #859, opened 2023-03-24
[oneDPL][host][hetero] + sort_by_key API

MikeDvorskiy:
[oneDPL][host][hetero] + sort_by_key API

PR #916, opened 2023-04-28
Allow sort to use non-const comparison functor

adamfidel:
The PR #859 accidentally introduced a change that required the comparison functor to provide a const `operator()`. This PR relaxes that requirement by changing the introduced lambda to be `mutable`.


danhoeflinger:
I think one of the clang format suggestions is relevant here.  Other than that it LGTM.

timmiesmith (approved):
LGTM

danhoeflinger (approved):
LGTM

PR #703, opened 2022-11-30
[oneDPL] + projection support for radix_sort pattern, ranges::sort API

MikeDvorskiy:
[oneDPL] + projection support for radix_sort pattern, ranges::sort API.

We use oneapi::dpl::__internal::__no_op, because std::identity is available from C++20.

PR #917, opened 2023-05-01
Adding value_type to replicate_start_view_simple

danhoeflinger:
This adds the missing `value_type` to `replicate_start_view_simple` that became a requirement with changed introduced in #703.


timmiesmith (approved):
LGTM.

timmiesmith:
@danhoeflinger should the clang-format issue be addressed?

danhoeflinger:
I fixed one clang format issue, one still remains from this file, but it is not related to this change.  I think it was intentionally left from a previous change.  (plug for #715 which would show this as a green check)

PR #918, opened 2023-05-02
[KT] Improvements to sort signed and floating point types

akukanov:
There are also general improvements here and there.

### DIFF
--- a/test/parallel_api/experimental/radix_sort_esimd.pass.cpp
+++ b/test/parallel_api/experimental/radix_sort_esimd.pass.cpp
@@ -194,7 +194,7 @@ int main()
             test_general_cases<uint32_t>(size);
             test_general_cases<int>(size);
             test_general_cases<float>(size);
-            test_general_cases<double>(size);
+            // test_general_cases<double>(size);
         }
         test_small_sizes();
     }
--- a/test/parallel_api/experimental/radix_sort_esimd.pass.cpp
+++ b/test/parallel_api/experimental/radix_sort_esimd.pass.cpp
@@ -121,7 +121,7 @@ void test_usm(std::size_t size)
 
     std::cout << "\nInput (size=" << size << "):" << std::endl;
     for( int i = 0; i < std::min(size, out_limit); ++i )
-        std::cout << ref[i] << std::endl;
+        std::cout << std::hex << ref[i] << std::endl;
 
     q.copy(ref, input, size).wait();
     std::sort(ref, ref + size);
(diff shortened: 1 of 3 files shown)


test/parallel_api/experimental/radix_sort_esimd.pass.cpp
```diff
@@ -192,9 +192,9 @@ int main()
         for(auto size: sizes)
         {
             test_general_cases<uint32_t>(size);
-            // test_general_cases<int>(size);
-            // test_general_cases<float>(size);
-            // test_general_cases<double>(size);
+            test_general_cases<int>(size);
+            test_general_cases<float>(size);
+            test_general_cases<double>(size);
```
dmitriy-sobolev, line 197: Let's comment it out, because the current implementation does not support `double` type. For example, it uses [gather](https://intel.github.io/llvm-docs/doxygen/group__sycl__esimd__memory.html#gaf7753dcd28985606893834239883c6d5) instruction, which has the following restriction:
> Tx	Element type, must be of size 4 or less.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
```diff
@@ -119,13 +130,17 @@ scan(sycl::ext::intel::esimd::simd<T, 16> src)
...
+template <::std::uint16_t __radix_mask, typename _T, int _N>
+sycl::ext::intel::esimd::simd<::std::uint16_t, _N>
 __get_bucket(sycl::ext::intel::esimd::simd<_T, _N> __value, ::std::uint32_t __radix_offset)
 {
-    return (__value >> __radix_offset) & sycl::ext::intel::esimd::simd<_T, _N>(__radix_mask);
+    return sycl::ext::intel::esimd::simd<::std::uint16_t, _N>(__value >> __radix_offset) & __radix_mask;
 }
 
+template <typename T, bool __is_ascending>
+inline constexpr T __sort_neutral =
```
dmitriy-sobolev, line 141: I like the name of the value given its connection to abstract algebra. You can also consider `__sort_identity`, because `identity` seems to be used more frequently, e.g. it is used in SYCL specification.

test/parallel_api/experimental/radix_sort_esimd.pass.cpp
akukanov: That's a pretty severe limitation; we need to understand how we address it later.

dmitriy-sobolev (dismissed):

akukanov:
I split generally useful improvements into a couple of separate PRs.

akukanov:
All the useful changes were moved to other branches and partially integrated.

PR #919, opened 2023-05-03
Support creating a view over an entire range with take_view_simple

mmichel11:
Currently, take_view_simple can only create a view up to the first `n - 1` elements of the range. We are extending this to support creating a view over all `n` elements which is consistent with the behavior of `std::ranges::take_view` in C++20.

This PR was identified by and fixes a bug in `reduce_by_segment` where the modified assertion fails when each segment is of size `1`.

### DIFF
--- a/include/oneapi/dpl/pstl/utils_ranges.h
+++ b/include/oneapi/dpl/pstl/utils_ranges.h
@@ -281,7 +281,7 @@ struct take_view_simple
     _R __r;
     _Size __n;
 
-    take_view_simple(_R __rng, _Size __size) : __r(__rng), __n(__size) { assert(__n >= 0 && __n < __r.size()); }
+    take_view_simple(_R __rng, _Size __size) : __r(__rng), __n(__size) { assert(__n >= 0 && __n <= __r.size()); }
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
--- a/include/oneapi/dpl/pstl/utils_ranges.h
+++ b/include/oneapi/dpl/pstl/utils_ranges.h
@@ -281,7 +281,10 @@ struct take_view_simple
     _R __r;
     _Size __n;
 
-    take_view_simple(_R __rng, _Size __size) : __r(__rng), __n(__size) { assert(__n >= 0 && __n <= __r.size()); }
+    take_view_simple(_R __rng, _Size __size) : __r(__rng), __n(::std::min(__size, _Size(__r.size())))
+    {
+        assert(__n >= 0);
+    }
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
@@ -318,7 +321,10 @@ struct drop_view_simple
     _R __r;
     _Size __n;
 
-    drop_view_simple(_R __rng, _Size __size) : __r(__rng), __n(__size) { assert(__n >= 0 && __n < __r.size()); }
+    drop_view_simple(_R __rng, _Size __size) : __r(__rng), __n(::std::min(__size, _Size(__r.size())))
+    {
+        assert(__n >= 0);
+    }
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
--- a/include/oneapi/dpl/pstl/utils_ranges.h
+++ b/include/oneapi/dpl/pstl/utils_ranges.h
@@ -272,7 +272,8 @@ struct reverse_view_simple
     }
 };
 
-//It is kind of pseudo-view for take_view support.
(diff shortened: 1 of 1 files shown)


danhoeflinger:
This look like a good fix.  Two things which came up in our offline conversation which I want to raise here:

1.  We should also fix `drop_view_simple` as it suffers from the same bug.
2.  @mmichel11 brought up that the c++20 `std::ranges::take_view` allows `n > __r.size()`, and just appropriately returns the size of the view as `min(__n, __r.size())`, as seen in the example https://en.cppreference.com/w/cpp/ranges/take_view.
  a. @akukanov Should we take this opportunity to align with the C++20 standard on this and allow `n > __r.size()` for `take_view_simple`
  b. If so, I think the same should apply for `drop_view_simple`, returning size of `max(0,__r.size() - __n)`

danhoeflinger:
> > Should we take this opportunity to align with the C++20 standard on this and allow n > __r.size() for take_view_simple?
> 
> In principle, why not - but it might depend on the amount of efforts/changes needed.

@mmichel11  can weigh in, but it seems like just a simple change to this assert and the `size` function for each of these two views.  
Any cases we are changing the behavior of would've been hitting the assert previously, so I don't see it as risky.

mmichel11:
> > > Should we take this opportunity to align with the C++20 standard on this and allow n > __r.size() for take_view_simple?
> > 
> > 
> > In principle, why not - but it might depend on the amount of efforts/changes needed.
> 
> @mmichel11 can weigh in, but it seems like just a simple change to this assert and the `size` function for each of these two views. Any cases we are changing the behavior of would've been hitting the assert previously, so I don't see it as risky to break any existing functionality.

Yes, this change to better align with C++20 is straightforward, so I agree that it is best to do this. I will push this change for `take_view_simple` and `drop_view_simple`.

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -281,7 +281,10 @@ struct take_view_simple
     _R __r;
     _Size __n;
 
-    take_view_simple(_R __rng, _Size __size) : __r(__rng), __n(__size) { assert(__n >= 0 && __n < __r.size()); }
+    take_view_simple(_R __rng, _Size __size) : __r(__rng), __n(::std::min(__size, _Size(__r.size())))
+    {
+        assert(__n >= 0);
+    }
```
danhoeflinger, line 287: Handling it like this makes `size()` faster (not recalculating `std::min` for every `size()` call), but loses a bit of information as `__n` is public and may not match the input `__size` when it exceeds the input range.  

I still think this is correct, but I'm mentioning it here.  Perhaps it is worth a comment above `__n` in case anyone is expecting different semantics (like `__n` to be always equal to the input `__size`).

akukanov: `__n` is public in the C++ sense but not supposed to be used directly, i.e. it's an undocumented implementation detail.

However, I am not sure if such an implementation is correct. We assume that the size of the underlying range is fixed, i.e. does not change during the lifetime of the view. But is it a correct assumption?

mmichel11: @akukanov I have taken a look and you are correct. If the underlying range is modified (ex: gets smaller), then it should still be safe. I will make the changes to ensure this.

mmichel11:
Clang format is failing due to a formatting difference that exists outside of these proposed changes.

include/oneapi/dpl/pstl/utils_ranges.h
akukanov: The question is, however, whether the underlying range **can** get bigger or smaller. As far as I understand, these views are for internal use and are intentionally simple, avoiding overheads where generality is not needed. So maybe the assumption about the size is correct, and all that we need to do is document it in the code.

danhoeflinger: If this is the direction we would want to take,  I think I may prefer to go back to limiting it via assert to be within `0<=n<=size`.  This seems a safer way to limit generality in favor of performance.

mmichel11: I agree with both points. I think just modifying the initial assert for correctness in `take_view_simple` and `drop_view_simple` will be the best way to go.

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -272,7 +272,8 @@ struct reverse_view_simple
     }
 };
 
-//It is kind of pseudo-view for take_view support.
+//It is kind of pseudo-view for take_view support. We assume that the underlying range will not shrink
+//after creation of the view to favor performance.
```
danhoeflinger, line 276: The comments need to be reverted as well as they are no longer the approach / assumption.

danhoeflinger (approved):
LGTM

PR #922, opened 2023-05-04
Update release notes for oneDPL 2022.1.1

ValentinaKats:
Signed-off-by: Valentina Kats valentina.kats@intel.com

### DIFF
--- a/documentation/release_notes.rst
+++ b/documentation/release_notes.rst
@@ -18,7 +18,7 @@ Fixes Issues
 ------------
 - Fixed an error that caused segmentation faults in ``transform_reduce``, ``minmax_element``, and related algorithms when ran on CPU devices. 
 - Fixed a compilation error in ``transform_reduce``, ``minmax_element``, and related algorithms on FPGAs.
-- Fixed an issue with oneDPL permutation iterator. 
+- Fixed permutation_iterator to support C-style array as a permutation map.
 - Fixed a radix-sort issue with 64-bit signed integer types.
 
 New in 2022.1.0
--- a/documentation/release_notes.rst
+++ b/documentation/release_notes.rst
@@ -13,15 +13,13 @@ New in 2022.1.1
 New Features
 ------------
 - Improved ``sort`` algorithm performance for the arithmetic data types with ``std::less`` or ``std::greater`` comparison operator and DPC++ policy.
-- AOT compilation for CPU devices is now possible with the use of ``ONEDPL_USE_AOT_COMPILATION`` cmake option.
 
 Fixes Issues
 ------------
-- Fixed ``Device name missing`` error occurring with the following cmake options: ``ONEDPL_USE_AOT_COMPILATION=ON``, ``ONEDPL_DEVICE_BACKEND=GPU`` or not specified and ``CMAKE_BUILD_TYPE`` is not specified.
 - Fixed an error that caused segmentation faults in ``transform_reduce``, ``minmax_element``, and related algorithms when ran on CPU devices. 
 - Fixed a compilation error in ``transform_reduce``, ``minmax_element``, and related algorithms on FPGAs.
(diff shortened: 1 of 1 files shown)


documentation/release_notes.rst
```diff
@@ -8,6 +8,21 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
 and provides high-productivity APIs aimed to minimize programming efforts of C++ developers
 creating efficient heterogeneous applications.
 
+New in 2022.1.1
+===============
+New Features
+------------
+- Improved ``sort`` algorithm performance for the arithmetic data types with ``std::less`` or ``std::greater`` comparison operator and DPC++ policy.
+- AOT compilation for CPU devices is now possible with the use of ``ONEDPL_USE_AOT_COMPILATION`` cmake option.
```
akukanov, line 16: As far as I remember, it was supposed to only affect the tests, and be named appropriately.
If we plan to change the name, maybe it makes sense to postpone this note until we make that change. If we do not plan to change the name, at the very least the note should clarify the applicability scope.

documentation/release_notes.rst
```diff
@@ -8,6 +8,21 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
...
+
+Fixes Issues
+------------
+- Fixed ``Device name missing`` error occurring with the following cmake options: ``ONEDPL_USE_AOT_COMPILATION=ON``, ``ONEDPL_DEVICE_BACKEND=GPU`` or not specified and ``CMAKE_BUILD_TYPE`` is not specified.
```
akukanov, line 20: Because of the options, that also only affects the tests, right? If so, is it important enough to mention? If the impact is more than just our tests, we need to explain it more generally, not referring to Cmake macros.

documentation/release_notes.rst
```diff
@@ -8,6 +8,21 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
...
+- Fixed an error that caused segmentation faults in ``transform_reduce``, ``minmax_element``, and related algorithms when ran on CPU devices. 
+- Fixed a compilation error in ``transform_reduce``, ``minmax_element``, and related algorithms on FPGAs.
+- Fixed an issue with oneDPL permutation iterator. 
+- Fixed an issue with 64-bit integer types with DPC++ policy.
```
akukanov, line 24: How have these issues impacted (maybe potentially) user code? That should be clear enough so that users could correlate to their problems. If no impact on user code, why is it worth mentioning?

documentation/release_notes.rst
dmitriy-sobolev: Currently, this macro affects external targets as well. However, it might be better to use it only for the test only (prons: it is not related to oneDPL; cons: still, it provides some shortcut for external targets) and thus do not document it in release notes.

dmitriy-sobolev: I suggest omitting the notes about `ONEDPL_USE_AOT_COMPILATION` cmake option.

ValentinaKats: Omitted the notes about ONEDPL_USE_AOT_COMPILATION cmake option as was suggested.

documentation/release_notes.rst
ValentinaKats: Omitted the notes about ONEDPL_USE_AOT_COMPILATION cmake option as was suggested.

documentation/release_notes.rst
akukanov: For (1), I suggest "Fixed permutation_iterator to support C-style array as a permutation map"

ValentinaKats: Added the suggested.

akukanov (approved):

PR #926, opened 2023-05-09
[Dynamic Selection] Make internal implementation of concurrent queue generic

AnuyaWelling2801:
Used a std::deque instead of std::queue.


vossmjp (approved):
looks good  to me.

PR #927, opened 2023-05-09
[Dynamic Selection] fix wrap around for next context in round robin policy

AnuyaWelling2801:


include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy_impl.h
```diff
@@ -66,7 +66,11 @@ namespace experimental{
 
     template<typename ...Args>
     selection_handle_t select(Args&&...) {
-      auto i = next_context_++ % num_contexts_;
+      size_t i=0;
+      if(next_context_==std::numeric_limits<universe_container_size_t>::max())
```
vossmjp, line 70: This logic doesn't seem right to me. You are typically incrementing next_context_ and then using % to map to one of the contexts. Once next_context_ reaches the max possible value based on its numeric type the if condition is true, and instead of computing an i, you set i to 0. However, you do not modify next_context_. Won't it just get stuck here then and stop round-robin-ing?

vossmjp (changes requested):
Logic looks wrong to me.

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy_impl.h
AnuyaWelling2801: Fixed.

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy_impl.h
```diff
@@ -66,7 +66,13 @@ namespace experimental{
 
     template<typename ...Args>
     selection_handle_t select(Args&&...) {
-      auto i = next_context_++ % num_contexts_;
+      size_t i=0;
+      if(next_context_==std::numeric_limits<universe_container_size_t>::max()){
+          next_context_=0;
```
vossmjp, line 71: Should "next_context_=0" actually be "next_context_ = next_context_%num_contexts_ + 1"?  In the unlikely case that you actually overflow, you'd could skip some devices by always going back to the 1st device, even if you were part way through the list when you overflowed.

vossmjp: I was so focused on the update logic, I didn't notice that the reset logic is not thread safe. You need to add thread safety to select.  When it was simple increment of an atomic, things were fine, but now that the logic is more complicated, it's not.

AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy_impl.h
```diff
@@ -66,7 +66,20 @@ namespace experimental{
...
+      universe_container_size_t current_context_;
+      while(true){
+          if(current_context_ == std::numeric_limits<universe_container_size_t>::max()){
+            universe_container_size_t new_context_ = (current_context_%num_contexts_)+1;
+            if(next_context_.compare_exchange_weak(current_context_, new_context_)){
+                i = new_context_;
+                break;
+            }
+          }else{
+              i = next_context_.fetch_add(1)%num_contexts_;
```
vossmjp, line 79: still need to do compare_exchange in the else to be sure that value has not changed since the if condition was checked. Otherwise, by the time you reach this line it could now be max and so the wrap-around is necessary.

AnuyaWelling2801: Fixed.

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy_impl.h
```diff
@@ -77,7 +77,22 @@ namespace experimental{
 
     template<typename ...Args>
     selection_handle_t select(Args&&...) {
-      auto i = unit_->next_context_++ % unit_->num_contexts_;
+      size_t i=0;
+      while(true){
+          universe_container_size_t current_context_ = unit_->next_context_.load();
```
vossmjp, line 82: Looks fine now. Not a huge fan of unit_ as the name but its an implementation detail, so it's ok. Eventually, unit_ should be private though. But that might wait to refactoring with new names, etc.

vossmjp (approved):
Looks ok to me now.

PR #931, opened 2023-05-11
Clarification of tests list for switching off checking iterator

SergeyKopienko:
In this PR we make some clarifications of test list for Windows build for which required to switch off checking iterator too:

The list of tests:
- test\general\header_order_ranges_1.pass.cpp
- test\parallel_api\algorithm\alg.sorting\alg.min.max\min_init_list.pass.cpp
- test\parallel_api\algorithm\alg.sorting\alg.min.max\min_init_list_comp.pass.cpp
- test\parallel_api\algorithm\alg.sorting\alg.min.max\max_init_list_comp.pass.cpp
- test\parallel_api\algorithm\alg.sorting\alg.min.max\max_init_list.pass.cpp
- test\parallel_api\ranges\transform_ranges_factory_sycl.pass.cpp
- test\parallel_api\ranges\fill_generate_factory.pass.cpp
- test\parallel_api\ranges\transform_ranges_sycl.pass.cpp
- test\parallel_api\ranges\move_ranges_sycl.pass.cpp
- test\parallel_api\ranges\copy_ranges_factory_sycl.pass.cpp
- test\parallel_api\ranges\copy_ranges_sycl.pass.cpp
- test\parallel_api\ranges\search_n_ranges_sycl.pass.cpp
- test\parallel_api\ranges\replace_copy_ranges_sycl.pass.cpp
- test\parallel_api\ranges\transform2_ranges_factory_sycl.pass.cpp
- test\parallel_api\ranges\transform2_ranges_sycl.pass.cpp
- test\xpu_api\random\device_tests\lognormal_distribution_device_test.pass.cpp
- test\xpu_api\random\device_tests\normal_distribution_device_test.pass.cpp
- test\xpu_api\random\interface_tests\distributions_methods.pass.cpp
- test\xpu_api\random\statistics_tests\lognormal_distribution_test.pass.cpp
- test\xpu_api\random\statistics_tests\normal_distribution_test.pass.cpp


test/CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
 endmacro()
 
 set(_regexp_dpcpp_backend_required "(xpu_api/algorithms|xpu_api/functional|xpu_api/numerics)")
+set(_regexp_switch_off_checked_it  "(xpu_api/algorithms|xpu_api/functional|xpu_api/numerics|test/general/header_order_ranges|test/parallel_api/algorithm/alg.sorting/alg.min.max|test/parallel_api/ranges|test/xpu_api/random/device_tests|test/xpu_api/random/interface_tests|test/xpu_api/random/statistics_tests)")
```
adamfidel, line 147: I don't think you need to include `xpu_api/algorithms|xpu_api/functional|xpu_api/numerics` in the regex, since all of those tests are covered by the statement `if (_file MATCHES "${_regexp_dpcpp_backend_required}")` and so they will never reach the `else` to be compared against this regex.

SergeyKopienko: Good point. Fixed.

danhoeflinger:
The regex will also turn off checked iterator for many tests not mentioned in your description.  Is the intention to tightly scope the regex, or is it OK to turn off checked iterators in some places where it may not be necessary?

Extras from the first two parts of the regex as compared to the list in the description:

- test\general\header_order_ranges_0.pass.cpp (this should likely also need it if its sibling does)
- test/parallel_api/algorithm/alg.sorting/alg.min.max/max_comp.pass.cpp
- test/parallel_api/algorithm/alg.sorting/alg.min.max/max.pass.cpp
- test/parallel_api/algorithm/alg.sorting/alg.min.max/min_comp.pass.cpp
- test/parallel_api/algorithm/alg.sorting/alg.min.max/min.pass.cpp
- test/parallel_api/algorithm/alg.sorting/alg.min.max/minmax_element.pass.cpp

Many more are included from `test/parallel_api/ranges` as well, I did not check the rest.

SergeyKopienko:
> The regex will also turn off checked iterator for many tests not mentioned in your description. Is the intention to tightly scope the regex, or is it OK to turn off checked iterators in some places where it may not be necessary?

@From my point of view - it's ok: usually we still separate this check state by folders.
@akukanov what do you think ?

test/CMakeLists.txt
```diff
@@ -144,6 +144,7 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
 endmacro()
 
 set(_regexp_dpcpp_backend_required "(xpu_api/algorithms|xpu_api/functional|xpu_api/numerics)")
+set(_regexp_switch_off_checked_it  "(test/general/header_order_ranges|test/parallel_api/algorithm/alg.sorting/alg.min.max|test/parallel_api/ranges|test/xpu_api/random/device_tests|test/xpu_api/random/interface_tests|test/xpu_api/random/statistics_tests)")
```
adamfidel, line 147: I have found that the `transform_scan.pass` test also requires turning off checked iterators. It will compile fine without the flag to disable checked iterators, but it will crash at runtime.

```suggestion
set(_regexp_switch_off_checked_it  "(test/general/header_order_ranges|test/parallel_api/algorithm/alg.sorting/alg.min.max|test/parallel_api/ranges|test/xpu_api/random/device_tests|test/xpu_api/random/interface_tests|test/xpu_api/random/statistics_tests|test/parallel_api/numeric/numeric.ops/transform_scan)")
```

adamfidel (approved):
@danhoeflinger I believe it is okay to disable checked iterators for some tests that may not explicitly require it being disabled.

Seeing that this PR allows us to have more passing tests for Windows debug targets that weren't passing before, I think that this PR is good to be merged as is.

danhoeflinger:
> @danhoeflinger I believe it is okay to disable checked iterators for some tests that may not explicitly require it being disabled.

Yes, this works for me

PR #932, opened 2023-05-11
Update software.intel.com URLs

dcbenito:
Updated all URLs with the old schema software.intel.com and updated some URLs that were resulting in a 404.


ValentinaKats (approved):
LGTM. Thank you!

PR #934, opened 2023-05-12
Remove unused things + mem management

SergeyKopienko:
In this PR we remove from radix sort some unused things:
- dynamic_job_queue_t;
- job_queue;
- p_lookup buffer;
- use std::shared_ptr for sycl::free call.


include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
```diff
@@ -611,20 +591,19 @@ onesweep(_ExecutionPolicy&& __exec, _Range&& __rng, ::std::size_t __n)
...
+
+    ::std::shared_ptr<uint8_t> tmp_buffer(sycl::malloc_device<uint8_t>(temp_buffer_size, __exec.queue()),
+                                          [__exec](uint8_t* ptr) { sycl::free(ptr, __exec.queue()); });
+    auto p_tmp_buffer = tmp_buffer.get();
+    auto e_init = __exec.queue().memset(p_tmp_buffer, 0, temp_buffer_size);
+
+    auto p_global_offset = reinterpret_cast<uint32_t*>(p_tmp_buffer);
+    auto p_sync_buffer = reinterpret_cast<uint32_t*>(p_tmp_buffer + GLOBAL_OFFSET_SIZE);
+
+    ::std::shared_ptr<uint32_t> p_output(sycl::malloc_device<uint32_t>(__n, __exec.queue()),
```
dmitriy-sobolev, line 604: I think it is better to leave raw pointers and free calls, because:
- There is no difficult memory management; we create and destroy these object in the same function
- It does not really improve readability
- It incurs overhead

dmitriy-sobolev (approved):

PR #939, opened 2023-05-15
Updating CMake minimum required versions

danhoeflinger:
Increasing minimum required CMake version:

- Windows: 3.20
  - Required for baseline CMake support for `icx`, `icx-cl` to allow workarounds to work 
- Linux: 3.11
  - Required for CMake support of `icpc` with required C++17 feature 

Adding in minimum CMake requirements for oneDPL via package inclusion (previously none).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,13 @@
 #
 ##===----------------------------------------------------------------------===##
 
-cmake_minimum_required(VERSION 3.4.3)
+if (CMAKE_HOST_WIN32)
+    # Requires version 3.20 for baseline support of icx, icx-cl
+    cmake_minimum_required(VERSION 3.20)
+else()
+    # Requires version 3.11 for use of icpc with c++17 requirement
+    cmake_minimum_required(VERSION 3.11)
+endif()
 
 # oneDPL is a subproject only when PARENT_DIRECTORY is defined
 get_directory_property(_onedpl_is_subproject PARENT_DIRECTORY)
--- a/cmake/oneDPLWindowsIntelLLVMConfig.cmake
+++ b/cmake/oneDPLWindowsIntelLLVMConfig.cmake
@@ -12,7 +12,7 @@
 #
 ##===----------------------------------------------------------------------===##
 
-if(CMAKE_HOST_WIN32)
+if (CMAKE_HOST_WIN32)
     if (CMAKE_VERSION VERSION_LESS 3.20)
         set(REASON_FAILURE "oneDPLWindowsIntelLLVM requires CMake 3.20 or later on Windows.")
         set(oneDPLWindowsIntelLLVM_FOUND FALSE)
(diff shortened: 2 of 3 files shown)


dmitriy-sobolev (dismissed):

CMakeLists.txt
```diff
@@ -12,7 +12,13 @@
 #
 ##===----------------------------------------------------------------------===##
 
-cmake_minimum_required(VERSION 3.4.3)
+if (CMAKE_HOST_WIN32)
```
adamfidel, line 15: I've noticed that we use the variables `CMAKE_HOST_WIN32` or `WIN32` pretty much interchangeably. Do you think we should converge on a single one?

cmake/oneDPLWindowsIntelLLVMConfig.cmake
```diff
@@ -11,7 +11,17 @@
 # See https://llvm.org/LICENSE.txt for license information.
 #
 ##===----------------------------------------------------------------------===##
-cmake_minimum_required(VERSION 3.15.0)
+
+if(CMAKE_HOST_WIN32)
```
adamfidel, line 15: I would prefer if there was a space after the `if` to be consistent with the other `if` statements.

CMakeLists.txt
danhoeflinger: They have different meanings, `CMAKE_HOST_WIN32` is for the host system, where `WIN32` is the for the target (cross compilation).  

I think for these purposes, it is the host system that we are interested in, because it is about getting things to properly compile using the Windows versions of CMake and the intel compilers or Linux versions of the same. 

Taking a quick look around the rest of the CMake files:

- I just see one usage of `WIN32` in the `test/CMakeLists`, with the recent work on checked iterators, and I think that is appropriate.  
- A few others in examples which I think are OK. 
- Finally there is one in the generation script for the CMake config.  I think that one is also appropriate as `WIN32`.  

To me they seem appropriate but I may be mistaken somewhere.

cmake/oneDPLWindowsIntelLLVMConfig.cmake
danhoeflinger: done, thanks.

danhoeflinger:
> A general question: do we know if all of our Windows CI systems meet the minimum version requirement?

This is a good question.  I believe so, but @timmiesmith perhaps you can confirm or pull in the right person to confirm?

danhoeflinger:
> This is a good question. I believe so, but @timmiesmith perhaps you can confirm or pull in the right person to confirm?

After an offline discussion, I believe that this is the case, but I am running the testing on this branch to confirm.

timmiesmith:
> > This is a good question. I believe so, but @timmiesmith perhaps you can confirm or pull in the right person to confirm?
> 
> After an offline discussion, I believe that this is the case, but I am running the testing on this branch to confirm.

It looks like our testing on Windows picks up the Cmake provided by Visual Studio, or something explicitly installed as part of our environment.  In either case it will meet the minimum we're defining now, yes.

danhoeflinger:
> It looks like our testing on Windows picks up the Cmake provided by Visual Studio, or something explicitly installed as part of our environment. In either case it will meet the minimum we're defining now, yes.

I've looked through the testing results, and nothing makes me think that this will be introducing new problems.  

@dmitriy-sobolev, @adamfidel: if one of you are willing to approve, I will merge this in.

PR #941, opened 2023-05-15
Add checks for char8_t library support in numeric limits tests

mmichel11:
For the C++20 `char8_t` type, we only check for presence of the `char8_t` feature macro in our `std::numeric_limits` tests. However, it is necessary to also check for the presence of library support for `char8_t` to ensure that the `std::numeric_limits` specializations are implemented.

### DIFF
--- a/test/support/test_macros.h
+++ b/test/support/test_macros.h
@@ -370,14 +370,8 @@ inline void DoNotOptimize(Tp const& value) {
 #  define TEST_HAS_NO_LOCALIZATION
 #endif
 
-#if TEST_STD_VER <= 17 || !defined(__cpp_char8_t)
+#if TEST_STD_VER <= 17 || !defined(__cpp_char8_t) || !defined(__cpp_lib_char8_t)
 #  define TEST_HAS_NO_CHAR8_T
-#endif
-
-#if TEST_STD_VER > 17 && defined(__cpp_char8_t) && defined(__cpp_lib_char8_t)
-#   define _ENABLE_CHAR8_T 1
-#else
-#   define _ENABLE_CHAR8_T 0
 #endif
 
 #if defined(_LIBCPP_HAS_NO_THREADS)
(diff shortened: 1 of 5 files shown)


danhoeflinger:
I notice in `test/support/test_macros.h` we defined `TEST_HAS_NO_CHAR8_T` with the negation of these conditions (but it goes unused).  

Does it make sense to update that define (and perhaps rename it `TEST_HAS_NO_CHAR8_LIB_SUPPORT_T`), then use this in place of the repeated checks?

danhoeflinger:
> As for me better to add this into some new define in the file test/support/test_config.h For example may be used `_ENABLE_RANGES_TESTING`

@SergeyKopienko Is there a reason to add this to `test/support/test_config.h` rather than augmenting and using the the currently unused `TEST_HAS_NO_CHAR8_T` in `test/support/test_macros.h`?

SergeyKopienko:
I think better to define ```TEST_STD_VER > 17 && defined(__cpp_char8_t) && defined(__cpp_lib_char8_t)``` as one macro in one place and use it.

mmichel11:
@danhoeflinger @SergeyKopienko Thanks for the feedback. I will define this macro as suggested and add it to the affected tests.

mmichel11:
> I think better to define `TEST_STD_VER > 17 && defined(__cpp_char8_t) && defined(__cpp_lib_char8_t)` as one macro in one place and use it.

I added the `_ENABLE_CHAR8_T` macro. I actually put it into `test_macros.h` since `TEST_STD_VER` may not be defined if its put into `test_config.h`.

mmichel11:
> > I think better to define `TEST_STD_VER > 17 && defined(__cpp_char8_t) && defined(__cpp_lib_char8_t)` as one macro in one place and use it.
> 
> I added the `_ENABLE_CHAR8_T` macro. I actually put it into `test_macros.h` since `TEST_STD_VER` may not be defined if its put into `test_config.h`.

After an offline conversation with @danhoeflinger, I decided it would be best to update the `TEST_HAS_NO_CHAR8_T` macro instead. This way is consistent with how we check for the presence of other types in the affected tests.

SergeyKopienko (approved):
LGTM

julianmi (approved):
LGTM

danhoeflinger (approved):
LGTM

PR #943, opened 2023-05-16
Define missed variables in CMakeLists

dmitriy-sobolev:
CMakeLists relies on `INTEL_LLVM_COMPILER` and `INTEL_LLVM_COMPILER_VERSION` variables. Currently, they are not defined unless `ONEDPL_TEST_WIN_ICX_FIXES` is on.

A side effect of the change is that when `ONEDPL_TEST_WIN_ICX_FIXES` is on and this `CMakeLists` is used, these variables are set again inside `oneDPLWindowsIntelLLVMConfig.cmake` and have the same values. This does not change CMake behavior.

I additionally corrected a message for `ONEDPL_USE_AOT_COMPILATION`, because AOT compilation does not depend on OCLOC compiler (only) since #794.


danhoeflinger (approved):
LGTM

adamfidel (approved):
LGTM.

PR #944, opened 2023-05-16
Ensure absence of overflow due to unintentional use of 32-bit arithmetic

akukanov:
Address another potential issue found by static code analysis: intermediate computations are done in 32 bit while the result is then assigned to 64-bit variables


danhoeflinger (approved):
LGTM

PR #381, opened 2021-10-19
[pstl] LLVM sync In: Initial implementation of OpenMP backend

MikeDvorskiy:
1) pstl] LLVM sync In: Initial implementation of OpenMP backend
LLVM Phabricator Review:
https://reviews.llvm.org/D99836
A couple of parallel patterns still remains serial - "Parallel partial sort", and "Parallel transform scan" - there are //TODOs in the code.

2) building and CI stuff for OpenMP backend

PR #945, opened 2023-05-17
Add the library version header, and update oneDPL version to 2022.7.0

akukanov:
This patch introduces the special `<oneapi/dpl/version>` header for the library version and feature macros.

- It's included in `onedpl_config.h`, and so its definitions are added into and included with each library public header.
- It includes the standard `<version>` header

Please suggest other feature macros which users might have reasons to check.


include/oneapi/dpl/version
```diff
@@ -0,0 +1,34 @@
+// -*- C++ -*-
+//===----------------------------------------------------------------------===//
+//
+// Copyright (C) Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef __ONEDPL_version_H
```
SergeyKopienko, line 10: May be ```_ONEDPL_VERSION_H``` ?

include/oneapi/dpl/version
```diff
@@ -0,0 +1,34 @@
...
+#    define _PSTL_VERSION 14000
+#    define _PSTL_VERSION_MAJOR (_PSTL_VERSION / 1000)
+#    define _PSTL_VERSION_MINOR ((_PSTL_VERSION % 1000) / 10)
+#    define _PSTL_VERSION_PATCH (_PSTL_VERSION % 10)
+#endif
+
+// Feature macros
+#define ONEDPL_HAS_RANDOM_NUMBERS 202210L
+
+#endif // __ONEDPL_version_H
```
SergeyKopienko, line 34: the same.

include/oneapi/dpl/version
```diff
@@ -0,0 +1,34 @@
...
+// Copyright (C) Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef __ONEDPL_version_H
+#define __ONEDPL_version_H
+
+#include "oneapi/dpl/internal/common_config.h"
```
rarutyun, line 13: This is a logic change. onedpl_config had never have includes. Now it includes common_config.h and furthermore, `<version>`

I am not sure if it's critical or not. Need to double check

test/general/version.pass.cpp
```diff
@@ -15,14 +15,10 @@
...
-#include _PSTL_TEST_HEADER(execution)
-
-#include <oneapi/dpl/pstl/onedpl_config.h>
-
-#include "support/utils.h"
+#include _PSTL_TEST_HEADER(version)
 
 #if !__has_include(<execution>)
-static_assert(_PSTL_VERSION == 11000, "");
+static_assert(_PSTL_VERSION == 14000, "");
```
rarutyun, line 21: Just curious, didn't we really have a sync?

rarutyun (commented):
Other than that looks good to me

include/oneapi/dpl/version
akukanov: As of today, the config file already includes `<version>`, so I guess it's not an issue.

test/general/version.pass.cpp
akukanov: We need to double-check which PSTL upstream version we are currently synced with.

documentation/library_guide/macros.rst
```diff
@@ -34,8 +34,14 @@ Use these macros to test presence of specific |onedpl_short| functionality.
 ================================== ===============================================
 Macro                              Macro values and the functionality
 ================================== ===============================================
-``ONEDPL_HAS_RANDOM_NUMBERS``      ``202210L`` for pseudo-random number generators and distributions,
-                                   as defined in |onedpl_short| version 2022.2 or later
+``ONEDPL_HAS_RANDOM_NUMBERS``      Pseudo-random number generators and distributions.
+
+                                   * ``202409L`` - added support of comparison and I/O stream operators and an experimental Philox engine
```
danhoeflinger, line 39: Should we be listing previous values with there descriptions here as well, or just the current one?
`202210L`

test/general/version.pass.cpp
danhoeflinger: As far as I can tell, this is the last sync #381 (plus some minor cleanup right after).  
I was not involved in these sync efforts though so its possible I'm not great at searching for this.

danhoeflinger (dismissed):
LGTM
Edit:  documentation question answered.

documentation/library_guide/macros.rst
akukanov: We will be listing all values, but this is the first one, as  `202210L` was never released.

danhoeflinger: Aha whoops, you are right.  I mistook a delta commit, and thought perhaps it existed in some form prior.  
Overall LGTM.

include/oneapi/dpl/version
```diff
@@ -0,0 +1,42 @@
...
+#ifndef _PSTL_VERSION
+#    define _PSTL_VERSION 14000
+#    define _PSTL_VERSION_MAJOR (_PSTL_VERSION / 1000)
+#    define _PSTL_VERSION_MINOR ((_PSTL_VERSION % 1000) / 10)
+#    define _PSTL_VERSION_PATCH (_PSTL_VERSION % 10)
+#endif
+
+// Feature macros
+#define ONEDPL_HAS_RANDOM_NUMBERS         202409L
+#if _ONEDPL_CPP20_RANGES_PRESENT
```
rarutyun, line 38: I was thinking about this logic and I think it's correct. We may say that our feature test macro is available independently of the used compiler but then we return `true` for clang less then 16 and the users see the compile-time errors that they (and we) don't understand.

include/oneapi/dpl/version
```diff
@@ -0,0 +1,42 @@
...
+
+#ifndef _PSTL_VERSION
+#    define _PSTL_VERSION 14000
+#    define _PSTL_VERSION_MAJOR (_PSTL_VERSION / 1000)
+#    define _PSTL_VERSION_MINOR ((_PSTL_VERSION % 1000) / 10)
+#    define _PSTL_VERSION_PATCH (_PSTL_VERSION % 10)
+#endif
+
+// Feature macros
+#define ONEDPL_HAS_RANDOM_NUMBERS         202409L
```
rarutyun, line 37: I think I would preferer to change the name schema. To me `HAS` implies `bool`, while we also provide the version of a feature. So, I would name them like
- `ONEDPL_RANDOM_NUMBER_GENERATION`
- `ONEDPL_RANGE_ALGORITHMS`

I can see with some might prefer to use `_HAS_` in the name. It makes a perfect sense for the
`#if ONEDPL_HAS_RANGE_ALGORITHMS` expression.

But for `#if ONEDPL_HAS_RANGE_ALGORITHMS == 202409L` it doesn't look good in my opinion.

test/general/version.pass.cpp
```diff
@@ -15,23 +15,21 @@
...
 static_assert(_PSTL_VERSION_MINOR == 00);
 static_assert(_PSTL_VERSION_PATCH == 0);
 #endif
 
 static_assert(ONEDPL_VERSION_MAJOR == 2022);
-static_assert(ONEDPL_VERSION_MINOR == 6);
+static_assert(ONEDPL_VERSION_MINOR == 7);
 static_assert(ONEDPL_VERSION_PATCH == 0);
 
+#include "support/utils.h"
```
rarutyun, line 31: What's the point to move it below `static_assert`s? Is it because we want to make sure that we `<version>` header has everything we need?

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -16,21 +16,8 @@
...
-
-#define ONEDPL_VERSION_MAJOR 2022
-#define ONEDPL_VERSION_MINOR 6
-#define ONEDPL_VERSION_PATCH 0
-
-#if _ONEDPL___cplusplus >= 202002L && __has_include(<version>)
-#    include <version>
-#    define _ONEDPL_STD_FEATURE_MACROS_PRESENT 1
-#endif
+#include "../version"
```
rarutyun, line 19: To me this line looks like the indication that we want to move `onedpl_config.h` one level higher in the directory hierarchy. Of course, I don't propose to do it in this PR but having it within `pstl` folder confuses me.

rarutyun (commented):
Overall changes look good to me. They should not bring any troubles. The only thing that I suggest to reconsider is the naming schema for feature test macros themselves. If it's only my preference, so be it.

rarutyun:
Just remembered one more question. What about `ONEDPL_HISTOGRAM_ALGORITHM` feature test macro?

include/oneapi/dpl/pstl/onedpl_config.h
SergeyKopienko, line 19:
```suggestion
#include <oneapi/dpl/version>
```

danhoeflinger: I think our preference is to use relative includes from within oneDPL.

include/oneapi/dpl/version
danhoeflinger: This is a good point.  I agree with the name change away from `_HAS_`.

include/oneapi/dpl/pstl/onedpl_config.h
SergeyKopienko: We have about 100 includes inside oneDPL code (excepting tests) like I propose.

danhoeflinger:
> Just remembered one more question. What about `ONEDPL_HISTOGRAM_ALGORITHM` feature test macro?
@rarutyun 
Alexey and I discussed this.  I think it is unnecessary after the decision we made to just treat it as a bugfix and rearrange the template parameters.  
There shouldn't be a situation where a user would want to check for the presence of this macro and make one action or another in their code.  Our guidance for users should just be to remove explicit template specification if they encounter issues, and if they are relying on it to force a conversion, to replace that with an explicit conversion.  This will work with both before 2022.7.0 and after.

include/oneapi/dpl/pstl/onedpl_config.h
danhoeflinger: This is just my understanding of our convention.  I think we do both within the code.  It seems the norm in this directory is relative paths.  The norm one level up is what you suggest.

test/general/version.pass.cpp
akukanov: Exactly.
It is actually a good practice when writing tests to include any utility headers as late as possible, to minimize chances of artificially introducing extra code (incl. standard headers) that might interfere with the tested API.

include/oneapi/dpl/pstl/onedpl_config.h
akukanov: One level upper are only public headers, and the config header does not belong there.

Actually, including the new version header into the config is just the easiest way to ensure two things:
- The version information and feature macros are available to all headers that may need those;
- The content of the standard `<version>` header is available for the config.
It can be done in a different way, for example, by splitting our version header into the public and internal ones.

akukanov: We have about 100 includes starting from "oneapi/dpl/", about 100 includes starting from "../", and more than 200 other includes not starting from either.

Is there some real motivation why in this very case having the "full" path would be preferred?

include/oneapi/dpl/version
akukanov: We have discussed that last year and the schema was agreed back then. It's not ideal but we come to that as a satisfactory compromise, and I am not going to repeat that discussion. My personal preference, for example, was to only use capitals for ONEDPL and the rest to be in lower case, but that was not supported, and even though I still do not like it, it will be used.

A particular reason for `_HAS_` is that it's good for all the feature macros to have a common prefix of suffix, to ease searching those if needed (including by users), and this one is concise and meaningful.

include/oneapi/dpl/pstl/onedpl_config.h
SergeyKopienko: My assumption the full file name here is better due the header file name `version` is the same like in standard library.

akukanov: Switched to use an internal header here.

include/oneapi/dpl/version
akukanov: I can agree on the extra `_VER` suffix if that helps: `#if ONEDPL_HAS_RANGE_ALGORITHMS_VER == 202409L`. Sounds weirder to me but fine.

danhoeflinger: I see the point about having some shared prefix which isn't merely `ONEDPL_`, and it is hard to think of a better "small" prefix than `_HAS_`... 

Could `_VER_` or `_VERS_` be an alternative small prefix, without a postfix?
I don't have a strong preference between HAS and VER, but my preference would be against a postfix and prefix together just to keep names of reasonable length.

akukanov: `_FEAT_`? :)

include/oneapi/dpl/internal/version_impl.h
```diff
@@ -0,0 +1,40 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _ONEDPL_VERSION_IMPL_H
+#define _ONEDPL_VERSION_IMPL_H
+
+#if _ONEDPL___cplusplus >= 202002L && __has_include(<version>)
+#    include <version> // The standard C++20 header
+#    define _ONEDPL_STD_FEATURE_MACROS_PRESENT 1
```
SergeyKopienko, line 15: In the file https://github.com/oneapi-src/oneDPL/blob/0414309eccb641e5e250af215e904cedead9f2f8/include/oneapi/dpl/pstl/onedpl_config.h#L32
we have the code:
```C++
#if _ONEDPL___cplusplus >= 202002L && __has_include(<version>)
#    include <version>
#    define _ONEDPL_STD_FEATURE_MACROS_PRESENT 1
#endif
```

Should we add `#else` branch there?

SergeyKopienko: Or may be better to use `_ONEDPL_STD_FEATURE_MACROS_PRESENT` in tests ?

akukanov: I would leave the tests as they are, at least until there is a need to explicitly define the macro to 0.

SergeyKopienko: ok

include/oneapi/dpl/internal/version_impl.h
```diff
@@ -0,0 +1,40 @@
...
+
+#ifndef _PSTL_VERSION
+#    define _PSTL_VERSION 14000
+#    define _PSTL_VERSION_MAJOR (_PSTL_VERSION / 1000)
+#    define _PSTL_VERSION_MINOR ((_PSTL_VERSION % 1000) / 10)
+#    define _PSTL_VERSION_PATCH (_PSTL_VERSION % 10)
+#endif
+
+// Feature macros
+#define ONEDPL_HAS_RANDOM_NUMBERS         202409L
```
SergeyKopienko, line 35: I didn't see the use of `ONEDPL_HAS_RANDOM_NUMBERS` in the code.
Is it ok?

SergeyKopienko: Should we have some test for check `ONEDPL_HAS_RANDOM_NUMBERS` state?

akukanov: This is a good question. I will think what would appropriate for testing.

akukanov: Added a few static assertions.

danhoeflinger (approved):
From my perspective this looks good. 
A few discussions still open, but I don't have any issues with the PR as is.

include/oneapi/dpl/version
danhoeflinger: After this discussion, my preference is probably this order:
`ONEDPL_VER_XXX`
`ONEDPL_FEAT_XXX`
`ONEDPL_HAS_XXX`
`ONEDPL_XXX`
`ONEDPL_HAS_XXX_VER`

But, really any work for me.

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -22,6 +22,8 @@
 
 #if _ENABLE_STD_RANGES_TESTING
 
+static_assert(ONEDPL_HAS_RANGE_ALGORITHMS >= 202409L);
```
SergeyKopienko, line 25: Why this `static_assert` has placed under `#if _ENABLE_STD_RANGES_TESTING` check?

test/xpu_api/random/interface_tests/distributions_methods.pass.cpp
```diff
@@ -20,10 +20,12 @@
 #include "support/utils.h"
 
 #if TEST_UNNAMED_LAMBDAS
+#include <oneapi/dpl/random>
+static_assert(ONEDPL_HAS_RANDOM_NUMBERS >= 202409L);
```
SergeyKopienko, line 24: Why this `static_assert` has placed under `#if TEST_UNNAMED_LAMBDAS` check?

cmake/scripts/generate_config.cmake
```diff
@@ -34,7 +34,7 @@ endif()
 
 set(ONEDPL_ROOT "${CMAKE_CURRENT_LIST_DIR}/../..")
 
-file(READ ${ONEDPL_ROOT}/include/oneapi/dpl/pstl/onedpl_config.h _onedpl_version_info LIMIT 1024)
+file(READ ${ONEDPL_ROOT}/include/oneapi/dpl/internal/version_impl.h _onedpl_version_info LIMIT 1024)
```
SergeyKopienko, line 37: It seems to me this constant '1024' similar to some tricks that rely on the actual content of the `include/oneapi/dpl/internal/version_impl.h` file and it may breaks if this file will edited later.

test/xpu_api/random/interface_tests/distributions_methods.pass.cpp
SergeyKopienko: Probably we should check `ONEDPL_HAS_RANDOM_NUMBERS` and `ONEDPL_HAS_RANGE_ALGORITHMS` in tests by the same way like our customers will check it in the code.

In this case we will see is it user-friendly way or no.

test/parallel_api/ranges/std_ranges_test.h
akukanov: Because the feature is available and the feature macro is defined only under similar conditions.

test/xpu_api/random/interface_tests/distributions_methods.pass.cpp
akukanov: > Why this static_assert has placed under #if TEST_UNNAMED_LAMBDAS check?

Actually, it is placed after the header. Since it is a library defined macro, you need to include some header first.

The customers will likely use it mainly as a condition in `#if` preprocessor directives.
I think `static_assert` is good enough as it will fail if either the macro is undefined (in which case its name will not be recognized as a known token and compilation will fail) or defined but not into a proper value (in which case the asserted condition will not be met).

test/parallel_api/ranges/std_ranges_test.h
SergeyKopienko: ok

test/xpu_api/random/interface_tests/distributions_methods.pass.cpp
SergeyKopienko: ok

cmake/scripts/generate_config.cmake
akukanov: https://cmake.org/cmake/help/latest/command/file.html
```
file(READ <filename> <variable> [OFFSET <offset>] [LIMIT <max-in>] [HEX])
Read content from a file called <filename> and store it in a <variable>.
Optionally start from the given <offset> and read at most <max-in> bytes.
```

Currently the needed lines fit into the first 1024 bytes. The new file is ~10x smaller than `onedpl_config.h`, currently ~1.4K. So maybe the limit has now less value, but I would let others to decide.

SergeyKopienko: @dmitriy-sobolev, @danhoeflinger what do you think about this?

danhoeflinger: I think it is fine as is.  At most we could add a comment in version_impl.h saying it should not grow beyond 1024 bytes without changes in this file, but I think were probably OK without it.

dmitriy-sobolev: I have no strong opinion on that. You can follow Dan's suggestion, or remove LIMIT. I do not expect that the file will significantly inflate (I suppose LIMIT was added to speed up CMake configuration).

SergeyKopienko: OK. So let's save as is.

SergeyKopienko (dismissed):
LGTM

cmake/scripts/generate_config.cmake
SergeyKopienko: Just for note: we reads this file in two places:
- in the `scripts/generate_config.cmake` file;
- in the `CMakeLists.txt` file.
I mean if we decide to remove `LIMIT` usage, required to remove it in both places.

akukanov: I moved the library version macros to the top of the file.

timmiesmith (approved):
I agree with Alexey on the macro naming approach.  The rest of the PR looks good to me. Reapplying previous approvals from Sergey and Dan that were dismissed to address final coments.

SergeyKopienko (commented):
LGTM

PR #950, opened 2023-05-18
[Tests] Fix UB in the tests (may happen with multiple TUs)

dmitriy-sobolev:
C++ specification says:
> If an inline function or variable (since C++17) with external linkage is defined differently in different translation units, the behavior is undefined.

`get_test_queue()` uses `my_queue`, which is linked internally and thus has different addresses in each TU.

### DIFF
--- a/test/support/utils_sycl.h
+++ b/test/support/utils_sycl.h
@@ -134,7 +134,7 @@ make_new_policy(_Policy&& __policy)
 #endif     // ONEDPL_FPGA_DEVICE
 
 // create the queue with custom asynchronous exceptions handler
-static auto my_queue = sycl::queue(default_selector, async_handler);
+inline auto my_queue = sycl::queue(default_selector, async_handler);
 
 inline
 sycl::queue get_test_queue()
--- a/test/support/utils_sycl.h
+++ b/test/support/utils_sycl.h
@@ -134,7 +134,7 @@ make_new_policy(_Policy&& __policy)
 #endif     // ONEDPL_FPGA_DEVICE
 
 inline
-sycl::queue& get_test_queue()
+sycl::queue get_test_queue()
 {
     // create the queue with custom asynchronous exceptions handler
     static sycl::queue my_queue(default_selector, async_handler);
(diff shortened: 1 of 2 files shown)


akukanov:
Is it a problem that each TU has its own copy of `my_queue`? As far as I understand, static variables do not have external linkage, so the rules are not violated.

dmitriy-sobolev:
> This problem absent in our code from point of view test organization structure: in our `CMakeLists.txt` file each test consist only from one TU.

Yes, the test harness currently creates executables with only one TU. So the issue does not appear. But why not fix the issue?

akukanov:
> To be more precise, the problem is with `get_test_queue()`, which uses `my_queue`. `get_test_queue()` must be identical in each TU, but `my_queue` has different addresses (each TU has its own copy of it). 

I get it, thanks. But then, why should we even have `my_queue` outside of the function scope? Make it the static variable in the function.

SergeyKopienko (dismissed):
LGTM

test/support/utils_sycl.h
```diff
@@ -133,12 +133,11 @@ make_new_policy(_Policy&& __policy)
 #    endif // ONEDPL_USE_PREDEFINED_POLICIES
 #endif     // ONEDPL_FPGA_DEVICE
 
-// create the queue with custom asynchronous exceptions handler
-static auto my_queue = sycl::queue(default_selector, async_handler);
-
 inline
-sycl::queue get_test_queue()
+sycl::queue& get_test_queue()
```
julianmi, line 137: I think the return type should be changed in `test/support/utils_invoke.h:103` as well (see failing tests).

SergeyKopienko: As for me - it's not required because sycl::queue has pimpl-implementation.

dmitriy-sobolev: There used to be a mismatch between the definition and a forward declaration in another header. It must have been fixed to allow the test to compile.

SergeyKopienko:
I believe the replacement from ```sycl::queue get_test_queue()``` to ```sycl::queue& get_test_queue()``` is not required.
And this change hasn't related to the original problem (from my point of view).

test/support/utils_sycl.h
dmitriy-sobolev: @SergeyKopienko, now I got what you suggested. The suggestion was about returning a copy in both declarations. It initially seemed to me as if it was fine to leave a copy at `test/support/utils_invoke.h:103` while having a reference here.

SergeyKopienko (approved):
LGTM

julianmi (approved):

Issue #949, opened 2023-05-17
Using zip_iterator -> transform_iterator with std::ignore to discard output


danhoeflinger:
Thanks for the additional information.  I think it is fine here in the context of this PR.
              I'll make an issue to discuss this `__get_tuple_type<>` workaround in general.  I think it is there to enable a usage pattern which should be discouraged (and can be rewritten to not require the workaround). Also, there are other algorithm(s?) for which a similar usage pattern will cause the same compilation errors that don't currently use this workaround.

_Originally posted by @danhoeflinger in https://github.com/oneapi-src/oneDPL/pull/909#discussion_r1197037914_

danhoeflinger:
The workaround `__get_tuple_type<>`  from `include/oneapi/dpl/pstl/tuple_impl.h` is used in a few places:  
- `__partition_by_mask`
- `__copy_by_mask`
- `__global_scan_functor` (which is used by `__pattern_transform_scan_multi_group`,  but not `__pattern_transform_scan_single_group`)
- `__parallel_copy_if` in #909

It is meant to cover the case where the input type to the high level algorithm is a `zip_iterator`, and the output type is an iterator pointing of `std::tuple`.  This is shown in the `zip_iterator.pass` test case: 
https://github.com/oneapi-src/oneDPL/blob/2f70dac1d2718af37832c63562c8ae1186fd1c34/test/parallel_api/iterator/zip_iterator.pass.cpp#L736-L748

It seems to me that this usage pattern should be discouraged and rewritten with `oneapi::dpl::discard_iterator()` instead, which does not require the workaround.  However, removing the workaround could break existing code which employs this sort of pattern.  
 
**I believe we need to add the `__get_tuple_type<>` workaround to the single group scan, as that may have worked for users previously, and it should have the same functionality as multi-group scan.** 

I was able to break compilation of `oneapi::dpl::copy` with a similar usage pattern as it does not use this workaround internally. 
 There are likely other affected APIs.  I question whether we need to expand the use of the workaround to other APIs like `oneapi::dpl::copy`, as they could be achieved successfully with `discard_iterator`.  I'm interested to hear other opinions here though.

danhoeflinger:
#951 has been approved and merged.  Leaving this open for now in case there is feedback / recommendations for further action.

danhoeflinger:
Summary for users who may be experiencing this problem
---
[...]

PR #951, opened 2023-05-18
Add zip_iterator with discard_iterator example

danhoeflinger:
Adds a test case to `zip_iterator.pass.cpp` using `oneapi::dpl::zip_iterator`  and `oneapi::dpl::discard_iterator`.  Provided as an example of an simpler rewrite of the previous test case which uses `oneapi::dpl::transform_iterator` with `std::forward_as_tuple` and `std::ignore` to accomplish the same effect as described in #949.


SergeyKopienko (approved):
LGTM

timmiesmith (approved):
LGTM

timmiesmith (approved):
LGTM

PR #953, opened 2023-05-18
Correct radix sort's order preserving cast for descending sort

mmichel11:
Our radix sort implementation has been observed to produce incorrect results when performing a descending sort of `int16_t` numbers containing negative values.

The bitwise complement operation adheres to integral promotion rules. In `__order_preserving_cast` of our radix sort, we require a mask of the form 011..11 for the descending case. In order to ensure this, we need the right shift operator to pad with zeros which is only guaranteed when shifting unsigned types. We must add an additional cast to ensure that we are right-shifting an 
unsigned type. This change corrects the observed sorting bug.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
@@ -62,7 +62,8 @@ __order_preserving_cast(_Int __val)
 {
     using _UInt = ::std::make_unsigned_t<_Int>;
     // mask: 100..0 for ascending, 011..1 for descending
-    constexpr _UInt __mask = (__is_ascending) ? _UInt(1) << ::std::numeric_limits<_Int>::digits : _UInt(~_UInt(0)) >> 1;
+    constexpr _UInt __mask =
+        (__is_ascending) ? _UInt(1) << ::std::numeric_limits<_Int>::digits : ::std::numeric_limits<_UInt>::max() >> 1;
     return __val ^ __mask;
 }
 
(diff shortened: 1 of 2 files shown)


danhoeflinger:
Perhaps it makes sense to add an `int16_t` test to `sort.pass` to provide a regression test for this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -62,7 +62,7 @@ __order_preserving_cast(_Int __val)
 {
     using _UInt = ::std::make_unsigned_t<_Int>;
     // mask: 100..0 for ascending, 011..1 for descending
-    constexpr _UInt __mask = (__is_ascending) ? _UInt(1) << ::std::numeric_limits<_Int>::digits : ~_UInt(0) >> 1;
+    constexpr _UInt __mask = (__is_ascending) ? _UInt(1) << ::std::numeric_limits<_Int>::digits : _UInt(~_UInt(0)) >> 1;
```
akukanov, line 65: Maybe it's better to use `::std::numeric_limits<_Uint>::max`?

mmichel11: Yes, I agree this is a lot cleaner. I have pushed this change.

mmichel11:
> Perhaps it makes sense to add an `int16_t` test to `sort.pass` to provide a regression test for this.

I agree. I have added a test that covers this case: sorting `std::int16_t` with negative values and `std::greater<int16_t>` as the comparison functor.

danhoeflinger (approved):
LGTM

mmichel11:
It seems that this test file had some kernel name clashes in it which was recently fixed, so I rebased with main and adjusted my additions accordingly.

danhoeflinger (approved):

PR #956, opened 2023-05-22
[KT] Support for descending sort order

akukanov:
No implementation changes appear necessary, except for onesweep to use the order as designed.

In the test, I enable the possibility to test everything in descending order, off by default. It's similar to other template parameters - WG size, etc. We will need to find out how to test for different parameter values without сombinatorial explosion.


SergeyKopienko (approved):
LGTM

SergeyKopienko:
About "We will need to find out how to test for different parameter values without сombinatorial explosion."
This approach already implemented in the ```TestUtils``` namespace but for another entities. For example please take a look at https://github.com/oneapi-src/oneDPL/blob/main/test/support/utils_invoke.h

PR #958, opened 2023-05-22
Bugfix for CMake openMP setup

danhoeflinger:
Bugfix in CMake setup for `cmake/templates/oneDPLConfig.cmake.in`:
This PR uses the variable `_openmp_flag` to set up `CMAKE_REQUIRED_LIBRARIES`.  Previously, it was using an uninitialized variable `_openmp_option`.  It seems that this was just a typo.
Following similar structure in `CMakeLists.txt`, and following the intention as stated in the comment two lines above.

### DIFF
--- a/cmake/templates/oneDPLConfig.cmake.in
+++ b/cmake/templates/oneDPLConfig.cmake.in
@@ -64,7 +64,7 @@ if (EXISTS "${_onedpl_headers}")
 
             # Some compilers may fail if _openmp_flag is not in CMAKE_REQUIRED_LIBRARIES.
             set(_onedpl_saved_required_libs ${CMAKE_REQUIRED_LIBRARIES})
-            set(CMAKE_REQUIRED_LIBRARIES ${_openmp_option})
+            set(CMAKE_REQUIRED_LIBRARIES ${_openmp_flag})
             check_cxx_compiler_flag(${_openmp_flag} _openmp_option)
             set(CMAKE_REQUIRED_LIBRARIES ${_onedpl_saved_required_libs})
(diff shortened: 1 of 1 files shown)


dmitriy-sobolev (commented):
Do you know what is the issue?

I performed some source code archeology and was not able to find any description of the environment and the error. 
It is unclear to me if it is even relevant to keep that workaround because the typo has been here for more than 2 years. 

I suggest removing this workaround from both files (oneDPLConfig.cmake and CMakeLists.txt) after the upcoming release and thoroughly testing the result.

danhoeflinger:
> Do you know what is the issue?
> 
> I performed some source code archeology and was not able to find any description of the environment and the error. It is unclear to me if it is even relevant to keep that workaround because the typo has been here for more than 2 years.
> 
> I suggest removing this workaround from both files (oneDPLConfig.cmake and CMakeLists.txt) after the upcoming release and thoroughly testing the result.

Sure, I think this is the appropriate plan, I can make an issue to track this.  I do not know the compiler(s) which require this, but like you say, it must not be a common case as this bug has existed for a long time.  I'll see if I can find the source of the workaround.

I would say that this PR does improve the CMake code even if the workaround is no longer necessary.

danhoeflinger:
After an offline discussion, I think it is best to hold off merging this prior to the release, as we have no known error case and this has existed for 2 years without complaint.  More risk is introduced by changing this than leaving it until after the release.  We can pursue the removal of the workaround as discussed in #960 then.

danhoeflinger:
Closing in favor of #974.

PR #959, opened 2023-05-22
Fix cmake diagnostics of the used backend

dmitriy-sobolev:
This fixes an issue when cmake provides wrong information on the used backend. 

It happens when all the conditions are satisfied:
- The compiler supports `-fsycl` feature
- SYCL headers are not available
- `ONEDPL_BACKEND` is equal to `dpcpp` or `dpcpp_only` or not explicitly provided. 

In that case, cmake prints a message that DPCPP backend is going to be used, but oneDPL headers do not enable it, because SYCL headers are not available. 

Examples of such compilers: Clang++ v13, v14.


danhoeflinger (approved):
LGTM.  
As a confirmation this feature (`CHECK_INCLUDE_FILE_CXX`) and the way we use it is available for CMake 3.11, so this does not impact our minimum CMake version requirements.

dmitriy-sobolev:
I am going to merge it after the next release in order not to bring more risks.

PR #961, opened 2023-05-23
Add support on float type into onesweep esimd radix sort

SergeyKopienko:
In this PR we :
- refactor of radix_sort_esimd.pass.cpp test to avoid duplicated code;
- add support on float type into ```onesweep``` esimd radix sort.


test/parallel_api/experimental/radix_sort_esimd.pass.cpp
```diff
@@ -119,7 +162,7 @@ void test_all_view(std::size_t size)
     EXPECT_EQ_RANGES(ref, input, msg.c_str());
 }
 
-template<typename T>
```
dmitriy-sobolev, line 122: I think it is enough to test different USM allocation only inside `test_usm`.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
```diff
@@ -87,7 +160,7 @@ void global_histogram(sycl::nd_item<1> idx, size_t __n, const InputT& input, uin
                 sycl::ext::intel::esimd::simd offset((read_addr + s + lane_id)*sizeof(KeyT));
                 simd<KeyT, 16> source = lsc_gather<KeyT, 1, lsc_data_size::default_size, cache_hint::cached, cache_hint::cached, 16>(input, offset, m);
 
-                keys.template select<16, 1>(s) = merge(source, simd<KeyT, 16>(utils::__sort_identity<KeyT, IsAscending>), m);
+                keys.template select<16, 1>(s) =merge(source, simd<KeyT, 16>(__full_sort_identity<KeyT, IsAscending>()), m);
```
dmitriy-sobolev, line 163: Missing space

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
```diff
@@ -20,6 +20,79 @@
 
 namespace oneapi::dpl::experimental::esimd::impl
 {
+// float, 32bit
+template <typename T, bool __is_ascending, std::enable_if_t<std::is_same<T, float>::value && sizeof(T) == sizeof(::std::uint32_t), int> = 0>
+constexpr T
+__full_sort_identity()
```
dmitriy-sobolev, line 26: What do you think about moving this implementation into the utils header alongside with `__sort_identity`? 
I also suggest modifying the original `__sort_identity` so that it is a callable, and renaming `__full_sort_identity` to `__sort_identity` so it is just a specialization.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
```diff
@@ -20,6 +20,79 @@
...
+{
+    static_assert(std::is_same<T, float>::value, "");
+
+    //                                                                    sign bit
+    //                                                                      |      NaN bit
+    //                                                                      |        |
+    //                                                                      V        V
+    static_assert(::std::numeric_limits<T>::max()    == sycl::bit_cast<T>(0b01111111'01111111'11111111'11111111), "");
+    static_assert(::std::numeric_limits<T>::lowest() == sycl::bit_cast<T>(0b11111111'01111111'11111111'11111111), "");
```
dmitriy-sobolev, line 35: Perhaps, these static asserts are not needed here. It may be even dangerous, because as far as I know c++ standard does not state that the representation of a floating point number is consistent with IEEE-754.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
```diff
@@ -20,6 +20,79 @@
...
+
+    if constexpr (__is_ascending)
+    {
+        //                       clear sign bit
+        //                         |      setup NaN bit
+        //                         |        |
+        //                         V        V
+        return sycl::bit_cast<T>(0b01111111'11111111'11111111'11111111);
```
dmitriy-sobolev, line 43: Let's use Hex representation. It is consistent with other functions doing bit-related operations.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
```diff
@@ -20,6 +20,79 @@
 
 namespace oneapi::dpl::experimental::esimd::impl
 {
+// float, 32bit
```
dmitriy-sobolev, line 23: Could you also leave a comment describing why using `std::numeric_limits<T>::max()` or `std::numeric_limits<T>::lowest()` as an identity does not work when performing radix sort of floating point numbers?

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
dmitriy-sobolev: Here is a note in [cppreference](https://en.cppreference.com/w/c/language/arithmetic_types):
> float - single precision floating-point type. Matches [IEEE-754 binary32 format](https://en.wikipedia.org/wiki/Single-precision_floating-point_format) if supported.

test/parallel_api/experimental/radix_sort_esimd.pass.cpp
```diff
@@ -39,15 +39,58 @@
 #include <limits>
 
 constexpr bool Ascending = true;
+constexpr bool Descending = !Ascending;
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed (removed)

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
SergeyKopienko: Fixed

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
SergeyKopienko: The main idea of this static assert to give compile error if we have the different ```float``` presentation in memory.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
SergeyKopienko: I believe we able to do this later. Just now this binary format more clear for me.

test/parallel_api/experimental/radix_sort_esimd.pass.cpp
SergeyKopienko: This test organized like all out other tests from this point of view.
I don't think that we should have some special in this test.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
dmitriy-sobolev: What is the purpose of aborting compilation if the representation is different? As far as I see the new sort identity does not depend on it, it just sets all the bits to 1 (with the exception of descending order with the sign bit set to 0).

test/parallel_api/experimental/radix_sort_esimd.pass.cpp
dmitriy-sobolev: Compilation time of that test is already huge and we agreed not to make "combinatoric explosion" of the cases. I believe this case does not bring much value and it is better not to include it.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
dmitriy-sobolev: We discussed it offline and agreed to leave it as is for now

include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
dmitriy-sobolev: We discussed it offline and agreed to leave it as is for easier debugging and return to it during productization.

dmitriy-sobolev (approved):

PR #962, opened 2023-05-23
Adding sycl header check

danhoeflinger:
Adding commit b6ea2df988fd0f5f4ea02b9abc38b42f181a9a11 from #906 into dynamic_selection branch:

Adding checks for the presence of sycl headers prior to enabling sycl backend.
This is required when compilers support -fsycl option and define SYCL_LANGUAGE_VERSION, without making SYCL implementation available in practice.


timmiesmith (approved):
LGTM

PR #965, opened 2023-05-24
Fix repeated kernel names for__reduce_mid_main_kernel

julianmi:
This PR addresses a clash in kernel names for `__reduce_mid_main_kernel` introduced in #907. We currently use the same value for `__iters_per_work_item1`, which leads to the same name. This can be observed wen compiling with `ONEDPL_USE_UNNAMED_LAMBDA=OFF`.

This fix adds both `__iters_per_work_item` parameters to the kernel name provider to come up with unique names. It further prevents potential issues in the future in case we want to vary `__iters_per_work_item1` for different kernel invocations.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -244,10 +244,13 @@ __parallel_transform_reduce_mid_impl(_ExecutionPolicy&& __exec, _Size __n, const
 {
     using _Policy = typename ::std::decay<_ExecutionPolicy>::type;
     using _CustomName = typename _Policy::kernel_name;
+    using __iters_per_work_item1_t = ::std::integral_constant<::std::uint8_t, __iters_per_work_item1>;
+    using __iters_per_work_item2_t = ::std::integral_constant<::std::uint8_t, __iters_per_work_item2>;
+
     using _ReduceMainKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-        __reduce_mid_main_kernel<::std::integral_constant<::std::uint8_t, __iters_per_work_item1>, _CustomName>>;
+        __reduce_mid_main_kernel<__iters_per_work_item1_t, __iters_per_work_item2_t, _CustomName>>;
     using _ReduceLeafKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-        __reduce_mid_leaf_kernel<::std::integral_constant<::std::uint8_t, __iters_per_work_item2>, _CustomName>>;
+        __reduce_mid_leaf_kernel<__iters_per_work_item1_t, __iters_per_work_item2_t, _CustomName>>;
 
     return __parallel_transform_reduce_mid_submitter<__work_group_size, __iters_per_work_item1, __iters_per_work_item2,
                                                      _Tp, _ReduceMainKernel, _ReduceLeafKernel>()(
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -50,9 +50,8 @@ template <typename _Tp, typename _NDItemId, typename _Size, typename _TransformP
           typename _InitType, typename _AccLocal, typename _Res, typename... _Acc>
 void
 __work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Size __n_items,
-                           _TransformPattern __transform_pattern, _ReducePattern __reduce_pattern,
-                           _InitType __init, const _AccLocal& __local_mem, const _Res& __res_acc,
-                           const _Acc&... __acc)
+                           _TransformPattern __transform_pattern, _ReducePattern __reduce_pattern, _InitType __init,
+                           const _AccLocal& __local_mem, const _Res& __res_acc, const _Acc&... __acc)
 {
     auto __local_idx = __item_id.get_local_id(0);
     // 1. Initialization (transform part). Fill local memory
@@ -104,8 +103,8 @@ struct __parallel_transform_reduce_small_submitter<__work_group_size, __iters_pe
               oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0,
               typename... _Ranges>
     auto
-    operator()(_ExecutionPolicy&& __exec, const _Size __n, _ReduceOp __reduce_op,
-               _TransformOp __transform_op, _InitType __init, _Ranges&&... __rngs) const
+    operator()(_ExecutionPolicy&& __exec, const _Size __n, _ReduceOp __reduce_op, _TransformOp __transform_op,
+               _InitType __init, _Ranges&&... __rngs) const
     {
         auto __transform_pattern =
             unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _TransformOp>{
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -50,8 +50,8 @@ template <typename _Tp, typename _NDItemId, typename _Size, typename _TransformP
           typename _InitType, typename _AccLocal, typename _Res, typename... _Acc>
 void
 __work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Size __n_items,
-                           const _TransformPattern& __transform_pattern, const _ReducePattern& __reduce_pattern,
-                           const _InitType& __init, const _AccLocal& __local_mem, const _Res& __res_acc,
+                           _TransformPattern __transform_pattern, _ReducePattern __reduce_pattern,
+                           _InitType __init, const _AccLocal& __local_mem, const _Res& __res_acc,
                            const _Acc&... __acc)
 {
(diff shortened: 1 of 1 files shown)


julianmi:
> I'm failing to understand why this is necessary, and therefore I'm unsure this is the right fix.
> 
> From my reading, only `__iters_per_work_item1` is used in `__reduce_mid_main_kernel`, via `__transform_pattern1`, and only `__iters_per_work_item2` is used in `__reduce_mid_leaf_kernel`, via `__transform_pattern2`.
> 
> It seems like any clashes in name generation would still be there after his change, as we are adding variables to the name which don't impact the kernels being generated.
> 
> I'm sure I'm missing something but I don't see it.

The idea is to use the combination of `__iters_per_work_item1` and `__iters_per_work_item2` in both kernel names since the combination is unique. `__reduce_mid_main_kernel` currently uses the same value for `__iters_per_work_item1` for all calls since 32 has shown to work well. What we vary in the mid-sized kernel is how many elements we assign to the second kernel. Thus, only `__iters_per_work_item2` varies. I have added both in case we want to vary `__iters_per_work_item1` in the future as well.

danhoeflinger:
> The idea is to use the combination of `__iters_per_work_item1` and `__iters_per_work_item2` in both kernel names since the combination is unique. `__reduce_mid_main_kernel` currently uses the same value for `__iters_per_work_item1` for all calls since 32 has shown to work well. What we vary in the mid-sized kernel is how many elements we assign to the second kernel. Thus, only `__iters_per_work_item2` varies. I have added both in case we want to vary `__iters_per_work_item1` in the future as well.

But if we always use the same `__iters_per_work_item1` value for all `__reduce_mid_main_kernel`, then it should generate the kernel once and rely on it multiple times, not have a naming clash unless something else which impacts the kernel is not incorporated into the name.  It seems like we are missing something else that needs to be in the kernel name because it is resulting in different kernels.

SergeyKopienko (approved):
LGTM

julianmi:
> > The idea is to use the combination of `__iters_per_work_item1` and `__iters_per_work_item2` in both kernel names since the combination is unique. `__reduce_mid_main_kernel` currently uses the same value for `__iters_per_work_item1` for all calls since 32 has shown to work well. What we vary in the mid-sized kernel is how many elements we assign to the second kernel. Thus, only `__iters_per_work_item2` varies. I have added both in case we want to vary `__iters_per_work_item1` in the future as well.
> 
> But if we always use the same `__iters_per_work_item1` value for all `__reduce_mid_main_kernel`, then it should generate the kernel once and rely on it multiple times, not have a naming clash unless something else which impacts the kernel is not incorporated into the name. It seems like we are missing something else that needs to be in the kernel name because it is resulting in different kernels.

Isn't the problem that `__parallel_transform_reduce_mid_impl` is called multiple times and each calls `__kernel_name_provider` with the same parameters? Thus, we end up with multiple kernels with the same name. Do you suggest to extract `__reduce_mid_main_kernel` such that we need to compile it only once?

danhoeflinger:
> > > The idea is to use the combination of `__iters_per_work_item1` and `__iters_per_work_item2` in both kernel names since the combination is unique. `__reduce_mid_main_kernel` currently uses the same value for `__iters_per_work_item1` for all calls since 32 has shown to work well. What we vary in the mid-sized kernel is how many elements we assign to the second kernel. Thus, only `__iters_per_work_item2` varies. I have added both in case we want to vary `__iters_per_work_item1` in the future as well.
> > 
> > 
> > But if we always use the same `__iters_per_work_item1` value for all `__reduce_mid_main_kernel`, then it should generate the kernel once and rely on it multiple times, not have a naming clash unless something else which impacts the kernel is not incorporated into the name. It seems like we are missing something else that needs to be in the kernel name because it is resulting in different kernels.
> 
> Isn't the problem that `__parallel_transform_reduce_mid_impl` is called multiple times and each calls `__kernel_name_provider` with the same parameters? Thus, we end up with multiple kernels with the same name. Do you suggest to extract `__reduce_mid_main_kernel` such that we need to compile it only once?

Perhaps I am misunderstanding the way that the sycl compiler works, but I believe if the exact same kernel is called multiple times, it should reuse the previously compiled binaries / name without issue.

I believe that there are some other compile time parameters which are missing from the kernel name which impact the kernel generation.  I'm looking through the code now to try to figure out what is missing.  I'm thinking now that `_Tp`, `_ReduceOp`, `_TransformOp`  are needed.  I thought they were already folded into the `_CustomName` previously, but I don't see being done anywhere.  (I'm not sure that list is correct or complete yet)

danhoeflinger (dismissed):
I don't think that we need to add the other `__iters_per_work_item`, but rather the following 
`_Tp`, `_ReduceOp`, `_TransformOp`, `_Ranges...` to main kernel and 
`_Tp`, `_ReduceOp`, `_InitType`, `_Ranges...` to leaf kernel.

(edited to add `Ranges...`)

julianmi:
> I don't think that we need to add the other `__iters_per_work_item`, but rather the following `_Tp`, `_ReduceOp`, `_TransformOp` to main kernel and `_Tp`, `_ReduceOp`, `_InitType` to leaf kernel.

This does not solve the issue. @rarutyun would you mind having a look please? The issue seems to be only with the `__reduce_mid_main_kernel` where we define the same kernel and name multiples times.

danhoeflinger:
> > I don't think that we need to add the other `__iters_per_work_item`, but rather the following `_Tp`, `_ReduceOp`, `_TransformOp` to main kernel and `_Tp`, `_ReduceOp`, `_InitType` to leaf kernel.
> 
> This does not solve the issue. @rarutyun would you mind having a look please? The issue seems to be only with the `__reduce_mid_main_kernel` where we define the same kernel and name multiples times.

I guess there are two separate things at play here:
1) `_reduce_mid_main_kernel` is defined from multiple different template-generated origin locations but resulting in the same generated name / kernel.  In a discussion with @julianmi offline, I think this can be resolved by separating the submitters for `__reduce_mid_main_kernel` and `__reduce_mid_leaf_kernel` where they each only have their relevant template parameters.  The result would be that there is only one unified location for each unique kernel definition.

2) The question of does `__kernel_name_provider` have all the parameters it needs here?  It seems to me that it requires the extra parameters mentioned above to be unique for each generated kernel.  However, looking around the code it seems that in other locations when `__kernel_name_provider` is used rather than `__kernel_name_generator`, fewer template parameters are passed than I would expect.  
My understanding of the difference between these is that `__kernel_name_generator` should be used for kernels where precompilation is necessary to query properties, and `__kernel_name_provider` should be used otherwise to avoid reliance on a builtin for name creation.
@rarutyun Are there different requirements for the "uniqueness" of the name provided to these two options?

@julianmi mentioned and I agree that we should document the result of this discussion into the code so that it is more clear which kernel name function to use and what the specific requirements are for each.

julianmi:
While the previous solution fixes the kernel name clashes, multiple kernels are stills used for the first kernel of the mid-sized array pattern. This is not needed since we use the same template parameters. However, `__parallel_transform_reduce_mid_impl` is parameterized with both `__iters_per_work_item` parameters which instantiates the submitter for each call in the `else if` statements in `__parallel_transform_reduce`. @rarutyun and @danhoeflinger proposed to split the mid-sized submitter to only have one instance. This resolves the kernel name issue and improves compilation performance and binary sizes.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -222,37 +236,51 @@ struct __parallel_transform_reduce_mid_submitter<__work_group_size, __iters_per_
...
+
+    // number of buffer elements processed within workgroup
+    constexpr _Size __size_per_work_group = __iters_per_work_item1 * __work_group_size;
+    const _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
+    sycl::buffer<_Tp> __temp{sycl::range<1>(__n_groups)};
+
+    sycl::event __reduce_event =
+        __parallel_transform_reduce_device_kernel_submitter<__work_group_size, __iters_per_work_item1, _Tp,
+                                                            _ReduceDeviceKernel>()(
+            ::std::forward<_ExecutionPolicy>(__exec), __n, __reduce_op, __transform_op, __init, __temp,
```
rarutyun, line 277: Forwarding execution policy twice is incorrect. The second one is below

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -222,37 +236,51 @@ struct __parallel_transform_reduce_mid_submitter<__work_group_size, __iters_per_
...
+            ::std::forward<_Ranges>(__rngs)...);
+
+    __n = __n_groups; // Number of preliminary results from the device kernel.
+    return __parallel_transform_reduce_work_group_kernel_submitter<__work_group_size, __iters_per_work_item2, _Tp,
```
rarutyun, line 281: I think ``__iters_per_work_item1`` ``__iters_per_work_item2`` deserve renaming. The name as it is today does not tell anything and it's easy to mess up with 1 and 2 numbers at the end.

rarutyun (commented):
Overall, looks good to me except two comments

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: Applied.

danhoeflinger (approved):
LGTM.
I think we can improve the documentation for `__kernel_name_provider` and `__kernel_name_generator` in a separate PR which adds some comments to that code.  Thanks to @rarutyun and @julianmi for discussions there.

rarutyun (approved):

danhoeflinger (approved):

danhoeflinger (approved):

PR #966, opened 2023-05-25
Implement & test esimd radix sort for int16_t and uint16_t

SergeyKopienko:
In this PR we add support of ```int16_t``` and ```uint16_t``` into esimd radix sort.

For that we made change in ```one_wg_kernel``` function:
- we replace block of code
```
            #pragma unroll
            for (uint32_t s = 0; s<PROCESS_SIZE; s+=64){
                keys.template select<64, 1>(s) = lsc_slm_block_load<KeyT, 64>(slm_reorder_this_thread+s*sizeof(KeyT));
            }
```
to
```
            keys = utils::BlockLoad<KeyT, PROCESS_SIZE>(slm_reorder_this_thread);
```
to avoid compile error:
```
../include/sycl/ext/intel/experimental/esimd/memory.hpp:381:3: error: static assertion failed due to requirement '_DS == lsc_data_size::u32 || _DS == lsc_data_size::u64': Transposed load is supported only for data size u32 or u64
```

### DIFF
--- a/test/parallel_api/experimental/radix_sort_esimd.pass.cpp
+++ b/test/parallel_api/experimental/radix_sort_esimd.pass.cpp
@@ -310,6 +310,7 @@ int main()
         {
             test_usm<uint32_t>(size);
             test_usm<int>(size);
+            test_usm<int16_t>(size);
             test_usm<float>(size);
         }
         test_small_sizes();
--- a/test/parallel_api/experimental/radix_sort_esimd.pass.cpp
+++ b/test/parallel_api/experimental/radix_sort_esimd.pass.cpp
@@ -68,6 +68,13 @@ struct TypeInfo
     }
 
     template <>
+    const std::string& name<uint16_t>()
+    {
+        static const std::string kTypeName = "uint16_t";
+        return kTypeName;
+    }
+
+    template <>
     const std::string& name<uint32_t>()
     {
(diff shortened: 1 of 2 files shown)


test/parallel_api/experimental/radix_sort_esimd.pass.cpp
```diff
@@ -296,23 +315,32 @@ int main()
...
             test_usm<int>(size);
+            test_usm<int16_t>(size);
+            test_usm<uint16_t>(size);
             test_usm<float>(size);
         }
-        test_small_sizes();
+        test_small_sizes<uint32_t>();
+        test_small_sizes<int>();
+        test_small_sizes<int16_t>();
+        test_small_sizes<uint16_t>();
```
dmitriy-sobolev, line 343: Let's leave only on case as before. The purpose of that test case is to make sure the input consequence is not changed when the number of elements to sort is 0 or 1. There is no need to test different types.

SergeyKopienko: Fixed

dmitriy-sobolev (approved):
LGTM. Could you also describe the issue being fixed with `lsc_slm_block_load<KeyT, 64>(slm_reorder_this_thread+s*sizeof(KeyT));`? I understood it during offline discussion, but it may be useful to other developers.

PR #972, opened 2023-05-30
Fix for zip_view test on Windows

danhoeflinger:
This PR fixes a bug in zip_view.pass.exe on Windows.  

`iota_view` bases its' value type on the first argument,  making an iota with `0L` on Windows results in a 32 signed integer type, which overflows during the test when using the large numbers (designed to test overflow of 32 bit signed types).  This is resolved by explicitly creating a 64 bit integer in the iota creation.


MikeDvorskiy (approved):
LGTM

Issue #955, opened 2023-05-22
Wrong work-group size computed in radix_sort


joeatodd:
I think I found a bug while looking at the current radix sort implementation.

On [this line](https://github.com/oneapi-src/oneDPL/blob/8b03924ddbaa1de82fac794a1fa18903f77231f9/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h#LL607C20-L607C20), the computed maximum work-group size  is divided by the number of radix states:

``` c++
__block_size = oneapi::dpl::__internal::__dpl_ceiling_div(__max_count_wg_size, __radix_states);
```

On my system with an RTX 3060, this results in a `__block_size` of 2. I suspect what is intended here is to round `__max_count_wg_size` up to a multiple of __radix_states, e.g.:

``` c++
__block_size = oneapi::dpl::__internal::__dpl_ceiling_div(__max_count_wg_size, __radix_states) *  __radix_states;
```

At present, this bug is made semi-benign by the next line:

``` c++
__block_size = sycl::max(oneapi::dpl::__internal::__dpl_bit_floor(__block_size), ::std::size_t(__radix_states));
```

which produces `__block_size` of 16, assuming `__radix_states = 16`. Obviously this still isn't ideal, esp. on Nvidia hardware.

MikeDvorskiy:
Hello, @joeatodd 
Why "this still isn't ideal, esp. on Nvidia hardware."? Do you mean subgroup size should be 32 in case of NVidia warp?

joeatodd:
Hi @MikeDvorskiy,

Yes the fix I mentioned is what I'm using in my local branch. It works. I flagged this as an issue (rather than PR) because I wasn't totally confident it was a bug. If you agree it's a bug, I'm happy to push a PR to fix it.

> Do you mean subgroup size should be 32 in case of NVidia warp?

Yes that's right. And with the proposed fix, the block_size is indeed 32 on Nvidia hardware.

MikeDvorskiy:
Good.  We just have discussed in our team internally and tend to conclusion that it is a bug. So, you may  push PR to fix.
Thank you!

dmitriy-sobolev:
I guess the issue has been solve with merging #975. I am closing it.

PR #975, opened 2023-06-02
Fix block size computation bug in sycl radix_sort

joeatodd:
This is a bugfix for #955, which lead to the wrong block size being computed in sycl `radix_sort`.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
@@ -605,7 +605,7 @@ struct __parallel_radix_sort_iteration
         // correct __block_size according to local memory limit in count phase
         const auto __max_count_wg_size = oneapi::dpl::__internal::__slm_adjusted_work_group_size(
             __exec, sizeof(typename __decay_t<_TmpBuf>::value_type) * __radix_states, __block_size);
-        __block_size = oneapi::dpl::__internal::__dpl_ceiling_div(__max_count_wg_size, __radix_states) * __radix_states;
+        __block_size = static_cast<::std::size_t>((__max_count_wg_size / __radix_states)) * __radix_states;
 
         // block size must be a power of 2 and not less than the number of states.
         // TODO: Check how to get rid of that restriction.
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -605,7 +605,7 @@ struct __parallel_radix_sort_iteration
         // correct __block_size according to local memory limit in count phase
         const auto __max_count_wg_size = oneapi::dpl::__internal::__slm_adjusted_work_group_size(
             __exec, sizeof(typename __decay_t<_TmpBuf>::value_type) * __radix_states, __block_size);
-        __block_size = oneapi::dpl::__internal::__dpl_ceiling_div(__max_count_wg_size, __radix_states);
+        __block_size = oneapi::dpl::__internal::__dpl_ceiling_div(__max_count_wg_size, __radix_states) * __radix_states;
```
dmitriy-sobolev, line 608: Let's do "floor" division instead of "ceil" division.

```suggestion
        __block_size = static_cast<::std::size_t>((__max_count_wg_size / __radix_states)) * __radix_states;
```

 `__slm_adjusted_work_group_size` returns the maximum number of work-items in a work-group according to SLM usage per each work-item. It is `__max_count_wg_size`. With "ceil" division, we may end up with `__block_size` bigger than `__max_count_wg_size` and more SLM allocated than the device allows.

joeatodd: Yeah, good point. I guess in that case we might also want to assert that we don't end up with a `__block_size` of zero?

dmitriy-sobolev: Thanks for noticing it. I would suggest leaving it as is right now. Unfortunately, I do not see an easy way to handle that case properly. I've created #977 with a suggestion to implement a fallback implementation of count kernel which does not use SLM.

Line 612 will correct `0` to `__radix_bits`. However, If we get `0`, it will indicate that we do not have enough SLM. I expect that SYCL backend runtime will throw an exception about not having enough resources.

dmitriy-sobolev (commented):
LGTM, 
@MikeDvorskiy do you have any objections to merge it? The solution slightly differs from on proposed in #955.

dmitriy-sobolev:
clang-format fails due to misalignments in places which are not tackled in this PR. OpenMP-related issue is fixed in #978.

dmitriy-sobolev:
Thanks, @joeatodd. The change is expected to be a part of the next oneDPL release (2022.3).

PR #976, opened 2023-06-02
Fix duplicate kernel names in copy_if

adamfidel:
The kernel names for `copy_if` were missing static parameters (e.g., work-group size) and thus multiple kernels were being generated with the same name when compiled with `-fno-sycl-unnamed-lambda`. This PR adds relevant parameters to the kernel names to fix the duplicate naming errors.


timmiesmith (approved):
LGTM.

PR #982, opened 2023-06-14
[KT] Fix sync error in cooperative impl of esimd radix sort

SergeyKopienko:
In this PR we fix syncronization error in cooperative impl of esimd radix sort.
This error have place for 64-bit data types.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
@@ -22,19 +22,24 @@
 namespace oneapi::dpl::experimental::esimd::impl
 {
 
+template <uint32_t STAGES>
 void inline init_global_sync(uint32_t * psync, uint32_t tg_id) {
     using namespace __ESIMD_NS;
     using namespace __ESIMD_ENS;
-    simd<uint32_t, 16> lane_id(0, 4);
-    simd<uint32_t, 16> old_value = lsc_atomic_update<atomic_op::load, uint32_t, 16>(psync, lane_id, 1);
+
+    constexpr uint32_t SYNC_BUFFER_PER_STAGE = 4;
+    constexpr uint32_t SYNC_BUFFER_SIZE = SYNC_BUFFER_PER_STAGE * STAGES;
+    
+    simd<uint32_t, SYNC_BUFFER_SIZE> lane_id(0, 4);
+    simd<uint32_t, SYNC_BUFFER_SIZE> old_value = lsc_atomic_update<atomic_op::load, uint32_t, SYNC_BUFFER_SIZE>(psync, lane_id, 1);
     if (tg_id == 0) {
         if (!(old_value==1).all()) {
-            lsc_atomic_update<atomic_op::store, uint32_t, 16>(psync, lane_id, 1, 1);
+            lsc_atomic_update<atomic_op::store, uint32_t, SYNC_BUFFER_SIZE>(psync, lane_id, 1, 1);
         }
     } else {
         uint32_t try_count = 0;
         while (!(old_value==1).all()) {
-            old_value = lsc_atomic_update<atomic_op::load, uint32_t, 16>(psync, lane_id, 1);
+            old_value = lsc_atomic_update<atomic_op::load, uint32_t, SYNC_BUFFER_SIZE>(psync, lane_id, 1);
             if (try_count++ > 10240) break;
         }
     }
@@ -110,7 +115,7 @@ void cooperative_kernel(sycl::nd_item<1> idx, size_t n, const InputT& input, Key
     constexpr uint32_t BIN_WIDTH_UD = BIN_COUNT / BIN_GROUPS * sizeof(hist_t) / sizeof(uint32_t);
     constexpr uint32_t BIN_HEIGHT = THREAD_PER_TG / THREAD_PER_BIN_GROUP;
 
-    if (local_tid == 0) init_global_sync(p_sync_buffer, tg_id);
+    if (local_tid == 0) init_global_sync<STAGES>(p_sync_buffer, tg_id);
     barrier();
 
     #pragma unroll
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
@@ -22,12 +22,13 @@
 namespace oneapi::dpl::experimental::esimd::impl
 {
 
+constexpr uint32_t SYNC_BUFFER_PER_STAGE = 4;
+
 template <uint32_t STAGES>
 void inline init_global_sync(uint32_t * psync, uint32_t tg_id) {
     using namespace __ESIMD_NS;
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
```diff
@@ -22,19 +22,24 @@
...
 
+template <uint32_t STAGES>
 void inline init_global_sync(uint32_t * psync, uint32_t tg_id) {
     using namespace __ESIMD_NS;
     using namespace __ESIMD_ENS;
-    simd<uint32_t, 16> lane_id(0, 4);
-    simd<uint32_t, 16> old_value = lsc_atomic_update<atomic_op::load, uint32_t, 16>(psync, lane_id, 1);
+
+    constexpr uint32_t SYNC_BUFFER_PER_STAGE = 4;
+    constexpr uint32_t SYNC_BUFFER_SIZE = SYNC_BUFFER_PER_STAGE * STAGES;
```
SergeyKopienko, line 31: We should have here 4 items for each stage.
So initial const 16 is OK for 4 stages only.
For 8 stages we should have 32 instead 16.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
```diff
@@ -110,7 +115,7 @@ void cooperative_kernel(sycl::nd_item<1> idx, size_t n, const InputT& input, Key
     constexpr uint32_t BIN_WIDTH_UD = BIN_COUNT / BIN_GROUPS * sizeof(hist_t) / sizeof(uint32_t);
     constexpr uint32_t BIN_HEIGHT = THREAD_PER_TG / THREAD_PER_BIN_GROUP;
 
-    if (local_tid == 0) init_global_sync(p_sync_buffer, tg_id);
+    if (local_tid == 0) init_global_sync<STAGES>(p_sync_buffer, tg_id);
```
SergeyKopienko, line 118: Now STAGES count required in the ```init_global_sync``` function.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
```diff
@@ -22,19 +22,25 @@
 namespace oneapi::dpl::experimental::esimd::impl
 {
 
+constexpr uint32_t SYNC_BUFFER_PER_STAGE = 4;
```
SergeyKopienko, line 25: We should have here 4 items for each stage.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
```diff
@@ -201,7 +207,7 @@ void cooperative_kernel(sycl::nd_item<1> idx, size_t n, const InputT& input, Key
             // global sync()
             // each tg read global scan buffer, add with prev tg, then put to slm incoming buffer
             if (local_tid == 0) {
-                global_sync(p_sync_buffer, stage*4+0, tg_count, tg_id, local_tid);
+                global_sync(p_sync_buffer, stage*SYNC_BUFFER_PER_STAGE+0, tg_count, tg_id, local_tid);
```
SergeyKopienko, line 210: We should have here 4 items for each stage.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
```diff
@@ -234,7 +240,7 @@ void cooperative_kernel(sycl::nd_item<1> idx, size_t n, const InputT& input, Key
                     lsc_fence<lsc_memory_kind::untyped_global, lsc_fence_op::evict, lsc_scope::gpu>();
                 }
                 if (local_tid == 0)  {
-                    global_sync(p_sync_buffer, stage*4+1, tg_count, tg_id, local_tid);
+                    global_sync(p_sync_buffer, stage*SYNC_BUFFER_PER_STAGE+1, tg_count, tg_id, local_tid);
```
SergeyKopienko, line 243: We should have here 4 items for each stage.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
```diff
@@ -290,7 +296,7 @@ void cooperative_kernel(sycl::nd_item<1> idx, size_t n, const InputT& input, Key
             }
             lsc_fence<lsc_memory_kind::untyped_global, lsc_fence_op::evict, lsc_scope::gpu>();
             barrier();
-            if (local_tid == 0) {global_sync(p_sync_buffer, stage*4+2, tg_count, tg_id, local_tid);}
+            if (local_tid == 0) {global_sync(p_sync_buffer, stage*SYNC_BUFFER_PER_STAGE+2, tg_count, tg_id, local_tid);}
```
SergeyKopienko, line 299: We should have here 4 items for each stage.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
```diff
@@ -22,19 +22,25 @@
...
+template <uint32_t STAGES>
 void inline init_global_sync(uint32_t * psync, uint32_t tg_id) {
     using namespace __ESIMD_NS;
     using namespace __ESIMD_ENS;
-    simd<uint32_t, 16> lane_id(0, 4);
-    simd<uint32_t, 16> old_value = lsc_atomic_update<atomic_op::load, uint32_t, 16>(psync, lane_id, 1);
+
+    constexpr uint32_t SYNC_BUFFER_SIZE = SYNC_BUFFER_PER_STAGE * STAGES;
+    
+    simd<uint32_t, SYNC_BUFFER_SIZE> lane_id(0, 4);
```
akukanov, line 34: Is 4 here the same constant?

include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
akukanov, line 210: As far as I can see, `local_tid` is always 0 when `global_sync` is called, and moreover the corresponding parameter is not used by the function. If it is not needed, let's remove.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_cooperative.h
SergeyKopienko: I think - no: as I understand ```0``` - value of first item of simd, ```4``` - step to increment each next value.

akukanov: But is not the step supposed to be equal to SYNC_BUFFER_PER_STAGE?

SergeyKopienko: You are right. Fixed.

akukanov: No, I analyzed it and I think I was wrong :) and this `4` really means `sizeof(uint32_t)`.

akukanov (approved):

PR #983, opened 2023-06-14
Bugfix for set_intersection copying from second iterator range

danhoeflinger:
Bug Description
---
As implemented before this PR, `set_intersection` using `oneapi::dpl::execution::par` has a bug which results in incorrect semantics for some situations.  
For performance reasons, in some situations, we would prefer to reverse the order of arguments in the parallel loop for input iterators.  However, this results in copying from the second set of iterators which is against semantics found here:

https://en.cppreference.com/w/cpp/algorithm/set_intersection
..."elements will be copied from [first1, last1) to the output range, preserving order."  

In cases where the comparator does not compare ALL of the input type, this can be very important.  For instance, a comparator for a key value pair which only considers the key.  

Fix Summary
---
This PR fixes this bug by passing a functor to `__set_intersection_construct` which selects which element to copy to the output range.  This allows us to keep the structure of the reversed iterator arguments without the issue of copying from the wrong input set.

This PR also adds a test which fails to compile without this fix using a key value pair example which only compares the key.  It uses a `oneapi::dpl::counting_iterator<int>` as a "value" for the first set of iterators, a `oneapi::dpl::discard_iterator` as a "value" for the second set of iterators, and a sequence of integers for the output "values".  If the implementation were to attempt to copy from the second set of iterators, a compiler error will occur when attempting to assign the dereferenced `oneapi::dpl::discard_iterator` to an integer.

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3551,6 +3551,7 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                    _RandomAccessIterator2 __last2, _T* __result, _Compare __comp) {
                     return oneapi::dpl::__utils::__set_intersection_construct(__first1, __last1, __first2, __last2,
                                                                               __result, __comp,
+                                                                              oneapi::dpl::__internal::__op_uninitialized_copy<_ExecutionPolicy>{},
                                                                               /*CopyFromFirstSet = */ ::std::true_type());
                 }, __is_vector);
         });
@@ -3568,6 +3569,7 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                    _RandomAccessIterator2 __last2, _T* __result, _Compare __comp) {
                     return oneapi::dpl::__utils::__set_intersection_construct(__first2, __last2, __first1, __last1,
                                                                               __result, __comp,
+                                                                              oneapi::dpl::__internal::__op_uninitialized_copy<_ExecutionPolicy>{},
                                                                               /*CopyFromFirstSet = */ ::std::false_type());
                 }, __is_vector);
             return __result;
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3552,7 +3552,7 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                     return oneapi::dpl::__utils::__set_intersection_construct(
                         __first1, __last1, __first2, __last2, __result, __comp,
                         oneapi::dpl::__internal::__op_uninitialized_copy<_ExecutionPolicy>{},
-                        /*CopyFromFirstSet = */ ::std::true_type());
+                        /*CopyFromFirstSet = */ std::true_type{});
                 });
         });
     }
@@ -3570,7 +3570,7 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                     return oneapi::dpl::__utils::__set_intersection_construct(
                         __first2, __last2, __first1, __last1, __result, __comp,
                         oneapi::dpl::__internal::__op_uninitialized_copy<_ExecutionPolicy>{},
-                        /*CopyFromFirstSet = */ ::std::false_type());
+                        /*CopyFromFirstSet = */ std::false_type{});
                 });
             return __result;
         });
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3550,7 +3550,7 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                 [](_RandomAccessIterator1 __first1, _RandomAccessIterator1 __last1, _RandomAccessIterator2 __first2,
                    _RandomAccessIterator2 __last2, _T* __result, _Compare __comp) {
                     return oneapi::dpl::__utils::__set_intersection_construct(__first1, __last1, __first2, __last2,
-                                                                              __result, __comp);
+                                                                              __result, __comp, [](auto&& a, auto&& b) { return std::forward<decltype(a)>(a); }); //copy first to output
                 });
         });
     }
@@ -3566,7 +3566,7 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                 [](_RandomAccessIterator1 __first1, _RandomAccessIterator1 __last1, _RandomAccessIterator2 __first2,
                    _RandomAccessIterator2 __last2, _T* __result, _Compare __comp) {
                     return oneapi::dpl::__utils::__set_intersection_construct(__first2, __last2, __first1, __last1,
-                                                                              __result, __comp);
+                                                                              __result, __comp, [](auto&& a, auto&& b) { return std::forward<decltype(b)>(b); }); //copy second to output                                                                              
                 });
             return __result;
         });
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3553,7 +3553,7 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                                                                               __result, __comp,
                                                                               oneapi::dpl::__internal::__op_uninitialized_copy<_ExecutionPolicy>{},
                                                                               /*CopyFromFirstSet = */ ::std::true_type());
-                }, __is_vector);
+                });
         });
     }
 
@@ -3571,7 +3571,7 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                                                                               __result, __comp,
                                                                               oneapi::dpl::__internal::__op_uninitialized_copy<_ExecutionPolicy>{},
                                                                               /*CopyFromFirstSet = */ ::std::false_type());
-                }, __is_vector);
+                });
             return __result;
         });
     }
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -186,7 +188,7 @@ __set_intersection_construct(_ForwardIterator1 __first1, _ForwardIterator1 __las
         {
             if (!__comp(*__first2, *__first1))
             {
-                ::new (::std::addressof(*__result)) _Tp(*__first1);
+                ::new (::std::addressof(*__result)) _Tp(*__select_copy(__first1, __first2));
```
mmichel11, line 191: Since there are just two cases, would it be clearer to just pass a `std::true_type`, `std::false_type`, or just a boolean as a template parameter for whether we copy from the first or second iterator instead of taking a lambda argument?

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3422,8 +3422,9 @@ __pattern_set_intersection(_ExecutionPolicy&& __exec, _RandomAccessIterator1 __f
             [](_DifferenceType __n, _DifferenceType __m) { return ::std::min(__n, __m); },
             [](_RandomAccessIterator1 __first1, _RandomAccessIterator1 __last1, _RandomAccessIterator2 __first2,
                _RandomAccessIterator2 __last2, _T* __result, _Compare __comp) {
-                return oneapi::dpl::__utils::__set_intersection_construct(__first1, __last1, __first2, __last2,
-                                                                          __result, __comp);
+                return oneapi::dpl::__utils::__set_intersection_construct(
+                    __first1, __last1, __first2, __last2, __result, __comp,
+                    [](auto&& a, auto&& b) { return std::forward<decltype(a)>(a); }); //copy first to output
```
mmichel11, line 3427: `std::forward` -> `::std::forward` in this return and the one below.

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: I considered this, but I thought this may be more descriptive / less disruptive to the code.  

Do you think it would be better to have `if constexpr` in the body of `_set_intersection_construct`? and just a 
`/*CopyFromFirstSet = */ ::std::true_type` here?

I don't have a strong opinion either way.

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: thanks, fixed.

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: I looked at this a little more. Since there is no ternary constexpr, its ugly and replicates code to use `if constexpr` in the body in this context.  You'd need to make a pair of copy selection function helpers which would end up pretty similar to this version with lambdas.

SergeyKopienko, line 191: As for me we able to make this line better (to increase performance).
For example I mean we may take a look to another place in our code:
```
        auto __move_values = [](_RandomAccessIterator __x, _Tp* __z) {
            if constexpr (::std::is_trivial_v<_Tp>)
                *__z = ::std::move(*__x);
            else
                ::new (::std::addressof(*__z)) _Tp(::std::move(*__x));
        };
```

mmichel11: Okay, I think the current way you've implemented is best then.

danhoeflinger: I don't think we have any guarantees that the selected iterator's `value_type` matches the `value_type` of `__result`.  I think we only know require that it can be converted to the `value_type` of result. 

Also, I don't think we should use `std::move` on input data.  This would move instead of copy from the input set to the output set, which changes the semantics of `std::set_intersection`.

If we adapt this code by removing `::std::move()` in both locations, I think it could fit.  However, would that improve performance to assign for trivial types rather than copy construct?

test/parallel_api/algorithm/alg.sorting/alg.set.operations/set.pass.cpp
```diff
@@ -125,6 +125,20 @@ struct test_set_union
...
 
+// Compare the first of a tuple using the supplied comparator
+template <typename _Comp>
+struct comp_select_first
+{
+    _Comp comp;
+    comp_select_first(_Comp __comp) : comp(__comp) {}
+    template <typename _T1, typename _T2>
+    bool
+    operator()(_T1 t1, _T2 t2) const
```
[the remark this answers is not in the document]
danhoeflinger: Thanks, fixed.

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: Sure,  I can add it, but without the `std::move`.

danhoeflinger: I found a decent way to do this in `_set_intersection_construct` which seems slightly better.

danhoeflinger: I've added this.  Is this what you had in mind?

mmichel11: This new solution seems the best to me.

SergeyKopienko: Yes. thank you!

danhoeflinger:
> I'm investigating the failure for windows, cl, but I'm thus far unable to reproduce it...

I'm still unable to reproduce the failure in practice.  However, examining the code a bit more, I think that our usage of `::new (::std::addressof(*__result)) _Tp(*(__select_element(__first1, __first2)));` is generally incorrect when `__result` can be a `zip_iterator`.  

For a zip_iterator, `operator*()` returns a rvalue `oneapi::dpl::__internal::tuple` of references to each zipped element.  The `addressof` of an rvalue should be deleted.  Its surprizing to me that this new test case works at all in hindsight.

It seems that we need a specialization for zip_iterator which results in a placement `::new` uninitialized copy for each element in the tuple, rather than the tuple as a whole.  I've prototyped this and I'm looking to add this.  

I believe there are other places in the code which will also require similar changes.  Probably 

In `memory_impl.h`, we have many similar usages of input iterators with `::std::addressof` for `uninitialized_copy`, `uinitialized_move`, `uninitialized_fill`, `uninitialized_default_construct`, and `__op_uninitialized_value_construct`.  

Also, we have similar usage in `__pattern_inplace_merge` which will likely need adjustment.

Unless I am missing something, all of these are affected and may be broken for zip_iterators.

danhoeflinger:
> I'm still unable to reproduce the failure in practice. However, examining the code a bit more, I think that our usage of `::new (::std::addressof(*__result)) _Tp(*(__select_element(__first1, __first2)));` is generally incorrect when `__result` can be a `zip_iterator`.

I was missing something in `__parallel_set_op` where it creates a temporary `__par_backend::__buffer` of the output `value_type` to write to, before using  `__brick_move_destroy` to move it to the final result (`zip_iterator`).  This means that this instance of `std::addressof` isn't invalid because we aren't dealing with a `zip_iterator`, but rather a buffer of `value_type` tuples.

I am still investigating the source of the error for windows-2019, cl and tbb.

(Also the general issue of `std::addressof` is still worth investigating and I will make an issue for it.)

danhoeflinger:
After rebasing with current main, it seems the failure for windows-2019, cl and tbb is no longer there.  I think we can go ahead and re-review the PR and consider merging.

mmichel11 (approved):
I have looked through both the patch and testing changes and they LGTM.

danhoeflinger:
@SergeyKopienko (since you have left feedback already) and others: 
I plan to merge ~April 16th with the existing approval if there are no objections.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3549,8 +3549,10 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                 __result, __comp, [](_DifferenceType __n, _DifferenceType __m) { return ::std::min(__n, __m); },
                 [](_RandomAccessIterator1 __first1, _RandomAccessIterator1 __last1, _RandomAccessIterator2 __first2,
                    _RandomAccessIterator2 __last2, _T* __result, _Compare __comp) {
-                    return oneapi::dpl::__utils::__set_intersection_construct(__first1, __last1, __first2, __last2,
-                                                                              __result, __comp);
+                    return oneapi::dpl::__utils::__set_intersection_construct(
+                        __first1, __last1, __first2, __last2, __result, __comp,
+                        oneapi::dpl::__internal::__op_uninitialized_copy<_ExecutionPolicy>{},
+                        /*CopyFromFirstSet = */ ::std::true_type());
```
SergeyKopienko, line 3555: I think `std::true_type{}` variant is more useful in our code.

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -186,7 +188,14 @@ __set_intersection_construct(_ForwardIterator1 __first1, _ForwardIterator1 __las
         {
             if (!__comp(*__first2, *__first1))
             {
-                ::new (::std::addressof(*__result)) _Tp(*__first1);
+
+                auto __select_element = [](auto&& _f1, auto&& _f2) {
```
SergeyKopienko, line 192: As far as I understood this code, `_f1' and `_f2` are iterators. So why we using `std::forward` here?
Moreover, may be `__select_element` lambda not required at all?

SergeyKopienko: What about `_copy(*(_CopyFromFirstSet::value ? _f1 : _f2), *__result);`

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: thanks, applied.

SergeyKopienko (approved):
LGTM

julianmi (approved):
LGTM

danhoeflinger:
Thanks all for the reviews.  With merging with the existing approvals.

PR #991, opened 2023-06-20
Use ND-range kernel in parallel_for kernel implementation to improve performance 

mmichel11:
Performance issues have been identified within our `parallel_for` kernel on Intel® Data Center GPU Max for input sizes that are not powers of two. The root cause of this has been identified as the usage of a SYCL basic parallel kernel in our implementation.

This PR switches to using an explicit nd-range kernel for GPU devices only within the `parallel_for` pattern. A grid-strided memory access pattern is used which successfully resolves the identified performance issues. Additionally, performance improvements have been observed for powers of 2 as well.

Algorithms that internally rely on this kernel should see improvements on GPUs (ex: `for_each`, `copy`, etc.).

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -255,43 +255,23 @@ struct __parallel_for_submitter</*_IsGPU=*/::std::true_type, __internal::__optio
     {
         assert(oneapi::dpl::__ranges::__get_first_range_size(__rngs...) > 0);
         _PRINT_INFO_IN_DEBUG_MODE(__exec);
-        constexpr ::uint16_t __vals_per_witem = 8;
         ::std::size_t __wgroup_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
-        ::std::size_t __ngroups = oneapi::dpl::__internal::__dpl_ceiling_div(__count, __vals_per_witem * __wgroup_size);
+        ::std::size_t __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__count, __wgroup_size);
+        ::std::size_t __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
+        __n_groups = ::std::min(__n_groups, __max_cu);
         auto __event = __exec.queue().submit(
-            [&__rngs..., &__brick, __count, __vals_per_witem, __wgroup_size, __ngroups](sycl::handler& __cgh) {
+            [&__rngs..., &__brick, __count, __wgroup_size, __n_groups, __max_cu](sycl::handler& __cgh) {
                 //get an access to data under SYCL buffer:
                 oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
 
                 __cgh.parallel_for<_Name...>(
-                    sycl::nd_range</*dim=*/1>(__ngroups * __wgroup_size, __wgroup_size),
+                    sycl::nd_range</*dim=*/1>(__n_groups * __wgroup_size, __wgroup_size),
                     [=](sycl::nd_item</*dim=*/1> __item_id) {
-                        auto __stride = __item_id.get_sub_group().get_local_range();
-                        auto __sg_local_id = __item_id.get_sub_group().get_local_id();
-                        auto __sg_id = __item_id.get_sub_group().get_group_id();
                         auto __local_id = __item_id.get_local_id(0);
-                        auto __group_id = __item_id.get_group(0);
-                        auto __start_idx = __vals_per_witem * (__group_id * __wgroup_size + __sg_id * __stride);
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -242,6 +245,38 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+template <typename... _Name>
+struct __parallel_for_submitter</*_IsGPU=*/::std::true_type, __internal::__optional_kernel_name<_Name...>>
+{
+    // GPU specialization with strided access memory pattern
+    template <typename _ExecutionPolicy, typename _Fp, typename _Index, typename... _Ranges>
+    auto
+    operator()(_ExecutionPolicy&& __exec, _Fp __brick, _Index __count, _Ranges&&... __rngs) const
+    {
+        assert(oneapi::dpl::__ranges::__get_first_range_size(__rngs...) > 0);
+        _PRINT_INFO_IN_DEBUG_MODE(__exec);
```
julianmi, line 257: Should we move these debug statements into the calling `__parallel_for` to not have to duplicate them?

julianmi (commented):
Looks good to me. As discussed offline, I think it makes sense to investigate whether other algorithms suffer from these performance limitations as well.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -210,18 +210,21 @@ class __scan_single_wg_dynamic_kernel;
...
 // Use the trick with incomplete type and partial specialization to deduce the kernel name
 // as the parameter pack that can be empty (for unnamed kernels) or contain exactly one
 // type (for explicitly specified name by the user)
-template <typename _KernelName>
+template <typename _IsGPU, typename _KernelName>
 struct __parallel_for_submitter;
 
 template <typename... _Name>
-struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
+struct __parallel_for_submitter</*_IsGPU=*/::std::false_type, __internal::__optional_kernel_name<_Name...>>
```
danhoeflinger, line 227: For maintainability, it would be great to just have a single implementation which has different configuration settings for GPU and CPU targets. Is this possible to do with configuration settings and the same underlying `nd_range` implementation?

mmichel11: I think this is a good idea for maintainability. I will look into and see if I can find a good way to unite the two implementations with a single `nd_range` kernel.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Yes, I will do this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I have looked into this a bit more and ran some benchmarks on SPR with some different nd range kernels. I was able to tune the kernel to perform slightly better than the SYCL basic kernel by changing grain sizes, but I suppose this might be specific to SPR with the input type I used (fp32).

I think it may be the simplest and most generic solution to just leave the SYCL basic kernel for CPU devices as is and have the GPU specialization to ensure memory reads / writes are coalesced.

mmichel11:
I am marking this PR as draft for now. I have seen substantial performance improvements with basic SYCL kernels in our parallel for pattern with recent compiler / driver updates so this patch may no longer be needed. I will do some wider evaluation on this. Ideally, keeping this kernel as generic as possible is best to reduce the need for future tuning with new architectures / devices.

mmichel11:
I have looked back at this PR again to see if it is something that should still be pursued since it is marked as priority.

After further thought, I think this PR can be closed. I still observe that we can get improvements in performance of this pattern by tuning, but I think it will ultimately end up being overtuned for certain architectures and operators / data types.  From my testing, we would also have to differentiate between vendor backends to get the best performance on different vendors. This would ultimately add more complexity then it is worth.

Parallel for is a very simple pattern, so if we find poor performance with the basic kernels, then I think we should likely push for improvements within the SYCL basic kernel implementation which I have seen happen. The performance issue that motivated this PR (massive runtime spikes for random non-power of 2 input sizes) no longer exists.

Are there any objections to me closing this PR? @akukanov @julianmi @danhoeflinger

danhoeflinger:
> Are there any objections to me closing this PR? @akukanov @julianmi @danhoeflinger

No objections here.  I think your argument is sound.  Its best if the we can rely on simple code for such a simple pattern.  The SYCL implementation should be able to run this efficiently on any hardware (and if it doesn't we should pursue the problem there in the long run).  As long as we aren't sacrificing huge performance benefits, I'm fine with closing it.

PR #996, opened 2023-06-23
[oneDPL][doc] + description for 'sort_by_key' algo

MikeDvorskiy:
Added description for 'sort_by_key' algorithm (extension)

### DIFF
--- a/documentation/library_guide/parallel_api/additional_algorithms.rst
+++ b/documentation/library_guide/parallel_api/additional_algorithms.rst
@@ -76,10 +76,10 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
     search sequence: [0, 2, 4, 7, 6]
     result sequence: [1, 4, 8, 10, 10]
 
-* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts sequence's keys according to 
-  a comparioson operator. If no comparator is provided, then a function object that uses ``operator<``
-  to compare the elements is used. The sequence's values are permutated according to the sorted sequence's keys.  
-  For correct computation, a size of number of a key sequence and a size of a value sequence should be the same.
+* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts the sequence's keys according to 
+  a comparioson operator. If no comparator is provided, then the elements are compared with ``operator<``.
+  The sequence's values are permutated according to the sorted sequence's keys. The prerequisite for correct
+  behavior is that the size for both keys sequence and values sequence shall be the same.  
   For example::
 
     keys:   [3,    5,   0,   4,   3,   0]


documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -74,4 +74,15 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
 
     input sequence:  [0, 2, 2, 2, 3, 3, 3, 3, 6, 6]
     search sequence: [0, 2, 4, 7, 6]
-    result sequence: [1, 4, 8, 10, 10]
\ No newline at end of file
+    result sequence: [1, 4, 8, 10, 10]
+
+* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts sequence's keys according to
```
dcbenito, line 79: My suggestions:
``sort_by_key``: performs a stable key-value sort. The algorithm sorts the sequence's keys according to a comparison operator. If no comparator is provided, then a function object that uses ``operator<`` to compare the elements is used. The sequence's values are permutated according to the sorted sequence's keys. For correct computation, the number size of a key sequence and the size of a value sequence should be the same. For example::

dcbenito (dismissed):

documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -74,4 +74,15 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
...
     input sequence:  [0, 2, 2, 2, 3, 3, 3, 3, 6, 6]
     search sequence: [0, 2, 4, 7, 6]
-    result sequence: [1, 4, 8, 10, 10]
\ No newline at end of file
+    result sequence: [1, 4, 8, 10, 10]
+
+* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts sequence's keys according to 
+  a comparioson operator. If no comparator is provided, then a function object that uses ``operator<``
+  to compare the elements is used. The sequence's values are permutated according to the sorted sequence's keys.  
+  For correct computation, a size of number of a key sequence and a size of a value sequence should be the same.
```
rarutyun, line 82: This sentence is hard to read and doesn't have strong enough statements. I wouldn't also document it with Undefined Behavior term because it's more specification term rather than documentation one.

I would say something like:

"The prerequisite for correct behavior is that the size for both keys sequence and values sequence shall be the same."
The wording of the sentence above might be further polished.

dcbenito: I think both suggestions are good. Does this work for all?

sort_by_key: performs a stable key-value sort. The algorithm sorts the sequence's keys according to a comparison operator. If no comparator is provided, then the elements are compared with ``operator<``. The sequence's values are permutated according to the sorted sequence's keys. The size for key and sequence values must be the same. For example::

dcbenito: Thank you for the explanation, let's go with your original sentence: "The prerequisite for correct behavior is that the size for both keys sequence and values sequence shall be the same."

timmiesmith: I think I've captured the update needed for this change.

rarutyun (approved):

documentation/library_guide/parallel_api/additional_algorithms.rst
rarutyun: I believe we iterate over the wording several times, so after the latest update we are good to go

rarutyun (approved):

PR #997, opened 2023-06-24
[KT] Pass number of stages to histogram kernel

dmitriy-sobolev:
Let's calculate the number of stages only once.


SergeyKopienko:
Why so many differences? Why we have new (added) files in this PR ?

dmitriy-sobolev:
> Why so many differences? Why we have new (added) files in this PR ?

I guess it happened after rebasing of the base branch ([dev/dmitriy-sobolev/esimd-radix-sort](https://github.com/oneapi-src/oneDPL/tree/dev/dmitriy-sobolev/esimd-radix-sort)). I will rebase this one to fix it.

SergeyKopienko (approved):
LGTM

PR #999, opened 2023-06-26
[oneDPL] radix sort: one work group optimization: counters to SM

MikeDvorskiy:
[oneDPL] radix sort: one work group optimization: move counters to SLM for 4K..16K.
When SLM is not enough for counters + temporary buffer, we keep SLM just for the counters, for the values - a buffer in the global memory.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
@@ -48,12 +48,15 @@ struct __subgroup_radix_sort
 
         using _KeyT = oneapi::dpl::__internal::__value_t<_RangeIn>;
         //check SLM size
-        if (__check_slm_size<_KeyT>(__q, __src.size()))
+        const auto __SLM_available = __check_slm_size<_KeyT>(__q, __src.size());
+        if (__SLM_available.first && __SLM_available.second)
             return __one_group_submitter<_SortKernelLoc>()(__q, ::std::forward<_RangeIn>(__src), __proj,
-                                                           std::true_type{} /*SLM*/);
-        else
+                                                           ::std::true_type{} /*SLM*/, ::std::true_type{} /*SLM*/);
+        if (__SLM_available.second)
             return __one_group_submitter<_SortKernelGlob>()(__q, ::std::forward<_RangeIn>(__src), __proj,
-                                                            std::false_type{} /*No SLM*/);
+                                                            ::std::false_type{} /*No SLM*/, ::std::true_type{} /*SLM*/);
+        return __one_group_submitter<_SortKernelGlob>()(__q, ::std::forward<_RangeIn>(__src), __proj,
+                                                            ::std::false_type{} /*No SLM*/, ::std::false_type{} /*No SLM*/);
     }
 
   private:
@@ -61,7 +64,7 @@ struct __subgroup_radix_sort
     class _TempBuf;
 
     template <typename _KeyT>
-    class _TempBuf<_KeyT, std::true_type /*shared local memory buffer*/>
+    class _TempBuf<_KeyT, ::std::true_type /*shared local memory buffer*/>
     {
         uint16_t __buf_size;
 
@@ -75,7 +78,7 @@ struct __subgroup_radix_sort
     };
 
     template <typename _KeyT>
-    class _TempBuf<_KeyT, std::false_type /*global memory buffer*/>
+    class _TempBuf<_KeyT, ::std::false_type /*global memory buffer*/>
     {
         sycl::buffer<_KeyT> __buf;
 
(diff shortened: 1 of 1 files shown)


dmitriy-sobolev (commented):
I think a more robust solution would be to check the available amount of SLM separately:
 1. Check SLM for the counters. Use SLM only if the desired amount is available.
 2. Check if the remaining amount of SLM is sufficient for the output buffer.

So, you would need two SLM tags. What do you think?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -111,15 +111,16 @@ struct __subgroup_radix_sort
     {
         assert(__n <= 1 << 16); //the kernel is designed for data size <= 64K
 
+        const auto __req_slm_size_counters = __counter_buf_sz * sizeof(uint32_t);
```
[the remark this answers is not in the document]
MikeDvorskiy: `__req_slm_size_counters` is size in bytes.

danhoeflinger (commented):
This change assumes that the counters always fit into SLM, without a backup plan.

For radix of 4, and a wg_size ranging from 64 to 512 depending on the input size, and 32 bit counts, this is a maximum of slightly more than 32KB (with the extra element).  

I am not sure if this is safe to assume, but we could have a third option which uses no SLM if even this does not fit.  It would just require augmenting the return of `__check_slm_size` to return a pair of bools, then utilizing it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -111,15 +111,16 @@ struct __subgroup_radix_sort
...
+
         // Pessimistically only use half of the memory to take into account memory used by compiled kernel
```
danhoeflinger, line 116: While we are here, can we make this change to be more specific?

```suggestion
        // Pessimistically only use half of the memory to take into account memory used by group scan algorithm
```

MikeDvorskiy: I am confused - why "scan algorithm"?

danhoeflinger: I believe the reason we are using only half the SLM is because of the use of `__exclusive_scan_over_group`.  Is it correct, or from some other reason?

This is not necessarily relevant to this PR, but the comment confused me as I was investigating it.

MikeDvorskiy: Unfortunately, we could not understand what is true reason... and just reduced SLM size in 2 times...
Regarding to "group scan" - I am not sure that is a reason, because "group_scan" is launched here on the private memory, not SLM.

danhoeflinger: I believe `__exclusive_scan_over_group` uses some amount of SLM temporary data, and there is not a way to query how much.  In other places of oneDPL, we have reduced the amount of SLM by half to account for this for both the `...over_group` and `joint_...` with `scan` and `reduce`.  @adamfidel @julianmi  perhaps you can confirm this or provide some more information here?

MikeDvorskiy: In that case a more clear comment should be look like that:
`// Pessimistically only use half of the memory to take into account a SYCL group algorithm might use a portion of SLM`

danhoeflinger: Yes this is better than my suggestion.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -106,20 +109,25 @@ struct __subgroup_radix_sort
...
     __check_slm_size(sycl::queue __q, _Size __n)
     {
         assert(__n <= 1 << 16); //the kernel is designed for data size <= 64K
 
+        const auto __req_slm_size_counters = __counter_buf_sz * sizeof(uint32_t);
+
         // Pessimistically only use half of the memory to take into account memory used by compiled kernel
         const ::std::size_t __max_slm_size =
-            __q.get_device().template get_info<sycl::info::device::local_mem_size>() / 2;
+            (__q.get_device().template get_info<sycl::info::device::local_mem_size>() - __req_slm_size_counters) / 2;
```
danhoeflinger, line 121: I don't think we want to remove `__req_slm_size_counters` here.  We are incorporating that size below in the conditionals instead (lines 126, 128).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -48,12 +48,15 @@ struct __subgroup_radix_sort
 
         using _KeyT = oneapi::dpl::__internal::__value_t<_RangeIn>;
         //check SLM size
-        if (__check_slm_size<_KeyT>(__q, __src.size()))
+        const auto __SLM_available = __check_slm_size<_KeyT>(__q, __src.size());
+        if (__SLM_available.first && __SLM_available.second)
             return __one_group_submitter<_SortKernelLoc>()(__q, ::std::forward<_RangeIn>(__src), __proj,
-                                                           std::true_type{} /*SLM*/);
-        else
+                                                           std::true_type{} /*SLM*/, std::true_type{} /*SLM*/);
```
danhoeflinger, line 54: `std` -> `::std` in here and in a few more places in the next few lines of code.

danhoeflinger (commented):
A couple minor things to fix, but this resolves my main issue at a high level.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
MikeDvorskiy: dont catch....

We consider the total SLM size as `device::local_mem_size/2`.. right?
After that we checks is it enough for values and counters or just for counters... or  - not for both

danhoeflinger:
```suggestion
            __q.get_device().template get_info<sycl::info::device::local_mem_size>() / 2;
```


Further explanation:

` const ::std::size_t __max_slm_size = (__q.get_device().template get_info<sycl::info::device::local_mem_size>() - __req_slm_size_counters) / 2;`
gets the max SLM size, subtracts the calculated required size the counters would require, and then divides by 2.  I think instead it should just get the max SLM size, and divide by 2.

`if(__req_slm_size_val + __req_slm_size_counters <= __max_slm_size)` 
Checks if both the values and the counters fit into max memory (but as written in the PR, the counter size has already been removed).

`if(__req_slm_size_counters <= __max_slm_size)`
Checks if only the counters fit into max memory (but as written in the PR, the counter size has already been removed).

MikeDvorskiy: fixed

danhoeflinger (approved):
LGTM

PR #1001, opened 2023-06-28
[KT] Get rid of excessive code in onesweep

dmitriy-sobolev:
Measurements show that this does not affect performance, at least on Intel® Data Center GPU Max. So the changes might be treated as refactoring.

Nontrivial changes:
- `ResetBinCounters` is removed. `RankSLM` does the same, but without using `fence` which is unnecessary as well.


dmitriy-sobolev:
This will be handled as a part of #1024

PR #992, opened 2023-06-21
Unify exit message in the tests

dmitriy-sobolev:
Let's unify the output of the tests for easier handling with scripts.

PR #1002, opened 2023-06-28
[Tests] Fix xpu tests (regression since #992)

dmitriy-sobolev:
It fixes "error: no member named 'done' in namespace 'TestUtils'". The error is introduced by #992.


mmichel11 (approved):
LGTM. I have also verified that the three unchanged files from PR #992 already include utils.h

SergeyKopienko (approved):
LGTM

PR #1003, opened 2023-06-29
[KT] Test 8-bit types

dmitriy-sobolev:
Additional changes: 
 - Thin out types to test due to long compilation time
 - Get rid of 'k' suffix in the names of some constants

Unfortunately, the testing has shown that onesweep does not work with 8-bit types.


SergeyKopienko (approved):
LGTM

PR #1004, opened 2023-06-30
[oneDPL] radix sort optimization: + skip iteration if is possible

MikeDvorskiy:
Radix sort optimization:
1. Checks statistic on count/scan phase.
2. If there is no variety (the all keys are the same,), don't  start  re-order phase, do just copying to the opposite buffer.


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -616,15 +661,20 @@ struct __parallel_radix_sort_iteration
...
-            __exec, __segments, __block_size, __radix_offset, __in_rng, __tmp_buf, __dependency_event, __proj
+            __exec, __segments, __block_size, __radix_offset, __in_rng, __tmp_buf,
+            __dependency_event, __proj
 #if _ONEDPL_COMPILE_KERNEL
             , __count_kernel
 #endif
         );
 
+        //A flag to specify "no re-order execution" if data are the same on the current sort iteration.
+        uint16_t __no_op = 0;
```
dmitriy-sobolev, line 672: Let's make it aligned with the rest of the file using fixed-width integral types from `::std` namespace. Also, should the type be changed to `::std::uint8_t` or `bool`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -467,6 +502,15 @@ __radix_sort_reorder_submit(_ExecutionPolicy&& __exec, ::std::size_t __segments,
             __kernel,
 #endif
             sycl::nd_range<1>(__segments * __sg_size, __sg_size), [=](sycl::nd_item<1> __self_item) {
+
+                //Optimization: skip re-order phase if the all keys are the same, do just copying
+                if(__no_op_acc[0])
+                {
+                    __copy_kernel_for_radix_sort(__segments, __blocks_per_segment * __block_size, __sg_size, __self_item,
```
dmitriy-sobolev, line 509: Have you considered avoiding copying at all? This would require adding some logic in case of odd number of such skipped copies, though.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
MikeDvorskiy: probably, it makes sense

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -728,9 +778,6 @@ __parallel_radix_sort(_ExecutionPolicy&& __exec, _Range&& __in_rng, _Proj __proj
     else if (__n <= 16384 && __wg_size * 8 <= __max_wg_size)
         __event = __subgroup_radix_sort<_RadixSortKernel, __wg_size * 8, 32, __radix_bits, __is_ascending>{}(
             __exec.queue(), ::std::forward<_Range>(__in_rng), __proj);
-    else if (__n <= 32768 && __wg_size * 16 <= __max_wg_size)
-        __event = __subgroup_radix_sort<_RadixSortKernel, __wg_size * 16, 32, __radix_bits, __is_ascending>{}(
-            __exec.queue(), ::std::forward<_Range>(__in_rng), __proj);
```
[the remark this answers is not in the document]
MikeDvorskiy: it is a kind of merge artifact...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
dmitriy-sobolev: I believe there is no need in barriers, but yes, it will result in extra overheads because of:
- Need to support an additional flag to indicate if the source and destination buffers must be swapped. It will add an extra global memory read and write in `scan` kernel.
- Need to support logic which copies the data into the main buffer if it is the last iteration and the data is in the temporary buffer. This will add 2 extra if conditions into `reorder` kernel (checking the new flag before potential swapping, checking if it is the last stage), swap operation, and an additional read of the new flag.

Anyway, thanks for the investigation. If necessary, it can be further investigated in a separate PR.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -616,15 +661,20 @@ struct __parallel_radix_sort_iteration
...
+            __exec, __segments, __block_size, __radix_offset, __in_rng, __tmp_buf,
+            __dependency_event, __proj
 #if _ONEDPL_COMPILE_KERNEL
             , __count_kernel
 #endif
         );
 
+        //A flag to specify "no re-order execution" if data are the same on the current sort iteration.
+        uint16_t __no_op = 0;
+        sycl::buffer<uint16_t, 1> __no_op_flag(&__no_op, 1);
```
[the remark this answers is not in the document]
dmitriy-sobolev: There is probably not that much value in optimizing that piece of code. So you can close this conversation.

I hope that the compiler is clever enough to interleave the allocations (or cache them) and copying a single element onto a device with the `count` and `scan` kernels in the current implementation. But, it can always be verified with a profiler.

But let me add some comment:
1. You can profile the code, e.g. with [onetrace](https://github.com/intel/pti-gpu/tree/master/tools/onetrace) to make sure that there are no extra allocations.  
2. I do not see any need in barriers. You only need to fill the buffer with a single work-item (there is no cooperation). You may also call `sycl::queue::fill`: that operation may be done in parallel with `count` and `scan` kernels; there is no need to copy data between host and device with that approach.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
dmitriy-sobolev: Actually, we would need a global synchronization if we perform copying at the end of `reorder` kernel when the data is not in the main buffer after the last sort iteration. Thus it must be done in a separate kernel. So yes, it is difficult to implement, I agree with the current approach.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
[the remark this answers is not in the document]
MikeDvorskiy: `sycl::queue::fill` leads to additional submit... I believe it will be a reason of bigger overhead...

`Equivalent to submitting a command-group containing handler::fill(ptr, pattern, count).`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -274,6 +276,7 @@ __radix_sort_scan_submit(_ExecutionPolicy&& __exec, ::std::size_t __scan_wg_size
         __hdl.depends_on(__dependency_event);
         // access the counters for all work groups
         oneapi::dpl::__ranges::__require_access(__hdl, __count_rng);
+        sycl::accessor __no_op_acc(__no_op_flag, __hdl, sycl::write_only);
```
danhoeflinger, line 279: I think you should be able to avoid a copy to the device of this flag by adding `sycl::no_init`, and adjusting the if statement in the kernel to have an `else` which assigns it to false (but you'll have to separate the selection of the work item and the real conditional).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -287,6 +296,13 @@ __radix_sort_scan_submit(_ExecutionPolicy&& __exec, ::std::size_t __scan_wg_size
                 // TODO: consider another approach with use of local memory
                 __dpl_sycl::__joint_exclusive_scan(__self_item.get_group(), __begin, __begin + __scan_size, __begin,
                                                    _CountT(0), __dpl_sycl::__plus<_CountT>{});
+                const auto __wi = __self_item.get_local_linear_id();
+                //That condition may be truth (by algo semantic) just on one WG, one WI, so there is no race here.
+                if(__wi == __scan_wg_size - 1 && *(__begin + __scan_size - 1) == __n)
+                {
+                    auto& __no_op_flag = __count_rng[__no_op_flag_idx];
+                    __no_op_flag = 1; //set flag if the all values got into one bin
```
dmitriy-sobolev, line 304:
```suggestion
                    __count_rng[__no_op_flag_idx]= 1; //set flag if the all values got into one bin
```

dmitriy-sobolev (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -232,6 +233,13 @@ __radix_sort_count_submit(_ExecutionPolicy&& __exec, ::std::size_t __segments, :
                     // thus splitting __count_rng into __radix_states regions
                     __count_rng[(__segments + 1) * __self_lidx + __wgroup_idx] = __count_lacc[__self_lidx];
                 }
+
+                // 3.0 side work: reset 'no operation flag', which specifies whether to skip re-order phase
+                if(__wgroup_idx == 0 && __self_lidx == 0)
```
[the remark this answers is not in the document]
MikeDvorskiy: In my opinion my variant is more understandable - zero group, zero wi does this action.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -232,6 +233,13 @@ __radix_sort_count_submit(_ExecutionPolicy&& __exec, ::std::size_t __segments, :
...
+                {
+                    auto& __no_op_flag = __count_rng[__no_op_flag_idx];
+                    __no_op_flag = 0;
```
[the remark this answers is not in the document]
MikeDvorskiy: 1. I specially added additional reference to make that place is more readable and self-documented.
2. A compiler does optimization here, of course. So, there is not an additional variable in real code.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
MikeDvorskiy: That code and comment is not actual any more.

danhoeflinger: yep, thanks.

danhoeflinger (approved):
I believe some of the clang-format suggestions are valid and should be taken.  Otherwise, LGTM

danhoeflinger (approved):

PR #1005, opened 2023-07-01
[KT] Fix sort identity for floating point numbers

dmitriy-sobolev:
Additionally, fulfil TODOs for `__full_sort_identity`.

PR #1006, opened 2023-07-02
[KT] Enable support of 64-bit types in histogram kernel

dmitriy-sobolev:
Additional changes:
 - Add a test for the histogram kernel. It can be removed when the kernel is proven to be working with other parameters changed, e.g. `RADIX_BITS`, `WORK_GROUPS` and `WORK_GROUP_SIZE`.
 - Rename variables in histogram kernel for better readability.
 - Replace `slm_block_load` with `utils::BlockLoad` due to `slm_block_load` limitation (`N * sizeof(Tx) must be 1, 2, 4 or 8 owords long` i.e. it cannot be safely used to load more than 8 * 16 = 128 bytes or 32 `uint32_t` elements).

Currently, the test fails to correctly calculate a histogram for `double` type elements, but it is fixed in #1005.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
@@ -102,94 +102,110 @@ __full_sort_identity()
     return utils::__sort_identity<T, __is_ascending>;
 }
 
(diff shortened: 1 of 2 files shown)


SergeyKopienko (approved):
LGTM

PR #1007, opened 2023-07-02
[KT] Enable support of 8-bit types in onesweep

dmitriy-sobolev:
Make sure the sorted data is in the input range upon work completion


SergeyKopienko (approved):
LGTM

PR #1008, opened 2023-07-03
[KT] Enable all types (excepting double) in esimd radix sort test

SergeyKopienko:
In this PR we define by default PROCESS_SIZE for ```onesweep``` impl in esimd radix sort as:

- ```384``` for all types with size <= 4 bytes;
- ```192``` for all types with size == 8 bytes;
- enable all types in esimd radix sort test: ```double``` will be included in separate PR.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
@@ -26,6 +26,19 @@
 
 namespace oneapi::dpl::experimental::esimd::impl
 {
+template <typename T, ::std::enable_if_t<sizeof(T) <= sizeof(::std::uint32_t), int> = 0>
+constexpr ::std::uint32_t
+__onesweep_process_size()
+{
+    return 384;
+}
+
+template <typename T, ::std::enable_if_t<sizeof(T) == sizeof(::std::uint64_t), int> = 0>
+constexpr ::std::uint32_t
+__onesweep_process_size()
+{
+    return 192;
+}
 
 /*
     Interface:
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -74,7 +87,7 @@ radix_sort(_ExecutionPolicy&& __exec, _Range&& __rng)
         // TODO: avoid kernel duplication (generate the output storage with the same type as input storage and use swap)
         // TODO: pass _ProcessSize according to DataPerWorkItem
```
dmitriy-sobolev, line 88: It might require more time, but what do you think about fulfilling this TODO and passing a desired `DataPerWorkItem` from within `radix_sort_esimd.pass.cpp`?

SergeyKopienko: At current moment we have ```DataPerWorkItem``` defined as ```constexpr ::std::uint16_t DataPerWorkItem = 16;```
So we haven't any dependencies at this place now.

dmitriy-sobolev (commented):
There are code sections such:
```c++
  if constexpr (sizeof(T) <= sizeof(::std::uint32_t))
  {
      test_sycl_iterators<T, Ascending>(size);
      test_sycl_iterators<T, Descending>(size);
  }
  else
  {
      // TODO required to implement
  }
```
I guess, such if-else conditions should be removed.

SergeyKopienko:
> There are code sections such:
> 
> ```c++
>   if constexpr (sizeof(T) <= sizeof(::std::uint32_t))
>   {
>       test_sycl_iterators<T, Ascending>(size);
>       test_sycl_iterators<T, Descending>(size);
>   }
>   else
>   {
>       // TODO required to implement
>   }
> ```
> 
> I guess, such if-else conditions should be removed.

fixed.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
dmitriy-sobolev: I see two ways:
 -  either leave a TODO comment,
 -  or modify the test therefore different values of `DataPerWorkItem` are used for single-work-group, cooperative and onesweep implementation. Whereas, it varies for onesweep depending on a data type.

dmitriy-sobolev: It seems there is a suitable TODO comment at (`TODO: pass _ProcessSize according to DataPerWorkItem`)

dmitriy-sobolev (approved):
LGTM.

However, it might be better to merge #1005 first and enable testing of `double` type here, in a single PR.

PR #1012, opened 2023-07-03
[KT] Using of TEST_LONG_RUN define in test

SergeyKopienko:
In this PR we using of ```TEST_LONG_RUN``` define in esimd radix sort test test instead of our previous define ```TEST_ALL_INPUTS```


dmitriy-sobolev (approved):
LGTM

PR #608, opened 2022-07-26
SYCL Rewrites of inclusive_scan_by_segment and exclusive_scan_by_segment

mmichel11:
# Original Description
To increase performance of segmented reduction and segmented scans, the SYCL backend implementations of the algorithms have been rewritten using the SYCL group and sub-group functions. 

The new algorithms operate in two primary phases:
1. Perform a segmented scan or reduction within each individual work-group with equal work assignment, storing partially completed scan / reduction values to global memory.
2. Apply the partial reduction / scan values from previous work-group(s) to the current.

# Update 1/18/2023
The scope of this PR has been reduced to only include the inclusive and exclusive segmented scan rewrites. The sycl rewrite for reduce by segment has been moved to a separate branch and will be worked on separately. This updated PR contains correctness fixes from its initial state along with expanded tests for the segmented scans.

## Algorithmic Descriptions
The segmented scans are implemented in a two kernel approach using the lower level SYCL libraries in contrast to the existing implementation utilizing internal parallel patterns. The segmented scans require the provided binary operator to be associative. Furthermore, the implementation is only invoked if the underlying datatype possesses a known identity element. Otherwise, the legacy implementation will be invoked.

The algorithms proceed as follows:
1. The first kernel is launched and each work group is assigned a range of elements to process equally divided among work items. Each work item is by default assigned four elements to process which has yielded the best performance empirically. Within a work group, the work items perform a segmented inclusive or exclusive scan writing the results to the global output buffer. The last work item outputs its final scan output to global memory along with a flag stating if the work group encountered a segment end.
2. The second kernel is launched upon the first kernel's completion. Each work group collectively searches the global segment end flag buffer within the range `[0, work group id - 1]` to find the work group with the closest id that contains a segment end while reducing the partial scan outputs in the range `[closest seg end id, work group id - 1]`. Once this prefix value is determined, work items will apply the prefix to the values stored in the global output buffer up until the first segment end in the work group (if it exists).

## Performance Results
Performance results show significant speedup over the existing segmented scan implementations for varying segment sizes. To obtain performance results of these new rewrites, please contact Matthew Michel for more details.

PR #1013, opened 2023-07-03
Do not forward or move variables that are used later in the code

timmiesmith:
Static analysis tools run since #862 and #608 were merged have identified more cases where variables were forwarded and then used later in the code.

### DIFF
--- a/include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
@@ -66,7 +66,7 @@ pattern_exclusive_scan_by_segment(Policy&& policy, InputIterator1 first1, InputI
     auto flags = _flags.get();
     flags[0] = 1;
 
-    transform(::std::forward<Policy>(policy), first1, last1 - 1, first1 + 1, _flags.get() + 1,
+    transform(policy, first1, last1 - 1, first1 + 1, _flags.get() + 1,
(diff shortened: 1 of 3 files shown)


timmiesmith:
Addressing the remaining issues reported by clang format I think would make the code less readable. Please let me know if you disagree and I can address them.

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
```diff
@@ -127,7 +127,7 @@ exclusive_scan_by_segment_impl(Policy&& policy, InputIterator1 first1, InputIter
         flags[0] = 1;
     }
 
-    transform(::std::forward<Policy>(policy), first1, last1 - 1, first1 + 1, _flags.get() + 1,
```
danhoeflinger, line 130: I think this is correct, but I also think we can add forwarding to the `policy2` creation @ line 154 below.

danhoeflinger (commented):
One minor fix, otherwise, LGTM.

mmichel11:
LGTM once Dan's comment is addressed.

include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
timmiesmith: Done.

danhoeflinger (approved):
LGTM

PR #1014, opened 2023-07-04
[oneDPL][async] fixed bug with reduce_async ignoring __binary_op parameter

lslusarczyk:
`reduce_async` function with `binary_op` parameter was ignoring the parameter and always used `std::plus`.
I exposed the issue by a test and made a fix.

Please merge this bugfix and include in next RC, as the issue fails distributed-ranges project code (see `inclusive_scan` implemented in https://github.com/oneapi-src/distributed-ranges/pull/349)

### DIFF
--- a/include/oneapi/dpl/internal/async_impl/glue_async_impl.h
+++ b/include/oneapi/dpl/internal/async_impl/glue_async_impl.h
@@ -131,7 +131,6 @@ auto
 reduce_async(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _Tp __init,
              _BinaryOperation __binary_op, _Events&&... __dependencies)
 {
-    using _ValueType = typename ::std::iterator_traits<_ForwardIterator>::value_type;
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/internal/async_impl/glue_async_impl.h
```diff
@@ -133,9 +133,9 @@ reduce_async(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterat
 {
     using _ValueType = typename ::std::iterator_traits<_ForwardIterator>::value_type;
```
MikeDvorskiy, line 134: As far as `_ValueType` is not used anymore, this line can be removed as well.

MikeDvorskiy (commented):
Overall look goods, excepting one minor comment.

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
lslusarczyk: good point, fixed - could you please approve now?

PR #1019, opened 2023-07-07
Add support for using zip iterators as input values in segmented algorithms

mmichel11:
This PR adds support for using zip iterators with segmented algorithms by remedying compiler errors related to constness of internal tuple elements when the scans / reductions are performed.

### DIFF
--- a/include/oneapi/dpl/internal/function.h
+++ b/include/oneapi/dpl/internal/function.h
@@ -134,7 +134,7 @@ struct segmented_scan_fun
     {
         using ::std::get;
         using x_t = ::std::tuple_element_t<0, _T1>;
-        x_t new_x = get<1>(y) ? get<0>(y) : binary_op(get<0>(x), get<0>(y));
+        auto new_x = get<1>(y) ? x_t(get<0>(y)) : x_t(binary_op(get<0>(x), get<0>(y)));
         auto new_y = get<1>(x) | get<1>(y);
         return _T1(new_x, new_y);
     }
--- a/include/oneapi/dpl/internal/function.h
+++ b/include/oneapi/dpl/internal/function.h
@@ -133,7 +133,8 @@ struct segmented_scan_fun
     operator()(const _T1& x, const _T2& y) const
     {
         using ::std::get;
-        auto new_x = get<1>(y) ? get<0>(y) : binary_op(get<0>(x), get<0>(y));
+        using x_t = ::std::tuple_element_t<0, _T1>;
+        auto new_x = get<1>(y) ? x_t(get<0>(y)) : x_t(binary_op(get<0>(x), get<0>(y)));
         auto new_y = get<1>(x) | get<1>(y);
         return _T1(new_x, new_y);
     }
--- a/include/oneapi/dpl/internal/function.h
+++ b/include/oneapi/dpl/internal/function.h
@@ -134,7 +134,7 @@ struct segmented_scan_fun
     {
         using ::std::get;
         using x_t = ::std::tuple_element_t<0, _T1>;
-        auto new_x = get<1>(y) ? x_t(get<0>(y)) : x_t(binary_op(get<0>(x), get<0>(y)));
+        x_t new_x = get<1>(y) ? get<0>(y) : binary_op(get<0>(x), get<0>(y));
         auto new_y = get<1>(x) | get<1>(y);
         return _T1(new_x, new_y);
     }
--- a/include/oneapi/dpl/pstl/hetero/utils_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/utils_hetero.h
@@ -169,6 +169,9 @@ struct __decay_with_tuple_specialization<::std::tuple<_Args...>>
     using type = ::std::tuple<::std::decay_t<_Args>...>;
 };
 
+template <typename... _Args>
+using __decay_with_tuple_specialization_t = typename __decay_with_tuple_specialization<_Args...>::type;
+
 } // namespace __internal
 } // namespace dpl
 } // namespace oneapi
--- a/include/oneapi/dpl/pstl/hetero/utils_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/utils_hetero.h
@@ -21,6 +21,8 @@
 #ifndef _ONEDPL_UTILS_HETERO_H
 #define _ONEDPL_UTILS_HETERO_H
 
+#include "../tuple_impl.h"
+
 namespace oneapi
 {
 namespace dpl
@@ -148,6 +150,25 @@ struct acc_handler_count
         return (__predicate(acc[gidx]) ? 1 : 0);
     }
 };
+
+template <typename _T>
+struct __decay_with_tuple_specialization
+{
+    using type = ::std::decay_t<_T>;
+};
+
+template <typename... _Args>
+struct __decay_with_tuple_specialization<oneapi::dpl::__internal::tuple<_Args...>>
+{
+    using type = oneapi::dpl::__internal::tuple<::std::decay_t<_Args>...>;
+};
+
+template <typename... _Args>
+struct __decay_with_tuple_specialization<::std::tuple<_Args...>>
+{
+    using type = ::std::tuple<::std::decay_t<_Args>...>;
+};
+
 } // namespace __internal
 } // namespace dpl
 } // namespace oneapi
--- a/include/oneapi/dpl/pstl/hetero/utils_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/utils_hetero.h
@@ -21,8 +21,6 @@
 #ifndef _ONEDPL_UTILS_HETERO_H
 #define _ONEDPL_UTILS_HETERO_H
 
-#include "../tuple_impl.h"
-
 namespace oneapi
 {
 namespace dpl
@@ -150,28 +148,6 @@ struct acc_handler_count
         return (__predicate(acc[gidx]) ? 1 : 0);
     }
 };
-
-template <typename _T>
-struct __decay_with_tuple_specialization
-{
-    using type = ::std::decay_t<_T>;
-};
-
-template <typename... _Args>
-struct __decay_with_tuple_specialization<oneapi::dpl::__internal::tuple<_Args...>>
-{
-    using type = oneapi::dpl::__internal::tuple<::std::decay_t<_Args>...>;
-};
-
-template <typename... _Args>
-struct __decay_with_tuple_specialization<::std::tuple<_Args...>>
-{
-    using type = ::std::tuple<::std::decay_t<_Args>...>;
-};
-
-template <typename... _Args>
-using __decay_with_tuple_specialization_t = typename __decay_with_tuple_specialization<_Args...>::type;
-
 } // namespace __internal
 } // namespace dpl
 } // namespace oneapi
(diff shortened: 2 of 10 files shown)


include/oneapi/dpl/pstl/hetero/utils_hetero.h
```diff
@@ -148,6 +150,28 @@ struct acc_handler_count
...
+
+template <typename... _Args>
+struct __decay_with_tuple_specialization<::std::tuple<_Args...>>
+{
+    using type = ::std::tuple<::std::decay_t<_Args>...>;
+};
+
+template <typename... _Args>
+using __decay_with_tuple_specialization_t = typename __decay_with_tuple_specialization<_Args...>::type;
+
```
danhoeflinger, line 174: Is there a reason this is here in `utils_hetero?`  It seems like this should be accessible to any backend.  On first inspection, perhaps it belongs in `tuple_impl.h`?

include/oneapi/dpl/internal/function.h
```diff
@@ -133,7 +133,8 @@ struct segmented_scan_fun
     operator()(const _T1& x, const _T2& y) const
     {
         using ::std::get;
-        auto new_x = get<1>(y) ? get<0>(y) : binary_op(get<0>(x), get<0>(y));
+        using x_t = ::std::tuple_element_t<0, _T1>;
+        auto new_x = get<1>(y) ? x_t(get<0>(y)) : x_t(binary_op(get<0>(x), get<0>(y)));
```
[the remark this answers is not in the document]
mmichel11: Actually, after taking a closer look, this change results in the following compiler error:

`error: conditional expression is ambiguous`

It seems the compiler is trying to cast to `x_t` after evaluating the conditional. I think the previous syntax is necessary.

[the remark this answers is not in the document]
danhoeflinger: Ah yes, both sides of the ternary expression need the same type. I'm fine with going back to how it was.

mmichel11:
After some offline discussion, I have removed the changes to the `zip_iterator.pass` tests and instead updated `reduce_by_segment_zip.pass` to pass zip iterators as input / output values and added `inclusive_scan_by_segment_zip.pass` and `exclusive_scan_by_segment_zip.pass`.

test/parallel_api/numeric/numeric.ops/inclusive_scan_by_segment_zip.pass.cpp
```diff
@@ -0,0 +1,112 @@
...
+    //data initialization
+    int keys1  [n] = { 11, 11, 21, 20, 21, 21, 21, 37, 37 };
+    int keys2  [n] = { 11, 11, 20, 20, 20, 21, 21, 37, 37 };
+    int values1[n] = {  0,  1,  2,  3,  4,  5,  6,  7,  8 };
+    int values2[n] = {  0,  1,  2,  3,  4,  5,  6,  7,  8 };
+    int output_values1[n] = { };
+    int output_values2[n] = { };
+
+    // allocate USM memory and copying data to USM shared/device memory
+    TestUtils::usm_data_transfer<alloc_type, int> dt_helper1(q, std::begin(keys1),          std::end(keys1));
```
SergeyKopienko, line 49: May be better to use something like ```TestUtils::usm_data_transfer<alloc_type, int> dt_helper1(q, keys1, n);``` ?

test/parallel_api/numeric/numeric.ops/inclusive_scan_by_segment_zip.pass.cpp
```diff
@@ -0,0 +1,112 @@
...
+    auto begin_vals_in = oneapi::dpl::make_zip_iterator(d_values1, d_values2);
+    auto begin_vals_out= oneapi::dpl::make_zip_iterator(d_output_values1, d_output_values2);
+
+    //run inclusive_scan_by_segment algorithm 
+    oneapi::dpl::inclusive_scan_by_segment(
+        TestUtils::make_device_policy<KernelName>(q), begin_keys_in,
+        end_keys_in, begin_vals_in, begin_vals_out,
+        ::std::equal_to<>(), TestUtils::tuple_add_fn());
+
+    q.wait();
```
SergeyKopienko, line 74: Is this call really required here?

test/parallel_api/numeric/numeric.ops/exclusive_scan_by_segment_zip.pass.cpp
```diff
@@ -0,0 +1,112 @@
...
+    auto begin_vals_in = oneapi::dpl::make_zip_iterator(d_values1, d_values2);
+    auto begin_vals_out= oneapi::dpl::make_zip_iterator(d_output_values1, d_output_values2);
+
+    //run exclusive_scan_by_segment algorithm 
+    oneapi::dpl::exclusive_scan_by_segment(
+        TestUtils::make_device_policy<KernelName>(q), begin_keys_in,
+        end_keys_in, begin_vals_in, begin_vals_out, ::std::make_tuple(int(1), int(1)),
+        ::std::equal_to<>(), TestUtils::tuple_add_fn());
+
+    q.wait();
```
SergeyKopienko, line 74: Is this call really required here?

test/parallel_api/numeric/numeric.ops/exclusive_scan_by_segment_zip.pass.cpp
```diff
@@ -0,0 +1,112 @@
...
+    //data initialization
+    int keys1  [n] = { 11, 11, 21, 20, 21, 21, 21, 37, 37 };
+    int keys2  [n] = { 11, 11, 20, 20, 20, 21, 21, 37, 37 };
+    int values1[n] = {  0,  1,  2,  3,  4,  5,  6,  7,  8 };
+    int values2[n] = {  0,  1,  2,  3,  4,  5,  6,  7,  8 };
+    int output_values1[n] = { };
+    int output_values2[n] = { };
+
+    // allocate USM memory and copying data to USM shared/device memory
+    TestUtils::usm_data_transfer<alloc_type, int> dt_helper1(q, std::begin(keys1),          std::end(keys1));
```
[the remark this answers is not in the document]
mmichel11: Yes, this syntax is better. I have made this change.

test/parallel_api/numeric/numeric.ops/inclusive_scan_by_segment_zip.pass.cpp
mmichel11: Yes, this syntax is better. I have made this change.

test/parallel_api/numeric/numeric.ops/inclusive_scan_by_segment_zip.pass.cpp
mmichel11: No, I have removed it from all locations.

test/parallel_api/numeric/numeric.ops/exclusive_scan_by_segment_zip.pass.cpp
mmichel11: No, I have removed it from all locations.

mmichel11: No, I have removed this.

test/support/utils.h
```diff
@@ -754,6 +755,18 @@ struct MaxAbsFunctor<::std::complex<_Tp>>
     }
 };
 
+struct tuple_add_fn
```
danhoeflinger, line 758:
```suggestion
struct TupleAddFunctor
```
...to match the style of other functors here, obviously update the usages as well.

test/support/utils.h
```diff
@@ -754,6 +755,18 @@ struct MaxAbsFunctor<::std::complex<_Tp>>
...
+{
+    template <typename Tup1, typename Tup2>
+    auto operator()(Tup1 lhs, const Tup2& rhs) const
+    {
+        using ::std::get;
+        get<0>(lhs) += get<1>(rhs);
+        get<1>(lhs) += get<1>(rhs);
+        return lhs;
+    }
```
danhoeflinger, line 767: I don't have a strong preference here, but I'm curious why not use `const Tup1& lhs` and make a new tuple for the return?
Also, for the way we use this functor, can't we just have a single template argument for `lhs`, `rhs` and return type?

danhoeflinger (commented):
A couple minor things I'd like changes for, but I think otherwise LGTM.

test/support/utils.h
mmichel11: I have made this change.

test/support/utils.h
mmichel11: I have switched to taken both parameters by const reference and creating a new tuple to return.

The reason we can't use a single template argument is due to constness differences between `lhs` and `rhs`. `lhs` will always have decayed tuple elements since it is the type of the tuple that collects the reduction value. The type of `rhs` may differ since it is passed directly from the input values buffer [here](https://github.com/oneapi-src/oneDPL/blob/a678aee8b1afd1b4fe098897e0cf88a0743e7386/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L945C67-L945C67). Using a single template type causes a compiler error.

test/support/utils.h
```diff
@@ -755,15 +755,15 @@ struct MaxAbsFunctor<::std::complex<_Tp>>
     }
 };
 
-struct tuple_add_fn
+struct TupleAddFunctor
 {
     template <typename Tup1, typename Tup2>
-    auto operator()(Tup1 lhs, const Tup2& rhs) const
+    auto
+    operator()(const Tup1& lhs, const Tup2& rhs) const
```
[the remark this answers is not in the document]
mmichel11: I think in this case using a const reference is best although I may be mistaken. We don't need to forward the arguments anywhere and temporary objects can bind to const references, so I don't think a forward reference is necessary.

Is there any reason forwarding references would be preferred here?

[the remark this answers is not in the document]
SergeyKopienko: I think the form ```Tup1&& lhs, Tup2&& rhs`` contains less requirements about params value kind: I mean l-value, r-value and so on.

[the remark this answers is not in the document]
danhoeflinger: I believe that `const Tup1& lhs` should bind to all incoming options (meaning it shouldn't put any additional requirements as compared to universal references).

I think the advantage to universal references like `Tup1&& lhs` is, as @mmichel11 mentions, is that you can keep rvalues as rvalues and forward them along as rvalues.  However, it seems that doesn't provide us anything we need in this setting.  

https://stackoverflow.com/questions/3106110/what-is-move-semantics/11540204#11540204
This is a useful table for binding from the above link:
```
            lvalue   const lvalue   rvalue   const rvalue
---------------------------------------------------------              
X&          yes
const X&    yes      yes            yes      yes
X&&                                 yes
const X&&                           yes      yes
```

I may be missing something, but I think const lvalue references are best here because universal references add complication without a clear benefit.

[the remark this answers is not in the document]
SergeyKopienko: OK

mmichel11:
We are seeing the known `reduce_by_segment` optimizer crash in the CI. This first occurred after my commit this morning which does not effect the `reduce_by_segment.pass` test at all, so I believe this is unrelated to this PR.

danhoeflinger (approved):
LGTM

PR #1024, opened 2023-07-24
[KT] Add radix sort by key (onesweep)

dmitriy-sobolev:
It duplicates a large share of existing radix sort code with some sparse modifications to support radix sort by key.
Probably, it is easier to avoid changes in other (non-related to radix sort by key) code at this point; then proceed to refactoring in a follow up PR.


dmitriy-sobolev:
I am closing this PR to create a newer one with esimd-radix-sort аs a base branch.

PR #1026, opened 2023-07-25
Document algorithmic design and assumptions of the family of reduce patterns

### LINKED ISSUE #904: Document algorithmic design and assumptions of the family of reduce patterns
Multiple algorithms in oneDPL use the generic `transform_reduce` pattern including `count`, `is_partitioned`, `lexicographical_compare`, `min_element`, `minmax_element`, `max_element`, `reduce`. Several of these algorithms require in order processing of the elements. These assumptions should be documented:

- `is_partitioned`, `lexicographical_compare`: Require in-order processing in general.
- `min_element`, `minmax_element`, `max_element`: The current implementation requires in-order processing and ascending indices. They further require that the first element compared has a lower index than the second. The existing implementation of `min_element` and `max_element` requires that already.  #902 requires it for `minmax_element` as well.

There are further many `__identity_init_fn` and `__identity_reduce_fn` with very different meanings. The naming of these should be improved.

julianmi:
Multiple algorithms in oneDPL use the generic `transform_reduce` pattern including `count`, `is_partitioned`, `lexicographical_compare`, `min_element`, `minmax_element`, `max_element`, `reduce`. Several of these algorithms require in order processing of the elements, i.e., they are non-commutative. This patch documents this behavior for each algorithm using the reduction pattern by introducing a template parameter `_Commutative`.

- `is_partitioned`, `lexicographical_compare`: Require in-order processing in general.
- `min_element`, `minmax_element`, `max_element`: The current implementation requires in-order processing since the indices are not tracked for performance reasons in case of equal minimum or maximum values.

Further, the ambiguous use of `__identity_init_fn` and `__identity_reduce_fn` was clarified by specifying the operation (reduction or transformation) and the name of the pattern.

Closes #904.


include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1195,17 +1206,16 @@ __pattern_sort_by_key(_ExecutionPolicy&& __exec, _Iterator1 __keys_first, _Itera
                       _Iterator2 __values_first, _Compare __comp, /*vector=*/::std::true_type,
                       /*parallel=*/::std::true_type)
 {
-    static_assert(::std::is_move_constructible_v<typename ::std::iterator_traits<_Iterator1>::value_type>
-        && ::std::is_move_constructible_v<typename ::std::iterator_traits<_Iterator2>::value_type>,
-        "The keys and values should be move constructible in case of parallel execution.");
+    static_assert(::std::is_move_constructible_v<typename ::std::iterator_traits<_Iterator1>::value_type> &&
```
mmichel11, line 1209: Is this change to formatting intentional or is it the result of applying `clang-format` across the whole file? There are some other instances in this PR where just the formatting has changed like this.

julianmi:
I noticed a performance regression introduced by switching to structs for the reduce and transform operators. I changed the reduce-based patterns to use lambda operators instead. The performance now matches the mainline.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -384,23 +384,28 @@ __pattern_min_element(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __
         typename ::std::make_unsigned<typename ::std::iterator_traits<_Iterator>::difference_type>::type;
     using _ReduceValueType = tuple<_IndexValueType, _IteratorValueType>;
 
-    auto __identity_init_fn = __acc_handler_minelement<_ReduceValueType>{};
-    auto __identity_reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
+    // This operator doesn't track the lowest found index in case of equal min. or max. values. Thus, this operator is
+    // not commutative.
+    auto __identity_reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) -> _ReduceValueType {
```
akukanov, line 389: Does the word "identity" in the names have some descriptive meaning? I do not quite get it.

akukanov, line 389: Here and in other lambdas, explicitly declaring the return type seems redundant, as it's deducible and also does not add much (if anything) for readability.

julianmi: Not that I know of. This naming scheme was previously used. I will remove it if there are no objections.

julianmi: Good point, I will change that. I think the only needed for `count` where we cast the return value of the predicate to an integer.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1294,11 +1289,20 @@ __pattern_lexicographical_compare(_ExecutionPolicy&& __exec, _Iterator1 __first1
...
-    auto __identity_reduce_fn = [](_ReduceValueType __a, _ReduceValueType __b) -> _ReduceValueType {
+    auto __reduce_fn = [](_ReduceValueType __a, _ReduceValueType __b) {
         bool __is_mismatched = __a != 0;
         return __a * __is_mismatched + __b * !__is_mismatched;
     };
+    auto __transform_fn = [__comp](auto __gidx, auto __acc1, auto __acc2) {
+        auto __s1_val = __acc1[__gidx];
+        auto __s2_val = __acc2[__gidx];
+
+        int32_t __is_s1_val_less = __comp(__s1_val, __s2_val);
```
mmichel11, line 1300:
```suggestion
        ::std::int32_t __is_s1_val_less = __comp(__s1_val, __s2_val);
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -193,15 +193,15 @@ struct __parallel_transform_reduce_device_kernel_submitter<__work_group_size, __
 // Submits the second kernel of the parallel_transform_reduce for mid-sized arrays.
 // Uses a single work groups to reduce __n preliminary results stored in __temp and returns a future object with the
 // result buffer.
-template <::std::uint16_t __work_group_size, ::std::uint8_t __iters_per_work_item, typename _Tp, typename _KernelName>
+template <typename _Tp, ::std::uint16_t __work_group_size, ::std::uint8_t __iters_per_work_itemx, typename _KernelName>
```
mmichel11, line 196: I think `__iters_per_work_itemx` is a typo here.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1294,11 +1289,20 @@ __pattern_lexicographical_compare(_ExecutionPolicy&& __exec, _Iterator1 __first1
...
     using _ReduceValueType = int32_t;
 
-    auto __identity_init_fn = acc_handler_lexicographical_compare<_Compare, _ReduceValueType>{__comp};
-    auto __identity_reduce_fn = [](_ReduceValueType __a, _ReduceValueType __b) -> _ReduceValueType {
+    auto __reduce_fn = [](_ReduceValueType __a, _ReduceValueType __b) {
         bool __is_mismatched = __a != 0;
         return __a * __is_mismatched + __b * !__is_mismatched;
     };
+    auto __transform_fn = [__comp](auto __gidx, auto __acc1, auto __acc2) {
+        auto __s1_val = __acc1[__gidx];
```
mmichel11, line 1297: Is it better to store `__s1_val` and `__s2_val` as const references since we do not need a local copy of them?

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -582,12 +592,36 @@ __pattern_minmax_element(_ExecutionPolicy&& __exec, _Range&& __rng, _Compare __c
     using _ReduceValueType =
         oneapi::dpl::__internal::tuple<_IndexValueType, _IndexValueType, _IteratorValueType, _IteratorValueType>;
 
-    auto __identity_init_fn = __acc_handler_minmaxelement<_ReduceValueType>{};
+    // This operator doesn't track the lowest found index in case of equal min. values and the highest found index in
+    // case of equal max. values. Thus, this operator is not commutative.
+    auto __reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
```
mmichel11, line 597: Since `__a` and `__b` are taken by value, we might want to use `::std::move` when assigning to `__chosen_for_min` and `__chosen_for_max` to avoid unnecessary copies with larger data types.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -396,8 +400,9 @@ struct __parallel_transform_reduce_impl
 // Mid-sized arrays use two tree reductions with independent __iters_per_work_item.
 // Big arrays are processed with a recursive tree reduction. __work_group_size * __iters_per_work_item elements are
 // reduced in each step.
-template <typename _Tp, typename _ReduceOp, typename _TransformOp, typename _ExecutionPolicy, typename _InitType,
-          oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0, typename... _Ranges>
+template <typename _Tp, typename _ReduceOp, typename _TransformOp, typename _Commutative, typename _ExecutionPolicy,
+          typename _InitType, oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0,
+          typename... _Ranges>
```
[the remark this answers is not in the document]
julianmi: I changed the order.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: Good catch, thanks.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1294,11 +1289,20 @@ __pattern_lexicographical_compare(_ExecutionPolicy&& __exec, _Iterator1 __first1
...
+        auto __s2_val = __acc2[__gidx];
+
+        int32_t __is_s1_val_less = __comp(__s1_val, __s2_val);
+        int32_t __is_s1_val_greater = __comp(__s2_val, __s1_val);
```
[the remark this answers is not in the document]
julianmi: Done.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
julianmi: I changed that.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -438,19 +440,43 @@ __pattern_minmax_element(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator
     using _IndexValueType =
         typename ::std::make_unsigned<typename ::std::iterator_traits<_Iterator>::difference_type>::type;
     using _ReduceValueType = ::std::tuple<_IndexValueType, _IndexValueType, _IteratorValueType, _IteratorValueType>;
-    using _ReduceFnType = __identity_reduce_fn<_Compare>;
 
-    auto __identity_init_fn = __acc_handler_minmaxelement<_ReduceValueType>{};
+    // This operator doesn't track the lowest found index in case of equal min. values and the highest found index in
+    // case of equal max. values. Thus, this operator is not commutative.
+    auto __reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
```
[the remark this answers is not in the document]
julianmi: Done.

danhoeflinger (commented):
I understand the point of this PR is not to necessarily improve the predicate code that is being moved around, but rather to document if the different algorithms are require commutative properties or not and refactor some things for clarity.
However, we are touching it, so we can at least discuss it, and hopefully grab some potential performance gains.  They may only help for more complex types if at all.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -532,15 +556,19 @@ __pattern_count(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __last,
 
     using _ReduceValueType = typename ::std::iterator_traits<_Iterator>::difference_type;
 
-    auto __identity_init_fn = acc_handler_count<_Predicate>{__predicate};
-    auto __identity_reduce_fn = ::std::plus<_ReduceValueType>{};
+    auto __reduce_fn = ::std::plus<_ReduceValueType>{};
+    // int is being implicitly casted to difference_type
+    // otherwise we can only pass the difference_type as a functor template parameter
+    auto __transform_fn = [__predicate](auto __gidx, auto __acc) -> int {
+        return (__predicate(__acc[__gidx]) ? 1 : 0);
```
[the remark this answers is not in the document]
akukanov: The compiler should be smart enough to optimize that.
Besides being somewhat less obvious, an expression `int(__predicate(__acc[__gidx]))` results in 1 or 0 only if the predicate returns a bool. If however it returns something implicitly convertible to bool, the ternary operator and the cast might have different results, or even the cast will not compile if the thing converts to bool but not to int.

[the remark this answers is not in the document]
danhoeflinger: Good points, works for me.  I see that the requirement is only that the predicate return is convertible to bool, not that it must be bool.  Thanks.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -312,12 +312,16 @@ __pattern_count(_ExecutionPolicy&& __exec, _Range&& __rng, _Predicate __predicat
 
     using _ReduceValueType = oneapi::dpl::__internal::__difference_t<_Range>;
 
-    auto __identity_init_fn = acc_handler_count<_Predicate>{__predicate};
-    auto __identity_reduce_fn = ::std::plus<_ReduceValueType>{};
+    auto __reduce_fn = ::std::plus<_ReduceValueType>{};
+    // int is being implicitly casted to difference_type
+    // otherwise we can only pass the difference_type as a functor template parameter
+    auto __transform_fn = [__predicate](auto __gidx, auto __acc) -> int {
+        return (__predicate(__acc[__gidx]) ? 1 : 0);
```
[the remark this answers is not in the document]
danhoeflinger: See above, retracted.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -384,23 +384,25 @@ __pattern_min_element(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __
         typename ::std::make_unsigned<typename ::std::iterator_traits<_Iterator>::difference_type>::type;
     using _ReduceValueType = tuple<_IndexValueType, _IteratorValueType>;
 
-    auto __identity_init_fn = __acc_handler_minelement<_ReduceValueType>{};
-    auto __identity_reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
+    // This operator doesn't track the lowest found index in case of equal min. or max. values. Thus, this operator is
+    // not commutative.
+    auto __reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
```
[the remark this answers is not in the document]
akukanov: It depends on which data we want to optimize for. For arithmetic data types and perhaps even small structs (like pairs, complex etc.) I think by-value operations tend to be better optimized by modern compilers. More sophisticated data types might better be passed by reference, but I would prefer any such changes - especially "for many of the lambdas" to be driven by actual performance data. Times of "common wisdom" for such low-level optimizations have gone :)

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -582,12 +592,35 @@ __pattern_minmax_element(_ExecutionPolicy&& __exec, _Range&& __rng, _Compare __c
     using _ReduceValueType =
         oneapi::dpl::__internal::tuple<_IndexValueType, _IndexValueType, _IteratorValueType, _IteratorValueType>;
 
-    auto __identity_init_fn = __acc_handler_minmaxelement<_ReduceValueType>{};
+    // This operator doesn't track the lowest found index in case of equal min. values and the highest found index in
+    // case of equal max. values. Thus, this operator is not commutative.
+    auto __reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
```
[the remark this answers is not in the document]
akukanov: Can you show with data that it makes a performance difference? Taking copies, especially for small size objects, are well optimized by modern compilers.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -545,16 +549,22 @@ __pattern_min_element(_ExecutionPolicy&& __exec, _Range&& __rng, _Compare __comp
...
+    // This operator doesn't track the lowest found index in case of equal min. or max. values. Thus, this operator is
+    // not commutative.
+    auto __reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
         using ::std::get;
-        return __comp(get<1>(__b), get<1>(__a)) ? __b : __a;
+        if (__comp(get<1>(__b), get<1>(__a)))
+        {
+            return __b;
+        }
+        return __a;
```
akukanov, line 560: What was wrong with the ?: operator?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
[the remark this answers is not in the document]
danhoeflinger: Yes, you are right.  The performance numbers should drive the answers to these questions, and we are likely more interested in preferring arithmetic data types.  
Given that this isn't the point of this PR perhaps its not the setting to investigate this.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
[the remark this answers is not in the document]
danhoeflinger: I'll take a look and see if I get any improvement.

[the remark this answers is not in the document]
danhoeflinger: As you suspected, all differences are within the margin of error here for `int` and `float` across our normal benchmark sizes.  I don't have an easy way to test more complex types, that that would of course be heavily dependent on the type used.
  
I'll withdraw the suggestion, as it complicates the code.

danhoeflinger (commented):
After some investigation on my own, I think other than the formatting-only change this LGTM

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1197,7 +1210,7 @@ __pattern_sort_by_key(_ExecutionPolicy&& __exec, _Iterator1 __keys_first, _Itera
 {
     static_assert(::std::is_move_constructible_v<typename ::std::iterator_traits<_Iterator1>::value_type>
         && ::std::is_move_constructible_v<typename ::std::iterator_traits<_Iterator2>::value_type>,
-        "The keys and values should be move constructible in case of parallel execution.");
+                  "The keys and values should be move constructible in case of parallel execution.");
```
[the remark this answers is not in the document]
julianmi: I've reverted this clang-format change.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
akukanov: I see. For brevity, I would use ?: in both cases, but that's up to you.

danhoeflinger (approved):
LGTM

PR #1029, opened 2023-07-26
[KT] Support work with accessors in scatter and gather functions for 8-byte data types

SergeyKopienko:
In this PR we implement work with accessors in ```oneapi::dpl::experimental::esimd::impl::utils::scatter``` and ```oneapi::dpl::experimental::esimd::impl::utils::gather``` functions for 8-byte data types.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
@@ -66,13 +66,22 @@ gather(const T* input, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets
     return sycl::ext::intel::esimd::gather(input + base_offset, offsets * size32<T>, mask);
 }
 
-template <typename T, int N, sycl::access_mode Mode, sycl::access::placeholder P>
+template <typename T, int N, sycl::access_mode Mode, sycl::access::placeholder P, ::std::enable_if_t<sizeof(T) <= sizeof(::std::uint32_t), int> = 0>
 sycl::ext::intel::esimd::simd<T, N>
 gather(sycl::accessor<T, 1, Mode, sycl::target::device, P> input, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets,
        ::std::uint32_t base_offset, sycl::ext::intel::esimd::simd_mask<N> mask = 1)
 {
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
```diff
@@ -66,14 +66,28 @@ gather(const T* input, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets
...
 sycl::ext::intel::esimd::simd<T, N>
 gather(sycl::accessor<T, 1, Mode, sycl::target::device, P> input, sycl::ext::intel::esimd::simd<::std::uint32_t, N> offsets,
        ::std::uint32_t base_offset, sycl::ext::intel::esimd::simd_mask<N> mask = 1)
 {
     return sycl::ext::intel::esimd::gather<T>(input, offsets * size32<T>, base_offset * size32<T>, mask);
 }
 
+template <typename T, int N, sycl::access_mode Mode, sycl::access::placeholder P, ::std::enable_if_t<sizeof(T) == sizeof(::std::uint64_t), int> = 0>
+sycl::ext::intel::esimd::simd<T, N>
+gather(sycl::accessor<T, 1, Mode, sycl::target::device, P> input,
```
dmitriy-sobolev, line 79: I suggest that we get rid of `utils::gather` and `utils::scatter` functions in favor of using `lsc_gather` and `lsc_scatter` directly, as it already done in cooperative and onesweep implementations.

LSC counterparts do not need abstractions due to having compatible API for accessors and pointers.

SergeyKopienko: I am not sure that we should do these changes.
As for me better to use ```utils::gather``` and ```utils::scatter``` functions in all radix sort implementations.

dmitriy-sobolev (approved):
This looks descent at this point as a fix of an issue with 8-byte wide types.

I am still suggesting getting rid of `utils;:gather` and `utils::scatter` as described in https://github.com/oneapi-src/oneDPL/pull/1029/files#r1274804801.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_utils.h
dmitriy-sobolev: Probably, adding abstractions to avoid adding things like `uint32_t(sizeof(KeyT))` worth, because conversion to uint32_t is not straightforward.

PR #1031, opened 2023-07-27
[CI] Fix execution status of Windows CI jobs

dmitriy-sobolev:
This PR fixes most possible cases when CI provides "green" status despite of finishing with errors. This is a continuation of #886. 

Example:
image
This happens because of default behaviour of `shell: cmd`, which is described in https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#exit-codes-and-error-action-preference

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,9 +248,9 @@ jobs:
           set CPATH=%CONDA_PREFIX%\include
           set TBB_DLL_PATH=%CONDA_PREFIX%\Library\bin
           reg add HKLM\SOFTWARE\khronos\OpenCL\Vendors /v intelocl64.dll /t REG_DWORD /d 00000000
-          cmake -G "Ninja" -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DCMAKE_CXX_STANDARD=${{ matrix.std }} -DCMAKE_CXX_COMPILER=${{ matrix.cxx_compiler }} -DONEDPL_BACKEND=${{ matrix.backend }} -DONEDPL_DEVICE_TYPE=${{ matrix.device_type }} .. || goto fail_fast
-          ninja -j 2 -v %ninja_targets% || goto fail_fast
-          ctest --timeout %TEST_TIMEOUT% -C ${{ matrix.build_type }} --output-on-failure %ctest_flags% || goto fail_fast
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -248,9 +248,13 @@ jobs:
...
           reg add HKLM\SOFTWARE\khronos\OpenCL\Vendors /v intelocl64.dll /t REG_DWORD /d 00000000
-          cmake -G "Ninja" -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DCMAKE_CXX_STANDARD=${{ matrix.std }} -DCMAKE_CXX_COMPILER=${{ matrix.cxx_compiler }} -DONEDPL_BACKEND=${{ matrix.backend }} -DONEDPL_DEVICE_TYPE=${{ matrix.device_type }} ..
-          ninja -j 2 -v %ninja_targets%
-          ctest --timeout %TEST_TIMEOUT% -C ${{ matrix.build_type }} --output-on-failure %ctest_flags%
+          cmake -G "Ninja" -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} -DCMAKE_CXX_STANDARD=${{ matrix.std }} -DCMAKE_CXX_COMPILER=${{ matrix.cxx_compiler }} -DONEDPL_BACKEND=${{ matrix.backend }} -DONEDPL_DEVICE_TYPE=${{ matrix.device_type }} .. || goto :fail_fast
+          ninja -j 2 -v %ninja_targets% || goto :fail_fast
+          ctest --timeout %TEST_TIMEOUT% -C ${{ matrix.build_type }} --output-on-failure %ctest_flags% || goto :fail_fast
+          exit /b 0
+          :: modify the default behaviour of shell:cmd, which exits with the status of a last command, in order not to unintentially miss an error
+          :fail_fast
```
danhoeflinger, line 256: I wonder if the naming for `fail_fast` should be changed to something else, like `short_circuit_fail` or `fail_out` to prevent confusion. 

The term `fail_fast` has meaning for github actions, and is defined as `false` in the strategy for this task.  I believe it means if one task fails in a matrix, it skips the rest.  https://www.edwardthomson.com/blog/github_actions_6_fail_fast_matrix_workflows.html

danhoeflinger (commented):
Other than one small suggestion, LGTM, I think this will be a nice change.

.github/workflows/ci.yml
dmitriy-sobolev: Thanks. That's done.

danhoeflinger (approved):
LGTM

PR #1034, opened 2023-07-27
[Dynamic Selection] Adding host task to sycl scheduler and adding a load-based policy

AnuyaWelling2801:
1. Change sycl scheduler to make it not call complete event only on wait. Needs to create a depenent task that executes at completion. Used sycl::host_task to accomplish this.
2. Added a dynamic load policy that chooses a execution resource based on the load on the device.
3. Added tests for dynamic load policy


include/oneapi/dpl/internal/dynamic_selection_impl/dynamic_load_policy_impl.h
```diff
@@ -0,0 +1,164 @@
...
+    }
+
+    template<typename Function, typename ...Args>
+    auto invoke_async(selection_handle_t e, Function&& f, Args&&... args) {
+      return sched_->submit(e, std::forward<Function>(f), std::forward<Args>(args)...);
+    }
+
+    template<typename Function, typename ...Args>
+    auto invoke(Function&& f, Args&&... args) {
+      return wait_for_all(sched_->submit(select(std::forward<Function>(f), std::forward<Args>(args)...),
```
vossmjp, line 138: double forward

vossmjp (commented):
Minor issues.  I think we're still waiting on updated test.

include/oneapi/dpl/internal/dynamic_selection_impl/dynamic_load_policy_impl.h
```diff
@@ -0,0 +1,164 @@
...
+
+      property_handle_t h_;
+
+      selection_handle_t(property_handle_t h) : h_(h) {}
+      native_resource_t get_native() { return h_.resource_ptr_->e_.get_native(); }
+      auto get_property_handle() { return h_; }
+    };
+
+    std::shared_ptr<scheduler_t> sched_;
+    universe_container_t universe_;
```
vossmjp, line 67: Need to have both sched_ and universe_ accessed via shared_ptr to support reference semantics. And also should like make private things private instead of struct with all public.

PR #1037, opened 2023-08-02
Add check for in-place scan before invoking __simd_scan

mmichel11:
On compilers in which we enable the `__simd_scan` codepath (such as `icpc 2021.8`), our `scan.pass` test cases have been failing for `unseq` and `unseq_par` host policies. In particular, the specific cases that test in-place scans fail. This is due to the following restriction in the [OpenMP specification:](https://www.openmp.org/spec-html/5.0/openmpsu45.html)
```
"Intra-iteration dependences from a statement in the structured block preceding a scan directive to a statement in the 
structured block following a scan directive must not exist, except for dependences for the list items specified in an 
inclusive or exclusive clause."
```
When an in-place scan is performed, the `__first` and `__result` iterators in [__simd_scan](https://github.com/oneapi-src/oneDPL/blob/0b57a685568ea1ccfab50ad266da9c71fd8bf7d4/include/oneapi/dpl/pstl/unseq_backend_simd.h#L547) correspond to the same start address in memory. The scan update performed in the [scan phase](https://github.com/oneapi-src/oneDPL/blob/0b57a685568ea1ccfab50ad266da9c71fd8bf7d4/include/oneapi/dpl/pstl/unseq_backend_simd.h#L553) must be performed after the update to the sum variable in the [input phase](https://github.com/oneapi-src/oneDPL/blob/0b57a685568ea1ccfab50ad266da9c71fd8bf7d4/include/oneapi/dpl/pstl/unseq_backend_simd.h#L555) for the results to be correct. This represents a dependency outside of the list-item between the input and scan phase in the same loop iteration. This dependency exists for both inclusive and exclusive scans, and we should not use the scan directive in this case.

To fix this issue, we can check if the `__first` and `__last` iterators begin at the same start address and use the non-vectorized implementation. This resolves the encountered issues in the `scan.pass` test.

### DIFF
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -207,7 +207,7 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
 {
 #if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
     // An in-place scan violates the intra-iteration dependency restriction of the omp scan directive. We must call our serial brick.
-    if (::std::addressof(*__first) != ::std::addressof(*__result))
+    if (reinterpret_cast<void*>(::std::addressof(*__first)) != reinterpret_cast<void*>(::std::addressof(*__result)))
     {
         return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
                                             _Inclusive());
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -206,13 +206,16 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
                        /*is_vector=*/::std::true_type) noexcept
 {
 #if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
-    return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
-                                        _Inclusive());
-#else
+    // An in-place scan violates the intra-iteration dependency restriction of the omp scan directive. We must call our serial brick.
+    if (::std::addressof(*__first) != ::std::addressof(*__result))
+    {
+        return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
+                                            _Inclusive());
+    }
+#endif
     // We need to call serial brick here to call function for inclusive and exclusive scan that depends on _Inclusive() value
     return __internal::__brick_transform_scan(__first, __last, __result, __unary_op, __init, __binary_op, _Inclusive(),
                                               /*is_vector=*/::std::false_type());
-#endif
 }
 
 template <class _RandomAccessIterator, class _OutputIterator, class _UnaryOperation, class _Tp, class _BinaryOperation,
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -206,8 +206,10 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
                        /*is_vector=*/::std::true_type) noexcept
 {
 #if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
+    const auto& __first_tmp = *__first;
+    const auto& __result_tmp = *__result;
     // An in-place scan violates the intra-iteration dependency restriction of the omp scan directive. We must call our serial brick.
-    if (static_cast<const void*>(::std::addressof(*__first)) != static_cast<const void*>(::std::addressof(*__result)))
+    if (static_cast<const void*>(::std::addressof(__first_tmp)) != static_cast<const void*>(::std::addressof(__result_tmp)))
     {
         return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
                                             _Inclusive());
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -207,7 +207,7 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
 {
 #if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
     // An in-place scan violates the intra-iteration dependency restriction of the omp scan directive. We must call our serial brick.
-    if (reinterpret_cast<void*>(::std::addressof(*__first)) != reinterpret_cast<void*>(::std::addressof(*__result)))
+    if (static_cast<const void*>(::std::addressof(*__first)) != static_cast<const void*>(::std::addressof(*__result)))
     {
         return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
                                             _Inclusive());
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -209,7 +209,8 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
     const auto& __first_tmp = *__first;
     const auto& __result_tmp = *__result;
     // An in-place scan violates the intra-iteration dependency restriction of the omp scan directive. We must call our serial brick.
-    if (static_cast<const void*>(::std::addressof(__first_tmp)) != static_cast<const void*>(::std::addressof(__result_tmp)))
+    if (static_cast<const void*>(::std::addressof(__first_tmp)) !=
+        static_cast<const void*>(::std::addressof(__result_tmp)))
     {
         return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
                                             _Inclusive());
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -206,13 +206,16 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
                        /*is_vector=*/::std::true_type) noexcept
 {
 #if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
-    return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
-                                        _Inclusive());
-#else
+    if (::std::is_same_v<_Inclusive, ::std::true_type> ||
+        !oneapi::dpl::__internal::__iterators_possibly_equal(__first, __result))
+    {
+        return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
+                                            _Inclusive());
+    }
+#endif
     // We need to call serial brick here to call function for inclusive and exclusive scan that depends on _Inclusive() value
     return __internal::__brick_transform_scan(__first, __last, __result, __unary_op, __init, __binary_op, _Inclusive(),
                                               /*is_vector=*/::std::false_type());
-#endif
 }
 
 template <class _RandomAccessIterator, class _OutputIterator, class _UnaryOperation, class _Tp, class _BinaryOperation,
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -202,11 +202,11 @@ template <class _RandomAccessIterator, class _OutputIterator, class _UnaryOperat
           class _Inclusive>
 ::std::enable_if_t<!is_arithmetic_udop<_Tp, _BinaryOperation>::value, ::std::pair<_OutputIterator, _Tp>>
 __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __last, _OutputIterator __result,
-                       _UnaryOperation __unary_op, _Tp __init, _BinaryOperation __binary_op, _Inclusive,
+                       _UnaryOperation __unary_op, _Tp __init, _BinaryOperation __binary_op, _Inclusive __is_inclusive,
                        /*is_vector=*/::std::true_type) noexcept
 {
 #if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
-    if (_Inclusive{} || !oneapi::dpl::__internal::__iterators_possibly_equal(__first, __result))
+    if (__is_inclusive || !oneapi::dpl::__internal::__iterators_possibly_equal(__first, __result))
     {
         return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
                                             _Inclusive());
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -206,19 +206,13 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
                        /*is_vector=*/::std::true_type) noexcept
 {
 #if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
-    const auto& __first_tmp = *__first;
-    const auto& __result_tmp = *__result;
-    // An in-place scan violates the intra-iteration dependency restriction of the omp scan directive. We must call our serial brick.
-    if (static_cast<const void*>(::std::addressof(__first_tmp)) !=
-        static_cast<const void*>(::std::addressof(__result_tmp)))
-    {
-        return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
-                                            _Inclusive());
-    }
-#endif
+    return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
+                                        _Inclusive());
+#else
     // We need to call serial brick here to call function for inclusive and exclusive scan that depends on _Inclusive() value
     return __internal::__brick_transform_scan(__first, __last, __result, __unary_op, __init, __binary_op, _Inclusive(),
                                               /*is_vector=*/::std::false_type());
+#endif
 }
 
 template <class _RandomAccessIterator, class _OutputIterator, class _UnaryOperation, class _Tp, class _BinaryOperation,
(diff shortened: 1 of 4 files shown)


timmiesmith (dismissed):
LGTM, but wait for another review please.

SergeyKopienko (changes requested):
I think that this solution may be incorrect.
We have another approach for solve this task: for details please see ```__iterators_possibly_equal``` and it's usage.
If sycl iterators are possible in this code here, the solution from this PR, of course, shouldn't work.
Sorry, if I am wrong.

SergeyKopienko:
@danhoeflinger, up to you: if our __iterators_possibly_equal is useful here, please move it into any place.

danhoeflinger:
Something has seemed out of place with this, so I looked at this a bit further... 

I don't think we support `sycl_iterator` at this location in this code.  The normal way to translate a `sycl_iterator` into something which can be dereferenced is via sending it through `__get_sycl_range` which detects the `sycl_iterator` and creates an `all_view` from it which holds internally an accessor and can be used to get access to the underlying data in the sycl buffer.  

You can see that there is no such code which does this prior to dereferencing the iterators below this location, so they are not able to be `sycl_iterators`.  

I believe I have confirmed this by calling `oneapi::dpl::inclusive_scan` with a `par_unseq` policy and `oneapi::dpl::begin(buf)` and `oneapi::dpl::end(buf)`.  This results in compiler errors in the host parallel implementation when it attempts to dereference the input iterator.  I'm not sure if this is documented, but it does not seem that we support host parallel execution policy in combination with `sycl_iterator` inputs.

There may still be a better way to compare "in-place-ness" than using the address of the dereferenced element, but I don't think there is value in supporting sycl iterators at this code location.

danhoeflinger:
Following up on the documentation front:  
https://oneapi-src.github.io/oneDPL/parallel_api/pass_data_algorithms.html
`oneapi:dpl::begin` and `oneapi::dpl::end` functions are only mentioned in the context of passing data to algorithms executed with a device policy.  I think this is proper, and the implementation is in line with this.

Also it should be noted that those functions return an "unspecified type", and the term `sycl_iterator` is internal only.

mmichel11:
> Following up on the documentation front: https://oneapi-src.github.io/oneDPL/parallel_api/pass_data_algorithms.html `oneapi:dpl::begin` and `oneapi::dpl::end` functions are only mentioned in the context of passing data to algorithms executed with a device policy. I think this is proper, and the implementation is in line with this.
> 
> Also it should be noted that those functions return an "unspecified type", and the term `sycl_iterator` is internal only.

Thanks for taking a look into this. I will take a deeper look as well to verify what you have found.

To me it seems that pointer comparison of the address of dereferenced input and output begin iterators is the most reliable way to check the in-place property. It is my understanding that iterator comparison is only valid within the same sequence of data and comparing iterators from two different different sequences (the standard out-of-place case) is undefined behavior. Pointer equality and inequality comparison is valid as long as the pointer types are the same which is ensured through the cast to `const void*`. 

As you said there may be a better way to check if the scan is in-place, but I am unsure what this would be.

mmichel11:
I have taken a second look at this and @danhoeflinger is correct that `oneapi::dpl::begin` and `oneapi::dpl::end` over SYCL buffers cannot be used in this function (and similar functions intended for host execution only). We do not need to worry about a `sycl_iterator` being passed as input here.

@danhoeflinger I also took a look into `zip_iterator` usage in this code and believe this is still a gap that must be covered. In the in-place case, `__first_tmp` and `__result_tmp` will be different tuples with different addresses, but the corresponding tuple elements will each start at the same address. I will look into this.

include/oneapi/dpl/pstl/numeric_impl.h
```diff
@@ -206,13 +206,19 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
                        /*is_vector=*/::std::true_type) noexcept
 {
 #if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
-    return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
-                                        _Inclusive());
-#else
+    const auto& __first_tmp = *__first;
+    const auto& __result_tmp = *__result;
+    // An in-place scan violates the intra-iteration dependency restriction of the omp scan directive. We must call our serial brick.
+    if (static_cast<const void*>(::std::addressof(__first_tmp)) !=
```
MikeDvorskiy, line 212: Unfortunately,  a comparison of pointers is not reliable here, due to a dereferenced result might be a kind of proxy objects... For example, in case of output transform iterator or something like that,

akukanov:
Actually, our code in the `__simd_scan` brick is not friendly for in-place scan. I mean, even if we remove the SIMD pragmas and execute serially, the result will be incorrect for in-place exclusive scan.
Therefore, I think the issue should be addressed in the brick itself, and not in disabling it for in-place scan. The code there should be like this:
```
    _ONEDPL_PRAGMA_SIMD_SCAN(+ : __init)
    for (_Size __i = 0; __i < __n; ++__i)
    {
        _Tp __tmp = __init;
        _ONEDPL_PRAGMA_SIMD_EXCLUSIVE_SCAN(__init)
        {
            __init += __unary_op(__first[__i]);
            __result[__i] = __tmp;
        }
    }
```
I recommend also to check the serial brick, just in case, as it might have a similar error. Also, we need to prepare a fix for the LLVM PSTL upstream implementation.

mmichel11:
> Actually, our code in the `__simd_scan` brick is not friendly for in-place scan. I mean, even if we remove the SIMD pragmas and execute serially, the result will be incorrect for in-place exclusive scan. Therefore, I think the issue should be addressed in the brick itself, and not in disabling it for in-place scan. The code there should be like this:
> 
> ```
>     _ONEDPL_PRAGMA_SIMD_SCAN(+ : __init)
>     for (_Size __i = 0; __i < __n; ++__i)
>     {
>         _Tp __tmp = __init;
>         _ONEDPL_PRAGMA_SIMD_EXCLUSIVE_SCAN(__init)
>         {
>             __init += __unary_op(__first[__i]);
>             __result[__i] = __tmp;
>         }
>     }
> ```
> 
> I recommend also to check the serial brick, just in case, as it might have a similar error. Also, we need to prepare a fix for the LLVM PSTL upstream implementation.

Thanks for the assistance. I will prepare these changes to the bricks, so we can

> _Tp __tmp = __init;



> Actually, our code in the `__simd_scan` brick is not friendly for in-place scan. I mean, even if we remove the SIMD pragmas and execute serially, the result will be incorrect for in-place exclusive scan. Therefore, I think the issue should be addressed in the brick itself, and not in disabling it for in-place scan. The code there should be like this:
> 
> ```
>     _ONEDPL_PRAGMA_SIMD_SCAN(+ : __init)
>     for (_Size __i = 0; __i < __n; ++__i)
>     {
>         _Tp __tmp = __init;
>         _ONEDPL_PRAGMA_SIMD_EXCLUSIVE_SCAN(__init)
>         {
>             __init += __unary_op(__first[__i]);
>             __result[__i] = __tmp;
>         }
>     }
> ```
> 
> I recommend also to check the serial brick, just in case, as it might have a similar error. Also, we need to prepare a fix for the LLVM PSTL upstream implementation.

Thanks for the assistance. I have confirmed this fix resolves the issue. I will update this today and check the serial bricks as well, so we can get this in for the milestone.

mmichel11:
> Actually, our code in the `__simd_scan` brick is not friendly for in-place scan. I mean, even if we remove the SIMD pragmas and execute serially, the result will be incorrect for in-place exclusive scan. Therefore, I think the issue should be addressed in the brick itself, and not in disabling it for in-place scan. The code there should be like this:
> 
> ```
>     _ONEDPL_PRAGMA_SIMD_SCAN(+ : __init)
>     for (_Size __i = 0; __i < __n; ++__i)
>     {
>         _Tp __tmp = __init;
>         _ONEDPL_PRAGMA_SIMD_EXCLUSIVE_SCAN(__init)
>         {
>             __init += __unary_op(__first[__i]);
>             __result[__i] = __tmp;
>         }
>     }
> ```
> 
> I recommend also to check the serial brick, just in case, as it might have a similar error. Also, we need to prepare a fix for the LLVM PSTL upstream implementation.

I have made a similar change to what you have suggested. I have noticed a correctness issue in CI and locally when using the proposed code change. By specifying `__tmp` as the list-item, we ensure the intra-iteration dependency between its assignment and the assignment of `__result[__i]` is properly handled. Making this change fixes the encountered issue.

include/oneapi/dpl/pstl/unseq_backend_simd.h
```diff
@@ -548,12 +548,16 @@ typename ::std::enable_if<is_arithmetic_plus<_Tp, _BinaryOperation>::value, ::st
...
-    _ONEDPL_PRAGMA_SIMD_SCAN(+ : __init)
+    _Tp __cur_scan_value;
+    _ONEDPL_PRAGMA_SIMD_SCAN(+ : __cur_scan_value)
     for (_Size __i = 0; __i < __n; ++__i)
     {
-        __result[__i] = __init;
-        _ONEDPL_PRAGMA_SIMD_EXCLUSIVE_SCAN(__init)
-        __init += __unary_op(__first[__i]);
+        __cur_scan_value = __init;
+        _ONEDPL_PRAGMA_SIMD_EXCLUSIVE_SCAN(__cur_scan_value)
```
akukanov, line 556: @mmichel11, I read your description, yet I am not convinced that using `__cur_scan_value` in the scan directive is correct. Meanwhile, while this change eliminates an intra-iteration dependency, it creates a cross-iteration dependency on `__init` that is also prohibited.

Eventually, I think we need to use what practically works (even if the code formally violates the OpenMP specification), and it might be compiler-dependent. I will take a look at the CI failures, and try to make some judgement about that all. Maybe we will need to ask the compiler team experts for a recommendation.

mmichel11: Yes, I see the cross-iteration dependency now. Furthermore, the [OpenMP 5.2 spec](https://www.openmp.org/spec-html/5.2/openmpse28.html#x92-960005.6) states that list items should not be modified in the scan phase which my change also violates.

I am unsure if there is an OpenMP compliant way to do this. I agree with your suggestion that we should ask the compiler team for a recommendation on this case.

danhoeflinger: To add some of our offline discussion into this conversation to the github PR,  I had suggested reforming it to cache the element of `__first[__i]` rather than the accumulation variable, and adding the temporary as a list item.
```
    _ONEDPL_PRAGMA_SIMD_SCAN(+ : __init, __tmp)
    for (_Size __i = 0; __i < __n; ++__i)
    {
        __tmp = __first[__i];
        __result[__i] = __init;
        _ONEDPL_PRAGMA_SIMD_EXCLUSIVE_SCAN(__init, __tmp)
        {
            __init += __unary_op(__tmp);
        }
    }
```
This *almost* works, as with the extra list item `__tmp` it allows `__tmp` to have an intra-iteration dependency across the pragma, but it also disobeys the provision mentioned above that "list items should not be modified in the scan phase"

---
I don't see a way to fit this into the simd scan specification (though in practice, the implementation seems to currently work through our testing). 

Effectively 4 things need to happen in an iteration:

1. capture `__first[i]`
2. capture previous running accumulation `__init`
3. write (2) to `__result[i]`
4. increment running accumulation with (1)

The dependencies are 
[1->4], [2->3], and when in-place, we add [1->3] so we don't write over our input before capturing it.  

For omp simd exclusive scan iteration we have: 
scan phase (apply / write the accumulation)
#pragma 
input phase (do accumulation)

I believe the scan phase should include (3), and the input phase should include (4).  However, with the paraphrased provisions, 
(a) intra-iterations between scan and input phase must not exist except with list items
(b) list items should not be modified in the scan phase

This means that (1) must also be in the input phase, which would force it to be after (3), breaking the dependency added from in-place.

akukanov: I agree that in-place exclusive scan cannot be expressed in a spec compliant way. However, I believe it is vectorizable, with the algorithm being like this (for the vector of size 4):

1. capture `__first[i..i+3]` into a vector `v[0..3]`
2. build a vector `sv = {__init, __init+v[0], __init+v[0]+v[1], __init+v[0]+v[1]+v[2]}`
3. write `sv` to `__result[i..i+3]
4. `__init += v[0]+v[1]+v[2]+v[3]`
5. `i+=4`

and it sort of matches the code that Dan has suggested:

```
for( int i=0; i<N; ++i) {
    auto v = __first[i];  // input phase?
    __result[i] = __init; // scan phase
    __init += v;          // input phase
}
```
Also note that in the above vectorization algorithm and a one for inclusive scan would only differ in the step 2, by shifting values in `sv` to the "left", and appending by the same value as computed in step 4.

mmichel11: Yes, I agree with your proposed algorithm that this code is vectorizable. If we are unable to have a guarantee in the spec (or an update to it), then I believe this could be something we could look to write ourselves with OpenMP simd directives. I also have messaged a contact on the compiler team to see if there are are any compiler-specific guarantees about this behavior.

In the short-term, what is the best step forward? Should we leave the known limitation with a release note so vectorized code can still be ran for the out-of-place case for exclusive_scan, or would it be best to disable this to ensure correctness in the in-place case? This was listed as a known issue for the previous release with the icpc compiler.

akukanov: > In the short-term, what is the best step forward?

It's clear that we will not fix `__simd_scan` for this release. Let's try to disable it for the in-place exclusive scan (I believe that the inclusive scan should be fine, correct me if not). But we should use an approach similar to `__iterators_possibly_equal `, instead of comparing addresses.

mmichel11: In that case, I think it is best to move `__iterators_possibly_equal` into the general utils header and directly invoke it as it provides a general iterator equality case we can reuse here. 

I have made this change.

include/oneapi/dpl/pstl/numeric_impl.h
```diff
@@ -206,13 +206,16 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
                        /*is_vector=*/::std::true_type) noexcept
 {
 #if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
-    return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
-                                        _Inclusive());
-#else
+    if (::std::is_same_v<_Inclusive, ::std::true_type> ||
```
SergeyKopienko, line 209: From my point of view better to write something like ```_Inclusive{}```

mmichel11: Done

danhoeflinger: We deliberately throw out the incoming argument with type `_Inclusive`, then here we would create a new one using the type.  Should we not just name the incoming argument `__is_inclusive` and use it?

include/oneapi/dpl/pstl/unseq_backend_simd.h
danhoeflinger: I created an issue for implementation of a vectorized brick, #1190 which is now outside the scope of this PR. So I am resolving this thread of discussion for this PR.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -646,6 +646,61 @@ __dpl_ceiling_div(_T1 __number, _T2 __divisor)
...
+// All with implemented operator ==
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable<
+    _Iterator1, _Iterator2,
+    std::void_t<decltype(::std::declval<::std::decay_t<_Iterator1>>() == ::std::declval<::std::decay_t<_Iterator2>>())>>
+    : std::true_type
+{
+};
+
+#if _ONEDPL_BACKEND_SYCL
```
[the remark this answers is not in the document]
mmichel11: Could there ever be a need to use `__iterators_possible_equal` with the sycl iterator specialization elsewhere in the codebase? If so, I think it would be best to keep its implementation in a utility header, so it could be used in other places without including `numeric_impl_hetero.h`. Is the concern with adding the SYCL specific code into `utils.h`?

Do any others have an opinion on this?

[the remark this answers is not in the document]
SergeyKopienko: The idea is that if we use sycl iterator, the header file `include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h` already included. For other types of iterator this code is not needed.

mmichel11:
I am seeing some failing CI that seems to be caused be a network issue during testing. It is unrelated to this PR.

include/oneapi/dpl/pstl/utils.h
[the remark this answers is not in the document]
mmichel11: Sounds good. I will make this change then.

include/oneapi/dpl/pstl/numeric_impl.h
danhoeflinger: I suppose we probably should also replace the `_Inclusive()` in the call itself with such a variable.

include/oneapi/dpl/pstl/utils.h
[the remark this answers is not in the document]
MikeDvorskiy: yes, SYCL specific code is allow from `.../hetero/dpcpp` level.

SergeyKopienko (approved):
LGTM

mmichel11:
@akukanov @SergeyKopienko @MikeDvorskiy One last thing I think should be discussed with usage of `__iterators_possibly_equal` in this place.

When comparing equality of [forward iterators](https://en.cppreference.com/w/cpp/named_req/ForwardIterator), "Equality and inequality comparison is defined over all iterators for the same underlying sequence". This means that for the in-place case we will correctly detect and handle this case as long as the iterators are equality comparable.

However, for the out-of-place case, I believe this is technically undefined behavior. @danhoeflinger and I had some discussion about this and in practice it seems that this is correctly handled due to iterator equality implementations, but it is not ensured in the standard. In the worst possible case, this would result in a false in-place detection with the non-vectorized codepath being invoked instead of the vectorized, but it seems unlikely given that it would require iterator equality from two separate containers to return true.

danhoeflinger:
> @akukanov @SergeyKopienko @MikeDvorskiy One last thing I think should be discussed with usage of `__iterators_possibly_equal` in this place.
> 
> When comparing equality of [forward iterators](https://en.cppreference.com/w/cpp/named_req/ForwardIterator), "Equality and inequality comparison is defined over all iterators for the same underlying sequence". This means that for the in-place case we will correctly detect and handle this case as long as the iterators are equality comparable.
> 
> However, for the out-of-place case, I believe this is technically undefined behavior. @danhoeflinger and I had some discussion about this and in practice it seems that this is correctly handled due to iterator equality implementations, but it is not ensured in the standard. In the worst possible case, this would result in a false in-place detection with the non-vectorized codepath being invoked instead of the vectorized, but it seems unlikely given that it would require iterator equality from two separate containers to return true.

Unless there is some way that calling the `==` operator itself causes a problem (like an exception), I think we can live with this.  Falsely detecting 2 independent iterators as in-place seems like something which is unlikely to happen in practice, and even if it did, it would just result in slower performance, not an incorrect result.

danhoeflinger (approved):
LGTM

SergeyKopienko (approved):
LGTM

PR #1039, opened 2023-08-04
Wait Container for async submissions

egfefey:
Added a wait container for async submissions. This is what gets returned by get_wait_list. Waiting on the container will wait on all waiters. The group of native sync objects can also be obtained from the container. (Using old terminology)


AnuyaWelling2801 (changes requested):

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_scheduler.h
```diff
@@ -85,22 +86,16 @@ namespace experimental {
...
-      waiters_.push(w);
+      waiters.add_waiter(w);
       return *w;
     }
 
     auto get_wait_list(){
-       std::list<async_wait_t*> wlist;
-       waiters_.pop_all(wlist);
-       return wlist;
+      return &waiters;
```
vossmjp, line 94: This doesn't seem to be destructive. The old implementation removed everything from the list, placing all of the active submission into the new wlist. This just directly returns a pointer to the container.

vossmjp (changes requested):
Semantics of get_wait_list seem to have changed.

PR #1040, opened 2023-08-07
[KT] API adjustments

akukanov:
- Adding a dedicated namespace under `experimental`
- Move generic kernel configuration parameters into a special class
- Changing an execution policy to a SYCL queue
- Some uglification
- A fix for a compilation error in the test

### DIFF
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
@@ -23,7 +23,7 @@
 #include <cstdint>
 #include <type_traits>
 
-namespace oneapi::dpl::experimental::esimd::impl
+namespace oneapi::dpl::experimental::kt::esimd::impl
 {
 template <typename T, ::std::enable_if_t<sizeof(T) <= sizeof(::std::uint32_t), int> = 0>
 constexpr ::std::uint32_t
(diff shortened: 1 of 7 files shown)


include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -40,63 +40,58 @@ __onesweep_process_size()
 }
 
 // TODO: allow calling it only for all_view (accessor) and guard_view (USM) ranges, views::subrange and sycl_iterator
-template <std::uint16_t WorkGroupSize, std::uint16_t DataPerWorkItem, bool IsAscending, std::uint32_t RadixBits,
-          typename _ExecutionPolicy, typename _Range,
-          oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0>
+template <bool _IsAscending, std::uint32_t _RadixBits, typename _KernelParam, typename _Range>
```
SergeyKopienko, line 43: May be better to have some predefined ```DefauleKernelParam``` and use it as
```typename _KernelParam=DefauleKernelParam``` ?

File: include/oneapi/dpl/experimental/kt/esimd_radix_sort.h [hunk 2]
```diff
@@ -40,63 +40,58 @@ __onesweep_process_size()
...
 void
-radix_sort(_ExecutionPolicy&& __exec, _Range&& __rng)
+radix_sort(sycl::queue __q, _Range&& __rng, _KernelParam __param)
```
SergeyKopienko, line 45: ... and then we may use here it's instance: ````_KernelParam __param = {}```

akukanov: See my answer above.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -40,63 +40,58 @@ __onesweep_process_size()
...
-} // oneapi::dpl::experimental::esimd::impl
+} // oneapi::dpl::experimental::kt::esimd::impl
 
-namespace oneapi::dpl::experimental::esimd
+namespace oneapi::dpl::experimental::kt::esimd
 {
-template <std::uint16_t WorkGroupSize, std::uint16_t DataPerWorkItem, bool IsAscending = true,
-          std::uint32_t RadixBits = 8, typename _ExecutionPolicy, typename _Range>
+
+template <bool _IsAscending = true, std::uint32_t _RadixBits = 8, typename _KernelParam, typename _Range>
```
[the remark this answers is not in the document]
akukanov: Maybe later.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
SergeyKopienko, line 45: Additional moment. Should we really have ```_KernelParam __param``` as run-time parameter?
Is it not enough to have it as template parameter?
Because all that we will use from this param is compile-time things...

akukanov: Technically a runtime parameter is not needed (and it has not state as you can see), but it allows to deduce the argument type and therefore not specify all preceding template arguments if you only want to configure kernel parameters and keep default values for sort order and radix bits.

akukanov: Though it is less needed for the internal function, and can be removed here if you prefer so.

SergeyKopienko (approved):
LGTM

PR #1043, opened 2023-08-09
Fix UB when async algorithms called without save result

SergeyKopienko:
In this PR we trying to fix the case with UB when we calls some async algorithm, receive result but doesn't save it on stack in some local variable:
 - in this case we can destroy some objects, which used in async algorithm implementation before async algorithm finished.

The example of problem code:
```
    const int n = 100;
    {
        sycl::buffer<int> x{n};

        auto my_policy = TestUtils::make_device_policy<class Copy1>(oneapi::dpl::execution::dpcpp_default);
        oneapi::dpl::experimental::copy_async(my_policy, oneapi::dpl::counting_iterator<int>(0),
                                                            oneapi::dpl::counting_iterator<int>(n),
                                                            oneapi::dpl::begin(x)); // x = [0..n]
        // UB is here (after exit from this scope) because we able to exit from scope
        // and destroy sycl::buffer x before copy_async finished
    }
```

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -534,6 +534,11 @@ class __future : private std::tuple<_Args...>
   public:
     __future(_Event __e, _Args... __args) : std::tuple<_Args...>(__args...), __my_event(__e) {}
     __future(_Event __e, std::tuple<_Args...> __t) : std::tuple<_Args...>(__t), __my_event(__e) {}
+    ~__future()
+    {
+        // TODO: we able to have exceptions in this place -> in this case we will have chrash
+        wait();
+    }
 
     auto
     event() const
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -536,7 +536,7 @@ class __future : private std::tuple<_Args...>
     __future(_Event __e, std::tuple<_Args...> __t) : std::tuple<_Args...>(__t), __my_event(__e) {}
     ~__future()
     {
-        // TODO: we able to have exceptions in this place -> in this case we will have chrash
+        // TODO: we able to have exceptions in this place -> in this case we will have crash
         wait();
     }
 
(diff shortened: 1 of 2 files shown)


rarutyun:
I would not say it's UB. It's a design choice we need to make for our async algorithms.

You example is not different if comparing with:

```cpp
constexpr std::size_t size = 100;
sycl::queue q;
{
    auto ptr = (int*)sycl::malloc_shared(size, q);
    q.submit([&](auto& cgh){
        cgh.parallel_for(size, [=](auto i) { ptr[i] = i; });
    });
    
    // Forget to wait be either way
    sycl::free(ptr, q);
}
```

The following code is slightly different by semantically illustrates the same issue:

```cpp
std::promise<std::size_t> p;

std::thread t;
{
    auto fut = p.get_future();
    std::vector a{1,2,3,4};
    t = std::thread([&](){
        std::size_t sum{};
        for (int i = 0; i < 20; ++i) {
            sum += a[i % a.size()];
        }
        p.set_value(sum);
    });
    // forget to wait on the fut and obtain the result
}
t.join();
```
What this code illustrates is both `sycl::event` and `std::future` (except special case) don't block on the destructor.
As far as I remember `tbb::task_group` also doesn't block (it's different example of async API but still)

When users are dealing with async API of any kind they should understand what they are doing. I am not sure that blocking on the destructor for the future is a good choice, especially when we think about having multiple backends for async API and providing interoperability between different futures. Non-blocking destructor seems like a choice that eventually gives us more freedom and opportunity to create a consistent API behavior-wise.

rarutyun:
> @rarutyun I understood what you mean. And agree with your examples. But we have one difference in our code (Is far as I understand it):
> 
> * in your example future / promise simple save result and managed all this for this mechanic;
> * in our code `__future` class are used for managing some resources like `sycl::buffer`.
>   And my point that it's not quite correct in corner case (when we don't save result of async call) and we should do something with this.

Those are implementation details. TBB-based async API would not have `sycl::buffer` in the returned future-like. I also believe not all our algorithms return the future with sycl::buffer within (it would be strange otherwise, or I don't understand something).

Our async API is still experimental and far from complete. Without the proper designing procedure and agreement I don't think we need this fix.

aleksei-fedotov:
One point not to forget about is that user may wanted to just make a "fire-and-forget" task without specifying the point when it needs to complete. Something like by saying "copy that to that and I don't care when it is going to be done", expecting that the runtime decide on its own when better to do that task.
Is such a use case of an interest to us?

akukanov:
Unfortunately, blocking in the destructor of a future is not an option at all. We have already got practical feedback from users that just want to call an async algorithm and do not bother with the lifetime of events/futures etc. Any problem with the lifetime management of internal storage should be solved in different ways, for example via a host task tied to the completion of the kernel(s).

PR #1044, opened 2023-08-09
[KT] Remove data() function from range view classes

SergeyKopienko:
This PR created for discussion how to avoid two new ```data()``` methods in esimd radix sort code.


SergeyKopienko:
@MikeDvorskiy what do you think about this?

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -245,6 +245,9 @@ class guard_view
     // Unified funciton to get pointer or accessor to use inside ESIMD kernels
     _Iterator data()
```
SergeyKopienko, line 246: I believe we are able to remove this function only

SergeyKopienko:
@MikeDvorskiy, the begin() isn't enough due existing restrictions from https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-0/explicit-simd-sycl-extension.html : we are unable to work with data address from sycl::accessor together with esimd implementation. @dmitriy-sobolev early tried to do this.

PR #1046, opened 2023-08-10
[KT] Skip all test cases with an error "LLVM ERROR: SLM size exceeds target limits"

SergeyKopienko:
In this PR skip all test cases with an error "LLVM ERROR: SLM size exceeds target limits"

### DIFF
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
@@ -33,12 +33,12 @@ radix_sort(sycl::queue __q, _Range&& __rng, _KernelParam __param)
 {
     static_assert(_RadixBits == 7 || _RadixBits == 8);
 
-    static_assert(param.data_per_workitem == 32 || param.data_per_workitem == 64 || param.data_per_workitem == 96 ||
-                  param.data_per_workitem == 128 || param.data_per_workitem == 160 || param.data_per_workitem == 192 ||
-                  param.data_per_workitem == 224 || param.data_per_workitem == 256 || param.data_per_workitem == 288 ||
-                  param.data_per_workitem == 320 || param.data_per_workitem == 352 || param.data_per_workitem == 384 ||
-                  param.data_per_workitem == 416 || param.data_per_workitem == 448 || param.data_per_workitem == 480 ||
-                  param.data_per_workitem == 512);
+    static_assert(__param.data_per_workitem == 32 || __param.data_per_workitem == 64 || __param.data_per_workitem == 96 ||
+                  __param.data_per_workitem == 128 || __param.data_per_workitem == 160 || __param.data_per_workitem == 192 ||
+                  __param.data_per_workitem == 224 || __param.data_per_workitem == 256 || __param.data_per_workitem == 288 ||
+                  __param.data_per_workitem == 320 || __param.data_per_workitem == 352 || __param.data_per_workitem == 384 ||
+                  __param.data_per_workitem == 416 || __param.data_per_workitem == 448 || __param.data_per_workitem == 480 ||
+                  __param.data_per_workitem == 512);
 
     const ::std::size_t __n = __rng.size();
     assert(__n > 1);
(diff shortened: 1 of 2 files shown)


test/kt/esimd_radix_sort.cpp
```diff
@@ -37,6 +37,7 @@
 #include <iostream>
 #include <cmath>
 #include <limits>
+#include <string_view>
```
dmitriy-sobolev, line 40: What is the purpose of this include?

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -31,6 +31,15 @@ template <bool _IsAscending, std::uint8_t _RadixBits, typename _KernelParam, typ
 sycl::event
 radix_sort(sycl::queue __q, _Range&& __rng, _KernelParam __param)
 {
+    static_assert(_RadixBits == 7 || _RadixBits == 8);
```
dmitriy-sobolev, line 34: I would suggest leaving only 8 for now. 7 has not been verified with the majority of configurations.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -31,6 +31,15 @@ template <bool _IsAscending, std::uint8_t _RadixBits, typename _KernelParam, typ
...
+
+    static_assert(param.data_per_workitem == 32 || param.data_per_workitem == 64 || param.data_per_workitem == 96 ||
+                  param.data_per_workitem == 128 || param.data_per_workitem == 160 || param.data_per_workitem == 192 ||
+                  param.data_per_workitem == 224 || param.data_per_workitem == 256 || param.data_per_workitem == 288 ||
+                  param.data_per_workitem == 320 || param.data_per_workitem == 352 || param.data_per_workitem == 384 ||
+                  param.data_per_workitem == 416 || param.data_per_workitem == 448 || param.data_per_workitem == 480 ||
+                  param.data_per_workitem == 512);
```
dmitriy-sobolev: Can you use `%` for simplicity?

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
SergeyKopienko: Fixed

test/kt/esimd_radix_sort.cpp
SergeyKopienko: Removed

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
SergeyKopienko: As for me better to have full list of available values without any calculations.

test/kt/esimd_radix_sort.cpp
```diff
@@ -386,38 +388,53 @@ void test_general_cases(std::size_t size, KernelParam param)
 }
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
+template <typename T, typename KernelParam>
+bool
+can_run_test(KernelParam param)
+{
+    // uint64_t, int64_t, double - skip tests with error: LLVM ERROR: SLM size exceeds target limits
+    if (sizeof(T) == sizeof(::std::uint64_t) && param.data_per_workitem >= 288 && param.workgroup_size == 64)
```
dmitriy-sobolev, line 396: Could you investigate how much memory is used by the implementations and compare it against available memory, so the check is more complete? 

`info::device::local_mem_size` device descriptor shows available memory. Our experiments have shown that compiler can implicitly use some SLM, so we can utilize only a part of the device SLM. 
Example: 
https://github.com/oneapi-src/oneDPL/blob/8528820d414921ce04169157e212d5bc6421ba1e/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h#L912

An example is capped to only a half of SLM, I consider it is too strict for ESIMD kernels basing on testing experience. Could you find the most appropriate coefficient?

dmitriy-sobolev (approved):

PR #1050, opened 2023-08-11
[KT] Extend supported WorkGroupSize

dmitriy-sobolev:
Pin histogram kernel configuration, because it is not as versatile as onesweep kernel.  This is needed to support more WorkGroupSize options.


SergeyKopienko (approved):
LGTM

test/kt/CMakeLists.txt
```diff
@@ -15,7 +15,7 @@ function(generate_esimd_radix_sort_tests)
 
     set(_key_type_all "char" "uint16_t" "int" "uint64_t" "float" "double")
     set(_data_per_work_item_all "32" "64" "96" "128" "160" "192" "224" "256" "288" "320" "352" "384" "416" "448" "480" "512")
-    set(_work_group_size_all "64")
+    set(_work_group_size_all "32" "64")
```
SergeyKopienko, line 18: @dmitriy-sobolev Looks like with 32 value we have a lot of new failed tests - wrong test results.
My proposal to remove this 32 value until we will ready to fix new errors in tests.

PR #1051, opened 2023-08-14
[KT] Dev/skopienko/esimd/esimd radix sort use large grf

SergeyKopienko:
Looks like this solution doesn't work correctly:
- I have an error ```Exception: Native API failed. Native API returns: -1 (PI_ERROR_DEVICE_NOT_FOUND) -1 (PI_ERROR_DEVICE_NOT_FOUND)``` at all tests.


include/oneapi/dpl/experimental/kt/esimd_radix_sort_onesweep.h
```diff
@@ -38,7 +39,8 @@ void global_histogram(sycl::nd_item<1> idx, size_t __n, const InputT& input, uin
 
     // Cap the number of histograms to reduce in SLM per input range read pass
     // due to excessive GRF usage for thread-local histograms
-    constexpr uint32_t STAGES_PER_BLOCK = sizeof(_KeyT) < 4? sizeof(_KeyT) : 4;
+    //constexpr uint32_t STAGES_PER_BLOCK = sizeof(_KeyT) < 4? sizeof(_KeyT) : 4;
+    constexpr uint32_t STAGES_PER_BLOCK = sizeof(_KeyT);
```
SergeyKopienko, line 43: @dmitriy-sobolev, is it correct way to exclude your workaround?

dmitriy-sobolev: Yes, there will be only one stage with that change as it was prior introduction of the stages...

dmitriy-sobolev:
@SergeyKopienko, does it still make sense to keep this draft PR opened?

PR #1052, opened 2023-08-14
Split set.pass to separate tests

SergeyKopienko:
In this PR we split ```set.pass``` test to separate tests to avoid timeout error during testing:
- ```set.pass.cpp```;
- ```set_diff.pass.cpp```;
- ```set_intersection.pass.cpp```;
- ```set_symm_diff.pass.cpp```;
- ```set_union.pass.cpp```.


timmiesmith (dismissed):
LGTM. Thank you.

test/parallel_api/algorithm/alg.sorting/alg.set.operations/set.pass.h
```diff
@@ -0,0 +1,365 @@
...
+#include <cmath>
+#include <chrono>
+
+#if  !defined(_PSTL_TEST_SET_UNION) && !defined(_PSTL_TEST_SET_DIFFERENCE) && !defined(_PSTL_TEST_SET_INTERSECTION) &&\
+     !defined(_PSTL_TEST_SET_SYMMETRIC_DIFFERENCE)
+#define _PSTL_TEST_SET_UNION
+#define _PSTL_TEST_SET_DIFFERENCE
+#define _PSTL_TEST_SET_INTERSECTION
+#define _PSTL_TEST_SET_SYMMETRIC_DIFFERENCE
+#endif
```
dmitriy-sobolev, line 31: These macros are not required anymore, because there are separated files for each algorithm. 
Some time ago, they were added to build a test for a specific algorithm to determine an algorithm-related pass rate.

Could you remove them?

test/parallel_api/algorithm/alg.sorting/alg.set.operations/set.pass.cpp
```diff
@@ -13,357 +13,7 @@
...
-struct test_non_const_set_union
-{
-    template <typename Policy, typename InputIterator, typename OutputInterator>
-    void
-    operator()(Policy&& exec, InputIterator input_iter, OutputInterator out_iter)
-    {
-        set_union(exec, input_iter, input_iter, input_iter, input_iter, out_iter, non_const(::std::less<T>()));
-    }
-};
+#include "set.pass.h"
```
dmitriy-sobolev, line 16: There are no files named with "*.pass.h" pattern. Could you remove "pass" from the name?

dmitriy-sobolev (commented):
Thanks for the changes. Here is a minor comment. 
Could you rename the tests so that the file names match the algorithms? I guess a developer would look for "set_symmetric_difference" to test that algorithm rather than "set_sym_diff".

test/parallel_api/algorithm/alg.sorting/alg.set.operations/set.pass.cpp
SergeyKopienko: In this case we will have file with name ```set.h```, it's too much from my point of view :-)

dmitriy-sobolev: I see your point. There might some clash with the header names (I guess you meant that). What do you think about set_utils.h?

dmitriy-sobolev: Or "set_common.h"?

SergeyKopienko: Fixed

test/parallel_api/algorithm/alg.sorting/alg.set.operations/set.pass.h
SergeyKopienko: No: this macro still required in each test and simple moved from original file to new header file.

SergeyKopienko:
> Thanks for the changes. Here is a minor comment. Could you rename the tests so that the file names match the algorithms? I guess a developer would look for "set_symmetric_difference" to test that algorithm rather than "set_sym_diff".

Fixed: renamed.

test/parallel_api/algorithm/alg.sorting/alg.set.operations/set.pass.h
dmitriy-sobolev: I see that removing the macros from `set_common.h` is not straightforward, because they are now used in `test_set`.

Meanwhile, removing them from the newly created files, which are intended for a certain algorithm, is easy and makes sense (to me at least). There is no any application in building a test with a macro which turns it off entirely. It used to be useful when all the algorithms were tested in a single file - to filter algorithms to test.

You may also consider splitting `test_usm` function, so that it tests certain algorithm in a designated file. For example, `set_union.pass.cpp` has all `set_union` cases and so on. Then, `set.pass.cpp` would be not required. This is optional.

dmitriy-sobolev: It depends on a situation. Sometimes it is reasonable and sometimes it is not.

Some tests are entirely skipped when a certain backend is not available. For example, `buffer_wrapper.pass.cpp` is skipped when dpcpp backend is not available. This is done on purpose.

Having a macro e.g. `_PSTL_TEST_SET_UNION` to skip the whole `set_union` test is not applicable... These macros became not relevant after splitting a test for each algorithm. That is why I suggest removing them, at least from files designated for a specific algorithm.

test/parallel_api/algorithm/alg.sorting/alg.set.operations/set_common.h
```diff
@@ -0,0 +1,320 @@
...
+//
+// This file incorporates work covered by the following copyright and permission
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#include "support/test_config.h"
```
dmitriy-sobolev, line 15: Could you add a guard to the header?

SergeyKopienko: Fixed: guard added.

test/parallel_api/algorithm/alg.sorting/alg.set.operations/set.pass.h
SergeyKopienko: My proposal to do this in separate PR.

dmitriy-sobolev (approved):
LGTM. Please, wait until CI finishes and passes before merging.

PR #1054, opened 2023-08-15
Specialize work-group sizes for scan-based algorithms

adamfidel:
This PR allows us to change the work-group size and iterations per work-item for the scan family of algorithms. We choose values for these parameters that optimize performance for devices of interest.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -802,11 +802,11 @@ __pattern_mismatch(_ExecutionPolicy&& __exec, _Iterator1 __first1, _Iterator1 __
 // copy_if
 //------------------------------------------------------------------------
 
-template <typename _ExecutionPolicy, typename _Iterator1, typename _IteratorOrTuple, typename _CreateMaskOp,
+template <typename _ExecutionPolicy, typename _Iterator1, typename _IteratorOrTuple, typename _AlgoType, typename _CreateMaskOp,
           typename _CopyByMaskOp>
 oneapi::dpl::__internal::__enable_if_hetero_execution_policy<
     _ExecutionPolicy, ::std::pair<_IteratorOrTuple, typename ::std::iterator_traits<_Iterator1>::difference_type>>
-__pattern_scan_copy(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __last, _IteratorOrTuple __output_first,
+__pattern_scan_copy(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __last, _IteratorOrTuple __output_first, _AlgoType,
                     _CreateMaskOp __create_mask_op, _CopyByMaskOp __copy_by_mask_op)
 {
     using _It1DifferenceType = typename ::std::iterator_traits<_Iterator1>::difference_type;
@@ -824,7 +824,7 @@ __pattern_scan_copy(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __
 
     auto __res =
         __par_backend_hetero::__parallel_scan_copy(::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(),
-                                                   __buf2.all_view(), __n, __create_mask_op, __copy_by_mask_op);
+                                                   __buf2.all_view(), __n, _AlgoType{}, __create_mask_op, __copy_by_mask_op);
 
     ::std::size_t __num_copied = __res.get();
     return ::std::make_pair(__output_first + __n, __num_copied);
@@ -870,16 +870,17 @@ __pattern_partition_copy(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterato
 
     using _It1DifferenceType = typename ::std::iterator_traits<_Iterator1>::difference_type;
     using _ReduceOp = ::std::plus<_It1DifferenceType>;
+    using _AlgoType = ::std::integral_constant<::oneapi::dpl::__internal::__algorithm_type, ::oneapi::dpl::__internal::__algorithm_type::partition>;
 
     unseq_backend::__create_mask<_UnaryPredicate, _It1DifferenceType> __create_mask_op{__pred};
     unseq_backend::__partition_by_mask<_ReduceOp, /*inclusive*/ ::std::true_type> __copy_by_mask_op{_ReduceOp{}};
 
     auto __result = __pattern_scan_copy(
-        ::std::forward<_ExecutionPolicy>(__exec), __first, __last,
+        ::std::forward<_ExecutionPolicy>(__exec), __first, __last, 
         __par_backend_hetero::zip(
             __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::write>(__result1),
             __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::write>(__result2)),
-        __create_mask_op, __copy_by_mask_op);
+        _AlgoType{}, __create_mask_op, __copy_by_mask_op);
 
     return ::std::make_pair(__result1 + __result.second, __result2 + (__last - __first - __result.second));
 }
@@ -893,6 +894,7 @@ oneapi::dpl::__internal::__enable_if_hetero_execution_policy<_ExecutionPolicy, _
 __pattern_unique_copy(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __last, _Iterator2 __result_first,
                       _BinaryPredicate __pred, /*vector*/ ::std::true_type, /*parallel*/ ::std::true_type)
 {
+    using _AlgoType = ::std::integral_constant<::oneapi::dpl::__internal::__algorithm_type, ::oneapi::dpl::__internal::__algorithm_type::unique>;
     using _It1DifferenceType = typename ::std::iterator_traits<_Iterator1>::difference_type;
     unseq_backend::__copy_by_mask<::std::plus<_It1DifferenceType>, oneapi::dpl::__internal::__pstl_assign,
                                   /*inclusive*/ ::std::true_type, 1>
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -293,11 +293,52 @@ struct __parallel_scan_submitter;
...
+            return 1;
+        case __algorithm_type::unique:
+            return 2;
+        case __algorithm_type::set_operation:
+            return 4;
+        case __algorithm_type::scan_copy:
+            return 16;
+        case __algorithm_type::transform_scan:
+            return 32;
+        default:
```
SergeyKopienko, line 312: Better to remove ```default``` branch and call ```return 16;``` after switch/case: we will have compile warnings or errors if we extend this enum.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -293,11 +293,52 @@ struct __parallel_scan_submitter;
...
+        case __algorithm_type::transform_scan:
+            return 128;
+        case __algorithm_type::scan_copy:
+            return 256;
+        case __algorithm_type::unique:
+        case __algorithm_type::set_operation:
+            return 512;
+        case __algorithm_type::partition:
+        default:
+            return ::std::numeric_limits<::std::size_t>::max();
```
SergeyKopienko, line 332: Could you please explain the goal of return ```::std::numeric_limits<::std::size_t>::max()``` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -310,7 +351,10 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         auto __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
         // get the work group size adjusted to the local memory limit
         // TODO: find a way to generalize getting of reliable work-group sizes
-        ::std::size_t __wgroup_size = oneapi::dpl::__internal::__slm_adjusted_work_group_size(__exec, sizeof(_Type));
+        ::std::size_t __wgroup_size_max =
+            oneapi::dpl::__internal::__slm_adjusted_work_group_size(__exec, sizeof(_Type));
+
+        ::std::size_t __wgroup_size = ::std::min(__wgroup_size_max, __preferred_wgroup_size(_AlgoType::value));
```
SergeyKopienko, line 357: May we pack this call ```std::min``` into ```__preferred_wgroup_size``` function and pass ```__wgroup_size_max``` into them?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -322,8 +366,20 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
...
-        constexpr decltype(__wgroup_size) __iters_per_witem = 16;
+#ifdef _ONEDPL_SCAN_ITER_SIZE
+        constexpr decltype(__wgroup_size) __iters_per_witem = _ONEDPL_SCAN_ITER_SIZE;
+#else
+        constexpr decltype(__wgroup_size) __iters_per_witem = __iters_per_item(_AlgoType::value);
+#endif
+
+#ifdef _ONEDPL_SCAN_TUNING_MODE
+        // Override wgsize if performing a tuning sweep
+        if (const char* __wgroup_size_override = ::std::getenv("_ONEDPL_SCAN_WG_SIZE"))
```
SergeyKopienko, line 377: As far as I found, this is the first call of ```std::getenv``` in our code.
May be better to avoid it and make this logic somehow in```CMakeLists.txt``` file?

include/oneapi/dpl/pstl/utils.h
```diff
@@ -48,6 +48,16 @@ namespace dpl
...
 {
 
+enum class __algorithm_type
+{
+    transform_scan,
+    set_operation,
+    scan_copy,
+    partition,
+    unique,
+    other
```
SergeyKopienko, line 58: Looks like this enum value ```other``` not used in your patch in other places. Is this enum value really required?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: It is an artifact from when I was testing with different work-group sizes. I don't see much much benefit from leaving it in the code, so I will remove it along with the other override macro for the iterations per work-item.

include/oneapi/dpl/pstl/utils.h
reble: Also we might want to give elements explicit integer values. Just as an extra caution for future modifications.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: The idea here is that the preferred work-group size is the largest allowed by the device. It is clipped down using `std::min` after this call.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: We could, but we would have to modify the switch statement. Instead of returning the number, we would update a local variable and them at the end of the function, we would return the min of that variable and the maximum work-group size. I do not know if that is cleaner.

include/oneapi/dpl/pstl/utils.h
adamfidel: Good points for both. I will remove `other` and give explicit values for the enums.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1621,6 +1625,8 @@ __pattern_hetero_set_op(_ExecutionPolicy&& __exec, _ForwardIterator1 __first1, _
     using _MaskAssigner = unseq_backend::__mask_assigner<2>;
     using _InitType = unseq_backend::__no_init_value<_Size1>;
     using _DataAcc = unseq_backend::walk_n<_ExecutionPolicy, oneapi::dpl::__internal::__no_op>;
+    using _AlgoType = ::std::integral_constant<::oneapi::dpl::__internal::__algorithm_type,
+                                               ::oneapi::dpl::__internal::__algorithm_type::set_operation>;
```
danhoeflinger, line 1629: It seems like this and similar instances would benefit from a unified helper to shorten it to require only the second template argument.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -836,31 +880,29 @@ struct __invoke_single_group_copy_if
...
                 _SizeType, __num_elems_per_item, __wg_size, false,
-                   oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
+                oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
                     __scan_copy_single_wg_kernel<::std::integral_constant<::std::uint16_t, __wg_size>,
-                                                ::std::integral_constant<::std::uint16_t, __num_elems_per_item>,
-                                                /* _IsFullGroup= */ std::false_type, _CustomName>>
-                >()(
+                                                 ::std::integral_constant<::std::uint16_t, __num_elems_per_item>,
+                                                 /* _IsFullGroup= */ std::false_type, _CustomName>>>()(
                 __exec, ::std::forward<_InRng>(__in_rng), ::std::forward<_OutRng>(__out_rng), __n, _InitType{},
```
danhoeflinger, line 896: It seems like these are formatting-only changes? Are they intentional?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -879,7 +881,7 @@ __pattern_partition_copy(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterato
         __par_backend_hetero::zip(
             __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::write>(__result1),
             __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::write>(__result2)),
-        __create_mask_op, __copy_by_mask_op);
+        _AlgoType{}, __create_mask_op, __copy_by_mask_op);
```
danhoeflinger, line 884: I don't really like passing the default constructed version of `_AlgoType` and discarding it, but it is of course better than specifying all the template arguments.

Putting `_AlgoType` first in the template argument list is an option, and there are other instances of doing this in oneDPL (`unseq_backend::__scan` and `__brick_fill`), but I'm not sure if it is the right call.

danhoeflinger (commented):
I think something like this is necessary for optimal performance while keeping shared code for many algorithms, and this is a good first step.  I believe it will need infrastructure to support it in a sustainable way to keep it beneficial.  

As this progresses I think we need to consider how to make this painless to update, and extend.  I expect we will ultimately want to have more than some "best" value across all hardware and software.  Understanding how to provide information to this system to select best values will be important.  Those sorts of changes would likely come in a separate PR.

For this PR, some minor changes and questions:

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -293,11 +293,51 @@ struct __parallel_scan_submitter;
 template <typename _CustomName, typename... _PropagateScanName>
 struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name<_PropagateScanName...>>
 {
+    static constexpr int
+    __iters_per_item(::oneapi::dpl::__internal::__algorithm_type __algo_type)
```
mmichel11, line 297: We might want to think about how we wish to store tuning parameters and maybe consider finding a way to tag / label the specific architecture they are tuned for. In the future, we may have a new set of best performing magic numbers for a newer architecture and may need to choose which to use.

timmiesmith:
Closing this in favor of the scan-based algorithm improvements implemented in the reduce-then-scan approach.

PR #1055, opened 2023-08-15
[Dynamic Selection] Introducing policy traits

AnuyaWelling2801:
1.Type traits defined by policies 
2. Type traits defined by scheduler
3. Added a policy traits class
4. Made changes to use policy_traits<Policy>::trait_name where its needed
5. PR does not add the submission_group_type class, since it makes sense to add it after #1039 is merged.


vossmjp (approved):
Looks good as a base for upcoming changes.

PR #1056, opened 2023-08-16
Round robin updates

egfefey:
This is to ensure that the round robin changes are on the right track.
Notable updates are the new names (without expanding out the _type yet), use of a boolean passed to the backend for deferred initialization.

### DIFF
--- a/include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
+++ b/include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
@@ -31,75 +31,71 @@ namespace experimental{
   private:
     std::shared_ptr<Backend> backend_;
 
-    struct unit_t{
+    struct state_t{
         resource_container_t resource_;
         resource_container_size_t num_contexts_;
         std::atomic<resource_container_size_t> next_context_;
 	int offset_;
-	bool deferred_;
     };
 
-    std::shared_ptr<unit_t> unit_;
+    std::shared_ptr<state_t> state_;
   public:
-    round_robin_policy(int offset=0, bool deferred=false) : backend_{std::make_shared<Backend>()}, unit_{std::make_shared<unit_t>()}  {
-      unit_->resource_ = get_resources();
-      unit_->num_contexts_ = unit_->resource_.size();
-      unit_->offset_ = offset;
-      unit_->next_context_ = unit_->offset_;
-      unit_->deferred_ = deferred;
-    }
+    round_robin_policy(int offset=0) : backend_{std::make_shared<Backend>()}, state_{std::make_shared<state_t>()}  {
+      state_->resource_ = get_resources();
+      state_->num_contexts_ = state_->resource_.size();
+      state_->offset_ = offset;
+      state_->next_context_ = state_->offset_;
+  }
 
-    round_robin_policy(resource_container_t u, int offset=0, bool deferred=false) : backend_{std::make_shared<Backend>()}, unit_{std::make_shared<unit_t>()}  {
+    round_robin_policy(resource_container_t u, int offset=0) : backend_{std::make_shared<Backend>()}, state_{std::make_shared<state_t>()}  {
       backend_->initialize(u);
-      unit_->resource_ = get_resources();
-      unit_->num_contexts_ = unit_->resource_.size();
-      unit_->offset_ = offset;
-      unit_->next_context_ = unit_->offset_;
-      unit_->deferred_ = deferred;
+      state_->resource_ = get_resources();
+      state_->num_contexts_ = state_->resource_.size();
+      state_->offset_ = offset;
+      state_->next_context_ = state_->offset_;
     }
 
     template<typename ...Args>
-    round_robin_policy_impl(Args&&... args) : backend_{std::make_shared<Backend>(std::forward<Args>(args)...)}, unit_{std::make_shared<unit_t>()} {
-      unit_->resource_ = backend_->get_resources();
-      unit_->num_contexts_ = unit_->resource_.size();
-      unit_->next_context_ = 0;
+    round_robin_policy_impl(Args&&... args) : backend_{std::make_shared<Backend>(std::forward<Args>(args)...)}, state_{std::make_shared<state_t>()} {
+      state_->resource_ = backend_->get_resources();
+      state_->num_contexts_ = state_->resource_.size();
+      state_->next_context_ = 0;
     }
 
     auto get_resources() const noexcept {
       return backend_->get_resources();
     }
 
-    void initialize(int offset=0, bool deferred=false) {
-      unit_->offset_ = offset;
-      unit_->deferred_ = deferred;
-      backend_->initialize(deferred);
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
AnuyaWelling2801, line 1: The old file - round_robin_policy_impl still exists and should be removed. 
This file name should be mentioned in include/oneapi/dpl/internal/dynamic_selection.h
Maybe changes need to be made the the tests too

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
```diff
@@ -0,0 +1,123 @@
...
+
+    void initialize(resource_container_t u, int offset=0, bool deferred=false) {
+      unit_->offset_ = offset;
+      unit_->deferred_ = deferred;
+      backend_->initialize(u);
+    }
+
+    template<typename ...Args>
+    selection_t select(Args&&...) {
+      size_t i=unit_->offset_;
```
AnuyaWelling2801, line 86: The tests will have to change with the introduction of offset, or we would need some more tests

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
```diff
@@ -0,0 +1,123 @@
...
+  public:
+    round_robin_policy(int offset=0, bool deferred=false) : backend_{std::make_shared<Backend>()}, unit_{std::make_shared<unit_t>()}  {
+      unit_->resource_ = get_resources();
+      unit_->num_contexts_ = unit_->resource_.size();
+      unit_->offset_ = offset;
+      unit_->next_context_ = unit_->offset_;
+      unit_->deferred_ = deferred;
+    }
+
+    round_robin_policy(resource_container_t u, int offset=0, bool deferred=false) : backend_{std::make_shared<Backend>()}, unit_{std::make_shared<unit_t>()}  {
```
AnuyaWelling2801, line 52: If deferred=true,the backend should not be initialized

AnuyaWelling2801: maybe check if (deferred==true) backend_->initialize();

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
```diff
@@ -0,0 +1,123 @@
...
+      unit_->resource_ = backend_->get_resources();
+      unit_->num_contexts_ = unit_->resource_.size();
+      unit_->next_context_ = 0;
+    }
+
+    auto get_resources() const noexcept {
+      return backend_->get_resources();
+    }
+
+    void initialize(int offset=0, bool deferred=false) {
```
AnuyaWelling2801, line 72: Does initialize need deferred?

AnuyaWelling2801 (changes requested):

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
egfefey: Yes, we'll need more tests for offset and deferred. However, since they have defaults, the current tests should be fine.

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
egfefey: The thinking was to have the backend decide on what to do and not the policy making the decision.

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
```diff
@@ -0,0 +1,123 @@
...
+        resource_container_t resource_;
+        resource_container_size_t num_contexts_;
+        std::atomic<resource_container_size_t> next_context_;
+	int offset_;
+	bool deferred_;
+    };
+
+    std::shared_ptr<unit_t> unit_;
+  public:
+    round_robin_policy(int offset=0, bool deferred=false) : backend_{std::make_shared<Backend>()}, unit_{std::make_shared<unit_t>()}  {
```
vossmjp, line 44: Should be a pre-defined deferred value for the offset, not an additional argument. And if deferred nothing should be initialized in policy. Since backends do require an initialize function, it should be ok to assume they can also be constructed as deferred. So it would be ok to create backend in deferred state.

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
```diff
@@ -0,0 +1,123 @@
...
+  public:
+    using resource_t = typename Backend::resource_t; 
+    using selection_t = oneapi::dpl::experimental::basic_selection_t<resource_t>; 
+    using wait_t = typename Backend::wait_t; 
+    using submission_group_t = Backend::submission_group_t; 
+
+  private:
+    std::shared_ptr<Backend> backend_;
+
+    struct unit_t{
```
vossmjp, line 34: Let's use state_t  and state_ instead of unit_t and unit_.

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
```diff
@@ -0,0 +1,123 @@
...
+        resource_container_t resource_;
+        resource_container_size_t num_contexts_;
+        std::atomic<resource_container_size_t> next_context_;
+	int offset_;
+	bool deferred_;
```
vossmjp, line 39: probably can be removed if offset_ == deferred means the same thing.

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
vossmjp, line 72: Only a single argument that is offset.

vossmjp: I don't think it makes sense to call initialize with deferred. It is a choice to silently ignored initialization if not deferred, but definitely should not reinitialize.  So need to check if already initialized.

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
```diff
@@ -0,0 +1,123 @@
...
+      unit_->next_context_ = 0;
+    }
+
+    auto get_resources() const noexcept {
+      return backend_->get_resources();
+    }
+
+    void initialize(int offset=0, bool deferred=false) {
+      unit_->offset_ = offset;
+      unit_->deferred_ = deferred;
```
vossmjp, line 74: If you want to use this to track state of initialization, maybe just call it initialized instead. So, if deferred, it starts out false.

vossmjp (commented):
Needs some changes w.r.t. to deferred initialization.

PR #1066, opened 2023-08-17
[onedpl] Merge Path algorithm for oneDPL merge algorithm (SYCL backend)

MikeDvorskiy:
[onedpl] Merge Path algorithm for oneDPL merge algorithm (SYCL backend).

This is a new implementation of the oneDPL merge algorithm  (merging two sorted ranges) based on new approach known as Merge Path parallel merging. It gives a significant speed up comparing to the current oneDPL implementation.
The main idea  - each work-item is searching for the start points in range 1 and range 2. After each work-item does M steps of sequential merging, where M = (N1+N2) / p, where  p -  number of threads. So, we have got well balanced parallel merging.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1466,8 +1466,8 @@ struct __partial_merge_kernel
     }
 };
 
-//Searching for an intersection of virtual matrix (n1, n2) diagonal with the Merge Path to define sub-ranges
-//to serial merge. For example, a virtual matrix for [0,1,1,2,3] and [0,0,2,3] shown below:
+//Searching for an intersection of a merge matrix (n1, n2) diagonal with the Merge Path to define sub-ranges
+//to serial merge. For example, a merge matrix for [0,1,1,2,3] and [0,0,2,3] is shown below:
 //     0   1  1  2   3
 //    ------------------
 //   |--->
@@ -1478,81 +1478,88 @@ struct __partial_merge_kernel
 // 2 | 0   0  0  0 | 1
 //   |             ---->
 // 3 | 0   0  0  0   0 |
-template<typename Rng1, typename Rng2, typename _Index, typename _Index1, typename _Index2>
-auto __find_start_point(Rng1& __rng1, Rng2& __rng2, _Index __i_elem, _Index1 __n_1, _Index2 __n_2)
+template <typename _Rng1, typename _Rng2, typename _Index, typename _Index1, typename _Index2, typename _Compare>
+auto
+__find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, _Index __i_elem, _Index1 __n1, _Index2 __n2,
+                   _Compare __comp)
 {
-    _Index1 __end1 = 0;
-    _Index2 __end2 = 0;
-    if (__i_elem < __n_1) //a condition to specify upper or down part of the merge matrix to be processed
+    _Index1 __start1 = 0;
+    _Index2 __start2 = 0;
+    if (__i_elem < __n1) //a condition to specify upper or lower part of the merge matrix to be processed
     {
-        auto __q = __i_elem; //digonal index
-        auto __n_diag = __q < __n_2 ? __q : __n_2; //digonal size
+        auto __q = __i_elem; //diagonal index
+        auto __n_diag = ::std::min<_Index2>(__q, __n2); //diagonal size
 
+        //searching for the first '1', a lower bound for a diagonal [0, 0,..., 0, 1, 1,.... 1, 1]
         oneapi::dpl::counting_iterator<_Index> __diag_it(0);
-        auto __res = std::lower_bound(__diag_it, __diag_it + __n_diag, 1, [&__rng1, &__rng2, __q](auto __v1, auto __v2)
+        auto __res = ::std::lower_bound(__diag_it, __diag_it + __n_diag, 1/*value to find*/,
+            [&__rng1, &__rng2, __q, __comp](const auto& __i_diag, const auto& __value)
             {
-                auto __val1 = __rng1[__q - __v1 - 1] < __rng2[__v1];
-                return __val1 < __v2;
+                auto __zero_or_one = __comp(__rng1[__q - __i_diag - 1], __rng2[__i_diag]);
+                return __zero_or_one < __value;
             });
-        __end1 = __q - *__res;
-        __end2 = *__res;
+        __start1 = __q - *__res;
+        __start2 = *__res;
     }
     else
     {
-        auto __q = __i_elem - __n_1; //diagonal index
+        auto __q = __i_elem - __n1; //diagonal index
 
-        auto __n_diag = __n_2 - __q;
-        __n_diag = __n_diag < __n_1 ? __n_diag : __n_1; //digonal size
+        auto __n_diag = ::std::min<_Index1>(__n2 - __q, __n1); //diagonal size
 
-        counting_iterator<_Index> __diag_it(0);
-        auto __res = std::lower_bound(__diag_it, __diag_it + __n_diag, 1, [&__rng1, &__rng2, __n_1, __q](auto __v1, auto __v2)
+        //searching for the first '1', a lower bound for a diagonal [0, 0,..., 0, 1, 1,.... 1, 1]
+        oneapi::dpl::counting_iterator<_Index> __diag_it(0);
+        auto __res = ::std::lower_bound(__diag_it, __diag_it + __n_diag, 1/*value to find*/,
+            [&__rng1, &__rng2, __n1, __q, __comp](const auto& __i_diag, const auto& __value)
             {
-                auto __val1 = __rng1[__n_1 - __v1 - 1] < __rng2[__q + __v1];
-                return __val1 < __v2;
+                auto __zero_or_one = __comp(__rng1[__n1 - __i_diag - 1], __rng2[__q + __i_diag]);
+                return __zero_or_one < __value;
             });
 
-        __end1 = __n_1 - *__res;
-        __end2 = __q + *__res;
+        __start1 = __n1 - *__res;
+        __start2 = __q + *__res;
     }
-    return std::make_pair(__end1, __end2);
+    return std::make_pair(__start1, __start2);
 }
 
-template<typename _Rng1, typename _Rng2, typename _Rng3, typename _Index1, typename _Index2, typename _Index3,
-         typename _Size, typename _Size1, typename _Size2, typename _Compare>
+// Do serial merge of the data from rng1 (starting from start1) and rng2 (starting from start2) and writing
+// to rng3 (starting from start3) in 'chunk' steps, but do not exceed the total size of the sequences (n1 and n2)
+template <typename _Rng1, typename _Rng2, typename _Rng3, typename _Index1, typename _Index2, typename _Index3,
+          typename _Size1, typename _Size2, typename _Compare>
 void
 __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index1 __start1, _Index2 __start2,
-               _Index3 __start3, _Size __chunk, _Size1 __n1, _Size2 __n2, _Compare __comp)
+               _Index3 __start3, ::std::uint8_t __chunk, _Size1 __n1, _Size2 __n2, _Compare __comp)
 {
     if (__start1 >= __n1)
     {
-        //copying a residual of the the second seq
-        const auto __n = std::min<_Index2>(__n2 - __start2, __chunk);
+        //copying a residual of the second seq
+        const auto __n = ::std::min<_Index2>(__n2 - __start2, __chunk);
         _ONEDPL_PRAGMA_UNROLL
-        for (auto __i = 0; __i < __n; ++__i)
+        for (::std::uint8_t __i = 0; __i < __n; ++__i)
             __rng3[__start3 + __i] = __rng2[__start2 + __i];
     }
     else if (__start2 >= __n2)
     {
-        //copying a residual of the the first seq
-        const auto __n = std::min<_Index1>(__n1 - __start1, __chunk);
+        //copying a residual of the first seq
+        const auto __n = ::std::min<_Index1>(__n1 - __start1, __chunk);
         _ONEDPL_PRAGMA_UNROLL
-        for (auto __i = 0; __i < __n; ++__i)
+        for (::std::uint8_t __i = 0; __i < __n; ++__i)
             __rng3[__start3 + __i] = __rng1[__start1 + __i];
     }
     else
     {
-        const auto __n = std::min<_Index1>(__n1 + __n2 - __start3, __chunk);
+        const auto __n = ::std::min<_Index1>(__n1 + __n2 - __start3, __chunk);
         _ONEDPL_PRAGMA_UNROLL
-        for (auto __i = 0; __i < __n; ++__i)
+        for (::std::uint8_t __i = 0; __i < __n; ++__i)
         {
-            const auto __val1 = __rng1[__start1];
-            const auto __val2 = __rng2[__start2];
+            const auto& __val1 = __rng1[__start1];
+            const auto& __val2 = __rng2[__start2];
             if (__comp(__val2, __val1))
             {
                 __rng3[__start3 + __i] = __val2;
                   if(++__start2 == __n2)
                   {
-                      //copying a residual of the the first seq
+                      //copying a residual of the first seq
                       _ONEDPL_PRAGMA_UNROLL
                       for (++__i; __i < __n; ++__i)
                       {
@@ -1566,7 +1573,7 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index1 
                 __rng3[__start3 + __i] = __val1;
                 if (++__start1 == __n1)
                 {
-                    //copying a residual of the the second seq
+                    //copying a residual of the second seq
                     _ONEDPL_PRAGMA_UNROLL
                     for (++__i; __i < __n; ++__i)
                     {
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1466,6 +1466,119 @@ struct __partial_merge_kernel
...
+// 3 | 0   0  0  0   0 |
+template<typename Rng1, typename Rng2, typename _Index, typename _Index1, typename _Index2>
+auto __find_start_point(Rng1& __rng1, Rng2& __rng2, _Index __i_elem, _Index1 __n_1, _Index2 __n_2)
+{
+    _Index1 __end1 = 0;
+    _Index2 __end2 = 0;
+    if (__i_elem < __n_1) //a condition to specify upper or down part of the merge matrix to be processed
+    {
+        auto __q = __i_elem; //digonal index
+        auto __n_diag = __q < __n_2 ? __q : __n_2; //digonal size
```
danhoeflinger, line 1489: spelling

```suggestion
        auto __q = __i_elem; //diagonal index
        auto __n_diag = __q < __n_2 ? __q : __n_2; //diagonal size
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1466,6 +1466,119 @@ struct __partial_merge_kernel
...
+
+        oneapi::dpl::counting_iterator<_Index> __diag_it(0);
+        auto __res = std::lower_bound(__diag_it, __diag_it + __n_diag, 1, [&__rng1, &__rng2, __q](auto __v1, auto __v2)
+            {
+                auto __val1 = __rng1[__q - __v1 - 1] < __rng2[__v1];
```
danhoeflinger, line 1494: `__val1` is a bool as it is the result of the `<` comparison, then we compare it to the diagonal index `__v2`?

Is there a better name for `__val1` to indicate what it is supposed to represent?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1466,6 +1466,119 @@ struct __partial_merge_kernel
...
+            });
+        __end1 = __q - *__res;
+        __end2 = *__res;
+    }
+    else
+    {
+        auto __q = __i_elem - __n_1; //diagonal index
+
+        auto __n_diag = __n_2 - __q;
+        __n_diag = __n_diag < __n_1 ? __n_diag : __n_1; //digonal size
```
danhoeflinger, line 1505: Can this be `auto __n_diag = std::min(__n_2 - q, __n_1)` for better clarity?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger, line 1489: Can this be `auto __n_diag = std::min(__q, __n_2)`  for better clarity?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1466,6 +1466,119 @@ struct __partial_merge_kernel
...
+void
+__serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index1 __start1, _Index2 __start2,
+               _Index3 __start3, _Size __chunk, _Size1 __n1, _Size2 __n2, _Compare __comp)
+{
+    if (__start1 >= __n1)
+    {
+        //copying a residual of the the second seq
+        const auto __n = std::min<_Index2>(__n2 - __start2, __chunk);
+        _ONEDPL_PRAGMA_UNROLL
+        for (auto __i = 0; __i < __n; ++__i)
```
danhoeflinger, line 1531: `__i` needs to be explicitly typed `__Index2` to accommodate `__n`.    This appears in a few more places below.


`_Size` should be included in the numbered `_SizeX`, types.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11, line 1494: Should this comparison use the provided binary comparator?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1466,6 +1466,119 @@ struct __partial_merge_kernel
...
+//   |--->
+// 0 | 0 | 1  1  1   1
+//   |   |
+// 0 | 0 | 1  1  1   1
+//   |   ---------->
+// 2 | 0   0  0  0 | 1
+//   |             ---->
+// 3 | 0   0  0  0   0 |
+template<typename Rng1, typename Rng2, typename _Index, typename _Index1, typename _Index2>
+auto __find_start_point(Rng1& __rng1, Rng2& __rng2, _Index __i_elem, _Index1 __n_1, _Index2 __n_2)
```
mmichel11, line 1482: Can `Rng1` and `Rng2` be const references here? We also may want to rename to `_Rng1` and `_Rng2` for consistency.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1477,22 +1590,26 @@ struct __parallel_merge_submitter<__internal::__optional_kernel_name<_Name...>>
...
+        assert(__n_1 > 0 || __n_2 > 0);
 
         _PRINT_INFO_IN_DEBUG_MODE(__exec);
 
-        const ::std::size_t __chunk = __exec.queue().get_device().is_cpu() ? 128 : 8;
-        const auto __max_n = ::std::max(__n, static_cast<decltype(__n)>(__n_2));
-        const ::std::size_t __steps = ((__max_n - 1) / __chunk) + 1;
+        const unsigned int __chunk = __exec.queue().get_device().is_cpu() ? 128 : 4;
+        const _Size __n = __n_1 + __n_2;
+        const _Size __steps = ((__n - 1) / __chunk) + 1;
```
mmichel11, line 1604: We can probably use `__dpl_ceiling_div` in this line.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1477,22 +1590,26 @@ struct __parallel_merge_submitter<__internal::__optional_kernel_name<_Name...>>
...
 
-        assert(__n > 0 || __n_2 > 0);
+        assert(__n_1 > 0 || __n_2 > 0);
 
         _PRINT_INFO_IN_DEBUG_MODE(__exec);
 
-        const ::std::size_t __chunk = __exec.queue().get_device().is_cpu() ? 128 : 8;
-        const auto __max_n = ::std::max(__n, static_cast<decltype(__n)>(__n_2));
-        const ::std::size_t __steps = ((__max_n - 1) / __chunk) + 1;
+        const unsigned int __chunk = __exec.queue().get_device().is_cpu() ? 128 : 4;
```
mmichel11, line 1602: Should we use a smaller data type?

danhoeflinger (commented):
Still understanding the algorithm, but adding these now to prevent redundant comments, as others are reviewing.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1477,22 +1590,26 @@ struct __parallel_merge_submitter<__internal::__optional_kernel_name<_Name...>>
...
+        const _Size __steps = ((__n - 1) / __chunk) + 1;
 
         auto __event = __exec.queue().submit([&](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rng1, __rng2, __rng3);
             __cgh.parallel_for<_Name...>(sycl::range</*dim=*/1>(__steps), [=](sycl::item</*dim=*/1> __item_id) {
-                __full_merge_kernel()(__item_id.get_linear_id() * __chunk, __rng1, decltype(__n)(0), __n, __rng2,
-                                      decltype(__n_2)(0), __n_2, __rng3, decltype(__n)(0), __comp, __chunk);
+
+                const unsigned int __i_elem = __item_id.get_linear_id() * __chunk;
+                const auto __start = __find_start_point(__rng1, __rng2, __i_elem, __n_1, __n_2);
```
mmichel11, line 1611: Could there be any benefit to loading a work group's data into SLM with a coalesced access pattern before passing it to this function? Since work items can't coalesce their accesses in `__find_start_point`, putting it into SLM first may be beneficial.

danhoeflinger: If I'm understanding the algorithm correctly, this only works if ALL input data fits into SLM.  

Also, we don't have any guarantees on the data here (and how they will be merged), its possible that you may be loading all the input data to end up using only a part of it within the work group.  It would take some investigation to understand if this is of benefit for a "normal" use case, and you would need to come up with a strategy to handle the case where only part of the input sequence fits into SLM, and/or a backup where you don't use SLM.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1466,6 +1466,119 @@ struct __partial_merge_kernel
...
+            __rng3[__start3 + __i] = __rng1[__start1 + __i];
+    }
+    else
+    {
+        const auto __n = std::min<_Index1>(__n1 + __n2 - __start3, __chunk);
+        _ONEDPL_PRAGMA_UNROLL
+        for (auto __i = 0; __i < __n; ++__i)
+        {
+            const auto __val1 = __rng1[__start1];
+            const auto __val2 = __rng2[__start2];
```
danhoeflinger, line 1549: Seems like these could be `const auto&`  to save some copies.

danhoeflinger (commented):
Some more minor things to fix.

At a higher level, I'm still not quite understanding the `__find_start_point` algorithm yet, but I'm getting there.  I'm confused by the `<` comparison between the `bool` (`__val1`) and the diagonal index (`__v2`).  I'll keep looking at this tomorrow.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: No, should not. Here we compare juts zeros and ones.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: Probably, is to try to use SLM makes sense... But in that case we have got the additional n1+n2 reading and n1+n2 additional writing. Together with fact that is almost no computation - just fast binary search + m steps of merging, it is no obvious that we have got a performance gain....
BTW, I tried pre-caching  'chunk' elements form the first seq and 'chunk' elements form the second seq to private memory (registers), and got the performance slow down to 10%.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1466,6 +1466,119 @@ struct __partial_merge_kernel
...
+            });
+
+        __end1 = __n_1 - *__res;
+        __end2 = __q + *__res;
+    }
+    return std::make_pair(__end1, __end2);
+}
+
+template<typename _Rng1, typename _Rng2, typename _Rng3, typename _Index1, typename _Index2, typename _Index3,
+         typename _Size, typename _Size1, typename _Size2, typename _Compare>
```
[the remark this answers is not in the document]
MikeDvorskiy: what do u mean?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: '__v2' is not diagonal index - this is a value to be found. We are searching for '1'.

dmitriy-sobolev, line 1494: As far as I see, it is possible to get integer an integer underflow when `__q` and `v1` are `0`.

dmitriy-sobolev: I agree with Matthew. Moreover the element type would have to support `<` operator. There is no such a requirement for C++ merge. Please note, that `merge.pass.cpp` do not use predicates other than `std::less`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
danhoeflinger:
```suggestion
template<typename _Rng1, typename _Rng2, typename _Rng3, typename _Index1, typename _Index2, typename _Index3,
         typename _Size1, typename _Size2, typename _Size3, typename _Compare>
```

For better consistency of naming.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I believe they are talking about 

`auto __val1 = __rng1[__q - __v1 - 1] < __rng2[__v1];`
->
`auto __val1 = comp(__rng1[__q - __v1 - 1], __rng2[__v1]);`

Not the comparison on the next line.  Does this not compare values of the input ranges?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
danhoeflinger: Yes, I can see your point, It would be bad to change the indices of the related template arguments to not match, or mislead that the chunk size as corresponding to the `_Rng3`.  

It may be more clear that it is intentional if you reorder the parameters to have `_Size __chunk` second to last.  However, I'm fine with however you decide to do this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: I see.... here- yes, it compares the values of of the input ranges, right, thanks.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1466,6 +1466,119 @@ struct __partial_merge_kernel
...
+void
+__serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index1 __start1, _Index2 __start2,
+               _Index3 __start3, _Size __chunk, _Size1 __n1, _Size2 __n2, _Compare __comp)
+{
+    if (__start1 >= __n1)
+    {
+        //copying a residual of the the second seq
+        const auto __n = std::min<_Index2>(__n2 - __start2, __chunk);
+        _ONEDPL_PRAGMA_UNROLL
```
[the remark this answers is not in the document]
MikeDvorskiy: I suggest trying it after the upcoming release. I can make a "reference in new issue"...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: thanks, that helps.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
MikeDvorskiy: "_Size __chunk second to last." also breakers a call semantic... 
Indeed,
```
__serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index1 __start1, _Index2 __start2,
               _Index3 __start3, _Size __chunk, _Size1 __n1, _Size2 __n2, _Compare __comp)
```
Tells:
"Does serial merge the data from rng1(starting from start1) and rng2(starting from start2) and writting to rng3 (starting from start3) in '__chunk' steps , but doesn't exceed of total size of  sequences  - n1 and n2."

May be it is better if '__chunk' will be at the first palce:
"Does  '__chunk' steps of serial merging of  the data from rng1 and rng2....."

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Actually, I'm still not convinced about `__v2`.  Is it not just the next element of the `counting_iterator`?

It seems like you are wanting the input iterator sequence to be already transformed into  `__rng1[__q - idx - 1] < __rng2[idx]` where `idx` is  is the `counting_iterator` element, but that isn't the case as written.  

I think you would need to zip together the input ranges, have a transform view and a functor which applies the above expression, then pass in that transform iterator into the `lower_bound` algorithm to get the result you are looking for here.

Unless I'm missing something, `__v2` is just the next element of the input counting iterator after `__v1` and it has not been similarly transformed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: You've changed the type to `::std::uint8_t`, which I believe will work if `__serial_merge` is called in the context of this algorithm.  However, if this is the intention, then we should probably just remove the `_Size` template argument, and just explicitly type `__chunk` as `::std::uint8_t`.  This solves the other discussion as well.

If instead we want `__serial_merge` to be general and available for use in other algorithms, then `__i` needs to be one of `_Size`, `_Index2` or `_Size2` to make sure there is no overflow.  Ideally it could be typed as the smallest of those types, and we could write some compile time helpers to accomplish that.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: Cppreference tells about std::lower_bound:
```
Returns an iterator pointing to the first element in the range [first, last) such that element < value (or comp(element, value)) is false, (i.e. that is greater than or equal to value), or last if no such element is found.

The range [first, last) must be partitioned with respect to the expression element < value (or comp(element, value)), i.e., all elements for which the expression is true must precede all elements for which the expression is false. A fully-sorted range meets this criterion.
```

danhoeflinger: Sorry, I've been thinking about it wrong.  My mistake.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I think to have any performance benefit with SLM, we will have to look at the GPU Merge Path variant of Merge Path which uses a tiling strategy. This can probably be a future optimization if it is difficult to implement since this implementation already provides significant performance benefit.

dmitriy-sobolev (approved):
LGTM. Please, double check the CI results before merging. You also might want to get an approval from the other reviewers.

dmitriy-sobolev (approved):
Reapproving after some minor changes (e.g. `__n_1` is renamed to `__n1` for consistency).

danhoeflinger (dismissed):
LGTM

dmitriy-sobolev (approved):
Reapproving after some formatting changes.

PR #1067, opened 2023-08-17
Removing free functions for set_universe and get_universe size

AnuyaWelling2801:
Removing free functions for set_universe and get_universe size
The set_universe function in sycl_scheduler, round_robin_policy and static policy still exists and needs to be replaced in future PRs


vossmjp (approved):
Looks fine.

PR #1073, opened 2023-08-18
[KT] Remove non-used operations from onesweep kernel

dmitriy-sobolev:
`ResetBinCounters` is removed. `RankSLM` does the same, but without using fence which is unnecessary as well.
`bins` are not changed, so the second calculation is not necessary.


SergeyKopienko (approved):
LGTM: please check all tests

dmitriy-sobolev:
> LGTM: please check all tests

All the test pass with the most recent GPU driver.

dmitriy-sobolev:
@SergeyKopienko, I have rebased the changes (with applying them both for key and key-value sort). All the tests pass.

PR #1079, opened 2023-08-21
RR updates

egfefey:
Added offset and initialization

### DIFF
--- a/include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy_impl.h
+++ b/include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy_impl.h
@@ -31,70 +31,73 @@ namespace experimental{
     using resource_type = typename backend_t::resource_type;
     using wait_type = typename backend_t::wait_type;
 
-
     std::shared_ptr<backend_t> backend_;
 
-
-    struct unit_t{
+    struct state_t{
         resource_container_t resources_;
         resource_container_size_t num_contexts_;
         std::atomic<resource_container_size_t> next_context_;
+	int offset_;
     };
 
-    std::shared_ptr<unit_t> unit_;
+    std::shared_ptr<state_t> state_;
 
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy_impl.h
```diff
@@ -31,70 +31,73 @@ namespace experimental{
...
+      state_->offset_ = offset;
+      state_->next_context_ = state_->offset_;
     }
 
     template<typename ...Args>
-    round_robin_policy_impl(Args&&... args) : backend_{std::make_shared<backend_t>(std::forward<Args>(args)...)}, unit_{std::make_shared<unit_t>()} {
-      unit_->resources_ = backend_->get_resources();
-      unit_->num_contexts_ = unit_->resources_.size();
-      unit_->next_context_ = 0;
+    round_robin_policy_impl(Args&&... args) : backend_{std::make_shared<backend_t>(std::forward<Args>(args)...)}, state_{std::make_shared<state_t>()} {
```
egfefey, line 61: A lot of our tests have a resource function in addition to the resource container. This overload supports that. Can anyone confirm or not that we're doing away with this feature?

It's of the form:
template<typename Policy, typename UniverseContainer, typename ResourceFunction>
int test_submit_and_wait_on_sync(UniverseContainer u, ResourceFunction&& f)

vossmjp: The ResourceFunction is used by the tests to return the expected resource. For example, the round_robin function returns u[(i-1)%4] to mimic round-robin-ing through the universe, while the static policy returns u[0] all the time, since that's what it selects. So the universe and function serve two different purposes in the tests. The ResourceFunction is very test specific, it has assumptions built-in so that it works for the specific test scenario. It's only a feature of the tests, not the policies themselves.

vossmjp (approved):

PR #1081, opened 2023-08-22
Fixed resource policy updates

egfefey:
Added offset, initialize and changed name to fixed resource policy.
Will change the file names in a different PR. Didn't want to bite too much.

### DIFF
--- a/include/oneapi/dpl/internal/dynamic_selection_impl/static_policy_impl.h
+++ b/include/oneapi/dpl/internal/dynamic_selection_impl/static_policy_impl.h
@@ -17,55 +17,59 @@ namespace dpl {
 namespace experimental {
 
   template <typename Backend>
-  struct static_policy_impl {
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/internal/dynamic_selection_impl/static_policy_impl.h
```diff
@@ -17,55 +17,59 @@ namespace dpl {
...
+      backend_->initialize();
+    }
+
+    void initialize(resource_container_t u, int offset=0) {
+      state_->offset_ = offset;
+      backend_->initialize(u);
     }
 
     template<typename ...Args>
     selection_type select(Args&&...) {
```
vossmjp, line 70: Need to account for deferred inititalization here too.

vossmjp (approved):
The changes are ok, but other changes need to made. Perhaps as an future PR.

PR #1083, opened 2023-08-22
[Docs] Adding req of const function objects for device policy algorithms 

danhoeflinger:
Adding a difference from the Standard C++ Parallel Algorithms.
Algorithms executed with a device policy must provide `const` function call operators for any provided functions objects.  

I'm not sure if it belongs here or as a "Known Limitations", as it is due to conflicting specifications between Standard Library and SYCL, but I prefer this placement.  There is no matching explicit requirement on function objects in the Standard C++ Parallel Algorithms.  

I don't expect SYCL to ever release this requirement, as using non-`const` functions of function objects in this context has some fundamental problems which run counter to performance and/or correctness.

---
Details of why this is required:
The SYCL spec requires kernels to use only `const` function call operators of function objects:

> 4.12.1 Defining kernels as named function objects
> “The operator() member function must be const-qualified” … “If the operator() function writes to any of the member variables, the behavior is undefined.”
> 4.12.2 Defining kernels as lambda functions
> “The kernel lambda must use copy for all of its captures (i.e. [=]), and the lambda must not use the mutable specifier.”

For a function object to be used in a SYCL kernel, it must either (1) be a member variable of named kernel function struct, or (2) it must be passed in to a lambda via a copy `=` capture.

1) If the kernel function struct will be using a `const` function call operator (4.12.1), which cannot modify any member functions of the kernel function struct. This means that it cannot call any non-`const` operators of member variables like the function object provided.
2) If a function object is passed by copy capture into the kernel lambda, it is marked `const` by default during the capture, and therefore non-`const` operators may not be called. If the lambda is marked as `mutable`, this removes the const qualification from the copy capture values, but this is explicitly disallowed for SYCL kernels (4.12.2).

### DIFF
--- a/documentation/library_guide/introduction.rst
+++ b/documentation/library_guide/introduction.rst
@@ -63,8 +63,6 @@ To use Parallel API with the device execution policies, you need to install the 
 Difference with Standard C++ Parallel Algorithms
 ************************************************
 
-* For algorithms executed with device policies, function object parameters which are passed must
-provide `const` function call operators.
 * oneDPL execution policies only result in parallel execution if random access iterators are provided,
   the execution will remain serial for other iterator types.
 * For the following algorithms, par_unseq and unseq policies do not result in vectorized execution:
@@ -93,6 +91,8 @@ When called with |dpcpp_short| execution policies, |onedpl_short| algorithms app
 Known Limitations
 *****************
 
+* Due to requirements of the SYCL specification, function object parameters passed in to algorithms executed with device
+  policies must provide `const` function call operators.
 * For ``transform_exclusive_scan``, ``transform_inclusive_scan`` algorithms the result of the unary operation should be
   convertible to the type of the initial value if one is provided, otherwise it is convertible to the type of values
   in the processed data sequence: ``std::iterator_traits<IteratorType>::value_type``.
@@ -126,7 +126,6 @@ Known Limitations
   the dereferenced value type of the provided iterators should satisfy the ``DefaultConstructible`` requirements.
 * For ``remove``, ``remove_if``, ``unique`` the dereferenced value type of the provided
   iterators should be ``MoveConstructible``.
-        
 
 Build Your Code with |onedpl_short|
 ===================================
--- a/documentation/library_guide/introduction.rst
+++ b/documentation/library_guide/introduction.rst
@@ -71,6 +71,8 @@ Difference with Standard C++ Parallel Algorithms
 * The following algorithms require additional O(n) memory space for parallel execution:
   ``copy_if``, ``inplace_merge``, ``partial_sort``, ``partial_sort_copy``, ``partition_copy``,
   ``remove``, ``remove_if``, ``rotate``, ``sort``, ``stable_sort``, ``unique``, ``unique_copy``.
+* For algorithms executed with device policies, callable function object parameters must provide `const`
+  function call operators.
 
 
 Restrictions
--- a/documentation/library_guide/introduction.rst
+++ b/documentation/library_guide/introduction.rst
@@ -63,6 +63,8 @@ To use Parallel API with the device execution policies, you need to install the 
 Difference with Standard C++ Parallel Algorithms
 ************************************************
 
+* Function object parameters passed in to algorithms executed with device policies must provide `const` function call
+  operators, due to requirements of the SYCL specification.
 * oneDPL execution policies only result in parallel execution if random access iterators are provided,
   the execution will remain serial for other iterator types.
 * For the following algorithms, par_unseq and unseq policies do not result in vectorized execution:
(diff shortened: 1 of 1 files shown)


documentation/library_guide/introduction.rst
```diff
@@ -63,6 +63,8 @@ To use Parallel API with the device execution policies, you need to install the
 Difference with Standard C++ Parallel Algorithms
 ************************************************
 
+* Function object parameters passed in to algorithms executed with device policies must provide `const` function call
```
rarutyun, line 66: I think adding something like that is useful. But probably it should be stated differently.

What we have here is not sufficient because SYCL specification also says that writing to "member variables" (I really hope that they meant "non-static data members") in the SYCL kernel is undefined behavior. In simple words it means that even if `operator()` is `const`-qualified user is not allowed to make members of the class as `mutable` and attempt to right to those.

That makes me think that we need to either reference to SYCL specification, or make a slightly different statement in our doc, or do some combination of the two mentioned approaches.

rarutyun, line 66: Even if we add that I think it shouldn't go first because it's related to the algorithms being launched with device policies. The following statement about iterators is applicable to everything making it more general.

danhoeflinger: This is a good point.  We could make it a bit stronger with something like:

"Function object parameters passed in to algorithms executed with device policies must provide `const` function call operators. The SYCL spec (4.12) states that writing to such an object during a SYCL kernel is undefined behavior." 

This should make it more clear that even marking as mutable within a wrapper is not supported, if you are in fact writing to the mutable object.

dcbenito: I like the updated instance. I think that it is very clear and I have no updates to the text.

dcbenito: I agree that it would work well as the second bullet.

danhoeflinger: Updated the language to what I proposed.

danhoeflinger: Moved to second bullet.

dcbenito (dismissed):
LGTM

documentation/library_guide/introduction.rst
```diff
@@ -65,6 +65,8 @@ Difference with Standard C++ Parallel Algorithms
 
 * oneDPL execution policies only result in parallel execution if random access iterators are provided,
   the execution will remain serial for other iterator types.
+* Function object parameters passed in to algorithms executed with device policies must provide const function call
```
rarutyun, line 68:
```suggestion
* Function objects passed in to algorithms executed with device policies much provide ``const``-qualified ``operator()``.
```

documentation/library_guide/introduction.rst
```diff
@@ -65,6 +65,8 @@ Difference with Standard C++ Parallel Algorithms
...
+  operators. The SYCL spec (4.12) states that writing to such an object during a SYCL kernel is undefined behavior.
```
rarutyun, line 69:
```suggestion
The SYCL specification states that writing to such an object during a SYCL kernel is undefined behavior.
```

documentation/library_guide/introduction.rst
dcbenito: "much" should be "must", otherwise it looks good to me.

danhoeflinger: Thanks, done.

documentation/library_guide/introduction.rst
danhoeflinger: Thanks, done.

rarutyun (dismissed):
The only thing that we might want to improve (to me) is making the "SYCL specification" as hyperlink. Other than that, LGTM.

danhoeflinger:
> The only thing that we might want to improve (to me) is making the "SYCL specification" as hyperlink. Other than that, LGTM.

I was looking at this:  
@dcbenito Do you think https://registry.khronos.org/SYCL/ is the correct link to ensure it is stable?  (not linking to a specific version of the standard)

Also, I notice another mention of the SYCL specification in this file without link ("Restrictions"), should we add a link to that as well?

danhoeflinger:
I've added the link, only to the new bullet point at this time, with what should be an evergreen URL.  If we want to switch it to point to SYCL 2020 html specification URL, I believe that would be the other reasonable option.

dcbenito:
I think that the link to the main Khronos page is good for where we want to point our users. I don't feel like we need to add a link to the second instance, since it is available in the text above.

danhoeflinger:
> I think that the link to the main Khronos page is good for where we want to point our users. I don't feel like we need to add a link to the second instance, since it is available in the text above.

@dcbenito 
Great, that should be what is in the PR currently.  If you could give a final check and approval, I'll go ahead and merge.

dcbenito (approved):
LGTM

PR #1089, opened 2023-08-23
Adding sycl guards, name updates

egfefey:
Guards for sycl vrs non-sycl
Fixed resource policy name updates
Round robin name updates


vossmjp (approved):
Looks good to me.

PR #1093, opened 2023-08-23
Using submission_group

AnuyaWelling2801:
Added submission group as the waiter container to the sycl_backend and inline_backend

### DIFF
--- a/include/oneapi/dpl/internal/dynamic_selection_impl/submission_group.h
+++ b/include/oneapi/dpl/internal/dynamic_selection_impl/submission_group.h
@@ -18,31 +18,29 @@ namespace dpl {
 namespace experimental {
   template<typename T>
   class submission_group_t {
-    using submission_container_t = util::concurrent_queue<T>;
+    using submission_container_t = std::shared_ptr<util::concurrent_queue<T>>;
     using submissions_t = std::vector<T>;
     submission_container_t submission_container;
     submissions_t submissions;
 
     public:
-    submission_group_t() = default;
-    submission_group_t(const submission_group_t&) = delete;
-    submission_group_t& operator=(const submission_group_t&) = delete;
+    submission_group_t() : submission_container(std::make_shared<util::concurrent_queue<T>>()) {};
 
     //add to submission group after submit
-    void add_submission(T w) { 
-      submission_container.push_back(w);	
-    }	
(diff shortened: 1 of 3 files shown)


include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -14,7 +14,7 @@
 #include "oneapi/dpl/internal/dynamic_selection.h"
 #include "oneapi/dpl/internal/dynamic_selection_impl/scoring_policy_defs.h"
 #include "oneapi/dpl/internal/dynamic_selection_impl/backend_defs.h"
-#include "oneapi/dpl/internal/dynamic_selection_impl/concurrent_queue.h"
+#include "oneapi/dpl/internal/dynamic_selection_impl/submission_group.h"
```
akukanov, line 17: Should this patch also remove `concurrent_queue.h`?

include/oneapi/dpl/internal/dynamic_selection_impl/submission_group.h
```diff
@@ -18,31 +18,29 @@ namespace dpl {
 namespace experimental {
   template<typename T>
   class submission_group_t {
-    using submission_container_t = util::concurrent_queue<T>;
+    using submission_container_t = std::shared_ptr<util::concurrent_queue<T>>;
     using submissions_t = std::vector<T>;
     submission_container_t submission_container;
     submissions_t submissions;
```
akukanov, line 24: What's the difference of `submission_container` and `submissions`? It seems the latter is not used anywhere.

include/oneapi/dpl/internal/dynamic_selection_impl/submission_group.h
```diff
@@ -18,31 +18,29 @@ namespace dpl {
...
-      while(!submission_container.empty()){
-        T w;  
-        submission_container.pop(w);
+      while(!submission_container->empty()){
+        T w;
+        submission_container->pop(w);
         w->wait();
         delete w;
       }
     }
```
akukanov, line 42: If this function is supposed to be used by multiple threads, it is incorrect. Different threads will pop different elements from the queue and therefore will wait only on a part of submissions.

PR #1094, opened 2023-08-23
Fix warning: 'get_pointer' is deprecated

dmitriy-sobolev:
`get_pointer` methods for buffer and local accessors recently became deprecated according to SYCL 2020 specification revision 8. 

You can find the information about the deprecation in [sycl-2020.pdf](https://github.com/oneapi-src/oneDPL/files/12422386/sycl-2020.pdf) (revision 8 is not published, this file has been generated from the latest sources on https://github.com/KhronosGroup/SYCL-Docs/tree/SYCL-2020/master). See Table 51 (Member functions of the accessor class) and Table 72 (Member functions of the local_accessor class). 

These APIs are already deprecated in the compiler in favor of `get_mult_ptr` (https://github.com/intel/llvm/pull/10174). This results in a warning, e.g:
> parallel_backend_sycl.h:634:32: warning: 'get_pointer' is deprecated: local_accessor::get_pointer() is deprecated, please use get_multi_ptr()

`get_pointer` returns a raw pointer, so it was replaced by `get_multi_ptr<sycl::access::decorated::no>().get()`, which also returns a raw, non-decorated, pointer. 

As a follow up task, we can investigate applicability of `sycl::access::decorated::yes`, but I am concerned about portability of it. Here are some notes from the specification: 
>Multi-pointer class
>If the value of access::decorated is access::decorated::yes, the interface exposes pointers and references type that are decorated by an address space. The decoration is implementation dependent and relies on device compiler extensions.
> ... 
> Member functions of multi_ptr class
> __unspecified__* get_decorated() const | Returns the underlying pointer decorated by the address space that it addresses. Note that the support involves implementation-defined device compiler extensions.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -610,7 +610,7 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
 #else
                     constexpr bool __can_use_subgroup_load_store = false;
 #endif
-                    auto __lacc_ptr = __get_accessor_ptr(__lacc);
+                    auto __lacc_ptr = __dpl_sycl::__get_accessor_ptr(__lacc);
                     if constexpr (__can_use_subgroup_load_store)
                     {
                         _ONEDPL_PRAGMA_UNROLL
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -444,8 +444,8 @@ struct __parallel_transform_scan_dynamic_single_group_submitter<_Inclusive,
                         __lacc[__idx] = __unary_op(__in_rng[__idx]);
                     }
 
-                    __scan_work_group<_ValueType, _Inclusive>(__group, __lacc.get_pointer(), __lacc.get_pointer() + __n,
-                                                              __lacc.get_pointer(), __bin_op, __init);
+                    auto __ptr = __dpl_sycl::__get_accessor_ptr(__lacc);
+                    __scan_work_group<_ValueType, _Inclusive>(__group, __ptr, __ptr + __n, __ptr, __bin_op, __init);
 
                     for (::std::uint16_t __idx = __item_id; __idx < __n; __idx += __wg_size)
                     {
@@ -505,6 +505,7 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
                     constexpr bool __can_use_subgroup_load_store = false;
 #endif
 
+                    auto __lacc_ptr = __dpl_sycl::__get_accessor_ptr(__lacc);
                     if constexpr (__can_use_subgroup_load_store)
                     {
                         _ONEDPL_PRAGMA_UNROLL
@@ -512,7 +513,7 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
                         {
                             auto __idx = __i * _WGSize + __subgroup_id * __subgroup_size;
                             auto __val = __unary_op(__subgroup.load(__in_rng.begin() + __idx));
-                            __subgroup.store(__lacc.get_pointer() + __idx, __val);
+                            __subgroup.store(__lacc_ptr + __idx, __val);
                         }
                     }
                     else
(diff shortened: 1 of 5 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -379,6 +379,16 @@ __get_host_access(_Buf&& __buf)
 #endif
 }
 
+template <typename _Acc>
+auto __get_accessor_ptr(const _Acc& __acc)
```
adamfidel, line 383: I would've said that the `const` here is unnecessary but I looked up in the spec that `get_pointer` is a const method (that internally `const_cast`s away the constness) so this seems fine to me.

adamfidel (dismissed):
LGTM.

I've confirmed that the libsycl version that deprecates `get_pointer` is 7.0.0.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -343,7 +343,7 @@ struct __peer_prefix_helper<_OffsetT, __peer_prefix_algo::atomic_fetch_or>
 
     __peer_prefix_helper(sycl::nd_item<1> __self_item, _TempStorageT __lacc)
         : __sgroup(__self_item.get_sub_group()), __self_lidx(__self_item.get_local_linear_id()),
-          __item_mask(~(~0u << (__self_lidx))), __atomic_peer_mask(*__lacc.get_pointer())
+          __item_mask(~(~0u << (__self_lidx))), __atomic_peer_mask(*__get_accessor_ptr(__lacc))
```
danhoeflinger, line 346: I think this needs `__dpl_sycl::` prior to the call.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -673,7 +673,7 @@ struct __parallel_radix_sort_iteration
 
         // work-group size must be a power of 2 and not less than the number of states.
         // TODO: Check how to get rid of that restriction.
-        __count_wg_size = 
+        __count_wg_size =
```
danhoeflinger, line 676: formatting only change

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -609,15 +610,15 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
 #else
                     constexpr bool __can_use_subgroup_load_store = false;
 #endif
-
+                    auto __lacc_ptr = __get_accessor_ptr(__lacc);
```
danhoeflinger, line 613: I think this also needs `__dpl_sycl::`

danhoeflinger:
I'm seeing usage of `.get_pointer()` in the tests and in some examples as well.  Those don't need to be changed in this PR, but should be addressed at some point.    For the tests, I suppose we could use `__dpl_sycl::` utilities, but we would likely just want to rewrite the examples to use the new API or avoid the need to get a pointer.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
dmitriy-sobolev: Would you mind if I leave that change?

dmitriy-sobolev:
> I'm seeing usage of `.get_pointer()` in the tests and in some examples as well. Those don't need to be changed in this PR, but should be addressed at some point. For the tests, I suppose we could use `__dpl_sycl::` utilities, but we would likely just want to rewrite the examples to use the new API or avoid the need to get a pointer.

My intention was to use a newer API in the tests in a separate PR, so a more important part, the library, is updated faster. I would create a duplicate of `__dpl_sycl::__get_accessor_ptr` in the test utilities in order not to rely on the internal API and e.g. to allow backward compatibility testing (older tests, newer library).

Thanks for the idea about removing `.get_pointer()` from the tests/examples if it is not necessary. I will take a look at it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
danhoeflinger: I don't really mind.  I know others feel strongly to keep the git blame clean, but I am ok with fixing this formatting issue here.

danhoeflinger (approved):
LGTM

PR #1097, opened 2023-08-24
Fix an error in __has_known_identity

SergeyKopienko:
In this PR we add check of ```std::plus<void>``` and ```__dpl_sycl::__plus<void>``` functional objects into definition of ```__has_known_identity```

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
@@ -48,7 +48,9 @@ using __has_known_identity =
     typename ::std::conjunction<
         ::std::is_arithmetic<_Tp>, __dpl_sycl::__has_known_identity<_BinaryOp, _Tp>,
         ::std::disjunction<::std::is_same<typename ::std::decay<_BinaryOp>::type, ::std::plus<_Tp>>,
+                           ::std::is_same<typename ::std::decay<_BinaryOp>::type, ::std::plus<void>>,
                            ::std::is_same<typename ::std::decay<_BinaryOp>::type, __dpl_sycl::__plus<_Tp>>,
+                           ::std::is_same<typename ::std::decay<_BinaryOp>::type, __dpl_sycl::__plus<void>>,
                            ::std::is_same<typename ::std::decay<_BinaryOp>::type, __dpl_sycl::__minimum<_Tp>>,
                            ::std::is_same<typename ::std::decay<_BinaryOp>::type, __dpl_sycl::__maximum<_Tp>>>>;
 #    else  //_ONEDPL_LIBSYCL_VERSION >= 50200


include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -48,7 +48,9 @@ using __has_known_identity =
     typename ::std::conjunction<
         ::std::is_arithmetic<_Tp>, __dpl_sycl::__has_known_identity<_BinaryOp, _Tp>,
         ::std::disjunction<::std::is_same<typename ::std::decay<_BinaryOp>::type, ::std::plus<_Tp>>,
+                           ::std::is_same<typename ::std::decay<_BinaryOp>::type, ::std::plus<void>>,
```
timmiesmith, line 51: Should these additional cases be added to the #else branch below?

SergeyKopienko: Agreed, fixed.

timmiesmith (approved):
LGTM. Thank you.

timmiesmith:
Confirmed clang-format suggestions are for lines not impacted by this PR.

akukanov:
Not caused by this patch, but the implementation seems to have a logical flaw.

```
template <typename _BinaryOp, typename _Tp>
using __has_known_identity =
#    if _ONEDPL_LIBSYCL_VERSION >= 50200
    typename ::std::conjunction<
        ::std::is_arithmetic<_Tp>, __dpl_sycl::__has_known_identity<_BinaryOp, _Tp>,
        ::std::disjunction<::std::is_same<typename ::std::decay<_BinaryOp>::type, ::std::plus<_Tp>>,
                           ::std::is_same<typename ::std::decay<_BinaryOp>::type, __dpl_sycl::__plus<_Tp>>,
                           ::std::is_same<typename ::std::decay<_BinaryOp>::type, __dpl_sycl::__minimum<_Tp>>,
                           ::std::is_same<typename ::std::decay<_BinaryOp>::type, __dpl_sycl::__maximum<_Tp>>>>;
```
This deciphers as "the type is arithmetic AND there is a known identity for the {type,operation} pair AND the operation is one of those known to work for the type". But really it should be "there is a known identity for {type,operation} OR (the type is arithmetic AND the operation is known to work for the type)"

SergeyKopienko:
> 

In the table https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.identities the rules is more complicated:
- for example, ```sycl::plus``` is "Available Only When" ```std::is_arithmetic_v<AccumulatorT> || std::is_same_v<std::remove_cv_t<AccumulatorT>, sycl::half>``` and so on.

SergeyKopienko:
@akukanov What do you think about this: https://github.com/oneapi-src/oneDPL/pull/1104/files#r1305430409
Is that what you mean in your comment?

SergeyKopienko:
Fixed: looks like ```__dpl_sycl::__has_known_identity<_BinaryOp, _Tp>``` return correct result now.

akukanov (approved):

PR #1109, opened 2023-08-25
Revert ternary conditional operator in min_element introduced in #1026

julianmi:
Switching to a ternary conditional operator in the reduce function of `min_element` as introduced in #1026 results in slowdowns on PVC. This reverts the change to the lambda.


danhoeflinger (approved):
LGTM

mmichel11 (approved):
LGTM

PR #1112, opened 2023-08-26
[Dynamic Selection] submission_group now waits on queues

vossmjp:
Changes submission_group in sycl_backend to wait on queues not individual submissions. Also fixes up inline_backed in tests to match requirements and report events in the most appropriate locations.


AnuyaWelling2801 (approved):
One change would be to delete the old files for concurrent queue and submission_group, but can be merged after that

PR #1117, opened 2023-08-28
Check for device support of required subgroup size in single work group radix sort

mmichel11:
In our single work group radix sort implementation, we provide an attribute to the kernel requiring a subgroup size of 16 or 32 dependent on the size of input. However, we do not perform a check to see if the required subgroup size is supported on the target device, so we will encounter a runtime exception on devices that do not support both subgroup sizes of 16 and 32. In practice, this has been encountered when running on CUDA devices as only subgroup sizes of 32 are supported here.

This PR adds a preprocessor check to see if we are compiling to SPIRV and only requires the specified subgroup size in this case.


danhoeflinger:
This is going to be further complicated by https://github.com/oneapi-src/oneDPL/pull/999, which adds more kernel instantiations as well.

mmichel11:
> This is going to be further complicated by #999, which adds more kernel instantiations as well.

Instead of checking for subgroup size support during runtime and selecting the correct kernel, I have added a preprocessor check if we are targeting the NVPTX backend.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -158,7 +158,12 @@ struct __subgroup_radix_sort
                 auto __counter_lacc = __buf_count.get_acc(__cgh);
 
                 __cgh.parallel_for<_Name...>(
-                    __range, ([=](sycl::nd_item<1> __it)[[_ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(__req_sub_group_size)]] {
+                    __range, ([=](sycl::nd_item<1> __it)
+                        // CUDA devices only support sg size of 32, so we disable this requirement
+#if !_ONEDPL_DETECT_NVPTX_COMPILATION
+                        [[_ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(__req_sub_group_size)]]
+#endif
+                    {
```
danhoeflinger, line 166: Would it be better to add a macro which folds this logic in?

Something like this:

`__range, ([=](sycl::nd_item<1> __it) _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE_IF_SUPPORTED(__req_sub_group_size) {`

And then keep the logic to check specific hardware in `sycl_defs.h`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -158,7 +158,12 @@ struct __subgroup_radix_sort
                 auto __counter_lacc = __buf_count.get_acc(__cgh);
 
                 __cgh.parallel_for<_Name...>(
-                    __range, ([=](sycl::nd_item<1> __it)[[_ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(__req_sub_group_size)]] {
+                    __range, ([=](sycl::nd_item<1> __it)
+                        // CUDA devices only support sg size of 32, so we disable this requirement
```
julianmi, line 162: Would we need to guard this for other backends as well? AMD seems to support 64 and 32 depending on their configuration (https://github.com/RadeonOpenCompute/ROCm/discussions/2318).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
mmichel11: Yes, I will add this change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
mmichel11: Yes, if there is a reliable way to determine if we are compiling for the HIP backend, then we should guard against this as well. I will take a look to see if we can do this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
danhoeflinger: It looks like its valid to have an empty attribute `[[ ]]`, so we can keep the syntax close to the baseline SYCL, and it can look like this:
`__range, ([=](sycl::nd_item<1> __it) [[_ONEDPL_SYCL_REQD_SUB_GROUP_SIZE_IF_SUPPORTED]] (__req_sub_group_size) {`

MikeDvorskiy (approved):
LGTM

julianmi (approved):

PR #1119, opened 2023-08-29
aligning headers

egfefey:
Aligned the headers

PR #1122, opened 2023-08-29
[Dynamic Selection] Added dynamic_selection_traits.h

AnuyaWelling2801:
Moved the header files to the out dynamic_selection and renamed the file dynamic_selection.h to dynamic_selection_traits.h

### DIFF
--- a/include/oneapi/dpl/dynamic_selection
+++ b/include/oneapi/dpl/dynamic_selection
@@ -21,9 +21,6 @@
 
 #include "oneapi/dpl/internal/dynamic_selection_traits.h"
 
-#if _DS_BACKEND_SYCL != 0
-    #include "oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h"
-#endif
 #include "oneapi/dpl/internal/dynamic_selection_impl/fixed_resource_policy.h"
 #include "oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h"
 #include "oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h"
--- a/include/oneapi/dpl/dynamic_selection
+++ b/include/oneapi/dpl/dynamic_selection
@@ -19,6 +19,13 @@
     #define _DS_BACKEND_SYCL 0
 #endif
 
-#include "oneapi/dpl/internal/dynamic_selection.h"
+#include "oneapi/dpl/internal/dynamic_selection_traits.h"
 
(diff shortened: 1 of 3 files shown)


include/oneapi/dpl/dynamic_selection
```diff
@@ -19,6 +19,13 @@
     #define _DS_BACKEND_SYCL 0
 #endif
 
-#include "oneapi/dpl/internal/dynamic_selection.h"
+#include "oneapi/dpl/internal/dynamic_selection_traits.h"
 
+#if _DS_BACKEND_SYCL != 0
```
vossmjp, line 24: include of backend should be removed. Each policy will include it as needed. This file should not.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -11,7 +11,7 @@
 #define _ONEDPL_SYCL_BACKEND_IMPL_H
 
 #include <CL/sycl.hpp>
-#include "oneapi/dpl/internal/dynamic_selection.h"
+#include "oneapi/dpl/dynamic_selection"
```
vossmjp, line 14: This is a cyclic include. Needs to be addressed.  This is seems necessary because of the use of the execution_info reporting.  Reporting is part of the contract between the policies and backends, so probably does not need to be directly in the <dynamic_selection> header, but instead included by the policies and backends that opt-in to the use of execution_info contract.

vossmjp (changes requested):
Need to address some includes. Move execution_info into its own header and then include it in policies and backends that participate in reporting / receiving execution_info. Also no need to include backend in main <dynamic_selection> header.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
vossmjp: If this include the new dynamic_selection_traits.h we can break the cycle.

AnuyaWelling2801: Will be fixed in #1119

include/oneapi/dpl/dynamic_selection
AnuyaWelling2801: Fixed

vossmjp (approved):
Looks ok (assuming followed by 1119)

PR #1124, opened 2023-08-30
Removed offset of 2

AnuyaWelling2801:
To make sure the tests pass even if the gpu_queue is absent


vossmjp (approved):
Looks good and I agree with the new underlying assumption.

PR #1126, opened 2023-08-30
Fix generated kernel name to consider ExecutionPolicy

rarutyun:
The generated kernel name should include `ExecutionPolicy` template parameter among others to avoid the conflict.

It turns out, all other (besides fixed) places take `ExecutionPolicy` into account implicitly via Callable type.

Example of "capturing" the policy via Callable type: `using _Functor = unseq_backend::walk_n<_ExecutionPolicy, _BinaryOperation2>;`

When the `ExecutionPolicy` is not taken into account the example of different instantiations of the function the produces the same kernel name is with `device_policy` and with `const device_policy`

The fix is temporary, due to time constraints. Need to think if we can redesign the backend to never accept the policy by forwarding reference

### DIFF
--- a/include/oneapi/dpl/internal/reduce_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/reduce_by_segment_impl.h
@@ -215,13 +215,17 @@ __sycl_reduce_by_segment(_ExecutionPolicy&& __exec, _Range1&& __keys, _Range2&& 
     using _CustomName = typename _Policy::kernel_name;
 
     using _SegReduceCountKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-        _SegReduceCountPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
+        _SegReduceCountPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate,
+        _BinaryOperator>;
     using _SegReduceOffsetKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-        _SegReduceOffsetPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
+        _SegReduceOffsetPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate,
+        _BinaryOperator>;
     using _SegReduceWgKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-        _SegReduceWgPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
+        _SegReduceWgPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate,
+        _BinaryOperator>;
     using _SegReducePrefixKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-        _SegReducePrefixPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
+        _SegReducePrefixPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate,
+        _BinaryOperator>;
 
     using __diff_type = oneapi::dpl::__internal::__difference_t<_Range3>;
     using __key_type = oneapi::dpl::__internal::__value_t<_Range1>;
--- a/include/oneapi/dpl/internal/reduce_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/reduce_by_segment_impl.h
@@ -215,13 +215,13 @@ __sycl_reduce_by_segment(_ExecutionPolicy&& __exec, _Range1&& __keys, _Range2&& 
     using _CustomName = typename _Policy::kernel_name;
 
     using _SegReduceCountKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-        _SegReduceCountPhase, _CustomName, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
+        _SegReduceCountPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
     using _SegReduceOffsetKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-        _SegReduceOffsetPhase, _CustomName, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
+        _SegReduceOffsetPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
     using _SegReduceWgKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-        _SegReduceWgPhase, _CustomName, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
+        _SegReduceWgPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
     using _SegReducePrefixKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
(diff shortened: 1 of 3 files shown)


SergeyKopienko:
@rarutyun Are these errors detected in our tests?

danhoeflinger:
Looks like we are hitting some compilation issues in the CI with these changes using dpcpp backend.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -632,12 +632,12 @@ struct __parallel_radix_sort_iteration
...
-            __internal::__kernel_name_generator<__count_phase, _CustomName, __decay_t<_InRange>, __decay_t<_TmpBuf>>;
+            __internal::__kernel_name_generator<__count_phase, _CustomName, _ExecutionPolicy __decay_t<_InRange>, __decay_t<_TmpBuf>>;
         using _RadixLocalScanKernel =
-            __internal::__kernel_name_generator<__local_scan_phase, _CustomName, __decay_t<_TmpBuf>>;
-        using _RadixReorderPeerKernel = __internal::__kernel_name_generator<__reorder_peer_phase, _CustomName,
+            __internal::__kernel_name_generator<__local_scan_phase, _CustomName, _ExecutionPolicy __decay_t<_TmpBuf>>;
+        using _RadixReorderPeerKernel = __internal::__kernel_name_generator<__reorder_peer_phase, _CustomName, _ExecutionPolicy
                                                                             __decay_t<_InRange>, __decay_t<_OutRange>>;
-        using _RadixReorderKernel = __internal::__kernel_name_generator<__reorder_phase, _CustomName,
+        using _RadixReorderKernel = __internal::__kernel_name_generator<__reorder_phase, _CustomName, _ExecutionPolicy
```
danhoeflinger, line 640: Looks like these are missing commas, which are causing the compilation errors.

kboyarinov:
> @rarutyun Are these errors detected in our tests?

Currently the issue is not reproducing during our regular testing. I guess we should definately test is somehow, for example execute algorithms with both `device_policy` and `const device_policy` as a regression test for how the issue was found. Of course if binary execution time allows.

mmichel11:
> > @rarutyun Are these errors detected in our tests?
> 
> Currently the issue is not reproducing during our regular testing. I guess we should definately test is somehow, for example execute algorithms with both `device_policy` and `const device_policy` as a regression test for how the issue was found. Of course if binary execution time allows.

As a follow-up to what @SergeyKopienko, do we see an easy way to modify our tests to assess this maybe by modifying `invoke_on_all_hetero_policies` to test with policies of different constness for each run or a subset of runs? I think it would be great if this could be tested in CI to validate what has been manually found but understand if it's too late to do so.

SergeyKopienko:
> > > @rarutyun Are these errors detected in our tests?
> > 
> > 
> > Currently the issue is not reproducing during our regular testing. I guess we should definately test is somehow, for example execute algorithms with both `device_policy` and `const device_policy` as a regression test for how the issue was found. Of course if binary execution time allows.
> 
> As a follow-up to what @SergeyKopienko, do we see an easy way to modify our tests to assess this maybe by modifying `invoke_on_all_hetero_policies` to test with policies of different constness for each run or a subset of runs? I think it would be great if this could be tested in CI to validate what has been manually found but understand if it's too late to do so.

Good idea, @mmichel11 !
I have tried to implement this in my commit https://github.com/oneapi-src/oneDPL/commit/462e82b2e60339492bffb6363c6bfd26204b5c94
Some Kernel name errors were detected.
So it works.
But I am not sure should we apply this commit or not because because the execution time of tests will increase.

In any case, @rarutyun may apply this commit or make the same and compile tests in this branch.

kboyarinov:
> 

Just a minor comment about the implementation - I guess we can simply add two `iterator_invoker`s. Something like:

``
iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(my_policy, op, rest...);
 iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(std::as_const(my_policy), op, ::std::forward<Args>(rest)...);
``

SergeyKopienko:
> > > > > @rarutyun Are these errors detected in our tests?
> > > > 
> > > > 
> > > > Currently the issue is not reproducing during our regular testing. I guess we should definately test is somehow, for example execute algorithms with both `device_policy` and `const device_policy` as a regression test for how the issue was found. Of course if binary execution time allows.
> > > 
> > > 
> > > As a follow-up to what @SergeyKopienko, do we see an easy way to modify our tests to assess this maybe by modifying `invoke_on_all_hetero_policies` to test with policies of different constness for each run or a subset of runs? I think it would be great if this could be tested in CI to validate what has been manually found but understand if it's too late to do so.
> > 
> > 
> > Good idea, @mmichel11 ! I have tried to implement this in my commit [462e82b](https://github.com/oneapi-src/oneDPL/commit/462e82b2e60339492bffb6363c6bfd26204b5c94) Some Kernel name errors were detected. So it works. But I am not sure should we apply this commit or not because because the execution time of tests will increase.
> > In any case, @rarutyun may apply this commit or make the same and compile tests in this branch.
> 
> Just a minor comment about the implementation - I guess we can simply add two `iterator_invoker`s. Something like:
> 
> `iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(my_policy, op, rest...); iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(std::as_const(my_policy), op, ::std::forward<Args>(rest)...);`

Good point, @kboyarinov 
I think we able to apply this approach if we make decision to apply commit https://github.com/oneapi-src/oneDPL/commit/462e82b2e60339492bffb6363c6bfd26204b5c94

SergeyKopienko (dismissed):
LGTM

rarutyun:
> > > > > > @rarutyun Are these errors detected in our tests?
> > > > > 
> > > > > 
> > > > > Currently the issue is not reproducing during our regular testing. I guess we should definately test is somehow, for example execute algorithms with both `device_policy` and `const device_policy` as a regression test for how the issue was found. Of course if binary execution time allows.
> > > > 
> > > > 
> > > > As a follow-up to what @SergeyKopienko, do we see an easy way to modify our tests to assess this maybe by modifying `invoke_on_all_hetero_policies` to test with policies of different constness for each run or a subset of runs? I think it would be great if this could be tested in CI to validate what has been manually found but understand if it's too late to do so.
> > > 
> > > 
> > > Good idea, @mmichel11 ! I have tried to implement this in my commit [462e82b](https://github.com/oneapi-src/oneDPL/commit/462e82b2e60339492bffb6363c6bfd26204b5c94) Some Kernel name errors were detected. So it works. But I am not sure should we apply this commit or not because because the execution time of tests will increase.
> > > In any case, @rarutyun may apply this commit or make the same and compile tests in this branch.
> > 
> > 
> > Just a minor comment about the implementation - I guess we can simply add two `iterator_invoker`s. Something like:
> > `iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(my_policy, op, rest...); iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(std::as_const(my_policy), op, ::std::forward<Args>(rest)...);`
> 
> Good point, @kboyarinov I think we able to apply this approach if we make decision to apply commit [462e82b](https://github.com/oneapi-src/oneDPL/commit/462e82b2e60339492bffb6363c6bfd26204b5c94)

Just to make sure everybody can see it. I believe adding tests for different value categories and cv-qualifiers (`const Ex&`, `Ex&&`, etc.) is overkill. I would rather try to redesign our backend to always to `const _ExecutionPolicy&` to do not introduce extra template instantiations. Yeah, it would be whitebox in a sense we are relying on our backend implementation details, but still I believe it's a right approach. Unfortunately, it requires more thinking, so I won't be able to finish it on time for the upcoming milestone. Let's consider this fix as temporary

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
rarutyun: Thanks, fixed.

include/oneapi/dpl/internal/reduce_by_segment_impl.h
```diff
@@ -215,13 +215,13 @@ __sycl_reduce_by_segment(_ExecutionPolicy&& __exec, _Range1&& __keys, _Range2&&
     using _CustomName = typename _Policy::kernel_name;
 
     using _SegReduceCountKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-        _SegReduceCountPhase, _CustomName, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
+        _SegReduceCountPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _Range4, _BinaryPredicate, _BinaryOperator>;
```
dmitriy-sobolev, line 218: This is optional. Please, consider splitting this long line and other similar lines, which are longer than 120 symbols.

rarutyun: I wanted to apply some clang-format suggestions anyway

SergeyKopienko (approved):
LGTM

mmichel11 (approved):
LGTM

PR #1127, opened 2023-08-30
[Dynamic Selection] Addressing comments for recursive submit and others 

AnuyaWelling2801:
1. Resolved commit regarding recursive submits
2. decay<T>::type to decay_t<T>
3. Removed redundant ==true in if condition


vossmjp (approved):
Looks ok.

PR #1128, opened 2023-08-30
[Dynamic Selection] Tests with offset

AnuyaWelling2801:
Added tests that use offset depending on the size of the universe


AnuyaWelling2801:
Not required

PR #1131, opened 2023-08-31
PSTL offload: check interop between system and overloaded allocation

Alexandr-Konovalov:


test/pstl_offload/memory/usm_memory_replacement.cpp
```diff
@@ -19,6 +19,8 @@
 
 #include "support/utils.h"
 
+#include "system_allocations.h"
```
kboyarinov, line 22: May be it would be better to move this test into the separate test file?
Currently, we use `usm_memory_replacement.pass` as part of subsets in CI and other testings where testing of the memory across TU is not critical.

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,148 @@
...
+typedef struct _IO_FILE FILE;
+
+extern FILE *stderr;
+int fprintf(FILE *stream, const char *format, ...);
+void exit(int status);
+}
+
+// to have native allocation, we must not include possibly overloaded header
+#if DO_STD_INJECTION
+namespace std {
```
kboyarinov, line 24: Could you please explain, what is the reason for not including any C++ standard headers into this file?
If we want a non-replaced `malloc` functions in `perform_system_allocations`, we can either build the second TU without `fsycl-pstl-offload` option or define the `TOP_LEVEL` guarding macro before including `cstdlib` or `new`.

I guess in this case, these strange forward declarations can be removed

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,148 @@
...
+    char *arr_nothrow_new_ptr;
+    char *aligned_new_ptr;
+    char *aligned_nothrow_new_ptr;
+    char *aligned_arr_new_ptr;
+    char *aligned_nothrow_arr_new_ptr;
+
+    static const std::size_t malloc_size = 1024;
+};
+
+static inline void perform_allocations_impl(system_allocs& na) {
```
kboyarinov, line 93: I guess `inline` can be removed

test/pstl_offload/memory/usm_memory_replacement.cpp
Alexandr-Konovalov: Let's discuss. May be I'm not fully understand what is "critical" here, at least it cheap.

I modified existing test vs added new one, because:
* Additional execution time is low vs compile time for a new test.
* Added checks do exactly what is in the test name, i.e. checks "usm memory replacement".

test/pstl_offload/memory/system_allocations.h
Alexandr-Konovalov: Thanks! Guarding macro is much better.

test/pstl_offload/memory/system_allocations.h
Alexandr-Konovalov: Done.

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,148 @@
...
+    na.libc_calloc_ptr = __libc_calloc(10, 10);
+    na.libc_realloc_ptr = __libc_realloc(nullptr, 10);
+    na.libc_memalign_ptr = __libc_memalign(16, 10);
+#endif
+
+    na.new_ptr = new char;
+    na.arr_new_ptr = new char[12];
+    na.nothrow_new_ptr = new (std::nothrow) char;
+    na.arr_nothrow_new_ptr = new (std::nothrow) char[12];
+    na.aligned_new_ptr = new (std::align_val_t(64)) char;
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done.

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,148 @@
...
+
+    static const std::size_t malloc_size = 1024;
+};
+
+static inline void perform_allocations_impl(system_allocs& na) {
+    na.malloc_ptr = malloc(system_allocs::malloc_size);
+    na.calloc_ptr = calloc(10, 10);
+    na.realloc_ptr = realloc(nullptr, 10);
+#if __linux__
+    EXPECT_TRUE(malloc_usable_size(na.malloc_ptr) >= system_allocs::malloc_size, "Invalid object size");
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done.

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,148 @@
...
+    na.aligned_nothrow_arr_new_ptr = new (std::align_val_t(64), std::nothrow) char[12];
+}
+
+static inline void perform_deallocations_impl(const system_allocs& na) {
+#if __linux__
+    EXPECT_TRUE(malloc_usable_size(na.malloc_ptr) >= system_allocs::malloc_size, "Invalid object size");
+#endif
+    free(na.malloc_ptr);
+    free(na.calloc_ptr);
+    (void)realloc(na.realloc_ptr, 0);
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done. Let's just call `free`.

test/pstl_offload/memory/usm_memory_replacement.cpp
```diff
@@ -160,5 +162,100 @@ int main() {
...
+        EXPECT_TRUE(sycl::get_pointer_type(na.aligned_arr_new_ptr, memory_context) == sycl::usm::alloc::unknown,
+            "Expect pointer from system allocation");
+        EXPECT_TRUE(sycl::get_pointer_type(na.aligned_nothrow_arr_new_ptr, memory_context) == sycl::usm::alloc::unknown,
+            "Expect pointer from system allocation");
+
+        perform_deallocations_impl(na);
+    }
+    // check that system deallocation able to release memory allocated via overloaded allocation
+    {
+        system_allocs na;
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done.

test/pstl_offload/memory/usm_memory_replacement.cpp
```diff
@@ -160,5 +162,100 @@ int main() {
...
+        perform_allocations_impl(na);
+
+        EXPECT_TRUE(sycl::get_pointer_type(na.malloc_ptr, memory_context) == sycl::usm::alloc::shared,
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done.

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,148 @@
...
+    free(na.libc_calloc_ptr);
+    free(na.libc_realloc_ptr);
+    free(na.libc_memalign_ptr);
+#endif
+
+    delete na.new_ptr;
+    delete[] na.arr_new_ptr;
+    delete na.nothrow_new_ptr;
+    delete[] na.arr_nothrow_new_ptr;
+    operator delete(na.aligned_new_ptr, std::align_val_t(64));
```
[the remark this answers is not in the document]
kboyarinov: Yes, as I see it is not critical for the test logic to have an array of exactly N elements allocated with the new expression.
I think having just a pointer to the single element should be enough.

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,120 @@
...
+
+static void check_true(bool expected, bool condition, const char* file, int line, const char* message) {
+    if (condition != expected) {
+        fprintf(stderr, "error at %s:%d - %s\n", file, line, message);
+        exit(1);
+    }
+}
+
+#ifndef EXPECT_TRUE
+#define EXPECT_TRUE(condition, message) check_true(true, condition, __FILE__, __LINE__, message)
```
kboyarinov, line 27: I guess we can include `support/utils.h`  under the guard and remove these assertions as well

test/pstl_offload/memory/usm_memory_replacement_native_alloc.cpp
```diff
@@ -0,0 +1,23 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#if !__SYCL_PSTL_OFFLOAD__
+#error "PSTL offload compiler mode should be enabled to run this test"
+#endif
+
+#define DO_STD_INJECTION 1
```
kboyarinov, line 14: It seems like this is redundant

test/pstl_offload/memory/usm_memory_replacement_native_alloc.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "system_allocations.h"
+
+void perform_system_allocations(allocs& na) {
+    perform_allocations_impl(na);
```
kboyarinov, line 18: I would also add a comment that this cpp file emulates system allocations while compiling with pstl offload option

test/pstl_offload/memory/usm_memory_replacement_native_alloc.cpp
Alexandr-Konovalov: Done.

test/pstl_offload/memory/system_allocations.h
Alexandr-Konovalov: Done.

test/pstl_offload/memory/usm_memory_replacement_native_alloc.cpp
Alexandr-Konovalov: Done.

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,121 @@
...
+
+    na.new_ptr = new char;
+    na.arr_new_ptr = new char[allocs::array_size];
+    na.nothrow_new_ptr = new (std::nothrow) char;
+    na.arr_nothrow_new_ptr = new (std::nothrow) char[allocs::array_size];
+    na.aligned_new_ptr = new (std::align_val_t(allocs::alignment)) char;
+    na.aligned_nothrow_new_ptr = new (std::align_val_t(allocs::alignment), std::nothrow) char;
+    na.aligned_arr_new_ptr = new (std::align_val_t(allocs::alignment)) char[allocs::array_size];
+    na.aligned_nothrow_arr_new_ptr =
+        new (std::align_val_t(allocs::alignment), std::nothrow) char[allocs::array_size];
```
kboyarinov, line 86: I would rewrite this part as well to use `operator new` instead of new expression. Currently, if non-trivial types would be used in the test, the explicit destructor call would be required before calling `operator delete`.
```suggestion
    na.new_ptr = ::operator new(allocs::size);
    na.arr_new_ptr = ::operator new[](allocs::size);
    na.nothrow_new_ptr = ::operator new (allocs::size, std::nothrow);
    na.arr_nothrow_new_ptr = ::operator new[](allocs::size, std::nothrow);
    na.aligned_new_ptr = ::operator new (allocs::size, std::align_val_t(allocs::alignment));
    na.aligned_nothrow_new_ptr = ::operator new (allocs::size, std::align_val_t(allocs::alignment), std::nothrow);
    na.aligned_arr_new_ptr = ::operator new[] (allocs::size, std::align_val_t(allocs::alignment));
    na.aligned_nothrow_arr_new_ptr = ::operator
        new (allocs::size, std::align_val_t(allocs::alignment), std::nothrow);
```

Where `allocs::size` can be `sizeof(char)`

test/pstl_offload/memory/usm_memory_replacement.cpp
```diff
@@ -160,5 +203,20 @@ int main() {
...
     }
 
+    // check that it's possible to release memory allocated via system allocation
+    {
+        allocs na;
+        perform_system_allocations(na);
+        check_memory_ownership(na, sycl::usm::alloc::unknown);
+        perform_deallocations_impl(na);
+    }
+    // check that system deallocation able to release memory allocated via overloaded allocation
```
kboyarinov, line 213: Actually, we don't test system deallocation function - it is our global replacement functions. May be it would be better to rewrite both comments to something like "Check the ability to deallocate USM memory from the translation unit built without PSTL offload mode"? What do you think?

test/CMakeLists.txt
```diff
@@ -201,3 +204,9 @@ foreach (_file IN LISTS UNIT_TESTS)
     endif()
 
 endforeach()
+
+if (NOT _ONEDPL_PSTL_OFFLOAD STREQUAL off)
+    file(GLOB_RECURSE _test_file "pstl_offload/memory/usm_memory_replacement.cpp")
+    file(GLOB_RECURSE _test_file1 "pstl_offload/memory/usm_memory_replacement_native_alloc.cpp")
+    onedpl_add_test(${_test_file} true ${_test_file1} ${_test_file2})
```
kboyarinov, line 211: It seems like we don't have a make target named `usm_memory_replacement.pass`, now it is named `usm_memory_replacement`. I think that is why we see CI failures in PSTL offload mode. 
We need to either modify the CI configuration or replace the part with 2 TUs into the separate test, as I mentioned before. Lets discuss separately.

test/pstl_offload/memory/usm_memory_replacement.cpp
Alexandr-Konovalov: Good point, the terminology is not consistent. Are names below clear or something different it's better?
* native allocation
* global allocation
* locally-overloaded allocation

test/CMakeLists.txt
Alexandr-Konovalov: It seems suffix pass is expected to be meaningful. At least is libc++ I see `*.verify.cpp` in addition to pass. So I may want to restore it.

kboyarinov: I think restoration of `.pass` suffix is ok, but in this case `usm_memory_replacement.pass` should be skipped in the main test processing loop in CMake to avoid issues with test target duplication.

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,121 @@
...
+
+#ifndef _SYSTEM_ALLOCATIONS_H
+#define _SYSTEM_ALLOCATIONS_H
+
+#define _ONEDPL_PSTL_OFFLOAD_TOP_LEVEL
+#include <new>
+#include <cstdlib>
+#include <malloc.h>
+#include "support/utils.h"
+
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done.

test/pstl_offload/memory/system_allocations.h
Alexandr-Konovalov: Done.

test/pstl_offload/memory/system_allocations.h
```diff
@@ -0,0 +1,121 @@
...
+extern "C"
+{
+void *__libc_malloc(std::size_t);
+void *__libc_calloc(std::size_t, std::size_t);
+void *__libc_memalign(std::size_t, std::size_t);
+void *__libc_realloc(void *, std::size_t);
+}
+
+#if TEST_DELETE_OVERLOAD
```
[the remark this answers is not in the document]
Alexandr-Konovalov: I added a check to not forget to extend the test if/when we got a compiler with `__cpp_sized_deallocation`.

test/pstl_offload/memory/usm_memory_replacement.cpp
Alexandr-Konovalov: I tried to rewrite the comments.

test/pstl_offload/memory/allocation_utils.h
```diff
@@ -0,0 +1,127 @@
...
+    free(na.libc_calloc_ptr);
+    free(na.libc_realloc_ptr);
+    free(na.libc_memalign_ptr);
+#endif // __linux__
+
+    operator delete(na.new_ptr);
+    operator delete[](na.arr_new_ptr);
+#if TEST_DELETE_OVERLOAD
+    operator delete(na.nothrow_new_ptr, allocs::alloc_size);
+    operator delete[](na.arr_nothrow_new_ptr, allocs::array_size*allocs::alloc_size);
```
kboyarinov, line 113: It seems like there is a mismatch in number of bytes allocated and deallocated here - we allocate `alloc_size` and deallocate `alloc_size * array_size`. I guess `array_size` should be removed. And after removal, `array_size` would be unused and the definition also can be removed.

test/pstl_offload/memory/usm_memory_replacement_system_alloc.cpp
```diff
@@ -0,0 +1,25 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#if !__SYCL_PSTL_OFFLOAD__
+#error "PSTL offload compiler mode should be enabled to run this test"
+#endif
+
+// in this translation unit we use system allocations while compiling with pstl offload option,
```
kboyarinov, line 14: Could you please also mention that this test is a second TU for usm_memory_replacement.pass.cpp?

test/pstl_offload/memory/usm_memory_replacement.pass.cpp
```diff
@@ -37,6 +38,48 @@ void test_new(std::size_t count, NewArgs... new_args) {
     ::operator delete(ptr, new_args...);
 }
 
+void check_memory_ownership(const allocs &na, sycl::usm::alloc expected_type) {
+        EXPECT_TRUE(sycl::get_pointer_type(na.malloc_ptr, memory_context) == expected_type,
+            "Expect pointer from system allocation");
```
kboyarinov, line 43: It seems like the error message should be also changed to something like "Unexpected pointer type"

test/pstl_offload/memory/allocation_utils.h
```diff
@@ -0,0 +1,127 @@
...
+void *__libc_calloc(std::size_t, std::size_t);
+void *__libc_memalign(std::size_t, std::size_t);
+void *__libc_realloc(void *, std::size_t);
+}
+
+#if __cpp_sized_deallocation
+#error "The test must be updated to cover a compiler with sized decallocation support."
+#endif
+
+#if TEST_DELETE_OVERLOAD
```
kboyarinov, line 32: I would suggest to rename this macro to `TEST_DELETE_SIZED_OVERLOAD`

test/pstl_offload/memory/usm_memory_replacement.cpp
kboyarinov: I guess you are right. And also `make usm_memory_replacement.pass` works now, so I am closing this issue.

test/pstl_offload/memory/allocation_utils.h
Alexandr-Konovalov: Done.

test/pstl_offload/memory/usm_memory_replacement_system_alloc.cpp
Alexandr-Konovalov: Done.

test/pstl_offload/memory/usm_memory_replacement.pass.cpp
Alexandr-Konovalov: Done.

test/pstl_offload/memory/allocation_utils.h
Alexandr-Konovalov: This macro was eliminated.

test/pstl_offload/memory/usm_memory_replacement.pass.cpp
```diff
@@ -160,5 +196,20 @@ int main() {
         EXPECT_TRUE(errno == ENOMEM, "Incorrect errno");
     }
 
+    // check the ability to release system memory allocated inside another translation unit without local allocation overload
+    {
+        allocs na;
+        perform_system_allocations(na);
```
kboyarinov, line 202:
```suggestion
        perform_system_allocations(na); // declared in usm_memory_replacement_system_alloc.cpp
```

test/pstl_offload/memory/usm_memory_replacement.pass.cpp
```diff
@@ -160,5 +196,20 @@ int main() {
...
+        check_memory_ownership(na, sycl::usm::alloc::unknown);
+        perform_deallocations_impl(na);
+    }
+    // check the ability to release USM memory inside another translation unit without local allocation overload
+    {
+        allocs na;
+        perform_allocations_impl(na);
+        check_memory_ownership(na, sycl::usm::alloc::shared);
+        perform_system_deallocations(na);
```
kboyarinov, line 211:
```suggestion
        perform_system_deallocations(na); // declared in usm_memory_replacement_system_alloc.cpp
```

test/CMakeLists.txt
```diff
@@ -201,3 +204,9 @@ foreach (_file IN LISTS UNIT_TESTS)
...
+endif()
```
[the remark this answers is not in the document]
Alexandr-Konovalov: I tried to support it in semi-universal way.

test/pstl_offload/memory/usm_memory_replacement.pass.cpp
kboyarinov: Was discussed offline

test/pstl_offload/memory/usm_memory_replacement.pass.cpp
kboyarinov: Was discussed offline

test/pstl_offload/memory/usm_memory_replacement.cpp
kboyarinov: Since `usm_memory_replacement.pass` is part of testing subset on the compiler side, let's provide only single-TU tests to them and hence move this test into separate file.

Alexandr-Konovalov: Done, this is sound reason.

kboyarinov (dismissed):
LGTM

test/CMakeLists.txt
```diff
@@ -67,7 +67,14 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
     get_filename_component(_test_name ${test_source_file} NAME)
     string(REPLACE "\.cpp" "" _test_name ${_test_name})
 
-    add_executable(${_test_name} EXCLUDE_FROM_ALL "${test_source_file}")
+    # add additional translation units to current test according to mask
```
danhoeflinger, line 70: Do you think we will ever want the same extra translational unit (`shared_util.cpp`) to be added for multiple different tests?  

If so, it may be better to instead have some list for each multi translational unit test which CMake could parse to find out the extra translational units that are required.  With the current method, we would need multiple copies of the same file named differently to achieve that.  

Instead you could have a file `${_test_name_without_pass}_extra_TU_List.txt` which contains a list of extra source files to add.  CMake can check if that file exists and iterate through the list if it does.  

On the other hand, if we never think this is useful, it isn't worth the extra complexity.

test/CMakeLists.txt
```diff
@@ -67,7 +67,14 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
...
+    get_filename_component(_test_path ${test_source_file} DIRECTORY)
+    string(REPLACE "\.pass" "" _test_name_without_pass ${_test_name})
+    file(GLOB_RECURSE additional_src_files "${_test_path}/${_test_name_without_pass}_*.cpp")
```
danhoeflinger, line 73: I think `GLOB`  with `${_test_path}` specified as the directory seems to be the intention here instead of `GLOB_RECURSE`, which searches all subdirectories.

test/CMakeLists.txt
Alexandr-Konovalov: That is interesting. Personally I'd prefer more explicit approach contrary to "everything covered by some glob mask". But probably cmake is powerful enough, so this can be expressed in cmake. What do you think?
```
set_property(SOURCE ${CMAKE_CURRENT_LIST_DIR}/pstl_offload/memory/interop_allocs.pass.cpp
    APPEND PROPERTY SRC "pstl_offload/memory/interop_allocs_system.cpp;pstl_offload/memory/interop_allocs_another.cpp")
...
    get_property(additional_src_files SOURCE ${test_source_file} PROPERTY SRC)
    if (DEFINED additional_src_files)
        unset(files)
        foreach(f ${additional_src_files})
            file(GLOB f ${f})
            list(APPEND files ${f})
        endforeach()
        set(additional_src_files ${files})
    endif()
```

danhoeflinger: Yes, this is probably preferable to some extra file to parse.  I was formatting it like that because up until this point, we have relied on the files structure to dictate the tests rather than having some explicit test listing.  My proposed change would keep the CMake code generic and unaware of the specifics of the tests.  However, I have no real preference that it stays that way.

I think since instances of these multi-TU tests are likely rare, it is manageable to list them explicitly in CMake like this.  If you think that many of these tests may exist, then perhaps I would prefer something more generic.

test/CMakeLists.txt
Alexandr-Konovalov: Fixed.

test/CMakeLists.txt
```diff
@@ -57,6 +57,10 @@ add_custom_target(run-onedpl-tests
     DEPENDS build-onedpl-tests
     COMMENT "Build and run all oneDPL tests")
 
+# proterty's value is list with additional TUs to be linked together with property's name
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Fixed.

test/CMakeLists.txt
danhoeflinger: I also don't know of an idiom for this.  What you've done here looks reasonable to me.

dmitriy-sobolev: I am not aware of such an idiom as well, but [target_sources](https://cmake.org/cmake/help/latest/command/target_sources.html) may be a good candidate to simplify the logic. It adds source files into the existing target and its dependent targets. 

You can place this after `foreach (_file IN LISTS UNIT_TESTS)` loop. 
```
if (TARGET interop_allocs.pass)
    target_sources(interop_allocs.pass PRIVATE "pstl_offload/memory/interop_allocs_system.cpp")
endif()

```

Still, it is a bit hacky, but it is simpler. 

Probably, it also makes sense to define these customized targets in separate CMakeLists.txt, but I have not investigated this approach yet. I will let you know if I find something.

Alexandr-Konovalov: Thanks, this is definitely more clear.

dmitriy-sobolev (commented):
CMake part looks good enough to me.

danhoeflinger:
CMake part LGTM, I have not reviewed the full PR

test/pstl_offload/memory/interop_allocs_system.cpp
```diff
@@ -0,0 +1,25 @@
+// -*- C++ -*-
```
rarutyun, line 1: What if we just do a symmetric translation unit with USM that has a function like `perform_usm_allocations` and the main what just call `perform_usm_allocations` and `perform_system_allocations` and check the results?

test/pstl_offload/memory/allocation_utils.h
```diff
@@ -0,0 +1,119 @@
...
+    operator delete[](na.arr_new_ptr);
+    operator delete(na.nothrow_new_ptr, allocs::alloc_size);
+    operator delete[](na.arr_nothrow_new_ptr, allocs::alloc_size);
+    operator delete(na.aligned_new_ptr, std::align_val_t(allocs::alignment));
+    operator delete(na.aligned_nothrow_new_ptr, std::align_val_t(allocs::alignment));
+    operator delete[](na.aligned_arr_new_ptr, std::align_val_t(allocs::alignment));
+    operator delete[](na.aligned_nothrow_arr_new_ptr, std::align_val_t(allocs::alignment));
+}
+
+void perform_system_allocations(allocs&);
```
rarutyun, line 116: How about applying the following suggestion to have a more symmetric API with malloc/free? 
```suggestion
allocs perform_system_allocations();
```

test/pstl_offload/memory/allocation_utils.h
```diff
@@ -0,0 +1,119 @@
...
+#define _ALLOCATION_UTILS_H
+
+#define _ONEDPL_PSTL_OFFLOAD_TOP_LEVEL
+#include <new>
+#include <cstdlib>
+#if __linux__
+#include <malloc.h>
+#endif
+#include "support/utils.h"
+#undef _ONEDPL_PSTL_OFFLOAD_TOP_LEVEL
```
rarutyun, line 20: what if we just extract these lines to a separate header (or just include those in the file where we have ``perform_system_allocations`` functions? In this case utils would just be utils and the decision about memory redirection would be done by translation unit.

I am not sure if it would be better or not but I am suggesting to experiment with that.

test/pstl_offload/memory/interop_allocs.pass.cpp
```diff
@@ -0,0 +1,80 @@
...
+//===----------------------------------------------------------------------===//
+
+#if !__SYCL_PSTL_OFFLOAD__
+#error "PSTL offload compiler mode should be enabled to run this test"
+#endif
+
+#include "allocation_utils.h"
+
+// in this translation unit we have overload to USM
+#include <new> // include just to provide the local allocation overload
```
rarutyun, line 17:
```suggestion
// in this translation unit we redirect memory to USM
#include <new> // Necessary to make memory redirection
```

test/pstl_offload/memory/interop_allocs_system.cpp
```diff
@@ -0,0 +1,25 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#if !__SYCL_PSTL_OFFLOAD__
+#error "PSTL offload compiler mode should be enabled to run this test"
+#endif
+
+// in this translation unit (TU) we use system allocations while compiling with pstl offload option,
+// it should be no overload to USM, that is done in another TU
```
rarutyun, line 15:
```suggestion
// it should be no redirection to USM
```

Alexandr-Konovalov: Done.

test/pstl_offload/memory/allocation_utils.h
```diff
@@ -0,0 +1,119 @@
...
+    void *arr_new_ptr;
+    void *nothrow_new_ptr;
+    void *arr_nothrow_new_ptr;
+    void *aligned_new_ptr;
+    void *aligned_nothrow_new_ptr;
+    void *aligned_arr_new_ptr;
+    void *aligned_nothrow_arr_new_ptr;
+
+    static const std::size_t alloc_size = 1024;
+    static const std::size_t alignment = 16;
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done.

test/pstl_offload/memory/interop_allocs_system.cpp
Alexandr-Konovalov: Done.

test/pstl_offload/memory/allocation_utils.h
Alexandr-Konovalov: Done.

test/pstl_offload/memory/allocation_utils.h
Alexandr-Konovalov: Done.

test/pstl_offload/memory/interop_allocs.pass.cpp
Alexandr-Konovalov: "Overload" was dropped, but wording is different.

test/CMakeLists.txt
```diff
@@ -199,3 +199,9 @@ foreach (_file IN LISTS UNIT_TESTS)
     endif()
 
 endforeach()
+
+# add additional TUs if required
+if (TARGET interop_allocs.pass)
+    target_sources(interop_allocs.pass PRIVATE "pstl_offload/memory/interop_allocs_system.cpp")
+    target_sources(interop_allocs.pass PRIVATE "pstl_offload/memory/interop_allocs_usm.cpp")
```
dmitriy-sobolev, line 206: The minimum supported cmake is 3.11, however `target_sources` can handle relative paths only since 3.13. 
 `${CMAKE_CURRENT_LIST_DIR}` provides the path to the folder with the currently processed `CMakeLists.txt`. 

Also, `target_source` supports multiple items.

```suggestion
    target_sources(interop_allocs.pass PRIVATE 
                   "${CMAKE_CURRENT_LIST_DIR}/pstl_offload/memory/interop_allocs_system.cpp"
                   "${CMAKE_CURRENT_LIST_DIR}/pstl_offload/memory/interop_allocs_usm.cpp")
```

Alexandr-Konovalov: Done.

test/pstl_offload/memory/interop_allocs.pass.cpp
```diff
@@ -0,0 +1,76 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#if !__SYCL_PSTL_OFFLOAD__
+#error "PSTL offload compiler mode should be enabled to run this test"
+#endif
+
+#include "interop_allocs_headers.h"
```
rarutyun, line 14:
```suggestion
#include "support/utils.h"
```

rarutyun: Discussed. This header is necessary.

rarutyun (approved):

PR #1132, opened 2023-08-31
[Dynamic Selection] Removed backend defs

AnuyaWelling2801:
1. Removed usage of backend_defs.h in sycl_backend.h and inline_backend.h
2. Modified the execution resource to be the native resource in sycl_backend.h
3. Added a basic_execution_resource_t as a wrapper over native_resource in inline_backend.h


vossmjp (approved):
I wonder if we can even simplify this more by requiring backends to only define execution_resource_t if it is different than resource_type.  Right now policies look at both and assume they are different. Maybe this is something to do later.

PR #1133, opened 2023-08-31
[Dynamic Selection] No more recursion for submit and wait

AnuyaWelling2801:


include/oneapi/dpl/internal/dynamic_selection_traits.h
```diff
@@ -152,10 +152,10 @@ namespace internal {
         return std::forward<T>(t).submit_and_wait(std::forward<Function>(f), std::forward<Args>(args)...);
       } else if constexpr ( internal::has_submit_and_wait_handle<T, typename std::decay_t<T>::selection_type, Function, Args...>::value) {
         // has the optional submit_and_wait for a selection, so select and call
-        return submit_and_wait(std::forward<T>(t).select(f, args...), std::forward<Function>(f), std::forward<Args>(args)...);
+        return std::forward<T>(t).submit_and_wait(t.select(f, args...), std::forward<Function>(f), std::forward<Args>(args)...);
       } else {
         // does not have the optional submit_and_wait(f, args...) or (s, f, args...)
-        return wait(submit(std::forward<T>(t), std::forward<Function>(f), std::forward<Args>(args)...));
+        return wait(std::forward<T>(t).submit(t.select(f, args...), std::forward<Function>(f), std::forward<Args>(args)...));
```
vossmjp, line 158: This first change look ok. But we wanted to keep this one at line 158 as-is since the policy might have a customization for submit(f, args...). With this change we would force the use of select and not choose that customization of submit if it exists.

vossmjp (changes requested):

PR #1153, opened 2023-09-06
[oneDPL] + a fix for the oneDPL merge (merge Path)

MikeDvorskiy:
[oneDPL] + a fix for the oneDPL merge (merge Path).
There was a kind of mess with ranges order to be merged, r1 and r2. 
A "merge" matrix was define for r2 and r1, but real merging for r1 and r2.
Actually a "merge" matrix has been "transposed" for the fix.

### DIFF
--- a/test/parallel_api/algorithm/alg.merge/merge.pass.cpp
+++ b/test/parallel_api/algorithm/alg.merge/merge.pass.cpp
@@ -150,6 +150,19 @@ struct test_non_const
     }
 };
 
+struct test_merge_tuple
+{
+    template <typename Policy, typename InputIterator1, typename InputIterator2, typename OutputIterator,
+    typename Compare, typename Checker>
+    void
+    operator()(Policy&& exec, InputIterator1 first1, InputIterator1 last1, InputIterator2 first2, InputIterator2 last2,
+               OutputIterator out_first, Compare comp, Checker check)
+    {
+        std::merge(exec, first1, last1, first2, last2, out_first, comp);
+        check();
+    }
+};
+
 int
 main()
 {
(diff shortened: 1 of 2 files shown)


mmichel11:
This fix with the merge matrix transpose looks correct, but I just want to clarify my understanding of the problem.

Is my understanding correct that `__find_start_point` was considering an `n_1 x n_2` merge matrix (sizes of range 1 and 2) while the `__serial_merge` phase was considering an `n_2 x n_1` merge matrix which causes a disconnect that leads to the encountered failures?

MikeDvorskiy:
> This fix with the merge matrix transpose looks correct, but I just want to clarify my understanding of the problem.
> 
> Is my understanding correct that `__find_start_point` was considering an `n_1 x n_2` merge matrix (sizes of range 1 and 2) while the `__serial_merge` phase was considering an `n_2 x n_1` merge matrix which causes a disconnect that leads to the encountered failures?

yes, like that, but otherwise - 
"__find_start_point was considering an n_2 x n_1, while the `__serial_merge` phase was considering an `n_1 x n_2` merge

test/parallel_api/algorithm/alg.merge/merge.pass.cpp
```diff
@@ -165,5 +175,26 @@ main()
     test_algo_basic_double<std::int32_t>(run_for_rnd_fw<test_non_const<std::int32_t>>());
 #endif
 
+    using T = std::tuple<std::int32_t, std::int32_t>; //a pair (key, value)
```
dmitriy-sobolev, line 178: Could you make sure that std::tuple can be used in a kernel when compiling with an oldest supported compiler: Intel® oneAPI DPC++/C++ Compiler 2022.3?

dmitriy-sobolev: I checked some issues on intel-llvm repository related to using std::tuple in kernels and I think, the support was enabled in 2021 major release or even earlier (e.g. https://github.com/intel/llvm/issues/4806 is related to 2021.4)

test/parallel_api/algorithm/alg.merge/merge.pass.cpp
```diff
@@ -165,5 +175,26 @@ main()
...
+    std::vector<T> a = { {1, 2}, {1, 2}, {1,2}, {1,2}, {1, 2}, {1, 2} };
+    std::vector<T> b = { {1, 1}, {1, 1}, {1, 1}, {1, 1}, {1, 1}, {1, 1}, {1, 1} };
+    std::vector<T> merged(a.size() + b.size());
+
+    auto comp = [](auto a, auto b) { return std::get<0>(b) < std::get<0>(a); }; //greater by key
+
+    invoke_on_all_policies<6>()(test_merge_compare<T>(), a.begin(), a.end(), b.cbegin(), b.cend(), merged.begin(), comp,
```
dmitriy-sobolev, line 185: Could you wrap the case into some function, e.g. `test_merge_zip` and pass e.g. `100` as a kernel name augmentation in order to avoid unintentionally making kernel duplicates upon changing other parts of the test?

dmitriy-sobolev (commented):
The changes in the algorithm look good to me. However, there are suggestions for keeping the test tidy.

test/parallel_api/algorithm/alg.merge/merge.pass.cpp
MikeDvorskiy: rather `test_merge_tuple`, because we have not explicit zip_iterator usage here.

dmitriy-sobolev (approved):
LGTM

mmichel11 (approved):
LGTM

Issue #1060, opened 2023-08-17
Using oneDPL forces client to always run with "SYCL device" present


krasznaa:
I seem to have run into a pretty fundamental design issue in the code. :frowning: As discussed in https://github.com/acts-project/traccc/pull/442, apparently a oneDPL using application can never be executed on a host that has no "SYCL capable devices" at all.

In our applications we've seen a couple of times already that we must not create `sycl::queue` objects in a global scope. We use [gtest_discover_tests(...)](https://cmake.org/cmake/help/latest/module/GoogleTest.html#command:gtest_discover_tests) to set up tests in our projects to [CTest](https://cmake.org/cmake/help/latest/manual/ctest.1.html). But for this to work in a CI, which generally doesn't provide us with "SYCL capable devices", the test code must only try to create `sycl::queue`-s once a test is actually running. (In practice this just meant not ever declaring "global" `sycl::queue` objects.)

Unfortunately as soon as we include anything from oneDPL, our tests fail to even build in our CI system. :frowning: Because oneDPL itself apparently creates at least one (but as far as I can see actually multiple) `sycl::queue` object globally just by being included.

```
Thread 1 "traccc_test_syc" hit Catchpoint 1 (exception thrown), 0x00007ffff1cd1662 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
(gdb) bt
#0  0x00007ffff1cd1662 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#1  0x00007ffff1a74613 in sycl::_V1::detail::select_device(std::function<int (sycl::_V1::device const&)>, std::vector<sycl::_V1::device, std::allocator<sycl::_V1::device> >&) () from /software/intel/oneapi-2023.1.0/compiler/2023.1.0/linux/lib/libsycl.so.6
#2  0x00007ffff1a750b9 in sycl::_V1::detail::select_device(std::function<int (sycl::_V1::device const&)> const&) () from /software/intel/oneapi-2023.1.0/compiler/2023.1.0/linux/lib/libsycl.so.6
#3  0x00007ffff1a757fc in sycl::_V1::device_selector::select_device() const () from /software/intel/oneapi-2023.1.0/compiler/2023.1.0/linux/lib/libsycl.so.6
[...]

PR #1154, opened 2023-09-06
Lazy initialization of queue inside dpcpp_default policy

adamfidel:
There are a few potential issues with oneDPL's global `sycl::queue`, including:

- Requirement that a SYCL device is available whenever oneDPL is included as mentioned in #1060
- Static initialization order fiasco with regard to queues being defined in other translation units

To address the issue, this PR lazily initializes the `sycl::queue` inside the `dpcpp_default` execution policy on first use.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -48,48 +48,52 @@ class device_policy
     template <typename>
     friend class device_policy;
 
+    template <typename T>
+    static auto
+    lock_and_forward(T&& t, std::mutex& mtx)
+    {
+        ::std::scoped_lock lock{mtx};
+        return std::forward<T>(t);
+    }
+
   public:
     using kernel_name = KernelName;
 
     device_policy() = default;
+    explicit device_policy(sycl::queue q_) : q(q_) {}
+    explicit device_policy(sycl::device d_) { q.emplace(d_); }
 
     template <typename OtherName>
-    device_policy(const device_policy<OtherName>& other)
+    device_policy(const device_policy<OtherName>& other) : q(device_policy::lock_and_forward(other.q, other.mtx))
     {
-        ::std::scoped_lock lock{other.mtx};
-        q = other.q;
     }
 
-    device_policy(const device_policy& other)
-    {
-        ::std::scoped_lock lock{other.mtx};
-        q = other.q;
-    }
+    device_policy(const device_policy& other) : q(device_policy::lock_and_forward(other.q, other.mtx)) {}
 
-    device_policy(device_policy&& other)
-    {
-        ::std::scoped_lock lock{other.mtx};
-        q.swap(other.q);
-    }
+    device_policy(device_policy&& other) : q(device_policy::lock_and_forward(::std::move(other.q), other.mtx)) {}
 
     device_policy&
     operator=(const device_policy& other)
     {
-        ::std::scoped_lock lock{mtx, other.mtx};
-        q = other.q;
+        if (this != &other)
+        {
+            ::std::scoped_lock lock{mtx, other.mtx};
+            q = other.q;
+        }
         return *this;
     }
 
     device_policy&
     operator=(device_policy&& other)
     {
-        ::std::scoped_lock lock{mtx, other.mtx};
-        q.swap(other.q);
+        if (this != &other)
+        {
+            ::std::scoped_lock lock{mtx, other.mtx};
+            q = ::std::move(other.q);
+        }
         return *this;
     }
 
-    explicit device_policy(sycl::queue q_) : q(q_) {}
-    explicit device_policy(sycl::device d_) { q.emplace(d_); }
     operator sycl::queue() const { return queue(); }
     sycl::queue
     queue() const
@@ -121,8 +125,8 @@ class device_policy
     }
 
   protected:
-    mutable ::std::optional<sycl::queue> q;
     mutable ::std::mutex mtx;
+    mutable ::std::optional<sycl::queue> q;
 };
 
 #if _ONEDPL_FPGA_DEVICE
@@ -137,7 +141,9 @@ class fpga_policy : public device_policy<KernelName>
 
     fpga_policy() = default;
     template <unsigned int other_factor, typename OtherName>
-    fpga_policy(const fpga_policy<other_factor, OtherName>& other) : base(other.queue()){};
+    fpga_policy(const fpga_policy<other_factor, OtherName>& other) : base(other.queue())
+    {
+    }
     explicit fpga_policy(sycl::queue q) : base(q) {}
     explicit fpga_policy(sycl::device d) : base(d) {}
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -51,12 +51,14 @@ class device_policy
     {
     }
     explicit device_policy(sycl::queue q_) : q(q_) {}
-    explicit device_policy(sycl::device d_) : q(d_) {}
-    operator sycl::queue() const { return q; }
+    explicit device_policy(sycl::device d_) { q.emplace(d_); }
+    operator sycl::queue() const { return queue(); }
     sycl::queue
     queue() const
     {
-        return q;
+        if (!q)
+            q.emplace();
+        return *q;
     }
 
     // For internal use only
@@ -78,7 +80,7 @@ class device_policy
     }
 
   private:
-    sycl::queue q;
+    mutable ::std::optional<sycl::queue> q;
 };
 
 #if _ONEDPL_FPGA_DEVICE
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -106,9 +106,9 @@ class fpga_policy : public device_policy<KernelName>
         if (!this->q)
             this->q.emplace(
 #    if _ONEDPL_FPGA_EMU
-              __dpl_sycl::__fpga_emulator_selector()
+                __dpl_sycl::__fpga_emulator_selector()
 #    else
-              __dpl_sycl::__fpga_selector()
+                __dpl_sycl::__fpga_selector()
 #    endif // _ONEDPL_FPGA_EMU
             );
         return *this->q;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -55,15 +55,17 @@ class device_policy
 
     device_policy(const device_policy& other) : q(other.queue()) {}
     device_policy(device_policy&& other) : q(other.queue()) {}
-    device_policy& operator=(const device_policy& other)
+    device_policy&
+    operator=(const device_policy& other)
     {
-      q = other.queue();
-      return *this;
+        q = other.queue();
+        return *this;
     }
-    device_policy& operator=(device_policy&& other)
+    device_policy&
+    operator=(device_policy&& other)
     {
-      q.swap(other.q);
-      return *this;
+        q.swap(other.q);
+        return *this;
     }
 
     explicit device_policy(sycl::queue q_) : q(q_) {}
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -21,6 +21,8 @@
 
 #include "sycl_defs.h"
 
+#include <mutex>
+#include <optional>
 #include <type_traits>
 
 namespace oneapi
@@ -50,6 +52,20 @@ class device_policy
     device_policy(const device_policy<OtherName>& other) : q(other.queue())
     {
     }
+
+    device_policy(const device_policy& other) : q(other.queue()) {}
+    device_policy(device_policy&& other) : q(other.queue()) {}
+    device_policy& operator=(const device_policy& other)
+    {
+      q = other.queue();
+      return *this;
+    }
+    device_policy& operator=(device_policy&& other)
+    {
+      q.swap(other.q);
+      return *this;
+    }
+
     explicit device_policy(sycl::queue q_) : q(q_) {}
     explicit device_policy(sycl::device d_) { q.emplace(d_); }
     operator sycl::queue() const { return queue(); }
@@ -57,7 +73,11 @@ class device_policy
     queue() const
     {
         if (!q)
-            q.emplace();
+        {
+            ::std::scoped_lock lock{mtx};
+            if (!q)
+                q.emplace();
+        }
         return *q;
     }
 
@@ -81,6 +101,7 @@ class device_policy
 
   protected:
     mutable ::std::optional<sycl::queue> q;
+    mutable ::std::mutex mtx;
 };
 
 #if _ONEDPL_FPGA_DEVICE
@@ -104,13 +125,17 @@ class fpga_policy : public device_policy<KernelName>
     queue() const
     {
         if (!this->q)
-            this->q.emplace(
+        {
+            ::std::scoped_lock lock{this->mtx};
+            if (!this->q)
+                this->q.emplace(
 #    if _ONEDPL_FPGA_EMU
-                __dpl_sycl::__fpga_emulator_selector()
+                    __dpl_sycl::__fpga_emulator_selector()
 #    else
-                __dpl_sycl::__fpga_selector()
+                    __dpl_sycl::__fpga_selector()
 #    endif // _ONEDPL_FPGA_EMU
-            );
+                );
+        }
         return *this->q;
     }
 };
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -44,26 +44,46 @@ struct DefaultKernelName;
 template <typename KernelName = DefaultKernelName>
 class device_policy
 {
+    // Needed for the copy constructor that rebinds the kernel name
+    template <typename>
+    friend class device_policy;
+
   public:
     using kernel_name = KernelName;
 
     device_policy() = default;
+
     template <typename OtherName>
-    device_policy(const device_policy<OtherName>& other) : q(other.queue())
+    device_policy(const device_policy<OtherName>& other)
     {
+        ::std::scoped_lock lock{other.mtx};
+        q = other.q;
     }
 
-    device_policy(const device_policy& other) : q(other.queue()) {}
-    device_policy(device_policy&& other) : q(other.queue()) {}
+    device_policy(const device_policy& other)
+    {
+        ::std::scoped_lock lock{other.mtx};
+        q = other.q;
+    }
+
+    device_policy(device_policy&& other)
+    {
+        ::std::scoped_lock lock{other.mtx};
+        q.swap(other.q);
+    }
+
     device_policy&
     operator=(const device_policy& other)
     {
-        q = other.queue();
+        ::std::scoped_lock lock{mtx, other.mtx};
+        q = other.q;
         return *this;
     }
+
     device_policy&
     operator=(device_policy&& other)
     {
+        ::std::scoped_lock lock{mtx, other.mtx};
         q.swap(other.q);
         return *this;
     }
@@ -74,11 +94,10 @@ class device_policy
     sycl::queue
     queue() const
     {
+        ::std::scoped_lock lock{mtx};
         if (!q)
         {
-            ::std::scoped_lock lock{mtx};
-            if (!q)
-                q.emplace();
+            q.emplace();
         }
         return *q;
     }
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -77,8 +79,8 @@ class device_policy
         return ::std::true_type{};
     }
 
-  private:
-    sycl::queue q;
+  protected:
+    mutable ::std::optional<sycl::queue> q;
```
mmichel11, line 83: I think we might need to include the optional header at the top of this file since I do not think it is included anywhere.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
julianmi, line 1: Could we also add a test that fails before this change?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -50,14 +52,34 @@ class device_policy
     device_policy(const device_policy<OtherName>& other) : q(other.queue())
     {
     }
+
+    device_policy(const device_policy& other) : q(other.queue()) {}
+    device_policy(device_policy&& other) : q(other.queue()) {}
+    device_policy&
+    operator=(const device_policy& other)
```
SergeyKopienko, line 59: Looks like not thread-safe ```operator=``` implementation.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -50,14 +52,34 @@ class device_policy
...
+    {
+        q = other.queue();
```
SergeyKopienko, line 61: Should we really need to create ```sycl::queue``` instance here in ```other.queue()``` call ?
May be good enough simple copy ```q``` state?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -50,14 +52,34 @@ class device_policy
...
+        return *this;
+    }
+    device_policy&
+    operator=(device_policy&& other)
+    {
+        q.swap(other.q);
```
SergeyKopienko, line 67: As form it's looks like not thread-safe operation too.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: We might be able to add a test that just includes oneDPL but doesn't call anything, and then invoke it with a device filter that disables all devices (e.g., `ONEAPI_DEVICE_SELECTOR='!*:*'`). In this case, the old code should throw an exception whereas this PR should not.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: Good catch. I will need to change the assignment operators to grab a lock from both mutexes.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: Will change.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: Will change with the other assignment operator.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: It should be sufficient to just copy `q` directly as you suggest.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -50,13 +52,35 @@ class device_policy
...
     explicit device_policy(sycl::queue q_) : q(q_) {}
-    explicit device_policy(sycl::device d_) : q(d_) {}
-    operator sycl::queue() const { return q; }
+    explicit device_policy(sycl::device d_) { q.emplace(d_); }
+    operator sycl::queue() const { return queue(); }
     sycl::queue
     queue() const
     {
-        return q;
+        if (!q)
```
SergeyKopienko, line 77: As form it's looks like not thread-safe operation too.
The problem is in this condition, I think:
- thread (1) make call of this function and because ```if (!q)``` is true -> started to create queue under mutex.
   On last step of this process it's modify ```q``` state
- thread (2) make call of this function - it's check ```q``` state without synchronization (error 1) and may return broken state of ```q``` (error 2) while thread 1 not finished modification of ```q``` state.

adamfidel: I suppose it depends on whether `std::optional<T>::emplace` can leave the optional in a state where it reports that it has a value before the value is actually available.

The alternative would be to move the lock to the outer scope, which would make the `queue()` method always lock. I would like to avoid that if possible.

SergeyKopienko: I think https://en.cppreference.com/w/cpp/utility/optional class is not thread-safe.

akukanov: Looks like Sergey is right and the double check locking pattern (DCLP) is not safe when applied to `std::optional`. Even if the implementation sets the "has a value" flag after assigning the value, it's not enough as a compiler or a processor can reorder the operations. To prevent that, a memory fence is needed, and I doubt that any (not even talking about all) implementations of `optional` does that,
So my original suggestion of DCLP does not work here, and in lieu of enough time let's just always take the lock.

adamfidel: Okay, I changed it to always grab the lock now.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: I have manually confirmed that this test works in principle. I created the following test:
```c++

int main()
{
    return 0;
}
```
Running it as `ONEAPI_DEVICE_SELECTOR='!*:*' ./build/test/header_inclusion_only.pass` crashes on main but does not crash with this branch.

I feel that there is benefit in adding this test, but it will require some special changes to the CMake to modify the environment specifically for this test. It is further complicated by the fact that the CMake files currently already do something similar with the variable `$ONEAPI_DEVICE_SELECTOR`, which is either set to `ONEAPI_DEVICE_SELECTOR=gpu/acc/fpga` or `SYCL_DEVICE_FILTER=gpu/acc/fpga` depending on the compiler version and device type.

The dynamic selection branch also makes changes to this variable. In all, I believe that there is value for adding this test but the risks are high to unintentionally break something with this being so close to the code freeze.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -42,21 +44,62 @@ struct DefaultKernelName;
...
+    {
+        ::std::scoped_lock lock{mtx, other.mtx};
+        q.swap(other.q);
+        return *this;
+    }
+
     explicit device_policy(sycl::queue q_) : q(q_) {}
-    explicit device_policy(sycl::device d_) : q(d_) {}
-    operator sycl::queue() const { return q; }
+    explicit device_policy(sycl::device d_) { q.emplace(d_); }
```
akukanov, line 92: Could you reorder the code so that these existing constructors come before the new copy & move ones, perhaps right after the line 54?

akukanov (commented):
I do not see any serious problems in the current version of the patch. However, the implementations of the policy classes could use some more uglification.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -91,21 +135,29 @@ class fpga_policy : public device_policy<KernelName>
...
-
+    fpga_policy() = default;
     template <unsigned int other_factor, typename OtherName>
     fpga_policy(const fpga_policy<other_factor, OtherName>& other) : base(other.queue()){};
     explicit fpga_policy(sycl::queue q) : base(q) {}
     explicit fpga_policy(sycl::device d) : base(d) {}
+
+    operator sycl::queue() const { return queue(); }
+    sycl::queue
+    queue() const
```
SergeyKopienko, line 146: Am I correct that this function ```sycl::queue queue()``` should be declared as ```virtual``` now in this and base class?
1) Let's assume we have created the ```policy``` instance of ```fpga_policy```
    After this PR this instance doesn't have created ```sycl::queue``` instance inside as it's lazy initialized.
2) In some code we save result of ```static_cast<device_policy&>(policy)``` into ```auto& d_policy```:
```C++
    fpga_policy policy;
    auto& d_policy = static_cast<device_policy&>(policy);
```
3) and after that we call ```queue()``` function from reference to the base class:
```C++
    fpga_policy policy;
    auto& d_policy = static_cast<device_policy&>(policy);
    auto q = d_policy.queue();                   // call device_policy::queue() !?
```
Looks like we will call ```queue()``` method of the base class and creation of ```sycl::queue``` instance will be executed incorrectly.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -77,8 +120,9 @@ class device_policy
         return ::std::true_type{};
     }
 
-  private:
-    sycl::queue q;
+  protected:
+    mutable ::std::optional<sycl::queue> q;
+    mutable ::std::mutex mtx;
```
SergeyKopienko, line 125: From my point of view better to declare synchronization staff at first.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: Or we need some another approach to resolve this issue.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -91,21 +135,29 @@ class fpga_policy : public device_policy<KernelName>
...
-#    else
-              __dpl_sycl::__fpga_selector()
-#    endif // _ONEDPL_FPGA_EMU
-              ))
-    {
-    }
-
+    fpga_policy() = default;
     template <unsigned int other_factor, typename OtherName>
     fpga_policy(const fpga_policy<other_factor, OtherName>& other) : base(other.queue()){};
```
SergeyKopienko, line 140: Lazy initialization concept would be broken at this place: ```queue()``` call will create ```sycl::queue``` instance

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: May be able to apply something like https://github.com/oneapi-src/oneDPL/pull/1158
but this PR doesn't resolve the problem of broken lazy initialization in constructor of
```C++
    template <unsigned int other_factor, typename OtherName>
    fpga_policy(const fpga_policy<other_factor, OtherName>& other) : base(other.queue()){};
```
See my comment https://github.com/oneapi-src/oneDPL/pull/1154/files#r1318324154

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko, line 140: Ending ```;``` no required

SergeyKopienko: May be we should pass ```static_cast<const base&>(other)``` into constructor of base class to avoid this broken case?

SergeyKopienko: Also looks like move constructor of this format and copy/move assignment constructors required in this class too.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -42,21 +44,62 @@ struct DefaultKernelName;
...
+    device_policy(const device_policy& other)
+    {
+        ::std::scoped_lock lock{other.mtx};
+        q = other.q;
+    }
+
+    device_policy(device_policy&& other)
     {
+        ::std::scoped_lock lock{other.mtx};
+        q.swap(other.q);
```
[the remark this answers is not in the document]
SergeyKopienko: Also header file for ```std::swap``` not included.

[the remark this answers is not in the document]
rarutyun: I also don't understand why we do use `swap` member function for move and don't use it for copy. I feel like reusing copy/move-assignment from `optional` is the right way to go. But if you have arguments why you did `swap` please clarify.

My other question is can we make _member-init-list_ there instead of default constructing and then assigning? For example, a`static` function that that locks and returns optional with the correct value category?

> Also header file for std::swap not included.

That's something I don't understand at all. we don't use std::swap (at least in this line). For member function including `<optional>` is enough

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -42,21 +44,62 @@ struct DefaultKernelName;
...
     }
+
+    device_policy&
+    operator=(const device_policy& other)
+    {
+        ::std::scoped_lock lock{mtx, other.mtx};
```
rarutyun, line 78: I believe it's not done. What if it's self assignment? How does `scoped_lock` behave for the same mutex passed twice? My bet it's a deadlock (better to say, lifelock)

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
rarutyun: no

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -42,21 +44,62 @@ struct DefaultKernelName;
...
     using kernel_name = KernelName;
 
     device_policy() = default;
+
     template <typename OtherName>
-    device_policy(const device_policy<OtherName>& other) : q(other.queue())
+    device_policy(const device_policy<OtherName>& other)
+    {
+        ::std::scoped_lock lock{other.mtx};
+        q = other.q;
```
rarutyun, line 60: I am wondering what is the correct design? should we call `other.queue()` or `other.q`? This is relevant for all places. I understand the difference in the behavior and I am trying to understand what is the right one.

rarutyun:
Just to state it explicitly. Those are not all comments that I have. I am thinking about Sergey's comment about virtuality and also about additional things for copy/move semantics

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
rarutyun: > Lazy initialization concept would be broken at this place: `queue()` call will create `sycl::queue` instance

That's exactly the question what behavior we want for non-default constructors and assignments. I think I want them being initialized (I can change my mind, though), but I also can see the issues with that since we have inheritance

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
rarutyun: I feel like that:
- If we do that, this is not the design intent we want
- If we don't do that - this is the API breaking change

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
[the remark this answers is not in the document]
SergeyKopienko: I mean if we take a loot into https://en.cppreference.com/w/cpp/algorithm/swap,
there are 3 headers in which ```std::swap``` declared:
```
Defined in header [<algorithm>](https://en.cppreference.com/w/cpp/header/algorithm) (until C++11)
Defined in header [<utility>](https://en.cppreference.com/w/cpp/header/utility) (since C++11)
Defined in header [<string_view>](https://en.cppreference.com/w/cpp/header/string_view) (since C++17)
```
And when I see introduced ```std::swap``` call I'm waiting for the inclusion of one from these headers.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: Yes, @rarutyun, you are right. Agreed.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: I modified the code to now check for self assignment.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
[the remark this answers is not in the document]
adamfidel: I switched the swap with the move constructor and rewrote the copy/move constructor as suggested by @rarutyun.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
adamfidel: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -42,21 +44,66 @@ struct DefaultKernelName;
 template <typename KernelName = DefaultKernelName>
 class device_policy
 {
+    // Needed for the copy constructor that rebinds the kernel name
+    template <typename>
+    friend class device_policy;
+
+    template <typename T>
+    static auto
+    lock_and_forward(T&& t, std::mutex& mtx)
```
SergeyKopienko, line 53: From my point of view this pattern looks like not quite correct.
For example let's analyze the situation when we create new object ```device_policy``` by ```const device_policy&``` from another copy.
What's happen when we call ```lock_and_forward``` :
- we pass ```const device_policy&``` and mutex into this function;
- lock mutex;
- return our argument ```const device_policy&``` as result;
- unlock mutex on exit.
What we made safe by this mutex? Looks like we made safe only return of ```const device_policy&```.
I think it wasn't our goal.

SergeyKopienko: I think this pattern will work if we return simple copy of ```t```.
May be the idea was in that?

SergeyKopienko: Discussed offline that because we return ```auto``` this mean we return copy - agree, ok.

SergeyKopienko: From "Working Draft, Standard for Programming Language C++"

```
9.6.3 The return statement
.....
3 The copy-initialization of the result of the call is sequenced before the destruction of temporaries at the end of the full-expression established by the operand of the return statement, which, in turn, is sequenced before the destruction of local variables (9.6) of the block enclosing the return statement.
```

So looks like this solution with ```lock_and_forward``` static function is OK.

SergeyKopienko: @adamfidel, @rarutyun I think now we can resolve this conversation?

akukanov:
After merging #1652, this patch is outdated.

SergeyKopienko:
> After merging #1652, this patch is outdated.

Why do you think so?

PR #1159, opened 2023-09-07
[KT] Uglification

dmitriy-sobolev:
Also, `class simd2d` was removed because it is not used.


SergeyKopienko (approved):
LGTM
But I think required to rerun our tests.

dmitriy-sobolev:
> LGTM But I think required to rerun our tests.

All the tests pass

PR #1160, opened 2023-09-07
[KT] Fix an error in the async free memory in esimd radix sort

SergeyKopienko:
Introduced in https://github.com/oneapi-src/oneDPL/pull/1057


dmitriy-sobolev (approved):
LGTM.

As we discussed offline, capturing by copy is needed to make sure the objects are alive upon exiting from the onesweep function. 

Probably the same fix is needed in other submissions in the onesweep function, but I agree to merge it right now.

PR #1162, opened 2023-09-07
Improve support for device selection incompatible with offload policy

Alexandr-Konovalov:
Previously, if ONEAPI_DEVICE_SELECTOR is incompatible with a selected offload policy, an exception is thrown from static ctor from header code included in user's code. Such exception is hard to process. After the fix the exception is catch and re-thrown during usage of the  execution policy. In the covered situation memory allocation is done via system allocation, as if no offload occurs.


kboyarinov (dismissed):
LGTM

kboyarinov (approved):

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +53,58 @@ __get_offload_device()
...
     using __set_active_device_func_type = void (*)(sycl::device*);
+    using __get_offload_device_func_type = sycl::device (*)();
 
   public:
     // Since the global object of __offload_policy_holder_type is static but the constructor
     // of the class is inline, we need to avoid calling static functions inside of the constructor
     // and pass the pointer to exact function as an argument to guarantee that the correct __active_device
     // would be stored in each translation unit
-    __offload_policy_holder_type(sycl::device&& __offload_device,
+    __offload_policy_holder_type(__get_offload_device_func_type __get_offload_device_func,
```
kboyarinov, line 63: May be it would be better to change `get_offload_device` to return device selector and pass the selector into the constructor?

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +53,58 @@ __get_offload_device()
...
+        try
+        {
+            _M_offload_device = __get_offload_device_func();
+        }
+        catch (const sycl::exception& e)
+        {
+            // e.code() == sycl::errc::runtime means occurs when unable to get offload device. Do not pass
+            // an exception, as ctor is called for a static object and the exception can't be processed.
+            // Remember the situation and re-throw exception when asked for the policy from user's code.
+            if (e.code() != sycl::errc::runtime)
```
kboyarinov, line 76: What kind of other exceptions can be thrown when constructing the device? 
If we are re-throwing all exceptions other then `sycl::errc::runtime`, it seems like there would be an other exception thrown by the constructor of the global object and hence it cannot be catched by the user.

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +53,58 @@ __get_offload_device()
...
+                throw;
+        }
+
+        if (_M_offload_device.has_value())
+        {
+            _M_set_active_device(&*_M_offload_device);
+            oneapi::dpl::execution::device_policy<> _offload_policy(*_M_offload_device);
+            _M_offload_policy = _offload_policy;
```
kboyarinov, line 84:
```suggestion
            _M_offload_policy = oneapi::dpl::execition::device_policy<>{*_M_offload_device};
```

test/CMakeLists.txt
```diff
@@ -110,6 +110,10 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
 
     set_tests_properties(${_test_name} PROPERTIES SKIP_RETURN_CODE 77)
 
+    if ("${_test_name}" STREQUAL selected_different_device.pass AND _ONEDPL_PSTL_OFFLOAD MATCHES "(cpu|gpu)")
+        set(DEVICE_SELECTION_LINE "${DEVICE_SELECTION_VARIABLE}=${ONEDPL_DEVICE_BACKEND}:${_ONEDPL_PSTL_OFFLOAD}")
+    endif()
```
kboyarinov, line 115: I don't completely understand what these lines are intended for. Could you please explain?

test/pstl_offload/general/selected_different_device.pass.cpp
```diff
@@ -0,0 +1,42 @@
...
+#error "PSTL offload is not enabled or the selected value is unsupported"
+#endif
+
+#include <algorithm>
+#include "support/utils.h"
+
+// The test assumes that ONEAPI_DEVICE_SELECTOR is set according to __SYCL_PSTL_OFFLOAD__
+// and check that changed __SYCL_PSTL_OFFLOAD__ leads to an exception exactly from parallel algorithm.
+int main() {
+#if __SYCL_PSTL_OFFLOAD__ == 1
```
kboyarinov, line 27: It seems like it would be more correct to return `TestUtils::done(1)`  here. This call will print "Skipped" and return the skip code.

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +53,58 @@ __get_offload_device()
...
     auto
     __get_policy()
     {
+        if (!_M_offload_device.has_value())
+            throw sycl::exception(sycl::errc::runtime);
         return _M_offload_policy;
     }
   private:
-    sycl::device _M_offload_device;
+    std::optional<sycl::device> _M_offload_device;
```
kboyarinov, line 102: Don't we need to `#include <optional>` in this header file?

CMakeLists.txt
```diff
@@ -207,15 +216,16 @@ if (ONEDPL_BACKEND MATCHES "^(tbb|dpcpp|dpcpp_only)$")
...
         endif()
 
+        if (INTEL_LLVM_COMPILER AND INTEL_LLVM_COMPILER_VERSION VERSION_LESS 2023.1)
+            set(DEVICE_SELECTION_VARIABLE "SYCL_DEVICE_FILTER")
+            set(FPGA_DEVICE_TYPE "acc")
+        else()
+            set(DEVICE_SELECTION_VARIABLE "ONEAPI_DEVICE_SELECTOR")
+            set(FPGA_DEVICE_TYPE "fpga")
+        endif()
+
```
kboyarinov, line 226: Why this logic should be moved outside the `if`. It seems like it would define `DEVICE_SELECTION_VARIABLE` for all backends now even if `ONEDPL_DEVICE_TYPE` is not defined? Is it correct?

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: > If we are re-throwing all exceptions other then sycl::errc::runtime, it seems like there would be an other exception thrown by the constructor of the global object and hence it cannot be catched by the user.

Exactly, and that's the plan. We are not going to mitigate situation "something bad occurs, let's hide it for a while", but only cover the expected situation.

Possible alternative is `std::exception_ptr`, `std::rethrow_exception()` etc. I don't like it, because it's non-trivial processing of something unknown.

Sure, this might be discussed.

CMakeLists.txt
Alexandr-Konovalov: Yes, exactly. I'll need `DEVICE_SELECTION_VARIABLE` later.

test/CMakeLists.txt
Alexandr-Konovalov: I'll add a comment, but generally there are 2 options.

1. Set ONEAPI_DEVICE_SELECTOR always according to PSTL offload policy.
2. Set it only when required.

I'm slightly for (2), as it more aligned with possible working mode. It should be noted that (1) leads to more clear cmake code, but this is not a goal.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Done.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Done.

test/pstl_offload/general/selected_different_device.pass.cpp
Alexandr-Konovalov: Done.

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +55,57 @@ __get_offload_device()
...
     using __set_active_device_func_type = void (*)(sycl::device*);
+    using __get_offload_device_func_type = sycl::device (*)();
 
   public:
     // Since the global object of __offload_policy_holder_type is static but the constructor
     // of the class is inline, we need to avoid calling static functions inside of the constructor
     // and pass the pointer to exact function as an argument to guarantee that the correct __active_device
     // would be stored in each translation unit
-    __offload_policy_holder_type(sycl::device&& __offload_device,
+    __offload_policy_holder_type(__device_selector_type __device_selector,
```
kboyarinov, line 65: I guess we can simply make this constructor template (in the same manner as a `sycl::device` constructor):

```suggestion
    template <typename _DeviceSelector>
    __offload_policy_holder_type(const _DeviceSelector& __device_selector,
```

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +55,57 @@ __get_offload_device()
...
                                  __set_active_device_func_type __set_active_device_func)
-        : _M_offload_device(std::move(__offload_device)), _M_offload_policy(_M_offload_device),
-          _M_set_active_device(__set_active_device_func)
+        : _M_set_active_device(__set_active_device_func)
     {
-        _M_set_active_device(&_M_offload_device);
+        try
+        {
+            _M_offload_device = sycl::device{__device_selector};
```
kboyarinov, line 71:
```suggestion
            _M_offload_device.emplace(__device_selector);
```

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -36,15 +36,17 @@ __set_active_device(sycl::device* __new_active_device)
     __active_device.store(__new_active_device, std::memory_order_release);
 }
 
-static sycl::device
-__get_offload_device()
+using __device_selector_type = int (*)(const sycl::device &);
+
+static __device_selector_type
```
kboyarinov, line 41:
```suggestion
static auto
```

If the constructor would be `template` - `__device_selector_type` can be removed

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +55,57 @@ __get_offload_device()
...
+            // e.code() == sycl::errc::runtime means occurs when unable to get offload device. Do not pass
+            // an exception, as ctor is called for a static object and the exception can't be processed.
+            // Remember the situation and re-throw exception when asked for the policy from user's code.
+            if (e.code() != sycl::errc::runtime)
+                throw;
+        }
+
+        if (_M_offload_device.has_value())
+        {
+            _M_set_active_device(&*_M_offload_device);
```
kboyarinov, line 84: Having `set_active_device` only if the device was constructed will cause the memory replacement function to always use system malloc and the user will identify the error only when the first parallel algorithm would be called. 
I guess it makes sense to store some special value into the active device in the `else` branch here and throw an exception from the allocating functions as well.

include/pstl_offload/internal/usm_memory_replacement.h
kboyarinov: It is not guaranteed that sycl includes `<optional>`, isn't it? I think it would be more clear to include it explicitly.

Alexandr-Konovalov: Ahh, to support usage of this version of oneDPL with some future compiler, that is possible while PSTL offload is a part of compiler. Got it.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Could you please clarify the reason a little? I mean, we know the exact type, we are going to use only this exact type, but not want to mention it because...

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Interesting point. As you may see from the test, I definitely want to support seamless regression to a non-PSTL offload implementation. I.e., if TSLT offload is unavailable, a user may call original `par`/`seq` algorithm (better yet, non-offload `par_unseq`). But what should she/he use after a `malloc` throws an exception? Call another `malloc` in another TU?

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Done.

include/pstl_offload/internal/usm_memory_replacement.h
kboyarinov: As far as I understand from the spec, device selector is not guaranteed to be pointer to function, the only thing we know about it - it satisfies the requirements of `Callable` as described [here](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-selection:~:text=The%20interface%20for%20a%20device%20selector%20is%20any%20object%20that%20meets%20the%20C%2B%2B%20named%20requirement%20Callable%2C%20taking%20a%20parameter%20of%20type%20const%20device%20%26%20and%20returning%20a%20value%20that%20is%20implicitly%20convertible%20to%20int.). It covers also other invocable objects (lambdas, pointers to member functions/objects, etc.)

I would prefer to be consistent with SYCL classes (device, queue) as take/pass the selector argument in the same manner.

Alexandr-Konovalov: > As far as I understand from the spec, device selector is not guaranteed to be pointer to function, the only thing we know about it - it satisfies the requirements of Callable as described [here](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:device-selection:~:text=The%20interface%20for%20a%20device%20selector%20is%20any%20object%20that%20meets%20the%20C%2B%2B%20named%20requirement%20Callable%2C%20taking%20a%20parameter%20of%20type%20const%20device%20%26%20and%20returning%20a%20value%20that%20is%20implicitly%20convertible%20to%20int.).

Thank you, missed that.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Done.

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +54,59 @@ __get_offload_device()
...
   public:
     // Since the global object of __offload_policy_holder_type is static but the constructor
     // of the class is inline, we need to avoid calling static functions inside of the constructor
     // and pass the pointer to exact function as an argument to guarantee that the correct __active_device
     // would be stored in each translation unit
-    __offload_policy_holder_type(sycl::device&& __offload_device,
+    template <typename _DeviceSelector>
+    __offload_policy_holder_type(const _DeviceSelector& __device_selector,
                                  __set_active_device_func_type __set_active_device_func)
-        : _M_offload_device(std::move(__offload_device)), _M_offload_policy(_M_offload_device),
```
kboyarinov, line 64: It seems like we need to also make `_M_offload_policy` std::optional. Because otherwise, the default constructor of the `device_policy` would be called that will create unnecessary `sycl::queue` with the default selector inside.

include/pstl_offload/internal/usm_memory_replacement.h
kboyarinov: As discussed offline, one of the main purposes to have "delayed" throw here is to allow user to fallback into seq/par/unseq version of the algorithm in case of exception from the parallel algorithm. In this case, having an error handling inside allocating functions is not needed because we want allocating functions to provide memory that can be used in seq/par/unseq algorithm calls.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Good point, thank you.

CMakeLists.txt
```diff
@@ -184,14 +192,15 @@ if (ONEDPL_BACKEND MATCHES "^(tbb|dpcpp|dpcpp_only)$")
             set(USE_FSYCL_OPTION TRUE)
         endif()
 
-        # check device type
+        # check device type for oneDPL test targets
         if (ONEDPL_DEVICE_TYPE MATCHES "^(CPU|GPU|FPGA_EMU|FPGA_HW)$")
```
danhoeflinger, line 196: Does it make sense to warn or throw an error if this `ONEDPL_DEVICE_TYPE` is specified at the same time as `_ONEDPL_PSTL_OFFLOAD`?

CMakeLists.txt
```diff
@@ -207,15 +216,16 @@ if (ONEDPL_BACKEND MATCHES "^(tbb|dpcpp|dpcpp_only)$")
...
+        if (INTEL_LLVM_COMPILER AND INTEL_LLVM_COMPILER_VERSION VERSION_LESS 2023.1)
+            set(DEVICE_SELECTION_VARIABLE "SYCL_DEVICE_FILTER")
+            set(FPGA_DEVICE_TYPE "acc")
+        else()
+            set(DEVICE_SELECTION_VARIABLE "ONEAPI_DEVICE_SELECTOR")
+            set(FPGA_DEVICE_TYPE "fpga")
+        endif()
+
         # set up device selection line basing on ONEDPL_DEVICE_BACKEND and ONEDPL_DEVICE_TYPE values
         if (DEFINED ONEDPL_DEVICE_BACKEND AND DEFINED ONEDPL_DEVICE_TYPE)
```
danhoeflinger, line 228: Following up on above, will this code conflict with PSTL offload device selection?  (and should we prevent this in some way?)

CMakeLists.txt
Alexandr-Konovalov: I added a check against mutual set of `_ONEDPL_PSTL_OFFLOAD` and `ONEDPL_DEVICE_TYPE`.

CMakeLists.txt
Alexandr-Konovalov: Thank you, that was a bug in `DEVICE_SELECTION_VARIABLE` usage. I added a check on `onedpl_add_test()` level.

CMakeLists.txt
danhoeflinger: Looks good.

CMakeLists.txt
danhoeflinger: Looks good.

danhoeflinger:
CMake part LGTM, I have not reviewed the full PR

CMakeLists.txt
```diff
@@ -147,6 +147,14 @@ else()
     message(STATUS "oneDPL: ONEDPL_ENABLE_SIMD is OFF, corresponding compiler flag is not used")
 endif()
 
+if (NOT _ONEDPL_PSTL_OFFLOAD)
+    set(_ONEDPL_PSTL_OFFLOAD "off")
+else()
+    if (NOT ONEDPL_BACKEND MATCHES "^(dpcpp)$")
```
dmitriy-sobolev, line 153: Could you also add a check for `dpcpp_only`?

Alexandr-Konovalov: Done.

dmitriy-sobolev (commented):
CMake changes look good enough to me as a short-term fix.

However, I suggest considering these questions for future cmake-only PRs for PSTL offload feature:
 - Can `_ONEDPL_PSTL_OFFLOAD` use `ONEDPL_DEVICE_TYPE`? 
 - Can `_ONEDPL_PSTL_OFFLOAD` be replaced with `ONEDPL_TEST_PSTL_OFFLOAD` (`ONEDPL_TEST*` prefix is used for options affecting only tests) and be documented e.g. like `ONEDPL_TEST_EXPLICIT_KERNEL_NAMES` from [Readme](https://github.com/oneapi-src/oneDPL/blob/main/cmake/README.md)?

Alexandr-Konovalov:
> However, I suggest considering these questions for future cmake-only PRs for PSTL offload feature:
> 
> * Can `_ONEDPL_PSTL_OFFLOAD` use `ONEDPL_DEVICE_TYPE`?

`_ONEDPL_PSTL_OFFLOAD` servers 2 proposes: it enables build of `libpstloffload.so` and sets target offload device during compile time. `ONEDPL_DEVICE_TYPE` basically sets `ONEAPI_DEVICE_SELECTOR` environment variable. The `ONEAPI_DEVICE_SELECTOR` is not needed during PSTL offload testing, because offload device is set during compile time (exception is one case, when we test that incompatible values of them are processed correctly).

So, I don't see how to unify them.

> * Can `_ONEDPL_PSTL_OFFLOAD` be replaced with `ONEDPL_TEST_PSTL_OFFLOAD` (`ONEDPL_TEST*` prefix is used for options affecting only tests) and be documented e.g. like `ONEDPL_TEST_EXPLICIT_KERNEL_NAMES` from [Readme](https://github.com/oneapi-src/oneDPL/blob/main/cmake/README.md)?

The underscore in `_ONEDPL_PSTL_OFFLOAD` was added because this is not a product-quality feature yet. And one was not documented by same reason. Hope this can be changed sooner rather than later.

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +54,59 @@ __get_offload_device()
 class __offload_policy_holder_type
 {
     using __set_active_device_func_type = void (*)(sycl::device*);
+    using __get_offload_device_func_type = sycl::device (*)();
```
rarutyun, line 57: Looks unused

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +54,59 @@ __get_offload_device()
...
+            // the exception can't be processed.
+            // Remember the situation and re-throw exception when asked for the policy from user's code.
+            if (e.code() != sycl::errc::runtime)
+                throw;
+        }
+
+        if (_M_offload_device.has_value())
+        {
+            _M_set_active_device(&*_M_offload_device);
+            _M_offload_policy.emplace(*_M_offload_device);
```
rarutyun, line 86: Can we move those two lines in `try`-block and without if?

I guess, we can. Am I missing something?

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -53,34 +54,59 @@ __get_offload_device()
...
+            // with required type. Do not pass an exception, as ctor is called for a static object and
+            // the exception can't be processed.
+            // Remember the situation and re-throw exception when asked for the policy from user's code.
+            if (e.code() != sycl::errc::runtime)
+                throw;
+        }
+
+        if (_M_offload_device.has_value())
+        {
+            _M_set_active_device(&*_M_offload_device);
```
rarutyun, line 85:
```suggestion
            _M_set_active_device(&_M_offload_device.value());
```
How about that? To me it's just simpler to read comparing with `&*`

test/pstl_offload/general/selected_different_device.pass.cpp
```diff
@@ -0,0 +1,43 @@
+// -*- C++ -*-
+//===----------------------------------------------------------------------===//
+//
+// Copyright (C) 2023 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#if __SYCL_PSTL_OFFLOAD__ == 1
```
rarutyun, line 10: Can we enabling skipping on the CMake level? I guess we are trying to do that. Did I misunderstand?

test/pstl_offload/general/selected_different_device.pass.cpp
```diff
@@ -0,0 +1,43 @@
...
+    return TestUtils::done(0); // to get Skipped
+#else
+    size_t num = 10LLU*1024*1024;
+    int* ptr = new int[num];
+    try {
+        std::fill(std::execution::par_unseq, ptr, ptr + num, 7);
+        EXPECT_TRUE(false, "Device must be unavailable, expecting exception from std::fill().");
+    } catch (const sycl::exception &) {
+    }
+    delete []ptr;
```
rarutyun, line 39:
```suggestion
    delete[] ptr;
```
Doesn't have to be applied

test/pstl_offload/general/selected_different_device.pass.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+// The test assumes that ONEAPI_DEVICE_SELECTOR is set according to __SYCL_PSTL_OFFLOAD__
+// and check that changed __SYCL_PSTL_OFFLOAD__ leads to an exception exactly from parallel algorithm.
+int main() {
+#if __SYCL_PSTL_OFFLOAD__ == 1
+    // skip test when value of _ONEDPL_PSTL_OFFLOAD is default
+    // as ONEAPI_DEVICE_SELECTOR allows any offload policy in this case
+    return TestUtils::done(0); // to get Skipped
+#else
+    size_t num = 10LLU*1024*1024;
```
rarutyun, line 32:
```suggestion
    std::size_t num = 10LLU * 1024 * 1024;
```

test/pstl_offload/general/selected_different_device.pass.cpp
Alexandr-Konovalov: It's definitely possible but requires ugly changes in cmake, while skipping on the test level is straightforward, so I'm for the current implementation.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Done.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Done.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Done.

test/pstl_offload/general/selected_different_device.pass.cpp
Alexandr-Konovalov: Done.

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -59,28 +60,49 @@ class __offload_policy_holder_type
...
                                  __set_active_device_func_type __set_active_device_func)
-        : _M_offload_device(std::move(__offload_device)), _M_offload_policy(_M_offload_device),
-          _M_set_active_device(__set_active_device_func)
+        : _M_set_active_device(__set_active_device_func)
     {
-        _M_set_active_device(&_M_offload_device);
+        try
+        {
+            _M_offload_device.emplace(__device_selector);
+            _M_set_active_device(&_M_offload_device.value());
```
rarutyun, line 71:
```suggestion
            _M_set_active_device(&*_M_offload_device);
```
I apologize. I changed my mind because I found out the technical reason, why `*` is better. It's unchecked version and never throws. I didn't know it previously.

test/pstl_offload/general/selected_different_device.pass.cpp
```diff
@@ -0,0 +1,43 @@
...
+//===----------------------------------------------------------------------===//
+
+#if __SYCL_PSTL_OFFLOAD__ == 1
+#elif __SYCL_PSTL_OFFLOAD__ == 2
+    #undef __SYCL_PSTL_OFFLOAD__
+    #define __SYCL_PSTL_OFFLOAD__ 3
+#elif __SYCL_PSTL_OFFLOAD__ == 3
+    #undef __SYCL_PSTL_OFFLOAD__
+    #define __SYCL_PSTL_OFFLOAD__ 2
+#else
```
rarutyun, line 17: Based on discussion above wow about this?

```suggestion
#if __SYCL_PSTL_OFFLOAD__ == 1
#elif __SYCL_PSTL_OFFLOAD__ == 2
    #undef __SYCL_PSTL_OFFLOAD__
    #define __SYCL_PSTL_OFFLOAD__ 3
#elif __SYCL_PSTL_OFFLOAD__ == 3
    #undef __SYCL_PSTL_OFFLOAD__
    #define __SYCL_PSTL_OFFLOAD__ 2
#elif __SYCL_PSTL_OFFLOAD != 1
```
Is it better than empty if or not? @kboyarinov, @Alexandr-Konovalov, opinions?

test/pstl_offload/general/selected_different_device.pass.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+// The test assumes that ONEAPI_DEVICE_SELECTOR is set according to __SYCL_PSTL_OFFLOAD__
+// and check that changed __SYCL_PSTL_OFFLOAD__ leads to an exception exactly from parallel algorithm.
+int main() {
+#if __SYCL_PSTL_OFFLOAD__ == 1
+    // skip test when value of _ONEDPL_PSTL_OFFLOAD is default
+    // as ONEAPI_DEVICE_SELECTOR allows any offload policy in this case
+    return TestUtils::done(0); // to get Skipped
+#else
+    size_t num = 10LLU * 1024 * 1024;
```
rarutyun, line 32:
```suggestion
    std::size_t num = 10LLU * 1024 * 1024;
```

test/pstl_offload/general/selected_different_device.pass.cpp
Alexandr-Konovalov: I'm for `else` and for keeping all cases enumerated explicitly, as if someday we'll add yet another value for `__SYCL_PSTL_OFFLOAD__`, the test failed with clear diagnostics.

Alexandr-Konovalov: Both is working for me, I'm abstaining.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Done.

test/pstl_offload/general/selected_different_device.pass.cpp
Alexandr-Konovalov: Sorry, missed.

test/pstl_offload/general/selected_different_device.pass.cpp
kboyarinov: I would prefer the suggested option to avoid empty branches.
From the functional perspective, seems equivalent.

Alexandr-Konovalov: Sure, let's change it.

rarutyun:
LGTM, but I guess we should fix the PSTL offload step in CI before merging that. Let's try to make it in this PR. After it's done feel free to approve and merge.

kboyarinov (approved):
LGTM

PR #1167, opened 2023-09-08
Updating RNG tests to fix warning of deprecated get_pointer interface.

timmiesmith:
#1094 updated oneDPL headers to remove usage of the get_pointer methods for buffer and accessors. This PR updates the random number generator tests to remove use of the accessor get_pointer method.

### DIFF
--- a/test/xpu_api/random/device_tests/common_for_device_tests.h
+++ b/test/xpu_api/random/device_tests/common_for_device_tests.h
@@ -87,7 +87,7 @@ int device_copyable_test(sycl::queue& queue) {
                 Distr device_distr(distr);
                 device_engine.discard(offset);
                 result_type res = device_distr(device_engine);
-                res.store(idx.get_linear_id(), acc.get_pointer());
+                res.store(idx.get_linear_id(), __dpl_sycl::__get_accessor_ptr(acc));
             });
         });
     }
--- a/test/xpu_api/random/statistics_tests/bernoulli_distribution_test.pass.cpp
+++ b/test/xpu_api/random/statistics_tests/bernoulli_distribution_test.pass.cpp
@@ -69,7 +69,7 @@ test(sycl::queue& queue, double p, int nsamples)
                 oneapi::dpl::bernoulli_distribution<BoolType> distr(p);
 
                 sycl::vec<bool, num_elems> res = distr(engine);
-                res.store(idx.get_linear_id(), acc.get_pointer());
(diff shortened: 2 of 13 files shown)


mmichel11 (approved):
The changes LGTM. I do not see any other uses of `get_pointer` in device code.

test/xpu_api/random/device_tests/common_for_device_tests.h
```diff
@@ -87,7 +87,7 @@ int device_copyable_test(sycl::queue& queue) {
                 Distr device_distr(distr);
                 device_engine.discard(offset);
                 result_type res = device_distr(device_engine);
-                res.store(idx.get_linear_id(), acc.get_pointer());
+                res.store(idx.get_linear_id(), __dpl_sycl::__get_accessor_ptr(acc));
```
dmitriy-sobolev, line 90: Using a wrapper from internal API might be a descent solution in short-term. But I would recommend duplicating `__get_accessor_ptr` method in `test/utils_sycl.hpp`.

test/xpu_api/random/statistics_tests/bernoulli_distribution_test.pass.cpp
```diff
@@ -116,7 +116,7 @@ test_portion(sycl::queue& queue, double p, int nsamples, unsigned int part)
 
                 sycl::vec<bool, num_elems> res = distr(engine, part);
                 for (int i = 0; i < n_elems; ++i)
-                    acc.get_pointer()[idx.get_linear_id() * n_elems + i] = res[i];
+                    __dpl_sycl::__get_accessor_ptr(acc)[idx.get_linear_id() * n_elems + i] = res[i];
```
dmitriy-sobolev, line 119: Can we replace it with:
```suggestion
                    acc[idx.get_linear_id() * n_elems + i] = res[i];
```

dmitriy-sobolev:
@paveldyakov, could you take a look at this PR because it tackles RNG tests?

paveldyakov (approved):
Looks good to me, thank you for creating these changes!
Alina will also take a look.

aelizaro (approved):
Looks good, thank you!

test/xpu_api/random/device_tests/common_for_device_tests.h
akukanov: Can we do it properly now? Relying on any implementation detail in tests is bad. We should reduce that. not add new cases.

PR #1169, opened 2023-09-08
[Examples] Fix warning: 'get_pointer' is deprecated

dmitriy-sobolev:
More details about specification and compiler changes which led to this PR are in #1094 description.

### DIFF
--- a/examples/random/src/main.cpp
+++ b/examples/random/src/main.cpp
@@ -95,7 +95,7 @@ vector_example(sycl::queue& queue, std::uint32_t seed, std::vector<RealType>& x)
                         auto res = distr(engine);
 
                         // Store results from res to VecSize * offset position of x_acc
-                        res.store(idx.get_linear_id(), x_acc.get_pointer());
+                        res.store(idx.get_linear_id(), x_acc);
                     });
             });
     }


examples/random/src/main.cpp
```diff
@@ -95,7 +95,7 @@ vector_example(sycl::queue& queue, std::uint32_t seed, std::vector<RealType>& x)
                         auto res = distr(engine);
 
                         // Store results from res to VecSize * offset position of x_acc
-                        res.store(idx.get_linear_id(), x_acc.get_pointer());
+                        res.store(idx.get_linear_id(), x_acc);
```
dmitriy-sobolev, line 98: A more straightforward solution would be to pass `x_acc.template get_multi_ptr<sycl::access::decorate::yes>()`. But `multi_ptr` can be constructed using a buffer accessor, so we can pass the accessor, which looks simpler.

akukanov (approved):

dmitriy-sobolev:
@aelizaro, would you mind if I merge this PR?

aelizaro (approved):
@dmitriy-sobolev, sure, looks cleaner

PR #1170, opened 2023-09-09
[KT] Align names with the rest oneDPL code

dmitriy-sobolev:
oneDPL code uses snake_case for variables (even for constexpr ones), function names; PascalCase for types. 
The names are preserved for ESIMD function wrappers to be aligned with ESIMD.


SergeyKopienko (approved):
LGTM

PR #1171, opened 2023-09-11
[KT] Remove namespace injections

dmitriy-sobolev:


dmitriy-sobolev:
@SergeyKopienko, all the tests have passed

PR #1176, opened 2023-09-12
Fix clang-format issue in nanorange

dmitriy-sobolev:
I guess, this issue was introduced at some early stage of adaptation of `nanorange.cpp`. It is absent in the original repository.


danhoeflinger (approved):
LGTM

PR #1178, opened 2023-09-13
Skipping old compilers

egfefey:
This skips dynamic selection tests on oneAPI compilers before 2023.0.
(Some context: For the sycl backend, dynamic selection uses host tasks that is not available on older compilers)


SergeyKopienko (approved):
LGTM

PR #1182, opened 2023-09-14
Extend usage of features supported by Intel llvm-based compiler

dmitriy-sobolev:
### Summary
Intel® oneAPI DPC++/C++ Compiler supports some features which are supported by Intel® C++ Compiler Classic. This PR enables them in oneDPL. These are currently indicated by `__INTEL_COMPILER` macro.

### Details
__INTEL_COMPILER   occurance | Feature | __INTEL_LLVM_COMPILER support | Comment
-- | -- | -- | --
[glue_execution_defs.h#L28](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/glue_execution_defs.h#L28) | - |   | Looks like some legacy code. Clean up (probably in a separate PR).
[onedpl_config.h#L104](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L104) | omp simd | + | DPC++ fully supports OpenMP 4.5 since 2021.1: https://www.intel.com/content/www/us/en/developer/articles/technical/openmp-features-and-extensions-supported-in-icx.html
[onedpl_config.h#L105](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L105) | omp declare simd | + | DPC++ fully supports OpenMP 4.5 since 2021.1
[onedpl_config.h#L106](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L106) | omp simd reduction(X) | + | DPC++ fully supports OpenMP 4.5 since 2021.1
[onedpl_config.h#L118](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L118) | unroll | + |  The pragma is covered in https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-2/unroll-nounroll.html
[onedpl_config.h#L125](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L125) | forceinline | + |  The pragma is covered in https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-2/inline-noinline-forceinline.html#GUID-0ECDCF52-2977-4857-9A74-2472E634FD88
[onedpl_config.h#L133](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L133) | omp simd reduction(inscan, X) | since 2023.1 | It depends on `omp scan inclusive` or `omp scan exclusive`: "A list item that appears in the inclusive or exclusive clause must appear in a reduction clause with the inscan modifier on the enclosing worksharing-loop, worksharing-loop SIMD, or simd construct." (see https://www.openmp.org/spec-html/5.0/openmpsu45.html).
[onedpl_config.h#L134](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L134) | omp scan inclusive(X) | since 2023.1 | Check "scan" feature in   https://www.intel.com/content/www/us/en/developer/articles/technical/openmp-features-and-extensions-supported-in-icx.html
[onedpl_config.h#L135](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L135) | omp scan exclusive(X) | since 2023.1 | Check "scan" feature in   https://www.intel.com/content/www/us/en/developer/articles/technical/openmp-features-and-extensions-supported-in-icx.html
[onedpl_config.h#L165](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L165) | omp simd early_exit | - | specific to ICC: it is in   https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-10/intel-specific-pragma-reference.html,   but not in   https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-2/overview.html
[onedpl_config.h#L176](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L176) | omp ordered simd monotonic(X) | - | Absent in OpenMP spec; Looks specific to ICC;  The latest DPC++ shows warning   "warning: extra tokens at the end of '#pragma omp ordered' are   ignored"
[onedpl_config.h#L177](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L177) | omp ordered simd monotonic(X, Y) | - | Absent in OpenMP spec; Looks specific to ICC;  The latest DPC++ shows warning   "warning: extra tokens at the end of '#pragma omp ordered' are   ignored"
[onedpl_config.h#L185](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L185) | omp declare reduction(…) | + | User-Defined-Reduction is a part of OpenMP 4.5; DPC++ claims full support   of OpenMP 4.5
[onedpl_config.h#L190](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L190) |  omp simd scan clause with a user-defined binary | since 2023.1 | It should be used at least since 2023.1 to have access to `omp scan ...`
[onedpl_config.h#L209](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L209) | vector unaligned | - | This pragma is in   https://www.intel.com/content/www/us/en/docs/cpp-compiler/developer-guide-reference/2021-10/vector-001.html#GUID-2B528C72-29B4-4DBC-8B91-EE4D1A03A850,   but it is absent in   https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-2/vector.html
[onedpl_config.h#L217](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L217) | vector nontemporal | + | The pragma is in https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-2/vector.html
[onedpl_config.h#L224](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/onedpl_config.h#L224) | _ONEDPL_PRAGMA_LOCATION | ? | It does not look very important. To investigate if time remains.
[iterator_impl.h#L792](https://github.com/oneapi-src/oneDPL/blob/e941380cc272fb7aaed67b0e4344bbba97e3e7e0/include/oneapi/dpl/pstl/iterator_impl.h#L792) | _ONEDPL_CONSTEXPR_FIX |  + | DPC++ compiler is implicitly covered because of `__clang__`.
[test_config.h](https://github.com/oneapi-src/oneDPL/blob/2d943a30993eb864cc25768a509dd411ce5df92d/test/support/test_config.h#L30C17-L30C34) | - |   | Legacy "Broken" macros, which cover non-supported ICC versions. Clean up (probably in a separate PR).

Note that this does not cover open-source Intel llvm-based compier, [oneAPI DPC++](https://github.com/intel/llvm) compiler: it does not provide `__INTEL_LLVM_COMPILER` macro.

### Manual testing

Ubuntu 22.04:
- TBB backend:
  - Intel(R) oneAPI DPC++/C++ Compiler 2022.2.1: **199/199** tests pass
  - Intel(R) oneAPI DPC++/C++ Compiler 2023.2.0:  **199/199** tests pass
  - Intel(R) oneAPI DPC++/C++ Compiler 202x.x.x (latest build): **199/199** tests pass
- OpenMP backend
  - Intel(R) oneAPI DPC++/C++ Compiler 2022.2.1: **199/199** tests pass
  - Intel(R) oneAPI DPC++/C++ Compiler 2023.2.0:  **199/199** tests pass
  - Intel(R) oneAPI DPC++/C++ Compiler 202x.x.x (latest build): **199/199** tests pass


include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -115,21 +116,22 @@
...
+#if (__INTEL_LLVM_COMPILER || __INTEL_COMPILER)
 #    define _ONEDPL_PRAGMA_FORCEINLINE _ONEDPL_PRAGMA(forceinline)
 #elif defined(_PSTL_PRAGMA_FORCEINLINE)
 #    define _ONEDPL_PRAGMA_FORCEINLINE _PSTL_PRAGMA_FORCEINLINE
 #else
 #    define _ONEDPL_PRAGMA_FORCEINLINE
 #endif
 
-#if (__INTEL_COMPILER >= 1900)
+#if (__INTEL_LLVM_COMPILER >= 20230100 || __INTEL_COMPILER >= 1900)
```
MikeDvorskiy, line 134: What' about compilers `__INTEL_LLVM_COMPILER  < 20230100` ?
Are  these covered by `__INTEL_COMPILER >= 1900` condition?

include/oneapi/dpl/pstl/pstl_config.h
```diff
@@ -66,7 +66,8 @@
 #endif
 
 // Enable SIMD for compilers that support OpenMP 4.0
-#if (_OPENMP >= 201307) || (__INTEL_COMPILER >= 1600) || (!defined(__INTEL_COMPILER) && _PSTL_GCC_VERSION >= 40900) || \
+#if (__INTEL_LLVM_COMPILER || __INTEL_COMPILER ||                                                                      \
```
MikeDvorskiy, line 69: it looks that we lost `(_OPENMP >= 201307) ` condition

include/oneapi/dpl/pstl/pstl_config.h
```diff
@@ -81,13 +82,13 @@
...
 
-#if (__INTEL_COMPILER)
+#if (__INTEL_LLVM_COMPILER || __INTEL_COMPILER)
 #    define _PSTL_PRAGMA_FORCEINLINE _PSTL_PRAGMA(forceinline)
 #else
 #    define _PSTL_PRAGMA_FORCEINLINE
 #endif
 
-#if (__INTEL_COMPILER >= 1900)
+#if (__INTEL_LLVM_COMPILER >= 20230100 || __INTEL_COMPILER >= 1900)
```
MikeDvorskiy, line 91: Is there a compiler version which doesn't  stratify` (__INTEL_LLVM_COMPILER >= 20230100 || __INTEL_COMPILER >= 1900)`?

dmitriy-sobolev: Sure, there are even among the supported ones. For example, Intel(R) oneAPI DPC++/C++ Compiler 2022.2.1.

julianmi:
> Note that this does not cover open-source Intel llvm-based compier, [oneAPI DPC++](https://github.com/intel/llvm) compiler: it does not provide `__INTEL_LLVM_COMPILER` macro.

Is there another macro that we could use to support the open-source compiler?

include/oneapi/dpl/pstl/pstl_config.h
MikeDvorskiy: and?.. at the same tine `__INTEL_COMPILER >= 1900 `is true?

include/oneapi/dpl/pstl/onedpl_config.h
MikeDvorskiy: you want to say that DPC++ witch __INTEL_LLVM_COMPILER < 20230100, doesnt support features defined on lines 135.....

include/oneapi/dpl/pstl/pstl_config.h
dmitriy-sobolev: DPC++ compilers do not define `__INTEL_COMPILER`.

include/oneapi/dpl/pstl/onedpl_config.h
MikeDvorskiy: I that case, I think, it makes sense to add a description in the code... otherwise, such condition looks as non quite correct  - we "skip"  the compliers which `__INTEL_LLVM_COMPILER < 20230100.`..

include/oneapi/dpl/pstl/pstl_config.h
dmitriy-sobolev: Fixed

dmitriy-sobolev:
> > Note that this does not cover open-source Intel llvm-based compier, [oneAPI DPC++](https://github.com/intel/llvm) compiler: it does not provide `__INTEL_LLVM_COMPILER` macro.
> 
> Is there another macro that we could use to support the open-source compiler?

I assume that [oneAPI DPC++](https://github.com/intel/llvm) is clang++ with SYCL extensions, so we should treat it as clang++ when working with host code. So we can rely on `__clang__` or `__llvm__`. But this is probably beyond the scope of this PR.

danhoeflinger:
>This is called UDS (User-Defined-SCAN?), but it seem the restrictions are the same as in User-Defined-Reduction; May be it also should be restricted to 2023.1

I can confirm that this is in fact user-defined-scan, and specific to the openmp simd scan clause with a user-defined binary operation as the scan operation (rather than `+` for instance).  It certainly requires 2023.1 (to have access to openmp simd scan).  I don't know yet for certain that user defined versions were fully supported as of that version.

danhoeflinger:
>The failing test is scan.pass and it is a regression since this PR. This case produces an error (wrong result from exclusive_scan with binary operator, at index 1):

I expect this failure is due to UDS, so it would seem to confirm that it works on 2023.2, but not 2022.2.1.  Its likely it works on 2023.1 as well, but we should check.

dmitriy-sobolev:
> I can confirm that this is in fact user-defined-scan, and specific to the openmp simd scan clause with a user-defined binary operation as the scan operation (rather than + for instance). It certainly requires 2023.1 (to have access to openmp simd scan). I don't know yet for certain that user defined versions were fully supported as of that version.

Thanks, If I limited UDS use to 2023.1 and it fixed the error. 

> I expect this failure is due to UDS, so it would seem to confirm that it works on 2023.2, but not 2022.2.1. Its likely it works on 2023.1 as well, but we should check.
> 
> edit: Actually I'm a bit confused, as I would expect a compilation error rather than a runtime error. However it seems to coincide with the UDS test.

A compiler can silently ignore non-supported pragmas according to https://en.cppreference.com/w/cpp/preprocessor/impl:
> Pragma directive controls implementation-specific behavior of the compiler, such as disabling compiler warnings or changing alignment requirements. Any pragma that is not recognized is ignored.

include/oneapi/dpl/pstl/pstl_config.h
```diff
@@ -105,13 +106,19 @@
...
 #else
 #    define _PSTL_UDR_PRESENT 0
 #endif
 
-#define _PSTL_UDS_PRESENT (__INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626)
+#if (__INTEL_LLVM_COMPILER >= 20230100) || (__INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626)
+#    define _ONEDPL_UDS_PRESENT 1
+#else
+#    define _ONEDPL_UDS_PRESENT 0
+#endif
```
danhoeflinger, line 121: Interestingly with 2023.1 on Windows (because that is where I have that version installed) and tbb backend, this results in a failure
```
.\build\test\scan.pass.exe
error at C:\Users\dhoeflin\source\repos\oneDPL\test\parallel_api\numeric\numeric.ops\scan.pass.cpp:137 - wrong result from inclusive_scan with binary operator, at index 1
```
If I disable UDS, it succeeds.  Its unclear if this is a windows issue or a 2023.1-> 2023.2 support issue.  Sadly I don't have 2023.1 on Linux, or 2023.2 on Windows to easily check.

danhoeflinger: Looks likely to be a Windows issue, 2023.2 TBB backend also fails with the same error, and passes when UDS is turned off.  Still remains to check Linux 2023.1.

mmichel11:
I believe our tests are still using `-fopenmp` and `-fopenmp-simd`. Would it be useful to test with `-fiopenmp` and `-fiopenmp-simd` as well?

mmichel11:
I believe our tests are still using `-fopenmp` and `-fopenmp-simd`. Would it be useful to test manually with `-fiopenmp` and `-fiopenmp-simd` as well?

include/oneapi/dpl/pstl/pstl_config.h
danhoeflinger: I just checked on 2023.1 and 2023.2 on Linux, and it seems to succeed.
My guess is that this is a bug on Windows, but that 2023.1 is the appropriate version to enable UDS on Linux.

danhoeflinger:
To add to the matrix on the linux side: 
I ran 2023.1.0 with tbb and omp backend on linux and got 199/199 passed for both with the current PR.

danhoeflinger:
> I believe our tests are still using `-fopenmp` and `-fopenmp-simd`. Would it be useful to test manually with `-fiopenmp` and `-fiopenmp-simd` as well?

Others may differ in opinion, but from my perspective and with the timeline we have, I'd voteto keep the exploration to the "recommended" build settings within CMake.  There is at least one known issue within `-fiopenmp` for some patterns, which is why we use `-fopenmp` currently.  It may be difficult to disentangle those issues from any possible ones introduced here.  

If we have a working version with the recommended build settings, and people encounter problems with other build settings, we can point them to use these recommended settings until we have more time to investigate and resolve issues with `-fiopenmp`.

dmitriy-sobolev:
uninitialized_fill_destroy fails in the following configuration:
 - Windows, 
 - /O2 or /O3 options, 
 - destroy algorithm
 - par or par_unseq policies
 - Wrapper\<std::int8_t*\> element type

It looks loke some compiler bug.

mmichel11:
> > I believe our tests are still using `-fopenmp` and `-fopenmp-simd`. Would it be useful to test manually with `-fiopenmp` and `-fiopenmp-simd` as well?
> 
> Others may differ in opinion, but from my perspective and with the timeline we have, I'd vote to keep the exploration to the "recommended" build settings within CMake. There is at least one known issue within `-fiopenmp` for some patterns, which is why we use `-fopenmp` currently. It may be difficult to disentangle those issues from any possible ones introduced here.
> 
> If we have a working version with the recommended build settings, and people encounter problems with other build settings, we can point them to use these recommended settings until we have more time to investigate and resolve issues with `-fiopenmp`.

I think this is reasonable to do and agree that we just keep looking at the default CMake build settings.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: > I that case, I think, it makes sense to add a description in the code... otherwise, such condition looks as non quite correct condition - we "skip" the compliers which __INTEL_LLVM_COMPILER < 20230100...

@MikeDvorskiy, I guess this does not look straightforward probably because of an assumption that DPCPP is a linear continuation of ICC, so let's say DPCPP 2021 should support features which are supported by ICC 2019. This does not look right when comparing e.g. [OpenMP* Features and Extensions Supported in Intel® oneAPI DPC++/C++ Compiler](https://www.intel.com/content/www/us/en/developer/articles/technical/openmp-features-and-extensions-supported-in-icx.html) and [A Survey of OpenMP* Features Implemented in Intel® Fortran and C++ Compilers](https://www.intel.com/content/www/us/en/developer/articles/technical/a-survey-of-openmp-features-implemented-in-intel-fortran-and-c-compilers.html).

Does it solve your confusion (if I guessed it right)? Would you still prefer adding an explanation?

include/oneapi/dpl/pstl/glue_memory_impl.h
```diff
@@ -232,9 +232,18 @@ destroy(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __
     if constexpr (!::std::is_trivially_destructible_v<_ValueType>)
     {
 #if _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDWOS_BROKEN
-        oneapi::dpl::__internal::__pattern_walk1(::std::forward<_ExecutionPolicy>(__exec), __first, __last,
-                                                 [](_ReferenceType __val) { __val.~_ValueType(); },
-                                                 /*__is_vector*/ ::std::false_type{}, __is_parallel);
+        if constexpr (oneapi::dpl::__internal::__enable_if_host_execution_policy<_ExecutionPolicy>::value)
```
SergeyKopienko, line 235: Do you really need duplicate if/else branches?
May be better to eval one parameter value?

include/oneapi/dpl/pstl/glue_memory_impl.h
```diff
@@ -262,9 +271,18 @@ destroy_n(_ExecutionPolicy&& __exec, _ForwardIterator __first, _Size __n)
     else
     {
 #if _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDWOS_BROKEN
-        return oneapi::dpl::__internal::__pattern_walk1_n(::std::forward<_ExecutionPolicy>(__exec), __first, __n,
-                                                          [](_ReferenceType __val) { __val.~_ValueType(); },
-                                                          /*__is_vector*/ ::std::false_type{}, __is_parallel);
+        if constexpr (oneapi::dpl::__internal::__enable_if_host_execution_policy<_ExecutionPolicy>::value)
```
SergeyKopienko, line 274: The same

include/oneapi/dpl/pstl/pstl_config.h
dmitriy-sobolev: Thanks, Dan. I have turned off UDS on Windows. Checking that we do not break anything with 2023.2 on Linux was also useful.

include/oneapi/dpl/pstl/pstl_config.h
```diff
@@ -105,13 +106,27 @@
...
 #    define _PSTL_UDR_PRESENT 1
 #else
 #    define _PSTL_UDR_PRESENT 0
 #endif
 
-#define _PSTL_UDS_PRESENT (__INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626)
+// UDS is broken with __INTEL_LLVM_COMPILER on Windows,
+// however it may be fixed in the future releases (ones which follow 20240000)
+#if (defined(_MSC_VER) && __INTEL_LLVM_COMPILER > 20240000) || (!defined(_MSC_VER) && __INTEL_LLVM_COMPILER > 20230100)
+#    define _ONEDPL_INTEL_LLVM_UDS_PRESENT 1
```
MikeDvorskiy, line 120: Why do we introduce `_ONEDPL_*` macro in `pstl_config.h` ?
the upstream pstl version has just `_PSTL_*` macro everywhere.

dmitriy-sobolev: Thanks for pointing it out. I have replaced ONDPL with PSTL in pstl_config.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -294,4 +310,12 @@
...
 #endif
 
+// Intel(R) oneAPI DPC++ Compiler cannot build some cases of destroy and destroy_n with unseq policies on Windows
+// It might be fixed in the future releases, which follow 20240000
+#if defined(_MSC_VER) && __INTEL_LLVM_COMPILER <= 20240000
+#    define _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDWOS_BROKEN 1
+#else
+#    define _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDWOS_BROKEN 0
+#endif
+
```
danhoeflinger, line 320: Do we want to predict that this will be fixed in `__INTEL_LLVM_COMPILER > 20240000`?
This seems like a ticking time bomb.  For me I would rather have just `__INTEL_LLVM_COMPILER` and along with a TODO to reenable it once we have a version which we know works (perhaps a github issue as well to help us remember).

include/oneapi/dpl/pstl/pstl_config.h
```diff
@@ -169,4 +184,12 @@
 // broken macros
 #define _PSTL_ICC_18_OMP_SIMD_BROKEN (__INTEL_COMPILER == 1800)
 
+// Intel(R) oneAPI DPC++ Compiler cannot build some cases of destroy and destroy_n with unseq policies on Windows
+// It might be fixed in the future releases, which follow 20240000
+#if defined(_MSC_VER) && __INTEL_LLVM_COMPILER <= 20240000
+#    define _PSTL_ICPX_OMP_SIMD_DESTROY_WINDWOS_BROKEN 1
+#else
+#    define _PSTL_ICPX_OMP_SIMD_DESTROY_WINDWOS_BROKEN 0
+#endif
```
danhoeflinger, line 193: typo, need to correct its usage too.
```suggestion
#if defined(_MSC_VER) && __INTEL_LLVM_COMPILER <= 20240000
#    define _PSTL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN 1
#else
#    define _PSTL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN 0
#endif
```

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -294,4 +310,12 @@
 #    define _ONEDPL_SYCL_INTEL_COMPILER 1
 #endif
 
+// Intel(R) oneAPI DPC++ Compiler cannot build some cases of destroy and destroy_n with unseq policies on Windows
+// It might be fixed in the future releases, which follow 20240000
+#if defined(_MSC_VER) && __INTEL_LLVM_COMPILER <= 20240000
+#    define _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDWOS_BROKEN 1
+#else
+#    define _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDWOS_BROKEN 0
+#endif
```
danhoeflinger, line 319: typo, need to correct its usage too.

```suggestion
#if defined(_MSC_VER) && __INTEL_LLVM_COMPILER <= 20240000
#    define _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN 1
#else
#    define _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN 0
#endif
```

include/oneapi/dpl/pstl/glue_memory_impl.h
danhoeflinger: Yes, I think this should be able to be collapsed back into a single line of code for the `__pattern_walk1` call, with another variable assigned based on `_ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN`.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: Yes, your suggestion looks better. I did it this way so we instantly notice that and raise the version if necessary. It might work well because we have early access to newest versions, but it is definitely not a good idea for pstl_config, which can be unstreamed.

danhoeflinger: I'm also thinking about the case where a user updates the compiler, but not oneDPL (for whatever reason).  
In the case where the compiler bug is one that is not fixed in 2024.1.0, lets say we then update the version in oneDPL for the matching oneAPI release because we see the failing test.  
If the user updates the compiler alone, it will break their code, and moreover look like a *new* compiler bug.  It is unlikely to be clear to them that upgrading to the newest oneDPL would be a resolution for their problem.  

I'd favor being conservative here, but taking some extra precautions (github issue) to see that it isn't forgotten about.

include/oneapi/dpl/pstl/onedpl_config.h
danhoeflinger: Unresolving until the usage is also fixed (so it isn't forgotten).

include/oneapi/dpl/pstl/glue_memory_impl.h
SergeyKopienko: May be something like this?

```C++
template <class _ExecutionPolicy>
constexpr bool
eval_expr_for_destroy(_ExecutionPolicy __exec)
{
#if _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDWOS_BROKEN
    if constexpr (oneapi::dpl::__internal::__is_host_execution_policy<_ExecutionPolicy>::value)
        return ::std::false_type{};
    else
#endif
        return oneapi::dpl::__internal::__is_vectorization_preferred<_ExecutionPolicy, _ForwardIterator>(__exec);
}

// [specialized.destroy]

template <class _ExecutionPolicy, class _ForwardIterator>
oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy>
destroy(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last)
{
    typedef typename ::std::iterator_traits<_ForwardIterator>::value_type _ValueType;
    typedef typename ::std::iterator_traits<_ForwardIterator>::reference _ReferenceType;

    const auto __is_parallel =
        oneapi::dpl::__internal::__is_parallelization_preferred<_ExecutionPolicy, _ForwardIterator>(__exec);

    oneapi::dpl::__internal::__pattern_walk1(::std::forward<_ExecutionPolicy>(__exec), __first, __last,
                                            [](_ReferenceType __val) { __val.~_ValueType(); },
                                            eval_expr_for_destroy<_ExecutionPolicy>(__exec), __is_parallel);
}
```

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: Fixed

include/oneapi/dpl/pstl/glue_memory_impl.h
dmitriy-sobolev: I am worried that it would not work. `eval_expr_for_destroy` is not evaluated at compile-time, because `__exec` run-time argument. `eval_expr_for_destroy` must be a `constexpr` to avoid compilation of a broken configuration.

dmitriy-sobolev: I have followed the suggestion to remove `#else` branch.

dmitriy-sobolev: I had to revert that change, because if `#else` branch is removed, then the default one with broken configuration is always compiled...

danhoeflinger: Thats a good point.  It needs to be remove from the code path as a constexpr to resolve the crash...

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -181,13 +183,27 @@
...
+     !defined(__INTEL_LLVM_COMPILER) && !defined(__INTEL_COMPILER) && _ONEDPL_GCC_VERSION >= 40900 ||                  \
+     _OPENMP >= 201307)
 #    define _ONEDPL_UDR_PRESENT 1
 #else
 #    define _ONEDPL_UDR_PRESENT 0
 #endif
 
-#define _ONEDPL_UDS_PRESENT (__INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626)
+// UDS is broken with Intel(R) oneAPI DPC++ Compiler on Windows
+// It might be fixed in the future releases, which follow 20240000
```
danhoeflinger, line 195: Probably don't want this comment anymore.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -294,4 +310,12 @@
 #    define _ONEDPL_SYCL_INTEL_COMPILER 1
 #endif
 
+// Intel(R) oneAPI DPC++ Compiler cannot build some cases of destroy and destroy_n with unseq policies on Windows
+// It might be fixed in the future releases, which follow 20240000
```
danhoeflinger, line 314: probably dont want this comment anymore

include/oneapi/dpl/pstl/pstl_config.h
```diff
@@ -105,13 +106,27 @@
...
+     !defined(__INTEL_LLVM_COMPILER) && !defined(__INTEL_COMPILER) && _PSTL_GCC_VERSION >= 40900 ||                    \
+     _OPENMP >= 201307)
 #    define _PSTL_UDR_PRESENT 1
 #else
 #    define _PSTL_UDR_PRESENT 0
 #endif
 
-#define _PSTL_UDS_PRESENT (__INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626)
+// UDS is broken with Intel(R) oneAPI DPC++ Compiler on Windows
+// It might be fixed in the future releases, which follow 20240000
```
danhoeflinger, line 118: same here, dont want this comment.

include/oneapi/dpl/pstl/pstl_config.h
```diff
@@ -169,4 +184,12 @@
 // broken macros
 #define _PSTL_ICC_18_OMP_SIMD_BROKEN (__INTEL_COMPILER == 1800)
 
+// Intel(R) oneAPI DPC++ Compiler cannot build some cases of destroy and destroy_n with unseq policies on Windows
+// It might be fixed in the future releases, which follow 20240000
```
danhoeflinger, line 188: again, remove comment.

danhoeflinger (dismissed):
Other than the minor issues with comments and considering the suggestions from clang-format, it LGTM.

include/oneapi/dpl/pstl/glue_memory_impl.h
SergeyKopienko: And without `constexpr` is this variant useful?

akukanov: What about this:

```
    const auto __is_vector =
#if _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN
        ::std::conditional_t<
            oneapi::dpl::__internal::__is_host_execution_policy<::std::decay_t<_ExecutionPolicy>>::value,
            ::std::false_type,
            oneapi::dpl::__internal::__is_vectorization_preferred<_ExecutionPolicy, _ForwardIterator>(__exec)
        >;
#else
        oneapi::dpl::__internal::__is_vectorization_preferred<_ExecutionPolicy, _ForwardIterator>(__exec);
#endif // _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN

        oneapi::dpl::__internal::__pattern_walk1(::std::forward<_ExecutionPolicy>(__exec), __first, __last,
                                                 [](_ReferenceType __val) { __val.~_ValueType(); },
                                                 __is_vector, __is_parallel);
```

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -181,13 +183,27 @@
 #    define _ONEDPL_PRAGMA_SIMD_ORDERED_MONOTONIC_2ARGS(PRM1, PRM2)
 #endif
 
-#if (__INTEL_COMPILER >= 1900 || !defined(__INTEL_COMPILER) && _ONEDPL_GCC_VERSION >= 40900 || _OPENMP >= 201307)
+#if (__INTEL_LLVM_COMPILER || __INTEL_COMPILER >= 1900 ||                                                              \
+     !defined(__INTEL_LLVM_COMPILER) && !defined(__INTEL_COMPILER) && _ONEDPL_GCC_VERSION >= 40900 ||                  \
+     _OPENMP >= 201307)
```
akukanov, line 188: Move `_OPENMP >= 201307` to the first line?

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -181,13 +183,27 @@
...
+#    define _ONEDPL_INTEL_LLVM_UDS_PRESENT 1
+#else
+#    define _ONEDPL_INTEL_LLVM_UDS_PRESENT 0
+#endif
+
+#if _ONEDPL_INTEL_LLVM_UDS_PRESENT || (__INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626)
+#    define _ONEDPL_UDS_PRESENT 1
+#else
+#    define _ONEDPL_UDS_PRESENT 0
+#endif
```
akukanov, line 206: Combine instead as `#if..#elif..#else`?

Also, I prefer to limit the compiler version to the known non-working, as if the next version would work. If it will not, we should see it in testing and increase the limit. It requires more effort, but has less chances to be forgotten once a fix is in place.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -294,4 +310,12 @@
 #    define _ONEDPL_SYCL_INTEL_COMPILER 1
 #endif
 
+// Intel(R) oneAPI DPC++ Compiler cannot build some cases of destroy and destroy_n with unseq policies on Windows
+// It might be fixed in the future releases, which follow 20240000
+#if defined(_MSC_VER) && __INTEL_LLVM_COMPILER
+#    define _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN 1
+#else
+#    define _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN 0
+#endif
```
akukanov, line 319: See the comment above about limiting applicability of "broken" macros.

akukanov (commented):
Comments for `onedpl_config.h` also may apply to `pstl_config.h`

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: There are several arguments, why it is better not to limit the version until we know when it is fixed: please, find them in https://github.com/oneapi-src/oneDPL/pull/1182#discussion_r1327230978. Would you still prefer limiting the version?

akukanov: Thanks for pointing out this discussion, it's useful to see the downsides, especially the potential customer impact.

Maybe we can introduce a special function-like macro like this:
```
#define _ONEDPL_BROKEN_AT_VERSION(M, V) (M) 
```
which for testing we can redefine to
```
#define _ONEDPL_BROKEN_AT_VERSION(M, V) (M <= V) 
```

But for now, OK to leave with a TODO.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: I left a TODO comment as it was suggested earlier.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: I left a TODO comment as it was suggested earlier.

include/oneapi/dpl/pstl/pstl_config.h
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/pstl_config.h
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: Done.

include/oneapi/dpl/pstl/glue_memory_impl.h
dmitriy-sobolev: Thanks for the idea, Alexey. I have changed the code accordingly with some minor corrections.

include/oneapi/dpl/pstl/glue_memory_impl.h
dmitriy-sobolev: It's been implemented as a part of https://github.com/oneapi-src/oneDPL/pull/1182#discussion_r1328439292

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: > Combine instead as #if..#elif..#else?

Done

dmitriy-sobolev: > Maybe we can introduce a special function-like macro like this: ...

I will do it in a separate PR and #1191 is a reminder.

include/oneapi/dpl/pstl/glue_memory_impl.h
```diff
@@ -226,14 +228,23 @@ destroy(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __
...
+#if _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN
+        ::std::conditional_t<
+            oneapi::dpl::__internal::__is_host_execution_policy<::std::decay_t<_ExecutionPolicy>>::value,
+            ::std::false_type,
+            decltype(oneapi::dpl::__internal::__is_vectorization_preferred<_ExecutionPolicy, _ForwardIterator>(
+                __exec))>;
+#else
+        decltype(oneapi::dpl::__internal::__is_vectorization_preferred<_ExecutionPolicy, _ForwardIterator>(__exec));
+#endif // _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN
+    const _is_vector_type __is_vector;
```
MikeDvorskiy, line 241: Whether  do we want to use `constexpr` here?

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -181,13 +183,21 @@
...
 #    define _ONEDPL_UDR_PRESENT 1
 #else
 #    define _ONEDPL_UDR_PRESENT 0
 #endif
 
-#define _ONEDPL_UDS_PRESENT (__INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626)
+// TODO: enable with Intel LLVM-based compiler on Windows when fixed
+#if __INTEL_LLVM_COMPILER >= 20230100 && !defined(_MSC_VER)
+#    define _ONEDPL_UDS_PRESENT 1
+#elif __INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626
```
MikeDvorskiy, line 196: Probably, it makes sense to write one "common" condition instead of two(194 and 196 lines) and keep just one `#define _ONEDPL_UDS_PRESENT 1`?

include/oneapi/dpl/pstl/glue_memory_impl.h
dmitriy-sobolev: This is not necessary for avoiding compilation of unseq version when configuration is broken. The type is inferred in compile type. 
I decided to leave `const` so this line is aligned with similar occurrences in this file.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -294,4 +304,11 @@
 #    define _ONEDPL_SYCL_INTEL_COMPILER 1
 #endif
 
+// TODO: enable when fixed
```
MikeDvorskiy, line 307: probably, the comment should be more informational. "who" fixed ?

include/oneapi/dpl/pstl/pstl_config.h
```diff
@@ -169,4 +177,11 @@
 // broken macros
 #define _PSTL_ICC_18_OMP_SIMD_BROKEN (__INTEL_COMPILER == 1800)
 
+// TODO: enable when fixed
```
MikeDvorskiy, line 180: Probably, the comment should be more informational.

include/oneapi/dpl/pstl/glue_memory_impl.h
```diff
@@ -246,8 +257,17 @@ destroy_n(_ExecutionPolicy&& __exec, _ForwardIterator __first, _Size __n)
...
```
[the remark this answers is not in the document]
dmitriy-sobolev: I have added it.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: It's been changed to have only one `#define _ONEDPL_UDS_PRESENT 1`.

include/oneapi/dpl/pstl/pstl_config.h
```diff
@@ -105,13 +105,21 @@
...
 #else
 #    define _PSTL_UDR_PRESENT 0
 #endif
 
-#define _PSTL_UDS_PRESENT (__INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626)
+// TODO: enable with Intel LLVM-based compiler on Windows when fixed
+#if __INTEL_LLVM_COMPILER >= 20230100 && !defined(_MSC_VER)
+#    define _PSTL_UDS_PRESENT 1
+#elif __INTEL_COMPILER >= 1900 && __INTEL_COMPILER_BUILD_DATE >= 20180626
+#    define _PSTL_UDS_PRESENT 1
```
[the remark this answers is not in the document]
dmitriy-sobolev: It's been changed to have only one `#define _ONEDPL_UDS_PRESENT 1`.

include/oneapi/dpl/pstl/pstl_config.h
dmitriy-sobolev: It's been clarified.

danhoeflinger (approved):
LGTM

include/oneapi/dpl/pstl/glue_memory_impl.h
```diff
@@ -226,14 +228,23 @@ destroy(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __
 
     const auto __is_parallel =
         oneapi::dpl::__internal::__is_parallelization_preferred<_ExecutionPolicy, _ForwardIterator>(__exec);
-    const auto __is_vector =
-        oneapi::dpl::__internal::__is_vectorization_preferred<_ExecutionPolicy, _ForwardIterator>(__exec);
+    using _is_vector_type =
+#if _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN
```
MikeDvorskiy, line 232: Sorry, I missed one more comment.
Every where in the current ondDPL code in the simular cases we check two macroses: 
 _ONEDPL_* and _PSTL_*..
 So, we have to add  _PSTL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN everywhere below.

dmitriy-sobolev: Thanks for noticing it. It has been fixed in #1195.

PR #1184, opened 2023-09-14
[pstl-offload] Add SDL flags into libpstloffload build

kboyarinov:
Flags that were not added:

* `-ipo` caused link-time issues on `icpx` because some required library not found in the compiler package. Investigation required.
* `-fsanitize=cfi` causes a warning that the option is ignored as unsupported for SPIRV.
* `-pie` causes a warning that the argument is unused

No performance degradations detected


timmiesmith (approved):
LGTM

Alexandr-Konovalov (approved):
LGTM

PR #1187, opened 2023-09-15
[oneDPL] [test] Fix heap use after free in uninitialized_fill_destroy.pass

mmichel11:
In `uninitialized_fill_destroy.pass.cpp`, we currently allocate memory into a smart pointer to manage its lifetime. However, there are two issues with this: since the memory is allocated with `new` it is already initialized when it is passed to the test function, and the test object's destructor will be called a second time after the `std::destroy` test before deallocation. This has crashes in this test on certain OSs (certain test configs with MacOS) and can be observed with address sanitizers elsewhere.

This fix manually allocates the memory with `malloc` so it is passed uninitialized to the test and manually deallocates it with `free`.

### DIFF
--- a/test/parallel_api/memory/specialized.algorithms/uninitialized_fill_destroy.pass.cpp
+++ b/test/parallel_api/memory/specialized.algorithms/uninitialized_fill_destroy.pass.cpp
@@ -174,7 +174,10 @@ test_uninitialized_fill_destroy_by_type()
     for (size_t n = 0; n <= N; n = n <= 16 ? n + 1 : size_t(3.1415 * n))
     {
 #if !TEST_DPCPP_BACKEND_PRESENT
-        auto p_begin = static_cast<T*>(::std::malloc(sizeof(T) * n));
+        auto free_allocation = [](auto ptr) { ::std::free(ptr); };
+        std::unique_ptr<T[], decltype(free_allocation)> p(static_cast<T*>(::std::malloc(sizeof(T) * n)),
+                                                          free_allocation);
+        auto p_begin = p.get();
 #else
         Sequence<T> p(n, [](size_t){ return T{}; });
         auto p_begin = p.begin();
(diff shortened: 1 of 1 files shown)


test/parallel_api/memory/specialized.algorithms/uninitialized_fill_destroy.pass.cpp
```diff
@@ -198,6 +198,7 @@ test_uninitialized_fill_destroy_by_type()
         invoke_on_all_policies<>()(test_destroy_n<T>(), p_begin, p_end, T(), n,
                                    ::std::is_trivial<T>());
 #endif
+        ::std::free(p_begin);
```
akukanov, line 201: `free` should be guarded by the same preprocessor conditions as `malloc`.

mmichel11: I have switched to using a custom deleter with `std::unique_ptr` that frees the allocation.

akukanov:
Perhaps it makes sense to add a comment why malloc/free is used here.

mmichel11:
> Perhaps it makes sense to add a comment why malloc/free is used here.

I agree. I have added a comment explaining why it is used.

MikeDvorskiy:
> > Perhaps it makes sense to add a comment why malloc/free is used here.
> 
> I agree. I have added a comment explaining why it is used.

I think the added comment doesn't  quite explain why we use "malloc / free" instead of a typical approach with T[] and smart pointer which  calls delete[] within its destructor..

I would suggest re-writting comment like this:

`We use malloc / free instead of a typical approach with operator new T[] and a smart pointer (which  calls operator delete[] within its destructor) to avoid double creation/destroy the elements of type Tin the test.`

MikeDvorskiy (approved):
LGTM

PR #1189, opened 2023-09-18
Remove extra policy decay from oneDPL code

SergeyKopienko:
In this PR we introduce in [include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h](https://github.com/oneapi-src/oneDPL/pull/1189/files#diff-31b8d8fdd06accc04eefb73a2735d34162170cdf1ac6a050510d060296d8de92) some new [alias templates](https://en.cppreference.com/w/cpp/language/type_alias) and use them in oneDPL code:
```C++
template <typename Policy>
using policy_kernel_name = typename ::std::decay_t<Policy>::kernel_name;

template <typename Policy>
inline constexpr unsigned int policy_unroll_factor = ::std::decay_t<Policy>::unroll_factor;
```

### `policy_kernel_name` usage in code:
before:
```C++
    using _Policy = ::std::decay_t<_ExecutionPolicy>;
    using _CustomName = typename _Policy::kernel_name;
    using _ForKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<_CustomName>;
```
after
```C++
    using _ForKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<_CustomName>;
```

### `policy_unroll_factor` usage in code:
before:
```C++
    return oneapi::dpl::execution::make_fpga_policy<::std::decay_t<_Policy>::unroll_factor,
                                                    _NewKernelName<oneapi::dpl::execution::policy_kernel_name<_Policy>>>(
        ::std::forward<_Policy>(__policy));
```
after:
```C++
    return oneapi::dpl::execution::make_fpga_policy<oneapi::dpl::__internal::__policy_unroll_factor<_Policy>,
                                                    _NewKernelName<oneapi::dpl::__internal::__policy_kernel_name<_Policy>>>(
        ::std::forward<_Policy>(__policy));
```

## fpga_policy
In the `fpga_policy` template class we introduce one new function for internal use only:
```C++
template <unsigned int factor = 1, typename KernelName = DefaultKernelNameFPGA>
class fpga_policy : public device_policy<KernelName>
{
    using base = device_policy<KernelName>;

//...

    // For internal use only

    const base&
    __device_policy() const
    {
        return static_cast<const base&>(*this);
    };
};
```
The goal of this - remove creation of the device policy in the file https://github.com/oneapi-src/oneDPL/pull/1189/files#diff-9dd135f546ffa3dee7d9734ea5d398504f22aa661c512d1780fd256071c2ff0f.

## Other

Also in this PR we remove some extra namespace qualificators like `__dpl` and `oneapi::dpl::__internal::`.

### DIFF
--- a/include/oneapi/dpl/pstl/execution_defs.h
+++ b/include/oneapi/dpl/pstl/execution_defs.h
@@ -123,29 +123,26 @@ inline constexpr unsequenced_policy unseq{};
 
 // 2.3, Execution policy type trait
 template <class T>
-struct __is_execution_policy_impl : ::std::false_type
+struct is_execution_policy : ::std::false_type
 {
 };
 
 template <>
-struct __is_execution_policy_impl<oneapi::dpl::execution::sequenced_policy> : ::std::true_type
+struct is_execution_policy<oneapi::dpl::execution::sequenced_policy> : ::std::true_type
 {
 };
 template <>
-struct __is_execution_policy_impl<oneapi::dpl::execution::parallel_policy> : ::std::true_type
+struct is_execution_policy<oneapi::dpl::execution::parallel_policy> : ::std::true_type
 {
 };
 template <>
-struct __is_execution_policy_impl<oneapi::dpl::execution::parallel_unsequenced_policy> : ::std::true_type
+struct is_execution_policy<oneapi::dpl::execution::parallel_unsequenced_policy> : ::std::true_type
 {
 };
 template <>
-struct __is_execution_policy_impl<oneapi::dpl::execution::unsequenced_policy> : ::std::true_type
+struct is_execution_policy<oneapi::dpl::execution::unsequenced_policy> : ::std::true_type
 {
 };
-
-template <typename T>
-using is_execution_policy = __is_execution_policy_impl<::std::decay_t<T>>;
 
 template <class T>
 inline constexpr bool is_execution_policy_v = oneapi::dpl::execution::is_execution_policy<T>::value;
@@ -186,7 +183,8 @@ template <typename _ExecPolicy>
 inline constexpr bool __is_host_execution_policy_v = __is_host_execution_policy<_ExecPolicy>::value;
 
 template <class _ExecPolicy, class _T = void>
-using __enable_if_execution_policy = ::std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<_ExecPolicy>, _T>;
+using __enable_if_execution_policy =
+    ::std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<::std::decay_t<_ExecPolicy>>, _T>;
 
 template <class _ExecPolicy, class _T = void>
 using __enable_if_host_execution_policy =
--- a/include/oneapi/dpl/pstl/execution_defs.h
+++ b/include/oneapi/dpl/pstl/execution_defs.h
@@ -176,17 +176,20 @@ struct __is_host_execution_policy<oneapi::dpl::execution::unsequenced_policy> : 
 {
 };
 
+template <typename _ExecPolicy>
+inline constexpr bool __is_host_execution_policy_v = __is_host_execution_policy<::std::decay_t<_ExecPolicy>>::value;
+
 template <class _ExecPolicy, class _T = void>
 using __enable_if_execution_policy =
     ::std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<::std::decay_t<_ExecPolicy>>, _T>;
 
 template <class _ExecPolicy, class _T = void>
 using __enable_if_host_execution_policy =
-    ::std::enable_if_t<oneapi::dpl::__internal::__is_host_execution_policy<::std::decay_t<_ExecPolicy>>::value, _T>;
+    ::std::enable_if_t<oneapi::dpl::__internal::__is_host_execution_policy_v<_ExecPolicy>, _T>;
 
 template <class _ExecPolicy, const bool __condition, class _T = void>
 using __enable_if_host_execution_policy_conditional = ::std::enable_if_t<
-    oneapi::dpl::__internal::__is_host_execution_policy<::std::decay_t<_ExecPolicy>>::value && __condition, _T>;
+    oneapi::dpl::__internal::__is_host_execution_policy_v<_ExecPolicy> && __condition, _T>;
 
 template <typename _ExecPolicy, typename _T>
 struct __ref_or_copy_impl
--- a/include/oneapi/dpl/pstl/execution_defs.h
+++ b/include/oneapi/dpl/pstl/execution_defs.h
@@ -145,7 +145,7 @@ struct is_execution_policy<oneapi::dpl::execution::unsequenced_policy> : ::std::
 };
 
 template <class T>
-inline constexpr bool is_execution_policy_v = oneapi::dpl::execution::is_execution_policy<T>::value;
+inline constexpr bool is_execution_policy_v = oneapi::dpl::execution::is_execution_policy<::std::decay_t<T>>::value;
 
 } // namespace v1
 } // namespace execution
--- a/include/oneapi/dpl/pstl/execution_defs.h
+++ b/include/oneapi/dpl/pstl/execution_defs.h
@@ -155,43 +155,38 @@ namespace __internal
 
 // Extension: host execution policy type trait
 template <class _T>
-struct __is_host_execution_policy_impl : ::std::false_type
+struct __is_host_execution_policy : ::std::false_type
 {
 };
 
 template <>
-struct __is_host_execution_policy_impl<oneapi::dpl::execution::sequenced_policy> : ::std::true_type
+struct __is_host_execution_policy<oneapi::dpl::execution::sequenced_policy> : ::std::true_type
 {
 };
 template <>
-struct __is_host_execution_policy_impl<oneapi::dpl::execution::parallel_policy> : ::std::true_type
+struct __is_host_execution_policy<oneapi::dpl::execution::parallel_policy> : ::std::true_type
 {
 };
 template <>
-struct __is_host_execution_policy_impl<oneapi::dpl::execution::parallel_unsequenced_policy> : ::std::true_type
+struct __is_host_execution_policy<oneapi::dpl::execution::parallel_unsequenced_policy> : ::std::true_type
 {
 };
 template <>
-struct __is_host_execution_policy_impl<oneapi::dpl::execution::unsequenced_policy> : ::std::true_type
+struct __is_host_execution_policy<oneapi::dpl::execution::unsequenced_policy> : ::std::true_type
 {
 };
-
-template <typename T>
-using __is_host_execution_policy = __is_host_execution_policy_impl<::std::decay_t<T>>;
-
-template <typename _ExecPolicy>
-inline constexpr bool __is_host_execution_policy_v = __is_host_execution_policy<_ExecPolicy>::value;
 
 template <class _ExecPolicy, class _T = void>
 using __enable_if_execution_policy =
     ::std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<::std::decay_t<_ExecPolicy>>, _T>;
 
 template <class _ExecPolicy, class _T = void>
-using __enable_if_host_execution_policy = ::std::enable_if_t<__is_host_execution_policy_v<_ExecPolicy>, _T>;
+using __enable_if_host_execution_policy =
+    ::std::enable_if_t<__is_host_execution_policy<::std::decay_t<_ExecPolicy>>::value, _T>;
 
 template <class _ExecPolicy, const bool __condition, class _T = void>
-using __enable_if_host_execution_policy_conditional =
-    ::std::enable_if_t<__is_host_execution_policy_v<_ExecPolicy> && __condition, _T>;
+using __enable_if_host_execution_policy_conditional = ::std::enable_if_t<
+    __is_host_execution_policy<::std::decay_t<_ExecPolicy>>::value && __condition, _T>;
 
 template <typename _ExecPolicy, typename _T>
 struct __ref_or_copy_impl
--- a/include/oneapi/dpl/pstl/execution_defs.h
+++ b/include/oneapi/dpl/pstl/execution_defs.h
@@ -181,7 +181,7 @@ inline constexpr bool __is_host_execution_policy_v = __is_host_execution_policy<
 
 template <class _ExecPolicy, class _T = void>
 using __enable_if_execution_policy =
-    ::std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<::std::decay_t<_ExecPolicy>>, _T>;
+    ::std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<_ExecPolicy>, _T>;
 
 template <class _ExecPolicy, class _T = void>
 using __enable_if_host_execution_policy =
--- a/include/oneapi/dpl/pstl/execution_defs.h
+++ b/include/oneapi/dpl/pstl/execution_defs.h
@@ -180,16 +180,15 @@ template <typename _ExecPolicy>
 inline constexpr bool __is_host_execution_policy_v = __is_host_execution_policy<::std::decay_t<_ExecPolicy>>::value;
 
 template <class _ExecPolicy, class _T = void>
-using __enable_if_execution_policy =
-    ::std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<_ExecPolicy>, _T>;
+using __enable_if_execution_policy = ::std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<_ExecPolicy>, _T>;
 
 template <class _ExecPolicy, class _T = void>
 using __enable_if_host_execution_policy =
     ::std::enable_if_t<oneapi::dpl::__internal::__is_host_execution_policy_v<_ExecPolicy>, _T>;
 
 template <class _ExecPolicy, const bool __condition, class _T = void>
-using __enable_if_host_execution_policy_conditional = ::std::enable_if_t<
-    oneapi::dpl::__internal::__is_host_execution_policy_v<_ExecPolicy> && __condition, _T>;
+using __enable_if_host_execution_policy_conditional =
+    ::std::enable_if_t<oneapi::dpl::__internal::__is_host_execution_policy_v<_ExecPolicy> && __condition, _T>;
 
 template <typename _ExecPolicy, typename _T>
 struct __ref_or_copy_impl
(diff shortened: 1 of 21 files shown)


test/support/utils_sycl.h
```diff
@@ -96,10 +96,10 @@ template <typename _NewKernelName, typename _Policy,
           oneapi::dpl::__internal::__enable_if_fpga_execution_policy<_Policy, int> = 0>
 auto
 make_new_policy(_Policy&& __policy)
-    -> decltype(TestUtils::make_fpga_policy<::std::decay_t<_Policy>::unroll_factor, _NewKernelName>(
+    -> decltype(TestUtils::make_fpga_policy<oneapi::dpl::execution::__dpl::policy_unroll_factor<_Policy>, _NewKernelName>(
         ::std::forward<_Policy>(__policy)))
 {
-    return TestUtils::make_fpga_policy<::std::decay_t<_Policy>::unroll_factor, _NewKernelName>(
+    return TestUtils::make_fpga_policy<oneapi::dpl::execution::__dpl::policy_unroll_factor<_Policy>, _NewKernelName>(
```
akukanov, line 102: These changes are incorrect due to use of implementation details in a test.

test/support/policy_helpers.h
```diff
@@ -0,0 +1,47 @@
...
+using enable_if_hetero_execution_policy_conditional = ::std::enable_if_t<
+    oneapi::dpl::__internal::__is_hetero_execution_policy_v<::std::decay_t<_ExecPolicy>> && __condition, _T>;
+#endif
+
+template <class _ExecPolicy, const bool __condition>
+using enable_if_non_hetero_execution_policy_conditional = ::std::enable_if_t<
+#if TEST_DPCPP_BACKEND_PRESENT
+    !oneapi::dpl::__internal::__is_hetero_execution_policy_v<::std::decay_t<_ExecPolicy>> &&
+#endif
+    __condition>;
```
akukanov, line 43: While these helpers replace similar constructs in several tests, still they use implementation details of the library. Moreover, combining the policy type check with another condition makes tests less self-explanatory. I prefer to add a single helper that checks if a policy is "heterogeneous" (read: requires the SYCL backend) or not, and only using the public API.

test/support/policy_helpers.h
```diff
@@ -0,0 +1,47 @@
...
+//
+// This file incorporates work covered by the following copyright and permission
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+#ifndef _POLICY_HELPERS_H
+#define _POLICY_HELPERS_H
```
akukanov, line 16: Do we really need one more support header? Why cannot we add the new stuff to `support/utils.h`, for example?

test/support/policy_helpers.h
SergeyKopienko: Fixed.

test/support/policy_helpers.h
SergeyKopienko: Fixed: file removed.

SergeyKopienko:
@akukanov @rarutyun could you please take a look to this PR once more?

SergeyKopienko:
@rarutyun @danhoeflinger could you please take a look at this PR ?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -247,6 +253,10 @@ struct __is_device_execution_policy<execution::device_policy<PolicyParams...>> :
 template <typename... PolicyParams>
 inline constexpr bool __is_device_execution_policy_v = __is_device_execution_policy<PolicyParams...>::value;
 
+template <typename _ExecPolicy, const bool __condition, typename _T = void>
+using __enable_if_device_execution_policy_conditional = ::std::enable_if_t<
+    oneapi::dpl::__internal::__is_device_execution_policy_v<::std::decay_t<_ExecPolicy>> && __condition, _T>;
```
danhoeflinger, line 258: Is there a reason this is inserted as an extra layer and not added in to `__is_device_execution_policy_v` to just decay the incoming `PolicyParams` there?
  
Or probably better still, make an "impl" layer of  `__is_device_execution_policy` to do the actual template specialization, and add the decay to  `__is_device_execution_policy`.

It doesn't make sense to me why this `conditional` version gets the automatic decay, but the others don't.  I'm not sure the `conditional` version needs to exist even though it seems to be used in concert with other conditions.

Also, on the topic of `__is_device_execution_policy_v` ... shouldn't that template parameter be named `_ExecPolicy` , and not be a variadic template?  (This is not part of the PR, but lets fix it if its broken.)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -668,7 +667,7 @@ __parallel_transform_scan_single_group(_ExecutionPolicy&& __exec, _InRng&& __in_
                                        ::std::size_t __n, _UnaryOperation __unary_op, _InitType __init,
                                        _BinaryOperation __binary_op, _Inclusive)
 {
-    using _CustomName = typename std::decay_t<_ExecutionPolicy>::kernel_name;
+    using _CustomName = oneapi::dpl::execution::policy_kernel_name<_ExecutionPolicy>;
```
danhoeflinger, line 670: In some places `oneapi::dpl::execution::policy_kernel_name<_ExecutionPolicy>` is used, in others `oneapi::dpl::execution::__dpl::policy_kernel_name<_ExecutionPolicy>` is used.  

Which is correct? (As implemented, it seems like the second one...)  Should this be made consistent?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
```diff
@@ -76,13 +76,23 @@ struct __parallel_for_fpga_submitter<__internal::__optional_kernel_name<_Name...
     }
 };
 
+template <typename _ExecutionPolicy>
+auto
+__device_policy(_ExecutionPolicy&& __exec)
+    -> decltype(oneapi::dpl::execution::make_device_policy<
+                oneapi::dpl::execution::policy_kernel_name<_ExecutionPolicy>>(__exec.queue()))
```
danhoeflinger, line 83: Is it required to list the type here?  With such a small function, it seems redundant unless it is required for some reason I am missing.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
```diff
@@ -266,14 +239,15 @@ template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typenam
 auto
 __parallel_merge(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2, _Range3&& __rng3, _Compare __comp)
     -> oneapi::dpl::__internal::__enable_if_fpga_execution_policy<
-        _ExecutionPolicy, decltype(oneapi::dpl::__par_backend_hetero::__parallel_merge(
-                              __device_policy(__exec), ::std::forward<_Range1>(__rng1), ::std::forward<_Range2>(__rng2),
-                              ::std::forward<_Range3>(__rng3), __comp))>
+        _ExecutionPolicy,
+        decltype(oneapi::dpl::__par_backend_hetero::__parallel_merge(
+            __device_policy(::std::forward<_ExecutionPolicy>(__exec)), ::std::forward<_Range1>(__rng1),
+            ::std::forward<_Range2>(__rng2), ::std::forward<_Range3>(__rng3), __comp))>
```
danhoeflinger, line 245: Again, is this type listing required?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: Fixed

danhoeflinger: This is better, but I think we can make it slightly better still.  Perhaps something like this? (whatever is done should also be done for `__is_hetero_execution_policy`
```
template <typename _T>
struct __is_device_execution_policy_impl : ::std::false_type
{
};

template <typename... PolicyParams>
struct __is_device_execution_policy_impl<execution::device_policy<PolicyParams...>> : ::std::true_type
{
};


template <typename _ExecPolicy>
using __is_device_execution_policy = __is_device_execution_policy_impl<::std::decay_t<_ExecPolicy>>;


template <typename _ExecPolicy>
inline constexpr bool __is_device_execution_policy_v = __is_device_execution_policy<_ExecPolicy>::value;
```


This does 2 things beyond what is currently in the PR:
1) Adds the decay to `__is_device_execution_policy` so that it can be consistent in whatever flavor someone may want
2) Changes the variadic `PolicyParams` template argument to `_ExecPolicy` at higher levels, where I believe we always want exactly one Execution Policy as input.  (this could be its own PR but why not improve things while we are here)

SergeyKopienko: Looks like the are not exactly the same:
```C++
struct __is_hetero_execution_policy : ::std::false_type
struct __is_hetero_execution_policy<execution::device_policy<PolicyParams...>> : ::std::true_type
 
struct __is_device_execution_policy : ::std::false_type
struct __is_device_execution_policy<execution::device_policy<PolicyParams...>> : ::std::true_type
struct __is_hetero_execution_policy<execution::fpga_policy<unroll_factor, PolicyParams...>> : ::std::true_type
```

danhoeflinger: Great, I think this looks good.  Thanks.

I've created #1294 to track an issue I brought up here which is beyond the scope of this PR, and could be looked at later.

SergeyKopienko: Probably you mean https://github.com/oneapi-src/oneDPL/issues/1298 ?

danhoeflinger: Yes you are correct, I made a typo.  #1298.

danhoeflinger (approved):
LGTM

SergeyKopienko:
@akukanov what is your opinion about this PR? Can I merge it or you think we should change something in this PR ?
Thanks.

include/oneapi/dpl/pstl/execution_defs.h
```diff
@@ -123,27 +123,30 @@ inline constexpr unsequenced_policy unseq{};
...
 };
 template <>
-struct is_execution_policy<oneapi::dpl::execution::unsequenced_policy> : ::std::true_type
+struct __is_execution_policy_impl<oneapi::dpl::execution::unsequenced_policy> : ::std::true_type
 {
 };
 
+template <typename T>
+using is_execution_policy = __is_execution_policy_impl<::std::decay_t<T>>;
+
```
kboyarinov, line 149: Since `is_execution_policy` is a public traits aligned with the C++ Standard definition, are we sure we need to consider cv- or ref- qualified execution policy classes as execution policies? Standard does not requires this trait to be true for any cv- and ref- qualifiers: https://timsong-cpp.github.io/cppwp/n4868/execpol.type. 
Microsoft implementation also defines it in the similar way as it was before this PR.

include/oneapi/dpl/pstl/execution_defs.h
```diff
@@ -123,27 +123,30 @@ inline constexpr unsequenced_policy unseq{};
...
 {
 };
 template <>
-struct is_execution_policy<oneapi::dpl::execution::unsequenced_policy> : ::std::true_type
+struct __is_execution_policy_impl<oneapi::dpl::execution::unsequenced_policy> : ::std::true_type
 {
 };
 
+template <typename T>
+using is_execution_policy = __is_execution_policy_impl<::std::decay_t<T>>;
```
danhoeflinger, line 148: After our offline discussion, I think `is_execution_policy` should not decay the type (as it changes the semantics of a public and defined struct). We should add a layer which we use internally to decay the type automatically if we want that functionality.

Direct usages of this public interface should go back to decaying the type explicitly (in the tests for instance).

The other (internal) versions of this can probably stay as they are.

danhoeflinger (changes requested):

include/oneapi/dpl/pstl/execution_defs.h
kboyarinov: > After our offline discussion, I think `is_execution_policy` should not decay the type (as it changes the semantics of a public and defined struct). We should add a layer which we use internally to decay the type automatically if we want that functionality.
> 
> Direct usages of this public interface should go back to decaying the type explicitly (in the tests for instance).
> 
> The other (internal) versions of this can probably stay as they are currently in the PR.

@danhoeflinger, thanks for explaining. That is exactly what I had in my mind to propose about this issue:)

danhoeflinger: Good to hear we are on the same page.  

The only thing I am unsure of is if we want our internal versions of this pattern to mirror the semantics of this public one.   
I think it is probably unnecessary and will just complicate code and make names longer, but its something worth considering.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -81,6 +81,9 @@ class device_policy
     sycl::queue q;
 };
 
+template <typename Policy>
+using policy_kernel_name = typename ::std::decay_t<Policy>::kernel_name;
+
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -260,7 +274,7 @@ struct __ref_or_copy_impl<execution::device_policy<PolicyParams...>, _T>
 // Extension: hetero execution policy type trait
 template <typename _T>
 using __is_hetero_execution_policy =
-    ::std::disjunction<typename __is_device_execution_policy<_T>::type, typename __is_fpga_execution_policy<_T>::type>;
```
SergeyKopienko, line 263: As proposed @kboyarinov in offline discussion, we able rewrite this definition in more shortly way.

SergeyKopienko: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -162,24 +162,26 @@ template <template <typename> class _NewKernelName, typename _Policy,
 auto
 make_wrapped_policy(_Policy&& __policy)
     -> decltype(oneapi::dpl::execution::make_device_policy<
-                _NewKernelName<typename ::std::decay_t<_Policy>::kernel_name>>(::std::forward<_Policy>(__policy)))
+                _NewKernelName<oneapi::dpl::__internal::__policy_kernel_name<_Policy>>>(
+        ::std::forward<_Policy>(__policy)))
```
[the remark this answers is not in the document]
SergeyKopienko: I think it's out of scope for current PR.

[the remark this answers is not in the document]
SergeyKopienko: Additionally, there are two `make_wrapped_policy` :
```C++
template <template <typename> class _NewKernelName, typename _Policy,
          oneapi::dpl::__internal::__enable_if_device_execution_policy<_Policy, int> = 0>
auto
make_wrapped_policy(_Policy&& __policy)
    -> decltype(oneapi::dpl::execution::make_device_policy<
                _NewKernelName<oneapi::dpl::__internal::__policy_kernel_name<_Policy>>>(
        ::std::forward<_Policy>(__policy)))
{
    return oneapi::dpl::execution::make_device_policy<
        _NewKernelName<oneapi::dpl::__internal::__policy_kernel_name<_Policy>>>(::std::forward<_Policy>(__policy));
}
```

```C++
template <template <typename> class _NewKernelName, typename _Policy,
          oneapi::dpl::__internal::__enable_if_fpga_execution_policy<_Policy, int> = 0>
auto
make_wrapped_policy(_Policy&& __policy)
    -> decltype(oneapi::dpl::execution::make_fpga_policy<
                oneapi::dpl::__internal::__policy_unroll_factor<_Policy>,
                _NewKernelName<oneapi::dpl::__internal::__policy_kernel_name<_Policy>>>(
        ::std::forward<_Policy>(__policy)))
{
    return oneapi::dpl::execution::make_fpga_policy<
        oneapi::dpl::__internal::__policy_unroll_factor<_Policy>,
        _NewKernelName<oneapi::dpl::__internal::__policy_kernel_name<_Policy>>>(::std::forward<_Policy>(__policy));
}
```

I think your proposal not applicable in this case.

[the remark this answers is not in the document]

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
```diff
@@ -39,6 +39,14 @@ namespace dpl
 {
 namespace __par_backend_hetero
 {
+template <typename _ExecutionPolicy>
+auto
+__device_policy(_ExecutionPolicy&& __exec)
```
kboyarinov, line 44: Do we really need to take `ExecutionPolicy` by forwarding reference here? 
The only action we do is reading the `queue` and it is never depends on the policy value category.
So maybe we can take it as const lvalue reference and avoid forwarding the policy while calling the function? What do you think?

SergeyKopienko: Looks like we should create new policy instance with `__kernel_name` from template param policy class.
I think it's the main idea.

kboyarinov: Yes, we need to create a new policy with the same kernel name and with the same queue. As far as I understand, forwarding reference is redundant here.

Also, in all places where this function is called we have a comment:

`// workaround until we implement more performant version for patterns`

If it relates to the policy re-creation with the same kernel name and queue, I guess the comment should be moved inside the new function.

kboyarinov (approved):
LGTM

akukanov (approved):

PR #1193, opened 2023-09-18
[Dynamic Selection] unwrap function with const

AnuyaWelling2801:
Made unwrap functions in selection_type and async_wait_type as const. 
Added tests

### DIFF
--- a/test/support/test_dynamic_load_utils.h
+++ b/test/support/test_dynamic_load_utils.h
@@ -94,8 +94,7 @@ test_unwrap(UniverseContainer u)
     }
 
     //Checking if the const unwrapped submission type is the same as const policy::wait_type
-    const typename oneapi::dpl::experimental::policy_traits<Policy>::wait_type const_wait_test{};
-    if(!std::is_same_v<decltype(const_submission_unwrap),decltype(const_wait_test)>){
+    if(!std::is_same_v<decltype(const_submission_unwrap),const typename oneapi::dpl::experimental::policy_traits<Policy>::wait_type>){
         pass=false;
         std::cout << "ERROR: Unwrapped const submission type is not equal to the policy's wait type\n";
     }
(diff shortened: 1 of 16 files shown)


test/support/test_dynamic_load_utils.h
```diff
@@ -60,6 +60,78 @@ test_dl_initialization(const std::vector<sycl::queue>& u)
...
+        p, [](typename oneapi::dpl::experimental::policy_traits<Policy>::resource_type e) {
+            if constexpr (std::is_same_v<
+                              typename oneapi::dpl::experimental::policy_traits<Policy>::resource_type, int>)
+                return e;
+            else
+                return typename oneapi::dpl::experimental::policy_traits<Policy>::wait_type{};
+        });
+    const auto const_submission_unwrap = oneapi::dpl::experimental::unwrap(const_submission);
+
+    if(std::is_same_v<decltype(const_submission),decltype(const_submission_unwrap)>){
```
vossmjp, line 91: I think the check can be made even stronger. We know that it should NOT be the same as the submission type, but we also know that it should be the resource_type. So, I think it would be better to also check that it returns the expected type.  (This applies for all of the checks/tests).

AnuyaWelling2801: Fixed

test/support/test_offset_utils.h
```diff
@@ -117,9 +140,20 @@ test_unwrap(UniverseContainer u)
         std::cout << "ERROR: Unwrapped selection type is equal to the actual selection type\n";
     }
 
+    //Checking if the unwrapped selection type is the same as policy::resource_type
+    typename oneapi::dpl::experimental::policy_traits<Policy>::resource_type resource_test{};
+    if(!std::is_same_v<decltype(selection_unwrap),decltype(resource_test)>)
```
vossmjp, line 145: Is it necessary to use decltype(resource_test) instead of just oneapi::dpl::experimental::policy_traits<Policy>::resource_type? Seems like the construction of resource_test is not necessary.

vossmjp (commented):
Question about need for resource object created just to get declytype.  Is it necessary?

test/support/test_offset_utils.h
AnuyaWelling2801: Fixed

PR #1195, opened 2023-09-19
Use PSTL-prefixed macro along with ONEDPL one

dmitriy-sobolev:
This is a fix after #1182 PR to make these changes valid for a version where PSTL is used (upstream). It follows https://github.com/oneapi-src/oneDPL/pull/1182#discussion_r1329945495 advice.


MikeDvorskiy (approved):
LGTM

PR #1201, opened 2023-09-25
[oneDPL][ranges] + a minor change for value_type definition

MikeDvorskiy:
[oneDPL][ranges] + a minor change for value_type definition.

Some standard views/genrators (example iota_view) doesn't have `value_type` type.
Our internal "views" are relied on that type of a base range. 
The patch fixes it by introducing another definition  (oneDPL internal) of "value_type".


rarutyun:
Please rebase because, I believe, in that case clang-format should show only changes related to your PR.

rarutyun (approved):
LGTM. Please merge it when CI is finished.

PR #1204, opened 2023-09-25
[pstl-offload] Add workaround for TBB headers tests failures

kboyarinov:
Workaround for the compiler issue when including oneTBB headers in pstl-offload mode - include `<execution>` before any public oneTBB headers. 

This patch replaces `<onedpl/execution>` include in usm memory replacement with private headers (that fixes most of the TBB failing tests) and adds `#include <execution>` as a workaround for other failing tests.

The issue is expected to be fixed on the compiler side and additional macros around changes in this patch would be required.

### DIFF
--- a/include/pstl_offload/execution
+++ b/include/pstl_offload/execution
@@ -22,4 +22,8 @@
 #    undef _ONEDPL_PSTL_OFFLOAD_TOP_LEVEL
 #endif // _ONEDPL_PSTL_OFFLOAD_TOP_LEVEL_EXECUTION
 
+#include <algorithm>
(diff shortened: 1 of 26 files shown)


rarutyun (commented):
I would add some comments to the code, of course, because changes are not obvious at all.
Other than that, looks good to me.

danhoeflinger:
The false positive issues with clang format should be resolved by #1205, if you rebase with main, you should have a more relevant clang-format report.

kboyarinov:
The issue was resolved on the compiler side. No more need in this workaround

PR #1208, opened 2023-09-26
[KT] Esimd radix sort: key-value

dmitriy-sobolev:
This PR adds key-value ESIMD sort and reorganizes internal header structure to avoid obvious duplication. Still, there are a lot of duplicates, but I suggest reconsideration of the architecture in the next PRs.

[[KT] key-value sort prototype](https://github.com/oneapi-src/oneDPL/commit/0e2a6128d421dae78e2148978f4221148a355e93) straightforwardly adds key-value sort.
[[KT] Get rid of duplicates](https://github.com/oneapi-src/oneDPL/commit/62de776a08eb1d5668008220c9a68e5adaf5cf34) removes obvious duplication by restructuring the headers. The new structure is: API (esimd_radix_sort) -> Dispatchers  -> Submitters -> Kernels. It aims to reflect call hierarchy.


include/oneapi/dpl/experimental/kt/esimd_radix_sort_dispatchers.h
```diff
@@ -0,0 +1,323 @@
...
+    constexpr ::std::uint32_t __stage_count = oneapi::dpl::__internal::__dpl_ceiling_div(__bit_count, __radix_bits);
+
+    constexpr ::std::uint32_t __hist_mem_size = __bin_count * sizeof(_GlobalHistT);
+    constexpr ::std::uint32_t __global_hist_mem_size = __stage_count * __hist_mem_size;
+    const ::std::uint32_t __group_hists_mem_size = __sweep_work_group_count * __stage_count * __hist_mem_size;
+    const ::std::uint32_t __tmp_keys_mem_size =  __n * sizeof(_KeyT);
+    const ::std::uint32_t __tmp_vals_mem_size =  __n * sizeof(_ValT);
+    const ::std::size_t __tmp_mem_size = __group_hists_mem_size + __global_hist_mem_size + __tmp_keys_mem_size + __tmp_vals_mem_size;
+
+    // ::std::uint8_t* __p_tmp_mem = sycl::malloc_device<::std::uint8_t>(__tmp_mem_size, __q);
```
SergeyKopienko, line 201: Is this comment required?

SergeyKopienko:
The implementation of the next two functions
`C++
template <typename _KernelName, bool __is_ascending, ::std::uint8_t __radix_bits, ::std::uint16_t __data_per_work_item,
          ::std::uint16_t __work_group_size, typename _KeysRng>
sycl::event
__onesweep(sycl::queue __q, _KeysRng&& __keys_rng, ::std::size_t __n)
`
and
`C++
template <typename _KernelName, bool __is_ascending, ::std::uint8_t __radix_bits, ::std::uint16_t __data_per_work_item,
          ::std::uint16_t __work_group_size, typename _KeysRng, typename _ValsRng>
sycl::event
__onesweep_by_key(sycl::queue __q, _KeysRng&& __keys_rng, _ValsRng&& __vals_rng, ::std::size_t __n)
`
looks like the same in some parts.
May be we should think about remove duplicated code from them?

include/oneapi/dpl/experimental/kt/esimd_radix_sort_submitters.h
```diff
@@ -0,0 +1,253 @@
...
+            [&](sycl::handler& __cgh)
+            {
+                oneapi::dpl::__ranges::__require_access(__cgh, __in_keys_rng, __out_keys_rng, __in_vals_rng, __out_vals_rng);
+                auto __in_keys_data = __in_keys_rng.data();
+                auto __out_keys_data = __out_keys_rng.data();
+                auto __in_vals_data = __in_vals_rng.data();
+                auto __out_vals_data = __out_vals_rng.data();
+                __cgh.depends_on(__e);
+                __radix_sort_onesweep_by_key_slm_reorder_kernel<__is_ascending, __radix_bits, __data_per_work_item, __work_group_size,
+                                                       _KeyT, _ValT, decltype(__in_keys_data), decltype(__out_keys_data), decltype(__in_vals_data), decltype(__out_vals_data)>
```
SergeyKopienko, line 211: `decltype(__in_keys_data), decltype(__out_keys_data), decltype(__in_vals_data), decltype(__out_vals_data)`
Do we really need declare this types?
If we skip all of them, will they deduced automatically?

SergeyKopienko:
If we take a look at
`template <bool __is_ascending, ::std::uint8_t __radix_bits, ::std::uint16_t __data_per_work_item,
          ::std::uint16_t __work_group_size, typename _KeyT, typename _InputT, typename _OutputT>
struct __radix_sort_onesweep_slm_reorder_kernel`
and
`template <bool __is_ascending, ::std::uint8_t __radix_bits, ::std::uint16_t __data_per_work_item,
          ::std::uint16_t __work_group_size, typename _KeyT, typename _ValT, typename _InKeysT, typename _OutKeysT, typename _InValsT, typename _OutValsT>
struct __radix_sort_onesweep_by_key_slm_reorder_kernel`

we may see there are a lot of duplicated code too.
For example - the function
`    inline void
    __load_keys(::std::uint32_t __io_offset, __dpl_esimd_ns::simd<_KeyT, __data_per_work_item>& __keys, _KeyT __default_key) const` looks like the same. May be some other parts too.
May be we should think how to share this duplicated code parts?

test/kt/esimd_radix_sort_by_key.cpp
```diff
@@ -0,0 +1,128 @@
...
+        sycl::buffer<KeyT> keys(actual_keys.data(), actual_keys.size());
+        sycl::buffer<ValueT> values(actual_values.data(), actual_values.size());
+        oneapi::dpl::experimental::kt::esimd::radix_sort_by_key<isAscending, RadixBits>(
+            q, oneapi::dpl::begin(keys), oneapi::dpl::end(keys), oneapi::dpl::begin(values), param).wait();
+    }
+
+    auto expected_first  = oneapi::dpl::make_zip_iterator(std::begin(expected_keys), std::begin(expected_values));
+    std::stable_sort(expected_first, expected_first + size, CompareKey<isAscending>{});
+
+    std::string parameters_msg = ", n: " + std::to_string(size) +
```
SergeyKopienko, line 51: I believe better to use `std::stringstream` in this case and simple write to stream all required info through operator `<<`.
In this case we don't need make the calls like `std::to_string`.

test/kt/esimd_radix_sort_by_key.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+    oneapi::dpl::experimental::kt::esimd::radix_sort_by_key<isAscending, RadixBits>(
+        q, keys.get_data(), keys.get_data() + size, values.get_data(), param).wait();
+
+    std::vector<KeyT> actual_keys(size);
+    std::vector<ValueT> actual_values(size);
+    keys.retrieve_data(actual_keys.begin());
+    values.retrieve_data(actual_values.begin());
+
+    std::string parameters_msg = ", n: " + std::to_string(size) +
```
SergeyKopienko, line 86: The same.

test/kt/esimd_radix_sort_by_key.cpp
```diff
@@ -0,0 +1,128 @@
...
+
+int main()
+{
+    const std::vector<std::size_t> sizes = {
+        6, 16, 43, 256, 316, 2048, 5072, 8192, 14001, 1<<14,
+        (1<<14)+1, 50000, 67543, 100'000, 1<<17, 179'581, 250'000, 1<<18,
+        (1<<18)+1, 500'000, 888'235, 1'000'000, 1<<20, 10'000'000
+    };
+
+    for(auto size: sizes)
```
SergeyKopienko, line 108: I believe better to organize this test like esimd_radix_sort.cpp and specify all test params from CMakeLists.txt file through CMake params.
Otherwise we increate build time for this test and may don't understand which combination of test params exactly failed.

SergeyKopienko:
These includes exist in 3 files:
`
#if __has_include(<sycl/sycl.hpp>)
#    include <sycl/sycl.hpp>
#else
#    include <CL/sycl.hpp>
#endif
`
Are we able to use these includes only in one place?

SergeyKopienko:
As proposal: all this code exist inside two namespaces:
- `namespace oneapi::dpl::experimental::kt::esimd`;
- `namespace oneapi::dpl::experimental::kt::esimd::__impl`.
Can we move all files with code from `namespace oneapi::dpl::experimental::kt::esimd::__impl` into some `internal` folder?
Because each user of this algorithms will open the folder with the `esimd_radix_sort.h` file and read what is available in this file and near of them. And it's may be inconvenient to read our implementations details.

test/kt/esimd_radix_sort_by_key.cpp
dmitriy-sobolev: It's a good idea in long-term, but it's a pretty big chunk of work. Would you mind if we do it in a some follow-up PR? 

The current `esimd_radix_sort_by_key` test builds in ~3-5 minutes and it covers the most basic cases (I believe so). It essentially proves that key-value sort works.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_submitters.h
dmitriy-sobolev: Indeed, it is not required. This approach is derived from the existing key sort implementation. I preserved it with the intention of enhancing code quality synchronously for both implementation in a separate PR.

include/oneapi/dpl/experimental/kt/esimd_radix_sort_dispatchers.h
dmitriy-sobolev: Thanks for noticing it. It seems I forgot to replace `malloc_shared` with` malloc_device` after some debugging. The comment has to be uncommented:)

dmitriy-sobolev:
> These includes exist in 3 files: ` #if __has_include(<sycl/sycl.hpp>)
> 
> # include <sycl/sycl.hpp>
> #else
> 
> # include <CL/sycl.hpp>
> #endif ` Are we able to use these includes only in one place?

Sure, we can do it. We can include `sycl_defs.h`, instead. The drawback is bringing many non-used entities. Alternatively, we can define a separate header with this conditional include. I think these 3 quite plain repetitions do not worth it.

SergeyKopienko (approved):
LGTM

dmitriy-sobolev:
> If we take a look at `template <bool __is_ascending, ::std::uint8_t __radix_bits, ::std::uint16_t __data_per_work_item, ::std::uint16_t __work_group_size, typename _KeyT, typename _InputT, typename _OutputT> struct __radix_sort_onesweep_slm_reorder_kernel` and `template <bool __is_ascending, ::std::uint8_t __radix_bits, ::std::uint16_t __data_per_work_item, ::std::uint16_t __work_group_size, typename _KeyT, typename _ValT, typename _InKeysT, typename _OutKeysT, typename _InValsT, typename _OutValsT> struct __radix_sort_onesweep_by_key_slm_reorder_kernel`
> 
> we may see there are a lot of duplicated code too. For example - the function ` inline void __load_keys(::std::uint32_t __io_offset, __dpl_esimd_ns::simd<_KeyT, __data_per_work_item>& __keys, _KeyT __default_key) const` looks like the same. May be some other parts too. May be we should think how to share this duplicated code parts?

We have discussed it offline and came up with a conclusion that it can be done in a follow-up PR.

SergeyKopienko (approved):
LGTM

dmitriy-sobolev:
> As proposal: all this code exist inside two namespaces:
> 
> * `namespace oneapi::dpl::experimental::kt::esimd`;
> * `namespace oneapi::dpl::experimental::kt::esimd::__impl`.
>   Can we move all files with code from `namespace oneapi::dpl::experimental::kt::esimd::__impl` into some `internal` folder?
>   Because each user of this algorithms will open the folder with the `esimd_radix_sort.h` file and read what is available in this file and near of them. And it's may be inconvenient to read our implementations details.

It was discussed offline as well. It's a good idea to implement, but it is orthogonal to this PR. We agreed to do it later.

PR #1211, opened 2023-09-29
clang-format fix for push event

danhoeflinger:
Yet another fix for clang format post-merge.  This now gets the syntax for the `push` event commit hashes correct.

It also cleans up the code a little.

Here is the documentation:
https://docs.github.com/en/webhooks/webhook-events-and-payloads#push
This lists the structure of the event objects for `pull_request` and `push`, and supports the identifiers used here.

The link below supports the syntax to allow the code cleanup.  Specifically "Note: empty string returns 0." and their example for ternary below.
https://docs.github.com/en/actions/learn-github-actions/expressions#operators

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-          ref:  ${{ github.event.pull_request.head.sha }}
+          ref:  ${{ github.event.pull_request.head.sha || github.event.push.after }}
       - uses: actions/setup-python@v3
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -31,15 +31,8 @@ jobs:
       - name: Applying clang-format for changed files
         shell: bash
```
rarutyun, line 32: do we need this line? it doesn't look like that after the latest modification?

danhoeflinger: No, we shouldn't need it anymore.  Good catch.

rarutyun (approved):

PR #1212, opened 2023-10-02
Remove clang-format check from post-merge action

danhoeflinger:
This PR changes the clang-format check to only run on `pull_request` events (commits to an active pull request or creation of one).  The CI will skip the job for post-merge push events to main.

A couple previous PRs (#1207 #1211) attempted to fix the post-merge version of this clang-format job without success, after some operations were added in #1205 requiring specific git commit hashes which broke it post-merge.  With additional thought, it seems that this job is unnecessary in the context of the post-merge CI action.   In fact, the way it was written, before #1205 (and even before #715) it would just pass unconditionally post-merge because the diff would be empty.

The clang-format check is a tool to help us craft better PRs, so let us run it in that context, but not otherwise.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,7 @@ env:
 
 jobs:
   clang-format:
+    if: github.event_name == 'push'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


rarutyun (commented):
Looks good to me. Hope, it finally fixes what we need to fix

rarutyun (approved):

PR #1213, opened 2023-10-02
[KT] Move internal headers into internal folder

dmitriy-sobolev:
Remove unused on_wg header, uglify missed functions.


include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -11,7 +11,7 @@
 #define _ONEDPL_KT_ESIMD_RADIX_SORT_H
 
 #include "kernel_param.h"
-#include "esimd_radix_sort_dispatchers.h"
+#include "internal/esimd_radix_sort_dispatchers.h"
```
SergeyKopienko, line 14: I think better to include headers from `internal` folder at last place.

dmitriy-sobolev: Internal headers are now included last.

SergeyKopienko (approved):
LGTM

PR #1215, opened 2023-10-05
Fix compile error in test_internal_sycl_scheduler.pass with nightly compiler

adamfidel:
The test `test_internal_sycl_scheduler.pass` fails to compile with the nightly intel/llvm compiler because the `TEST_DYNAMIC_SELECTION_AVAILABLE` macro evaluates to false, but some of the test cases are defined outside of the if guard.

This PR moves the affected tests within the if macro.


dmitriy-sobolev (approved):
LGTM

julianmi (approved):

danhoeflinger (approved):

PR #1217, opened 2023-10-06
Fix `dpl` namespace name to `oneapi::dpl`

SergeyKopienko:
In this PR we fix `dpl` namespace name to `oneapi::dpl` to be consistent with other places in the code.


adamfidel (approved):
Thanks. LGTM.

PR #1218, opened 2023-10-06
Remove usm based `class __buffer_impl'

SergeyKopienko:
As discussed at https://github.com/oneapi-src/oneDPL/pull/1197#discussion_r1348717655
this `class __buffer_impl` based on USM memory is not required anymore as unused in oneDPL code and has been deleted in this PR:
```C++
// impl for USM pointer
template <typename _ExecutionPolicy, typename _T, typename _BValueT>
class __buffer_impl<_ExecutionPolicy, _T, _BValueT*>
{
    // ...
};
```

Also in this PR we simplify the definitions of `__internal::__buffer_impl` and `__buffer` [alias template](https://en.cppreference.com/w/cpp/language/type_alias):
```C++
// impl for sycl::buffer<...>
class __buffer_impl
{
  private:
    using _Container = sycl::buffer<_T>;
    using __container_t = typename __local_buffer<_Container>::type;

    // ....
};

using __buffer = __internal::__buffer_impl<::std::decay_t<_ExecutionPolicy>, _T>;
```
Please see [sycl::buffer](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:buffers) documentation for details.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -381,15 +381,12 @@ struct __local_buffer<sycl::buffer<::std::tuple<_T...>, __dim, _AllocT>>
     using type = sycl::buffer<oneapi::dpl::__internal::tuple<_T...>, __dim, _AllocT>;
 };
 
-template <typename _ExecutionPolicy, typename _T, typename _Container>
-class __buffer_impl;
-
 // impl for sycl::buffer<...>
-template <typename _ExecutionPolicy, typename _T, typename _BValueT, int __dim, typename _AllocT>
-class __buffer_impl<_ExecutionPolicy, _T, sycl::buffer<_BValueT, __dim, _AllocT>>
+template <typename _ExecutionPolicy, typename _Container>
+class __buffer_impl
 {
   private:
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -381,15 +381,12 @@ struct __local_buffer<sycl::buffer<::std::tuple<_T...>, __dim, _AllocT>>
     using type = sycl::buffer<oneapi::dpl::__internal::tuple<_T...>, __dim, _AllocT>;
 };
 
-template <typename _ExecutionPolicy, typename _T, typename _Container>
-class __buffer_impl;
-
 // impl for sycl::buffer<...>
-template <typename _ExecutionPolicy, typename _T, typename _BValueT, int __dim, typename _AllocT>
-class __buffer_impl<_ExecutionPolicy, _T, sycl::buffer<_BValueT, __dim, _AllocT>>
+template <typename _ExecutionPolicy, typename _Container>
```
MikeDvorskiy, line 385: It seems a general type `_Container` doesn't  make sense here, because it can be only `sycl::buffer`.
So, I would suggest removing `_Container` template parameter and use `sycl::buffer` directly in `__buffer_impl` class.

MikeDvorskiy (approved):
LGTM

PR #1222, opened 2023-10-10
Simplify condition in tests adding

Alexandr-Konovalov:
Make more clear that onedpl_add_test() is supposed be called no more than once on each iteration.


danhoeflinger (approved):
LGTM

PR #1223, opened 2023-10-12
Use SYCL_IMPLEMENTATION_INTEL macro

AidanBeltonS:
This PR updates how `_ONEDPL_SYCL_INTEL_COMPILER` is not defined for the open source DPC++ compiler this means that for `has_known_identity` incorrectly evaluates to false in all cases.

I propose using the `SYCL_IMPLEMENTATION_INTEL` macro instead of `_ONEDPL_SYCL_INTEL_COMPILER` which is defined in the SYCL specification and is used by the open source compiler. This change is not for performance reasons but instead having the expected and correct code path used on opensource DPC++.

### DIFF
--- a/include/oneapi/dpl/pstl/onedpl_config.h
+++ b/include/oneapi/dpl/pstl/onedpl_config.h
@@ -299,7 +299,7 @@
 
 // When compile with Intel(R) oneAPI DPC++ Compiler macro is on
 // We need this macro to use compiler specific implementation of SYCL
-#if defined(SYCL_IMPLEMENTATION_INTEL)
+#if defined(__INTEL_LLVM_COMPILER) && defined(SYCL_LANGUAGE_VERSION)
 #    define _ONEDPL_SYCL_INTEL_COMPILER 1
 #endif
 
--- a/include/oneapi/dpl/pstl/onedpl_config.h
+++ b/include/oneapi/dpl/pstl/onedpl_config.h
@@ -297,12 +297,6 @@
 
 #define _ONEDPL_BUILT_IN_STABLE_NAME_PRESENT __has_builtin(__builtin_sycl_unique_stable_name)
 
-// When compile with Intel(R) oneAPI DPC++ Compiler macro is on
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -299,7 +299,7 @@
 
 // When compile with Intel(R) oneAPI DPC++ Compiler macro is on
 // We need this macro to use compiler specific implementation of SYCL
```
dmitriy-sobolev, line 301: Let's remove the comment. The first line is not relevant anymore, the second one is not required due to the self-descriptive name of the macro (`SYCL_IMPLEMENTATION_INTEL`).

```suggestion
```

AidanBeltonS: Some from internal talks I decided that using the `SYCL_IMPLEMENTATION_INTEL` macro directly made more sense then changing this. Given this change, would you still like me to remove this comment?

AidanBeltonS:
I have updated this PRs approach. It instead swaps `_ONEDPL_SYCL_INTEL_COMPILER` with `SYCL_IMPLEMENTATION_INTEL` for the `has_known_identity` guards.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: Thanks, it is even better. With that change, I would recommend removing the comment and `_ONEDPL_SYCL_INTEL_COMPILER` macro definition altogether: this macro will have applications.

AidanBeltonS: Sounds good. Iv removed the macro and the comment

AidanBeltonS:
> I am ready to approve the PR. @AidanBeltonS, could you rebase your branch onto the latest commit of the main branch to resolve the conflicts?

Done

dmitriy-sobolev (approved):
LGTM

PR #1224, opened 2023-10-12
Add coalesced loads for reduce

AidanBeltonS:
This PR adds a commutative code path which performs non-sequential coalesced global loads. This provides large improvements in performance for memory bound reduce cases; specifically large problem sizes. We are guarding the implementation for Intel hardware due to some compiler issues around code-gen. 

For a reduce problem of size `33554432` we see the performance time drop from `349.819` to `190.436`.
The performance measurements were collected on A100. Performance results collected on PVC indicated there was no change, and the sequential code path is being used.


include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -226,6 +227,70 @@ struct transform_reduce
...
+                __res = __binary_op(__res, __unary_op(__adjusted_global_id + __stride * __i, __acc...));
+            __local_mem[__local_idx] = __res;
+        }
+    }
+    template <typename _NDItemId, typename _Size, typename _AccLocal, typename... _Acc>
+    inline void
+    operator()(const _NDItemId& __item_id, const _Size& __n, const _Size& __global_offset, const _AccLocal& __local_mem,
+               const _Acc&... __acc) const
+    {
+#ifndef __SPIR__
```
mmichel11, line 269: There is a oneDPL macro recently introduced called [_ONEDPL_DETECT_SPIRV_COMPILATION](https://github.com/oneapi-src/oneDPL/blob/0b14dbf530865a2305a5e0dcc8e26925548b71ff/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h#L61). Could this be used here?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -226,6 +227,70 @@ struct transform_reduce
...
+            _ONEDPL_PRAGMA_UNROLL
+            for (_Size __i = 1; __i < __iters_per_work_item; ++__i)
+                __res = __binary_op(__res, __unary_op(__adjusted_global_id + __stride * __i, __acc...));
+            __local_mem[__local_idx] = __res;
+        }
+        else if (__adjusted_global_id < __adjusted_n)
+        {
+            // TODO: simplify to not use floats
+            const _Size __items_to_process =
+                std::max(((__adjusted_n - __adjusted_global_id - 1) / __stride) + 1, static_cast<_Size>(0));
```
mmichel11, line 257: Can we use `__dpl_ceiling_div` here?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -226,6 +227,70 @@ struct transform_reduce
...
+    output_size(const _Size& __n, const ::std::uint16_t& __work_group_size) const
+    {
+#ifndef __SPIR__
+        if constexpr (_Commutative::value)
+        {
+            _Size __items_per_work_group = __work_group_size * __iters_per_work_item;
+            _Size __full_group_contrib = (__n / __items_per_work_group) * __work_group_size;
+            _Size __last_wg_remainder = __n % __items_per_work_group;
+
+            _Size __last_wg_contrib = std::min(__last_wg_remainder, static_cast<_Size>(__work_group_size));
```
mmichel11, line 287: `::std::` is the convention we use.

mmichel11 (commented):
Minor comments from my side.

mmichel11:
One thing I think it may be worth considering is how we ensure these implementations are tested. With the compilation target specific codepaths, testing becomes vendor specific, and I think we are at a greater risk of missing correctness issues due to skipping this codepath in our testing. I would imagine this risk grows with the more compilation target specific code we add.

Would it maybe be useful if we have some testing macros to ensure both the coalesced and non-coalesced implementations are tested for correctness in our reduce tests regardless of the value of `__SPIR__`?

julianmi (commented):
Is this PR ready for review yet? I don't think this patch is correct. At least `output_size()` should be used to correctly set `__n_items`.

I agree with @mmichel11 on the testing challenges of the additional code path. Further, disabling this for SPIR-V in general will disable coalescing also for architectures which might benefit from this such as integrated GPUs.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
AidanBeltonS: Thanks for pointing this macro out! I have used this instead.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
AidanBeltonS: In this case they are not equivalent, mainly if the numerator is cleanly divisible by the denominator we still want to go up 1 value. That does not happen with ceiling division. So I have kept the current maths.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
AidanBeltonS: Done

AidanBeltonS:
> Would it maybe be useful if we have some testing macros to ensure both the coalesced and non-coalesced implementations are tested for correctness in our reduce tests regardless of the value of `__SPIR__`?

This is a good point. I think that we could run the tests twice overriding the `_ONEDPL_DETECT_SPIRV_COMPILATION` macro so we test both branches. Would this be a suitable approach for you?

AidanBeltonS:
> Is this PR ready for review yet? I don't think this patch is correct. At least `output_size()` should be used to correctly set `__n_items`.

I have resolved the issue with `__n_items`. The PR is ready for review in my opinion. If you find any more issues please feel free to point them out.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -361,10 +365,17 @@ struct __parallel_transform_reduce_impl
                         auto __local_idx = __item_id.get_local_id(0);
                         auto __group_idx = __item_id.get_group(0);
                         // 1. Initialization (transform part). Fill local memory
+                        _Size __n_items;
                         if (__is_first)
+                        {
                             __transform_pattern1(__item_id, __n, /*global_offset*/ (_Size)0, __temp_local, __rngs...);
+                            __n_items = __transform_pattern1.output_size(__n, __work_group_size);
```
julianmi, line 372: Should we do this calculation on the host instead?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -226,6 +227,70 @@ struct transform_reduce
...
+        {
+            typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
+            _ONEDPL_PRAGMA_UNROLL
+            for (_Size __i = 1; __i < __iters_per_work_item; ++__i)
+                __res = __binary_op(__res, __unary_op(__adjusted_global_id + __stride * __i, __acc...));
+            __local_mem[__local_idx] = __res;
+        }
+        else if (__adjusted_global_id < __adjusted_n)
+        {
+            // TODO: simplify to not use floats
```
julianmi, line 255: I think this comment is outdated.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -49,14 +49,16 @@ class __reduce_kernel;
...
+__work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, _TransformPattern __transform_pattern,
+                           _ReducePattern __reduce_pattern, _InitType __init, const _AccLocal& __local_mem,
+                           const _Res& __res_acc, const _Acc&... __acc)
 {
     auto __local_idx = __item_id.get_local_id(0);
+    auto __group_size = __item_id.get_local_range().size();
     // 1. Initialization (transform part). Fill local memory
     __transform_pattern(__item_id, __n, /*global_offset*/ (_Size)0, __local_mem, __acc...);
     __dpl_sycl::__group_barrier(__item_id);
+    const _Size __n_items = __transform_pattern.output_size(__n, __group_size);
```
julianmi, line 61: Should we do this calculation on the host instead?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -70,15 +72,17 @@ __work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Si
...
+                       _ReducePattern __reduce_pattern, const _AccLocal& __local_mem, const _Tmp& __temp_acc,
+                       const _Acc&... __acc)
 {
     auto __local_idx = __item_id.get_local_id(0);
     auto __group_idx = __item_id.get_group(0);
+    auto __group_size = __item_id.get_local_range().size();
     // 1. Initialization (transform part). Fill local memory
     __transform_pattern(__item_id, __n, /*global_offset*/ (_Size)0, __local_mem, __acc...);
     __dpl_sycl::__group_barrier(__item_id);
+    const _Size __n_items = __transform_pattern.output_size(__n, __group_size);
```
julianmi, line 85: Should we do this calculation on the host instead?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -226,6 +227,70 @@ struct transform_reduce
             __local_mem[__local_idx] = __res;
         }
     }
+
+    template <typename _NDItemId, typename _Size, typename _AccLocal, typename... _Acc>
+    void
+    nonseq_impl(const _NDItemId __item_id, const _Size __n, const _Size __global_offset, const _AccLocal& __local_mem,
+                const _Acc&... __acc) const
+    {
+        auto __global_idx = __item_id.get_global_id(0);
```
julianmi, line 236: I think this is not used.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
AidanBeltonS: No, this must be in the device side as the `output_size` is calculated differently it the device code target is `SPIR`. This is only known during device compilation, and therefore must be in the kernel.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
AidanBeltonS: Thanks for catching this, it's removed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
AidanBeltonS: See above

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
AidanBeltonS: See above

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
AidanBeltonS: Good catch. Iv removed it

mmichel11:
> > Would it maybe be useful if we have some testing macros to ensure both the coalesced and non-coalesced implementations are tested for correctness in our reduce tests regardless of the value of `__SPIR__`?
> 
> This is a good point. I think that we could run the tests twice overriding the `_ONEDPL_DETECT_SPIRV_COMPILATION` macro so we test both branches. Would this be a suitable approach for you?

I think some approach similar to this is what we need to do. It may also be worth it to create some testing utilities to do this as future algorithm improvements may have this branching as well.

There are a few things that I think we should consider that may need a larger group discussion. Should we just include this in a handful of reduce tests or all that use a commutative reduction (ex: `reduce.pass`, `count.pass`, `transform_reduce.pass`, certain iterator tests, etc.)? I think we need to be careful about kernel naming in this scenario as we test with `-fsycl-no-unnamed-lambda`. Lastly, we will probably want to evaluate the impact of these testing changes on the test compile time and runtime to ensure the increase is acceptable.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: That makes sense, thanks.

AidanBeltonS:
@mmichel11 I have added a commit testing one approach to adding this dual testing form. This was the simplest way without messing around with the CMake I could think of. I appreciate that this does not add a testing utility for this problem. Let me know what you think about the solution, and if you think a different approach should be taken?

As for the tests I think it would be good to test all that use the commutative reduce. So that we make sure to have no failing operators on nvidia/AMD hardware

mmichel11:
> @mmichel11 I have added a commit testing one approach to adding this dual testing form. This was the simplest way without messing around with the CMake I could think of. I appreciate that this does not add a testing utility for this problem. Let me know what you think about the solution, and if you think a different approach should be taken?
> 
> As for the tests I think it would be good to test all that use the commutative reduce. So that we make sure to have no failing operators on nvidia/AMD hardware

I agree with your statement about adding something like this to all the algorithm tests that can take this codepath. These test additions look good to me, and I'm curious about other reviewers' thoughts here.

AidanBeltonS:
@mmichel11, just checking on the status of this review. Are you waiting for someone or any changes to the code?

mmichel11:
> @mmichel11, just checking on the status of this review. Are you waiting for someone or any changes to the code?

The proposed changes LGTM. I would wait for @julianmi for approval when he returns the following week since he works the most on reduce.

Also, since this PR sets the precedent for differentiating between backends would some other reviewers like to take a look? @MikeDvorskiy @rarutyun @akukanov

AidanBeltonS:
@julianmi, @MikeDvorskiy, @rarutyun, @akukanov what is the status of this review?

julianmi:
> @julianmi, @MikeDvorskiy, @rarutyun, @akukanov what is the status of this review?

I think the review comments so far have been addressed. There is a conflict due to the refactoring done in #1225 but this should be fixed by a rebase. @akukanov will review the general design of splitting the backends.

AidanBeltonS:
> > @julianmi, @MikeDvorskiy, @rarutyun, @akukanov what is the status of this review?
> 
> I think the review comments so far have been addressed. There is a conflict due to the refactoring done in #1225 but this should be fixed by a rebase. @akukanov will review the general design of splitting the backends.

I'v resolved the conflicts with a rebase

danhoeflinger (commented):
This is different from my initial position in the offline meeting, but here is my recommendation regarding the testing changes:

I believe that we should try to use cmake to generate separate tests with `_ONEDPL_DETECT_SPIRV_COMPILATION=[0,1]` for each of the tests affected here.  We have multiple tests which need this, and it may be a temporary bifurcation of the code if coalesced load performance is improved on some hardware.

https://github.com/oneapi-src/oneDPL/pull/1257/files/096b9c7e1926870ce421a698bcaa79cedd23e005#diff-0287b897807f0026622cd61ac4bb801dc9782f1c693a371d0deeeab481c3dcbe provides a good (much more complex) example of how this can be accomplished, as it generates from one source file, many options of template inputs.  Here we just need to toggle one preprocessor definition. 

Basically, I think we can put the tests which require both codepaths in a list, then when generating for tests in the list, we just use `target_compile_definitions()`, to define `_ONEDPL_DETECT_SPIRV_COMPILATION` in the desired way, and name the test accordingly.

This should have the following benefits:
1) Reduce large code movement to headers
2) Have one implementation of this bifurcation in cmake, repeated for all tests in the list
3) allow easy reuse for other tests which should be added
4) ability to easily remove these tests in the future, should codepaths be unified for all hardware without moving large blocks of code back.

test/CMakeLists.txt
```diff
@@ -170,6 +165,25 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
...
+    string(REPLACE "\.cpp" "" _test_name ${_test_name})
+
+    onedpl_construct_exec(${test_source_file} ${_test_name} ${switch_off_checked_iterators} "")
+
+    ##if (_test_name STREQUAL "reduce.pass")
+    set(coal_tests "reduce.pass" "transform_reduce.pass" "count.pass")
+
+    if (_test_name IN_LIST coal_tests)
+        onedpl_construct_exec(${test_source_file} ${_test_name}.coal ${switch_off_checked_iterators} "-D_ONEDPL_DETECT_SPIRV_COMPILATION=0")
+    endif()
```
danhoeflinger, line 184: Do we want to explicitly turn off (`-D_ONEDPL_DETECT_SPIRV_COMPILATION=1`) coalesced loads for the normal test path (or adjust the code specify each option for `coal_tests`, but not specify it across all tests)?  

If I understand correctly what is currently written, if we are running the tests on a non-SPIRV target, we will end up with 2 of the same test (named differently).  My initial thought was to have both versions always run regardless of the target.  
I understand that we may not care so much about the non-coalesced version on a non-SPIRV target, but the result as written may give a false impression of what the result of the tests are.

julianmi: I agree. Also, should we extend this testing approach to `sycl_iterator_reduce.pass` to catch errors during PR CI?

AidanBeltonS: I agree that that approach would be better. I have adjusted the CMake to build both versions regardless of target. I have also added `sycl_iterator_reduce.pass` to the list, let me know if the CI time increase is too large and ill remove it

julianmi: Thanks, @AidanBeltonS. Could you please rebase this PR to include #1319 which fixes the CI? We can then compare the CI times.

AidanBeltonS: Done

julianmi: Comparing the latest successful PR #1319 with this change, we see an accumulated increase in test time from about 3h to 3h 10m. The waiting times for developers won't change much with the slowest test (host, OpenMP backend, macOS) taking only 16 seconds more. The most affected are the DPCPP backend tests with 30-40 seconds increases with host offloading and 65 seconds increase on FPGA emulation. CI runtimes fluctuate a bit, so this is only an approximation. I think this increase is acceptable.

julianmi (approved):
I think all feedback has been addressed and there have been no objections against the small increase in CI times. Please wait for @danhoeflinger approval as well to see if he is fine with the test changes.

danhoeflinger (approved):
Testing LGTM. I did not look through the rest of the PR in detail.

PR #1225, opened 2023-10-12
Refactor USM host or buffer storage

julianmi:
This PR targets to refactor the USM host or buffer storage used as a return type for the reduction. This was introduced in #874. This simplifies the `__storage` and `__accessor` classes and moves them in the `__internal` namespace to align with other helpers. These changes are based on input and discussions with @SergeyKopienko.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -114,7 +114,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
         const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
 
         const bool __use_usm = __use_USM_host_allocations(__exec.queue());
-        __storage __res_container = __storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);
+        __storage<_ExecutionPolicy, _Tp> __res_container(__exec, __use_usm, 1);
 
         sycl::event __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
@@ -229,7 +229,7 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<_Tp, __work_group
         const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
 
         const bool __use_usm = __use_USM_host_allocations(__exec.queue());
-        __storage __res_container = __storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);
+        __storage<_ExecutionPolicy, _Tp> __res_container(__exec, __use_usm, 1);
 
         __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
             __cgh.depends_on(__reduce_event);
@@ -325,7 +325,7 @@ struct __parallel_transform_reduce_impl
         // Create temporary global buffers to store temporary values
         sycl::buffer<_Tp> __temp(sycl::range<1>(2 * __n_groups));
         const bool __use_usm = __use_USM_host_allocations(__exec.queue());
-        __storage __res_container = __storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);
+        __storage<_ExecutionPolicy, _Tp> __res_container(__exec, __use_usm, 1);
         // __is_first == true. Reduce over each work_group
         // __is_first == false. Reduce between work groups
         bool __is_first = true;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -114,7 +114,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
         const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
 
         const bool __use_usm = __use_USM_host_allocations(__exec.queue());
-        __storage<_ExecutionPolicy, _Tp> __res_container(__exec, __use_usm, 1);
+        __storage __res_container = __storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);
 
         sycl::event __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
@@ -129,7 +129,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
                 });
         });
 
-        return __future(__reduce_event, ::std::move(__res_container));
+        return __future(__reduce_event, __res_container);
     }
 }; // struct __parallel_transform_reduce_small_submitter
 
@@ -229,7 +229,7 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<_Tp, __work_group
         const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
 
         const bool __use_usm = __use_USM_host_allocations(__exec.queue());
-        __storage<_ExecutionPolicy, _Tp> __res_container(__exec, __use_usm, 1);
+        __storage __res_container = __storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);
 
         __reduce_event = __exec.queue().submit([&, __n, __n_items](sycl::handler& __cgh) {
             __cgh.depends_on(__reduce_event);
@@ -247,7 +247,7 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<_Tp, __work_group
                 });
         });
 
-        return __future(__reduce_event, ::std::move(__res_container));
+        return __future(__reduce_event, __res_container);
     }
 }; // struct __parallel_transform_reduce_work_group_kernel_submitter
 
@@ -325,7 +325,7 @@ struct __parallel_transform_reduce_impl
         // Create temporary global buffers to store temporary values
         sycl::buffer<_Tp> __temp(sycl::range<1>(2 * __n_groups));
         const bool __use_usm = __use_USM_host_allocations(__exec.queue());
-        __storage<_ExecutionPolicy, _Tp> __res_container(__exec, __use_usm, 1);
+        __storage __res_container = __storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);
         // __is_first == true. Reduce over each work_group
         // __is_first == false. Reduce between work groups
         bool __is_first = true;
@@ -388,7 +388,7 @@ struct __parallel_transform_reduce_impl
             __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
         } while (__n > 1);
 
-        return __future(__reduce_event, ::std::move(__res_container));
+        return __future(__reduce_event, __res_container);
     }
 }; // struct __parallel_transform_reduce_impl
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -113,9 +113,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
 
         const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
 
-        const bool __use_usm = __use_USM_host_allocations(__exec.queue());
-        __usm_host_or_buffer_storage __res_container =
-            __usm_host_or_buffer_storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -569,7 +566,7 @@ class __future : private std::tuple<_Args...>
 
     template <typename _ExecutionPolicy, typename _T>
     constexpr auto
-    __wait_and_get_value(__storage<_ExecutionPolicy, _T>& __buf)
+    __wait_and_get_value(__internal::__storage_impl<_ExecutionPolicy, _T>& __buf)
```
SergeyKopienko, line 569: This change is not required.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -435,71 +435,28 @@ struct __sycl_usm_alloc
     }
 };
 
-//-----------------------------------------------------------------------
```
SergeyKopienko, line 438: Can we avoid moving of this code?
I think it is a little bit difficult to understands your changes when this code were moved.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -435,71 +435,28 @@ struct __sycl_usm_alloc
...
-
-template <typename T>
-using __repacked_tuple_t = typename __repacked_tuple<T>::type;
-
-template <typename _ContainerOrIterable>
-using __value_t = typename __internal::__memobj_traits<_ContainerOrIterable>::value_type;
-
-template <typename _T>
-struct __accessor
+struct __accessor_impl
```
SergeyKopienko, line 439: Looks like it's not required to change the name of this class: we doesn't use in code `std::decay` for params of this class and doesn't create any alias for them. So I preferer to have old class name.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: I agree that this convolutes the review. However, the intention was to move them into the `__internal` namespace to align better with other helpers.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: I've reverted these changes.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: ok

SergeyKopienko (dismissed):
LGTM

MikeDvorskiy (dismissed):
Looks reasonable.
(And I didn't find wrong places, looking by my eyes)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -512,31 +469,29 @@ struct __accessor
...
     {
-        if (__usm)
-        {
-            __usm_buf = std::shared_ptr<_T>(
-                __internal::__sycl_usm_alloc<_ExecutionPolicy, _T, sycl::usm::alloc::host>{__exec}(__n),
-                __internal::__sycl_usm_free<_ExecutionPolicy, _T>{__exec});
-        }
-        else
-            __sycl_buf = ::std::make_shared<__sycl_buffer_t>(__sycl_buffer_t(__n));
     }
```
akukanov, line 489: This code looks ugly and overcomplicated even comparing with the current implementation. What is the reason to refactor it like this? Essentially, that only optimizes away a default ctor and a move assignment for a shared pointer - is it a big win, really?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -435,71 +435,28 @@ struct __sycl_usm_alloc
...
-        : __usm(__u)
+    __accessor(sycl::handler& __cgh, bool __u, sycl::buffer<_T, 1>* __sycl_buf, _T* __usm_buf)
+        : __usm(__u), __ptr(__u ? __usm_buf : nullptr),
+          __acc(__u ? decltype(__acc){} : sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{}))
     {
-        if (__usm)
-            __ptr = __usm_buf.get();
-        else
-            __acc = sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{});
     }
```
akukanov, line 456: Same question: is it worth to change the code like this for sake of a default ctor and assignment of sycl::accessor? The logic of the code is much more obfuscated this way.

akukanov (commented):
See detailed notes below, but shortly: I think some of the changes sacrifice code readability for no good reason. If there is sufficient justification (such as e.g. the performance improvement ratio) , please provide it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
akukanov, line 1: Two more things:
1. Essentially, both `__accessor` and `__storage` types are sort of unions; so, does it make sense to try either a union or `std::variant` in the implementation?
2. Are these types used independently or there is always a match between the storage kind (USM/buffer) and the access kind (USM pointer/accessor)? If the latter, is there some way to better correlate/coordinate the implementations?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -435,71 +435,28 @@ struct __sycl_usm_alloc
...
 {
   private:
     using __accessor_t = sycl::accessor<_T, 1, sycl::access::mode::read_write, __dpl_sycl::__target_device,
                                         sycl::access::placeholder::false_t>;
-    __accessor_t __acc;
-    _T* __ptr = nullptr;
+
     bool __usm = false;
+    _T* __ptr = nullptr;
+    __accessor_t __acc;
```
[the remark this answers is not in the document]
SergeyKopienko: As in previous code, initialization of `__usm` member is the first, so this member should be at first position inside class layout as far as we implement initialization of another members inside initialization list of class constructor.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: The goal: better to initialize some smart pointers through constructor - I believe it's should work faster then assignment.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -512,31 +469,29 @@ struct __accessor
...
-struct __storage
+class __storage_impl
 {
   private:
     using __sycl_buffer_t = sycl::buffer<_T, 1>;
-    ::std::shared_ptr<__sycl_buffer_t> __sycl_buf;
-    ::std::shared_ptr<_T> __usm_buf;
     bool __usm;
+    ::std::shared_ptr<_T> __usm_buf;
+    ::std::shared_ptr<__sycl_buffer_t> __sycl_buf;
```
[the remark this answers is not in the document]
SergeyKopienko: As in previous code, initialization of `__usm` member is the first, so this member should be at first position inside class layout as far as we implement initialization of another members inside initialization list of class constructor.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: The goal: better to initialize some smart pointers through constructor - I believe it's should work faster then assignment.

SergeyKopienko: This class looks like some common used functional inside `oneDPL` library so I think better to implement it as more effective.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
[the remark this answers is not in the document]
julianmi: This leads to a compilation error with "no viable conversion from returned value of type to function return type".

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: 1. I tried both unions or `std::variant` but they ended up not being device-copyable.
2. Storage and accessor are always the same kind.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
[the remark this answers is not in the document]
akukanov: To prevent warnings, you can change the order in the init list - this is OK because `__usm` is not used to initialize other members.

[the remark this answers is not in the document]
SergeyKopienko: Of course. it's possible.
But from my point of view, `__usm` is the `boss` inside this class, so better to initialize it at first and have it at first place in class fields.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: @akukanov I switched the ctor changes back for better readability based on your feedback. There is no measurable performance benefit of these changes. `__accessor` is now constructed based on the USM usage in `__storage`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -435,71 +435,28 @@ struct __sycl_usm_alloc
...
-        if (__usm)
-            __ptr = __usm_buf.get();
-        else
-            __acc = sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{});
     }
 #else
-    __accessor(sycl::handler& __cgh, bool, ::std::shared_ptr<sycl::buffer<_T, 1>> __sycl_buf,
-               ::std::shared_ptr<_T> __usm_buf)
-        : __usm(false), __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{}))
+    __accessor(sycl::handler& __cgh, bool, sycl::buffer<_T, 1>* __sycl_buf, _T* __usm_buf)
```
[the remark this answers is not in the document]
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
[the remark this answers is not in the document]
akukanov: Besides warnings, it sounds like there is no solid technical arguments justifying your preference to put `__usm` first in the class layout. On the other hand, the ordering before the patch follows the rule of ordering class members from most aligned to least aligned, as this usually minimizes paddings in the class layout, and so possibly reduces its size.

For these very classes, there might be no difference as of now. But since no strong justification to change the order is provided, I prefer to fix the warnings by changing the member initialization order instead.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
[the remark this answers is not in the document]
julianmi: Done.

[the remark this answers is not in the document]
julianmi: I agree. I've moved this helper.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -113,9 +113,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
 
         const _Size __n_items = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __iters_per_work_item);
 
-        const bool __use_usm = __use_USM_host_allocations(__exec.queue());
-        __usm_host_or_buffer_storage __res_container =
-            __usm_host_or_buffer_storage<_ExecutionPolicy, _Tp>(__exec, __use_usm, 1);
+        __usm_host_or_buffer_storage __res_container = __usm_host_or_buffer_storage<_ExecutionPolicy, _Tp>(__exec, 1);
```
SergeyKopienko, line 116: What about 
```C++
__usm_host_or_buffer_storage<_ExecutionPolicy, _Tp> __res_container(__exec, 1);
```
at this place and in all other places like this?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -533,7 +552,7 @@ struct __usm_host_or_buffer_storage
                      : __usm_host_or_buffer_accessor<_T>(__cgh, __sycl_buf.get());
     }
 
-    auto
+    _T
     __get_value(size_t idx = 0)
```
SergeyKopienko, line 556: Can we declare this function as `const` ?

julianmi: Done.

akukanov (approved):

PR #1226, opened 2023-10-12
Add missing synchronization for scan-copy-based algorithms

adamfidel:
Testing with the latest drivers exposes a bug in scan-copy-based algorithms (e.g., `copy_if`, `unique_copy`, `partition`, `set_intersection`, etc) where we do not properly wait on the event returned by the implementation before returning from the algorithm.

This issue was partially obscured because the calls to `__res.get()` were creating a `host_accessor` which acted as a quasi-synchronization, but this relied on undefined behavior and has been changed in newer drivers. The fix is to explicitly wait on the SYCL event before returning in the algorithm.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -849,6 +849,7 @@ __pattern_scan_copy(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __
         __par_backend_hetero::__parallel_scan_copy(::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(),
                                                    __buf2.all_view(), __n, __create_mask_op, __copy_by_mask_op);
 
+    __res.event().wait();
     ::std::size_t __num_copied = __res.get();
     return ::std::make_pair(__output_first + __n, __num_copied);
 }
@@ -873,6 +874,7 @@ __pattern_copy_if(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __la
     auto __res = __par_backend_hetero::__parallel_copy_if(::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(),
                                                           __buf2.all_view(), __n, __pred);
 
+    __res.event().wait();
     ::std::size_t __num_copied = __res.get();
     return __result_first + __num_copied;
 }
@@ -1638,7 +1640,7 @@ __pattern_hetero_set_op(_ExecutionPolicy&& __exec, _ForwardIterator1 __first1, _
     auto __keep3 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _OutputIterator>();
     auto __buf3 = __keep3(__result, __result + __n1);
 
-    auto __result_size =
+    auto __res =
         __par_backend_hetero::__parallel_transform_scan_base(
             ::std::forward<_ExecutionPolicy>(__exec),
             oneapi::dpl::__ranges::make_zip_view(
@@ -1655,8 +1657,10 @@ __pattern_hetero_set_op(_ExecutionPolicy&& __exec, _ForwardIterator1 __first1, _
                                   _Assigner, _DataAcc, _InitType>{__reduce_op, __get_data_op, _NoAssign{}, __assign_op,
                                                                   __get_data_op},
             // global scan
-            __copy_by_mask_op)
-            .get();
+            __copy_by_mask_op);
+
+    __res.event().wait();
+    auto __result_size = __res.get();
 
     return __result + __result_size;
 }


julianmi:
I don't see a disadvantage of explicitly waiting but why is the `host_accessor` call not blocking? The spec (https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_synchronization_in_the_sycl_application) explicitly states so:

Host Accessors: The constructor for a host accessor waits for all kernels that modify the same buffer (or image) in any queues to complete and then copies data back to host memory before the constructor returns. Any command groups with requirements to the same memory object cannot execute until the host accessor is destroyed as shown on [Table 6](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#fig:host-acc).

If this is indeed an issue, we might need to add explicit waits to other algorithms or the `__future.get()` method.

SergeyKopienko:
From my point of view as far as `__future` initialized by `sycl::event` in constructor we should wait this event in all cases.
Looks like this `wait()` call not executed in our `__future` now in some cases: this call depends on `ONEDPL_ALLOW_DEFERRED_WAITING` macro state but not only:
```C++
    constexpr auto
    __wait_and_get_value(__storage<_ExecutionPolicy, _T>& __buf)
    {
        // Explicit wait in case of USM memory. Buffer accessors are synchronous.
        if (__buf.__get_usm())
            wait();
        return __buf.__get_value();
    }
```
So I agree that we need add explicit waits to other algorithms or the __future.get() method.

SergeyKopienko (commented):
Please see my comment at https://github.com/oneapi-src/oneDPL/pull/1226#issuecomment-1761137873

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -849,6 +849,7 @@ __pattern_scan_copy(_ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __
         __par_backend_hetero::__parallel_scan_copy(::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(),
                                                    __buf2.all_view(), __n, __create_mask_op, __copy_by_mask_op);
 
+    __res.event().wait();
```
MikeDvorskiy, line 852: I think we don't need explicit wait, due to we call "get" below.
"get" does `get_host_access` in case of SYCL buffer and `sycl::event::wait_and_throw` in case USM.

`get_host_access` has implicit wait while a kernel is finalized:

SYCL spec tells:
```
As said in [Section 3.9.8](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:synchronization), the only blocking operations in SYCL (apart from explicit wait operations) are:

host accessor constructor, which waits for any kernels enqueued before its creation that write to the corresponding object to finish and be copied back to host memory before it starts processing. The host accessor does not necessarily copy back to the same host memory as initially given by the user;
```

MikeDvorskiy:
> From my point of view as far as `__future` initialized by `sycl::event` in constructor we should wait this event in all cases. Looks like this `wait()` call not executed in our `__future` now in some cases: this call depends on `ONEDPL_ALLOW_DEFERRED_WAITING` macro state but not only:
> 
> ```c++
>     template <typename _ExecutionPolicy, typename _T>
>     constexpr auto
>     __wait_and_get_value(__storage<_ExecutionPolicy, _T>& __buf)
>     {
>         // Explicit wait in case of USM memory. Buffer accessors are synchronous.
>         if (__buf.__get_usm())
>             wait();
>         return __buf.__get_value();
>     }
> ```
> 
> So I agree with @julianmi that probably we need to add explicit waits, but propose to implement this `wait()` call unconditionally inside `__future::get()` method. Or may be better to call `wait_and_throw()` to pass errors into async error handler [Table 34. Member functions for the event class](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_queue_interface)
> 
> **Wait for the event and the command associated with it to complete.**
> 
> Any unconsumed [asynchronous errors](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#async-error) from any context that the event was waiting on executions from will be passed to the [async_handler](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#async-handler) associated with the context. If no user defined async_handler is associated with the context, then an implementation-defined default [async_handler](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#async-handler) is called to handle any errors, as described in [Section 4.13.1.2](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsubsec:exception.nohandler).
> 
> Also may be we should have two methods in our `__future` class implementation - `get()` which make call `sycl::event::wait()` inside and `get_and_throw()` which make call `sycl::event::wait_and_throw()` inside. But in any case, `sycl::event` looks like the main sign which we should wait, I think.

I think "get_host_access" (host_accessor constructor) has implicit wait while a kernel is finalized and we don't need call explicit wait... especially since it leads to the "big" overheads (which we observed some time ago).

"As said in [Section 3.9.8](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:synchronization), the only blocking operations in SYCL (apart from explicit wait operations) are:

host accessor constructor, which waits for any kernels enqueued before its creation that write to the corresponding object to finish and be copied back to host memory before it starts processing. The host accessor does not necessarily copy back to the same host memory as initially given by the user;"

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
adamfidel: I believe you are right. The call to `get_host_access` seems like it should be sufficient synchronization in these cases.

However, there are test failures with the latest drivers/compiler where there appears to be a race when reading the USM output array. This means that perhaps the latest drivers have a bug that is causing the synchronization provided by `get_host_access` to not correctly work.

MikeDvorskiy: "....when reading the USM output array" - in that case we do (must do, as we designed, at least) explicit wait before get a value from a future object. We have an non-sycl::buffer overload of `__wait_and_get_value`:
https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h#L582C5-L582C25

Or, probably, we are wrong with macro definition `ONEDPL_ALLOW_DEFERRED_WAITING`.
BTW, do we still need one?

MikeDvorskiy: BTW 2,
'future' type has `wait()` methods as a part of contract (of public future API). we don't need call wait relying on 'event' API (which is returned by future::event() method)

adamfidel:
Closing this PR as I do not believe the synchronizations are necessary according to the SYCL spec and the original test failures have been resolved with a driver update.

PR #1233, opened 2023-10-19
Support proxy objects on iterator dereference for transform algorithms

rarutyun:
* Allow binary ``tranform_functor`` to take rvalues.
* Prohibit the assignment to fundamental types for both unary and binary `transform`.
* Unite unary and binary transform functors to one entity.

Solved problem: binary transform could not work with `zip_iterator` as an output.

Thanks to @Alexandr-Konovalov for spotting the problem.
Thanks to @MikeDvorskiy for the discussion.

If we like the design the corresponding tests should be added.
Update from @SergeyKopienko : four new tests has been added into this PR.

Update from @MikeDvorskiy: 
1) + the same fix (a proxy object support as output type, like tuple<ref, ref>) for transform_if, replace_if, replace_copy_if, also + usage of universal reference for input values as well.
2) + test coverage for transform_unary, transform_binary with zip iterator.

Update2 from @MikeDvorskiy: 
The changes (1) were reverted, added "todo":

- TODO 1 : to do the same fix  for output type (by re-using __transform_functor if applicable) for the other functor below:
__transform_if_unary_functor, __transform_if_binary_functor, __replace_functor, __replace_copy_functor
- TODO 2 : to make input type consistently: const T& or T&&; to think which way is preferable


include/oneapi/dpl/pstl/utils.h
```diff
@@ -70,19 +70,6 @@ __except_handler(_Fp __f) -> decltype(__f())
     }
 }
 
-template <typename _Op>
-struct __invoke_unary_op
-{
-    mutable _Op __op;
```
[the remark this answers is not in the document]
rarutyun: Yeah, I notice that we lose `mutable`. This is the most interesting question of this review (expect prohibiting the assignment to rvalue fundamentals)

To me, we should do it consistent. Either both binary and unary transformations support `mutable` or neither. I don't see how unary transformation is special with that regard.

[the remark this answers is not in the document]
danhoeflinger: I agree. 
When I was investigating the current state of oneDPL with regard to const-qualified `operator()` for function objects, this was the biggest inconsistency I found (binary vs unary overloads for `transform`).  I could not find any legitimate reason to justify it in the git history that I have access.  
In the long run, I would be in favor or dropping this and most other instances of `mutable` in oneDPL.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -279,13 +266,29 @@ class __transform_functor
     _Pred _M_pred;
 
   public:
-    explicit __transform_functor(_Pred __pred) : _M_pred(__pred) {}
+    explicit __transform_functor(_Pred __pred) : _M_pred(std::move(__pred)) {}
 
     template <typename _Input1Type, typename _Input2Type, typename _OutputType>
     void
-    operator()(const _Input1Type& x, const _Input2Type& y, _OutputType& z) const
+    operator()(const _Input1Type& x, const _Input2Type& y, _OutputType&& output) const
```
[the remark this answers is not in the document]
danhoeflinger: That makes sense, thanks.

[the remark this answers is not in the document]
rarutyun: Apparently, you are right. I changed it to forwarding reference because unary transform functor previously had this behavior. It's again inconsistency with binary transform that I don't understand. But what I read in the standard makes me think that unary transform behavior is correct.

include/oneapi/dpl/pstl/utils.h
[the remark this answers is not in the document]
danhoeflinger: Perhaps we should try to think of a way to "deprecate" this support for `mutable` in a way that can give the user a warning (if possible), and then eventually remove its support.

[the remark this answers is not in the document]
danhoeflinger: This is probably a task for another PR, and perhaps once we have access to c++20 which should make something like this easier via concepts / constraints.  Also, this is likely only relevant to deprecate with the `dpcpp` backend.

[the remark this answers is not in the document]
SergeyKopienko: > Made both `mutable`, because we have such a test for unary transform. I don't want to occasionally introduce a breaking change.

May we think about do not mark this member as `mutable` but apply some logic
inside `void __transform_impl(_OutputType&& output, _Args&&... args) const` :
- if `_M_pred` has operator() declared as `const` - simple call it;
- otherwise create a local copy of `_M_pred` inside `__transform_impl` function and call `operator()` on it...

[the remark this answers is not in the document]
danhoeflinger: This is an interesting idea, but we trade one issue for another.   This adds a requirement that the function object is CopyConstructible which I don't think is required by the standard.  Still, it's worth considering.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -276,16 +263,32 @@ class __not_equal_value
...
+    {
+        __transform_impl(::std::forward<_OutputType>(output), x);
+    }
+
+private:
+    template <typename _OutputType, typename... _Args>
+    void __transform_impl(_OutputType&& output, _Args&&... args) const
     {
-        z = _M_pred(x, y);
+        static_assert(sizeof...(_Args) < 3, "A predicate supports either unary or binary transformation");
```
adamfidel, line 289: Is this `static_assert` necessary? This is a private method that is only called within this class with either exactly 1 or 2 arguments.

include/oneapi/dpl/pstl/glue_algorithm_impl.h
```diff
@@ -349,8 +349,7 @@ transform(_ExecutionPolicy&& __exec, _ForwardIterator1 __first1, _ForwardIterato
 {
     return oneapi::dpl::__internal::__pattern_walk3(
         ::std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2, __result,
-        oneapi::dpl::__internal::__transform_functor<
-            oneapi::dpl::__internal::__ref_or_copy<_ExecutionPolicy, _BinaryOperation>>(__op),
+        oneapi::dpl::__internal::__transform_functor<_BinaryOperation>(std::move(__op)),
```
adamfidel, line 352:
```suggestion
        oneapi::dpl::__internal::__transform_functor<_BinaryOperation>(::std::move(__op)),
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -276,16 +263,32 @@ class __not_equal_value
 template <typename _Pred>
 class __transform_functor
 {
-    _Pred _M_pred;
+    mutable _Pred _M_pred;
 
   public:
-    explicit __transform_functor(_Pred __pred) : _M_pred(__pred) {}
+    explicit __transform_functor(_Pred __pred) : _M_pred(std::move(__pred)) {}
```
adamfidel, line 269:
```suggestion
    explicit __transform_functor(_Pred __pred) : _M_pred(::std::move(__pred)) {}
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -276,16 +263,32 @@ class __not_equal_value
...
+    }
+
+private:
+    template <typename _OutputType, typename... _Args>
+    void __transform_impl(_OutputType&& output, _Args&&... args) const
     {
-        z = _M_pred(x, y);
+        static_assert(sizeof...(_Args) < 3, "A predicate supports either unary or binary transformation");
+        static_assert(::std::is_invocable_v<_Pred, _Args...>, "A predicate cannot be called with the passed arguments");
+        ::std::forward<_OutputType>(output) = _M_pred(std::forward<_Args>(args)...);
```
adamfidel, line 291:
```suggestion
        ::std::forward<_OutputType>(output) = _M_pred(::std::forward<_Args>(args)...);
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -276,16 +263,32 @@ class __not_equal_value
...
+    operator()(const _Input1Type& x, const _Input2Type& y, _OutputType&& output) const
+    {
+        __transform_impl(::std::forward<_OutputType>(output), x, y);
+    }
+
+    template <typename _InputType, typename _OutputType>
+    void
+    operator()(_InputType&& x, _OutputType&& output) const
+    {
+        __transform_impl(::std::forward<_OutputType>(output), std::forward<_InputType>(x));
```
danhoeflinger, line 282:
```suggestion
        __transform_impl(::std::forward<_OutputType>(output), ::std::forward<_InputType>(x));
```

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
```diff
@@ -62,8 +62,7 @@ transform_async(_ExecutionPolicy&& __exec, _ForwardIterator1 __first1, _ForwardI
     wait_for_all(::std::forward<_Events>(__dependencies)...);
     auto ret_val = oneapi::dpl::__internal::__pattern_walk3_async(
         ::std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2, __result,
-        oneapi::dpl::__internal::__transform_functor<
-            oneapi::dpl::__internal::__ref_or_copy<_ExecutionPolicy, _BinaryOperation>>(__op));
+        oneapi::dpl::__internal::__transform_functor<_BinaryOperation>>(std::move(__op)));
```
danhoeflinger, line 65:
```suggestion
        oneapi::dpl::__internal::__transform_functor<_BinaryOperation>>(::std::move(__op)));
```

danhoeflinger (dismissed):
Overall, LGTM.

There are a few missing`::` before `std` which are noted in the review, and a couple clang-format suggestions I agree with (and one I don't).

The only non-cosmetic thing I want to point out is that we now are adding `mutable` to the functor used in `__pattern_walk3`.  I think this is the lesser of two "evils" because it fixes the inconsistency between `__pattern_walk2` and `__pattern_walk3`. However, we should be aware that this is the affect because it is difficult to remove the `mutable` after adding it because it means we may "break" user code which takes advantage of it.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -276,16 +263,32 @@ class __not_equal_value
...
     void
-    operator()(const _Input1Type& x, const _Input2Type& y, _OutputType& z) const
+    operator()(const _Input1Type& x, const _Input2Type& y, _OutputType&& output) const
+    {
+        __transform_impl(::std::forward<_OutputType>(output), x, y);
+    }
+
+    template <typename _InputType, typename _OutputType>
+    void
+    operator()(_InputType&& x, _OutputType&& output) const
```
MikeDvorskiy, line 280: 1) We do we have `_InputType&&` as input type while in the overload above we have `const _Input1Type& x` ?
2) What happens in case of rvalue type for output? How does a caller get a result?

SergeyKopienko: Let me try to answer to your question(2): please see my example at https://godbolt.org/z/cerxrvMT6

test/parallel_api/experimental/transform_async_binary_zip.pass.cpp
```diff
@@ -0,0 +1,181 @@
...
+template <typename T1, typename T2, typename T3>
+struct test_one_policy
+{
+    template <typename Policy, typename InputIterator1, typename InputIterator2, typename OutputIterator,
+              typename BinaryOp>
+    void
+    operator()(Policy&& exec, InputIterator1 first1, InputIterator1 last1, InputIterator2 first2, InputIterator2 /* last2 */,
+               OutputIterator out_first, OutputIterator /* out_last */, BinaryOp op)
+    {
+        auto result = oneapi::dpl::experimental::transform_async(exec, first1, last1, first2, out_first, op);
```
MikeDvorskiy, line 81: Do I understand correctly that we have just one difference with call `oneapi::dpl::experimental::transform_async` instead of `::std::transform` (`dpl::transform`) comparing with file `transform_binary_zip.pass.cpp` ?

test/parallel_api/experimental/transform_async_unary_zip.pass.cpp
```diff
@@ -0,0 +1,124 @@
...
+
+template <typename T1, typename T2>
+struct test_one_policy
+{
+    template <typename Policy, typename InputIterator, typename OutputIterator, typename UnaryOp>
+    void
+    operator()(Policy&& exec, InputIterator first, InputIterator last, OutputIterator out_first,
+               OutputIterator out_last, UnaryOp op)
+    {
+        auto result = oneapi::dpl::experimental::transform_async(exec, first, last, out_first, op);
```
MikeDvorskiy, line 57: Do I understand correctly that we have just one difference with call oneapi::dpl::experimental::transform_async instead of ::std::transform (dpl::transform) comparing with file `transform_unary_zip.pass.cpp` ?

MikeDvorskiy (changes requested):

test/parallel_api/algorithm/alg.modifying.operations/transform_binary_zip.pass.cpp
```diff
@@ -0,0 +1,168 @@
+// -*- C++ -*-
+//===-- transform_binary.pass.cpp -----------------------------------------===//
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/parallel_api/experimental/transform_async_binary_zip.pass.cpp
```diff
@@ -0,0 +1,181 @@
+// -*- C++ -*-
+//===-- transform_binary.pass.cpp -----------------------------------------===//
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/parallel_api/experimental/transform_async_unary_zip.pass.cpp
```diff
@@ -0,0 +1,124 @@
+// -*- C++ -*-
+//===-- transform_unary.pass.cpp ------------------------------------------===//
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/parallel_api/experimental/transform_async_binary_zip.pass.cpp
SergeyKopienko: No, it's incorrect understanding: also predicate in these new tests working with `std::tuple`

test/parallel_api/algorithm/alg.modifying.operations/transform_binary_zip.pass.cpp
```diff
@@ -0,0 +1,168 @@
...
+check_and_reset(InputIterator1 first1, InputIterator1 last1, InputIterator2 first2, OutputIterator out_first)
+{
+    typedef typename ::std::iterator_traits<OutputIterator>::value_type Out;
+    typename ::std::iterator_traits<OutputIterator>::difference_type k = 0;
+    for (; first1 != last1; ++first1, ++first2, ++out_first, ++k)
+    {
+        // check
+        Out expected = Out(1.5) + (::std::get<0>(*first1) + ::std::get<1>(*first1)) - (::std::get<0>(*first2) + ::std::get<1>(*first2));
+        Out actual = *out_first;
+        if (::std::is_floating_point_v<Out>)
```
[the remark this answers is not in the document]
MikeDvorskiy: I we have compile time condition we should use compile time checking  that condition, `constexpr if` for example.
Copying a bad approach is not good way to develop new code.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: With regards to (1).  I believe the original intention was to avoid enabling non-const lvalue references for inputs to the binary overload (which universal/ forwarding references would do).  

@rarutyun originally wrote the code to use const lvalue references explicitly for inputs for both binary and unary functors, but then changed it to support forwarding references for unary to support existing behavior. 
https://github.com/oneapi-src/oneDPL/pull/1233/commits/3e3c043e91f44a0a55b3f3ac2807a235707644ea

I think that was to support dereferenced `zip_iterators` forwarding along them as rvalues, but I am not sure if that is the reason (other than to not change the existing code).  Since this was changed intentionally, and aligns with the previous code, I believe it is very likely required.

test/parallel_api/algorithm/alg.modifying.operations/transform_binary_zip.pass.cpp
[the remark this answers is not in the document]
SergeyKopienko: It's not quite new code. It's copied / modified test. And our main goal in these new tests not math functions checks. We have another goal.

test/parallel_api/experimental/transform_async_binary_zip.pass.cpp
MikeDvorskiy: Please have a look at the "diff" for these files..
[diff_transform_binary_zip_transform_async_binary_zip.diff.txt](https://github.com/oneapi-src/oneDPL/files/13979375/diff_transform_binary_zip_transform_async_binary_zip.diff.txt)

Essentially, I would say there is no significant difference expecting "wait()" and test invocation  just for hetero policies in case of `transform_async`.

To tell the truth I have an objection here. I don't like huge code duplication, because:
1) A reviewer has to review the same code twice.
2) A reviewer has to leave a feedback at the same places twice.
3) An author has to fix the found error twice.
4) In case of refactoring/modification/improving the duplicated code, a developer has to don't  miss the duplicated places to modify ones as well, and has to  write the same new code twice.

If the author doesn't know how how to get rid of the code duplication, please let me know - I will help with that.
If the author  is strongly disagree with my point of view regarding the code duplication, let's  make a voting :) If I am in the minority, I will follow "disagree and commit".

test/parallel_api/algorithm/alg.modifying.operations/transform_binary_zip.pass.cpp
```diff
@@ -0,0 +1,168 @@
...
+        {
+            EXPECT_TRUE((expected > actual ? expected - actual : actual - expected) < Out(1e-7),
```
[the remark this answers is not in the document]
SergeyKopienko: Partially fixed.

test/parallel_api/algorithm/alg.modifying.operations/transform_binary_zip.pass.cpp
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/parallel_api/experimental/transform_async_binary_zip.pass.cpp
SergeyKopienko: @MikeDvorskiy could you please help with this point?

test/parallel_api/experimental/transform_async_unary_zip.pass.cpp
MikeDvorskiy: 1) As far a I see we use predicate with std::tuple on the both files: `transform_unary_zip.pass.cpp` and `transform_async_unary_zip.pass.cpp`...
2) Please read my opinion above (about the code duplication).

SergeyKopienko: Could you please prepare some PR with fix for that?

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: > but I am not sure if that is the reason

Yes, it was done to support dereferenced type for `zip_iterator`. It is a rvalue (like `tuple<ref, ref>`) which doesn't  match to lvalue type (`outputType&`).

test/parallel_api/experimental/transform_async_binary_zip.pass.cpp
MikeDvorskiy: > @MikeDvorskiy could you please help with this point?

sure, but let me to think a little bit..

test/parallel_api/algorithm/alg.modifying.operations/transform_binary_zip.pass.cpp
```diff
@@ -0,0 +1,168 @@
...
+public:
+
+    TheOperation(TOutput v) : val(v) {}
+
+    TOutput
+    operator()(const ::std::tuple<const TInput1&, const TInput1&>& x,
+               const ::std::tuple<const TInput2&, const TInput2&>& y) const
+    {
+        return TOutput(val + (::std::get<0>(x) + ::std::get<1>(x)) - (::std::get<0>(y) + ::std::get<1>(y)));
+    }
```
danhoeflinger, line 50: From the PR description:
"Solved problem: binary transform could not work with zip_iterator as an output."   

I think the original motivation for this PR is to allow `zip_iterators` to be used as output sequences for binary transform operations.  It seems from this test that we are testing `zip_iterator` inputs, but not `zip_iterator` outputs.  Am I missing that somewhere?

include/oneapi/dpl/pstl/utils.h
danhoeflinger: OK yes, since the dereference return is not a simple rvalue, but a proxy object, it wont just simply bind to `const T&` and work the way we would want.

However, with this in mind, I'm wondering how the current tests for `transform_binary_zip.pass` compile + pass with `zip_iterators` as inputs when the code only has const lvalue for binary inputs.  Would this same concept not prevent us from accepting `zip_iterator` inputs to binary transforms as it is currently written?

test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
```diff
@@ -47,9 +58,9 @@ check_and_reset(InputIterator1 first1, InputIterator1 last1, InputIterator2 firs
         // check
         Out expected = Out(1.5) + *first1 - *first2;
         Out actual = *out_first;
-        if (::std::is_floating_point_v<Out>)
+        if constexpr (::std::is_floating_point_v<Out>)
```
MikeDvorskiy, line 61: I looked  on that one more time... 
And I think that more right way to add a proper (floating point) specialization(s) for  `expect_equal_val`, see https://github.com/oneapi-src/oneDPL/blob/main/test/support/utils.h#L102C1-L102C17
If we do it we will not need any `if constexpr (::std::is_floating_point_v<Out>) ` here.

test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
```diff
@@ -47,9 +58,9 @@ check_and_reset(InputIterator1 first1, InputIterator1 last1, InputIterator2 firs
...
         {
-            EXPECT_TRUE((expected > actual ? expected - actual : actual - expected) < Out(1e-7),
+            EXPECT_TRUE((expected > actual ? expected - actual : actual - expected) < get_epsilon<Out>(),
```
MikeDvorskiy, line 63: In regard to two floating point numbers comparison please have a look
https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/

include/oneapi/dpl/pstl/glue_algorithm_impl.h
```diff
@@ -350,8 +350,7 @@ transform(_ExecutionPolicy&& __exec, _ForwardIterator1 __first1, _ForwardIterato
 {
```
danhoeflinger, line 350: There is a comment on line 344:
`// we can't use non-const __op here` 

which should be removed with this PR, as it is no longer matches the implementation.

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: Dan, It seems I don't catch what you mean...

danhoeflinger: It seems now that the rvalue inputs for unary transform was not to support `zip_iterators` but rather to support those ill-formed operators in the `binary_search` algorithms (as fixed by your recent commit).  This makes more sense now how the tests pass using `zip_iterator` inputs.

danhoeflinger: To go a little further and try to summarize (stating what might be obvious to others):

The derefenced `zip_iterators` return a rvalue ref of `tuple<const lvalue ref, const lvalue ref>` which will still bind to const lvalue ref input variables in `__transform_functor::operator()` .  However, for the output variable, an rvalue ref of `tuple<const lvalue ref, const lvalue ref>` with *not* bind to a non-const lvalue ref output variable.  This requires we use rvalue references for output variables, and forward them.  

Functionally, rvalue references would work for the input variables, but is more permissive than we would like to be, as it includes non-const lvalue references.  The implementation for the binary search algorithms was explicitly using a functor with lvalue references as input parameters, which was causing problems.

include/oneapi/dpl/pstl/glue_algorithm_impl.h
MikeDvorskiy: Removed

test/parallel_api/algorithm/alg.modifying.operations/transform_binary_zip.pass.cpp
MikeDvorskiy: good catch!
yes, new test cases should cover  a proxy object like tuple<ref, ref> as an output value.

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -75,9 +75,7 @@ lower_bound_impl(Policy&& policy, InputIterator1 start, InputIterator1 end, Inpu
                  InputIterator2 value_end, OutputIterator result, StrictWeakOrdering comp)
 {
     return oneapi::dpl::transform(policy, value_start, value_end, result,
-                                  [=](typename ::std::iterator_traits<InputIterator2>::reference val) {
-                                      return ::std::lower_bound(start, end, val, comp) - start;
-                                  });
+                                  [=](auto val) { return ::std::lower_bound(start, end, val, comp) - start; });
```
SergeyKopienko, line 78: Looks like it's not equivalent change.

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -87,9 +85,7 @@ upper_bound_impl(Policy&& policy, InputIterator1 start, InputIterator1 end, Inpu
                  InputIterator2 value_end, OutputIterator result, StrictWeakOrdering comp)
 {
     return oneapi::dpl::transform(policy, value_start, value_end, result,
-                                  [=](typename ::std::iterator_traits<InputIterator2>::reference val) {
-                                      return ::std::upper_bound(start, end, val, comp) - start;
-                                  });
+                                  [=](auto val) { return ::std::upper_bound(start, end, val, comp) - start; });
```
SergeyKopienko, line 88: Looks like it's not equivalent change to replace
`typename ::std::iterator_traits<InputIterator2>::reference val` to `auto`

include/oneapi/dpl/internal/binary_search_impl.h
danhoeflinger: As we discussed offline, perhaps `const auto&` is most appropriate to avoid a copy.  The same for all three of these.
I double checked, and see no legitimate reason this was originally written to use a non-const lvalue ref.

MikeDvorskiy (dismissed):

test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
MikeDvorskiy: in result, the overloads of `expect_equal_val` for floating point numbers were added.

SergeyKopienko (dismissed):
~~Still resolved not all questions.~~

test/support/utils.h
```diff
@@ -110,6 +110,32 @@ expect_equal_val(const T& expected, const T& actual, const char* file, std::int3
     }
 }
 
+template <typename T>
+void
+expect_equal_val_impl(const T& expected, const T& actual, const char* file, std::int32_t line, const char* message)
+{
+    const auto eps = std::numeric_limits<T>::epsilon();
+    if (std::fabs(expected - actual) >= eps)
```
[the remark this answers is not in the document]
MikeDvorskiy: yes.... it makes sense

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -99,9 +95,7 @@ binary_search_impl(Policy&& policy, InputIterator1 start, InputIterator1 end, In
                    InputIterator2 value_end, OutputIterator result, StrictWeakOrdering comp)
 {
     return oneapi::dpl::transform(policy, value_start, value_end, result,
-                                  [=](typename ::std::iterator_traits<InputIterator2>::reference val) {
-                                      return ::std::binary_search(start, end, val, comp);
-                                  });
+                                  [=](auto val) { return ::std::binary_search(start, end, val, comp); });
```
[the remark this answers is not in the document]
MikeDvorskiy: Actually, `binary_search_impl.h` was reverted.

include/oneapi/dpl/internal/binary_search_impl.h
MikeDvorskiy: Actually, binary_search_impl.h was reverted.

test/parallel_api/experimental/transform_async_unary_zip.pass.cpp
MikeDvorskiy: @SergeyKopienko, you may review the new commits within this PR.

test/parallel_api/experimental/transform_async_binary_zip.pass.cpp
MikeDvorskiy: @SergeyKopienko, you may review the new commits within this PR.

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: Actually, `binary_search_impl.h` was reverted.
At the moment the compiling of `binary_search_impl.h` doesn't lead any issues.

MikeDvorskiy:
@SergeyKopienko, as far as you requested changes, do you still any objections here?

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -340,7 +340,7 @@ __pattern_walk2(_ExecutionPolicy&& __exec, _ForwardIterator1 __first1, _ForwardI
         typedef typename ::std::iterator_traits<_ForwardIterator2>::reference _ReferenceType2;
 
         __par_backend::__parallel_for_each(::std::forward<_ExecutionPolicy>(__exec), __begin, __end,
-                                           [&__f](const ::std::tuple<_ReferenceType1, _ReferenceType2>& __val) {
+                                           [&__f](::std::tuple<_ReferenceType1, _ReferenceType2> __val) {
```
danhoeflinger, line 343: @MikeDvorskiy 

For my edification, why are we changing this parameter type to pass the tuple by value rather than const lvalue ref?

Is it required for compilation or merely a better way of handling zip iterator references?

I understand that the zip iterator reference is a tuple of references, but I'm not immediately seeing the advantage of using the tuple passed by value over the lvalue reference to a tuple.

rarutyun: Short answer - it doesn't compile with `const tuple&`.

Longer answer - if `tuple` is `const` it preserves constness for its objects. So when you have `zip_forward_iterator<..., zip_iterator>` and dereference it and bind it to `const&` you have `const tuple<_whatever_refs_, tuple<_refs_>>&`. Then, you try to read inner `tuple` from outer `tuple` and the only thing you can get is the `const tuple<_refs_>&` (unless you do copy but this is the different story). But this is `transform` algorithm and eventually you are trying to write to `const tuple<_refs_>&` which is obviously compile-time error.

For even longer answer, don't hesitate to reach me personally :)

danhoeflinger: Alright, I think I understand. This element is from an outer zip iterator, which zips together both inputs and outputs on the way to the functor.  Works OK with normal lvalue references as output, but when a zip iterator is present as the output, you end up with a const where you don't want one. 

I was confused because I was conflating the inner and outer zip iterators.
Thanks!

test/support/utils.h
```diff
@@ -97,11 +97,22 @@ expect(bool expected, bool condition, const char* file, std::int32_t line, const
...
+expect_equal_val(const T1& expected, const T2& actual, const char* file, std::int32_t line, const char* message)
 {
-    if (!(expected == actual))
+    using T = std::common_type_t<T1, T2>;
+
+    bool __is_error = false;
+    if constexpr (std::is_floating_point_v<T>)
+    {
+        const auto eps = std::numeric_limits<T>::epsilon();
+        __is_error = std::fabs(T(expected) - T(actual)) >= eps;
```
SergeyKopienko, line 110: As described at https://en.cppreference.com/w/c/numeric/math/fabs, `std::fabs` defined in header `<math.h>`.
Where is this include?

SergeyKopienko:
> @SergeyKopienko, as far as you requested changes, do you still any objections here?
I as understood, you already rolled back all problem from my point of view changes.

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: About `const` I understood and agree, thank you @rarutyun.
But why we remove reference `&` too?

MikeDvorskiy: We removed the reference too, because a tuple argument passed as rvalue  doesn't  match to lvalue (tuple<..>&).

SergeyKopienko: ok

test/support/utils.h
MikeDvorskiy: Actually, C++ version `std::fabs` is defined in `<cmath>`.   - https://en.cppreference.com/w/cpp/header/cmath
 (`math.h` - header of C- library with `fabs` function, https://en.cppreference.com/w/c/numeric/math/fabs)
 
 At the moment `<cmath>` is included via `<complex>` implicitly, https://godbolt.org/z/1cqT148YP
 But, yes, probably, we should include `<cmath>` explicitly, because I could not find any guarantee on `cppreference` about  implicit including via `<complex>`, mentioned above.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -273,19 +260,39 @@ class __not_equal_value
     }
 };
 
+//TODO: to do the same fix  for output type (by re-using __transform_functor if applicable) for the other functor below:
```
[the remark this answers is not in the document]
MikeDvorskiy: https://github.com/oneapi-src/oneDPL/issues/1369

SergeyKopienko (dismissed):
LGTM

test/support/utils.h
```diff
@@ -328,19 +339,59 @@ HashBits(size_t i, size_t bits)
...
     }
 };
 
+struct ComplementZip
+{
+    std::int32_t val = 1;
+
+    template<typename T>
+    auto
+    operator()(const oneapi::dpl::__internal::tuple<T&>& t) const
```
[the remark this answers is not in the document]
MikeDvorskiy: No, `oneapi::dpl::__internal::tuple<T&>` doesn't match to `std::tuple<T&>`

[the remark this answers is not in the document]
MikeDvorskiy: To hide the private API, probably, we may try to infer  a tuple type via decltype.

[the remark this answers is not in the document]
MikeDvorskiy: ... Or via writing more general code via SFINAE approach, like
`template <typename T, std::enable_if_t<__check_tuple<T>(0), int> = 0>`

test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
```diff
@@ -22,62 +22,80 @@
...
+    get_out_type(T val)
+    {
+        return val;
+    }
+
+    template <typename T>
+    constexpr auto
+    get_out_type(oneapi::dpl::__internal::tuple<T&>&& t)
+    {
+        return std::get<0>(t);;
```
danhoeflinger, line 92:
```suggestion
        return std::get<0>(t);
```

test/parallel_api/algorithm/alg.modifying.operations/transform_unary.pass.cpp
```diff
@@ -91,11 +117,13 @@ main()
     test<std::int32_t, std::int32_t>();
     test<std::int32_t, float32_t>();
     test<std::uint16_t, float32_t>();
-    test<float32_t, float64_t>();
```
danhoeflinger, line 94: Is there a reason to remove the float32->float64 test?  Just time?

test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
```diff
@@ -118,12 +136,12 @@ main()
     test<std::int32_t, float32_t, float32_t>(non_const(TheOperation<std::int32_t, float32_t, float32_t>(1.5)));
     test<std::int64_t, float64_t, float32_t>(non_const(TheOperation<std::int64_t, float64_t, float32_t>(1.5)));
 #endif
-    // lambda
-    //TODO: wrong value in output sentence with std::int8_t
-    //test<std::int8_t, float64_t, std::int8_t>([](const std::int8_t& x, const float64_t& y) { return std::int8_t(std::int8_t(1.5) + x - y); });
```
danhoeflinger, line 123: Did we determine it is not worth pursuing this TODO with `int8_t`?

test/parallel_api/algorithm/alg.modifying.operations/transform_unary.pass.cpp
```diff
@@ -53,9 +35,51 @@ struct test_one_policy
...
+            EXPECT_EQ(expected, actual, "wrong value in output sequence");
+            // reset
+            actual = k % 7 != 4 ? 7 * k - 5 : 0;
+        }
+    }
+    template <typename InputT>
+    auto
+    get_expected(InputT val)
+    {
+        return 1 - val;
```
danhoeflinger, line 59: optional nitpick:

Since we are hardcoding the "expected" calculation here to be a complement operation, should we rename the `UnaryComplementOp` and `_ComplementOp` below to make it clear that arbitrary unary functions aren't appropriate for use here?  These template arguments are merely to support different types of complement operations.

danhoeflinger (commented):
Only minor comments and questioning some seemingly unrelated removals.
Otherwise, LGTM.

I'm interested in the result of the private tuple API usage discussion, but I don't have a solution.  If it seems like the `std::tuple` conversion should help us here, should we add an issue to investigate why it is not working and use it as is?

MikeDvorskiy:
> but I don't have a solution
We can writing more general code via SFINAE approach  for `get_expected `and `get_actual`, like
`template <typename T, std::enable_if_t<__check_tuple<T>(0), int> = 0>`

test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
MikeDvorskiy: I have just removed the commented code. 
Obviously, we have got integer overflow for our test with "int8" type...  If we want to test such cases we have to write the test more accurately with checking resutl.

test/parallel_api/algorithm/alg.modifying.operations/transform_unary.pass.cpp
MikeDvorskiy: I think such case is redundant, because we already have cases with different  in/out types.

test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
MikeDvorskiy: It was unused code and it was removed.

test/parallel_api/algorithm/alg.modifying.operations/transform_unary.pass.cpp
```diff
@@ -53,9 +35,51 @@ struct test_one_policy
...
+    template <typename OutputT>
+    auto&
+    get_actual(OutputT& val)
+    {
+        return val;
+    }
+
+    template <typename T>
+    auto
+    get_expected(oneapi::dpl::__internal::tuple<T&>&& t)
```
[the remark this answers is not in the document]
MikeDvorskiy: updated

test/support/utils.h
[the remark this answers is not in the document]
MikeDvorskiy: done (by introducing one more abstraction)

test/parallel_api/algorithm/alg.modifying.operations/transform_unary.pass.cpp
MikeDvorskiy: I "joined" operation and get_expected/get_actual in one stuff.

danhoeflinger (approved):
LGTM, sorry I thought I ad re-approved this yesterday.

PR #1234, opened 2023-10-19
Dynamic selection tests to oneDPL format

egfefey:
This is a PR based on https://github.com/oneapi-src/oneDPL/pull/1138 
It separates out the dynamic_selection tests in addition to support/utils.h that is needed by the tests.


test/parallel_api/dynamic_selection/sycl/test_dynamic_load_policy_sycl.pass.cpp
```diff
@@ -53,37 +56,37 @@ main()
...
-    constexpr bool call_select_before_submit = true;
-    if (test_dl_initialization(u) || test_select<policy_t, decltype(u), decltype(f2)&, false>(u, f2) ||
-        test_submit_and_wait_on_event<just_call_submit, policy_t>(u, f2) ||
-        test_submit_and_wait_on_event<call_select_before_submit, policy_t>(u, f2) ||
-        test_submit_and_wait<just_call_submit, policy_t>(u, f2) ||
-        test_submit_and_wait<call_select_before_submit, policy_t>(u, f2) ||
-        test_submit_and_wait_on_group<just_call_submit, policy_t>(u, f) ||
-        test_submit_and_wait_on_group<call_select_before_submit, policy_t>(u, f)
+        auto actual = test_dl_initialization(u);
+        EXPECT_EQ(0, actual, "");
```
SergeyKopienko, line 71: Let's enumerate messages as "#1", "#2" and so far and so on. Or insert some text.

test/parallel_api/dynamic_selection/sycl/test_fixed_resource_policy_sycl.pass.cpp
```diff
@@ -7,56 +7,64 @@
...
 int
 main()
 {
+    bool bProcessed = false;
+
 #if TEST_DYNAMIC_SELECTION_AVAILABLE
     using policy_t = oneapi::dpl::experimental::fixed_resource_policy<oneapi::dpl::experimental::sycl_backend>;
     std::vector<sycl::queue> u;
     build_universe(u);
     std::cout << "UNIVERSE SIZE " << u.size() << std::endl;
```
SergeyKopienko, line 28: I believe we should avoid output into std::cout
Or if required we may do it under some define...
Please fix here and in all other places.

test/parallel_api/dynamic_selection/sycl/test_dynamic_load_policy_sycl.pass.cpp
```diff
@@ -7,10 +7,11 @@
 //
 //===----------------------------------------------------------------------===//
 
-#include <iostream>
 #include "oneapi/dpl/dynamic_selection"
+#include <iostream>
```
SergeyKopienko, line 11: I believe this include and `std::cout` output is not required

SergeyKopienko:
@egfefey could you please rebase your code from current `main` branch state? Looks like code based on too old `main` branch state.

SergeyKopienko:
I believe required mandatory to avoid constructions like `catch (...)`

SergeyKopienko:
Shy you are catch exceptions like 'catch (const sycl::exception&)` ?
You really throw exactly `sycl::exception` in dynamic selection code?

SergeyKopienko:
May be better to print exception text in catch blocks? Or it's obsoletely normal scenario?

test/parallel_api/dynamic_selection/sycl/test_dynamic_load_policy_sycl.pass.cpp
egfefey: The actual tests do already have the macros in them, so the additional macro checking the return code here is redundant. Those that do not have macros have a message and a return code.

test/parallel_api/dynamic_selection/sycl/test_fixed_resource_policy_sycl.pass.cpp
egfefey: This is actually just informational. I don't know if there are informational macros currently in oneDPL that can be used. I'll remove this one.

test/parallel_api/dynamic_selection/sycl/test_dynamic_load_policy_sycl.pass.cpp
egfefey: In this case, it is informational for devices that were skipped. I haven't seen an informational macro in oneDPL so can maintain it for now.

PR #1236, opened 2023-10-20
Replace `struct __conjunction` to `std::conjunction` as supported since C++17

SergeyKopienko:
In this PR we remove implementations of `struct __conjunction` from oneDPL code
and replace it' usage to [std::conjunction](https://en.cppreference.com/w/cpp/types/conjunction) (since C++17)

Looks like ourself implementation of `struct __conjunction` have required in oneDPL code when we supported C++11 and C++14 standards. But now our minimal supported standard is C++17.

### DIFF
--- a/include/oneapi/dpl/pstl/tuple_impl.h
+++ b/include/oneapi/dpl/pstl/tuple_impl.h
@@ -341,8 +341,8 @@ struct tuple<T1, T...>
     template <typename _U1, typename... _U,
               typename = ::std::enable_if_t<
                   (sizeof...(_U) == sizeof...(T) &&
-                   oneapi::dpl::__internal::__conjunction<::std::is_constructible<T1, _U1&&>,
-                                                          ::std::is_constructible<T, _U&&>...>::value)>>
+                   std::conjunction<::std::is_constructible<T1, _U1&&>,
+                                    ::std::is_constructible<T, _U&&>...>::value)>>
     tuple(_U1&& _value, _U&&... _next) : holder(::std::forward<_U1>(_value)), next(::std::forward<_U>(_next)...)
     {
     }
--- a/include/oneapi/dpl/pstl/tuple_impl.h
+++ b/include/oneapi/dpl/pstl/tuple_impl.h
@@ -341,7 +341,7 @@ struct tuple<T1, T...>
     template <typename _U1, typename... _U,
               typename = ::std::enable_if_t<
                   (sizeof...(_U) == sizeof...(T) &&
-                   std::conjunction_v<::std::is_constructible<T1, _U1&&>, ::std::is_constructible<T, _U&&>...>)>>
+                   ::std::conjunction_v<::std::is_constructible<T1, _U1&&>, ::std::is_constructible<T, _U&&>...>)>>
     tuple(_U1&& _value, _U&&... _next) : holder(::std::forward<_U1>(_value)), next(::std::forward<_U>(_next)...)
     {
     }
--- a/include/oneapi/dpl/pstl/tuple_impl.h
+++ b/include/oneapi/dpl/pstl/tuple_impl.h
@@ -341,8 +341,7 @@ struct tuple<T1, T...>
     template <typename _U1, typename... _U,
               typename = ::std::enable_if_t<
                   (sizeof...(_U) == sizeof...(T) &&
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -341,8 +341,7 @@ struct tuple<T1, T...>
     template <typename _U1, typename... _U,
               typename = ::std::enable_if_t<
                   (sizeof...(_U) == sizeof...(T) &&
-                   oneapi::dpl::__internal::__conjunction<::std::is_constructible<T1, _U1&&>,
-                                                          ::std::is_constructible<T, _U&&>...>::value)>>
+                   std::conjunction<::std::is_constructible<T1, _U1&&>, ::std::is_constructible<T, _U&&>...>::value)>>
```
MikeDvorskiy, line 344: If we are changing the lines.... lets use `std::is_constructible_v` as well.

akukanov: `std::conjunction` works with types, not values. It should be `conjunction_v<is_constructible<...>,...>`

include/oneapi/dpl/pstl/utils.h
akukanov, line 1: Should this patch be submitted to LLVM PSTL instead?

rarutyun:
The only thing I would check is that from what version of C++ standard library std::conjunction was implemented. For example does GNU 8 (libstdc++) have conjunction? Does libc++ from clang 7 (or whatever minimal version we support) have conjunction? The same is for Microsoft STL.

Other than that - looks good to me.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: @rarutyun I found some code about this:
https://github.com/llvm/llvm-project/blob/2a32afddf5594b7ca8c6aa67a5731612134f29f9/libcxx/include/__type_traits/conjunction.h#L43

SergeyKopienko:
GCC 8.5 example: https://godbolt.org/z/9aEbrb533
CLANG 7.1.0 example: https://godbolt.org/z/T9d9qdx4W

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -341,8 +341,7 @@ struct tuple<T1, T...>
     template <typename _U1, typename... _U,
               typename = ::std::enable_if_t<
                   (sizeof...(_U) == sizeof...(T) &&
-                   oneapi::dpl::__internal::__conjunction<::std::is_constructible<T1, _U1&&>,
-                                                          ::std::is_constructible<T, _U&&>...>::value)>>
+                   std::conjunction_v<::std::is_constructible<T1, _U1&&>, ::std::is_constructible<T, _U&&>...>)>>
```
kboyarinov, line 344: Since we only use conjunction value here, not an integral constant itself, may be it makes sense to simply use fold expression here:

```suggestion
                   ::std::is_constructible_v<T1, _U1&&> &&
                   ( ::std::is_constructible_v<T, _U&&> && ... )
```

SergeyKopienko: I propose to apply this change in separate PR.
Let's save the goal of this PR simple and clear.

rarutyun: Let's not use fold. `conjunction` is better in a sense of template instantiations. It stops instantiating the tail when meets first `false`, while fold-expression instantiates all the templates always.

rarutyun, line 344:
```suggestion
                   ::std::conjunction_v<::std::is_constructible<T1, _U1&&>, ::std::is_constructible<T, _U&&>...>)>>
```

SergeyKopienko: @rarutyun, @kboyarinov  so looks like the current state of this PR is ok?

rarutyun (approved):

include/oneapi/dpl/pstl/tuple_impl.h
kboyarinov: Agree, let's keep everything as-is.

kboyarinov (approved):

PR #1237, opened 2023-10-20
Replace `__void_type` to `std::void_t` as supported since C++17

SergeyKopienko:
In this PR we remove implementations of `struct __make_void_type` and `__void_type` from oneDPL code
and replace it' usage to [std::void_t](https://en.cppreference.com/w/cpp/types/void_t) (since C++17)

### DIFF
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -471,14 +471,6 @@ __pstl_left_bound(_Buffer& __a, _Index __first, _Index __last, const _Value& __v
 using __or_semantic = ::std::true_type;
 using __first_semantic = ::std::false_type;
 
-// Define __void_type via this structure to handle redefinition issue.
-// See CWG 1558 for information about it.
-template <typename... _Ts>
-struct __make_void_type
-{
-    using __type = void;
-};
-
 // is_callable_object
 template <typename _Tp, typename = void>
 struct __is_callable_object : ::std::false_type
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/utils.h
```diff
@@ -471,25 +471,14 @@ __pstl_left_bound(_Buffer& __a, _Index __first, _Index __last, const _Value& __v
...
-
 // is_callable_object
 template <typename _Tp, typename = void>
 struct __is_callable_object : ::std::false_type
 {
 };
 
 template <typename _Tp>
-struct __is_callable_object<_Tp, __void_type<decltype(&_Tp::operator())>> : ::std::true_type
+struct __is_callable_object<_Tp, ::std::void_t<decltype(&_Tp::operator())>> : ::std::true_type
```
MikeDvorskiy, line 481: probably we can use `std::is_invocable`  instead of?

SergeyKopienko: > probably we can use `std::is_invocable` instead of?

I propose to do this in some separate PR.
Let's change in this PR only `__void_type` staff.

SergeyKopienko:
@akukanov, @rarutyun, @MikeDvorskiy so what do you think about this PR? Is it OK ?

rarutyun (approved):
I guess we should not care about potential bug with `void_t` anymore. I hope all C++17 compliant compilers (that are in our support matrix) implement it right.

PR #1238, opened 2023-10-20
Implement `__is_callable_object` as `std::is_invocable` as supported since C++17

SergeyKopienko:
In this PR we reimplement `struct __is_callable_object` as [std::is_invocable](https://en.cppreference.com/w/cpp/types/is_invocable) :
Before:
```C++
// is_callable_object
template <typename _Tp, typename = void>
struct __is_callable_object : ::std::false_type
{
};

template <typename _Tp>
struct __is_callable_object<_Tp, __void_type<decltype(&_Tp::operator())>> : ::std::true_type
{
};
```

After:
```C++
// is_callable_object
template <typename _Tp, typename = void>
using __is_callable_object = ::std::is_invocable<_Tp>;
```

Please see an example at [godbolt.org](https://godbolt.org/z/WbjeY7xhr) to demonstrate how it's work.
Another example: https://godbolt.org/z/6Pvdffr8j

### DIFF
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -515,14 +515,7 @@ using __first_semantic = ::std::false_type;
 
 // is_callable_object
 template <typename _Tp, typename = void>
-struct __is_callable_object : ::std::false_type
-{
-};
-
-template <typename _Tp>
-struct __is_callable_object<_Tp, ::std::void_t<decltype(&_Tp::operator())>> : ::std::true_type
-{
-};
+using __is_callable_object = ::std::is_invocable<_Tp>;
 
 // is_pointer_to_const_member
 template <typename _Tp>


include/oneapi/dpl/pstl/utils.h
```diff
@@ -509,7 +498,7 @@ struct __is_pointer_to_const_member_impl<_R (_U::*)(_Args...) const noexcept> :
 {
 };
 
-template <typename _Tp, bool = __is_callable_object<_Tp>::value>
```
rarutyun, line 512: It's a different semantics. I was looking in this trait for quite a while. I don't like how it's written (the whole family), though it was my suggestion is the past.

Clearly, `std::is_invocable` is not the same as `__callable_object`. The first one is much much wider. Please postpone this change unless we discuss the desirable design.

SergeyKopienko: Converted to 'draft'.
@rarutyun could you please demonstrate described difference on some example?

include/oneapi/dpl/pstl/utils.h
```diff
@@ -471,27 +471,9 @@ __pstl_left_bound(_Buffer& __a, _Index __first, _Index __last, const _Value& __v
...
-// Define __void_type via this structure to handle redefinition issue.
-// See CWG 1558 for information about it.
-template <typename... _Ts>
-struct __make_void_type
-{
-    using __type = void;
-};
-
-template <typename... _Ts>
-using __void_type = typename __make_void_type<_Ts...>::__type;
```
MikeDvorskiy, line 483: It looks that `__void_type` is still used here:
https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/pstl/iterator_defs.h#L41

SergeyKopienko:
FYI: on current state (commit https://github.com/oneapi-src/oneDPL/pull/1238/commits/a50abc16139225fbaa7e5733993a19533d949295) all tests compiler and passed on Linux + icpx + GPU + debug

include/oneapi/dpl/pstl/utils.h
kboyarinov: @SergeyKopienko,

`std::is_invocable` relies on the well-formness of `INVOKE` expression (similar to `std::invoke`) that allows three possible types of callable objects:

* "Regular" callable objects (structures, lambdas, pointers to functions),
* Pointers to non-static functions
* Pointers to non-static variables

Consider [this example](https://godbolt.org/z/Gx89bK66c) for the last two usages.

SergeyKopienko: @kboyarinov but could you please describe the situations, for which we have the different behavior in this PR ?

kboyarinov: Sure, since you use `std::is_invocable_v` without providing any arguments, pointers to non-static functions/variables cannot affect the behavior.
But there can be a difference if the type provided is pointer to global function or static function. 
See [here](https://godbolt.org/z/MbjcP8xqb) for more details.

And since for `__is_callable_object` we want to check for exactly callable objects (structures or lambdas) - I guess we cannot use `std::is_invocable` here. 
I did not deeply investigate where __is_callable_object is used and can the logic be rewritten to avoid this trait.

SergeyKopienko: @kboyarinov Agree, as we can see in your example at https://godbolt.org/z/MbjcP8xqb, the behavior for pointers to global or static function really changed.
But I am sure that we should interpret this result as an error.

SergeyKopienko: OK, let's close this draft PR.

PR #1240, opened 2023-10-23
Fix C++20 compilation issues in tests

kboyarinov:
Fix minor C++20 compilation issues in newly added `dpl::move_iterator` tests and one PSTL offload test:

* Rename `phi_v` to `phi` in `numbers.pass.cpp` because `phi_v` requires a template argument. Move the test into `requires_header_presence` directory to enable CMake check for C++20 header file before compiling.
* Change `post_incr.pass.cpp` because starting from C++20, post increment operator in `std::move_iterator` returns `void` for an iterator that is not forward iterator.
* Add definition of dereference operator for `DummyIt` in types test - since C++20 to satisfy constraints for `std::iter_rvalue_reference_t` that is used to determine `std::move_iterator::reference` type.


SergeyKopienko (approved):
LGTM

PR #1241, opened 2023-10-24
Basic structure for kt::gpu::radix_sort

joeatodd:
This is the first PR for the pure SYCL onesweep radix_sort implementation.

I am adopting a similar PR strategy as `esimd-radix-sort`: incremental PRs into the feature branch `gpu-radix-sort`, followed by eventual merging into `main`.

This PR simply declares the `oneapi::dpl::experimental::kt::gpu::radix_sort` function signatures. It copies the equivalent declarations for `oneapi::dpl::experimental::kt::esimd::radix_sort`, though we omit `radix_sort_by_key` since this is not yet implemented.

PR #1243, opened 2023-10-25
Dispatcher decls for kt::gpu::radix_sort

joeatodd:
This is the 2nd PR for the pure SYCL onesweep radix_sort implementation.

This PR builds on top of (and depends on) #1241, adding dispatcher declarations and `radix_sort` implementations.


dmitriy-sobolev (approved):
LGTM

PR #1247, opened 2023-10-25
Fix oneDPL header inclusion before any other STL headers on GCC9

kboyarinov:
There is an issue in oneDPL config file that can lead to compilation issues on the user side if some oneDPL public headers is included before any STL header. E.g:

```

int main() { ... }
```

In the case above, `<oneapi/dpl/execution>` will include `oneapi/dpl/internal/common_config.h` firstly which defines 
`PSTL_USE_PARALLEL_POLICIES` according to the `_GLIBCXX_RELEASE` macro value. 
But the issue is when oneDPL is included before any other STL headers, `_GLIBCXX_RELEASE` is not yet defined so the value of `PSTL_USE_PARALLEL_POLICIES` would be always `true` that can result in known compilation issues in case of compiling by GCC9 with modern oneTBB.

This patch adds an inclusion of `<version>` into the common_config to be sure that `_GLIBCXX_RELEASE` is defined.
From the standard perspective, `<version>` is a C++20 header file but it is present in GCC9 and is not expected to include any other STL headers. I have wrapped it into `__has_include` to make sure that other configurations would not be affected.

### DIFF
--- a/include/oneapi/dpl/internal/common_config.h
+++ b/include/oneapi/dpl/internal/common_config.h
@@ -16,6 +16,10 @@
 #ifndef _ONEDPL_COMMON_CONFIG_H
 #define _ONEDPL_COMMON_CONFIG_H
 
+#if __has_include(<version>)
+#    include <version> // for _GLIBCXX_RELEASE
+#endif
+
 #if __cplusplus >= 201703L
 #    define _ONEDPL___cplusplus __cplusplus
 #elif defined(_MSVC_LANG) && _MSVC_LANG >= 201703L


SergeyKopienko:
May be required add some check that `_GLIBCXX_RELEASE` before line 34 too?

kboyarinov:
> May be required add some check that `_GLIBCXX_RELEASE` before line 34 too?

@SergeyKopienko
Actually we can't check the presence of `_GLIBCXX_RELEASE` macro because it can still be undefined for systems without glibcxx.

SergeyKopienko:
@kboyarinov sure, but I mean to add check if `_GLIBCXX_RELEASE ` is defined or no before check the value of `_GLIBCXX_RELEASE`.
Looks like that in other places we already have some checks:
https://github.com/oneapi-src/oneDPL/blob/8e5334da68f8351db4d4e36aa52506d14584bef0/include/oneapi/dpl/ranges#L23

include/oneapi/dpl/internal/common_config.h
```diff
@@ -16,6 +16,10 @@
 #ifndef _ONEDPL_COMMON_CONFIG_H
 #define _ONEDPL_COMMON_CONFIG_H
 
+#if __has_include(<version>)
+#    include <version> // for _GLIBCXX_RELEASE
```
dmitriy-sobolev, line 20: [`<version>`](https://github.com/gcc-mirror/gcc/blob/13c83c4cc679ad5383ed57f359e53e8d518b7842/libstdc%2B%2B-v3/include/std/version#L35) from GCC 9 includes [`<bits/c++config>`](https://github.com/gcc-mirror/gcc/blob/13c83c4cc679ad5383ed57f359e53e8d518b7842/libstdc%2B%2B-v3/include/bits/c%2B%2Bconfig#L683) which sets flags like `_GLIBCXX_USE_TBB_PAR_BACKEND`. Perhaps, this should be included after the workarounds below.

kboyarinov:
Close for now as more deeply investigation required

PR #1251, opened 2023-10-26
[KT] Remove histogram test

dmitriy-sobolev:
Histogram test validates internal API, which is not aligned with other oneDPL tests. Let's remove it as a part of the preparation of `esimd-radix-sort` branch for being reviewed/merged. The test will be moved into a separate branch because it might be useful for making histogram kernel configurable.


SergeyKopienko (approved):
LGTM

PR #1252, opened 2023-10-27
Add common_config fix to STD tests integration branch

kboyarinov:
Testing together STD tests integration and common_config fix


kboyarinov:
Close because testing is done

PR #1254, opened 2023-10-30
[oneDPL][hetero] Simplifying of temporary sycl::buffer support

MikeDvorskiy:
1. Simplified temporary sycl::buffer support in case of host iterators (when it needs to copy data to a temp buffer and back). 
2. Removed lifetime extension for temporary  sycl::buffer, which used within some SYCL backend patterns. According to the SYCL spec, SYCL runtime manages lifetime of such buffer itself.
3. Permutation_iterator support fixes.
4. A mention about a limitation in the doc: https://github.com/oneapi-src/oneDPL/pull/1295/files
5. Compiler time error for some non-supported iterator types (according to the documentation)
6.  is_passed_directly and is_hetero traits refactoring
7. Usage of exclusive access sycl::buffer with pointer to host data and data size for better performance. (Instead of sycl::buffer with iterators constructor - slow perf because of element-wise copy)

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
@@ -809,7 +809,7 @@ __parallel_radix_sort(_ExecutionPolicy&& __exec, _Range&& __in_rng, _Proj __proj
         // 'No operation' flag specifies whether to skip re-order phase if the all keys are the same (lie in one bin)
         const ::std::size_t __tmp_buf_size = __segments * __radix_states + __radix_states + 1 /*no_op flag*/;
         // memory for storing count and offset values
-        auto __tmp_buf = sycl::buffer<::std::uint32_t, 1>(sycl::range<1>(__tmp_buf_size));
+        sycl::buffer<::std::uint32_t, 1> __tmp_buf{sycl::range<1>(__tmp_buf_size)};
 
         // memory for storing values sorted for an iteration
         oneapi::dpl::__par_backend_hetero::__buffer<_ExecutionPolicy, _ValueT> __out_buffer_holder{__exec, __n};
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
@@ -835,7 +835,7 @@ __parallel_radix_sort(_ExecutionPolicy&& __exec, _Range&& __in_rng, _Proj __proj
         }
     }
 
-    return __future(__event, __tmp_buf, __val_buf);
+    return __future(__event);
 }
 
 } // namespace __par_backend_hetero
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
@@ -813,8 +813,7 @@ __parallel_radix_sort(_ExecutionPolicy&& __exec, _Range&& __in_rng, _Proj __proj
 
         // memory for storing values sorted for an iteration
         oneapi::dpl::__par_backend_hetero::__buffer<_ExecutionPolicy, _ValueT> __out_buffer_holder{__exec, __n};
-        auto __out_rng =
-            oneapi::dpl::__ranges::all_view<_ValueT, __par_backend_hetero::access_mode::read_write>(
+        auto __out_rng = oneapi::dpl::__ranges::all_view<_ValueT, __par_backend_hetero::access_mode::read_write>(
             __out_buffer_holder.get_buffer());
 
         // iterations per each bucket
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
@@ -755,9 +755,6 @@ __parallel_radix_sort(_ExecutionPolicy&& __exec, _Range&& __in_rng, _Proj __proj
     // radix bits represent number of processed bits in each value during one iteration
     constexpr ::std::uint32_t __radix_bits = 4;
 
-    //sycl::buffer doesn't have a default constructor; so, we have to pass zero-range to create an empty buffer
-    sycl::buffer<::std::uint32_t, 1> __tmp_buf(sycl::range<1>(0));
-    sycl::buffer<_ValueT, 1> __val_buf(sycl::range<1>(0));
     sycl::event __event{};
 
     const auto __max_wg_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
@@ -812,13 +809,13 @@ __parallel_radix_sort(_ExecutionPolicy&& __exec, _Range&& __in_rng, _Proj __proj
         // 'No operation' flag specifies whether to skip re-order phase if the all keys are the same (lie in one bin)
         const ::std::size_t __tmp_buf_size = __segments * __radix_states + __radix_states + 1 /*no_op flag*/;
         // memory for storing count and offset values
-        __tmp_buf = sycl::buffer<::std::uint32_t, 1>(sycl::range<1>(__tmp_buf_size));
+        auto __tmp_buf = sycl::buffer<::std::uint32_t, 1>(sycl::range<1>(__tmp_buf_size));
 
         // memory for storing values sorted for an iteration
         oneapi::dpl::__par_backend_hetero::__buffer<_ExecutionPolicy, _ValueT> __out_buffer_holder{__exec, __n};
-        __val_buf = __out_buffer_holder.get_buffer();
         auto __out_rng =
-            oneapi::dpl::__ranges::all_view<_ValueT, __par_backend_hetero::access_mode::read_write>(__val_buf);
+            oneapi::dpl::__ranges::all_view<_ValueT, __par_backend_hetero::access_mode::read_write>(
+            __out_buffer_holder.get_buffer());
 
         // iterations per each bucket
         assert("Number of iterations must be even" && __radix_iters % 2 == 0);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -232,7 +232,7 @@ struct is_passed_directly<zip_iterator<Iters...>>: ::std::conjunction<is_passed_
 };
 
 template<typename Iter>
-inline constexpr bool is_passed_directly_v = is_passed_directly<Iter>::value;
+inline constexpr bool is_passed_directly_v = is_passed_directly<::std::remove_const_t<Iter>>::value;
 
 // A trait for checking if iterator is heterogeneous or not
 
@@ -247,7 +247,7 @@ struct is_hetero<oneapi::dpl::__internal::sycl_iterator<Mode, Types...>> : ::std
 };
 
 template<typename Iter>
-inline constexpr bool is_hetero_v = is_hetero<Iter>::value;
+inline constexpr bool is_hetero_v = is_hetero<::std::remove_const_t<Iter>>::value;
 
 //A trait for checking if it needs to create a temporary SYCL buffer or not
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -243,27 +243,6 @@ struct is_hetero<oneapi::dpl::__internal::sycl_iterator<Mode, Types...>> : ::std
 {
 };
 
-template <typename Iter>
-struct is_hetero<::std::reverse_iterator<Iter>> : is_hetero<Iter>
-{
-};
-
-template <typename Iter, typename Unary>
-struct is_hetero<oneapi::dpl::transform_iterator<Iter, Unary>> : is_hetero<Iter>
-{
-};
-
-template <typename SourceIterator, typename IndexIterator>
-struct is_hetero<oneapi::dpl::permutation_iterator<SourceIterator, IndexIterator>> 
-    : ::std::conjunction<is_hetero<SourceIterator>, is_hetero<IndexIterator>>
-{
-};
-
-template <typename... Iters>
-struct is_hetero<zip_iterator<Iters...>>: ::std::conjunction<is_hetero<Iters>...>
-{
-};
-
 //A trait for checking if it needs to create a temporary SYCL buffer or not
 
 template <typename _Iter, typename Void = void>
@@ -408,7 +387,7 @@ struct __get_sycl_range
             AccMode == sycl::access::mode::read_write || AccMode == sycl::access::mode::write;
             
     //SFINAE iterator type checks
-    template<typename It, typename T = decltype(std::addressof(*It{}))>
+    template<typename It, typename T = decltype(std::addressof(*::std::declval<It&>()))>
     static constexpr std::true_type __test_addressof(int) { return {};}
     template<typename It>
     static constexpr std::false_type __test_addressof(...) { return {};}
@@ -519,6 +498,15 @@ struct __get_sycl_range
         auto rng = __get_permutation_view(__first.base(), __first.map(), __n);
 
         return __range_holder<decltype(rng)>{rng};
+    }
+
+    template <typename _Iter, typename _Map, ::std::enable_if_t<!is_hetero<_Iter>::value
+              && !is_passed_directly<_Iter>::value, int> = 0>
+    void
+    operator()(oneapi::dpl::permutation_iterator<_Iter, _Map>, oneapi::dpl::permutation_iterator<_Iter, _Map>)
+    {
+        static_assert(std::is_same_v<oneapi::dpl::permutation_iterator<_Iter, _Map>, void>,
+        "error: the iterator type is not suported with a hetero policy");
     }
 
     //specialization for permutation discard iterator
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -232,22 +232,22 @@ struct is_passed_directly<zip_iterator<Iters...>>: ::std::conjunction<is_passed_
 };
 
 template<typename Iter>
-inline constexpr bool is_passed_directly_v = is_passed_directly<::std::remove_const_t<Iter>>::value;
+inline constexpr bool is_passed_directly_v = is_passed_directly<Iter>::value;
 
 // A trait for checking if iterator is heterogeneous or not
 
 template <typename Iter>
-struct is_hetero : ::std::false_type
+struct is_sycl_iterator : ::std::false_type
 {
 };
 
 template <oneapi::dpl::access_mode Mode, typename... Types>
-struct is_hetero<oneapi::dpl::__internal::sycl_iterator<Mode, Types...>> : ::std::true_type
+struct is_sycl_iterator<oneapi::dpl::__internal::sycl_iterator<Mode, Types...>> : ::std::true_type
 {
 };
 
 template<typename Iter>
-inline constexpr bool is_hetero_v = is_hetero<::std::remove_const_t<Iter>>::value;
+inline constexpr bool is_sycl_iterator_v = is_sycl_iterator<Iter>::value;
 
 //A trait for checking if it needs to create a temporary SYCL buffer or not
 
@@ -258,7 +258,7 @@ struct is_temp_buff : ::std::false_type
 
 template <typename _Iter>
 struct is_temp_buff<
-    _Iter, ::std::enable_if_t<!is_hetero_v<_Iter> && !::std::is_pointer_v<_Iter> && !is_passed_directly_v<_Iter>>>
+    _Iter, ::std::enable_if_t<!is_sycl_iterator_v<_Iter> && !::std::is_pointer_v<_Iter> && !is_passed_directly_v<_Iter>>>
     : ::std::true_type
 {
 };
@@ -485,7 +485,7 @@ struct __get_sycl_range
 
   public:
     //specialization for permutation_iterator using sycl_iterator as source
-    template <typename _It, typename _Map, ::std::enable_if_t<is_hetero_v<_It>, int> = 0>
+    template <typename _It, typename _Map, ::std::enable_if_t<is_sycl_iterator_v<_It>, int> = 0>
     auto
     operator()(oneapi::dpl::permutation_iterator<_It, _Map> __first,
                oneapi::dpl::permutation_iterator<_It, _Map> __last)
@@ -502,7 +502,7 @@ struct __get_sycl_range
     }
 
     //specialization for permutation_iterator using USM pointer or direct pass object as source
-    template <typename _Iter, typename _Map, ::std::enable_if_t<!is_hetero_v<_Iter>
+    template <typename _Iter, typename _Map, ::std::enable_if_t<!is_sycl_iterator_v<_Iter>
               && is_passed_directly_v<_Iter>, int> = 0>
     auto
     operator()(oneapi::dpl::permutation_iterator<_Iter, _Map> __first,
@@ -518,7 +518,7 @@ struct __get_sycl_range
         return __range_holder<decltype(rng)>{rng};
     }
 
-    template <typename _Iter, typename _Map, ::std::enable_if_t<!is_hetero_v<_Iter>
+    template <typename _Iter, typename _Map, ::std::enable_if_t<!is_sycl_iterator_v<_Iter>
               && !is_passed_directly_v<_Iter>, int> = 0>
     void
     operator()(oneapi::dpl::permutation_iterator<_Iter, _Map>, oneapi::dpl::permutation_iterator<_Iter, _Map>)
@@ -555,7 +555,7 @@ struct __get_sycl_range
     template <typename _Iter>
     auto
     operator()(_Iter __first, _Iter __last)
-        -> ::std::enable_if_t<is_hetero_v<_Iter>,
+        -> ::std::enable_if_t<is_sycl_iterator_v<_Iter>,
                               __range_holder<oneapi::dpl::__ranges::all_view<val_t<_Iter>, AccMode>>>
     {
         assert(__first < __last);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -193,19 +193,57 @@ struct is_permutation<Iter, ::std::enable_if_t<Iter::is_permutation::value>> : :
 {
 };
 
-template <typename _Iter>
-using is_hetero_it = oneapi::dpl::__internal::is_hetero_iterator<_Iter>;
+//is_passed_directly trait definition; specializations for the oneDPL iterators
 
-template <typename _Iter>
-inline constexpr bool is_hetero_it_v = is_hetero_it<_Iter>::value;
+template <typename Iter>
+struct is_passed_directly : ::std::is_pointer<Iter>
+{
+};
 
-template <typename _Iter>
-using is_passed_directly_it = oneapi::dpl::__internal::is_passed_directly<_Iter>;
+template <typename Ip>
+struct is_passed_directly<oneapi::dpl::counting_iterator<Ip>> : ::std::true_type
+{
+};
 
-template <typename _Iter>
-inline constexpr bool is_passed_directly_it_v = is_passed_directly_it<_Iter>::value;
+template <>
+struct is_passed_directly<oneapi::dpl::discard_iterator> : ::std::true_type
+{
+};
 
-//struct for checking if it needs to create a temporary SYCL buffer or not
+template <typename Iter>
+struct is_passed_directly<::std::reverse_iterator<Iter>> : is_passed_directly<Iter>
+{
+};
+
+template <typename Iter, typename Unary>
+struct is_passed_directly<oneapi::dpl::transform_iterator<Iter, Unary>> : is_passed_directly<Iter>
+{
+};
+
+template <typename SourceIterator, typename IndexIterator>
+struct is_passed_directly<oneapi::dpl::permutation_iterator<SourceIterator, IndexIterator>> 
+    : ::std::conjunction<is_passed_directly<SourceIterator>, is_passed_directly<IndexIterator>>
+{
+};
+
+template <typename... Iters>
+struct is_passed_directly<zip_iterator<Iters...>>: ::std::conjunction<is_passed_directly<Iters>...>
+{
+};
+
+// A trait for checking if iterator is heterogeneous or not
+
+template <typename Iter>
+struct is_hetero : ::std::false_type
+{
+};
+
+template <oneapi::dpl::access_mode Mode, typename... Types>
+struct is_hetero<oneapi::dpl::__internal::sycl_iterator<Mode, Types...>> : ::std::true_type
+{
+};
+
+//A trait for checking if it needs to create a temporary SYCL buffer or not
 
 template <typename _Iter, typename Void = void>
 struct is_temp_buff : ::std::false_type
@@ -214,7 +252,7 @@ struct is_temp_buff : ::std::false_type
 
 template <typename _Iter>
 struct is_temp_buff<
-    _Iter, ::std::enable_if_t<!is_hetero_it_v<_Iter> && !::std::is_pointer_v<_Iter> && !is_passed_directly_it_v<_Iter>>>
+    _Iter, ::std::enable_if_t<!is_hetero<_Iter>::value && !::std::is_pointer_v<_Iter> && !is_passed_directly<_Iter>::value>>
     : ::std::true_type
 {
 };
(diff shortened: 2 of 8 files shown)


danhoeflinger (commented):
The main change here seems to be that we are removing the direct references to the `sycl::buffers` in the futures, in favor of relying on the reference counting in the actual async routines themselves which run the kernels.  Since we have references to accessors to buffers in the launch of those kernels, that should keep the buffers alive.  

If I understand correctly, one impact of this may be that the copy-back will be done when the async routine completes (and the last buffer reference count gets decremented), rather than when the future is waited on / destroyed.  This seems like a distinct benefit.  If this is correct, has this been confirmed via testing?

If I am understanding correctly, I like this change.  

I'm adding @SergeyKopienko as he has had interest in this topic of temporary buffer lifetime in the past.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -538,25 +511,28 @@ struct __get_sycl_range
...
+                                    : sycl::buffer<_T, 1>(__last - __first);
+        buf.set_write_back(__is_copy_back);
 
+        //a temp buffer lifetime extension
         using BufferType = oneapi::dpl::__internal::__lifetime_keeper<decltype(buf)>;
         m_buffers.push_back(::std::make_shared<BufferType>(buf));
 
-        return __buffer_holder<val_t<_Iter>, AccMode>{buf};
+        return __range_holder<oneapi::dpl::__ranges::all_view<_T, AccMode>>{
+            oneapi::dpl::__ranges::all_view<_T, AccMode>(buf)};
```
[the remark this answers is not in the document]
MikeDvorskiy: Actually, a buffer lifetime extension via `__buffer_holder` was redundant, because we have already extended the lifetime in 
`m_buffers.push_back(::std::make_shared<BufferType>(buf)); `line.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -538,25 +511,28 @@ struct __get_sycl_range
...
-        buf = copy_back(__first, buf, copy_back_tag());
+        auto buf = __is_copy_direct ? sycl::buffer<_T, 1>(std::addressof(*__first), __last - __first)
+                                    : sycl::buffer<_T, 1>(__last - __first);
+        buf.set_write_back(__is_copy_back);
 
+        //a temp buffer lifetime extension
         using BufferType = oneapi::dpl::__internal::__lifetime_keeper<decltype(buf)>;
         m_buffers.push_back(::std::make_shared<BufferType>(buf));
 
-        return __buffer_holder<val_t<_Iter>, AccMode>{buf};
```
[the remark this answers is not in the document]
MikeDvorskiy: sure

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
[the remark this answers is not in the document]
danhoeflinger: Well, that relies on the `__get_sycl_range` object staying in scope, which I don't think will be the case for the async APIs, or am I missing something?

SergeyKopienko:
For which reasons we still have now the definition of `sycl::buffer<_ValueT, 1> __val_buf(sycl::range<1>(0));` [here](https://github.com/oneapi-src/oneDPL/blob/bb878cdeb79794a7c7f607c985653f275172196c/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h#L760) ?

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -354,33 +354,6 @@ struct __get_sycl_range
     //We have to keep sycl buffer(s) instance here by sync reasons;
     ::std::vector<::std::shared_ptr<oneapi::dpl::__internal::__lifetime_keeper_base>> m_buffers;
```
SergeyKopienko, line 355: For which reasons we define `m_buffers` based on `std::shared_ptr` ?
Looks like it's enough to define through `std::unique_ptr`...

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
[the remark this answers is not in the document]
SergeyKopienko: Looks like we should delete this [struct](https://github.com/oneapi-src/oneDPL/blob/bb878cdeb79794a7c7f607c985653f275172196c/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h#L334) as unused anymore.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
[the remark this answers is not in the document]
danhoeflinger: Now that I think about it, if the `sycl::accessor` is enough to keep the buffer alive, do we gain anything from having `m_buffers` as a member variable in the first place?  Since the scope of the `struct __get_sycl_range` likely does not survive beyond the accessor its `operator()` returns, perhaps we don't gain anything from it.  

Of course, we do need to be sure that the `sycl::accessor` does in fact keep the buffer alive to consider this.

MikeDvorskiy:
> for which reasons we still have now the definition of `sycl::buffer<_ValueT, 1> __val_buf(sycl::range<1>(0));` [here](https://github.com/oneapi-src/oneDPL/blob/bb878cdeb79794a7c7f607c985653f275172196c/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h#L760) ?

No reasons any more. Yes, we may remove that definition.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
MikeDvorskiy: cannot call to mind why `std::shared_ptr`.... we don't share ownership if such buffer, just keep ones within __get_sycl_range lifetime.

SergeyKopienko: May you change it to `std::unique_ptr` or may be remove as proposed @danhoeflinger [here](https://github.com/oneapi-src/oneDPL/pull/1254#discussion_r1380057764)?

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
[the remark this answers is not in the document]
MikeDvorskiy: Regarding  "the async APIs" - on one of "Pass data..." tech meetings we came to conclusion that we will no support such case - "host data + Async oneDPL call", just USM or hetero iterators for async cases.

MikeDvorskiy:
> . Since we have references to accessors to buffers in the launch of those kernels, that should keep the buffers alive.

Unfortunately, SYCL Runtime takes ownership of buffer (as Runtime object, not C++ sycl::buffer instance) after bounding buffer with a command group - after sycl::handler::require() call.  
And SYCL Runtime manages just memory of "buffer" as Runtime object, not C++ sycl::buffer instance.

_"the SYCL system frees any storage required for the buffer asynchronously when it is no longer in use in queues."_

So, talking about auto-copying data back we can rely just on sycl::buffer destructor.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -812,13 +809,13 @@ __parallel_radix_sort(_ExecutionPolicy&& __exec, _Range&& __in_rng, _Proj __proj
         // 'No operation' flag specifies whether to skip re-order phase if the all keys are the same (lie in one bin)
         const ::std::size_t __tmp_buf_size = __segments * __radix_states + __radix_states + 1 /*no_op flag*/;
         // memory for storing count and offset values
-        __tmp_buf = sycl::buffer<::std::uint32_t, 1>(sycl::range<1>(__tmp_buf_size));
+        auto __tmp_buf = sycl::buffer<::std::uint32_t, 1>(sycl::range<1>(__tmp_buf_size));
```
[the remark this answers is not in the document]
MikeDvorskiy: Actually, we are using both ways everywhere in oneDPL.
With performance  perspective it is the same, I think.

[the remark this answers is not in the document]
SergeyKopienko: Sure. But in case when you know exactly the type I believe better to declare variable starting from type name. And it's short string, yes?

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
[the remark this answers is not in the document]
danhoeflinger: Aha, OK I understand the situation fully now, thanks for the clarification.

On the other hand ... Would we be better off to remove the `m_buffers` member variable, and instead keep `struct __buffer_holder` return to accomplish the lifetime extension?  

With `__get_sycl_range` it is quite unclear that this must be kept in scope for the lifetime of all kernels using the data.   However, what the `operator()` returns should stay in scope as it is what we use to represent the data.  In practice, perhaps they are the same because almost surely people will capture the return with `auto` within the scope of the `__get_sycl_range` object.  Switching is not a required change, but something to consider.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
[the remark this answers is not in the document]
MikeDvorskiy: "I believe better to declare variable starting from type name" - in given case it is a kind of code author preferable what is "better".

Line is simple, so we see type in both of cases immediately, IMHO.
Actually, we have no strong recommendations about that in our code guidelines. As far
So, it is not a bug and length difference in 4 chars is fit in 120 symbols. So, it is a kind of code author preferable and should be considered as "up to code author".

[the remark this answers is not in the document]
SergeyKopienko: At this point I disagree with you.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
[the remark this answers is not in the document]
MikeDvorskiy: "instead keep struct __buffer_holder".
For simple cases - yes... and there was such approach in the first version of the code.
But there are more complicated cases zip_iterator of zip_iterators of host iterators and hetero iteratos. In result we have got some hierarchical type - range. That range will be pass  to the SYCL backend and futher to kernel. We cannot keep sycl::buffer on that range. So, it range is based on placeholder accessors. For "flat" zip_iterator with may keep sycl::buffer(s) within  `__get_sycl_range`  lifetime as tuple f.e., but for hierarchical zip it is so complicated to do it compile time. As simpler solution  - was introduced storing buffers in RT via vector of smart pointers.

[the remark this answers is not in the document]
danhoeflinger: I don't think I understand.  Prior to this PR, we return a `__buffer_holder` from the host_iterator specialization of `operator()`, and a `__range_holder` from the other specializations.  Why cant we just just continue to do that, and then remove `m_buffer`, which is only currently used for the host_iterator specialization?

To me it seems that it is just two ways of removing a redundancy.  Again, I think it is preference, but I want to understand why that option may not be feasible.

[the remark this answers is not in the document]
MikeDvorskiy: We cannot "just continue" because we have lost sycl::buffer (__buffer_holder) life on this line (in case of zip):
`return oneapi::dpl::__ranges::make_zip_view(::std::get<_Ip>(tmp).all_view()...); `
https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h#L397

On this we have got __buffer_holder "on the fly", just in "a call scope" and immediately get an accessor (`oneapi::dpl::__ranges::all_view`)  by calling `all_view()`.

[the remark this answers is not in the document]
danhoeflinger: Alright, I think I understand now.  Thanks for the continued explanation.  

To restate to make sure I understand:  we are using this same `__get_sycl_range` object, to recursively create a `__buffer_holder` and then generate the views of the wrapped iterators.  The views live-on in the returned fancy iterator `__range_holder`, but the created `__buffer_holder` for each wrapped iterator gets lost as it leaves scope in the helper function  `gen_zip_view` (as an example with `zip_iterator`).

Alright this makes sense.  I think the way you have it is best.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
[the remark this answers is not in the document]
SergeyKopienko: // 88 chars
```C++
auto  __tmp_buf = sycl::buffer<::std:: uint32_t , 1 >(sycl::range< 1 >(__tmp_buf_size));
```

// 77 chars without auto and operator=
```C++
sycl::buffer<::std::uint32_t, 1> __tmp_buf(sycl::range< 1 >(__tmp_buf_size));
```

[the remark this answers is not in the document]
SergeyKopienko: Also discussed here: https://stackoverflow.com/questions/6900459/the-new-keyword-auto-when-should-it-be-used-to-declare-a-variable-type

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -354,33 +337,6 @@ struct __get_sycl_range
...
```
SergeyKopienko, line 338: Required to declare as `::std::vector<::std::unique_ptr...`

danhoeflinger (commented):
I notice that we are failing CI for the PSTL offload test, this is a new failure (not present in other active PRs) and relevant to the changes here.
https://github.com/oneapi-src/oneDPL/actions/runs/6783956403/job/18439268824?pr=1254

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -538,25 +511,28 @@ struct __get_sycl_range
...
+        using _T = val_t<_Iter>;
+
+        const bool __is_copy_direct = AccMode == sycl::access::mode::read_write || AccMode == sycl::access::mode::read;
+        const bool __is_copy_back = AccMode == sycl::access::mode::read_write || AccMode == sycl::access::mode::write;
 
-        auto buf = copy_direct(__first, __last, copy_direct_tag());
-        buf = copy_back(__first, buf, copy_back_tag());
+        auto buf = __is_copy_direct ? sycl::buffer<_T, 1>(std::addressof(*__first), __last - __first)
+                                    : sycl::buffer<_T, 1>(__last - __first);
+        buf.set_write_back(__is_copy_back);
```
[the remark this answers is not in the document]
MikeDvorskiy: 1. It looks the implementation is not compliance with the spec :-)
Data is not copied back until set_final_data is called.

2. Also, it seems this change does oneDPL behavior worse... ""When the buffer is destroyed, the destructor will block until all work in queues"...Before this change  sycl::buffer(first, last) is used and it doesn't have blocking on the destructor if a mode is just read.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -538,25 +494,42 @@ struct __get_sycl_range
...
+        constexpr bool __is_copy_direct =
+            AccMode == sycl::access::mode::read_write || AccMode == sycl::access::mode::read;
+        constexpr bool __is_copy_back =
+            AccMode == sycl::access::mode::read_write || AccMode == sycl::access::mode::write;
+
+        auto __get_buf = [&]()
+        {
+            if constexpr(__is_copy_direct)
+                //buffer will wait and copying on destructor
+                return sycl::buffer<_T, 1>(std::addressof(*__first), __last - __first);
```
danhoeflinger, line 515: We have avoided `std::addressof(*__first)` for input iterators in other places with the justification that we don't know really what the `operator*()` will return.  It could be a proxy object rvalue for instance.  
This overload is supposed to be for iterators to host data (or iterators to `std::vector`).  However, this is only detected by process of elimination. 

We discussed this last week and you mentioned that the initialization of `sycl::buffers` with iterators rather than a pointer to host memory will cause an extra host-to-host copy to occur, which can be avoided when initializing a buffer with a pointer to host memory.  

Are there implementations of `std::vector` for which this would not work?  I don't think we get any guarantees that the data range `host_vec.begin()` -> `host_vec.end()` is a contiguous block of memory.  However, in practice I think this is going to be the case. 

What about `host_vec.rbegin()`?  I think we "support" this case, but it seems reverse iterators to host vectors would be broken after this change.

While I don't want to incur the extra host-to-host copy, I think we may have to for correctness. We would need to rely on `set_final_data` and `set_write_back` as well if we use iterators to initialize the buffer.

MikeDvorskiy: > I don't think we get any guarantees that the data range `host_vec.begin()` -> `host_vec.end()` is a contiguous block of memory

Actually, for `std::vector `container we have such guarantee:
https://en.cppreference.com/w/cpp/container/vector
`The elements are stored contiguously, which means that elements can be accessed not only through iterators, but also using offsets to regular pointers to elements. This means that a pointer to an element of a vector may be passed to any function that expects a pointer to an element of an array.`

danhoeflinger: > So, in my opinion, an ideal solution here - if `std::addressof(*__first)` is valid, we use exclusive mem access buffer `sycl::buffer(ptr,N)`, if is not - we do fallback to `sycl::buffer(It1, it2)` usage.

I agree, I think that would be ideal.  Is it possible to properly determine if `std::addressof` is valid?

MikeDvorskiy: > Is it possible to properly determine if `std::addressof` is valid?

yes, it is possible via SFINAE approach.
Additionally, we have to check "contiguous" property and to have custom code for reverse iterator

danhoeflinger: Ok, that makes sense.  Since `std::addressof` deletes the rvalue call, that takes care of those, then if we check contiguity and specialize for reverse iterator, I think that will be sufficient. 

Since we don't have c++20, but we do know that `std::addressof` is valid, for contiguity we can just check confirm that the addresses are the same distance away as the iterators, and perhaps an incremented iterator's address is 1 away from itself or something like that.

MikeDvorskiy: > an incremented iterator's address is 1 away from itself or something like that.

I thought about such check... but it doesn't guarantees contiguity in general..

danhoeflinger: You are correct, but without c++20 concepts I'm not seeing a robust to confirm this.  If both the full range is the appropriate distance and a single increment is then I would feel pretty good about it.  Of course, you could design an iterator to thwart this, but it does not seem likely (and perhaps we don't care about such cases because they would not be officially supported by our specification / documentation).

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -468,22 +500,18 @@ struct __get_sycl_range
         return __range_holder<decltype(rng)>{rng};
     }
 
-    // TODO Add specialization for general case, e.g., permutation_iterator using host
-    // or another fancy iterator.
-    //specialization for permutation_iterator using USM pointer as source
-    template <typename _It, typename _Map, ::std::enable_if_t<!is_hetero_it_v<_It>, int> = 0>
+    //specialization for permutation_iterator using USM pointer or direct pass object as source
+    template <typename _Iter, typename _Map, ::std::enable_if_t<!is_hetero<_Iter>::value
+              && is_passed_directly<_Iter>::value, int> = 0>
```
danhoeflinger, line 505: This is getting really "exotic" but what if the there is a `zip_iterator` or `permutation_iterator` that is split between 
a `is_hetero_it` and a `is_passed_directly_it`.  This should be a "valid" input, but will return false for both checks.

I think this is only important for `permutation_iterator(split_iterator, valid_input)` because it special case

To be more explicit:
`permutation_iterator(zip_iter(hetero_it,passed_directly_it), valid_input)`
`permutation_iterator(zip_iter(passed_directly_it, hetero_it), valid_input)`
`permutation_iterator(permutation_iterator(passed_directly_it, hetero_it), valid_input)`
`permutation_iterator(permutation_iterator(hetero_it, passed_directly_it), valid_input)`

I think none of these will find an overload that fits it.  

We could create a third trait `is_hetero_or_passed_directly`, which has the appropriate logic for these split iterators.  However, I don't know if it is possible to support such a case with another `permutation_iterator` overload.  I think you run into the problem of not knowing the size of the range of the split iterator that is needed, which is problematic when it is split.

I'd recommend just adding this as a known limitation perhaps.

danhoeflinger (commented):
I think it would be really helpful to add comments at it "recursive call"  of `this->operator()`, to indicate clearly where this recursion exists.  This is where the largest source of complexity and confusion is as I try to understand this code.

Other than the exotic case of the permutation_iterator of a split_iterator, I think this should work.  Of course, the tests added in #1293 will be good to check it.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
MikeDvorskiy: Yes, it seems there is no way to support  `permutation_iterator(non_passed_directly, valid_input)`. 
Regarding  "adding  known limitation perhaps" - in that case we have to also explain what is "passed directly"... But I am not sire that we want to do that.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -468,32 +485,36 @@ struct __get_sycl_range
...
     {
         auto __n = __last - __first;
         assert(__n > 0);
 
-        // The size of the source range is unknown. Use non-zero size to create the view.
-        // permutation_view_simple access is controlled by the map range view.
-        auto res_src = this->operator()(__first.base(), __first.base() + 1 /*source size*/);
-
-        auto rng = __get_permutation_view(res_src.all_view(), __first.map(), __n);
+        // The size of the source range is unknown. So, just base iterator is passing to permutation_view
```
danhoeflinger, line 498:
```suggestion
        // The size of the source range is unknown. So, just base iterator is passing to permutation_view
        //   _Map is handled by recursively calling __get_sycl_range() in __get_permutation_view.
```

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -452,7 +469,7 @@ struct __get_sycl_range
 
   public:
     //specialization for permutation_iterator using sycl_iterator as source
-    template <typename _It, typename _Map, ::std::enable_if_t<is_hetero_it_v<_It>, int> = 0>
+    template <typename _It, typename _Map, ::std::enable_if_t<is_hetero_v<_It>, int> = 0>
     auto
     operator()(oneapi::dpl::permutation_iterator<_It, _Map> __first,
                oneapi::dpl::permutation_iterator<_It, _Map> __last)
```
danhoeflinger, line 475: Can we add a comment before the `__get_permutation_view()` call to mention that it handles the map by recursively calling `__get_sycl_range()`?

Without such comment, it looks like this function only is concerned with the base, and not the map, and it looks like an mistake (it is not).

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -391,7 +402,6 @@ struct __get_sycl_range
     template <typename _TupleType, typename _DiffType, ::std::size_t... _Ip>
     auto
     gen_zip_view(_TupleType __t, _DiffType __n, ::std::index_sequence<_Ip...>)
-        -> decltype(oneapi::dpl::__ranges::make_zip_view(gen_view(*this, ::std::get<_Ip>(__t), __n).all_view()...))
     {
         auto tmp = oneapi::dpl::__internal::make_tuple(gen_view(*this, ::std::get<_Ip>(__t), __n)...);
```
danhoeflinger, line 406:
```suggestion
        // Send each zipped iterator to `gen_view` which recursively calls __get_sycl_range() to process them.
        auto tmp = oneapi::dpl::__internal::make_tuple(gen_view(*this, ::std::get<_Ip>(__t), __n)...);
```

danhoeflinger (commented):
Only superficial changes left, but it should be run through expanded tests considering the complexity and importance.

clang-format has some suggestions which should be taken.

I've added some suggested comments, other comments that might be nice would be the following:
1. at recursive call for transform_iterator overload
2. within `__get_permutation_view` stating that the base iterator has already been "handled", but that the map is recursed over during this call (for random access iterator version)

rarutyun (commented):
This is the first portion of the comments. As I can see some of them have been already addressed.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -187,19 +187,78 @@ struct is_permutation<Iter, ::std::enable_if_t<Iter::is_permutation::value>> : :
...
-using is_passed_directly_it = oneapi::dpl::__internal::is_passed_directly<_Iter>;
+template <typename Ip>
+struct is_passed_directly<oneapi::dpl::counting_iterator<Ip>> : ::std::true_type
+{
+};
 
-template <typename _Iter>
-inline constexpr bool is_passed_directly_it_v = is_passed_directly_it<_Iter>::value;
+template <>
+struct is_passed_directly<oneapi::dpl::discard_iterator> : ::std::true_type
```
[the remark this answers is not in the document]
MikeDvorskiy: We didn't care about `const `fancy iterators before.
I am not sure that we have to start to do it now.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -187,19 +187,78 @@ struct is_permutation<Iter, ::std::enable_if_t<Iter::is_permutation::value>> : :
...
+{
+};
+
+template <oneapi::dpl::access_mode Mode, typename... Types>
+struct is_hetero<oneapi::dpl::__internal::sycl_iterator<Mode, Types...>> : ::std::true_type
+{
+};
+
+template <typename Iter>
+struct is_hetero<::std::reverse_iterator<Iter>> : is_hetero<Iter>
```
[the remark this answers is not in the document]
MikeDvorskiy: I think we didnt support it before...
If we device to support one, it should be a separate PR.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -352,34 +394,18 @@ struct __get_sycl_range
...
+    ::std::vector<::std::unique_ptr<oneapi::dpl::__internal::__lifetime_keeper_base>> m_buffers;
+
+    static constexpr bool __is_copy_direct =
+            AccMode == sycl::access::mode::read_write || AccMode == sycl::access::mode::read;
+    static constexpr bool __is_copy_back =
+            AccMode == sycl::access::mode::read_write || AccMode == sycl::access::mode::write;
+            
+    //SFINAE iterator type checks
+    template<typename It, typename T = decltype(std::addressof(*It{}))>
+    static constexpr std::true_type __test_addressof(int) { return {};}
```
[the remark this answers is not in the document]
MikeDvorskiy: `static_assert(__test_addressof<_Iter>(0));` ?
is it evaluated context?

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -538,30 +571,81 @@ struct __get_sycl_range
...
 
-    //specialization for a host iterator
+    //SFINAE-overload for a contiguous host iterator
+    template <typename _Iter>
+    auto
+    operator()(_Iter __first, _Iter __last)
+        -> ::std::enable_if_t<is_temp_buff<_Iter>::value && __is_addressable_v<_Iter> && !is_zip<_Iter>::value &&
+        !is_permutation<_Iter>::value, __range_holder<oneapi::dpl::__ranges::all_view<val_t<_Iter>, AccMode>>>
+    {
+        static_assert(__is_addressable_v<_Iter>);
```
rarutyun, line 581: This `static_assert` looks redundant based on what I can see in `enable_if`

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -490,33 +496,36 @@ struct permutation_view_simple<_R, _M, ::std::enable_if_t<oneapi::dpl::__interna
     }
 
     auto
-    base() const -> decltype(__r)
+    base() const
```
rarutyun, line 499: The behavior might be different with your change. Let's make sure that the current one is what we want.

rarutyun (commented):
Overall, I have several suggestions about naming that I didn't mention. I suggest to spend some time tomorrow for this review discussion. I feel like it's very closed to be merged.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
MikeDvorskiy: it is just  for check that we have not a mistake in signature with enable_if. I think it is helpful.

include/oneapi/dpl/pstl/utils_ranges.h
MikeDvorskiy: why different?

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -358,34 +392,20 @@ struct __get_sycl_range
...
-    {
-        //create a SYCL buffer and copy data [first, last) or create a empty SYCL buffer with size = (last - first)
-        if constexpr (_CopyDirectTag::value)
-            return sycl::buffer<val_t<_Iter>, 1>(__first, __last);
-        else
-            return sycl::buffer<val_t<_Iter>, 1>(__last - __first);
-    }
+    //SFINAE iterator type checks
+    template<typename It, typename T = decltype(std::addressof(*::std::declval<It&>()))>
+    static constexpr std::true_type __is_addressable (int);
```
rarutyun, line 404: Please consider the following:

```suggestion
template<typename It>
static constexpr auto __is_addressable (int) -> decltype(std::addressof(*::std::declval<It&>()), std::true_type{});
```

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -358,34 +392,20 @@ struct __get_sycl_range
...
+    template<typename It>
+    static constexpr std::false_type __is_addressable (...);
+    template<typename It>
+    static constexpr bool __is_addressable_v = decltype(__is_addressable <It>(0))::value;
```
rarutyun, line 408:
```suggestion
    static constexpr bool __is_addressable_v = decltype(__is_addressable<It>(0))::value;
```

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
rarutyun: As we discussed I don't see how it is different from the rest of checks.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -358,34 +392,20 @@ struct __get_sycl_range
...
-        if constexpr (_CopyDirectTag::value)
-            return sycl::buffer<val_t<_Iter>, 1>(__first, __last);
-        else
-            return sycl::buffer<val_t<_Iter>, 1>(__last - __first);
-    }
+    //SFINAE iterator type checks
+    template<typename It, typename T = decltype(std::addressof(*::std::declval<It&>()))>
+    static constexpr std::true_type __is_addressable (int);
+    template<typename It>
+    static constexpr std::false_type __is_addressable (...);
```
rarutyun, line 406:
```suggestion
    static constexpr std::false_type __is_addressable(...);
```

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
MikeDvorskiy: removed

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -193,19 +193,63 @@ struct is_permutation<Iter, ::std::enable_if_t<Iter::is_permutation::value>> : :
...
+{
+};
+
+template <typename... Iters>
+struct is_passed_directly<zip_iterator<Iters...>>: ::std::conjunction<is_passed_directly<Iters>...>
+{
+};
+
+template<typename Iter>
+inline constexpr bool is_passed_directly_v = is_passed_directly<::std::remove_const_t<Iter>>::value;
```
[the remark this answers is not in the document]
MikeDvorskiy: removed

include/oneapi/dpl/pstl/utils_ranges.h
MikeDvorskiy: As we checked In given case there is difference.

rarutyun (approved):
Looks good to me

PR #1256, opened 2023-10-31
[KT] Force inlining of ESIMD kernel callables 

dmitriy-sobolev:
Inlining must be guaranteed according to [Static allocation of SLM using slm_init function](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_esimd/sycl_ext_intel_esimd.md#static-allocation-of-slm-using-slm_init-function):
> The call of slm_init must be placed in the beginning of the kernel. If slm_init is called in some function 'F' called from kernel, then inlining of 'F' to the kernel must be forced/guaranteed.

This will help to avoid this issue (usually happens with -O0/-O1/-g compile options):
> LLVM ERROR: GenXLowering failed for: <  
call void @llvm.genx.slm.init(i32 %globalOrPrivateLoad)>: SLM init call is supported only in kernels

### DIFF
--- a/include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
+++ b/include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
@@ -27,7 +27,7 @@ namespace oneapi::dpl::experimental::kt::esimd::__impl
 
 template <bool __is_ascending, ::std::uint8_t __radix_bits, ::std::uint16_t __data_per_work_item,
           ::std::uint16_t __work_group_size, typename _KeyT, typename _InputT>
-void
+inline __attribute__((always_inline)) void
 __one_wg_kernel(sycl::nd_item<1> __idx, ::std::uint32_t __n, const _InputT& __input)
 {
     using _BinT = ::std::uint16_t;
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -27,7 +27,7 @@ namespace oneapi::dpl::experimental::kt::esimd::__impl
 
 template <bool __is_ascending, ::std::uint8_t __radix_bits, ::std::uint16_t __data_per_work_item,
           ::std::uint16_t __work_group_size, typename _KeyT, typename _InputT>
-void
+inline __attribute__((always_inline)) void
```
SergeyKopienko, line 30: Could you please pack this part `inline __attribute__((always_inline)`
into some new define?

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: Done

PR #1257, opened 2023-11-01
[KT] ESIMD-based radix sort kernel templates

dmitriy-sobolev:
### High level architecture
1. API. `esimd_radix_sort.h`: contains API for ESIMD radix sort kernel templates. `kernel_param.h`: contains generic configuration structure for any kernel template. 
2. Dispatchers. `esimd_radix_sort_dispatchers.h`: contains entities which specify a basic structure of an algorithm and select specific block of code which depend in input parameters and, perhaps, on GPU architecture in the future.
3. Submitters. `esimd_radix_sort_submitters.h`: contain entities which submit kernels; they serve the same purpose as submitters in [regular oneDPL code](https://github.com/oneapi-src/oneDPL/blob/bb878cdeb79794a7c7f607c985653f275172196c/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h#L217#L221).
4. Kernels. `esimd_radix_sort_kernels.h`. 

This structure can be revised.

## Other
ESIMD documentation: https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_esimd/sycl_ext_intel_esimd.md


SergeyKopienko:
@dmitriy-sobolev Why you are using pictures in PR description?
May be better to copy C++ code ?

SergeyKopienko (dismissed):
LGTM

dmitriy-sobolev:
> @dmitriy-sobolev Why you are using pictures in PR description? May be better to copy C++ code ?

I have removed the pictures

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -0,0 +1,84 @@
...
+#include "internal/esimd_radix_sort_utils.h"
+#include "internal/esimd_radix_sort_dispatchers.h"
+
+namespace oneapi::dpl::experimental::kt::esimd
+{
+
+// TODO: make sure to provide sufficient diagnostic if input does not allow either reading or writing
+template <bool __is_ascending = true, ::std::uint8_t __radix_bits = 8, typename _KernelParam, typename _KeysRng>
+sycl::event
+radix_sort(sycl::queue __q, _KeysRng&& __keys_rng, _KernelParam __param = {})
```
SergeyKopienko, line 29: May be better to separate declaration of public API from implementation?

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_utils.h
```diff
@@ -0,0 +1,881 @@
...
+};
+
+template<>
+struct __rng_value_type_deducer<__rng_dummy>
+{
+    using __value_t = void;
+};
+
+template<typename _Rng1, typename _Rng2 = __rng_dummy>
+struct __rng_pack
```
SergeyKopienko, line 829: Looks like std::tuple

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_utils.h
```diff
@@ -0,0 +1,881 @@
...
+    static constexpr bool __has_values = !std::is_void_v<_ValT>;
+
+    const auto& __keys_rng() const { return __m_keys_rng; }
+    const auto& __vals_rng() const
+    {
+        static_assert(__has_values);
+        return __m_vals_rng;
+    }
+
+    __rng_pack(const _Rng1& __rng1, const _Rng2& __rng2 = __rng_dummy{}):
```
SergeyKopienko, line 842: Looks like these two constructors may be combined into one with universal reference

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -0,0 +1,84 @@
...
+
+#ifndef _ONEDPL_KT_ESIMD_RADIX_SORT_H
+#define _ONEDPL_KT_ESIMD_RADIX_SORT_H
+
+#include "kernel_param.h"
+#include "../../pstl/utils_ranges.h"
+#include "../../pstl/hetero/dpcpp/utils_ranges_sycl.h"
+#include "../../pstl/hetero/dpcpp/parallel_backend_sycl_utils.h"
+
+#include <cstdint>
```
MikeDvorskiy, line 18: Usually, the standard headers should be included first.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -0,0 +1,84 @@
...
+
+namespace oneapi::dpl::experimental::kt::esimd
+{
+
+// TODO: make sure to provide sufficient diagnostic if input does not allow either reading or writing
+template <bool __is_ascending = true, ::std::uint8_t __radix_bits = 8, typename _KernelParam, typename _KeysRng>
+sycl::event
+radix_sort(sycl::queue __q, _KeysRng&& __keys_rng, _KernelParam __param = {})
+{
+    if (__keys_rng.size() < 2)
```
MikeDvorskiy, line 31: oneDPL has such trivial sequence checks on the algorithm pattern (parallel/hetero)  level, for example https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h#L1216

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -0,0 +1,84 @@
...
+        return {};
+
+    auto __pack = __impl::__utils::__rng_pack{oneapi::dpl::__ranges::views::all(::std::forward<_KeysRng>(__keys_rng))};
```
MikeDvorskiy, line 34: What is a pack?
Why does the pack contain just one element? 
Even, it is reasonable, I guess to have a comment  here  would be nice.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,687 @@
...
+    using _GlobHistT = _SimdT<_GlobOffsetT, __bin_count>;
+
+    static constexpr ::std::uint32_t __bit_count = sizeof(_KeyT) * 8;
+    static constexpr _LocOffsetT __mask = __bin_count - 1;
+    static constexpr ::std::uint32_t __hist_stride = __bin_count * sizeof(_LocOffsetT);
+    static constexpr ::std::uint32_t __work_item_all_hists_size = __work_group_size * __hist_stride;
+    static constexpr ::std::uint32_t __group_hist_size = __hist_stride;
+    static constexpr ::std::uint32_t __global_hist_size = __bin_count * sizeof(_GlobOffsetT);
+
+    static constexpr ::std::uint32_t __calc_reorder_slm_size()
```
MikeDvorskiy, line 352: Probably, it makes sense to define that contexpr value as a static constexpr member?
`static constexpr ::std::uint32_t  __reorder_slm_size = __work_group_size * __data_per_work_item * (sizeof(_KeyT) + (__has_values ? sizeof(_ValT) : 0));`

dmitriy-sobolev: `_ValT` becomes `void` if this is a key-only sort. `sizeof(void)` will produce a compilation error in this expression.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -0,0 +1,327 @@
...
+    // TODO: make this more explicit to reduce coupling between __onesweep_memory_holder and __rng_pack
+    __onesweep_memory_holder<_GlobalHistT, _KeyT, _ValT> __mem_holder(__q);
+    __mem_holder.__keys_alloc_count(__n);
+    if constexpr (__has_values)
+    {
+        __mem_holder.__vals_alloc_count(__n);
+    }
+    __mem_holder.__global_hist_item_alloc_count(__global_hist_item_count);
+    __mem_holder.__group_hist_item_alloc_count(__group_hist_item_count);
+    __mem_holder.__allocate();
```
julianmi, line 218: I feel like these statements belong together. Do we need them separately or could we allocate the memory in the ctor?

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,687 @@
...
+// Copyright (C) 2023 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===------------------------------------------------------===//
+
+#ifndef _ONEDPL_KT_ESIMD_RADIX_SORT_KERNELS_H
+#define _ONEDPL_KT_ESIMD_RADIX_SORT_KERNELS_H
+
+#include "oneapi/dpl/pstl/onedpl_config.h"
```
julianmi: Should we move this include after the standard headers?

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,687 @@
...
+    __dpl_esimd_ns::simd<_DeviceAddrT, __data_per_work_item> __write_addr;
+    __dpl_esimd_ns::simd<_KeyT, __data_per_work_item> __keys;
+    __dpl_esimd_ns::simd<_BinT, __data_per_work_item> __bins;
+    __dpl_esimd_ns::simd<_DeviceAddrT, __data_per_step> __lane_id(0, 1);
+
+    const _DeviceAddrT __io_offset = __data_per_work_item * __local_tid;
+
+    static_assert(__data_per_work_item % __data_per_step == 0);
+    static_assert(__bin_count % 128 == 0);
+    static_assert(__bin_count % 32 == 0);
```
julianmi, line 70: Why do we need both checks? And didn't we check before that `__radix_bits` is eight or is this a general check for once we support other values?

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,687 @@
...
+            __write_addr[__s] = __bin_offset[__bins[__s]];
+            __bin_offset[__bins[__s]] += 1;
+        }
+
+        /*
+        first write to slm,
+        then do column scan by group, each thread to 32c*8r,
+        then last row do exclusive scan as incoming offset
+        then every thread add local sum with sum of previous group and incoming offset
+        */
```
julianmi, line 97: Does it make sense to move each of these comments before their respective section below to improve the readability? Similar to what we do below (`__global_histogram()`.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
dmitriy-sobolev: I have rearranged the headers.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,688 @@
...
+
+    inline void
+    __rank_global(_LocHistT& __subgroup_offset, _GlobHistT& __global_fix,
+                  ::std::uint32_t __local_tid, ::std::uint32_t __wg_id) const
+    {
+        /*
+        first do column scan by group, each thread do 32c,
+        then last row do exclusive scan as group incoming __offset
+        then every thread add local sum with sum of previous group and incoming __offset
+        */
```
julianmi, line 506: Similar comment as before regarding moving the comments to their respective code sections.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_utils.h
julianmi, line 1: Lots of these utilities seem general (not explicitly for radix sort) to me. Also, the namespace suggests general utilities. Does it make sense to split these up into generic utilities (e.g., load, store, copy, gather, scatter...) and radix sort specific ones?

dmitriy-sobolev: Great idea, I will do it

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: That's right. These general checks might be useful when adding the support for more configurations. 

They are applied to each loop to guarantee no tails are left. To be more precise:

```
    static_assert(__data_per_work_item % __data_per_step == 0);
    static_assert(__bin_count % 128 == 0);
    static_assert(__bin_count % 32 == 0);
...
   for (::std::uint32_t __s = 0; __s < __data_per_work_item; __s += __data_per_step)
   ...
   for (::std::uint32_t __s = 0; __s < __bin_count; __s += 128)
...
   for (::std::uint32_t __i = 32; __i < __bin_count; __i += 32)
```

test/kt/CMakeLists.txt
```diff
@@ -0,0 +1,105 @@
...
+        if (NOT "${_val_type}" STREQUAL "")
+            target_compile_definitions(${_target_name} PRIVATE TEST_VALUE_TYPE=${_val_type})
+        endif()
+    endif()
+endfunction()
+
+function(_generate_esimd_sort_tests _key_value_pairs)
+    set(_data_per_work_item_all "32" "64" "96" "128" "160" "192" "224" "256" "288" "320" "352" "384" "416" "448" "480" "512")
+    set(_work_group_size_all "32" "64")
+    set(_type_all "char" "uint16_t" "int" "uint64_t" "float" "double")
```
danhoeflinger, line 82: I think its worth adding some comments here to explain / justify the strategy here for generating / separating test cases, and randomly selecting a subset.

Eventually, it may be useful in other places, and we may want to abstract it out to reuse, but not in this PR.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -16,6 +16,7 @@
 #ifndef _ONEDPL_EXECUTION_SYCL_DEFS_H
 #define _ONEDPL_EXECUTION_SYCL_DEFS_H
 
+#include "../../../internal/common_config.h"
```
danhoeflinger, line 19: It seems `common_config.h` is only included from public top level headers until this addition. Perhaps this is unnecessary now, as it is already in `include/oneapi/dpl/experimental/kernel_templates`

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
dmitriy-sobolev: I tried moving them at `__impl::__radix_sort` level, but, alas, `oneapi::dpl::__ranges::__get_sycl_range::operator()` does not allow zero-sized ranges due to checks like `assert(__n > 0);` . So this is not a trivial thing.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_utils.h
dmitriy-sobolev: Done

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
dmitriy-sobolev: It has benefits and drawbacks:
 - (+) It is easier to reference the API here as documentation.
 - (-) Duplication.

I am inclined to leave it as is unless there are more benefits.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
dmitriy-sobolev: This is an abstraction which stores either keys or key-value pairs. Perhaps, `std::tuple` suits here better as noted by Sergey. Let me think about it.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
dmitriy-sobolev: I have removed it

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
danhoeflinger: Unless I am mistaken, I think it breaks convention with the rest of the library.  As far as I know, we have forward definitions of the public APIs throughout oneDPL. Of course, with KT, we can perhaps make a change in this decision.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
danhoeflinger: I assume you looked at representing these as `zip_iterators`, for which we have `__get_sycl_range` overloads.  If there is not something which blocks it from being possible, it may help in the discussion of the repeated `size < 2` checks mentioned elsewhere by delaying `__get_sycl_range` until further in after the "packing".

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: I have described this stage more thoroughly, each its step individually.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
dmitriy-sobolev: As far as I understand, other APIs require inclusion of two headers: `execution` for the policies and one among `algorithm` `memory`, `numeric`  and other headers for specific algorithms. Forward declarations are used to support different inclusion orders (when `execution` is included before or after e.g.  `algorithm`).

I am not sure if such a convention exists, because it looks to me as a solution of a technical problem which does not exist in KT. 

@MikeDvorskiy, Is there such a convention? Would it be better to have forward declarations in the KT?

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_submitters.h
```diff
@@ -0,0 +1,192 @@
...
+            {
+                __cgh.depends_on(__e);
+                __cgh.parallel_for<_Name...>(__nd_range,
+                                             [=](sycl::nd_item<1> __nd_item)
+                                             {
+                                                 ::std::uint32_t __offset = __nd_item.get_global_id(0);
+                                                 const auto __g = __nd_item.get_group();
+                                                 ::std::uint32_t __count = __global_offset_data[__offset];
+                                                 ::std::uint32_t __presum = sycl::exclusive_scan_over_group(
+                                                     __g, __count, sycl::plus<::std::uint32_t>());
```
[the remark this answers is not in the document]
dmitriy-sobolev: Thanks, I completely forgot about these wrappers. They are now in use.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_submitters.h
```diff
@@ -0,0 +1,192 @@
...
+            [&](sycl::handler& __cgh)
+            {
+                __cgh.depends_on(__e);
+                __cgh.parallel_for<_Name...>(__nd_range,
+                                             [=](sycl::nd_item<1> __nd_item)
+                                             {
+                                                 ::std::uint32_t __offset = __nd_item.get_global_id(0);
+                                                 const auto __g = __nd_item.get_group();
+                                                 ::std::uint32_t __count = __global_offset_data[__offset];
+                                                 ::std::uint32_t __presum = sycl::exclusive_scan_over_group(
```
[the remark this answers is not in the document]
dmitriy-sobolev: Not it is used

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_submitters.h
```diff
@@ -0,0 +1,192 @@
...
+#define _ONEDPL_KT_ESIMD_RADIX_SORT_SUBMITTERS_H
+
+#include <cstdint>
+#include <utility>
+#include <type_traits>
+
+#if __has_include(<sycl/sycl.hpp>)
+#    include <sycl/sycl.hpp>
+#else
+#    include <CL/sycl.hpp>
```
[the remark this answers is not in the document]
dmitriy-sobolev: Now, it is used

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,706 @@
...
+    static_assert(__data_per_work_item % __data_per_step == 0);
+    static_assert(__bin_count % 128 == 0);
+    static_assert(__bin_count % 32 == 0);
+    _ONEDPL_PRAGMA_UNROLL
+    for (::std::uint32_t __s = 0; __s < __data_per_work_item; __s += __data_per_step)
+    {
+        __dpl_esimd::__ns::simd_mask<__data_per_step> __m = (__io_offset + __lane_id + __s) < __n;
+        __keys.template select<__data_per_step, 1>(__s) =
+            __dpl_esimd::__ns::merge(__dpl_esimd::__gather<_KeyT, __data_per_step>(__rng_data(__rng_pack.__keys_rng()), __lane_id, __io_offset + __s, __m),
+                  __dpl_esimd::__ns::simd<_KeyT, __data_per_step>(__sort_identity<_KeyT, __is_ascending>()), __m);
```
danhoeflinger, line 71: Is `__gather` required here to on all input because incoming data user may not be aligned to our needs?
If so, aren't we paying a huge price to align it by gathering across all input data?  

I think we could probably get away with a single gather.

We are already padding at the end for when `__n` and a multiple of `__data_per_work_item * __work_group_size` do not align.  Can we also pad at the start with `__sort_identity<_KeyT, **!__is_ascending**>` to keep those elements at the beginning of the sorted list? 

I think this would allow us to special case this first load to get to the start of the aligned input data, and then have we can skip the gather for the rest of the copy-in.  

The same idea would be applied on the way out for the scatter.  Special case the first step, then deal in aligned data the rest of the way.

danhoeflinger (commented):
I've still not gotten too deep into the actual kernels in the review, but I have one potentially interesting comment for the one_wg case so far.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
MikeDvorskiy: oneDPL has one rule regarding "separate declaration of public API" for "standard" parallel algorithm API, located into algorithm, numeric, memory. A reason and logic of that is described here:
https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/numeric#L26
Lines  26-35

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,706 @@
...
+        }
+
+        if (__stage != __stage_count - 1)
+        {
+            _ONEDPL_PRAGMA_UNROLL
+            for (::std::uint32_t __s = 0; __s < __data_per_work_item; __s += __data_per_step)
+            {
+                __dpl_esimd::__vector_store<_KeyT, 1, __data_per_step>(
+                    __write_addr.template select<__data_per_step, 1>(__s) * sizeof(_KeyT),
+                    __keys.template select<__data_per_step, 1>(__s));
```
danhoeflinger, line 211: I think it would be more clear if these overloads of `__vector_store` were renamed to include `SLM` in the name, similar to the function(s) they wrap.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,706 @@
...
+            __dpl_esimd::__ns::barrier();
+            __keys = __dpl_esimd::__block_load_slm<_KeyT, __data_per_work_item>(__slm_reorder_this_thread);
+        }
+    }
+    _ONEDPL_PRAGMA_UNROLL
+    for (::std::uint32_t __s = 0; __s < __data_per_work_item; __s += __data_per_step)
+    {
+        __dpl_esimd::__scatter<_KeyT, __data_per_step>(__rng_data(__rng_pack.__keys_rng()), __write_addr.template select<__data_per_step, 1>(__s),
+                                             __keys.template select<__data_per_step, 1>(__s),
+                                             __write_addr.template select<__data_per_step, 1>(__s) < __n);
```
danhoeflinger, line 224: I suppose now with the full understanding, the output scatter cannot be saved or avoided (as I suggested above), as we are not just pulling it out of contiguous memory but also doing the final stage's re-order.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,706 @@
...
+    __dpl_esimd::__ns::simd<_BinT, __data_per_work_item> __bins;
+    __dpl_esimd::__ns::simd<_DeviceAddrT, __data_per_step> __lane_id(0, 1);
+
+    const _DeviceAddrT __io_offset = __data_per_work_item * __local_tid;
+
+    static_assert(__data_per_work_item % __data_per_step == 0);
+    static_assert(__bin_count % 128 == 0);
+    static_assert(__bin_count % 32 == 0);
+    _ONEDPL_PRAGMA_UNROLL
+    for (::std::uint32_t __s = 0; __s < __data_per_work_item; __s += __data_per_step)
```
danhoeflinger, line 69: I think we require that `__data_per_work_item` is a multiple of 32.  Is there a benefit or requirement from using 16 rather than 32?

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,706 @@
...
+            __bin_offset[__bins[__s]] += 1;
+        }
+
+        /*
+        first write to slm,
+        then do column scan by group, each thread to 32c*8r,
+        then last row do exclusive scan as incoming offset
+        then every thread add local sum with sum of previous group and incoming offset
+        */
+        {
```
danhoeflinger, line 93: Does this bracket scope provide anything more than a conceptual count and scan section?  If not I would say we should delete it and just add more comments, or break it up into helper functions.  I think comments are sufficient.

danhoeflinger (commented):
Alright, added some more feedback for the one_wg kernel.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: I have added them. They now describe the steps in the same fashion as it is done for the onesweep kernel.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: Extra scopes have been removed. Code blocks are now described.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
dmitriy-sobolev: It makes the code slightly more complex. This is what I've got: [mem_holder_ctor.patch](https://github.com/oneapi-src/oneDPL/files/13569940/mem_holder_ctor.patch)
Two constructors are needed to avoid calling `sizeof(_ValueT)` when `_ValueT` is `void`. This might be changed, but I need to thinkg about how to do it the proper way. Perhaps, the whole design `__onesweep_memory_holder` should be reconsidered.

By the way, the current implementation is inspired by https://refactoring.guru/design-patterns/builder.

danhoeflinger: I may be mistaken but I think you should be able to use an `enable_if` call checking the type of `_ValueT` to separate the two constructors at compile time based on the provided types without requiring a different number of arguments.  I think that should contain the information you need to know if it is key sort or pair sort.
For me, the important thing would be unifying the constructor call in your patch which is currently separated by a constexpr.


If we want to leave the current structure, it can at least be improved by renaming the functions to `__set_...`, to make it more clear what the individual functions are doing.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -0,0 +1,320 @@
...
+    {
+        return __onesweep<_KernelName, __is_ascending, __radix_bits, __data_per_workitem, __workgroup_size>(
+            __q, ::std::forward<_RngPack>(__pack), __n);
+    }
+    else
+    {
+        constexpr ::std::uint32_t __one_wg_cap = __data_per_workitem * __workgroup_size;
+        if (__n <= __one_wg_cap)
+        {
+            // TODO: support different RadixBits values (only 7 or 8 are currently supported), WorkGroupSize
```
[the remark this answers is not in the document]
dmitriy-sobolev: I have tested only single-work-group kernel with different `RadixBits`. All the tests successfully pass with `7`, `8` and `9`. This comment is only about this single-work-group specialization.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: > Is __gather required here to on all input because incoming data user may not be aligned to our needs?

Yes, but this is not entirely correct. User-provided will always be element-aligned (otherwise, the user does not follow an inherent C++ requirement on the alignment). `gather` only requires the passed offset to be element-aligned. However, there are other, potentially more efficient, instructions with no overalignment requirements.

E.g. [block_load](https://intel.github.io/llvm-docs/doxygen/group__sycl__esimd__memory__block.html#ga49ae87e4075911a7e92549fb1e25fdb8), which may be used to load non-resudial elements. The API reference is not very clear so far, but my understanding is that an offset passed to `block_load` must be at least element-aligned if you provide the alignment property. 
 
> I think we could probably get away with a single gather (or maybe none?).

We can  reuse `__load_simd` wrapper over `gather` from onesweep kernel. It also applies the mask only to the residual elements, which is potentially more performant.

I think, we still need to load data in a loop portion by portion if we use `gather`. Single use with large width may increase register pressure for keeping a long mask vector.

dmitriy-sobolev: > We are already padding at the end for when __n and a multiple of __data_per_work_item * __work_group_size do not align. 

This is not for the sake of alignment. The "sort identity" elements are added to avoid extra branching. They will always be at the end of the sequence after reordering. So, you just need to avoid scattering the last elements; no need to add extra `if` conditions in the code.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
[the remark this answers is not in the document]
danhoeflinger: There is a `static_assert(__radix_bits == 8);` at the top of this function, should it be removed or expanded then?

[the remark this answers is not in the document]
dmitriy-sobolev: The current approach is to restrict values of the parameters (`RadixBits`, `DataPerWorkItem`, `WorkGroupSize`) to the strictest values among all the used kernels. `onesweep` kernel only supports `static_assert(__radix_bits == 8)`, so we prohibit passing other values. 

Do you think that we should allow passing more arguments? If so, the documentation would need to have more details on the internals. E.g. if `n < DataPerWorkItem * WorkGroupSize` then you can also pass `7` or `8` as `RadixBits`.

Also, `one_wg` kernel also support wider range of `WorkGroupSize` parameters.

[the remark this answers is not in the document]
dmitriy-sobolev: Now I understand what you meant. I have moved these checks onto the API-level, where they belong.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,706 @@
...
+            __dpl_esimd::__ns::barrier();
+            _ONEDPL_PRAGMA_UNROLL
+            for (::std::uint32_t __s = 0; __s < __bin_count; __s += 128)
+            {
+                __dpl_esimd::__block_store_slm<::std::uint32_t, 64>(
+                    __slm_bin_hist_this_thread + __s * sizeof(_HistT),
+                    __bin_offset.template select<128, 1>(__s).template bit_cast_view<::std::uint32_t>());
+            }
```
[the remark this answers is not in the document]
dmitriy-sobolev: > Is there some restrictions on the type for __block_store_slm? Why cant we just store 128 elements of type _HistT?

There used to be a restriction on the types. Intel® oneAPI DPC++/C++ Compiler Release Notes (2023.1):
_Added 8/16-bit type support to lsc_block_load and lsc_block_store ESIMD API._

ESIMD sort KT is going to supported for 2023.2 and newer, so we can handle `_HistT` directly.

> If this isn't possible, we should calculate the size difference between _HistT and ::std::uint32_t and use that rather than assuming a factor of 2.

Agree. It will save us from issues when changing the code in the future.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: That's true. The elements are scattered at arbitrary positions.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,722 @@
...
+                                                __wg_offset * sizeof(_ValT);
+            __dpl_esimd::__vector_store<_ValT, 1, __data_per_work_item>(__wg_offset_vals, __pack.__vals);
+        }
+        __dpl_esimd::__ns::barrier();
+    }
+
+    void inline
+    __reorder_slm_to_glob(const _GlobHistT& __global_fix, ::std::uint32_t __local_tid, ::std::uint32_t __wg_size) const
+    {
+        auto __pack = __make_simd_pack<__data_per_work_item, _KeyT, _ValT>();
```
danhoeflinger, line 647: Is there a reason we call `__make_simd_pack` in the helper function rather than pass it in like is done in some of the other helpers?

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
dmitriy-sobolev: I have noted it in #1311 to address later.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,722 @@
...
+                    __simd.template select<__data_per_step, 1>(__s) = __gathered;
+                }
+            }
+        }
+    }
+
+    static inline __dpl_esimd::__ns::simd<::std::uint32_t, 32>
+    __match_bins(const __dpl_esimd::__ns::simd<::std::uint32_t, 32>& __bins, ::std::uint32_t __local_tid)
+    {
+        __dpl_esimd::__ns::fence<__dpl_esimd::__ns::fence_mask::sw_barrier>();
```
danhoeflinger, line 462: It would be good to add comments to explain the usage of `__dpl_esimd::__ns::fence` here and the other places it is present.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,706 @@
...
+            if (__local_tid < __bin_summary_group_size)
+            {
+                constexpr ::std::uint32_t __bin_width = __bin_count / __bin_summary_group_size;
+                constexpr ::std::uint32_t __bin_width_ud = __bin_width * sizeof(_HistT) / sizeof(::std::uint32_t);
+                ::std::uint32_t __slm_bin_hist_summary_offset = __local_tid * __bin_width * sizeof(_HistT);
+                __dpl_esimd::__ns::simd<_HistT, __bin_width> ___thread_grf_hist_summary;
+                __dpl_esimd::__ns::simd<::std::uint32_t, __bin_width_ud> __tmp;
+
+                ___thread_grf_hist_summary.template bit_cast_view<::std::uint32_t>() =
+                    __dpl_esimd::__block_load_slm<::std::uint32_t, __bin_width_ud>(__slm_bin_hist_summary_offset);
```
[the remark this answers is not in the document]
dmitriy-sobolev: This is also noted in https://github.com/oneapi-src/oneDPL/issues/1311.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,722 @@
...
+    template<typename _SimdPack>
+    void inline
+    __reorder_reg_to_slm(const _SimdPack& __pack, const _LocOffsetSimdT& __ranks,
+                         const _LocOffsetSimdT& __bins, const _LocHistT& __subgroup_offset,
+                         ::std::uint32_t __wg_size, ::std::uint32_t __thread_slm_offset) const
+    {
+        __slm_lookup<_LocOffsetT> __subgroup_lookup(__thread_slm_offset);
+        _LocOffsetSimdT __wg_offset =
+            __ranks + __subgroup_lookup.template __lookup<__data_per_work_item>(__subgroup_offset, __bins);
+        __dpl_esimd::__ns::barrier();
```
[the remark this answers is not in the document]
danhoeflinger: With more thought, I think this barrier is required.  We are setting up the SLM copy of the global_fix, and need to look up segments of it later on.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: I have noted it in #1311 for addressing in the future

include/oneapi/dpl/experimental/kt/internal/esimd_defs.h
```diff
@@ -0,0 +1,590 @@
...
+#include <limits>
+#include <cstdint>
+#include <type_traits>
+
+#include "../../../pstl/hetero/dpcpp/sycl_defs.h"
+#include <sycl/ext/intel/esimd.hpp>
+
+// The macro is created to guarantee inlining of functions which contain slm_init. See ESIMD spec for more details:
+// https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_esimd/sycl_ext_intel_esimd.md#static-allocation-of-slm-using-slm_init-function
+#define _ONEDPL_ESIMD_INLINE inline __attribute__((always_inline))
```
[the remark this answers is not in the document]
dmitriy-sobolev: I agree. I have noted it in #1311

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: This is noted in #1312. Thanks for the great comment!

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,706 @@
...
+
+    constexpr ::std::uint32_t __bin_count = 1 << __radix_bits;
+    constexpr ::std::uint32_t __bit_count = sizeof(_KeyT) * 8;
+    constexpr ::std::uint32_t __stage_count = oneapi::dpl::__internal::__dpl_ceiling_div(__bit_count, __radix_bits);
+    constexpr _BinT __mask = __bin_count - 1;
+    constexpr ::std::uint32_t __hist_stride = sizeof(_HistT) * __bin_count;
+
+    constexpr ::std::uint32_t __reorder_slm_size = __data_per_work_item * sizeof(_KeyT) * __work_group_size;
+    constexpr ::std::uint32_t __bin_hist_slm_size = __hist_stride * __work_group_size;
+    constexpr ::std::uint32_t __incoming_offset_slm_size = (__bin_count + 1) * sizeof(_HistT);
```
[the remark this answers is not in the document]
dmitriy-sobolev: It's been noted in #1312.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,706 @@
...
+    {
+        __bins = __get_bucket<__mask>(__order_preserving_cast<__is_ascending>(__keys), __stage * __radix_bits);
+
+        __bin_offset = 0;
+        _ONEDPL_PRAGMA_UNROLL
+        for (::std::uint32_t __s = 0; __s < __data_per_work_item; __s += 1)
+        {
+            __write_addr[__s] = __bin_offset[__bins[__s]];
+            __bin_offset[__bins[__s]] += 1;
+        }
```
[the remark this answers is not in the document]
dmitriy-sobolev: It's been noted in #1312 for addressing in the future PRs.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: I've noted it in #1312.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,722 @@
...
+    }
+
+    static inline __dpl_esimd::__ns::simd<::std::uint32_t, 32>
+    __match_bins(const __dpl_esimd::__ns::simd<::std::uint32_t, 32>& __bins, ::std::uint32_t __local_tid)
+    {
+        __dpl_esimd::__ns::fence<__dpl_esimd::__ns::fence_mask::sw_barrier>();
+        __dpl_esimd::__ns::simd<::std::uint32_t, 32> __matched_bins(0xffffffff);
+        _ONEDPL_PRAGMA_UNROLL
+        for (int __i = 0; __i < __radix_bits; __i++)
```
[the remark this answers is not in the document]
dmitriy-sobolev: I've noted it in #1312.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: I have added comments

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -0,0 +1,722 @@
...
+        constexpr ::std::uint32_t __global_accumulated = 0x40000000;
+        constexpr ::std::uint32_t __hist_updated = 0x80000000;
+        constexpr ::std::uint32_t __global_offset_mask = 0x3fffffff;
+
+        _GlobOffsetT* __p_this_group_hist = __p_group_hists + __bin_count * __wg_id;
+        // First group contains global histogram to propagate it to other groups during synchronization
+        _GlobOffsetT* __p_prev_group_hist = (0 == __wg_id)? __p_global_hist : __p_this_group_hist - __bin_count;
+
+        {
+            __dpl_esimd::__ns::barrier();
```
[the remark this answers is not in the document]
dmitriy-sobolev: Indeed. I have removed that barrier.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
dmitriy-sobolev: Fixed

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
dmitriy-sobolev: I left a not about this suggestion in #1311 for further code quality improvements

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_utils.h
dmitriy-sobolev: This will become obsolete if we remove `__rng_pack` entirely, e.g. if we use `zip_iterator` or `zip_view`. But thank you for the suggestion.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
dmitriy-sobolev: I have noted it in #1311 for further code quality enhancements.

PR #1259, opened 2023-11-02
[KT] Add option to enable/disable ESIMD KT tests

dmitriy-sobolev:
Add an option to enable/disable ESIMD KT tests. The tests are disable by default.
Refactor CMake: move the section with creation of a test target into a common location.
Add the license header.


SergeyKopienko (approved):
ДПЕЬ

PR #1260, opened 2023-11-02
[oneDPL] Added "in order" queue support

MikeDvorskiy:
[oneDPL] Added "in order" queue support.

We introduce `__dpl_sycl::__event` and `__dpl_sycl::sumbit(..)` wrappers.

1. For `out_of_order `queue the mentioned wrappers work as sycl::event and sycl::queue::submit, set sycl dependencies and return sycl event.
2. For `in_order ` queue `__dpl_sycl::sumbit` doesn't set sycl dependencies and dosn't return sycl event. To provide a sync oneDPL call `__dpl_sycl::__event` calls `sycl::queue::wait`.


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -416,6 +416,50 @@ __get_accessor_ptr(const _Acc& __acc)
...
 
+class __event
+{
+    sycl::event __m_event;
+    sycl::queue __m_queue;
+    bool __in_order = false;
+
+  public:
+    __event(sycl::event __e): __m_event(__e) {}
+    __event() {}
```
SergeyKopienko, line 427: `__event() = default;`

MikeDvorskiy: What's a advantages of that in given case?

SergeyKopienko: In common case `=default` doesn't change triviality state of class.
See https://ru.stackoverflow.com/questions/495342/%D0%A7%D0%B5%D0%BC-%D0%BF%D1%83%D1%81%D1%82%D0%BE%D0%B9-%D0%BA%D0%BE%D0%BD%D1%81%D1%82%D1%80%D1%83%D0%BA%D1%82%D0%BE%D1%80-%D0%BF%D0%BE-%D1%83%D0%BC%D0%BE%D0%BB%D1%87%D0%B0%D0%BD%D0%B8%D1%8E-%D0%BE%D1%82%D0%BB%D0%B8%D1%87%D0%B0%D0%B5%D1%82%D1%81%D1%8F-%D0%BE%D1%82-default for details.

MikeDvorskiy: "doesn't change triviality state of class." - yes...
But, do we really need it? or it can be useful in the Future?

SergeyKopienko: This general idea look like as applied in all cases.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -416,6 +416,50 @@ __get_accessor_ptr(const _Acc& __acc)
...
+
+    void wait()
+    {
+        if(__in_order)
+            __m_queue.wait_and_throw();
+        else
+            __m_event.wait_and_throw();
+    }
+
+    bool has_event() const { return !__in_order; }
```
danhoeflinger, line 439: Is this a problem if the default constructor was called to create this `__event`?  `__in_order` is false, but only a default constructed event exists.  

I suppose in the way it is used currently, this isnt a problem.  However, the naming of this function may be misleading.  Maybe we would just be better off to `assert(!__in_order);` directly in the conversion function.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -145,15 +145,15 @@ class __radix_sort_reorder_kernel;
...
 
 template <typename _KernelName, ::std::uint32_t __radix_bits, bool __is_ascending, typename _ExecutionPolicy,
-          typename _ValRange, typename _CountBuf, typename _Proj
+          typename _ValRange, typename _CountBuf, typename _Event, typename _Proj
 #if _ONEDPL_COMPILE_KERNEL
           , typename _Kernel
 #endif
           >
-sycl::event
+auto
```
danhoeflinger, line 153: I think it may be better for code maintainability to use `__dpl_sycl::__event` as the return type rather than `auto`.  (same for other similar functions in thie PR)

For the declarations, I don't mind the use of `auto`, but for returns of large functions, this seems like unnecessary obfuscation when the type is known and has a simple name.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -416,6 +416,50 @@ __get_accessor_ptr(const _Acc& __acc)
...
+    operator sycl::event() const
```
danhoeflinger, line 440: Would it be safer to make this an explicit function call like `get_sycl_event()` rather than a conversion function?

I'm worried we might get ourselves into a bit of trouble with a conversion function since this is not always interchangeable with a `sycl::event` (as evidenced by the assert).

Is this only used in the `__future` implementation within this PR currently?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -595,12 +595,15 @@ class __future : private std::tuple<_Args...>
...
     void
     wait()
     {
-#if !ONEDPL_ALLOW_DEFERRED_WAITING
-        __my_event.wait_and_throw();
-#endif
+//#if !ONEDPL_ALLOW_DEFERRED_WAITING
+        //__my_event.wait_and_throw();
+//#endif
+        __my_event.wait();
```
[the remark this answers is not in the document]
MikeDvorskiy: yes... (I forgot remove the old code)

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
MikeDvorskiy: Actually, I've redesigned that code a little bit.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -416,6 +416,72 @@ __get_accessor_ptr(const _Acc& __acc)
...
+           __hdl.depends_on(*__dep_int);
+       __body(__hdl);
+       }));
+}
+
+template <typename _Body>
+auto
+__submit(sycl::queue __queue, _Body __body)
+{
+    return __submit_impl(__queue, __body, true, NULL, NULL);
```
SergeyKopienko, line 475: Better to use `null_ptr` keyword here and below.

PR #1261, opened 2023-11-03
[oneDPL] A fix - added missed synchronization between two SYCL patterns.

MikeDvorskiy:
In some algorithm patterns there are calls of more then one SYCL backend patterns. As far as SYCL backend patterns are asynchronous, we have to make synch between the them. It might be set SYCL dependencies or "wait" of the first pattern at least.
"Setting SYCL dependencies" is more preferable and right approach. But, currently, SYCL backend API doesn't have possibility to pass the any dependencies. It should be re-designed in the  future... But now we can apply simple fix  - to wait of the first SYCL backend pattern call.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -948,7 +948,7 @@ __pattern_copy_if(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterato
     auto __res = __par_backend_hetero::__parallel_copy_if(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
                                                           __buf1.all_view(), __buf2.all_view(), __n, __pred);
 
-    ::std::size_t __num_copied = __res.get();
+    ::std::size_t __num_copied = __res.get(); //is a blocking call
     return __result_first + __num_copied;
 }
 
@@ -1028,8 +1028,11 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, 
 
     auto __copy_last = __pattern_copy_if(__tag, __exec, __first, __last, __copy_first, __not_pred<_Predicate>{__pred});
 
-    //TODO: optimize copy back depending on Iterator, i.e. set_final_data for host iterator/pointer
-    return __pattern_walk2(
+    //TODO: To optimize copy back depending on Iterator, i.e. set_final_data for host iterator/pointer
+    // __pattern_copy_if above may be async due to there is implicit synchronization on sycl::buffer and the accessors
+
+    //An explicit wait isn't required here because we have implicit synchronization on sycl::buffer destructor.
+    return __pattern_walk2</*_IsSync=*/std::false_type>(
         __tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper>(::std::forward<_ExecutionPolicy>(__exec)),
         __copy_first, __copy_last, __first, __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
 }
@@ -1049,7 +1052,9 @@ __pattern_unique(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _It
     auto __copy_last = __pattern_unique_copy(__tag, __exec, __first, __last, __copy_first, __pred);
 
     //TODO: optimize copy back depending on Iterator, i.e. set_final_data for host iterator/pointer
-    return __pattern_walk2</*_IsSync=*/::std::true_type, __par_backend_hetero::access_mode::read_write,
+
+    //An explicit wait isn't required here because we have implicit synchronization on sycl::buffer destructor.
+    return __pattern_walk2</*_IsSync=*/std::false_type, __par_backend_hetero::access_mode::read_write,
                            __par_backend_hetero::access_mode::read_write>(
         __tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper>(::std::forward<_ExecutionPolicy>(__exec)),
         __copy_first, __copy_last, __first, __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
@@ -1230,7 +1235,9 @@ __pattern_inplace_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __ex
         __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::write>(__copy_first), __comp);
 
     //TODO: optimize copy back depending on Iterator, i.e. set_final_data for host iterator/pointer
-    __pattern_walk2(
+
+    //An explicit wait isn't required here because we have implicit synchronization on sycl::buffer destructor.
+    __pattern_walk2</*_IsSync=*/std::false_type>(
         __tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper>(::std::forward<_ExecutionPolicy>(__exec)),
         __copy_first, __copy_last, __first, __brick_move<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
 }
@@ -1315,14 +1322,18 @@ __pattern_stable_partition(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& _
     auto true_count = copy_result.first - __true_result;
 
     //TODO: optimize copy back if possible (inplace, decrease number of submits)
-    __pattern_walk2</*_IsSync=*/::std::false_type>(
+    __pattern_walk2</*_IsSync=*/std::false_type>(
         __tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper>(__exec), __true_result, copy_result.first,
         __first, __brick_move<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
 
-    __pattern_walk2(
+    //We don't need synchronization between these patterns due to the data are being processed independently.
+
+    __pattern_walk2</*_IsSync=*/std::false_type>(
         __tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper2>(::std::forward<_ExecutionPolicy>(__exec)),
         __false_result, copy_result.second, __first + true_count,
         __brick_move<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
+
+    //An explicit wait isn't required here because we have implicit synchronization on sycl::buffer destructor.
 
     return __first + true_count;
 }
@@ -1486,11 +1497,17 @@ __pattern_partial_sort_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& 
     {
         // If our output buffer is larger than the input buffer, simply copy elements to the output and use
         // full sort on them.
-        auto __out_end = __pattern_walk2</*_IsSync=*/::std::false_type>(
-            __tag, __par_backend_hetero::make_wrapped_policy<__initial_copy_1>(__exec), __first, __last, __out_first,
-            __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
+        auto __out_end =
+            __pattern_walk2(__tag, __par_backend_hetero::make_wrapped_policy<__initial_copy_1>(__exec), __first, __last,
+                            __out_first, __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
 
-        // Use regular sort as partial_sort isn't required to be stable
+        // TODO: __pattern_walk2 is a blocking call here, so there is a synchronization between the patterns.
+        // But, when the input iterators are a kind of hetero iterator on top of sycl::buffer, SYCL
+        // runtime makes a dependency graph. In that case the call of __pattern_walk2 could be changed to 
+        // be asynchronous for better performance.
+
+        // Use regular sort as partial_sort isn't required to be stable.
+        //__pattern_sort is a blocking call.
         __pattern_sort(
             __tag,
             __par_backend_hetero::make_wrapped_policy<__partial_sort_1>(::std::forward<_ExecutionPolicy>(__exec)),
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1239,7 +1239,7 @@ __pattern_stable_partition(_ExecutionPolicy&& __exec, _Iterator __first, _Iterat
     auto true_count = copy_result.first - __true_result;
 
     //TODO: optimize copy back if possible (inplace, decrease number of submits)
-    __pattern_walk2</*_IsSync=*/::std::false_type>(
+    __pattern_walk2(
         __par_backend_hetero::make_wrapped_policy<copy_back_wrapper>(__exec),
```
danhoeflinger, line 1243: Here it seems that these two `__pattern_walk2` calls are independent of each other (don't require a specific ordering), but both do need to be waited on before returning from `__pattern_stable_partition`. 

Instead of enforcing ordering here, could we make them both async calls and then wait on the policy's queue here?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1411,7 +1411,7 @@ __pattern_partial_sort_copy(_ExecutionPolicy&& __exec, _InIterator __first, _InI
     {
         // If our output buffer is larger than the input buffer, simply copy elements to the output and use
         // full sort on them.
-        auto __out_end = __pattern_walk2</*_IsSync=*/::std::false_type>(
+        auto __out_end = __pattern_walk2(
             __par_backend_hetero::make_wrapped_policy<__initial_copy_1>(__exec), __first, __last, __out_first,
```
danhoeflinger, line 1415: This `__pattern_walk2` and the subsequent `__pattern_sort` do have a dependency with `__out_first`, where `__pattern_walk2` needs to be finished before `__pattern_sort` can begin.   

My first thought is that the accessors involved here provide the implicit data dependency to order these kernels appropriately and make this work as a pipeline (as described in the `_IsSync` comment in the definition of `__pattern_walk2`).

However, with USM pointers as input data, I believe we lose these implicit dependencies and the ordering of kernels.  Is that also your understanding (and the reason this extra synchronization is required)?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1545,20 +1545,22 @@ __pattern_rotate(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __new_f
...
 
-    auto __temp_rng =
+    auto __temp_rng_w =
         oneapi::dpl::__ranges::all_view<_Tp, __par_backend_hetero::access_mode::write>(__temp_buf.get_buffer());
 
     const auto __shift = __new_first - __first;
     oneapi::dpl::__par_backend_hetero::__parallel_for(
         oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__rotate_wrapper>(__exec),
         unseq_backend::__rotate_copy<typename ::std::iterator_traits<_Iterator>::difference_type>{__n, __shift}, __n,
-        __buf.all_view(), __temp_rng);
```
danhoeflinger, line 1555: This falls in a similar category as the `partial_sort_copy` case above.

For buffers, I believe the implicit data dependencies of the accessor to `__temp_buf` should order these appropriately, and the wait on the last `__parallel_for` should remove any race with the return.

However, considering USM pointer inputs I think we lose those and this can be a bug that requries the `wait()`

danhoeflinger (commented):
I agree with you that the long term fix is to use explicit dependencies between these pattern calls, and that sort of refactor is necessary.  

However, I think that we may be able to do slightly better with the `__pattern_stable_partition` case as described in my comment.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
MikeDvorskiy: I agree  here...
Yes, we have to call wait for two events here. But we still don't have such oneDPL API. 
Usage of SYCL API directly is not allowed on this oneDPL code layer.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
MikeDvorskiy: Actually, the case is more complicated what we see...
1) USM: ideal solution  - via `depends_on(first_event)`  (preferable) or wait() on the first event (future).
2) hetero iterators - you are right  - we have implicit SYCL data access dependency (data dependency graph) and we don't need explicit synch
3) host iterators: due to the first pattern contains a temporary `sycl::buffer `in its own scope, we cannot rely on SYCL data dependency graph. On the second pattern call we have got another temporary `sycl::buffer` and SYCL Runtime will not be aware that it is the same data, I think.

danhoeflinger: That makes sense.  For us to rely on implicit dependencies, we need only a single call which is reused to `__get_sycl_range`.  However, this is something that should be important anyway to avoid extra copies to and from the device with host data.  

Perhaps that rewrite does not belong in this PR, but it seems that may be something we need to audit.  If we are using multiple high level patterns with host iterators, we may be copying data to and from the device for each pattern unnecessarily.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: Yes, you are correct.  I agree that similar to the others, this requires a rewrite / refactor to improve by returning some event from `__pattern_walk2` rather than the simple true / false async we have.

danhoeflinger (approved):
After taking another look through this,  I agree that this should be applied as is (since we cannot improve it further without significant refactoring to provide some method of pipelining patterns via events.  It may result in a slow down in some usages, but will fix bugs in others, which is the preferred option between the two.


We should create an issue to properly describe the intricacies of the requirements for a refactor, specifically from this topic: 
 https://github.com/oneapi-src/oneDPL/pull/1261/files/2a9d0f0af2f64d582f9b0d770d94fe3085ea4112#r1383348739

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: Yes, it is not a great situation, but you are correct because `__temp_buf` is a temporary `sycl::buffer` these waits are unnecessary.  Assuming we remove both `wait()`,  we should add comments for each to describe the implicit synchronization as it is not clear. (I missed it even with this topic in my mind when looking back at this PR)

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1486,9 +1485,9 @@ __pattern_partial_sort_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&&
...
         // full sort on them.
-        auto __out_end = __pattern_walk2</*_IsSync=*/::std::false_type>(
-            __tag, __par_backend_hetero::make_wrapped_policy<__initial_copy_1>(__exec), __first, __last, __out_first,
-            __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
+        auto __out_end =
+            __pattern_walk2(__tag, __par_backend_hetero::make_wrapped_policy<__initial_copy_1>(__exec), __first, __last,
+                            __out_first, __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
 
         // Use regular sort as partial_sort isn't required to be stable
         __pattern_sort(
```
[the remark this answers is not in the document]
danhoeflinger: So, since we are not wanting to do some sort of large refactor to use event dependencies to pipeline patterns at this point...

What about something like this to avoid the perf degradation:
Create something like `__type_requires_exterior_synchronization<__out_first>` which would resolve to `true_type` for anything for non-sycl backend, and for sycl backend it could return `bool_constant<!is_sycl_iterator_v<__out_first>>`.

We can do this without mentioning sycl at this level.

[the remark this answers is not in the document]
danhoeflinger: This still leaves a lot to be desired vs the refactor (for host_iterators especially), but it at least removes the degradation.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: What about if we rewrite this code like
`C++
    auto future1 = __pattern_walk2_async(__tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper>(__exec),
                                         __true_result, copy_result.first, __first,
                                         __brick_move<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});

    auto future2 = __pattern_walk2_async(
        __tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper2>(::std::forward<_ExecutionPolicy>(__exec)),
        __false_result, copy_result.second, __first + true_count,
        __brick_move<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});

    future1.wait();
    future2.wait();
```

File: include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h (hunk 4 above)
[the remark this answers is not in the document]
Comment (MikeDvorskiy) [reply include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h:1493]: `is_sycl_iterator_v`...
What if we have a "fancy pipe" over a sycl_iterator? or what should we do in case of mix? zip(zip(sycl_iter, usm_ptr), usm_ptr) and so on?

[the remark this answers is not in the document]
Comment (danhoeflinger) [reply include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h:1493]: Yes, it would miss these, its not a perfect solution.  It provides some improvement though for raw sycl_iterator or legacy hetero iterators. The better solution is using events which can be passed as dependencies to patterns and then into sycl kernels.  

The solution described could be made better by adding a `__contains_sycl_iterator` which has specializations for our fancy iterators to pass through and check if any iterator child `__contains_sycl_iterator` and would therefore provide the synchronization.
zip and permutation would return the OR of their children iterators, transform would return its base's result.  If it is not a fancy iterator, it would just be defined as described above.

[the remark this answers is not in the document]
Comment (danhoeflinger) [reply include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h:1493]: Actually its a bit more complicated because we need the access pattern to actually dictate a dependency.  Probably permutation iterator should only pass-through to its base, not its map, because the map will be read-only.

It is of course the user of this structure's responsibility to make sure the usage of the sequence will generate the synchronization, but permutation iterator wont require write access for it's map even if it is used in write mode.

Read-only followed by read-only wouldn't require an ordering.  We need to be most careful to avoid believing there is a synchronization where there isn't one.

File: include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h [hunk 5]
```diff
@@ -1621,21 +1639,27 @@ __pattern_rotate(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator
...
 
     const auto __shift = __new_first - __first;
     oneapi::dpl::__par_backend_hetero::__parallel_for(
         _BackendTag{}, oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__rotate_wrapper>(__exec),
         unseq_backend::__rotate_copy<typename ::std::iterator_traits<_Iterator>::difference_type>{__n, __shift}, __n,
-        __buf.all_view(), __temp_rng);
+        __buf.all_view(), __temp_rng_w);
+
+    //An explicit wait doesn't need here because we are working with temporary sycl::buffer and sycl accessors and
+    //SYCL runtime makes a dependecny graph to prevent the races between two __parallel_for patterns.
```
danhoeflinger, line 1652:
```suggestion
    //An explicit wait isn't required here because we are working with a temporary sycl::buffer and sycl accessors and
    //SYCL runtime makes a dependency graph to prevent the races between two __parallel_for patterns.
```

There are a number of these in the new comments to correct.  
"doesn't need" ->  "isn't required"
"dependecny" -> "dependency"

danhoeflinger (commented):
I believe I agree with your removal of these synchronizations 
From 
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:buf-sync-rules

"When the buffer is destroyed, the destructor will block until all work in queues on the buffer have completed, then copy the contents of the buffer back to the host memory (if required) and then return."
It doesn't matter what mode the accessor is in, it will always block on the destructor.  

Other than the performance degradation for that one case, this looks good.  We can discuss in that thread if we want to try to fix that in this PR or leave as an issue / TODO.

MikeDvorskiy:
> leave as an issue / TODO.

I've Ieft a proper //TODO

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1486,11 +1497,16 @@ __pattern_partial_sort_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&&
...
-            __tag, __par_backend_hetero::make_wrapped_policy<__initial_copy_1>(__exec), __first, __last, __out_first,
-            __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
+        auto __out_end =
+            __pattern_walk2(__tag, __par_backend_hetero::make_wrapped_policy<__initial_copy_1>(__exec), __first, __last,
+                            __out_first, __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
 
-        // Use regular sort as partial_sort isn't required to be stable
+        // TODO: __pattern_walk2 is a blocking call here, so there is a synchronization between the patterns.
+        // But, when the input iterators are a kind of hetero iterator on top of sycl::buffer, SYCL
+        // runtime makes a dependency graph. In that case the call of __pattern_walk2 may be unblocking.
```
danhoeflinger, line 1506:
```suggestion
        // TODO: __pattern_walk2 is a blocking call here, so there is a synchronization between the patterns.
        // But, when the input iterators are a kind of hetero iterator on top of sycl::buffer, SYCL
        // runtime makes a dependency graph. In that case the call of __pattern_walk2 could be changed to 
        // be asynchronous for better performance.
```

danhoeflinger (approved):
LGTM

Issue #1253, opened 2023-10-28
`transform_inclusive_scan` always runs serially with `std::ranges::iota_view::begin/end` iterators


akrivx:
I'm using the TBB parallel backend of oneAPI 2023.2 with GCC 12.2.0 on Linux. I've also enabled C++ 20.

I have noticed that passing the `begin`/`end` iterators of a standard `iota_view` to `oneapi::dpl::transform_inclusive_scan` (I have yet to test other algorithms) always results in serial execution, even when the execution policy is `oneapi::dpl::execution::par`.

Replacing `r.begin()`/`r.end()` with `oneapi::dpl::counting_iterator`s works as expected, i.e. runs in parallel with `oneapi::dpl::execution::par`.

It's worth noting that `r.begin()` and `r.end()` in the following example are random-access iterators:

```cpp
const int n = 1'000'000;

// Always runs serially:
auto r = std::views::iota(0, n);
static_assert(std::random_access_iterator<decltype(r.begin())>); // OK
static_assert(std::random_access_iterator<decltype(r.end())>); // OK
oneapi::dpl::transform_inclusive_scan(
  oneapi::dpl::execution::par,
  r.begin(), r.end(),
  // ...
);

// Runs in parallel:
auto it = oneapi::dpl::counting_iterator<int>{0};
oneapi::dpl::transform_inclusive_scan(
  oneapi::dpl::execution::par,
  it, it + n,
  // ...
);
```

Do you know why that is?

mmichel11:
Hello, @akrivx. Thank you very much for reporting this issue. We have confirmed the problem you are encountering and have identified the cause in our internal random access iterator checks. We are discussing the best way to address this issue and will have a full response in the upcoming weeks.

akukanov:
Supposedly fixed by #1266

PR #1239, opened 2023-10-20
Implementation of tag dispatching on current codebase

SergeyKopienko:
Based on

PR #1266, opened 2023-11-09
Fix serial execution of parallel algorithms with C++20 iterators

kboyarinov:
I propose this PR as a fix for #1253.

Our internal random access iterator checks relied on `std::iterator_traits::iterator_category` type. 
But for C++20 iterators, like `std::ranges::iota_view::iterator`, `iterator_category` is equivalent to `std::input_iterator_tag` even if the iterator is random access. For these modern iterators, the real iterator category should be observed using `std::iterator_traits::iterator_concept`. 
This PR modifies `dpl::__internal::__is_random_access_iterator` traits to use `std::random_access_iterator` concept since C++20. 
This concept respects both `iterator_concept` (for modern iterators) and `iterator_category` (for "legacy" iterators).

This PR also removes `dpl::__internal::__iterator_traits` helper since `std::iterator_traits` is SFINAE friendly since C++17 and hence internal structure is redundant now.


SergeyKopienko:
@kboyarinov you may take a look to https://github.com/oneapi-src/oneDPL/pull/1239
There is some similar approach from @rarutyun

kboyarinov:
> @kboyarinov you may take a look to #1239 There is some similar approach from @rarutyun
> 
> * this file: https://github.com/oneapi-src/oneDPL/pull/1239/files#diff-deed20001ff532923c71880283f45817ed065a09752a00c0da1192ee65655fef

For me these PRs and approaches are orthogonal. The approach in the PR you mentioned is better than the existing one, but does not fix the problem described. It is required to complete it with the fix similar to this one. But I guess these PRs and their goals should be considered separately.

MikeDvorskiy (commented):
It looks reasonable with my perspective.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -114,6 +114,80 @@ struct sycl_iterator
         return *this - it < 0;
     }
 
+    bool
```
MikeDvorskiy, line 117: I think that relying on
`using iterator_category = ::std::random_access_iterator_tag;`
is enough. The iterator should not  be a "honest" random_access_iterator concept, IMHO. Also, oneAPI spec and doc doesn't require that.

include/oneapi/dpl/pstl/experimental/internal/for_loop_impl.h
```diff
@@ -269,9 +269,9 @@ __execute_loop_strided(_Ip __first, _Ip __last, _Function __f, _Sp __stride, _Pa
 }
 
 template <typename _Ip, typename _Function, typename _Sp, typename _Pack, typename _IndexType>
-::std::enable_if_t<::std::is_same_v<typename oneapi::dpl::__internal::__iterator_traits<_Ip>::iterator_category,
+::std::enable_if_t<::std::is_same_v<typename ::std::iterator_traits<_Ip>::iterator_category,
```
MikeDvorskiy, line 272: Do we whether want to use `is_base_of(input_iterator tag) ` here?
Because a 'forward' iterator is an 'input' iterator as well.

kboyarinov: I have checked the code and `__execute_loop_strided` only uses for bidirectional and forward iterators. I have changed the code a bit to use the newly introduced trait and removed `enable_if` here.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
kboyarinov: It is not enough, because C++20 allows defining the type of iterator using `iterator_concept` type alias. If we would check for `iterator_category` only, all "modern" iterators (in particular from C++ ranges) would be threated as input iterators.

include/oneapi/dpl/pstl/iterator_defs.h
```diff
@@ -27,25 +27,80 @@ namespace dpl
...
+{
+};
 
-template <typename... _IteratorTypes>
-auto
-__is_iterator_of(...) -> ::std::false_type;
+template <typename _IteratorTag, typename _IteratorType>
+struct __is_iterator_of_category_dispatch<_IteratorTag, _IteratorType,
+                                          ::std::void_t<typename ::std::iterator_traits<_IteratorType>::iterator_category>>
+: ::std::is_base_of<_IteratorTag, typename ::std::iterator_traits<_IteratorType>::iterator_category>
```
akukanov, line 40: We stopped using colons in front of `std` in the new code.

akukanov (approved):
The patch looks OK to me. I wonder if the dispatch could somehow be "flattened" instead of its current "hierarchical" set of classes, but that certainly is not critical.

PR #1270, opened 2023-11-10
[KT] Generate test cases for key-value sort test

dmitriy-sobolev:
Introduce random generation of cases to limit their number.

Line endings have been corrected in the whole `test/kt/esimd_radix_sort_utils.h` file. Please, ignore whitespace changes during the review.


test/kt/esimd_radix_sort.cpp
```diff
@@ -198,41 +178,25 @@ can_run_test(sycl::queue q, KernelParam param)
...
+                test_general_cases<TEST_KEY_TYPE, Descending, TestRadixBits>(q, size, kernel_parameters);
             }
-#endif // TEST_LONG_RUN
+            test_small_sizes<TEST_KEY_TYPE, Ascending, TestRadixBits>(q, kernel_parameters);
         }
         catch (const ::std::exception& exc)
         {
             std::cerr << "Exception: " << exc.what() << std::endl;
-            return EXIT_FAILURE;
+            return 1;
```
SergeyKopienko, line 197: For that this change required?

test/kt/esimd_radix_sort_by_key.cpp
```diff
@@ -94,35 +91,31 @@ void test_usm(std::size_t size, KernelParam param)
...
+                test_usm<TEST_KEY_TYPE, TEST_VALUE_TYPE, Descending, TestRadixBits, sycl::usm::alloc::shared>(
+                    q, size, kernel_parameters);
+                test_sycl_iterators<TEST_KEY_TYPE, TEST_VALUE_TYPE, Ascending, TestRadixBits>(q, size, kernel_parameters);
+                test_sycl_iterators<TEST_KEY_TYPE, TEST_VALUE_TYPE, Descending, TestRadixBits>(q, size, kernel_parameters);
+            }
+        }
+        catch (const ::std::exception& exc)
+        {
+            std::cerr << "Exception: " << exc.what() << std::endl;
+            return 1;
```
SergeyKopienko, line 116: Why return not predefined const?

test/kt/esimd_radix_sort_utils.h
```diff
@@ -1,268 +1,296 @@
...
+#include <cmath>
+#include <limits>
+#include <iostream>
+#include <cstdint>
+#include <vector>
+#include <type_traits>
+
+#if defined(TEST_DATA_PER_WORK_ITEM) && defined(TEST_WORK_GROUP_SIZE)
+using ParamType = oneapi::dpl::experimental::kt::kernel_param<TEST_DATA_PER_WORK_ITEM, TEST_WORK_GROUP_SIZE>;
+constexpr ParamType kernel_parameters;
```
SergeyKopienko, line 38: What if `kernel_parameters` is undefined? Looks like we will have compile error...

SergeyKopienko: This solution mean that the Kernel template param now is the same for sort and sort by key algorithms.
Are you sure that this two algorithms still have the same structure of Kernel template params?

test/kt/esimd_radix_sort.cpp
dmitriy-sobolev: I've made the change in order not to include `<cstdlib>`. This is a minor change, I do not mind to roll it back and include `<cstdlib>`.

test/kt/esimd_radix_sort_by_key.cpp
dmitriy-sobolev: Could you describe the necessity of it?

SergeyKopienko: The necessity of it is very simple - to have someone const which used as return code if test failed. This const already used inside `issue_error_message` function (called when some test results is incorrect):
```C++
inline void
issue_error_message(::std::stringstream& outstr)
{
    outstr << ::std::endl;
    ::std::cerr << outstr.str();
    ::std::exit(EXIT_FAILURE);
}
```
So please do not remove it's usage in esimd tests too.

test/kt/esimd_radix_sort.cpp
SergeyKopienko: See my comments below

test/kt/esimd_radix_sort_utils.h
dmitriy-sobolev: > What if kernel_parameters is undefined? Looks like we will have compile error...

Currently, the tests are designed to strictly require `TEST_DATA_PER_WORK_ITEM` and `TEST_WORK_GROUP_SIZE` being externally defined. The dependency on these macros is a bit convoluted because of being scattered across multiple files. I can get rid of this alias to simplify the logic.

> This solution mean that the Kernel template param now is the same for sort and sort by key algorithms. Are you sure that this two algorithms still have the same structure of Kernel template params?

`oneapi::dpl::experimental::kt::kernel_param` is a genetic structure supposed to be used with any kernel template, be it esimd sort, other sort, scan and etc. The test harness passes different `TEST_DATA_PER_WORK_ITEM` and `TEST_WORK_GROUP_SIZE` when compiling different tests. So there is no dependency between `radix_sort` and `radix_sort_by_key` algorithms on the same `TEST_DATA_PER_WORK_ITEM` and `TEST_WORK_GROUP_SIZE` configuration.

SergeyKopienko: ok

test/kt/esimd_radix_sort.cpp
dmitriy-sobolev: I have returned it

test/kt/esimd_radix_sort_by_key.cpp
dmitriy-sobolev: It's in use now

PR #1271, opened 2023-11-13
Cancel existing workflows for PR when new commit is pushed

danhoeflinger:
This change cancels all existing workflows queued or actively running for the same PR, in favor of only testing the most recent commit.

People need to be aware that this is being changed.  As it stands there is no way to turn this off via a commit message.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,10 @@ env:
   TEST_TIMEOUT: 360
   MACOS_ONEAPI_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/irc_nas/18358/m_cpp-compiler-classic_p_2022.0.0.62_offline.dmg
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   clang-format:
     if: github.event_name == 'pull_request'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ env:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
   clang-format:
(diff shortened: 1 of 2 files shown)


SergeyKopienko (approved):
LGTM

.github/workflows/ci.yml
```diff
@@ -15,6 +15,10 @@ env:
   TEST_TIMEOUT: 360
   MACOS_ONEAPI_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/irc_nas/18358/m_cpp-compiler-classic_p_2022.0.0.62_offline.dmg
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
```
dmitriy-sobolev, line 21: Will it also cancel runs for the main branch? If so, it is better add the main branch as an exception.

danhoeflinger: I believe since it references the `github.event.pull_request.number` it should not touch the workflows on the main branch.  I have not tested this though.

SergeyKopienko: > Will it also cancel runs for the main branch? If so, it is better add the main branch as an exception.

Why it's needed?

danhoeflinger: Oh actually, I understand the question.  Yes this may be a problem now that I think about it.  Thanks for bringing it up.

If two PRs are merged in quick succession, we don't want the second merged to cancel the first's run.  Let me look into this.  I'll try to find documentation about what those variables are for main branch, and how we might make an exception here.

SergeyKopienko: You mean we should have one build for each commit in the `main` branch?

danhoeflinger: We don't want to cancel post-merge CI runs on main.  @dmitriy-sobolev  provided a solution: https://stackoverflow.com/questions/68418857/how-to-cancel-existing-runs-when-a-new-push-happens-on-github-actions-but-only/70972844#70972844

dmitriy-sobolev: > > Will it also cancel runs for the main branch? If so, it is better add the main branch as an exception.
> 
> @dmitriy-sobolev Why it's needed?

I do not have strong reasoning, but `main` branch has more "coarse-grained" commits compared to feature branches, it also should not have "intermediate" states, which we can skip. It will also facilitate finding regressions if any.

dmitriy-sobolev (approved):
LGTM

SergeyKopienko (approved):
LGTM

PR #1278, opened 2023-11-16
fix compile issue

rscohn2:
DR library needs to support 2023.2 and 2024.0 compilers. This change is needed to compile clean with 2023.2 compiler.


julianmi:
@rscohn2 Thank you for raising this. Please have a look at the refactoring in #1225 that we merged on Monday. Please let me know if you still see compilation issues with the current mainline.

rscohn2:
It looks like your commit compiles with 2023 & 2024. Thanks.

PR #1280, opened 2023-11-16
Adding sourcing of vars.sh to linux testing

danhoeflinger:
When installing from conda, we did not source vars.sh previously.  This adds that to the script(s).


danhoeflinger:
This is not the right approach for conda.  Needs more investigation.

PR #1283, opened 2023-11-17
[pstl-offload] Add vars generation stage for PSTL offload build and source vars while testing with oneAPI compiler 2024.0

kboyarinov:
Fix PSTL offload CI jobs to work with 2024.0 compiler and later.

The problem was that previously pstl offload was tested in the "manual" mode (due to absence of the compiler supporting -fsycl-pstl-offload) - include directories and libpstloffload paths substituted manually into the build command.

The compiler of 2024.0 version supports -fsycl-pstl-offload allows to test "honestly" by specifying only `-fsycl-pstl-offload` compiler option. But setting proper DPL_ROOT, LIBRARY_PATH and LD_LIBRARY_PATH is required to make it work.
On the user side, it is made by sourcing the vars script from oneDPL package. During our testing, we did not have vars script and hence did not set variables mentioned above. 

This PR adds generation of simple temporary vars script that should be sourced before running the testing.

I have also added `_ONEDPL_PSTL_OFFLOAD_BINARY_PATH` variable for enabling into other CI workflows when the binary is building once and others should not be built.


Alexandr-Konovalov (approved):
LGTM

dmitriy-sobolev (approved):
LGTM as a fix.

You may also consider other approaches:
 - Simpler one. Get rid of the generated vars scripts in favor of setting the required env variables inside CMake.
 - Apparently a more conventional one. Use [install ](https://cmake.org/cmake/help/latest/command/install.html) command to deploy the library with built binaries and relevant vars scripts.

PR #1286, opened 2023-11-20
Fixing transform_if test for unnamed lamdba

danhoeflinger:
Fixing unnamed lambda in transform_if tests. (same mangled name)


SergeyKopienko (approved):
LGTM

PR #1287, opened 2023-11-20
[CI] Fix environment for FPGA emulator testing

dmitriy-sobolev:
**Summary**
`dpcpp_linux-64` `2024.0.0`, in contrast to `2023.2.0`, lacks some files which allow discovery of FPGA emulator device.
This PR switches testing of FPGA emulator from the latest compiler version (`2024.0.0`) to the previous (`2023.2.0`) where FPGA emulator is available.

**Notes**
I was able to fix the environment for the emulator on `dpcpp_linux-64` `2024.0.0` locally following these steps:
 - copy `cl.cfg` as `cl.fpga_emu.cfg`
 - set `OCL_ICD_FILENAMES` to point the location of `libintelocl_emu.so*` or write this location into `$CONDA_PREFIX/etc/OpenCL/vendors/intel-fpga_emu.icd`. 

But it does not work with the environment on a github-hosted runner. Let me know if it still requires investigation. Other possible option is to use other, OS specific package managers, e.g. APT.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,6 +180,11 @@ jobs:
           export PATH=$CONDA/lib:$PATH
           export CPATH=$CONDA/include:$CPATH
           export OCL_ICD_FILENAMES=$CONDA/lib/libintelocl.so
+          
+          # Workaround for partially absent FPGA emulator runtime in https://anaconda.org/intel/intel-opencl-rt 2024.0.0 package (part of dpcpp_linux-64)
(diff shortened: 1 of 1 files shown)


adamfidel (approved):
I feel that this solution is sufficient for now to get testing off the ground. However, I think it would be good if we create an issue to capture that the FPGA_EMU path is broken with 2024.0.0 so that we do not lose track of this problem.

danhoeflinger (approved):
I agree with @adamfidel here.  I think an issue should be created, but this LGTM.  
Thanks for looking into this @dmitriy-sobolev.

PR #1292, opened 2023-11-23
[KT] Test API with sycl::buffer

dmitriy-sobolev:
`test_sycl_iterators` was replaced by `test_buffer` in `esimd_radix_sort_by_key` to tests both APIs: ranges and iterators (not only iterators).

### DIFF
--- a/test/kt/esimd_radix_sort.cpp
+++ b/test/kt/esimd_radix_sort.cpp
@@ -136,6 +136,26 @@ test_sycl_iterators(sycl::queue q, std::size_t size, KernelParam param)
 
 template <typename T, bool IsAscending, std::uint8_t RadixBits, typename KernelParam>
 void
+test_buffer(sycl::queue q, std::size_t size, KernelParam param)
+{
+#if LOG_TEST_INFO
+    std::cout << "\t\ttest_buffer<" << TypeInfo().name<T>() << ">(" << size << ");" << std::endl;
+#endif
+    std::vector<T> input(size);
+    generate_data(input.data(), size, 42);
+    std::vector<T> ref(input);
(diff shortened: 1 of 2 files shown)


test/kt/esimd_radix_sort.cpp
```diff
@@ -134,6 +134,26 @@ test_sycl_iterators(sycl::queue q, std::size_t size, KernelParam param)
     EXPECT_EQ_RANGES(ref, input, msg.c_str());
 }
 
+template <typename T, bool IsAscending, std::uint8_t RadixBits, typename KernelParam>
+void
+test_buffer(sycl::queue q, std::size_t size, KernelParam param)
```
SergeyKopienko, line 139: `test_host_buffer` ?

test/kt/esimd_radix_sort_by_key.cpp
```diff
@@ -29,7 +29,7 @@
 #include "../support/sycl_alloc_utils.h"
 
 template<typename KeyT, typename ValueT, bool isAscending, std::uint32_t RadixBits, typename KernelParam>
-void test_sycl_iterators(sycl::queue q, std::size_t size, KernelParam param)
+void test_buffer(sycl::queue q, std::size_t size, KernelParam param)
```
SergeyKopienko, line 32: `test_host_buffer` ?

test/kt/esimd_radix_sort.cpp
dmitriy-sobolev: There is only `sycl::buffer`, so `test_host_buffer` would incur confusion. Perhaps, it is better to rename the function to `test_sycl_buffer`.

SergeyKopienko (approved):
LGTM

Issue #1294, opened 2023-11-23
_ONEDPL_PRAGMA_UNROLL is undefined on open-source DPC++


joeatodd:
While investigating performance of our new scan implementation, I found that `_ONEDPL_PRAGMA_UNROLL` is undefined for open-source DPC++.

I am using a recent build of the open-source compiler:
`clang version 18.0.0 (git@github.com:intel/llvm.git 05e7395ebe5333048b851dea608260b51cec5f69)`

I suspect this is because the reported GCC version is lower than it ought to be, as per this comment in `onedpl_config.h`:
```cpp
// note that when ICC or Clang is in use, _ONEDPL_GCC_VERSION might not fully match
// the actual GCC version on the system.
#define _ONEDPL_GCC_VERSION (__GNUC__ * 10000 + __GNUC_MINOR__ * 100 + __GNUC_PATCHLEVEL__)
```

If I run `echo | clang++ -fsycl -dM -E -x c++ - | grep -i '__GNUC'` on my machine I get:
```cpp
#define __GNUC_GNU_INLINE__ 1
#define __GNUC_MINOR__ 2
#define __GNUC_PATCHLEVEL__ 1
#define __GNUC__ 4
```
which means that this:
```cpp
// Enable loop unrolling pragmas where supported
#if (__INTEL_COMPILER || (!defined(__INTEL_COMPILER) && _ONEDPL_GCC_VERSION >= 80000))
#    define _ONEDPL_PRAGMA_UNROLL _ONEDPL_PRAGMA(unroll)
#else //no pragma unroll
#    define _ONEDPL_PRAGMA_UNROLL
#endif
```
makes `_ONEDPL_PRAGMA_UNROLL` undefined.

I also confirmed this empirically by replacing `_ONEDPL_PRAGMA_UNROLL` with `#pragma unroll` in my kernel and I got a consistent speedup.

joeatodd:
Could we use `_ONEDPL_CLANG_VERSION` to fix this? Does anyone know which was the first version of `clang` would be the earliest which supports pragma unroll?

joeatodd:
[...]

PR #1299, opened 2023-11-29
Enable pragma unroll for open-source DPC++

joeatodd:
This PR attempts to address #1294. Not sure if it's the best approach, but it should definitely catch all `clang` based compilers which support pragma unroll.

I haven't worked out a portable way to add a warning when undefined as per @jinz2014's suggestion on the issue.

### DIFF
--- a/include/oneapi/dpl/pstl/onedpl_config.h
+++ b/include/oneapi/dpl/pstl/onedpl_config.h
@@ -118,7 +118,7 @@
 // Enable loop unrolling pragmas where supported
 #if (__INTEL_LLVM_COMPILER || __INTEL_COMPILER ||                                                                      \
      (!defined(__INTEL_LLVM_COMPILER) && !defined(__INTEL_COMPILER) &&                                                 \
-      ((_ONEDPL_GCC_VERSION >= 80000) || (_ONEDPL_CLANG_VERSION >= 40000))))
+      ((_ONEDPL_GCC_VERSION >= 80000) || (_ONEDPL_CLANG_VERSION >= 37000))))
 #    define _ONEDPL_PRAGMA_UNROLL _ONEDPL_PRAGMA(unroll)
 #else //no pragma unroll
 #    define _ONEDPL_PRAGMA_UNROLL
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -117,7 +117,8 @@
 
 // Enable loop unrolling pragmas where supported
 #if (__INTEL_LLVM_COMPILER || __INTEL_COMPILER ||                                                                      \
-     (!defined(__INTEL_LLVM_COMPILER) && !defined(__INTEL_COMPILER) && _ONEDPL_GCC_VERSION >= 80000))
+     (!defined(__INTEL_LLVM_COMPILER) && !defined(__INTEL_COMPILER) &&                                                 \
+      ((_ONEDPL_GCC_VERSION >= 80000) || (_ONEDPL_CLANG_VERSION >= 40000))))
```
adamfidel, line 121: Maybe we can use the version `37000`? I can see in the [Clang 3.7 docs](https://releases.llvm.org/3.7.0/tools/clang/docs/AttributeReference.html#pragma-unroll-pragma-nounroll) that `#pragma unroll` is valid whereas it is absent in the [Clang 3.6 docs](https://releases.llvm.org/3.6.0/tools/clang/docs/AttributeReference.html).

joeatodd, line 121:
```suggestion
      ((_ONEDPL_GCC_VERSION >= 80000) || (_ONEDPL_CLANG_VERSION >= 37000))))
```

joeatodd: Good point! Done.

adamfidel (approved):
I added some other reviewers to take a quick look, but this looks good to me.

include/oneapi/dpl/pstl/onedpl_config.h
danhoeflinger: Should this instead be 30700?

This is how it is defined:
`#    define _ONEDPL_CLANG_VERSION (__clang_major__ * 10000 + __clang_minor__ * 100 + __clang_patchlevel__)`

PR #1307, opened 2023-12-05
[oneDPL][sycl][test] + no op for hetero iterator due to sycl::buffer with non-trivially copyable type is not device copyable

MikeDvorskiy:
+ "no op" for hetero iterator due to sycl::buffer with non-trivially copyable type is not device copyable; see SYCL spec for more details.

### DIFF
--- a/test/general/sycl_iterator/sycl_iterator_for.pass.cpp
+++ b/test/general/sycl_iterator/sycl_iterator_for.pass.cpp
@@ -79,9 +79,22 @@ struct SyclTypeWrapper
     T __value;
 
     explicit SyclTypeWrapper(const T& value = T{4}) : __value(value) {}
-    ~SyclTypeWrapper() { __value = -2; }
     bool
     operator==(const SyclTypeWrapper& other) const
+    {
+        return __value == other.__value;
+    }
+};
+
+template <typename T>
+struct SyclTypeWrapperD
+{
+    T __value;
+
+    explicit SyclTypeWrapperD(const T& value = T{4}) : __value(value) {}
+    ~SyclTypeWrapperD() { __value = -2; }
+    bool
+    operator==(const SyclTypeWrapperD& other) const
     {
         return __value == other.__value;
     }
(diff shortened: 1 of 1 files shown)


test/general/sycl_iterator/sycl_iterator_for.pass.cpp
```diff
@@ -243,10 +256,26 @@ DEFINE_TEST(test_uninitialized_value_construct_n)
...
+
 DEFINE_TEST(test_destroy)
 {
     DEFINE_TEST_CONSTRUCTOR(test_destroy)
 
+    template <typename Policy, typename Iterator1, typename Size>
+    void
+    operator()(Policy&& exec, __it_test_type<Iterator1> first1, __it_test_type<Iterator1> last1, Size n)
+    {
+	if constexpr (std::is_trivially_copyable_v<TestValueType>)
```
SergeyKopienko, line 270: Incorrect formatting

test/general/sycl_iterator/sycl_iterator_for.pass.cpp
```diff
@@ -274,6 +303,20 @@ DEFINE_TEST(test_destroy_n)
 {
     DEFINE_TEST_CONSTRUCTOR(test_destroy_n)
 
+
+    template <typename Policy, typename Iterator1, typename Size>
+    void
+    operator()(Policy&&, __it_test_type<Iterator1>, __it_test_type<Iterator1>, Size)
```
SergeyKopienko, line 309: Usually in tests we have sometimes two `operator()` implementations with `std::enble_if_t`.
In this case - one will have
```C++
::std::enable_if_t<std::is_trivially_copyable_v<TestValueType>>
```
and the second simple `void`

test/general/sycl_iterator/sycl_iterator_for.pass.cpp
SergeyKopienko: The same like in https://github.com/oneapi-src/oneDPL/pull/1307/files#r1415847734

test/general/sycl_iterator/sycl_iterator_for.pass.cpp
MikeDvorskiy: Actually, both  approach are correct.
Since we support C++14/17 `if contsexpr` also used construction in oneDPl code.

test/general/sycl_iterator/sycl_iterator_for.pass.cpp
MikeDvorskiy: Actually,
```
@github-actions
oneDPL CI / clang-format (pull_request) Successful in 11s
```

SergeyKopienko (approved):
LGTM

test/general/sycl_iterator/sycl_iterator_for.pass.cpp
MikeDvorskiy: The comment is not actual any more - the code was changed.

PR #1310, opened 2023-12-08
[pstl-offload] Support free() call while obtaining system free() address

Alexandr-Konovalov:
free() call is possible inside dlsym(). Because global free() is overloaded, this may lead to crash. The situation is detected via TLS flag and releasing of object from nested free() calls is delayed till obtaining system free() address.

### DIFF
--- a/src/pstl_offload.cpp
+++ b/src/pstl_offload.cpp
@@ -31,8 +31,13 @@ namespace __pstl_offload
 
 using __free_func_type = void (*)(void*);
 
+// list of objects for delayed releasing
 struct __delayed_free_header {
     __delayed_free_header* _M_next;
+    void*                  _M_to_free;
+
+    __delayed_free_header(__delayed_free_header* __next, void* __to_free)
+        : _M_next(__next), _M_to_free(__to_free) {}
 };
 
 // are we inside dlsym call?
@@ -78,26 +83,29 @@ __internal_free(void* __user_ptr)
         {
             if (__dlsym_called)
             {
-                // delay releasing till exit of dlsym
-                __delayed_free_header* __header = static_cast<__delayed_free_header*>(__user_ptr);
-                __header->_M_next = __delayed_free? __delayed_free : nullptr;
-                __delayed_free = __header;
+                // Delay releasing till exit of dlsym. We do not overload malloc globally,
+                // so can use it safely. Do not use new to able to use free() during
+                // __delayed_free_header releasing.
+                void* __buf = malloc(sizeof(__delayed_free_header));
+                __delayed_free_header* __h = new(__buf) __delayed_free_header(__delayed_free, __user_ptr);
+                __delayed_free = __h;
             }
             else
             {
                 static __free_func_type __orig_free = __get_original_free();
 
-                // release objects from delayed release list
+                // releasing objects from delayed release list
                 while (__delayed_free)
                 {
                     __delayed_free_header* __next = __delayed_free->_M_next;
                     // it's possible that an object to be released during 1st call of __internal_free
-                    // would be released 2nd time from inside dlsym call. To prevent "double free"
+                    // would be released 2nd time from inside nested dlsym call. To prevent "double free"
                     // situation, check for it explicitly.
-                    if (__user_ptr != __delayed_free)
+                    if (__user_ptr != __delayed_free->_M_to_free)
                     {
-                        __orig_free(__delayed_free);
+                        __orig_free(__delayed_free->_M_to_free);
                     }
+                    __orig_free(__delayed_free);
                     __delayed_free = __next;
                 }
                 __orig_free(__user_ptr);
--- a/src/pstl_offload.cpp
+++ b/src/pstl_offload.cpp
@@ -46,7 +46,7 @@ static thread_local bool __dlsym_called = false;
 static thread_local __delayed_free_header* __delayed_free = nullptr;
 
 static __free_func_type
-__get_original_free()
+__get_original_free(void* __ptr_to_free)
 {
     __dlsym_called = true;
     __free_func_type __orig_free = __free_func_type(dlsym(RTLD_NEXT, "free"));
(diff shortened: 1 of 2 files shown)


src/pstl_offload.cpp
```diff
@@ -61,7 +98,25 @@ __internal_free(void* __user_ptr)
...
+            {
+                // Delay releasing till exit of dlsym. We do not overload malloc globally,
+                // so can use it safely. Do not use new to able to use free() during
+                // __delayed_free_header releasing.
+                void* __buf = malloc(sizeof(__delayed_free_header));
+                if (!__buf)
+                {
+                    throw std::bad_alloc();
+                }
+                __delayed_free_header* __h = new(__buf) __delayed_free_header(__delayed_free, __user_ptr);
```
rarutyun, line 111:
```suggestion
                __delayed_free_list* __h = new(__buf) __delayed_free_list(__delayed_free, __user_ptr);
```

`delayed_free_header` doesn't reflect what this thing is doing at all.

Alexandr-Konovalov: Done.

src/pstl_offload.cpp
```diff
@@ -29,12 +29,49 @@
...
-__get_original_free()
+using __free_func_type = void (*)(void*);
+
+// list of objects for delayed releasing
+struct __delayed_free_list {
+    __delayed_free_list* _M_next;
+    void*                  _M_to_free;
+
+    __delayed_free_list(__delayed_free_list* __next, void* __to_free)
+        : _M_next(__next), _M_to_free(__to_free) {}
```
kboyarinov, line 40: Since `__delayed_free_list` is a trivial structure, may be it makes sense to remove the constructor (and corresponding placement new calls) and either assign fields or use aggregate initialization (similar to how `__block_header` is operated)

src/pstl_offload.cpp
```diff
@@ -61,7 +98,25 @@ __internal_free(void* __user_ptr)
...
+                    throw std::bad_alloc();
+                }
+                __delayed_free_list* __h = new(__buf) __delayed_free_list(__delayed_free, __user_ptr);
+                __delayed_free = __h;
+            }
+            else
+            {
+                static __free_func_type __orig_free = __get_original_free(__user_ptr);
+
+                __orig_free(__user_ptr);
```
kboyarinov, line 118: May be it would be better to rename `__get_original_free` to something like `__free_original(__user_ptr)` and move `__user_ptr` releasing inside of the function. I guess the logic would be more clear. What do you think?

src/pstl_offload.cpp
Alexandr-Konovalov: Done.

src/pstl_offload.cpp
Alexandr-Konovalov: Symmetry with rest of `__get_original*` functions is a good thing, so I tend to agree.

src/pstl_offload.cpp
```diff
@@ -29,12 +29,54 @@
...
 {
 
-static auto
-__get_original_free()
+using __free_func_type = void (*)(void*);
+
+// list of objects for delayed releasing
+struct __delayed_free_list {
+    __delayed_free_list* _M_next;
+    void*                  _M_to_free;
```
kboyarinov, line 37: It seems like the indentation should be fixed. May be it is a GitHub fault since the code in "suggestion" has different number of whitespaces.

src/pstl_offload.cpp
```diff
@@ -61,7 +103,23 @@ __internal_free(void* __user_ptr)
...
+            {
+                // Delay releasing till exit of dlsym. We do not overload malloc globally,
+                // so can use it safely. Do not use new to able to use free() during
+                // __delayed_free_list releasing.
+                void* __buf = malloc(sizeof(__delayed_free_list));
+                if (!__buf)
+                {
+                    throw std::bad_alloc();
+                }
+                __delayed_free_list* __h = new(__buf) __delayed_free_list{__delayed_free, __user_ptr};
```
kboyarinov, line 116: I guess we can avoid placement new here:

```suggestion
                __delayed_free_list* __h = static_cast<__delayed_free_list*>(malloc(sizeof(__delayed_free_list)));
                if (!__h)
                {
                    throw std::bad_alloc();
                }
                *__h = __delayed_free_list{__delayed_free, __user_ptr};
```

src/pstl_offload.cpp
```diff
@@ -29,12 +29,54 @@
...
+    }
+
+    return __orig_free;
+}
+
+static __free_func_type
+__get_original_free(void* __ptr_to_free)
 {
-    using __free_func_type = void (*)(void*);
+    static __free_func_type __orig_free = __get_original_free_checked(__ptr_to_free);
```
kboyarinov, line 78: May be it also makes sense to immediately release the `__ptr_to_free` and rename the function:

```suggestion
    static __free_func_type __orig_free = __get_original_free_checked(__ptr_to_free);
    __orig_free(__ptr_to_free);
```
It will allow to avoid passing this pointer twice on the calling side.

Alexandr-Konovalov: You are suggesting to `__user_ptr` in the function that keep `__orig_free`, am I correctly understand that? Well, thus we lack symmetry with rest of `__get_original*` functions.

src/pstl_offload.cpp
Alexandr-Konovalov: Done.

src/pstl_offload.cpp
rarutyun: I think I tend to agree with what Konstantin is proposing. Perhaps, we should extrapolate this approach to other places. The code with the changes Konstantin made, the code looks cleaner to me/

rarutyun (approved):
Unfortunately, we still have infrastructure problems in CI but I don't see how why it should block this PR from being merged.

PR #1315, opened 2023-12-14
[pstl-offload] Support alignment on page size and more.

Alexandr-Konovalov:
Pointers to objects with large alignment are placed to global table. During memory release this table must be accessed only when a user pointer and pointer to a potential header of the object is not on same page.


src/pstl_offload.cpp
```diff
@@ -48,43 +48,111 @@ __get_original_msize()
...
+
+struct __large_aligned_ptr_header
+{
+    sycl::device* _M_device;
+    std::size_t   _M_size;
+};
+
+std::mutex __large_aligned_ptrs_mtx;
+// map user pointer to header
+std::unordered_map<void*, __large_aligned_ptr_header, __hash_aligned_ptr> __large_aligned_ptrs;
```
kboyarinov, line 71: When I prototyped the same idea several months ago, I faced an issue (as far as I remember on LULESH) when `free` was called after destroying the pointers table and I have resolved it by introducing the special "is alive" flag. Did you face something similar?

src/pstl_offload.cpp
```diff
@@ -48,43 +48,111 @@ __get_original_msize()
...
 
+struct __hash_aligned_ptr
+{
+    uintptr_t operator()(void *p) const
+    {
+        // we know that addresses are at least page-aligned, so drop 11 right bits that are zeros,
+        // and multiply to some prime number for bits hashing
+        // TODO: should well-known hashing function be used?
+        constexpr unsigned shift = 11;
+        return ((uintptr_t)p >> shift) * 1111111111111111111LLU;
```
kboyarinov, line 59: I guess we can use std::hash<void*> with the shift.

src/pstl_offload.cpp
Alexandr-Konovalov: Good point. It's unsafe to touch an objects with non-trivial dtor unconditionally.

src/pstl_offload.cpp
Alexandr-Konovalov: Done.

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+struct __hash_aligned_ptr
+{
+    uintptr_t operator()(void* __p) const
+    {
+        // We know that addresses are at least page-aligned, so, expecting page at least
+        // 4K-aligned, drop 11 right bits that are zeros, and treat rest as a pointer,
+        // hoping that an underlying Standard Library support this well.
+        static constexpr unsigned __ptr_shift = 11;
+        // current page size is at least 4K
+        assert(__get_memory_page_size() >= 1 << __ptr_shift);
```
rarutyun, line 125:
```suggestion
        assert(__get_memory_page_size() >= (1 << __ptr_shift));
```

I understand that this is equivalent to what is in the PR but I would add parentheses for the sake of readability (without remembering operators priority)

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+    using __realloc_func_type = void* (*)(void*, std::size_t);
+
+    static __realloc_func_type __orig_realloc = __realloc_func_type(dlsym(RTLD_NEXT, "realloc"));
+    return __orig_realloc;
+}
+
+inline bool
+__is_ptr_page_aligned(void* __p)
+{
+    return (uintptr_t)__p % __get_memory_page_size() == 0;
```
rarutyun, line 113:
```suggestion
    return (uintptr_t)__p & (__get_memory_page_size() - 1) == 0;
```
Effectively it should be the same, maybe even from code gen perspective if compiler optimizes it well. You can leave the code as is if you like.

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+}
+
+struct __hash_aligned_ptr
+{
+    uintptr_t operator()(void* __p) const
```
rarutyun, line 118:
```suggestion
    std::uintptr_t operator()(void* __p) const
```
For all the places where we have `uintptr_t`

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+        __orig_free_allocator<std::pair<void* const, __ptr_desc>>>;
+    __map_ptr_to_object_prop* _M_map;
+
+public:
+    // We suppose that all users of libpstloffload have dependence on it, so it's impossible to
+    // register >=4K-aligned USM memory before ctor of static objects in libpstloffload is executed.
+    // So, no need for special support for adding to not-yet-created __large_aligned_ptrs_map.
+    __large_aligned_ptrs_map()
+    {
+        std::scoped_lock __l(__large_aligned_ptrs_map_mtx);
```
rarutyun, line 186: Do we suppose to have multiple of such objects? If not, do we synchronize with something else?

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+        _M_map = new __map_ptr_to_object_prop;
```
rarutyun, line 187: Which "allocator" do we use? Is it original one? Or is it USM-substituted one?

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+    {
+        assert(__is_ptr_page_aligned(__ptr));
+
+        std::scoped_lock __l(__large_aligned_ptrs_map_mtx);
+        [[maybe_unused]] auto __ret = _M_map->emplace(__ptr, __ptr_desc{std::move(__device_ptr), __size});
+        assert(__ret.second); // the pointer must be unique
+    }
+
+    // nullopt means "it's not our pointer"
+    static std::optional<__ptr_desc>
```
rarutyun, line 206: Do we really need `optional`? C/C++ pointer is sort of optional itself. Is there any particular advantage to use optional?

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+            return std::nullopt;
+        }
+        return __iter->second._M_requested_number_of_bytes;
+    }
+};
+
+static __large_aligned_ptrs_map __large_aligned_ptrs;
+
+void
+__block_header::__free()
```
rarutyun, line 259: Why did you decide to make it a member function?

rarutyun (commented):
First portion of the comments

src/pstl_offload.cpp
Alexandr-Konovalov: We are using `__orig_free_allocator` (see `__map_ptr_to_object_prop` alias creation). `__orig_free_allocator` uses `std::malloc` and `__original_free`. libpstloffload.so is build without malloc replacement.

src/pstl_offload.cpp
Alexandr-Konovalov: I'm afraid that no compilers are smart enough to deduce that `__get_memory_page_size()` returns only power of 2 values. So, let's change.

src/pstl_offload.cpp
Alexandr-Konovalov: Thanks, one is unneeded.

src/pstl_offload.cpp
Alexandr-Konovalov: Let's make it just function, if you prefer.

src/pstl_offload.cpp
Alexandr-Konovalov: Done.

src/pstl_offload.cpp
Alexandr-Konovalov: Done.

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+        if (__block_header* __header = static_cast<__block_header*>(__user_ptr) - 1;
+            __same_memory_page(__user_ptr, __header) && __header->_M_uniq_const == __uniq_type_const)
+        {
+            __header->__free();
+            return;
+        }
 
-        if (__same_memory_page(__user_ptr, __header) && __header->_M_uniq_const == __uniq_type_const)
+        if (std::optional<__large_aligned_ptrs_map::__ptr_desc>
+                __desc = __large_aligned_ptrs_map::__unregister_ptr(__large_aligned_ptrs, __user_ptr))
```
kboyarinov, line 281: As far as I understand, the `__unregister_ptr` have try semantics and may fail. May be it makes more sense to rename it to `__try_unregister_ptr` to better explain the semantics?
The same applies to `__get_size`.

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+    {
+        assert(__is_ptr_page_aligned(__ptr));
+
+        std::scoped_lock __l(__large_aligned_ptrs_map_mtx);
+        [[maybe_unused]] auto __ret = _M_map->emplace(__ptr, __ptr_desc{std::move(__device_ptr), __size});
+        assert(__ret.second); // the pointer must be unique
+    }
+
+    // nullopt means "it's not our pointer"
+    static std::optional<__ptr_desc>
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Good point. No reason to not unify.

src/pstl_offload.cpp
Alexandr-Konovalov: Interesting. Is "try semantics" kind of standard terminology? I agree, the names are not clean, can it be `__unregister_known_ptr` etc?

src/pstl_offload.cpp
```diff
@@ -96,59 +97,311 @@ __original_msize(void* __user_ptr)
...
+        return __result;
+    }
+
+    if (std::optional<__large_aligned_ptrs_map::__ptr_desc>
+            __desc = __large_aligned_ptrs_map::__unregister_ptr(__large_aligned_ptrs, __user_ptr))
+    {
+        if (__desc->_M_requested_number_of_bytes == __new_size)
+        {
+            return __user_ptr;
+        }
```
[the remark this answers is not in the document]
Alexandr-Konovalov: > First of all, I think the check should be >= instead of ==.

Why? I disliked interesting cases like "the object was aligned, but now it's just 8 bytes, but we keep it in the table".

> The second issue is if that branch would be reached, __user_ptr would be removed from the pointers map, but not released and the next call to free would result in "original" free call.

Thanks, fixed.

src/pstl_offload.cpp
```diff
@@ -113,43 +126,223 @@ __original_msize(void* __user_ptr)
...
+        std::unordered_map<void*, __ptr_desc, __hash_aligned_ptr, std::equal_to<void*>,
+        __orig_free_allocator<std::pair<void* const, __ptr_desc>>>;
+    __map_ptr_to_object_prop* _M_map;
+
+public:
+    // We suppose that all users of libpstloffload have dependence on it, so it's impossible to
+    // register >=4K-aligned USM memory before ctor of static objects in libpstloffload is executed.
+    // So, no need for special support for adding to not-yet-created __large_aligned_ptrs_map.
+    __large_aligned_ptrs_map() : _M_map(new __map_ptr_to_object_prop) { }
+
```
Alexandr-Konovalov, line 223: Delete ctors etc.

src/pstl_offload.cpp
```diff
@@ -113,43 +126,223 @@ __original_msize(void* __user_ptr)
...
+    {
+#if __linux__
+        __original_free(__ptr);
+#elif _WIN64
+        __original_free_ptr(__ptr);
+#endif
+    }
+};
+
+template<class T, class U>
```
Alexandr-Konovalov, line 193: Make them friends.

include/pstl_offload/internal/usm_memory_replacement_common.h
```diff
@@ -205,125 +242,20 @@ __allocate_shared_for_device(__sycl_device_shared_ptr __device_ptr, std::size_t
...
-{
-    assert(__header != nullptr);
-    __header->_M_uniq_const = 0;
-    sycl::context __context = __header->_M_device.__get_context();
-    __header->~__block_header();
-    sycl::free(__header->_M_original_pointer, __context);
-}
-
-inline void*
-__realloc_real_pointer(void* __user_ptr, std::size_t __new_size)
```
rarutyun, line 219: Do we move `__realloc_real_pointer` and to the `__aligned_realloc_real_pointer` to cpp file, because they need access to the unordered_map (in case of page+ alignment)?

src/pstl_offload.cpp
```diff
@@ -10,6 +10,7 @@
 #include <new>
 #include <cassert>
 #include <cstdint>
+#include <mutex> // std::scoped_lock
```
rarutyun, line 13: For some reasons the file permissions are changes (100644 → 100755)

src/pstl_offload.cpp
rarutyun: ok, a different and more important question. Why did we move it to the cpp part? is it because it is not used in headers anymore?

src/pstl_offload.cpp
```diff
@@ -113,43 +126,223 @@ __original_msize(void* __user_ptr)
...
+        if (__block_header* __header = static_cast<__block_header*>(__user_ptr) - 1;
+            __same_memory_page(__user_ptr, __header) && __header->_M_uniq_const == __uniq_type_const)
+        {
+            __header->__free();
+            return;
+        }
 
-        if (__same_memory_page(__user_ptr, __header) && __header->_M_uniq_const == __uniq_type_const)
+        if (std::optional<__large_aligned_ptrs_map::__ptr_desc>
+                __desc = __large_aligned_ptrs_map::__unregister_ptr(__large_aligned_ptrs, __user_ptr))
```
rarutyun, line 317: I would suggest to rewrite `std::optional` checks here and in other places to initialization statement as you have above to decouple initialization statement and bool expression. Whether to use `operator bool` or `has_value` is up to you. I would use `has_value` for the sake of readability but this is a very slight preference.

For example for this code snippet I would do something like:
```suggestion
        if (std::optional<__large_aligned_ptrs_map::__ptr_desc>
                __desc = __large_aligned_ptrs_map::__unregister_ptr(__large_aligned_ptrs, __user_ptr); __desc.has_value())
```

src/pstl_offload.cpp
```diff
@@ -113,43 +126,223 @@ __original_msize(void* __user_ptr)
...
+    };
+
+private:
+    // Find sycl::device and requested size by user pointer. Use __orig_free_allocator to not
+    // call global delete during delete processing (overloaded global delete includes
+    // __unregister_ptr call).
+    using __map_ptr_to_object_prop =
+        std::unordered_map<void*, __ptr_desc, __hash_aligned_ptr, std::equal_to<void*>,
+        __orig_free_allocator<std::pair<void* const, __ptr_desc>>>;
+    __map_ptr_to_object_prop* _M_map;
```
rarutyun, line 216: Could you please try the approach with 
```cpp
union {
    __map_ptr_to_object_prop _M_map;
}
```
and then *placement*-`new` here and let me know if it works. Then we can decide what we prefer

src/pstl_offload.cpp
rarutyun: Discussed offline. Let's stick to optional right now

src/pstl_offload.cpp
```diff
@@ -200,22 +393,162 @@ __internal_msize(void* __user_ptr)
...
 
     if (__same_memory_page(__user_ptr, __header) && __header->_M_uniq_const == __uniq_type_const)
     {
-        __res = __header->_M_requested_number_of_bytes;
+        return __header->_M_requested_number_of_bytes;
+    }
+
+    std::optional<std::size_t> __size = __large_aligned_ptrs_map::__get_size(__large_aligned_ptrs, __user_ptr);
+
+    return __size.has_value() ? __size.value() : __original_msize(__user_ptr);
```
rarutyun, line 405:
```suggestion
    return __size.has_value() ? *__size : __original_msize(__user_ptr);
```
Since we already checked if `optional` has a value it's better to call the API that doesn't throw.

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -382,6 +355,24 @@ inline void* __attribute__((always_inline)) __libc_realloc(void* __ptr, std::siz
...
 
+inline void* __attribute__((always_inline)) valloc(std::size_t __size)
+{
+    return memalign(__pstl_offload::__get_memory_page_size(), __size);
+}
+
+inline void* __attribute__((always_inline)) __libc_valloc(std::size_t __size) { return valloc(__size); }
+
+// __THROW to match system declaration of pvalloc
+inline void* __attribute__((always_inline)) pvalloc(std::size_t __size) __THROW
```
rarutyun, line 366: generic question. Do we need this function at all?

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -382,6 +355,24 @@ inline void* __attribute__((always_inline)) __libc_realloc(void* __ptr, std::siz
...
+{
+    size_t __page_size = __pstl_offload::__get_memory_page_size();
```
rarutyun, line 368:
```suggestion
    std::size_t __page_size = __pstl_offload::__get_memory_page_size();
```

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -382,6 +355,24 @@ inline void* __attribute__((always_inline)) __libc_realloc(void* __ptr, std::siz
...
+    // align size up to the page size
+    __size = __size ? ((__size - 1) | (__page_size - 1)) + 1 : __page_size;
```
rarutyun, line 370: I don't understand `size ?` part. Can our `__get_memory_page_size` return `0`?

rarutyun (commented):
Overall comment. Have you tried you changes with clover-leaf benchmark from UoB github repository? If I am not mistaken they have "more than page" alignment. Alternatively we could try something else but I would like us to check this patch on the real benchmark

Alexandr-Konovalov:
> Overall comment. Have you tried you changes with clover-leaf benchmark from UoB github repository? If I am not mistaken they have "more than page" alignment. Alternatively we could try something else but I would like us to check this patch on the real benchmark

The BabelStream benchmark require this, and the functionality is blocked by [Memory allocation with alignment > 64KB is disabled #1537](https://github.com/oneapi-src/unified-runtime/issues/1537).

include/pstl_offload/internal/usm_memory_replacement_common.h
Alexandr-Konovalov: Yes.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: > I don't understand size ? part. 
We mimic behavior of replaced `pvalloc()` . 

> Can our __get_memory_page_size return 0?

I'm afraid I'm not understand the question. Are you asking what we will do if somehow a system call returns completely invalid value? 

Just in case I will add checking of `sysconf()` return value.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Done.

src/pstl_offload.cpp
Alexandr-Konovalov: Sorry, restored.

src/pstl_offload.cpp
Alexandr-Konovalov: Done.

src/pstl_offload.cpp
Alexandr-Konovalov: Done.

src/pstl_offload.cpp
Alexandr-Konovalov: Done.

include/pstl_offload/internal/usm_memory_replacement.h
rarutyun: yep, my bad, I misread the variable names

src/pstl_offload.cpp
Alexandr-Konovalov: One (pretty marginal) point here is that we need a to use `__large_aligned_ptrs_map` methods before its ctor run. With pointer-based implementation, this flag is just pointer. For union-based, some additional flag is required.

src/pstl_offload.cpp
```diff
@@ -113,43 +120,229 @@ __original_msize(void* __user_ptr)
...
+}
+
+struct __hash_aligned_ptr
+{
+    std::uintptr_t operator()(void* __p) const
+    {
+        // We know that addresses are at least page-aligned, so, expecting page at least
+        // 4K-aligned, drop 11 right bits that are zeros, and treat rest as a pointer,
+        // hoping that an underlying Standard Library support this well.
+        static constexpr unsigned __ptr_shift = 11;
```
rarutyun, line 151:
```suggestion
        constexpr unsigned __ptr_shift = 11;
```

src/pstl_offload.cpp
```diff
@@ -113,43 +120,229 @@ __original_msize(void* __user_ptr)
...
+
+template<class T>
+struct __orig_free_allocator
+{
+    typedef T value_type;
+
+    __orig_free_allocator() = default;
+
+    template<class U>
+    constexpr __orig_free_allocator(const __orig_free_allocator <U>&) noexcept {}
```
rarutyun, line 166:
```suggestion
    constexpr __orig_free_allocator(const __orig_free_allocator<U>&) noexcept {}
```

src/pstl_offload.cpp
```diff
@@ -113,43 +120,229 @@ __original_msize(void* __user_ptr)
...
+        __original_free(__ptr);
+#elif _WIN64
+        __original_free_ptr(__ptr);
+#endif
+    }
+
+    template<class U>
+        friend bool operator==(const __orig_free_allocator <T>&, const __orig_free_allocator <U>&) { return true; }
+    template<class U>
+        friend bool operator!=(const __orig_free_allocator <T>&, const __orig_free_allocator <U>&) { return false; }
```
rarutyun, line 189:
```suggestion
    template <class U>
    friend bool operator==(const __orig_free_allocator <T>&, const __orig_free_allocator <U>&) { return true; }
    template <class U>
    friend bool operator!=(const __orig_free_allocator <T>&, const __orig_free_allocator <U>&) { return false; }
```

src/pstl_offload.cpp
```diff
@@ -113,43 +120,229 @@ __original_msize(void* __user_ptr)
...
+};
+
+template<class T>
+struct __orig_free_allocator
+{
+    typedef T value_type;
+
+    __orig_free_allocator() = default;
+
+    template<class U>
```
rarutyun, line 165:
```suggestion
    template <class U>
```

test/pstl_offload/general/free_after_unload_lib.cpp
```diff
@@ -11,23 +11,33 @@
...
+    ptrs = *p;
 }
 
 struct DelayedReleaser
 {
     ~DelayedReleaser()
     {
-        free(ptr);
+        free(ptrs.p1);
+        ::operator delete (ptrs.p2, std::align_val_t(8*1024));
```
rarutyun, line 34:
```suggestion
        ::operator delete (ptrs.p2, std::align_val_t(8 * 1024));
```

test/pstl_offload/general/free_after_unload_lib.h
```diff
@@ -10,9 +10,16 @@
...
+    void* p1;
+    int* p2;
+    void* p3;
+};
+
 #if free_after_unload_lib_EXPORTS && _WIN64
 __declspec(dllexport)
 #endif
-void register_mem_to_later_release(void *);
+void register_mem_to_later_release(pointers *);
```
rarutyun, line 23:
```suggestion
void register_mem_to_later_release(pointers*);
```

src/pstl_offload.cpp
```diff
@@ -113,43 +120,229 @@ __original_msize(void* __user_ptr)
...
+    using __map_ptr_to_object_prop =
+        std::unordered_map<void*, __ptr_desc, __hash_aligned_ptr, std::equal_to<void*>,
+        __orig_free_allocator<std::pair<void* const, __ptr_desc>>>;
+    __map_ptr_to_object_prop* _M_map;
+
+public:
+    // We suppose that all users of libpstloffload have dependence on it, so it's impossible to
+    // register >=4K-aligned USM memory before ctor of static objects in libpstloffload is executed.
+    // So, no need for special support for adding to not-yet-created __large_aligned_ptrs_map.
+    __large_aligned_ptrs_map() : _M_map(new __map_ptr_to_object_prop) { }
```
rarutyun, line 218: Please return lock of the mutex back in the constructor. Seems like we have a race with `unregister`. Please also add a comment about having this map in BSS, so we rely on the fact that `_M_map` is initialized with zeros and it's safe to do reading in `unregister` even before the object constructor was called.

test/pstl_offload/general/free_after_unload_lib.cpp
```diff
@@ -11,23 +11,33 @@
...
+
+        constexpr size_t updated_size = 1024;
```
rarutyun, line 36:
```suggestion
        constexpr std::size_t updated_size = 1024;
```

test/pstl_offload/memory/usm_memory_alignment.pass.cpp
```diff
@@ -54,12 +55,23 @@ void test_alignment_allocation(AllocatingFunction allocate, DeallocatingFunction
 // test_alignment_allocation tests different sizes values because other functions
 // only requires the alignment to be power of two
 void test_aligned_alloc_alignment() {
-    for (std::size_t alignment = 1; alignment < get_page_size(); alignment <<= 1) {
+    int cnt = 0;
+    for (std::size_t alignment = 1; alignment < 16*get_page_size(); alignment <<= 1) {
```
rarutyun, line 59:
```suggestion
    for (std::size_t alignment = 1; alignment < 16 * get_page_size(); alignment <<= 1) {
```

src/pstl_offload.cpp
```diff
@@ -113,43 +120,229 @@ __original_msize(void* __user_ptr)
...
+        // current page size is at least 4K
+        assert(__get_memory_page_size() >= (1 << __ptr_shift));
+        return std::hash<void*>()((void*)((std::uintptr_t)__p >> __ptr_shift));
+    }
+};
+
+template<class T>
+struct __orig_free_allocator
+{
+    typedef T value_type;
```
rarutyun, line 161:
```suggestion
    using value_type = T;
```

rarutyun:
We discussed this PR with @Alexandr-Konovalov pretty exhaustively. I am happy with the current state of the patch. I expect my comments to be addressed. But other than that looks good to me. @kboyarinov, feel free to approve this patch but please double check that my comments are addressed. Also, if you find something that we missed feel free to add more comments

rarutyun:
@kboyarinov. Important note. If you are going to approve this PR please make sure that merge is done correctly and we don't have anything from "Windows support" patch in this PR.

test/pstl_offload/general/free_after_unload.pass.cpp
```diff
@@ -41,10 +41,16 @@ int main()
 {
     sycl::context memory_context = TestUtils::get_pstl_offload_device().get_platform().ext_oneapi_get_default_context();
 
-    void* ptr = malloc(8);
-    EXPECT_TRUE(ptr, "Can't get memory while allocating with overloaded malloc");
-    EXPECT_TRUE(sycl::get_pointer_type(ptr, memory_context) == sycl::usm::alloc::shared, "Wrong pointer type while allocating with overloaded malloc");
-    register_mem_to_later_release(ptr);
+    pointers ptrs{malloc(8), new (std::align_val_t(8*1024)) int, std::aligned_alloc(8*1024, 10)};
```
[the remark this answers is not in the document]
Alexandr-Konovalov: Done.

include/pstl_offload/internal/usm_memory_replacement.h
```diff
@@ -18,7 +18,7 @@
 #include <atomic>
 #include <cassert>
 #include <cerrno>
-#include <optional>
+#include <mutex> // std::scoped_lock
```
kboyarinov, line 21: It seems like it is not needed anymore in this file

src/pstl_offload.cpp
```diff
@@ -113,43 +120,236 @@ __original_msize(void* __user_ptr)
...
+    template <class U>
+    constexpr __orig_free_allocator(const __orig_free_allocator<U>&) noexcept {}
+
+    T* allocate(std::size_t __n)
+    {
+        if (T *ptr = static_cast<T*>(std::malloc(__n * sizeof(T))))
+        {
+            return ptr;
+        }
+        throw std::bad_alloc();
```
kboyarinov, line 174: May be it would be more readable if we revert the logic?
if (!ptr) {
    throw bad_alloc{};
}
return ptr;

As far as I know the original code can result in warnings on some system because the else branch don't have a return statement.

include/pstl_offload/internal/usm_memory_replacement.h
Alexandr-Konovalov: Why? `std::scoped_lock` is used [below](https://github.com/oneapi-src/oneDPL/blob/902256bdb75f9005b4054f1d3f7d1d2ecd26c8db/include/pstl_offload/internal/usm_memory_replacement.h#L106).

src/pstl_offload.cpp
Alexandr-Konovalov: Done.

Very strange systems, I would say. In which case they expect `throw` not throwing?

kboyarinov (approved):
LGTM

test/pstl_offload/memory/usm_memory_alignment.pass.cpp
Alexandr-Konovalov: Done.

PR #1318, opened 2023-12-15
[oneDPL][test][perm_it] Some test cases were disabled, because it is not supported by oneDPL

MikeDvorskiy:
1. The test coverages for a case with permutation_iterator(host_iterator, valid_iterator) was disabled, because it is not supported by oneDPL. Before it was UB, now - compile time error.
2. Compile time diagnostic improved for the mentioned case.

### DIFF
--- a/test/parallel_api/numeric/numeric.ops/exclusive_scan_by_segment_perm_it.pass.cpp
+++ b/test/parallel_api/numeric/numeric.ops/exclusive_scan_by_segment_perm_it.pass.cpp
@@ -139,6 +139,8 @@ test_exclusive_scan(sycl::queue q, std::vector<size_t>& perms, std::vector<TestV
     EXPECT_EQ_RANGES(expectedResults, results, "wrong effect from exclusive_scan_by_segment #2");
 }
 
+#define _ONEDPL_PERM_BASE_ITERATOR_HOST_DEVICE_POL_SUPPORT 0
+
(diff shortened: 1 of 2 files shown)


test/parallel_api/numeric/numeric.ops/exclusive_scan_by_segment_perm_it.pass.cpp
```diff
@@ -139,6 +139,8 @@ test_exclusive_scan(sycl::queue q, std::vector<size_t>& perms, std::vector<TestV
     EXPECT_EQ_RANGES(expectedResults, results, "wrong effect from exclusive_scan_by_segment #2");
 }
 
+#define __ONEDPL_PERM_BASE_ITERATOR_HOST_DEVICE_POL_SUPPORT 0
```
dmitriy-sobolev, line 142: I am fine with that macro being in the test right now. Howerver, could you rename it? `__ONEDPL_PERM_BASE_ITERATOR_HOST_DEVICE_POL_SUPPORT` -> `_ONEDPL_PERM_BASE_ITERATOR_HOST_DEVICE_POL_SUPPORT` so there is only one leading underscore as in other internal macros. 

But in the long term, it might be useful to keep it inside the primary headers as a feature macro. e.g. as #945 suggests.

dmitriy-sobolev (approved):
LGTM

PR #1319, opened 2023-12-17
[CI] Fix CI by moving from Conda to other approaches

dmitriy-sobolev:
libpstloffload and FPGA emulator are not available through Conda packages. This PR fixes it by using other installation methods:

- APT on Linux
- Offline installers on Windows

Side effects:
 - Simpler environment setup (e.g. no registry manipulations, no manual env variables set up).
 - Component links should be manually updated upon each component release (Windows only)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -335,7 +335,7 @@ jobs:
             call "%WINDOWS_ONEAPI_PATH%\setvars.bat"
           )
           if "${{ matrix.cxx_compiler }}" == "cl" (
-            call "C:\Program Files (x86)\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64
+            call "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Auxiliary\Build\vcvarsall.bat" x64
           )
           powershell $output = cmake --version; Write-Host ::warning::CMake: $output
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,12 @@ env:
   BUILD_CONCURRENCY: 2
   MACOS_BUILD_CONCURRENCY: 3
   TEST_TIMEOUT: 360
+  WINDOWS_TBB_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/64957c0f-37bf-4408-909c-37ff52fe5119/w_tbb_oneapi_p_2021.11.0.49526.exe
+  WINDOWS_ICPX_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/dc441d10-698c-417d-be6f-44a172f7ab77/w_dpcpp-cpp-compiler_p_2024.0.1.28_offline.exe
+  WINDOWS_ONEAPI_PATH: C:\Program Files (x86)\Intel\oneAPI
+  LINUX_TBB_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/af3ad519-4c87-4534-87cb-5c7bda12754e/l_tbb_oneapi_p_2021.11.0.49527_offline.sh
+  LINUX_ICPX_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/c68c8f0a-47f5-4f26-8e8e-fa2627271279/l_dpcpp-cpp-compiler_p_2024.0.1.29_offline.sh
+  LINUX_ONEAPI_PATH: /opt/intel/oneapi
   MACOS_ONEAPI_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/irc_nas/18358/m_cpp-compiler-classic_p_2022.0.0.62_offline.dmg
 
 concurrency:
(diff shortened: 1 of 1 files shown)


danhoeflinger:
Have you considered using `apt` instead of direct links?  apt install should provide similar to conda but without the unique environment setup stuff that comes along with conda.
It seems like this should be able to provide the pros without the con of updating the links.  However, I have not investigated this in detail.

dmitriy-sobolev:
> Have you considered using `apt` instead of direct links? apt install should provide similar to conda but without the unique environment setup stuff that comes along with conda. It seems like this should be able to provide the pros without the con of updating the links. However, I have not investigated this in detail.

I have not tried using APT. However, I checked [the installation guide](https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-linux/2024-0/apt.html#GUID-B1770241-0BBB-4E9C-B85B-9268CA66F15C). It shows that there are toolkit and runtime packages. I think downloading and installing the whole base toolkit should be avoided: public GitHub CI runners have up to 14 GB space, while the toolkit takes up to 24 GB. The runtime packages are not shipped with FPGA emulator.

kboyarinov:
> > Have you considered using `apt` instead of direct links? apt install should provide similar to conda but without the unique environment setup stuff that comes along with conda. It seems like this should be able to provide the pros without the con of updating the links. However, I have not investigated this in detail.
> 
> I have not tried using APT. However, I checked [the installation guide](https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-linux/2024-0/apt.html#GUID-B1770241-0BBB-4E9C-B85B-9268CA66F15C). It shows that there are toolkit and runtime packages. I think downloading and installing the whole base toolkit should be avoided: public GitHub CI runners have up to 14 GB space, while the toolkit takes up to 24 GB. The runtime packages are not shipped with FPGA emulator (basing on internal sources).
> 
> Upd. 1/2/2024 I tried APT packages. They have FPGA emulator. The remaining CI issues:
> 
> * `HOST,bknd=tbb,cmplr=icpx,ubuntu-20.04,std=с++17,cfg=release`: well-known issue with reduce_by_segment test, it is expected to be fixed in [[Tests] Update broken macro for reduce_by_segment test #1324](https://github.com/oneapi-src/oneDPL/pull/1324)
> * `PSTL offload cpu,bknd=dpcpp,cmplr=icpx,ubuntu-22.04,std=с++17,cfg=release`: if I keep oneDPL headers, which are installed through APT, then the tests pass, otherwise they fail. It seems the issue is either in PSTL offload headers from GitHub or libpstloffload should be sourced differently. @kboyarinov, what do you think?

I have another PR #1283 that should fix this issue. Testing of this PR merged #1283 [runs successfully](https://github.com/oneapi-src/oneDPL/actions/runs/7421207098/job/20194080962).

.github/workflows/ci.yml
```diff
@@ -198,7 +214,6 @@ jobs:
             make_targets="build-onedpl-tests"
           fi
           mkdir build && cd build
-          lscpu
```
danhoeflinger, line 201: why are we removing `lscpu` for the `linux-testing`?

.github/workflows/ci.yml
```diff
@@ -15,6 +15,11 @@ env:
   BUILD_CONCURRENCY: 2
   MACOS_BUILD_CONCURRENCY: 3
   TEST_TIMEOUT: 360
+  WINDOWS_TBB_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/64957c0f-37bf-4408-909c-37ff52fe5119/w_tbb_oneapi_p_2021.11.0.49526.exe
+  WINDOWS_ICPX_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/94e15cb5-4bcc-4fdd-91cf-0f819a54e42e/w_dpcpp-cpp-compiler_p_2024.0.2.28_offline.exe
```
danhoeflinger, line 19: We can deal with this in another PR, but we should be able to install on windows via a package manager other than conda as well.  I think pip should offer a similar experience to apt, and allow us to get the "newest" version without updating this link each release. 

https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-windows/2024-0/install-using-package-managers.html#GUID-495DDF2D-2369-4DF5-B248-2A53123282DA

danhoeflinger (approved):
LGTM, we can potentially improve the Windows side in another PR I think with a pip or nuget.

.github/workflows/ci.yml
dmitriy-sobolev: I have a couple of caveats regarding PIP and NuGet: 
 
 - This is a speculation, but I suppose that PIP is almost identical to Conda: we will have to do the same things to configure the environment (registry patching and setting up env variables).
 - The [NuGet page](https://www.intel.com/content/www/us/en/docs/oneapi/installation-guide-windows/2024-0/nuget.html) does not contain information about DPC++ compiler.

It can be investigated later with a low priority as updating links quarterly is not much of a burden. I will create an issue.

.github/workflows/ci.yml
dmitriy-sobolev: I've reverted that change, thanks for noticing.

.github/workflows/ci.yml
danhoeflinger: I don't know this for certain, but when I've previously used pip in the context of python development, it seemed to be closer to apt than conda.  I don't think it relies so much on controlling the environment itself, and hopefully wouldn't require so many tricks.  I'll try to make some time to explore this sometime in the near future.

dmitriy-sobolev: Thank you a lot for the assistance, Dan. I really appreciate it.

dmitriy-sobolev: The suggestion has been noted: https://github.com/oneapi-src/oneDPL/issues/1337

PR #1320, opened 2023-12-18
Single-pass scan kernel template

adamfidel:
This PR provides an implementation of the single-pass scan algorithm as a kernel template.


include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,245 @@
...
+        return __sum;
+    }
+
+    _AtomicRefT __atomic_flag;
+};
+
+template <bool _Inclusive, typename _InRange, typename _OutRange, typename _BinaryOp, typename _KernelParam>
+sycl::event
+single_pass_scan_impl(sycl::queue __queue, _InRange&& __in_rng, _OutRange&& __out_rng, _BinaryOp __binary_op,
+                      _KernelParam)
```
SergeyKopienko, line 109: How it's already implemented for another algorithm:
```C++
template <bool __is_ascending = true, ::std::uint8_t __radix_bits = 8, typename _KernelParam, typename _KeysIterator>
sycl::event
radix_sort(sycl::queue __q, _KeysIterator __keys_first, _KeysIterator __keys_last, _KernelParam __param = {})
```

So better to have `_KernelParam __param = {}`

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,245 @@
...
+                    if (__next_offset > __n)
+                        __next_offset = __n;
+                    auto __in_begin = __in_rng.begin() + __current_offset;
+                    auto __in_end = __in_rng.begin() + __next_offset;
+                    auto __out_begin = __out_rng.begin() + __current_offset;
+
+                    if (__current_offset >= __n)
+                        return;
+
+                    _Type __local_sum = sycl::joint_reduce(__group, __in_begin, __in_end, __binary_op);
```
julianmi, line 188: This limits `__binary_op` to the SYCL supported binary operators (https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:function-objects), right? The same goes for `sycl::joint_inclusive_scan()` below. Should we document this limitation?
If we want to support other operators in the future, we might want to reuse our existing reduce and scan implementations for those.

mmichel11 (commented):
Here are my initial comments on this PR. I will wait to review over the design of the decoupled lookback until the considerations about splitting the synchronization flag and partial scan result are made.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,147 @@
...
+#endif
+}
+
+
+int
+main()
+{
+    using T = int;
+    constexpr int data_per_work_item = 8;
+    constexpr int work_group_size = 256;
```
dmitriy-sobolev, line 132: Do I get it right that the KT currently works only with these values?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,333 @@
...
+
+    auto __keep1 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _InIterator>();
+    auto __buf1 = __keep1(__in_begin, __in_end);
+    auto __keep2 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _OutIterator>();
+    auto __buf2 = __keep2(__out_begin, __out_begin + __n);
+
+    return __impl::__single_pass_scan<true>(__queue, __buf1.all_view(), __buf2.all_view(), __binary_op, __param);
+}
+
+} // namespace igpu
```
dmitriy-sobolev, line 329: What does igpu stands on? I am correct in assuming that this KT is targeted to intel GPUs?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,333 @@
...
+    });
+
+    return __free_event;
+}
+
+} // namespace __impl
+
+template <typename _InRng, typename _OutRng, typename _BinaryOp, typename _KernelParam>
+sycl::event
+inclusive_scan(sycl::queue __queue, _InRng __in_rng, _OutRng __out_rng, _BinaryOp __binary_op, _KernelParam __param)
```
dmitriy-sobolev, line 308: Let's pass `__in_rng` and `__out_rng` by forward references to avoid copying. 

I am not totally sure on how to pass the resulting range, given that C++ range algorithms accept an iterator as result, e.g. https://en.cppreference.com/w/cpp/algorithm/ranges/copy and I do not the reason for it. It seems better to pass it as a range e.g. to support API like `inclusive_scan(queue, sycl_buffer_in, sycl_buffer_out, ...)` and to make sure if the output container has enough space.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,147 @@
...
+template <typename T, typename KernelParam>
+void
+test_general_cases(sycl::queue q, std::size_t size, KernelParam param)
+{
+    test_usm<T, sycl::usm::alloc::shared>(q, size, param);
+    test_usm<T, sycl::usm::alloc::device>(q, size, param);
+    test_sycl_iterators<T>(q, size, param);
+#if _ENABLE_RANGES_TESTING
+    test_all_view<T>(q, size, param);
+#endif
```
dmitriy-sobolev, line 123: What's about adding a test case for `sycl::buffer`? It is implicitly converted to a range, so I expect it is already supported. 

The test case will contain something like this:
```c++
sycl::buffer<T> buf(input.data(), input.size());
sycl::buffer<T> buf_out(input.size());
auto e = oneapi::dpl::experimental::kt::inclusive_scan(q, buf, buf_out, ::std::plus<T>(), param);
e.wait();
```

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: Yes, this limitation should be documented. In the future, we can pull in the scan/reduce support for arbitrary binary operators as you suggest.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: Thank you for pointing this out. I have changed it to default initialize the kernel param struct.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,245 @@
...
+    {
+        __atomic_flag.store(__val | __full_mask);
+    }
+
+    template <typename _Subgroup, typename _BinOp>
+    _T
+    cooperative_lookback(::std::uint32_t __tile_id, const _Subgroup& __subgroup, _BinOp __bin_op,
+                         _StorageType* __flags_begin)
+    {
+        _T __sum{0};
```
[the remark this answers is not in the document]
adamfidel: You are correct. This will be changed to use a known identity.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,245 @@
...
+            bool __is_full = (__tile_val & __full_mask) && ((__tile_val & __partial_mask) == 0);
+            auto __is_full_ballot = sycl::ext::oneapi::group_ballot(__subgroup, __is_full);
+            ::std::uint32_t __is_full_ballot_bits{};
+            __is_full_ballot.extract_bits(__is_full_ballot_bits);
+
+            auto __lowest_item_with_full = sycl::ctz(__is_full_ballot_bits);
+            _T __contribution = __local_id <= __lowest_item_with_full ? __tile_val & __value_mask : _T{0};
+
+            // Sum all of the partial results from the tiles found, as well as the full contribution from the closest tile (if any)
+            __sum += sycl::reduce_over_group(__subgroup, __contribution, __bin_op);
```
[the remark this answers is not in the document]
adamfidel: It should indeed use the binary operator here. I've changed instances of the word "sum" to "reduction" to be more generic.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,245 @@
...
+    assert("This device does not support 64-bit atomics" &&
+           (sizeof(_Type) < 64 || __queue.get_device().has(sycl::aspect::atomic64)));
+
+    const ::std::size_t __n = __in_rng.size();
+
+    // We need to process the input array by 2^30 chunks for 32-bit ints
+    constexpr ::std::size_t __chunk_size = 1ul << (sizeof(_Type) * 8 - 2);
+    const ::std::size_t __num_chunks = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __chunk_size);
+
+    auto __max_cu = __queue.get_device().template get_info<sycl::info::device::max_compute_units>();
```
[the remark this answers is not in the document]
adamfidel: Good catch. I will remove it.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,245 @@
...
+                    auto __group = __item.get_group();
+                    auto __subgroup = __item.get_sub_group();
+
+                    // Obtain unique ID for this work-group that will be used in decoupled lookback
+                    if (__group.leader())
+                    {
+                        sycl::atomic_ref<_FlagStorageType, sycl::memory_order::relaxed, sycl::memory_scope::device,
+                                         sycl::access::address_space::global_space>
+                            __idx_atomic(__status_flags[__status_flags_size - 1]);
+                        __tile_id_lacc[0] = __idx_atomic.fetch_add(1);
```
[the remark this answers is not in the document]
adamfidel: That's a good suggestion. I will change it to use a `group_broadcast`, which will also allow us to remove the local accessor that is just used to communicate the tile ID within the wokgroup.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,245 @@
...
+        ::std::size_t __current_num_wgs =
+            oneapi::dpl::__internal::__dpl_ceiling_div(__current_chunk_size, __elems_in_tile);
+        ::std::size_t __current_num_items = __current_num_wgs * __wgsize;
+
+        auto __event = __queue.submit([&](sycl::handler& __hdl) {
+            auto __tile_id_lacc = sycl::local_accessor<std::uint32_t, 1>(sycl::range<1>{1}, __hdl);
+            __hdl.depends_on(__prev_event);
+
+            oneapi::dpl::__ranges::__require_access(__hdl, __in_rng, __out_rng);
+            __hdl.parallel_for<class scan_kt_main>(
```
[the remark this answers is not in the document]
adamfidel: I changed this to use the user-provided kernel name.

test/kt/single_pass_scan.cpp
adamfidel: No, it works with other values as well. I have manually tested it with data per work item = [1, 2, 4, 8, 16, 32] and work group size = [64, 128, 256, 512, 1024]. However, perhaps it is a good idea to include multiple configurations in these tests as well.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: Yes, it is for Intel GPUs as per the design doc.

dmitriy-sobolev: I am not sure if it was discussed but we can consider adding it into gpu namespace instead of igpu.

If the implementation does not use any instructions specific to Intel GPUs (and is not expected to use), then it might perform well on other cards due to being highly-configurable. 

What do you think about it?

test/kt/single_pass_scan.cpp
SergeyKopienko, line 132: Because we already have some technique to generate the set of tests in the `kt` test folder, I propose to use this technique to generate one test for one set of parameters values too.
For details please see `CMakeLists.txt` in this folder.

SergeyKopienko:
@adamfidel could you please rebase this branch from current `main` branch state?

test/kt/single_pass_scan.cpp
dmitriy-sobolev: Thanks, Adam. Yes, it would be great to test each configuration, which is going to be declared as supported.

dmitriy-sobolev: Thanks, Adam. It would be great to test each configuration, which is going to be declared as supported.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,337 @@
...
+               ::std::size_t __current_num_items) const
+    {
+        return __q.submit([&](sycl::handler& __hdl) {
+            auto __tile_vals = sycl::local_accessor<_Type, 1>(sycl::range<1>{__elems_in_tile}, __hdl);
+            __hdl.depends_on(__prev_event);
+
+            oneapi::dpl::__ranges::__require_access(__hdl, __in_rng, __out_rng);
+            __hdl.parallel_for<_Name...>(
+                sycl::nd_range<1>(__current_num_items, __workgroup_size),
+                [=](const sycl::nd_item<1>& __item) [[intel::reqd_sub_group_size(SUBGROUP_SIZE)]] {
```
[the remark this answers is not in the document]
mmichel11: Usage of `[[_ONEDPL_SYCL_REQD_SUB_GROUP_SIZE_IF_SUPPORTED(__req_sub_group_size)]]` could potentially be problematic here since other portions of code assume the subgroup size is always `32`. If we want this solution to be device generic, I think we would also need to query the subgroup size at runtime in the kernel to be correct.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,245 @@
...
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _ONEDPL_parallel_backend_sycl_scan_H
+#define _ONEDPL_parallel_backend_sycl_scan_H
+
```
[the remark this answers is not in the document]
adamfidel: Done.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,115 @@
...
+    std::vector<T> expected(size);
+    generate_data(expected.data(), size, 42);
+
+    TestUtils::usm_data_transfer<_alloc_type, T> dt_input(q, expected.begin(), expected.end());
+    TestUtils::usm_data_transfer<_alloc_type, T> dt_output(q, size);
+
+    std::inclusive_scan(expected.begin(), expected.end(), expected.begin());
+
+    oneapi::dpl::experimental::kt::single_pass_inclusive_scan(q, dt_input.get_data(), dt_input.get_data() + size, dt_output.get_data(), ::std::plus<T>(), param)
```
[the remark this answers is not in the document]
adamfidel: I added testing for `std::mutliplies` here as well.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,115 @@
...
+    test_usm<T, sycl::usm::alloc::shared>(q, size, param);
+    test_usm<T, sycl::usm::alloc::device>(q, size, param);
+    test_sycl_iterators<T>(q, size, param);
+}
+
+
+int
+main()
+{
+    using T = int;
```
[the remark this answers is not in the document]
adamfidel: 64-bit types are now working.

test/kt/single_pass_scan.cpp
adamfidel: The tests have been updated to test multiple configurations for these values.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: I changed the signature to take the ranges as forwarding refs.

test/kt/single_pass_scan.cpp
adamfidel: Done.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,147 @@
...
+#if _ENABLE_RANGES_TESTING
+    test_all_view<T>(q, size, param);
+#endif
+}
+
+
+int
+main()
+{
+    using T = int;
```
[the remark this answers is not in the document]
adamfidel: The supported types are now listed in the CMake files and are being used to generate different executables for type.

test/kt/single_pass_scan.cpp
adamfidel: Done.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,399 @@
...
+    _BinaryOp __binary_op;
+    ::std::size_t __n;
+    _StatusFlags __status_flags;
+    ::std::size_t __status_flags_size;
+    _StatusValues __status_vals_full;
+    _StatusValues __status_vals_partial;
+    ::std::size_t __current_num_items;
+    _TileVals __tile_vals;
+
+    [[intel::reqd_sub_group_size(SUBGROUP_SIZE)]] void
```
mmichel11, line 179: I think we can use `sycl::reqd_sub_group_size` here since it is a part of the SYCL 2020.

adamfidel: Done.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+
+    ::std::size_t __current_num_wgs = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __elems_in_tile);
+    ::std::size_t __current_num_items = __current_num_wgs * __workgroup_size;
+
+    auto __prev_event =
+        __lookback_submitter<__data_per_workitem, __workgroup_size, _Type, _FlagType, _LookbackKernel>{}(
+            __queue, __fill_event, __in_rng, __out_rng, __binary_op, __n, __status_flags, __status_flags_size,
+            __status_vals_full, __status_vals_partial, __current_num_items);
+
+    if (0)
```
mmichel11, line 369: Are we planning to keep this synchronous or add back in the host task submission to support asynchronous calls?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+          __partial_vals_begin(__partial_vals), __atomic_flag(*(__flags + __tile_id + __padding)),
+          __atomic_partial_value(*(__partial_vals + __tile_id + __padding)),
+          __atomic_full_value(*(__full_vals + __tile_id + __padding))
+    {
+    }
+
+    void
+    set_partial(const _T __val)
+    {
+        __atomic_partial_value.store(__val, sycl::memory_order::release);
```
mmichel11, line 67: Do we need a memfence operation between the partial value and flag stores to guarantee the operations are not reordered?

Otherwise, I think we may be at risk of flag being updated first which could lead to reading a partial value before one is present.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+        __atomic_flag.store(__partial_status, sycl::memory_order::release);
+    }
+
+    void
+    set_full(const _T __val)
+    {
+        __atomic_full_value.store(__val, sycl::memory_order::release);
```
mmichel11, line 74: Similar comment as above with regards to a memory fence.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: In my understanding, if a consumer work-item observes the write to the flag, then that work-item will also observe the write to the partial value since the work-items are synchronizing on the same atomic variable using release-acquire. When the producer work-item releases the flag, all writes that occurred by the producer up to that point will be observed by the consumer work-item when it acquires the flag. Since the writes to the partial value and flag cannot be reordered within the work-item (as they are tagged with release semantics), then it shouldn't be possible for the consumer work-item to observe an old partial value if it observes an update to the flag.

At least, that is how I understood the SYCL memory consistency model (§3.8.3) and how it adapts the standard C++ memory consistency model.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: My response to the above comment should apply here as well.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: That is a good question. I do not like how it is written currently, as it introduces a synchronization while still giving the appearance that the algorithm itself is asynchronous. However, the asynchronous free introduces an unacceptably high slowdown, which I suspect is potentially a compiler bug.

I am not sure what the right call would be here, but in order to actually see a speedup with the kernel template, the synchronous free is needed currently unfortunately.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
mmichel11: Thanks for the pointers. I have read through the semantics of acquire-release and agree that this is safe.

I no longer believe there is a risk for the operations to be reordered.

danhoeflinger (commented):
This is not a full review, but I have a few bits of feedback to start with.  I'll return with more review this week.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+            // keep going through tiles until we either find a full tile or we've completely
+            // recomputed the prefix using partial values
+            if (__is_full_ballot_bits)
+                break;
+        }
+        return __running;
+    }
+
+    const uint32_t __tile_id;
+    _FlagType* __flags_begin;
```
mmichel11, line 124: Since we have separated this from the partial / full scan values, should we use a smaller type than `uint32_t` as `_FlagType`?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+    constexpr ::std::size_t __data_per_workitem = _KernelParam::data_per_workitem;
+
+    // Avoid non_uniform n by padding up to a multiple of workgroup_size
+    std::uint32_t __elems_in_tile = __workgroup_size * __data_per_workitem;
+    ::std::size_t __num_wgs = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __elems_in_tile);
+
+    constexpr int __status_flag_padding = SUBGROUP_SIZE;
+    std::uint32_t __status_flags_size = __num_wgs + 1 + __status_flag_padding;
+
+    _FlagStorageType* __status_flags = sycl::malloc_device<_FlagStorageType>(__status_flags_size, __queue);
```
mmichel11, line 352: Do you think one large `sycl::malloc_device` call and setting `__status_vals_full`, `__status_vals_partial`, and `__status_flags` to offsets within the allocation may be quicker than three separate mallocs?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
mmichel11: I think the best course of action would be to document this synchronous for now and bring it up with the compiler team to be improved.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
danhoeflinger: I believe that we are limited to 32bit /64 bit types with [atomics](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:atomic-references).  Also, not all devices have support for 64 bit atomics (FPGAs).

mmichel11: Ah, I see. My comment can be ignored then.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+inclusive_scan(sycl::queue __queue, _InRng&& __in_rng, _OutRng&& __out_rng, _BinaryOp __binary_op,
+               _KernelParam __param = {})
+{
+    return __impl::__single_pass_scan<true>(__queue, ::std::forward<_InRng>(__in_rng), ::std::forward<_OutRng>(__out_rng), __binary_op, __param);
+}
+
+template <typename _InIterator, typename _OutIterator, typename _BinaryOp, typename _KernelParam>
+sycl::event
+inclusive_scan(sycl::queue __queue, _InIterator __in_begin, _InIterator __in_end, _OutIterator __out_begin,
+               _BinaryOp __binary_op, _KernelParam __param = {})
```
danhoeflinger, line 397: Do we need an overload with `init` for our current targets?

(If so, this brings up the question of what sort of data do we accept as init:  scalar, device_ptr, single element "range" / iterator?)

test/kt/CMakeLists.txt
```diff
@@ -122,3 +123,50 @@ if (ONEDPL_TEST_ENABLE_KT_ESIMD)
     # Pin some cases to track them, e.g. because they fail
     _generate_esimd_sort_test("256" "32" "double" "" 1000) # segfault
 endif()
+
+function (_generate_igpu_scan_test _data_per_work_item _work_group_size _type)
```
danhoeflinger, line 127: Optional: It seems like we should be able to refactor this function and pull out some redundancy with `_generate_esimd_sort_test`.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,181 @@
...
+    std::cout << "TEST_DATA_PER_WORK_ITEM : " << TEST_DATA_PER_WORK_ITEM << "\n"
+              << "TEST_WORK_GROUP_SIZE    : " << TEST_WORK_GROUP_SIZE << "\n"
+              << "TEST_TYPE               : " << TypeInfo().name<TEST_TYPE>() << std::endl;
+#endif
+
+    constexpr oneapi::dpl::experimental::kt::kernel_param<TEST_DATA_PER_WORK_ITEM, TEST_WORK_GROUP_SIZE> params;
+    auto q = TestUtils::get_test_queue();
+    try
+    {
+        for (auto size : scan_sizes)
```
danhoeflinger, line 171: Optional:
Do we expect different problems to occur with different input data types?
I'm wondering with all the permutations, if we could get away with only running a subset of the input sizes on some input sequence data types for the sake of time.  For instance, we could always run all for USM device input data, but for USM shared, sycl_iterator, all_view, perhaps we could pick a subset of sizes (one single wg, one multi wg or something).

I think the same question could be asked of esimd radix sort.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
danhoeflinger: This makes sense to me as well unless our current targets require asynchronous versions. (I don't think they do)

danhoeflinger (commented):
Happy to discuss these comments in more detail offline, feel free to schedule a meeting if you think that would be helpful.

adamfidel:
> I do think it would be best if we better understand the compiler warning messages regarding generic address space operations we are seeing before merging.
> 
> For context if any other reviewers have seen something similar before:
> 
> ```
> warning: Adding 25 occurrences of additional control flow due to presence of generic address space operations
> in function ...
> ```

I've found that all of these warnings are stemming from the call to `sycl::joint_reduce` and `sycl::joint_inclusive_scan`. I am passing in a raw pointer from a `local_accessor` so I thought that changing it to a decorated `sycl::multi_ptr` would fix it, but it did not. I still need to track this down.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+template <::std::uint16_t __data_per_workitem, ::std::uint16_t __workgroup_size, typename _Type, typename _FlagType,
+          typename _KernelName>
+struct __lookback_submitter;
+
+template <::std::uint16_t __data_per_workitem, ::std::uint16_t __workgroup_size, typename _Type, typename _FlagType,
+          typename _InRng, typename _OutRng, typename _BinaryOp, typename _StatusFlags, typename _StatusValues,
+          typename _TileVals>
+struct __lookback_kernel_func
+{
+    using _FlagStorageType = typename _FlagType::_FlagType;
```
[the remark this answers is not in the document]
adamfidel: > Its a little confusing to me to have this overloaded name `_FlagType` describing both a `__scan_status_flag<T>` and also a `uint32_t` as defined in `__scan_status_flag<T>::_FlagType`. Can we adjust the name of this and similar template params to make these distinct?

Yes, this was confusing. I renamed the internal type to `_FlagStorageType` to make it clearer.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+// This file incorporates work covered by the following copyright and permission
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _ONEDPL_parallel_backend_sycl_scan_H
+#define _ONEDPL_parallel_backend_sycl_scan_H
```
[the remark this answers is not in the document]
adamfidel: > Probably should be `_ONEDPL_KT_SINGLE_PASS_SCAN_H`.

Done.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+    // Avoid non_uniform n by padding up to a multiple of workgroup_size
+    std::uint32_t __elems_in_tile = __workgroup_size * __data_per_workitem;
+    ::std::size_t __num_wgs = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __elems_in_tile);
+
+    constexpr int __status_flag_padding = SUBGROUP_SIZE;
+    std::uint32_t __status_flags_size = __num_wgs + 1 + __status_flag_padding;
+
+    _FlagStorageType* __status_flags = sycl::malloc_device<_FlagStorageType>(__status_flags_size, __queue);
+    _Type* __status_vals_full = sycl::malloc_device<_Type>(__status_flags_size, __queue);
+    _Type* __status_vals_partial = sycl::malloc_device<_Type>(__status_flags_size, __queue);
```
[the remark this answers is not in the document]
adamfidel: You are right, the values were not being freed correctly. I added proper calls to free with the other cleanup code.

For now, I think think maintaining the allocation/deallocation outside of the flag class might be better in terms of separation of responsibilities. However, there is an argument to be made that perhaps creating a separate class or set of functions to deal with allocation/deallocation might be beneficial.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: > Do you think one large `sycl::malloc_device` call and setting `__status_vals_full`, `__status_vals_partial`, and `__status_flags` to offsets within the allocation may be quicker than three separate mallocs?

Yes, I think it is better to use a single allocation where possible. I have updated the code to do that.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+        _Type __local_reduction =
+            sycl::joint_reduce(__group, __tile_vals_ptr, __tile_vals_ptr + __wg_local_memory_size, __binary_op);
+        _Type __prev_tile_reduction{};
+
+        // The first sub-group will query the previous tiles to find a prefix
+        if (__subgroup.get_group_id() == 0)
+        {
+            _FlagType __flag(__status_flags, __status_vals_full, __status_vals_partial, __tile_id);
+
+            if (__group.leader())
```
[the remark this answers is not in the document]
adamfidel: According to the SYCL 2020 spec, "The leader of the work-group is guaranteed to be the work-item with a local id of 0." However, from what I can tell, I do not find any guarantees that work-item 0 is guaranteed to be in the first subgroup. I find "work-items that compose a sub-group are chosen in an implementation defined way" which implies that it actually is not guaranteed.

Practically, it might be the case, but I will restructure this code so that it does not rely on something not explicitly guaranteed by the spec. I believe rewriting it from `__group.leader()` to `__subgroup.get_local_id() == 0` would suffice.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
danhoeflinger: Works for me.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
danhoeflinger: Right, this is what I found as well.  I agree with your proposal.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,436 @@
...
+
+    if constexpr (::std::is_same_v<_Type, _FlagStorageType>)
+    {
+        __status_flags = sycl::malloc_device<_FlagStorageType>(3 * __status_flags_size, __queue);
+        __status_vals_full = __status_flags + __status_flags_size;
+        __status_vals_partial = __status_flags + (2 * __status_flags_size);
+    }
+    else
+    {
+        __status_flags = sycl::malloc_device<_FlagStorageType>(__status_flags_size, __queue);
```
mmichel11, line 363: If we have a different `_FlagStorageType` and `_Type`, then I think we can still use a single `malloc_device` so long as we appropriately pad it for the data-type and then reinterpret the pointer. I believe the sort-kt is currently doing this.

Do you have any objection to this?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
dmitriy-sobolev: What about `cstdint`, `cassert`, `type_traits`?

test/kt/single_pass_scan.cpp
dmitriy-sobolev: I suggest doing that after revising the API. We may end up in reducing the amount of possible imputs.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+    _AtomicFlagT __atomic_flag;
+    _AtomicValueT __atomic_partial_value;
+    _AtomicValueT __atomic_full_value;
+};
+
+template <typename _FlagType, typename _Type, typename _BinaryOp, typename _KernelName>
+struct __lookback_init_submitter;
+
+template <typename _FlagType, typename _Type, typename _BinaryOp, typename... _Name>
+struct __lookback_init_submitter<_FlagType, _Type, _BinaryOp,
```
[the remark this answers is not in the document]
adamfidel: It is a bit complicated because we need to initialize the flags to different values based on whether it is out-of-bounds (padding) or not. We could possibly create a buffer on the host and copy it over, but I haven't done any testing to see if there is a performance difference from copying data from the host vs initializing it on the device itself with the kernel.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: The plan is to include a version with a single scalar init value to match the functionality of the original scan.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,436 @@
...
```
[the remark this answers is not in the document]
adamfidel: Thank you for providing these experiments. Based on these results, it appears that besides the initial kernel compilation, the performance of the initialization kernel vs a memset is essentially identical. Given that we are already compiling the main kernel, in my opinion, keeping the initialization kernel as is is a viable option.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: Good suggestion. I used some reinterpret_casts to condense this into a single allocation.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
danhoeflinger: Yeah, I think if it requires extra host-> device copy it will be worse.  

My hope was to be able to handle this logic with a subset of threads in the main kernel, and use default constructed "0" to mean "uninitialized".  Certainly not a blocker for this PR but its maybe worth investigating to see if it is possible (and a benefit) in a follow up PR.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
danhoeflinger: OK, so does that need to go in this PR or do you plan to add that in a follow up PR?

adamfidel: That's a good question. My goal was to have it implemented in this PR, but it is taking longer to implement + debug correctness than I expected. I am still planning to update this PR with the changes for an initial value, but if it is dragging too long then I might defer it to a follow up PR.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+        {
+            _ONEDPL_PRAGMA_UNROLL
+            for (std::uint32_t __i = 0; __i < __data_per_workitem; ++__i)
+            {
+                if (__wg_current_offset + __local_id + __workgroup_size * __i < __n)
+                {
+                    __tile_vals[__local_id + __workgroup_size * __i] =
+                        __in_rng[__wg_current_offset + __local_id + __workgroup_size * __i];
+                }
+            }
```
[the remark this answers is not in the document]
adamfidel: I'm not sure I follow the suggestion. The outer branch here (`__wg_next_offset <= n`) is giving a measurable performance increase in a small experiment that I tried (around 5% for an input of size 4M).

This branch is only taken for the very last work-group. Are you suggesting to rewrite this loop?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+    using _FlagType = uint32_t;
+    using _AtomicFlagT = sycl::atomic_ref<_FlagType, sycl::memory_order::acq_rel, sycl::memory_scope::device,
+                                          sycl::access::address_space::global_space>;
+    using _AtomicValueT = sycl::atomic_ref<_T, sycl::memory_order::acq_rel, sycl::memory_scope::device,
+                                           sycl::access::address_space::global_space>;
+
+    static constexpr _FlagType __initialized_status = 0;
+    static constexpr _FlagType __partial_status = 1;
+    static constexpr _FlagType __full_status = 2;
+    static constexpr _FlagType __oob_status = 3;
```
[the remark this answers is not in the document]
adamfidel: These have been changed to `_FlagStorageType` to be more consistent. With that change, I do believe it makes sense for them to exist inside the `__scan_status_flag` struct as their type is now tied to its internal storage type.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
dmitriy-sobolev: Yes, I agree that we can leave it as is. Reduced kernel compilation time is in "nice to have" category right now and it is a small kernel. I suggest adding TODO/issue if you find that advice possible to implement.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
danhoeflinger: I agree that this outer branch is the correct choice.  I'm wondering about what is in the `else`.  

Right now you unroll and check if each index is under `__n` in the innermost loop.
Another option would be to have some logic to calculate for each work-item the number of elements it will do something with, then loop from 0 to that number.  This avoids the branching in the inner-most loop, but at the cost of not-unrolling the loop.

I'm not sure it will be a benefit, and may be a small margin of a small subset of the computation. 

Not a blocker.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,413 @@
...
+                {
+                    __tile_vals[__local_id + __workgroup_size * __i] =
+                        __in_rng[__wg_current_offset + __local_id + __workgroup_size * __i];
+                }
+            }
+        }
+
+        auto __tile_vals_ptr = __dpl_sycl::__get_accessor_ptr(__tile_vals);
+        _Type __local_reduction =
+            sycl::joint_reduce(__group, __tile_vals_ptr, __tile_vals_ptr + __wg_local_memory_size, __binary_op);
```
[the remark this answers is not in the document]
adamfidel: We do have custom implementations of our own work-group reduce written by @julianmi. It was written specifically for the device-wide reduce algorithm and I haven't tried to incorporate it into this algorithm, although that would be a good future direction.

My thought process is that the work-group reduce is a rather small portion of the algorithm in terms of runtime conceptually, yet I do not have any data breaking down the runtime percentages so this could be an incorrect statement. Providing a breakdown of the percentages here would be helpful and is something that I can explore.

As far as a custom work-group scan, we have one in the main scan algorithm, but it exists as an alternative for the `sycl::joint_` scans when they cannot be used due to constraints of identity values / value types. From my understanding, it was written for the purpose of improving compatibility when the `sycl::joint_` scan cannot be used rather than for performance.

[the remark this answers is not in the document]
julianmi: Here are my thoughts on reduce: We have a single work-group implementation that would be closest to `sycl::joint_reduce`. I don't expect performance benefits of our generalized implementation over the highly optimized SYCL function though. The typical SLM required is the number of work-items times the data type size. Reusing this temporary memory could be beneficial.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,181 @@
...
+template <typename T, typename BinOp, typename KernelParam>
+void
+test_general_cases(sycl::queue q, std::size_t size, BinOp bin_op, KernelParam param)
+{
+    test_usm<T, sycl::usm::alloc::shared>(q, size, bin_op, param);
+    test_usm<T, sycl::usm::alloc::device>(q, size, bin_op, param);
+    test_sycl_iterators<T>(q, size, bin_op, param);
+#if _ENABLE_RANGES_TESTING
+    test_all_view<T>(q, size, bin_op, param);
+#endif
```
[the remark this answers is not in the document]
adamfidel: I added support and a test for plain `sycl::buffer`.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,436 @@
...
+    if ((__n_uniform & (__n_uniform - 1)) != 0)
+        __n_uniform = oneapi::dpl::__internal::__dpl_bit_floor(__n) << 1;
+
+    // Perform a single-work group scan if the input is small
+    if (oneapi::dpl::__par_backend_hetero::__group_scan_fits_in_slm<_Type>(__queue, __n, __n_uniform))
+    {
+        return oneapi::dpl::__par_backend_hetero::__parallel_transform_scan_single_group(
+            oneapi::dpl::execution::__dpl::make_device_policy(__queue), ::std::forward<_InRange>(__in_rng),
+            ::std::forward<_OutRange>(__out_rng), __n, oneapi::dpl::__internal::__no_op{}, unseq_backend::__no_init_value<_Type>{}, __binary_op, ::std::true_type{});
+    }
```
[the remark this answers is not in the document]
adamfidel: Could you elaborate on using `kernel_param` to reduce the number of compiled kernels? I'm not sure I follow.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,436 @@
...
+
+    constexpr ::std::size_t __workgroup_size = _KernelParam::workgroup_size;
+    constexpr ::std::size_t __data_per_workitem = _KernelParam::data_per_workitem;
+
+    // Avoid non_uniform n by padding up to a multiple of workgroup_size
+    std::uint32_t __elems_in_tile = __workgroup_size * __data_per_workitem;
```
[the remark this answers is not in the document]
adamfidel: Thanks, I updated the sizes to be `size_t` to match.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
dmitriy-sobolev: `__parallel_transform_scan_single_group` creates multiple kernels under the hood (~11 kernels for specific `N`). If we had  a separate implementation utilizing `kernel_param`, then we would end up only with one kernel. 

With `kernel_param`, user provides `data_per_workitem` and `workgroup_size`, which make dispatching for different `N` unnecessary.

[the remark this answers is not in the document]
adamfidel: I see your point now. I anticipate making this change would require a decent amount of re-work for the single-work-group scan and might be good to target in a later PR.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
danhoeflinger: Lets not worry about this within the context of this PR.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
julianmi: I'm unaware of documentation regarding the SLM requirements of our internal implementation. My statement concerning the SLM requirements was solely for our internal implementation, which uses the number of work-items times the data type size SLM space. 

I don't see a comment regarding SLM in the SYCL spec for `sycl::joint reduce`. It looks like the Intel compiler doesn't use SLM but instead a local reduction followed by a `reduce_over_group`, which typically uses shuffle and similar instructions to exchange data between work-items. See: https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/group_algorithm.hpp#L349-L354.

I'm not sure if the SLM can be reused. I suspect that this depends on how the runtime optimizes SLM allocation and deallocation.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,426 @@
...
+    ::std::size_t __status_flags_size_mem_size =
+        __status_flags_size * sizeof(_FlagStorageType) + (__status_flags_size * sizeof(_Type)) * 2;
+    ::std::size_t __status_vals_full_offset = __status_flags_size * sizeof(_FlagStorageType);
+    ::std::size_t __status_vals_partial_offset = __status_vals_full_offset + __status_flags_size * sizeof(_Type);
+
+    ::std::byte* __device_mem = sycl::malloc_device<::std::byte>(__status_flags_size_mem_size, __queue);
+    assert(__device_mem);
+
+    _FlagStorageType* __status_flags = reinterpret_cast<_FlagStorageType*>(__device_mem);
+    _Type* __status_vals_full = reinterpret_cast<_Type*>(__device_mem + __status_vals_full_offset);
```
dmitriy-sobolev, line 360: It seems that the pointer may be not aligned for `_Type` (i.e. its address may not be multiple to `_Type`), especially when `sizeof(_Type) > sizeof(_FlagStorageType)`, because the offset is `__status_flags_size * sizeof(_FlagStorageType)`.

Could you improve memory management to guarantee that the pointers are aligned for their types?

dmitriy-sobolev: Here is an example on [godbolt](https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGIAMzSrgAyeAyYAHI%2BAEaYxCAAnKQADqgKhE4MHt6%2BAdKp6Y4CoeFRLLHxSXaYDplCBEzEBNk%2BfoG2mPZFDPWNBCWRMXGJtg1NLbntCmP9YYPlwwkAlLaoXsTI7Bzm/mHI3lgA1Cb%2Bbk7TxJisJ9gmGgCCO3sHmMenbCwkAJ43d49muwY%2By8RxObgIX2SmAA%2BgRiExCApfg8nkCXm83MhpvhBMj/oDgaDTqIFEomnjUYTXmCIVDYfDERSAc8QdTidisFQKSjmWjWRisQR0LQ8NEbocAPQSw7YkAgFhiWioUwoh4ETAsZIGdVg/ZMUmHAAqeOiqE8hzwCmhYjwwHC6AgaAY00OADdUHh0IcAFSHZJwpaHBioTCqTb%2B44AdisD0OccOTC8RAt/uIbwAIodLmF1cRkpcCNCSQQwbKQF4c6nYTcIKmlicY/d41nMAR1gxDmBIHhU8czABWBMiu2oKgQQ1Lev%2BRsmSPpv4LgGcuZG7BCQ3Q%2B7BACSAHEIth56r7jnDgqwhB6yjo39m1LDthVKwta8wskkwoAHTf2/xrzpIxDmhQ1ITZTN0DWaJ6AbX843/MJgCAgAxAxgHqEgmGATAQKhDMZSFOUK0EfwzGradYPw9A5XSAAvGECCA6FpiYNsrSoVCrVosDDkkSMYOPRddioQ5l3tcc1w3AB5AAlPdtwiLcrybeNqIIkAuNhRjmNY6F2Mwzi8Do6EPiYwywIo5toSYhodL04ADLon0ZTM0cIGhFDMPQ%2BEsJwzBA2sQ43Oslj/10jjTMc30uNc4DQMnJyzAbSVpUtQ4CAQV4DBdNBNQERgGP4NMPm%2BABaS5tUwL1kiYdAcWAE4kIo1SqPUszNKs7TQtdMQ2K8WhaGhUcqCUBiTkzDqbNCuyHNeKKXLHdzUK8zDsLi/jlLjZrqLahiJpCq1utoK1qqaPAxEGqhhtbPC9p0w7ev6i6rtGywtMmtjwq4pzooW3ypxnWMVPUtToi%2BdVfSsrBXTwTZjI1PDs0EOJ81bIt9RLU4tpAUHwZrMsFX65Ugs6j79IimETK4yd1ubfUyQIIKoZhimNX%2BhdAbjRbPKIbzVqhCHgtsjiEcwHNkYLNHpjBLm0J5lbfO9GtIcwaHYY%2BNmOaAhW3v26F7t0vraBFsW8wl4tpYVpXoSZtX4YC26up6g3HqGkaNY2rXQIFkm9adk7HDEY2kdN1HzdOWL%2Batm2WZYPsLB1u6/b6M6Btd1t3Z2Vw8C5ATmWE0TKvE9dNx3fdDyU5ssY0hiTJtO1oWqr0xucuiYr%2BmmgbLautN16boRxzAFBuwWps%2BszvvmtyPNljCfLW8jNarnaE8do7ndTy6Rv7sHB%2BHn2%2B6%2BubW9%2B%2BfG0r4GWu7h2DqT07zrTwsB6H5vr7CsnD5bzA29PpqL%2B2ozdpwxYNvdUz9/DjSYr3cKT844rxvmvKghsnpbxgfbEe8DjrJ3vpvVGqDXq12HAwBuNV1oUXvMrVWMcLRDzEJcGqXwhy2ntIcQqCYGAMIUKIfqjQ0qgVIEGVACZaoZAEIHaYmBkg0MuEGTAlVKosJIGlDKLC8DEBdEoAAjl4Rgmxf5lgHgLaOQCg65hRoWMObgsYGLxmpAmSpkBBRMk/amC8PZ0ziAzChzMgEZ01jLZac9%2BZwLfvZEx4tQ7o2ltPAJfNMCK38NgRmKtvHqw7ptP%2BrUAGMUuOeBgWA0wvyASA3eJUe5CzJk/NJboPToAFtEfUMJezN0RqYs2kTTjuk9PExJXjbaxzQQoKBFSd4KHds2TptTGJ13tA3OEeEsbTIgAswhbBBAXT1hbUCNx%2BGv31ogl2ODH4jJ2f3BpsyyDZI1AiPJcQxnxgjnE4JezDZhJDuY9pbgHndKCtMyq5y7mc21rs2%2BAcjbNNFsHMxksMafMtgk4m71fYIJeQMxF%2Bt/Yp2QbgkZGcARZxzo8TW94xokvAaSzMbgMrIAANYZnJSSzWZY0BJgxGCPspF0EhIUCAdlrLTjBOmscTWzYRV8rcOy/hlprSEMqjy8wZgxWBSlb89AUSlpy0CZgK2%2B8OLxTBGyssrhaBVKZWsUapw2XyuBci/qcqzAKv1fy61D0wXCtFfGR14r5WSqtCqu1DqLX8ogMqmVqrw6%2BW1Wip2%2BzaB6sDeKw1eTjWuObKallnr2XOuIXfWg/rFVZoxYHCy7q4wZu9dQ6VTDZW8ozcG31obNlQkjbrdFWDY3%2BXjZROURqTVqSoJcTASTKE%2BNIXODgKxc0cH7LwPwHAtCkFQJwNw1gAoKDWBsakAIeCkAIJocdKxqUgEkAANk/JIfw/h%2BzHv7JGDQx7/AaEjAkSM0hJ2SBnXuhdnBeDco0DuvdKw4CwBgIgEAoYahJhIOQSgjR7LKEMJ0IQCBUAAHdZ3bpyskOgLFMjwfCLQJDqHZ3zsw3QYYwAuAkVIKR%2BgxAIisC2LwGjcRJJJkI2hzg27wPIHuMQey37SDcfqDiAT/BBAiDEOwKQMhBCKBUOoOdOg9AGCMCgFdlh9Cim5ZAFYqB/SZG5RwXgqBXRxGIJ6Qe8AVjEC8IIPAbBDRmiOlZ1Y6xNh6GxGEPDiHkMce4LwFD8JkicfHfoTg07SDEeM5wR8EGeaHFUAADmPSVY9khDjAGQMgQ4lHPwKogMuyw1h%2BG4EIIonYXAli8F3YpycpBD1mES5%2BK9kYBwPqfSRfwx771hY4O%2B0gLAQD9j/VFr9RnbAgD/TVrQgGQMQCQGa98BBoOOlQJqMjdGGOcCSyltLGWss5f8Hl3glUyvmaolwGTwguFSekGJ%2BQSg1Cft0JdwLTBgv%2BYneFj9imxusYIEtw4o4EvJdS%2BlzL2Xcv5Y8Ot2jfZ/CVeqwBlYGUarDEvL1/rg3huRc/Yu8bv7/21YPSAMwkYz2Ri4AkRLJ7b2JcaxoBI/hev%2BB%2B/O/HSPaukCA6B5jZAKCrdh%2BRyjZg%2BB0FzNyiA0RP3RDCI0L4IXSCy%2BYMQL4klojaBqDVjDa3VkEEkgwWgCvftYGiF4YAbhFSGe3VgBURhxAm9UVrvApnDPzu40mLY26xaTvnSKaI8JVceCwJ%2BuE9nFemeIKaJQ6Yrn24QgBvgHEABqeBMAocklCdDvB7sSfENJ%2B7cmnu/Ze/oQwxh1M2H99pjHenuiGeM5H87lmdMdC6JkFweSJh%2BEuyEOYZQKh6AKCIrInhWhD7SCPgYA/hiXeqLUAQvRxhj9yHPzozvF8zGn0MeIc%2BZjd481v/vO%2BJArDXW5qTX2p1s%2BixwEHu3wcHah4FUr8WKtVaJzNknZhT2M8kP/hIY9e1ZLfwJnV9TgfrUbDnCbKbZHXrMwG/MbTnL/UgSPdIZwSQIAA). I can make a patch.

adamfidel: Thanks for the godbolt example. I can incorporate what you have written into this branch.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,230 @@
...
+    test_buffer<T>(q, size, bin_op, param);
+#endif
+}
+
+template <typename T, typename KernelParam>
+void
+test_all_cases(sycl::queue q, std::size_t size, KernelParam param)
+{
+    test_general_cases<T>(q, size, std::plus<T>{}, param);
+    test_general_cases<T>(q, size, std::multiplies<T>{}, param);
```
dmitriy-sobolev, line 204: Could you make sure that there are no overflows? Perhaps, a different way of data generation or different sizes are needed.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
dmitriy-sobolev: Here is the patch: [aligned.patch](https://github.com/oneapi-src/oneDPL/files/15014668/aligned.patch)

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,426 @@
...
+
+    // Next power of 2 greater than or equal to __n
+    auto __n_uniform = ::oneapi::dpl::__internal::__dpl_bit_ceil(__n);
+
+    // Perform a single-work group scan if the input is small
+    if (oneapi::dpl::__par_backend_hetero::__group_scan_fits_in_slm<_Type>(__queue, __n, __n_uniform))
+    {
+        return oneapi::dpl::__par_backend_hetero::__parallel_transform_scan_single_group(
+            oneapi::dpl::__internal::__device_backend_tag{},
+            oneapi::dpl::execution::__dpl::make_device_policy<typename _KernelParam::kernel_name>(__queue),
```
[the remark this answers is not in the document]
adamfidel: Thanks, I added the missing headers.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,230 @@
...
+#    endif
+    std::vector<T> input(size);
+    generate_scan_data(input.data(), size, 42);
+    std::vector<T> ref(input);
+    sycl::buffer<T> buf_out(input.size());
+
+    std::inclusive_scan(std::begin(ref), std::end(ref), std::begin(ref), bin_op);
+    {
+        sycl::buffer<T> buf(input.data(), input.size());
+        oneapi::dpl::experimental::ranges::all_view<T, sycl::access::mode::read_write> view(buf);
```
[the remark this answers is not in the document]
adamfidel: Yes, it has been changed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -752,6 +754,20 @@ __parallel_transform_scan_base(oneapi::dpl::__internal::__device_backend_tag, _E
         __binary_op, __init, __local_scan, __group_scan, __global_scan);
 }
 
+template <typename _Type, typename _Queue>
+bool __group_scan_fits_in_slm(_Queue&& __queue, ::std::size_t __n, ::std::size_t __n_uniform)
```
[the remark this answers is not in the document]
adamfidel: Yeah, the template is not necessary here.

test/kt/CMakeLists.txt
```diff
@@ -122,3 +123,50 @@ if (ONEDPL_TEST_ENABLE_KT_ESIMD)
...
+
+    if ((NOT TARGET build-igpu-tests) AND (NOT TARGET run-igpu-tests))
+        add_custom_target(build-igpu-tests COMMENT "Build all scan kernel template tests")
+        add_custom_target(run-igpu-tests
+            COMMAND "${CMAKE_CTEST_COMMAND}" -R "^run-igpu-tests$" --output-on-failure --no-label-summary
+            DEPENDS build-igpu-tests
```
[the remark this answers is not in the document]
adamfidel: I changed it to `build-scan-kt-tests` since a target called `build-gpu-tests` sounds ambiguous to me (i.e., it sounds like it should encompass more than just the scan kernel template tests, which is not the intent).

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: Thanks for the patch. I applied it.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,434 @@
...
+
+    ::std::size_t __mem_align_pad = sizeof(_Type);
+    ::std::size_t __status_flags_bytes = __status_flags_size * sizeof(_FlagStorageType);
+    ::std::size_t __status_vals_full_offset_bytes = __status_flags_size * sizeof(_Type);
+    ::std::size_t __status_vals_partial_offset_bytes = __status_flags_size * sizeof(_Type);
+    ::std::size_t __mem_bytes = __status_flags_bytes + __status_vals_full_offset_bytes +
+                                __status_vals_partial_offset_bytes + __mem_align_pad;
+
+    ::std::byte* __device_mem = reinterpret_cast<::std::byte*>(sycl::malloc_device(__mem_bytes, __queue));
+    assert(__device_mem);
```
mmichel11, line 361: Should we do something similar to PR #1487 and throw `std::bad_alloc` if this is a nullptr?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,434 @@
...
+
+template <typename _InRng, typename _OutRng, typename _BinaryOp, typename _KernelParam>
+sycl::event
+inclusive_scan(sycl::queue __queue, _InRng&& __in_rng, _OutRng&& __out_rng, _BinaryOp __binary_op,
+               _KernelParam __param = {})
+{
+    auto __in_view = oneapi::dpl::__ranges::views::all(::std::forward<_InRng>(__in_rng));
+    auto __out_view = oneapi::dpl::__ranges::views::all(::std::forward<_OutRng>(__out_rng));
+
+    return __impl::__single_pass_scan<true>(__queue, ::std::move(__in_view), ::std::move(__out_view), __binary_op,
```
mmichel11, line 411: I think we should use `std::forward` here instead of `std::move`

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,434 @@
...
+    using _FlagType = __scan_status_flag<_Type>;
+    using _FlagStorageType = typename _FlagType::_FlagStorageType;
+
+    using _KernelName = typename _KernelParam::kernel_name;
+    using _LookbackInitKernel =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__lookback_init_kernel<_KernelName>>;
+    using _LookbackKernel =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__lookback_kernel<_KernelName>>;
+
+    const ::std::size_t __n = __in_rng.size();
```
mmichel11, line 317: Stylistic suggestion: it might be worth it to go ahead and convert `::std::` to `std::` in this PR's additions since it seems that will be our convention going forward.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: Yes, I replaced the assert with a throw.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: Sounds good. I went ahead and made that change in this file and the test.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
adamfidel: Both `__in_view` and `__out_view` are created in this function directly above this line (they are not the ranges passed in). Since the function ends here, it doesn't hurt to explicitly move these views as they go out of scope after this line.

test/kt/single_pass_scan.cpp
dmitriy-sobolev: Here is what I suggest doing: 
[test_improvement.patch](https://github.com/oneapi-src/oneDPL/files/15027882/test_improvement.patch)

The patch improves data generation and checks if the kernel template can be executed.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,435 @@
...
+    if (!__device_mem)
+        throw std::bad_alloc();
+
+    _FlagStorageType* __status_flags = reinterpret_cast<_FlagStorageType*>(__device_mem);
+    std::size_t __remainder = __mem_bytes - __status_flags_bytes;
+    void* __vals_base_ptr = reinterpret_cast<void*>(__device_mem + __status_flags_bytes);
+    void* __vals_aligned_ptr = std::align(std::alignment_of_v<_Type>, __status_vals_full_offset_bytes,
+                                            __vals_base_ptr, __remainder);
+    _Type* __status_vals_full = reinterpret_cast<_Type*>(__vals_aligned_ptr);
+    _Type* __status_vals_partial = reinterpret_cast<_Type*>(__status_vals_full + __status_vals_full_offset_bytes);
```
dmitriy-sobolev, line 370: Address arithmetic fix:

```suggestion
    _Type* __status_vals_partial = reinterpret_cast<_Type*>(__status_vals_full +
                                                            __status_vals_full_offset_bytes / sizeof(_Type));
```
There should've been a number of elements rather than a number of bytes.

test/kt/CMakeLists.txt
```diff
@@ -122,3 +123,50 @@ if (ONEDPL_TEST_ENABLE_KT_ESIMD)
...
+
+    foreach (_data_per_work_item ${_data_per_work_item_all})
+        foreach (_work_group_size ${_work_group_size_all})
+            foreach (_type ${_type_all})
+                _generate_gpu_scan_test(${_data_per_work_item} ${_work_group_size} ${_type})
+            endforeach()
+        endforeach()
+    endforeach()
+
+    _generate_test("single_pass_scan" "single_pass_scan.cpp")
```
dmitriy-sobolev, line 167: This is not needed now, I suppose.

test/kt/single_pass_scan.cpp
adamfidel: Thanks for the improvements. I incorporated them into this branch.

test/kt/CMakeLists.txt
adamfidel: I think it's valuable to have a target called just `single_pass_scan` that will choose a configuration that works out of the box, instead of having to decide which combination of data type, work-group size and iteration count to use for a quick test. That said, this target was broken so I added the correct compile def flags to get it working again.

dmitriy-sobolev (commented):
I've left some minor comments. 
Meanwhile, I'm looking at the kernel internals and will be back with the comments soon if any.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -0,0 +1,441 @@
...
+template <typename _FlagType, typename _Type, typename _BinaryOp, typename... _Name>
+struct __lookback_init_submitter<_FlagType, _Type, _BinaryOp,
+                                 oneapi::dpl::__par_backend_hetero::__internal::__optional_kernel_name<_Name...>>
+{
+    template <typename _StatusFlags, typename _PartialValues>
+    sycl::event
+    operator()(sycl::queue __q, _StatusFlags&& __status_flags, _PartialValues&& __partial_values,
+               std::size_t __status_flags_size, std::uint16_t __status_flag_padding) const
+    {
+        using _KernelName = __lookback_init_kernel<_Name..., _Type, _BinaryOp>;
```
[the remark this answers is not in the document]
dmitriy-sobolev: I suggest rebasing and reusing a utility function, which Dan has recently added: https://github.com/oneapi-src/oneDPL/blob/e3d8ba2857511c03d77bb1270537e73b948a1eca/test/support/utils.h#L901

test/support/test_config.h
```diff
@@ -164,4 +164,9 @@
 #    define _PSTL_ICPX_TEST_RED_BY_SEG_BROKEN_64BIT_TYPES 1
 #endif
 
+// Group reduction produces wrong results with multiplication of 64-bit for certain driver versions
+// TODO: When a driver fix is provided to resolve this issue, consider altering this macro or checking the driver version at runtime
+// of the underlying sycl::device to determine whether to include or exclude 64-bit type tests.
+#define _PSTL_GROUP_REDUCTION_MULT_INT64_BROKEN 1
+
```
akukanov, line 171: Is this for the same problem as the macro right above (which is also documented as a known issue), or for a different one?

adamfidel: I thought it was a different problem, but looking at it more, I think it's the same issue that the macros `_PSTL_ICPX_TEST_RED_BY_SEG_BROKEN_64BIT_TYPES` and `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` are meant for.

I know the internal ticket number for this issue, but it's hard to correlate that this is the same issue that these other macros are for because we do not (rightfully so) associate our internal issue tracking with these comments.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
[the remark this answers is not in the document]
adamfidel: Thanks, I've fixed the optional kernel name trick and I'm now using Dan's kernel param utility.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,230 @@
...
+            auto v = exp2(dist_real(gen));
+            return dist_binary(gen) == 0 ? v : -v;
+        };
+        std::generate(input, input + size, [&] { return randomly_signed_real(); });
+    }
+}
+
+template <typename OldParam, typename KernelName>
+using make_kernel_param =
+    oneapi::dpl::experimental::kt::kernel_param<OldParam::data_per_workitem, OldParam::workgroup_size, KernelName>;
```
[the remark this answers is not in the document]
adamfidel: I have updated the code to now use the helper in #1439.

test/kt/single_pass_scan.cpp
```diff
@@ -0,0 +1,236 @@
...
+
+    std::string msg = "wrong results with oneapi::dpl::begin/end, n: " + std::to_string(size);
+    EXPECT_EQ_RANGES(ref, output, msg.c_str());
+}
+
+template <typename T, typename BinOp, typename KernelParam>
+void
+test_general_cases(sycl::queue q, std::size_t size, BinOp bin_op, KernelParam param)
+{
+    test_usm<T, sycl::usm::alloc::shared>(q, size, bin_op, param);
```
dmitriy-sobolev, line 182: Could you use `TestUtils::get_new_kernel_params` at this level too?

dmitriy-sobolev, line 182: It's a matter of taste, but using `TestUtils::get_new_kernel_params` here instead of the functions below may be better if we want to add the same at `test_all_cases` level.

test/support/test_config.h
adamfidel: After an offline discussion with @mmichel11, we found that `_PSTL_ICPX_TEST_RED_BY_SEG_BROKEN_64BIT_TYPES` is for a separate issue.

The macro `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` seems to be from the same root issue, but I feel that we should still use a separate macro for the following reasons:

1. The `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` macro is expected to be defined externally through CMake rather than defined in `test_config.h` with the other macros
2. The macro's name suggests that when defined, it will apply some workaround for 64-bit group reductions with multiplies, whereas in this case, we want to explictly disable the tests cases with 64-bit types and the `std::multiplies` binary operator

dmitriy-sobolev (commented):
I've checked everything besides atomics, especially the specifics of the used memory ordering. Everything what I've checked looks good to me.

danhoeflinger (approved):
This LGTM.  I checked it far more thoroughly earlier on, but I have been reviewing delta commits and following the comments, and they look good to me.  

My outstanding issues have been:
1) init value, which we have decided to address in a later PR
2) Memory footprint in SLM of joint algorithms, and how to communicate memory footprint requirements to the end user to provide them the proper guidance for selecting kernel params.  I think this remains somewhat unresolved (as the sycl feature is a black box), but I think we can make a GH issue, and improve on this in the future (if we feel unsatisfied with our guidance in the docs with the current code).   I don't think it is worth holding back this PR for this.

It would probably be good to get another approval from one of the others who have been working on this.

dmitriy-sobolev (approved):
LGTM. I've reviewed the atomic operations and found no race conditions.

One more suggestions to consider doing in a separate PR: use regular (non-atomic) variables for values.

adamfidel:
Thanks all for the reviews!

@danhoeflinger and @dmitriy-sobolev, I will create GH issues to address the next steps that you have mentioned in your approvals.

PR #1325, opened 2024-01-02
Bugfix for permutation_iterator operator[]

danhoeflinger:
Fix for implementation of `permutation_view_simple` to return from `operator[]` by reference rather than by value.

Adding test for fill to cover missed test case.

Fixing some rules for modifying source data in permutation iterator tests, **this was resulting in a number of skipped tests.**

Following with other tests in this permutation_iterator.pass, checking of result is skipped.  Another PR should provide the checks for permutation iterator to properly evaluate if the result is achieved appropriately, and should apply the same permutation safely on the host to check the result.

The bug we experience and are fixing in this PR is at compile time and it is resolved.

### DIFF
--- a/include/oneapi/dpl/pstl/utils_ranges.h
+++ b/include/oneapi/dpl/pstl/utils_ranges.h
@@ -477,7 +477,7 @@ struct permutation_view_simple<_Source, _M, ::std::enable_if_t<oneapi::dpl::__in
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
-    auto
+    value_type&
     operator[](Idx __i) const
     {
         return __src[__map_fn(__i)];
@@ -516,7 +516,7 @@ struct permutation_view_simple<_Source, _M, ::std::enable_if_t<is_map_view<_M>::
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
-    auto
+    value_type&
     operator[](Idx __i) const
     {
         return __src[__map[__i]];
--- a/include/oneapi/dpl/pstl/utils_ranges.h
+++ b/include/oneapi/dpl/pstl/utils_ranges.h
@@ -477,8 +477,7 @@ struct permutation_view_simple<_Source, _M, ::std::enable_if_t<oneapi::dpl::__in
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
-    auto
-    operator[](Idx __i) const -> decltype(__src[__map_fn(__i)])
+    decltype(auto) operator[](Idx __i) const
     {
         return __src[__map_fn(__i)];
     }
@@ -516,8 +515,7 @@ struct permutation_view_simple<_Source, _M, ::std::enable_if_t<is_map_view<_M>::
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
-    auto
-    operator[](Idx __i) const -> decltype(__src[__map[__i]])
+    decltype(auto) operator[](Idx __i) const
     {
         return __src[__map[__i]];
     }
--- a/include/oneapi/dpl/pstl/utils_ranges.h
+++ b/include/oneapi/dpl/pstl/utils_ranges.h
@@ -477,7 +477,8 @@ struct permutation_view_simple<_Source, _M, ::std::enable_if_t<oneapi::dpl::__in
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
-    decltype(auto) operator[](Idx __i) const
+    decltype(auto)
+    operator[](Idx __i) const
     {
         return __src[__map_fn(__i)];
     }
@@ -515,7 +516,8 @@ struct permutation_view_simple<_Source, _M, ::std::enable_if_t<is_map_view<_M>::
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
-    decltype(auto) operator[](Idx __i) const
+    decltype(auto)
+    operator[](Idx __i) const
     {
         return __src[__map[__i]];
     }
--- a/include/oneapi/dpl/pstl/utils_ranges.h
+++ b/include/oneapi/dpl/pstl/utils_ranges.h
@@ -477,8 +477,8 @@ struct permutation_view_simple<_Source, _M, ::std::enable_if_t<oneapi::dpl::__in
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
-    value_type&
-    operator[](Idx __i) const
+    auto
+    operator[](Idx __i) const -> decltype(__src[__map_fn(__i)])
     {
         return __src[__map_fn(__i)];
     }
@@ -516,8 +516,8 @@ struct permutation_view_simple<_Source, _M, ::std::enable_if_t<is_map_view<_M>::
 
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
-    value_type&
-    operator[](Idx __i) const
+    auto
+    operator[](Idx __i) const -> decltype(__src[__map[__i]])
     {
         return __src[__map[__i]];
     }
--- a/test/parallel_api/iterator/permutation_iterator.h
+++ b/test/parallel_api/iterator/permutation_iterator.h
@@ -204,8 +204,6 @@ struct test_through_permutation_iterator<TSourceIterator, TSourceDataSize, perm_
     void
     operator()(Operand op)
     {
-
-        auto indexes_begin = dpl::counting_iterator<TSourceDataSize>(0);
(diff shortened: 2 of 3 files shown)


include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -478,7 +478,7 @@ struct permutation_view_simple<_Source, _M, ::std::enable_if_t<oneapi::dpl::__in
     //TODO: to be consistent with C++ standard, this Idx should be changed to diff_type of underlying range
     template <typename Idx>
     auto
-    operator[](Idx __i) const
+    operator[](Idx __i) const -> decltype(__src[__map_fn(__i)])
```
akukanov, line 481: Why is not the compiler able to deduce the return value type from the return statement?

danhoeflinger: I believe that the compiler is deducing and returning the `value_type` of the base iterator without this guide, and we need the `operator[]` to return the `reference_type`, the same as the return of the `operator[]` for the base iterator.

I'm not sure how the compiler makes this decision, but this guide makes it explicit and resolves the error.  I can look up the rules for auto return in this scenario to try to bring more clarity here.

test/parallel_api/iterator/permutation_iterator.pass.cpp
```diff
@@ -59,7 +60,7 @@ struct is_able_to_to_modify_src_data_in_test : ::std::true_type { };
 #if TEST_DPCPP_BACKEND_PRESENT
 template <typename ExecutionPolicy>
 struct is_able_to_to_modify_src_data_in_test<ExecutionPolicy, perm_it_index_tags::counting>
```
SergeyKopienko, line 62: @danhoeflinger could you please explain why you have changed the source implementations of  `struct is_able_to_to_modify_src_data_in_test` in test and exactly in which cases our previous implementation were incorrect and why?
Thanks.

include/oneapi/dpl/pstl/utils_ranges.h
danhoeflinger: Following up, I believe that without any trailing `-> ...`, auto uses the normal template argument deduction, which will result in the **referenced** type being selected, not the reference type. https://en.cppreference.com/w/cpp/language/template_argument_deduction#Deduction_from_a_function_call (3)

I've adapted the example from [here](https://dice-research.org/news/2022-04-08_cpp_return_type_auto/) to explore this topic a bit below:
https://godbolt.org/z/f8dr5eEea


It seems a more concise way of accomplishing this would be to just use `decltype(auto)` as the return with no trailing definition of the type.  This skips the template argument deduction entirely and is a placeholder which uses the decltype of the returned expression.  I don't have a preference between the two options.

SergeyKopienko: May be better to use `-> decltype(__src[__map_fn(__i)])` as you did because this approach already used in our code and it has been used here before https://github.com/oneapi-src/oneDPL/pull/1254

test/parallel_api/iterator/permutation_iterator.pass.cpp
danhoeflinger: Sure.  Let me first state my understanding of `is_able_to_to_modify_src_data_in_test` generally.  (which I now notice has a typo `..._to_to_...`)

`is_able_to_to_modify_src_data_in_test` is a struct which determines if we are able to modify base iterator data within a kernel which has a specific execution policy and a specific index iterator type (and therefore if we should run the test with this configuration).  I believe we assume that the base iterator data is modify-able using the selected execution policy by our test setup.  Therefore, what is important is if we can safely dereference the index iterator type in a kernel associated with the execution policy specified.  

Now on to my changes:
`is_able_to_to_modify_src_data_in_test<ExecutionPolicy, perm_it_index_tags::counting>` -> `std::true_type`.
A counting iterator is passed directly, and should be usable for both host and device kernels.  Previously we were skipping all device execution policy tests which used counting iterator indices for permutation iterator tests, with my changes they are no longer skipped.

`is_able_to_to_modify_src_data_in_test<ExecutionPolicy, perm_it_index_tags::usm_shared>` -> `std::true_type`.
usm_shared memory should be accessible both on the host and device.  Previously we were skipping all device execution policy tests for types using usm_shared index iterators.  With my changes they are no longer skipped.

`is_able_to_to_modify_src_data_in_test<ExecutionPolicy, perm_it_index_tags::usm_device>` -> `oneapi::dpl::__internal::__is_hetero_execution_policy<::std::decay_t<ExecutionPolicy>>`.
usm_device memory should be accessible only on the device.  Previously we were skipping all device execution policy tests for types using usm_device index iterators, but actually permitting it for host side policies.  With my changes, we  test with usm_device index iterators, but only on the device.

SergeyKopienko: @MikeDvorskiy could you please take a look to this change?

danhoeflinger: > As far as I remember we are unable to modify source data (it test this mean - unable to call algorithms with in-place data modification) for hetero policy with permutation iterator. And previous implementation were the same for all tags, it's true.

I'm not sure why this would be the case, assuming the source data is a valid supported type for permutation iterator.

include/oneapi/dpl/pstl/utils_ranges.h
MikeDvorskiy: > this would be to just use `decltype(auto)`

I remember that Ruslan recommended to use this way.

SergeyKopienko:
@danhoeflinger I think may be better to merge this PR after my https://github.com/oneapi-src/oneDPL/pull/1338

danhoeflinger:
> @danhoeflinger I think may be better to merge this PR after my #1338

Sure that is fine, but lets try to get it in for 2022.5.0.

SergeyKopienko:
> > @danhoeflinger I think may be better to merge this PR after my #1338
> 
> Sure that is fine, but lets try to get it in for 2022.5.0.

Yes, of course.

include/oneapi/dpl/pstl/utils_ranges.h
danhoeflinger: Considering it further, I like `decltype(auto)` better as it removes a redundant bit of code and is more maintainable.  I've made the change.

test/parallel_api/iterator/permutation_iterator.pass.cpp
```diff
@@ -56,7 +56,7 @@ struct is_able_to_modify_src_data_in_test : ::std::true_type { };
 #if TEST_DPCPP_BACKEND_PRESENT
 template <typename ExecutionPolicy>
 struct is_able_to_modify_src_data_in_test<ExecutionPolicy, perm_it_index_tags::counting>
```
SergeyKopienko, line 58: I believe all specializations of 
struct is_able_to_modify_src_data_in_test<ExecutionPolicy, perm_it_index_tags::yyyyy>
derived from `std::true_type` not required anymore.

test/parallel_api/iterator/permutation_iterator.h
```diff
@@ -204,6 +204,8 @@ struct test_through_permutation_iterator<TSourceIterator, TSourceDataSize, perm_
     void
     operator()(Operand op)
     {
+
+        auto indexes_begin = dpl::counting_iterator<TSourceDataSize>(0);
```
SergeyKopienko, line 208: could you please move this line before line 221 ?

test/parallel_api/iterator/permutation_iterator.pass.cpp
danhoeflinger: Agreed.  I think we actually may be able to enable all these cases, and perhaps even remove this check...  Investigating it now.

SergeyKopienko:
@danhoeflinger, could you please refresh the table `// Table: Current test state` at https://github.com/oneapi-src/oneDPL/blob/f61bfad0bcfc1712a758ea054273528baea48778/test/parallel_api/iterator/permutation_iterator.pass.cpp#L38 ?

danhoeflinger:
> @danhoeflinger, could you please refresh the table `// Table: Current test state` at
> 
> https://github.com/oneapi-src/oneDPL/blob/f61bfad0bcfc1712a758ea054273528baea48778/test/parallel_api/iterator/permutation_iterator.pass.cpp#L38
> 
> ?

Yes, I will.  Tentatively, I think we just say all are now supported.  I'm confirming this, but I'll update it soon.

test/parallel_api/iterator/permutation_iterator.h
danhoeflinger: yes, done. thanks.

danhoeflinger:
>  I think we just say all are now supported. I'm confirming this, but I'll update it soon.

I attempted to do this, but hit a segmentation fault for `perm_it_index_tags::host` with one of the modifying tests.  I am not sure yet which test is the culprit, or why.  From my understanding, there should be nothing preventing the use of host iterators as the index data, as they should be properly processed into `sycl::buffers`.  

host iterators are a problem as the base/source data for `permutation_iterators` because we can't know the size requirements for the `sycl::buffer` to load them into, but host iterators as index data should be known in size and properly handled for any execution policy.

I'll continue to investigate this.  If it is an easy fix, I'll include it.  Otherwise, I suggest we merge and create an issue.

SergeyKopienko (approved):
LGTM

PR #1326, opened 2024-01-02
Adding legacy support for is_passed_directly trait

danhoeflinger:
This PR returns legacy support for types having the `is_passed_directly` trait provided.
#1254 changed the way that this was implemented internally by instead enumerating the types which were passed directly with template specialization of `struct is_passed_directly`.  

While this is an implementation detail, it is a useful one which has been used in the compatibility headers to support `device_iterator` and `device_pointer`, and may be used in some user code. This returns the support for the `is_passed_directly` trait, to augment the new system.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -231,8 +231,19 @@ struct is_passed_directly<zip_iterator<Iters...>> : ::std::conjunction<is_passed
 {
 };
 
+template <typename Iter, typename Void = void>
+struct is_passed_directly_legacy_support : ::std::false_type
+{
+};
+
+template <typename Iter> // for permutation_iterators
+struct is_passed_directly_legacy_support<Iter, ::std::enable_if_t<Iter::is_passed_directly::value>> : ::std::true_type
+{
+};
+
 template <typename Iter>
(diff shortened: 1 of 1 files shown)


mmichel11 (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -231,8 +231,19 @@ struct is_passed_directly<zip_iterator<Iters...>> : ::std::conjunction<is_passed
...
 
+template <typename Iter, typename Void = void>
+struct is_passed_directly_legacy_support : ::std::false_type
+{
+};
+
+template <typename Iter> // for permutation_iterators
+struct is_passed_directly_legacy_support<Iter, ::std::enable_if_t<Iter::is_passed_directly::value>> : ::std::true_type
+{
+};
```
akukanov, line 242: I would call it ``is_passed_directly_legacy_trait``.

Why does the comment mention permutation iterators?

danhoeflinger: Name changed, thanks.

I don't think `std::conditional` simplifies anything, because we need to handle 3 cases for `is_passed_directly`, `true_type`, `false_type` and missing.  

The comment was made in error, good catch.  A permutation iterator was triggering the error that hit this bug, but its used for more than just that.  I removed it.

MikeDvorskiy (approved):
Good fix to me.

PR #1327, opened 2024-01-03
Adding read-only permission to all jobs in the CI flow to ensure minimum required access

timmiesmith:
The change is being made to bring the repo into compliance with the recommended best practice of enforcing the policy of least privilege.  See https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions for the details.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,8 @@ on:
       - main
       - 'release/**'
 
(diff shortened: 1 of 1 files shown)


danhoeflinger (approved):
LGTM.
Looks like we could also make some changes to address "Pin actions to a full length commit SHA", but that is probably another PR.

PR #1332, opened 2024-01-05
[Dynamic Selection] Modify select functions in policies to initialize if resources are empty

AnuyaWelling2801:


include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
```diff
@@ -217,6 +217,10 @@ class auto_tune_policy
         static_assert(sizeof...(KeyArgs) == sizeof...(Args));
         if (state_)
         {
+            if (state_->resources_with_index_.empty())
+            {
+                initialize();
```
egfefey, line 222: This will be fine for backends like sycl that can auto-initialize resources. However, for other backends that resources are manually initialized, this will fail. A message indicating that "no devices were found" or some combination of the two might be better.

include/oneapi/dpl/internal/dynamic_selection_impl/dynamic_load_policy.h
```diff
@@ -155,6 +155,10 @@ struct dynamic_load_policy
             std::unique_lock<std::mutex> l(state_->m_);
             std::shared_ptr<resource_t> least_loaded;
             int least_load = std::numeric_limits<load_t>::max();
+            if (state_->resources_.empty())
+            {
+                initialize();
```
egfefey, line 160: same here

include/oneapi/dpl/internal/dynamic_selection_impl/fixed_resource_policy.h
```diff
@@ -111,7 +111,10 @@ struct fixed_resource_policy
             {
                 return selection_type{*this, state_->resources_[state_->offset_]};
             }
-            return selection_type{*this};
+            else{
+                initialize();
```
egfefey, line 115: same here

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
```diff
@@ -115,6 +115,11 @@ struct round_robin_policy
     {
         if (state_)
         {
+
+            if (state_->resources_.empty())
+            {
+                initialize();
```
egfefey, line 121: same here

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/dynamic_load_policy.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/fixed_resource_policy.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/round_robin_policy.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
vossmjp: How would we get into a situation where state_ is not null and resources_with_index is empty, but calling initialize might still succeed?  I'm not sure what scenario this PR is trying to address?

AnuyaWelling2801: The case would be that initialize from the backend (not be a sycl_backend, something in the future)  would return 0 resources.

PR #1340, opened 2024-01-12
Address issues in algorithms identified by static analysis

timmiesmith:
Static analysis has identified a function marked as noexcept where the SYCL code called may through an exception.  A separate issue identified is the use of an object after it was forwarded.  This PR addresses both issues.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/numeric_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/numeric_ranges_impl_hetero.h
@@ -93,6 +93,7 @@ __pattern_transform_scan_base(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Rang
 {
     if (__rng1.empty())
         return 0;
+    oneapi::dpl::__internal::__difference_t<_Range2> __rng1_size = __rng1.size();
 
     using _Type = typename _InitType::__value_type;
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/numeric_ranges_impl_hetero.h
```diff
@@ -93,6 +93,7 @@ __pattern_transform_scan_base(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Rang
 {
     if (__rng1.empty())
         return 0;
+    oneapi::dpl::__internal::__difference_t<_Range2> __rng1_size = __rng1.size();
```
mmichel11, line 96: Should this type be `oneapi::dpl::__internal::__difference_t<_Range1>`?

mmichel11 (approved):
LGTM

include/oneapi/dpl/pstl/hetero/numeric_ranges_impl_hetero.h
danhoeflinger: I'm fine with it as is or with `auto`, which would convert during the return (but would make the code less verbose).

danhoeflinger (approved):
LGTM

PR #1341, opened 2024-01-12
Dev/dhoefling/permutation iter host iterator index fix

danhoeflinger:
This PR fixes a bug with the handling of input data for `permutation_iterator` where a `std::vector` host iterator is provided as the map for the permutation iterator.  

Prior to this PR, when a `permutation_iterator` is marked as `sycl::access_mode::write`, we used this `access_mode` for both source data and map data.  However, the map data for a `permutation_iterator` will always require `sycl::access_mode::read` (only).  Even when a `permutation_iterator` is written to, the map is merely used as an index / offset into the base data, so it must be read, not written regardless of the requested access mode.

This change insulates the data processing functions within `__get_sycl_range` from the external `operator()` function, and adds a layer, `__process_input_iter`, to allow more discrete control of access mode, and specifically split access mode inputs within a single `__get_sycl_range` object like `permutation_iterator`.  

This PR also re-enables a some skipped tests for `permutation_iterator`, and removes unnecessary code in the tests.


test/support/iterator_utils.h
```diff
@@ -257,6 +257,12 @@ using is_base_of_iterator_category = ::std::is_base_of<Tag, typename iterator_tr
 template <typename Tag, typename Iter>
 inline constexpr bool is_base_of_iterator_category_v = is_base_of_iterator_category<Tag, Iter>::value;
 
+template <typename Iter>
+using is_base_of_iterator_random_access = is_base_of_iterator_category<::std::random_access_iterator_tag, Iter>;
+
+template <typename Iter>
+inline constexpr bool is_base_of_iterator_random_access_v = is_base_of_iterator_random_access<Iter>::value;
```
SergeyKopienko, line 264: Technically it's correct.
But in tests there are a lot of places when we should use this define too,
But we still use `::std::enable_if_t<is_base_of_iterator_category<::std::random_access_iterator_tag, OutputIterator>::value>` and so far and so on,
So my proposal - to use `is_base_of_iterator_category_v` in this test too.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -574,23 +575,23 @@ struct __get_sycl_range
...
+                              __range_holder<oneapi::dpl::__ranges::all_view<val_t<_Iter>, _LocalAccMode>>>
     {
         using _T = val_t<_Iter>;
 
-        return __process_host_iter_impl(__first, __last, [&]() {
-            if constexpr (__is_copy_direct)
+        return __process_host_iter_impl<_LocalAccMode>(__first, __last, [&]() {
+            if constexpr (__is_copy_direct_v<_LocalAccMode>)
             {
                 //wait and copy on a buffer destructor; an exclusive access buffer, good performance
```
SergeyKopienko, line 596: I propose to add before this line assertion check:
`assert(!__is_copy_back_v<_LocalAccMode>);`

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -631,28 +632,37 @@ struct __get_sycl_range
...
-        return __range_holder<oneapi::dpl::__ranges::all_view<_T, AccMode>>{
-            oneapi::dpl::__ranges::all_view<_T, AccMode>(__buf)};
+        return __range_holder<oneapi::dpl::__ranges::all_view<_T, _LocalAccMode>>{
+            oneapi::dpl::__ranges::all_view<_T, _LocalAccMode>(__buf)};
+    }
+
+  public:
+    template <typename... _ArgType>
+    auto
+    operator()(_ArgType... __arg)
```
MikeDvorskiy, line 662: A minor comment. Usually we use a plural form for "argument pack"  - `"operator()(_ArgTypes... __args)"`

MikeDvorskiy (commented):
Good fix.
Just one comment before an approval. It would be good to have 2 commits: 1st - a bugfix fo2 perm iterator, 2nd - changes in the test.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: I don't understand why this is needed.  Can you explain further?
I think  `__is_copy_back_v<_LocalAccMode> == true` is a valid state here, (and it will be handled below with `__buf.set_write_back(__is_copy_back_v<_LocalAccMode>);` within `__process_host_iter_impl()`

test/support/iterator_utils.h
danhoeflinger: I've removed this in favor of just using the existing infrastructure.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: Fixed.

danhoeflinger:
@MikeDvorskiy 
> It would be good to have 2 commits: 1st - a bugfix fo2 perm iterator, 2nd - changes in the test.

I agree, I've rebased to separate it to these two commits.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
SergeyKopienko: In the next function we have
```C++
                if constexpr (__is_copy_back_v<_LocalAccMode>)
                    __buf.set_final_data(__first); //SYCL API for non-contiguous iterators
```
In this place this code isn't required, because by semantic `__is_copy_back_v<_LocalAccMode>` is `false`.
I propose to add assertion for this case.

test/support/iterator_utils.h
SergeyKopienko: OK

MikeDvorskiy (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: I'm not seeing how `__is_copy_back_v<_LocalAccMode>` is `false` by semantic here.  It could be either by my understanding.  Copy back is dependent on the access mode, and has nothing to do with if the input data is contiguous, so we can't determine that copy back isn't required based on this.  We don't special case based on this here because it isn't required with this `sycl::buffer` constructor.

This function is for contiguous host iterators, so it is able to utilize the host pointer + range constructor for `sycl::buffer`, rather than the host container `sycl::buffer` (which requires an extra host side allocation and copy of the data).

For this `sycl::buffer`, constructor with the host pointer, the "copy back" can merely be provided by the `__buf.set_write_back(__is_copy_back_v<_LocalAccMode>);` which is called when this lambda returns in `__process_host_iter_impl()`.  For the other non-contiguous version, it requires the extra step of `__buf.set_final_data(__first)`, because it is using the host container `sycl::buffer` constructor.  This tells the `sycl::buffer` that when it is destroyed, it should copy the data back to a specific host iterator.  

For the host pointer constructor, it just uses the supplied memory address as its host side memory, so no extra host side copy back actually occurs (after the device -> host copy).

@MikeDvorskiy  does this agree with your understanding as well?

MikeDvorskiy: Yes, `__is_copy_back_v<_LocalAccMode> == true `is a valid state here.

The fact that we don't call set_final_data doesn't  tell that copying will no be, because a buffer with exclusive access does it by default (according to the spec).
Probably, it makes sense to add a detailed comment why buffer with exclusive access has different behavior and  does copy back on destructor by default while buffer(it1, it2) doesn't. Dan, could you please to add such additional comment?

SergeyKopienko: Hope @MikeDvorskiy will propose some comment about this for insert into the code.

danhoeflinger: Sure I can add these comments.

SergeyKopienko (approved):
LGTM

MikeDvorskiy (approved):
LGTM

PR #1342, opened 2024-01-12
Correct preprocessor SYCL target checks for reduce and min/max element

mmichel11:
The recently merged PRs https://github.com/oneapi-src/oneDPL/pull/1235 and https://github.com/oneapi-src/oneDPL/pull/1224 introduced performance improvements for architectures that benefit from coalescing.

Benchmarking these additions, I did not see any performance change. It seems that this is due to to the preprocessor checks that enable the coalescing paths. The initial checks looked to see if `_ONEDPL_DETECT_SPIRV_COMPILATION` is defined. However, for the SYCL backend this is always defined and set to a value of `0` or `1`.

What we should do here is check the value of this macro to enable the coalescing path when appropriate. I have confirmed that these changes enable the performance benefit of these PRs to be seen.


include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -428,7 +428,7 @@ __pattern_min_element(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __
     using _IteratorValueType = typename ::std::iterator_traits<_Iterator>::value_type;
     using _IndexValueType = ::std::make_unsigned_t<typename ::std::iterator_traits<_Iterator>::difference_type>;
     using _ReduceValueType = tuple<_IndexValueType, _IteratorValueType>;
-#ifdef _ONEDPL_DETECT_SPIRV_COMPILATION
+#if defined(_ONEDPL_DETECT_SPIRV_COMPILATION) && _ONEDPL_DETECT_SPIRV_COMPILATION
```
danhoeflinger, line 431: Why do we check if it is defined in this situation, but not in the other?
Looking at `_ONEDPL_SYCL_USM_HOST_PRESENT` and `_ONEDPL_LIBSYCL_VERSION` as an example, it seems that it is reasonable to not check if it is defined before referencing it directly.

mmichel11: I added this check here since in this file we are not guaranteed to be using the SYCL backend whereas we are in `unseq_backend_sycl.h`.

However, it is okay to just check `_ONEDPL_DETECT_SPIRV_COMPILATION` directly as it will evaluate to 0 if undefined. I will make this change.

mmichel11:
These changes are causing a few CI tests to fail which I am looking into.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: Aha, fair enough.  Yes, I believe the C++ standard states explicitly that these undefined PP identifiers are replaced by `0`.

mmichel11:
I have also fixed the CMake issues that I believe are causing the CI failures when the coalesced paths are enabled.

mmichel11:
The CI failures have been addressed by a patch to `__pattern_min_element` so this PR is ready for the remainder of review.

I have also added a comment in `_ONEDPL_DETECT_SPIRV_COMPILATION` to clarify how it should be used, so we can avoid this issue in the future.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -428,23 +428,22 @@ __pattern_min_element(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __
     using _IteratorValueType = typename ::std::iterator_traits<_Iterator>::value_type;
     using _IndexValueType = ::std::make_unsigned_t<typename ::std::iterator_traits<_Iterator>::difference_type>;
     using _ReduceValueType = tuple<_IndexValueType, _IteratorValueType>;
-#ifdef _ONEDPL_DETECT_SPIRV_COMPILATION
-    using _Commutative = ::std::false_type;
-    // This operator doesn't track the lowest found index in case of equal min. or max. values. Thus, this operator is
-    // not commutative.
+    using _Commutative = ::std::true_type;
```
danhoeflinger, line 431: I believe this is a bug.  We are hardcoding to be commutative = true, but it is not always the case in practice as it depends on `_ONEDPL_DETECT_SPIRV_COMPILATION`.  We check this in unseq_backend_sycl.h and can end up in code which does not match our operator.

Additionally, I think it is a design problem to embed this knowledge of commutativity in the kernel name by having it in the type if it might depend upon knowledge that can only be known at JIT time.  We may want to rethink it.


With this in mind, here is my tentative proposal:  
1) Add a new struct:
`_CommutativeUnknown` which is passed in place of `std::true_type` or `std::false_type` as the `_Commutative` template parameter to make it clear that this type should not be used to make actual decisions in this pattern as we do not know at host compilation time the status of this.

2) add a struct `__is_commutative_known` which determines if we can make a decision from the _Commutative value, or if we need to have special casing.  

3) add special handling in `unseq_backend_sycl.h` to use `__is_commutative_known` and if it is unknown, use a direct check of `_ONEDPL_DETECT_SPIRV_COMPILATION` and choose the appropriate code.  I believe these are the only active places where `_Commutative` is used in practice to functionally change the code, but this should be confirmed.  If there are more, we need similar handling of a third potential value for `_Commutative`.

This probably needs more thought, but I believe something like this will be needed.

test/CMakeLists.txt
```diff
@@ -178,8 +178,8 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
 
     set(coal_tests "reduce.pass" "transform_reduce.pass" "count.pass" "sycl_iterator_reduce.pass" "minmax_element.pass")
     if (_test_name IN_LIST coal_tests)
-        onedpl_construct_exec(${test_source_file} ${_test_name} ${switch_off_checked_iterators} "-D_ONEDPL_DETECT_SPIRV_COMPILATION=1")
-        onedpl_construct_exec(${test_source_file} ${_test_name}.coal ${switch_off_checked_iterators} "-D_ONEDPL_DETECT_SPIRV_COMPILATION=0")
+        onedpl_construct_exec(${test_source_file} ${_test_name} ${switch_off_checked_iterators} _ONEDPL_DETECT_SPIRV_COMPILATION=1)
```
adamfidel, line 181: What is the rationale for removing `-D` for these defines here? According to the CMake documentation for [target_compile_definitions](https://cmake.org/cmake/help/latest/command/target_compile_definitions.html), the `-D` should automatically be removed.

> Any leading -D on an item will be removed.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: After some offline discussion, I have some updates to this proposal.  

Add a new struct which provides a matrix of commutative true / false decisions for spriv and non-spirv targets, which the host knows and can specify at host compile time.  `template<typename _SpirvCommutative, typename _NonSpirvCommutative> struct _CommutativeMatrix`

Then add a helper struct to check if commutative which has overloads for `std::true_type` and `std::false` type for backward compatibility, but also for `_CommutativeMatrix` which  uses `_ONEDPL_DETECT_SPIRV_COMPILATION` to decide commutativity at kernel compilation time to result in the correct decision.
  
The kernel name will be consistent since all the information is in the `_Commutative` type, and the choice is at kernel compilation time.

test/CMakeLists.txt
danhoeflinger: Does the `"` wrapping get in the way here or is that removed by cmake in the passing of and evaluating the argument?

Its possible that the only cmake problem was the lack of `${...}`

mmichel11: I took a second look and the quotes here are not affecting correctness, so I will revert this change (along with removing -D).

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
mmichel11: I have went ahead and added my implementation of these new structs to address this issue and test CI.

I am looking at the best way to consolidate `_ONEDPL_DETECT_SPIRV_COMPILATION` checks preferably in the structs to cleanup up the code and have a single field in `__commutative_matrix` here.

test/CMakeLists.txt
mmichel11: I have reverted these unnecessary changes. The only CMake change that remains is the bugfix to properly use the `custom_define` variable.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
mmichel11: I have pushed the change to maintain a single `value` field in `__is_commutative_operator`.

As discussed offline, the `_ONEDPL_DETECT_SPIRV_COMPILATION` checks remain to force the non-commutative path for SPIRV targets.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -264,8 +265,10 @@ struct transform_reduce
     operator()(const _NDItemId& __item_id, const _Size& __n, const _Size& __global_offset, const _AccLocal& __local_mem,
                const _Acc&... __acc) const
     {
-#ifndef _ONEDPL_DETECT_SPIRV_COMPILATION
-        if constexpr (_Commutative::value)
+        // For non-SPIRV targets, we check if the operator is commutative before selecting the appropriate codepath.
+        // On SPIRV targets, we always force the non-commutative implementation as this is more performant.
```
danhoeflinger, line 269:
```suggestion
        // On SPIRV targets, we always force the sequential implementation as this is more performant.
```

Lets try to talk about sequential vs non-sequential implementations, and non-commutative vs commutative operators, so that we do our best to not conflate one with the other.  

The non-commutative operator forces sequential implementation, but the either implementation is acceptable for commutative operators.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -275,8 +278,8 @@ struct transform_reduce
     _Size
     output_size(const _Size& __n, const ::std::uint16_t& __work_group_size) const
     {
-#ifndef _ONEDPL_DETECT_SPIRV_COMPILATION
-        if constexpr (_Commutative::value)
+#if !_ONEDPL_DETECT_SPIRV_COMPILATION
```
danhoeflinger, line 281: If I understand correctly, this `output_size` call correlates with the sequential vs non-sequential implementations.  
It seems error prone to be split like this. 

For now, I'd settle for a comment similar to above but it seems like we should be able to provide a constexpr helper which determines "sequential" vs "non-sequential" and then includes the commutative check and the `_ONEDPL_DETECT_SPIRV_COMPILATION` in one consolidated place for the `transform_reduce` struct.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: looks good.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
mmichel11: Good point, I have added this change.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
akukanov: I like the idea overall, but I do not like that the implementation is de-facto split across 3 different files.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -681,6 +682,21 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
...
+};
+
+template <typename _SpirvCommutative, typename _NonSpirvCommutative>
+struct __is_commutative_operator<oneapi::dpl::__internal::__commutative_matrix<_SpirvCommutative, _NonSpirvCommutative>>
+    :
+#if _ONEDPL_DETECT_SPIRV_COMPILATION
+    _SpirvCommutative
+#else
+    _NonSpirvCommutative
+#endif
```
adamfidel, line 697: These preprocessor directives can be replaced in a single line using a `std::conditional`.

```suggestion
    : std::conditional_t<_ONEDPL_DETECT_SPIRV_COMPILATION, _SpirvCommutative, _NonSpirvCommutative>
```

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -259,24 +259,30 @@ struct transform_reduce
             __local_mem[__local_idx] = __res;
         }
     }
+
+    // For non-SPIR-V targets, we check if the operator is commutative before selecting the appropriate codepath.
+    // On SPIRV targets, we force the non-commutative implementation as this is currently more performant.
+    constexpr static bool
+    use_nonseq_impl()
+    {
+        return oneapi::dpl::__internal::__is_spirv_target::value && _Commutative::type::value;
```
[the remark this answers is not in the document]
mmichel11: Good catch. I have added this change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: Thanks for the suggestion. The code has been moved around a bit to consolidate the implementation.

In the case we are not using the SYCL backend, this macro will be undefined which will result in compilation issues with `std::conditional`, so I think using the preprocessor directives is best here.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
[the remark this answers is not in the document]
mmichel11: This comment has been moved into the `use_nonseq_impl` helper with the suggested changes.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
mmichel11: I have consolidated this in the `use_nonseq_impl` helper function.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -695,6 +695,20 @@ __iterators_possibly_equal(_Iterator1 __it1, _Iterator2 __it2)
...
+struct __is_spirv_target : ::std::conditional_t<_ONEDPL_DETECT_SPIRV_COMPILATION, ::std::true_type, ::std::false_type>
+{
+};
+
+// Conditionally sets type to _SpirvT if oneDPL is being compiled to a SPIR-V target with the SYCL backend and _NonSpirvT otherwise.
+// This should be used only in kernels.
+template <typename _SpirvT, typename _NonSpirvT>
+struct __spirv_target_selector
+{
+    using type = ::std::conditional_t<_ONEDPL_DETECT_SPIRV_COMPILATION, _SpirvT, _NonSpirvT>;
```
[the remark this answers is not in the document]
mmichel11: Added

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -57,7 +57,8 @@
 // TODO: determine which compiler configurations provide subgroup load/store
 #define _ONEDPL_SYCL_SUB_GROUP_LOAD_STORE_PRESENT false
 
-// Macro to check if we are compiling for Intel devices
+// Macro to check if we are compiling for Intel devices. This macro must only be used within
+// SYCL kernels for determining SPIRV compilation. Using this macro on the host may lead to incorrect behavior.
```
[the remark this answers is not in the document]
mmichel11: I have added a similar comment to the other macro and have changed "Intel devices" to "SPIR-V devices".

include/oneapi/dpl/pstl/utils.h
```diff
@@ -695,6 +695,26 @@ __iterators_possibly_equal(_Iterator1 __it1, _Iterator2 __it2)
...
+// Conditionally sets type to _SpirvT if oneDPL is being compiled to a SPIR-V target with the SYCL backend and _NonSpirvT otherwise.
+// This should be used only in kernels.
+template <typename _SpirvT, typename _NonSpirvT>
+struct __spirv_target_selector
+{
+    using type =
+#if _ONEDPL_DETECT_SPIRV_COMPILATION
+        _SpirvT;
+#else
+        _NonSpirvT;
```
rarutyun, line 707:
```suggestion
        std::bool_constant<!_SpirvT::value>;
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -695,6 +695,26 @@ __iterators_possibly_equal(_Iterator1 __it1, _Iterator2 __it2)
     }
 }
 
+// Conditionally sets type to _SpirvT if oneDPL is being compiled to a SPIR-V target with the SYCL backend and _NonSpirvT otherwise.
+// This should be used only in kernels.
+template <typename _SpirvT, typename _NonSpirvT>
```
rarutyun, line 700:
```suggestion
template <typename _IsSpirvT>
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -695,6 +695,26 @@ __iterators_possibly_equal(_Iterator1 __it1, _Iterator2 __it2)
...
+        _NonSpirvT;
+#endif
+};
+
+// Trait that has a true value if _ONEDPL_DETECT_SPIRV_COMPILATION is set and false otherwise. This may be used within kernels
+// to determine SPIR-V targets.
+struct __is_spirv_target : __spirv_target_selector<::std::true_type, ::std::false_type>
+{
+};
+inline constexpr bool __is_spirv_target_v = __is_spirv_target::type::value;
```
rarutyun, line 716:
```suggestion
inline constexpr bool __is_spirv_target_v = __is_spirv_target::value;
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -695,6 +695,26 @@ __iterators_possibly_equal(_Iterator1 __it1, _Iterator2 __it2)
...
+#if _ONEDPL_DETECT_SPIRV_COMPILATION
+        _SpirvT;
+#else
+        _NonSpirvT;
+#endif
+};
+
+// Trait that has a true value if _ONEDPL_DETECT_SPIRV_COMPILATION is set and false otherwise. This may be used within kernels
+// to determine SPIR-V targets.
+struct __is_spirv_target : __spirv_target_selector<::std::true_type, ::std::false_type>
```
rarutyun, line 713:
```suggestion
struct __is_spirv_target : __spirv_target_selector<::std::true_type>
```

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -259,24 +259,30 @@ struct transform_reduce
...
         }
     }
+
+    // For non-SPIR-V targets, we check if the operator is commutative before selecting the appropriate codepath.
+    // On SPIRV targets, we force the non-commutative implementation as this is currently more performant.
+    constexpr static bool
+    use_nonseq_impl()
+    {
+        return !oneapi::dpl::__internal::__is_spirv_target_v && _Commutative::type::value;
+    }
```
rarutyun, line 269:
```suggestion
    static constexpr bool use_nonseq_impl =  !oneapi::dpl::__internal::__is_spirv_target_v && _Commutative::value;
```

include/oneapi/dpl/pstl/utils.h
rarutyun, line 713: By the way, I am not sure but perhaps we don't even need a `struct`. We might try 

```cpp
inline constexpr bool __is_spirv_target_v = __spirv_target_selector<std::true_type>::value;
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -695,6 +695,26 @@ __iterators_possibly_equal(_Iterator1 __it1, _Iterator2 __it2)
     }
 }
 
+// Conditionally sets type to _SpirvT if oneDPL is being compiled to a SPIR-V target with the SYCL backend and _NonSpirvT otherwise.
+// This should be used only in kernels.
+template <typename _SpirvT, typename _NonSpirvT>
+struct __spirv_target_selector
```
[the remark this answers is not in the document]
mmichel11: I agree with having just one template parameter here. Added.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -428,33 +428,36 @@ __pattern_min_element(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __
     using _IteratorValueType = typename ::std::iterator_traits<_Iterator>::value_type;
     using _IndexValueType = ::std::make_unsigned_t<typename ::std::iterator_traits<_Iterator>::difference_type>;
     using _ReduceValueType = tuple<_IndexValueType, _IteratorValueType>;
-#ifdef _ONEDPL_DETECT_SPIRV_COMPILATION
-    using _Commutative = ::std::false_type;
-    // This operator doesn't track the lowest found index in case of equal min. or max. values. Thus, this operator is
-    // not commutative.
+    using _Commutative = oneapi::dpl::__internal::__spirv_target_selector</*_SpirvT*/ ::std::false_type,
```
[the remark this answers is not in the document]
mmichel11: I have made this change.

include/oneapi/dpl/pstl/utils.h
mmichel11: I have made this change.

include/oneapi/dpl/pstl/utils.h
mmichel11: I have made this change.

include/oneapi/dpl/pstl/utils.h
mmichel11: I have dropped the usage of this struct in favor of just directly defining the variable.

include/oneapi/dpl/pstl/utils.h
mmichel11: `__is_spirv_target` has been removed.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
mmichel11: I agree it is better for this to just be a variable. Changed.

include/oneapi/dpl/pstl/utils.h
mmichel11: Yes, I do not think the struct has much use. I have made this change.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -695,6 +695,21 @@ __iterators_possibly_equal(_Iterator1 __it1, _Iterator2 __it2)
...
+// Conditionally sets type to _IsSpirvT if oneDPL is being compiled to a SPIR-V target with the SYCL backend and the negation otherwise.
+template <typename _IsSpirvT>
+struct __spirv_target_selector :
+#if _ONEDPL_DETECT_SPIRV_COMPILATION
+    _IsSpirvT
+#else
+    ::std::bool_constant<!_IsSpirvT::value>
+#endif
+{
+};
```
akukanov, line 707: I am not sure I agree with the change to just one parameter. You have made an assumption that the parameter is always a bool_constant, and that the values are always opposite. That's sufficient for the current use case, but loses some generality and orthogonality of the original design (which could have been applied to any type, e.g. float and double). Also the code readability is worse, because `using _Commutative = oneapi::dpl::__internal::__spirv_target_conditional<::std::false_type, ::std::true_type>` is more self-explaining than `using _Commutative = oneapi::dpl::__internal::__spirv_target_selector<::std::false_type>`.

mmichel11: I agree that `using _Commutative = oneapi::dpl::__internal::__spirv_target_conditional<::std::false_type, ::std::true_type>` is more readable in the code that uses this utility and there seems to be stronger consensus to implement this in a more generic way to enable non-bool types.

@rarutyun Do you have any objection to going back to the original design of the `__spirv_target_selector` struct? This would add back the second template parameter type and define a `type` alias in the body in of the struct.

mmichel11: I had an offline discussion with @rarutyun and have reverted back to using the previous implementation of `__spirv_target_selector`.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -259,24 +259,27 @@ struct transform_reduce
             __local_mem[__local_idx] = __res;
         }
     }
+
+    // For non-SPIR-V targets, we check if the operator is commutative before selecting the appropriate codepath.
+    // On SPIRV targets, we force the non-commutative implementation as this is currently more performant.
```
akukanov, line 264:
```suggestion
    // For non-SPIR-V targets, we check if the operator is commutative before selecting the appropriate codepath.
    // On SPIRV targets, the sequential implementation with the non-commutative operator is currently more performant.
```

akukanov:
As you can see I only suggest to improve comments, while the implementation looks good to me (consider it approved).

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
mmichel11: I have added this suggestion (and changed SPIRV to SPIR-V to be consistent).

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -428,33 +428,36 @@ __pattern_min_element(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __
...
-#ifdef _ONEDPL_DETECT_SPIRV_COMPILATION
-    using _Commutative = ::std::false_type;
-    // This operator doesn't track the lowest found index in case of equal min. or max. values. Thus, this operator is
-    // not commutative.
+    using _Commutative = oneapi::dpl::__internal::__spirv_target_conditional</*_SpirvT*/ ::std::false_type,
+                                                                             /*_NonSpirvT*/ ::std::true_type>;
     auto __reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
         using ::std::get;
-        if (__comp(get<1>(__b), get<1>(__a)))
+        //TODO: Develop a long-term solution to the commutativity property, or remove the non commutative implementation for SPIRV
```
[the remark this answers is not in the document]
mmichel11: I agree that these first part of the comment with regards to a long-term solution can be removed. I'll wait for Dan's opinion before committing this.

danhoeflinger (approved):
LGTM

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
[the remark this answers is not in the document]
mmichel11: I have removed this part of the comment and refactored the wording in the second part.

danhoeflinger (approved):

mmichel11:
@adamfidel @rarutyun @akukanov Are there any additional review comments that you all would like to make in this PR, or are we ready to merge?

test/CMakeLists.txt
```diff
@@ -77,7 +77,7 @@ macro(onedpl_construct_exec test_source_file _test_name switch_off_checked_itera
     target_compile_definitions(${_test_name} PRIVATE _PSTL_TEST_SUCCESSFUL_KEYWORD=1)
 
     if (NOT ${custom_define} STREQUAL "")
-        target_compile_definitions(${_test_name} PRIVATE custom_define)
+        target_compile_definitions(${_test_name} PRIVATE ${custom_define})
```
SergeyKopienko, line 80: May we describe this change as separate  issue and fix it in the independent PR ?

mmichel11: Discussed offline and agreed that the CMake bugfix can be left as a part of this PR.

akukanov (approved):

PR #1346, opened 2024-01-15
Add workaround to parallel merge in TBB backend not to use input iterators default constructor

kboyarinov:
Add workaround to implementation of __parallel_merge function for the following use case:

```
#include <vector>

int main() {
    std::vector<int> input1 = {1, 2, 3};
    std::vector<int> input2 = {1, 2, 3};
    std::vector<int> output = {0, 0, 0};

    auto no_transformation = [](int i) { return i; };

    oneapi::dpl::merge(oneapi::dpl::execution::par,
                       oneapi::dpl::make_transform_iterator(input1.begin(), no_transformation),
                       oneapi::dpl::make_transform_iterator(input1.end(), no_transformation),
                       input2.begin(),
                       input2.end(),
                       output.begin());
}
```

The first input iterator in `merge` algorithm is `dpl::transform_iterator<vector_it, lambda>`. Since lambda is not default constructible, this use case results in the compilation error while trying to default construct iterator inside TBB backend.

From the C++ Standard perspective, it is not an issue, because default constructability is part of ForwardIterator requirements, but after the short offline discussion, it was decided to workaround this place in parallel backend TBB to make things better for our fancy iterators.

There is still no 100% guarantee, that no issues would be present for transform iterator with lambda body, because some algorithms we use from the C++ Standard Library are still allowed to call default constructor on the input iterator, e.g. `std::upper_bound`). This PR was tested together with libstdc++ and no issues was found.

### DIFF
--- a/test/parallel_api/iterator/iterators.pass.cpp
+++ b/test/parallel_api/iterator/iterators.pass.cpp
@@ -268,7 +268,7 @@ struct test_transform_iterator {
         auto list_it1 = oneapi::dpl::make_transform_iterator(f_list.begin(), ref_functor);
         auto list_it2 = oneapi::dpl::make_transform_iterator(f_list.end(), ref_functor);
         ::std::fill(list_it1, list_it2, 7);
-        EXPECT_TRUE(::std::all_of(f_list.begin(), f_list.end(), [](int x){ return x == 7; }),
+        EXPECT_TRUE(::std::all_of(f_list.begin(), f_list.end(), [](int x){ return x == 7; }), 
             "wrong result from fill with forward_iterator wrapped with transform_iterator");
 
         auto test_lambda = [](T2& x){ return x + 1; };


include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -424,7 +424,9 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc())
+    transform_iterator() = default;
```
SergeyKopienko, line 427: Is this new constructor really needed?
If yes, should it be explicit?

include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -1248,25 +1249,26 @@ operator()(__task* __self)
         return nullptr;
     }
 
-    _RandomAccessIterator1 __xm;
-    _RandomAccessIterator2 __ym;
+    ::std::optional<_RandomAccessIterator1> __xm;
```
SergeyKopienko, line 1252: As far as I found, we have another place like this for OpenMP backend.

include/oneapi/dpl/pstl/iterator_impl.h
kboyarinov: Yes, it is needed.
As I mentioned in the description:

"We are also required to make change in transform_iterator default constructor, because calling the default constructor of the std::optional with the type which default constructor is ill-formed (non-deleted) is also ill-formed on clang (see this example: https://godbolt.org/z/MM93c5GWx). Because of this we need the default constructor to be deleted if the body of the iterator is not default constructible. From the first glance, it is still aligned with the oneDPL specification."

I will try to formulate reasonable comment for the code before merging this.

include/oneapi/dpl/pstl/parallel_backend_tbb.h
kboyarinov: Yes, OMP backend should be fixed as well. Thanks.

include/oneapi/dpl/pstl/iterator_impl.h
SergeyKopienko: what about `explicit` ?

SergeyKopienko:
I think we are able to fix this issue without changes in premutation_iterator and it's more preferable: for that we should duplicate some code in `if` / 'else` parts inside https://github.com/oneapi-src/oneDPL/pull/1346/files#diff-9908a4470cd112b8a7c7c87c5d3dfa425d873a5e5f9bf353268e6bdf756f5af8.

include/oneapi/dpl/pstl/iterator_impl.h
kboyarinov: This part was removed from the PR.
In any case, `explicit` is orthogonal and would not fix the previous issue.

include/oneapi/dpl/pstl/parallel_backend_tbb.h
kboyarinov: Fixed

include/oneapi/dpl/pstl/iterator_impl.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/parallel_backend_tbb.h
SergeyKopienko: ok

kboyarinov:
> What about the code
> 
> ```c++
>         // Using callable object instead of lambda here to ensure transform iterator would be
>         // default constructible, that is part of the Forward Iterator requirements in the C++ standard.
>         struct NoTransform
>         {
>             ValueType operator()(const ValueType& val) const
>             {
>                 return val;
>             }
>         };
> ```
> 
> Should we use lambda now in
> 
> https://github.com/oneapi-src/oneDPL/blob/5740f585386bed798af83e1c18ae1ee92c613637/test/parallel_api/iterator/permutation_iterator.h#L211
> 
> ?

I would prefer to keep this test as-is because, as I mentioned this fix is not a 100% guarantee that merge with transform_iterator with lambda body would work. `std::upper_bound` and `std::lower_bound` implementations can still rely on default constructability of input iterators.
For today, this code works, but formally such an iterator violates the ForwardIterator requirements from the Standard. 
This patch is intended to minimize the risk of issues, as we discussed offline.

SergeyKopienko:
> > What about the code
> > ```c++
> >         // Using callable object instead of lambda here to ensure transform iterator would be
> >         // default constructible, that is part of the Forward Iterator requirements in the C++ standard.
> >         struct NoTransform
> >         {
> >             ValueType operator()(const ValueType& val) const
> >             {
> >                 return val;
> >             }
> >         };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Should we use lambda now in
> > https://github.com/oneapi-src/oneDPL/blob/5740f585386bed798af83e1c18ae1ee92c613637/test/parallel_api/iterator/permutation_iterator.h#L211
> > 
> > ?
> 
> I would prefer to keep this test as-is because, as I mentioned this fix is not a 100% guarantee that merge with transform_iterator with lambda body would work. `std::upper_bound` and `std::lower_bound` implementations can still rely on default constructability of input iterators. For today, this code works, but formally such an iterator violates the ForwardIterator requirements from the Standard. This patch is intended to minimize the risk of issues, as we discussed offline.

ok

SergeyKopienko (approved):
LGTM

akukanov:
Should the patch also be done in PSTL upstream?
@MikeDvorskiy @rarutyun

include/oneapi/dpl/pstl/omp/parallel_merge.h
```diff
@@ -39,20 +46,29 @@ __parallel_merge_body(std::size_t __size_x, std::size_t __size_y, _RandomAccessI
...
 
-    _RandomAccessIterator1 __xm;
-    _RandomAccessIterator2 __ym;
-
     if (__size_x < __size_y)
     {
-        __ym = __ys + (__size_y / 2);
-        __xm = std::upper_bound(__xs, __xe, *__ym, __comp);
+        _RandomAccessIterator2 __ym = __ys + (__size_y / 2);
+        __make_parallel_merge_leaf_tasks(::std::upper_bound(__xs, __xe, *__ym, __comp),
```
MikeDvorskiy, line 52: Recently, we came to conclusion that we should not write `::` operator before `std::` in new code, at least.

MikeDvorskiy:
> Should the patch also be done in PSTL upstream? @MikeDvorskiy @rarutyun

I guess, yes it should.

PR #1347, opened 2024-01-15
[Doc] Fix doc generation: pin sphinxcontrib_applehelp version

dmitriy-sobolev:
`sphinxcontrib_applehelp`, which `docutils` depends on, recently updated its version. This version is not compatible with `sphinx` versions older than `5`, while we use `3`. See **https://github.com/sphinx-doc/sphinxcontrib-applehelp/pull/15#issuecomment-1890824244** for more details.

### DIFF
--- a/documentation/library_guide/requirements.txt
+++ b/documentation/library_guide/requirements.txt
@@ -3,6 +3,8 @@ sphinx==3.4.0
 jinja2<3.1
 # https://github.com/sphinx-doc/sphinx/issues/9923
 docutils==0.15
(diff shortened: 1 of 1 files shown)


dmitriy-sobolev:
@ValentinaKats, This PR can be merged as a short-term or mid-term fix.
I am also going to try another approach separately: updating Sphinx to avoid such issues in the future. It may take more time to investigate given many other pinned dependencies.

ValentinaKats (approved):
LGTM

PR #1349, opened 2024-01-16
Split premutation_iterator test

SergeyKopienko:
In this PR we split `permutation_iterator` test to set of small independent tests - one test per pattern.
The goal - to reduce compile time, test run-time and detect potential bugs for each pattern easy.


MikeDvorskiy (approved):
LGTM

PR #1350, opened 2024-01-16
Update preprocessor check for sub-group mask extension

mmichel11:
The oneAPI sub-group mask extension used in radix sort has more than a single version. Currently, we check if the `SYCL_EXT_ONEAPI_SUB_GROUP_MASK` macro is equal to one which will miss capturing newer versions of this extension. This results in a significant loss of performance in radix sort.

Instead, we need to check if the version number of this macro is `1` or greater.


dmitriy-sobolev (approved):
Indeed, the macro has multiple values according to [the specification](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_sub_group_mask.asciidoc#feature-test-macro).

The change looks good to me. 

Minor (and optional) thing: you can also consider verifying if the macro is defined instead of relying on versions as suggested by the specification.

danhoeflinger (approved):
LGTM

mmichel11:
> Indeed, the macro has multiple values according to [the specification](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_sub_group_mask.asciidoc#feature-test-macro).
> 
> The change looks good to me.
> 
> Minor (and optional) thing: you can also consider verifying if the macro is defined instead of relying on versions as suggested by the specification.

I think verifying if the macro defined would require a bit code (probably some #if ... #else ... preprocessor conditional and manually setting the macro to 0 or 1), so I think leaving it as it is currently is the most concise.

PR #1352, opened 2024-01-17
Temp fix to commutivity issue

AidanBeltonS:
This PR adds a temporary fix to the min element lambda. It adds an assumption that the non-comm approach is taken for the SPIRV target path, regardless of the commutivity template parameter passed.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -430,7 +430,8 @@ __pattern_min_element(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __
     using _ReduceValueType = tuple<_IndexValueType, _IteratorValueType>;
     using _Commutative = ::std::true_type;
     auto __reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
-//TODO: Remove the non commutative implementation for SPIRV when we no longer always use sequential loads
+//TODO: Develop a long-term solution to the commutativity property, or remove the non commutative implementation for SPIRV
+//when we no longer always use sequential loads.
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -428,23 +428,21 @@ __pattern_min_element(_ExecutionPolicy&& __exec, _Iterator __first, _Iterator __
     using _IteratorValueType = typename ::std::iterator_traits<_Iterator>::value_type;
     using _IndexValueType = ::std::make_unsigned_t<typename ::std::iterator_traits<_Iterator>::difference_type>;
     using _ReduceValueType = tuple<_IndexValueType, _IteratorValueType>;
-#if _ONEDPL_DETECT_SPIRV_COMPILATION
-    using _Commutative = ::std::false_type;
-    // This operator doesn't track the lowest found index in case of equal min. or max. values. Thus, this operator is
-    // not commutative.
+    using _Commutative = ::std::true_type;
     auto __reduce_fn = [__comp](_ReduceValueType __a, _ReduceValueType __b) {
+//TODO: Remove the non commutative implementation for SPIRV when we no longer always use sequential loads
```
mmichel11, line 433: Small suggestion here: I think there is still potential that we may need to keep the non-commutative implementation, so I recommend changing this comment to something like, 

"TODO: Develop a long-term solution to manage the commutativity property, or remove the non commutative implementation for SPIRV when we no longer always use sequential loads."

mmichel11:
Just one comment about the TODO. Everything else LGTM.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
AidanBeltonS: Done

mmichel11 (approved):

PR #1354, opened 2024-01-18
Redefine transform_reduce's scratch & result mem

AidanBeltonS:
## Rationale

Algorithms using `transform_reduce` (`std::reduce`, `std::max_element`, etc) return a single value result. They also, except for the smallest cases, require intermediate scratch memory on the device to hold partial results.

For L0 backend, it's faster to have a 1-element USM host allocation and to write the final reduction result directly to that.

For Nvidia, host USM is expensive, and it's faster instead to have a single USM device allocation for both the **result** and the **intermediate scratch**.

## Approach

This PR combines the two approaches into a struct `__usm_host_or_unified_storage`, based on the previous `__usm_host_or_buffer_storage`. When host USM is supported and the backend is L0, this struct holds two memory allocations (device USM for intermediate scratch, and host USM for final result). In all other cases, this struct holds a single device USM allocation, large enough for both intermediate scratch and final result. In this latter case, a `memcpy` from device to host is needed to return the final result.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -183,9 +183,9 @@ struct __parallel_transform_reduce_device_kernel_submitter<_Tp, __work_group_siz
               oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0,
               typename... _Ranges>
     auto
-    operator()(_ExecutionPolicy&& __exec, _Size __n, _ReduceOp __reduce_op, _TransformOp __transform_op,
-               _InitType __init, __result_and_scratch_storage<_ExecutionPolicy2, _Tp> __scratch_container,
-               _Ranges&&... __rngs) const
+    operator()(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Size __n,
+               _ReduceOp __reduce_op, _TransformOp __transform_op, _InitType __init,
+               __result_and_scratch_storage<_ExecutionPolicy2, _Tp> __scratch_container, _Ranges&&... __rngs) const
     {
         auto __transform_pattern =
             unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _TransformOp, _Tp,
@@ -228,8 +228,8 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<
               typename _ExecutionPolicy2,
               oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0>
     auto
-    operator()(_ExecutionPolicy&& __exec, sycl::event& __reduce_event, _Size __n, _ReduceOp __reduce_op,
-               _TransformOp __transform_op, _InitType __init,
+    operator()(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, sycl::event& __reduce_event,
+               _Size __n, _ReduceOp __reduce_op, _TransformOp __transform_op, _InitType __init,
                __result_and_scratch_storage<_ExecutionPolicy2, _Tp> __scratch_container) const
     {
         using _NoOpFunctor = unseq_backend::walk_n<_ExecutionPolicy, oneapi::dpl::__internal::__no_op>;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -128,11 +128,11 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
                                             _Commutative>{__reduce_op, __transform_op};
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
 
-        __usm_host_or_buffer_storage<_ExecutionPolicy, _Tp> __res_container(__exec, 1);
+        __usm_host_or_unified_storage<_ExecutionPolicy, _Tp> __scratch_container(__exec, 0);
 
         sycl::event __reduce_event = __exec.queue().submit([&, __n](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
-            auto __res_acc = __res_container.__get_acc(__cgh);
+            auto __res_ptr = __scratch_container.__get_result_ptr();
             ::std::size_t __local_mem_size = __reduce_pattern.local_mem_req(__work_group_size);
             __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__local_mem_size), __cgh);
             __cgh.parallel_for<_Name...>(
@@ -146,7 +146,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
                 });
         });
 
-        return __future(__reduce_event, __res_container);
+        return __future(__reduce_event, __scratch_container);
     }
 }; // struct __parallel_transform_reduce_small_submitter
 
@@ -183,9 +183,8 @@ struct __parallel_transform_reduce_device_kernel_submitter<_Tp, __work_group_siz
     template <typename _ExecutionPolicy, typename _Size, typename _ReduceOp, typename _TransformOp, typename _InitType,
               typename... _Ranges>
     auto
-    operator()(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Size __n,
-               _ReduceOp __reduce_op, _TransformOp __transform_op, _InitType __init, sycl::buffer<_Tp>& __temp,
-               _Ranges&&... __rngs) const
+    operator()(_ExecutionPolicy&& __exec, _Size __n, _ReduceOp __reduce_op, _TransformOp __transform_op,
+               _InitType __init, _Tp* __temp_ptr, _Ranges&&... __rngs) const
     {
         auto __transform_pattern =
             unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _TransformOp, _Tp,
@@ -198,14 +197,13 @@ struct __parallel_transform_reduce_device_kernel_submitter<_Tp, __work_group_siz
 
         return __exec.queue().submit([&, __n](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
-            sycl::accessor __temp_acc{__temp, __cgh, sycl::write_only, __dpl_sycl::__no_init{}};
             ::std::size_t __local_mem_size = __reduce_pattern.local_mem_req(__work_group_size);
             __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__local_mem_size), __cgh);
             __cgh.parallel_for<_KernelName...>(
                 sycl::nd_range<1>(sycl::range<1>(__n_groups * __work_group_size), sycl::range<1>(__work_group_size)),
                 [=](sycl::nd_item<1> __item_id) {
                     __device_reduce_kernel<_Tp>(__item_id, __n, __transform_pattern, __reduce_pattern, __temp_local,
-                                                __temp_acc, __rngs...);
+                                                __temp_ptr, __rngs...);
                 });
         });
     }
@@ -225,15 +223,18 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<
 {
     template <typename _ExecutionPolicy, typename _Size, typename _ReduceOp, typename _TransformOp, typename _InitType>
     auto
-    operator()(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, sycl::event& __reduce_event,
-               _Size __n, _ReduceOp __reduce_op, _TransformOp __transform_op, _InitType __init,
-               sycl::buffer<_Tp>& __temp) const
+    operator()(_ExecutionPolicy&& __exec, sycl::event& __reduce_event, _Size __n, _ReduceOp __reduce_op,
+               _TransformOp __transform_op, _InitType __init,
+               __usm_host_or_unified_storage<_ExecutionPolicy, _Tp> __scratch_container) const
     {
         using _NoOpFunctor = unseq_backend::walk_n<_ExecutionPolicy, oneapi::dpl::__internal::__no_op>;
         auto __transform_pattern =
             unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _NoOpFunctor, _Tp,
                                             _Commutative>{__reduce_op, _NoOpFunctor{}};
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
+
+        _Tp* __temp_ptr = __scratch_container.__get_scratch_ptr();
+        _Tp* __res_ptr = __scratch_container.__get_result_ptr();
 
         // Lower the work group size of the second kernel to the next power of 2 if __n < __work_group_size.
         auto __work_group_size2 = __work_group_size;
@@ -247,13 +248,9 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<
             }
         }
 
-        __usm_host_or_buffer_storage<_ExecutionPolicy, _Tp> __res_container(__exec, 1);
-
         __reduce_event = __exec.queue().submit([&, __n](sycl::handler& __cgh) {
             __cgh.depends_on(__reduce_event);
 
-            sycl::accessor __temp_acc{__temp, __cgh, sycl::read_only};
-            auto __res_acc = __res_container.__get_acc(__cgh);
             ::std::size_t __local_mem_size = __reduce_pattern.local_mem_req(__work_group_size2);
             __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__local_mem_size), __cgh);
 
@@ -264,11 +261,11 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<
                         __usm_host_or_buffer_storage<_ExecutionPolicy, _Tp>::__get_usm_host_or_buffer_accessor_ptr(
                             __res_acc);
                     __work_group_reduce_kernel<_Tp>(__item_id, __n, __transform_pattern, __reduce_pattern, __init,
-                                                    __temp_local, __res_ptr, __temp_acc);
+                                                    __temp_local, __res_ptr, __temp_ptr);
                 });
         });
 
-        return __future(__reduce_event, __res_container);
+        return __future(__reduce_event, __scratch_container);
     }
 }; // struct __parallel_transform_reduce_work_group_kernel_submitter
 
@@ -293,19 +290,19 @@ __parallel_transform_reduce_mid_impl(oneapi::dpl::__internal::__device_backend_t
     // number of buffer elements processed within workgroup
     constexpr _Size __size_per_work_group = __iters_per_work_item_device_kernel * __work_group_size;
     const _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
-    sycl::buffer<_Tp> __temp{sycl::range<1>(__n_groups)};
+    __usm_host_or_unified_storage<_ExecutionPolicy, _Tp> __scratch_container(__exec, __n_groups);
 
     sycl::event __reduce_event =
         __parallel_transform_reduce_device_kernel_submitter<_Tp, __work_group_size, __iters_per_work_item_device_kernel,
                                                             _Commutative, _ReduceDeviceKernel>()(
-            __backend_tag, __exec, __n, __reduce_op, __transform_op, __init, __temp,
+            __backend_tag, __exec, __n, __reduce_op, __transform_op, __init, __scratch_container.__get_scratch_ptr(),
             ::std::forward<_Ranges>(__rngs)...);
 
     __n = __n_groups; // Number of preliminary results from the device kernel.
     return __parallel_transform_reduce_work_group_kernel_submitter<
         _Tp, __work_group_size, __iters_per_work_item_work_group_kernel, _Commutative, _ReduceWorkGroupKernel>()(
         __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __reduce_event, __n, __reduce_op, __transform_op,
-        __init, __temp);
+        __init, __scratch_container);
 }
 
 // General implementation using a tree reduction
@@ -343,8 +340,10 @@ struct __parallel_transform_reduce_impl
         _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
 
         // Create temporary global buffers to store temporary values
-        sycl::buffer<_Tp> __temp(sycl::range<1>(2 * __n_groups));
-        __usm_host_or_buffer_storage<_ExecutionPolicy, _Tp> __res_container(__exec, 1);
+        __usm_host_or_unified_storage<_ExecutionPolicy, _Tp> __scratch_container(__exec, 2 * __n_groups);
+        _Tp* __temp_ptr = __scratch_container.__get_scratch_ptr();
+        _Tp* __res_ptr = __scratch_container.__get_result_ptr();
+
         // __is_first == true. Reduce over each work_group
         // __is_first == false. Reduce between work groups
         bool __is_first = true;
@@ -363,8 +362,6 @@ struct __parallel_transform_reduce_impl
 
                 // get an access to data under SYCL buffer
                 oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
-                sycl::accessor __temp_acc{__temp, __cgh, sycl::read_write};
-                auto __res_acc = __res_container.__get_acc(__cgh);
                 ::std::size_t __local_mem_size = __reduce_pattern.local_mem_req(__work_group_size);
                 __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__local_mem_size), __cgh);
 #if _ONEDPL_COMPILE_KERNEL && _ONEDPL_KERNEL_BUNDLE_PRESENT
@@ -397,7 +394,7 @@ struct __parallel_transform_reduce_impl
                         }
                         else
                         {
-                            __transform_pattern2(__item_id, __n, __offset_2, __result, __temp_acc);
+                            __transform_pattern2(__item_id, __n, __offset_2, __result, __temp_ptr);
                             __n_items = __transform_pattern2.output_size(__n, __work_group_size);
                         }
                         // 2. Reduce within work group using local memory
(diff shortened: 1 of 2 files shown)


julianmi (changes requested):
I copied my feedback from the original PR (https://github.com/AidanBeltonS/oneDPL/pull/30#issuecomment-1850441506) since it is still relevant:

I think this is a good addition. We had a similar approach in an older PR #1106 that was superseded by the unified memory design currently in mainline. Tests on this PR showed reduced host overheads especially for small input arrays.

An issue I see with this approach is compatibility for devices without USM memory support. This could be solved by adding a third case that uses the existing buffer-based approach as a fallback.

AidanBeltonS:
> I copied my feedback from the original PR ([AidanBeltonS#30 (comment)](https://github.com/AidanBeltonS/oneDPL/pull/30#issuecomment-1850441506)) since it is still relevant:
> 
> I think this is a good addition. We had a similar approach in an older PR #1106 that was superseded by the unified memory design currently in mainline. Tests on this PR showed reduced host overheads especially for small input arrays.
> 
> An issue I see with this approach is compatibility for devices without USM memory support. This could be solved by adding a third case that uses the existing buffer-based approach as a fallback.

This is an interesting problem. The current approach passes memory as pointers, this would be made more complicated with a buffer option. As you would have to query if the device supported USM allocations (a runtime check) then execute either a ptr or buffer version of the kernel. This would require two instantiations for each kernel to support both memory types at runtime.

Do you have some existing infrastructure in oneDPL to handle these kinds of cases or propose another solution?

julianmi:
> > I copied my feedback from the original PR ([AidanBeltonS#30 (comment)](https://github.com/AidanBeltonS/oneDPL/pull/30#issuecomment-1850441506)) since it is still relevant:
> > I think this is a good addition. We had a similar approach in an older PR #1106 that was superseded by the unified memory design currently in mainline. Tests on this PR showed reduced host overheads especially for small input arrays.
> > An issue I see with this approach is compatibility for devices without USM memory support. This could be solved by adding a third case that uses the existing buffer-based approach as a fallback.
> 
> This is an interesting problem. The current approach passes memory as pointers, this would be made more complicated with a buffer option. As you would have to query if the device supported USM allocations (a runtime check) then execute either a ptr or buffer version of the kernel. This would require two instantiations for each kernel to support both memory types at runtime.
> 
> Do you have some existing infrastructure in oneDPL to handle these kinds of cases or propose another solution?

Could we use `__usm_host_or_buffer_accessor` instead of the current pointer solution?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -170,7 +170,7 @@ struct __parallel_transform_reduce_device_kernel_submitter<_Tp, __work_group_siz
               typename... _Ranges>
     auto
     operator()(_ExecutionPolicy&& __exec, _Size __n, _ReduceOp __reduce_op, _TransformOp __transform_op,
-               _InitType __init, sycl::buffer<_Tp>& __temp, _Ranges&&... __rngs) const
+               _InitType __init, __usm_host_or_unified_storage<_ExecutionPolicy, _Tp> __scratch_container, _Ranges&&... __rngs) const
```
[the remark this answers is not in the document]
AidanBeltonS: Yes I think that is the best approach. I have applied this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -211,7 +212,8 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<
               oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0>
     auto
     operator()(_ExecutionPolicy&& __exec, sycl::event& __reduce_event, _Size __n, _ReduceOp __reduce_op,
-               _TransformOp __transform_op, _InitType __init, sycl::buffer<_Tp>& __temp) const
+               _TransformOp __transform_op, _InitType __init,
+               __usm_host_or_unified_storage<_ExecutionPolicy, _Tp> __scratch_container) const
```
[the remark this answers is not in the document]
AidanBeltonS: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -524,43 +539,85 @@ struct __usm_host_or_buffer_storage
...
+
   public:
-    __usm_host_or_buffer_storage(_ExecutionPolicy& __exec, ::std::size_t __n)
+    __usm_host_or_unified_storage(_ExecutionPolicy& __exec, ::std::size_t __scratch_n)
+        : __exec{__exec}, __scratch_n{__scratch_n}, __host{__use_USM_host_allocations(__exec.queue())},
+        __supports_USM{__use_USM_allocations(__exec.queue())}
     {
-        __usm = __use_USM_host_allocations(__exec.queue());
-        if (__usm)
+        if (__host)
```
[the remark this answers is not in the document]
AidanBeltonS: Iv applied the name change

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -478,38 +478,53 @@ struct __usm_host_or_buffer_accessor
...
+    _ExecutionPolicy __exec;
+    ::std::shared_ptr<_T> __scratch_buf;
+    ::std::shared_ptr<_T> __result_buf;
     ::std::shared_ptr<__sycl_buffer_t> __sycl_buf;
-    ::std::shared_ptr<_T> __usm_buf;
-    bool __usm;
 
-    // Only use USM host allocations on L0 GPUs. Other devices show significant slowdowns and will use a buffer instead.
+    ::std::size_t __scratch_n;
+    bool __host;
```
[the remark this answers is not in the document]
AidanBeltonS: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -478,38 +478,53 @@ struct __usm_host_or_buffer_accessor
...
+    bool __supports_USM;
```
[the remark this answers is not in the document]
AidanBeltonS: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -478,38 +478,53 @@ struct __usm_host_or_buffer_accessor
...
 
   public:
     // Buffer accessor
     __usm_host_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf)
-        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false)
+        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false),
+        __offset(0)
+    {
+    }
+    __usm_host_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf, size_t __acc_offset)
```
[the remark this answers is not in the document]
AidanBeltonS: Ya the rename makes more sense, it is now `__usm_or_buffer_accessor`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -478,38 +478,53 @@ struct __usm_host_or_buffer_accessor
...
     __get_pointer() const // should be cached within a kernel
     {
-        return __usm ? __ptr : &__acc[0];
+        return __usm ? __ptr + __offset : &__acc[__offset];
     }
 };
 
 template <typename _ExecutionPolicy, typename _T>
-struct __usm_host_or_buffer_storage
+struct __usm_host_or_unified_storage
```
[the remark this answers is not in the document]
AidanBeltonS: This has been re-named to your suggestion

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -577,11 +634,9 @@ class __future : private std::tuple<_Args...>
 
     template <typename _ExecutionPolicy, typename _T>
     constexpr auto
-    __wait_and_get_value(__usm_host_or_buffer_storage<_ExecutionPolicy, _T>& __buf)
+    __wait_and_get_value(__usm_host_or_unified_storage<_ExecutionPolicy, _T>& __buf)
     {
-        // Explicit wait in case of USM memory. Buffer accessors are synchronous.
-        if (__buf.__get_usm())
-            wait();
+        wait();
```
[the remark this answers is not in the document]
AidanBeltonS: Agreed, this now only waits for USM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -524,43 +541,92 @@ struct __usm_host_or_buffer_storage
...
     }
 
+    inline bool
+    __use_USM_allocations(sycl::queue __queue)
+    {
+        auto __device = __queue.get_device();
+        if (!__device.has(sycl::aspect::usm_device_allocations))
+            return false;
+        return true;
+    }
```
julianmi, line 551: We might need to guard this feature similarly to `_ONEDPL_SYCL_USM_HOST_PRESENT` in case the underlying compiler issue is present for all types of USM memory. @AidanBeltonS please double-check this implementation with older release compilers.

AidanBeltonS: I have checked and added the macro to guard the USM approach to maintain backward compatibility

julianmi:
@AidanBeltonS Please rebase this since https://github.com/oneapi-src/oneDPL/pull/1410 has been merged. I think we can get this merged after the rebase and testing.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -522,19 +539,45 @@ struct __usm_host_or_buffer_storage
...
+    __result_and_scratch_storage(_ExecutionPolicy& __exec, ::std::size_t __scratch_n)
+        : __exec{__exec}, __scratch_n{__scratch_n}, __use_USM_host{__use_USM_host_allocations(__exec.queue())},
+          __supports_USM_device{__use_USM_allocations(__exec.queue())}
     {
-        __usm = __use_USM_host_allocations(__exec.queue());
-        if (__usm)
+        if (__use_USM_host && __supports_USM_device)
+        {
+            // Separate scratch (device) and result (host) allocations on performant backends (i.e. L0)
+            __scratch_buf = std::shared_ptr<_T>(
```
julianmi, line 563:
```suggestion
            __scratch_buf = ::std::shared_ptr<_T>(
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -211,11 +210,13 @@ template <typename _Tp, ::std::uint16_t __work_group_size, ::std::uint8_t __iter
 struct __parallel_transform_reduce_work_group_kernel_submitter<
     _Tp, __work_group_size, __iters_per_work_item, _Commutative, __internal::__optional_kernel_name<_KernelName...>>
 {
-    template <typename _ExecutionPolicy, typename _Size, typename _ReduceOp, typename _TransformOp, typename _InitType>
+    template <typename _ExecutionPolicy, typename _Size, typename _ReduceOp, typename _TransformOp, typename _InitType,
+              typename _ExecutionPolicy2,
+              oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0>
```
julianmi, line 215: I believe this line is a merge issue as well and can be remove.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -522,19 +539,45 @@ struct __usm_host_or_buffer_storage
...
+                __internal::__sycl_usm_alloc<_ExecutionPolicy, _T, sycl::usm::alloc::device>{__exec}(__scratch_n),
+                __internal::__sycl_usm_free<_ExecutionPolicy, _T>{__exec});
+            __result_buf = std::shared_ptr<_T>(
```
julianmi, line 566:
```suggestion
            __result_buf = ::std::shared_ptr<_T>(
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -522,19 +539,45 @@ struct __usm_host_or_buffer_storage
...
+                __internal::__sycl_usm_alloc<_ExecutionPolicy, _T, sycl::usm::alloc::host>{__exec}(1),
+                __internal::__sycl_usm_free<_ExecutionPolicy, _T>{__exec});
+        }
+        else if (__supports_USM_device)
         {
-            __usm_buf = std::shared_ptr<_T>(
-                __internal::__sycl_usm_alloc<_ExecutionPolicy, _T, sycl::usm::alloc::host>{__exec}(__n),
+            // If we don't use host memory, malloc only a single unified device allocation
+            __scratch_buf = std::shared_ptr<_T>(
```
julianmi, line 573:
```suggestion
            __scratch_buf = ::std::shared_ptr<_T>(
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -550,30 +593,53 @@ struct __usm_host_or_buffer_storage
...
+        }
+        else
+        {
+            return __sycl_buf->get_host_access(sycl::read_only)[__scratch_n];
+        }
     }
 };
 
 //A contract for future class: <sycl::event or other event, a value, sycl::buffers..., or __usm_host_or_buffer_storage>
+
```
julianmi, line 642: Why was this blank line added?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
AidanBeltonS: Removed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -550,26 +593,48 @@ struct __usm_host_or_buffer_storage
...
-        return __usm;
+        return __supports_USM_device;
+    }
+
+    // Note: this member function assumes the result is *ready*, since the __future has already
+    // waited on the relevant event.
+    _T
+    __get_value(size_t idx = 0) const
+    {
+        if (__use_USM_host)
```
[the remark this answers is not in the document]
AidanBeltonS: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -468,46 +468,63 @@ template <typename _ContainerOrIterable>
...
+    ::std::size_t __scratch_n;
+    bool __use_USM_host;
+    bool __supports_USM_device;
+
+    // Only use USM host allocations on L0 GPUs. Other devices show significant slowdowns and will use a device allocation instead.
     inline bool
     __use_USM_host_allocations(sycl::queue __queue)
     {
-// A buffer is used by default. Supporting compilers use the unified future on top of USM host memory or a buffer.
+// USM device memory is used by default. Supporting compilers use the unified future on top of USM host or device memory.
```
julianmi, line 527: I think we can remove this outdated comment since line 523 already explains the USM host specifics.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -472,46 +472,62 @@ template <typename _ContainerOrIterable>
...
-struct __usm_host_or_buffer_accessor
+struct __usm_or_buffer_accessor
 {
   private:
     using __accessor_t = sycl::accessor<_T, 1, sycl::access::mode::read_write, __dpl_sycl::__target_device,
                                         sycl::access::placeholder::false_t>;
     __accessor_t __acc;
     _T* __ptr = nullptr;
     bool __usm = false;
+    size_t __offset;
```
SergeyKopienko, line 483: If we write here ` = 0`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -472,46 +472,62 @@ template <typename _ContainerOrIterable>
...
 
   public:
     // Buffer accessor
-    __usm_host_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf)
-        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false)
+    __usm_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf)
+        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false),
+          __offset(0)
```
SergeyKopienko, line 489: ...then `__offset(0)` isn't required...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -472,46 +472,62 @@ template <typename _ContainerOrIterable>
...
     _T* __ptr = nullptr;
     bool __usm = false;
+    size_t __offset;
 
   public:
     // Buffer accessor
-    __usm_host_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf)
-        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false)
+    __usm_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf)
+        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false),
```
SergeyKopienko, line 488: `__usm(false)` - looks like not required?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -472,46 +472,62 @@ template <typename _ContainerOrIterable>
...
+    }
+    __usm_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf, size_t __acc_offset)
+        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false),
+          __offset(__acc_offset)
     {
     }
 
     // USM pointer
-    __usm_host_or_buffer_accessor(sycl::handler& __cgh, _T* __usm_buf) : __ptr(__usm_buf), __usm(true) {}
+    __usm_or_buffer_accessor(sycl::handler& __cgh, _T* __usm_buf) : __ptr(__usm_buf), __usm(true), __offset(0) {}
```
SergeyKopienko, line 499: ...then __offset(0) isn't required...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -472,46 +472,62 @@ template <typename _ContainerOrIterable>
...
     // Buffer accessor
-    __usm_host_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf)
-        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false)
+    __usm_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf)
+        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false),
+          __offset(0)
+    {
+    }
+    __usm_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf, size_t __acc_offset)
+        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{})), __usm(false),
```
SergeyKopienko, line 493: `__usm(false)` - isn't required

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -472,46 +472,62 @@ template <typename _ContainerOrIterable>
...
 
 template <typename _ExecutionPolicy, typename _T>
-struct __usm_host_or_buffer_storage
+struct __result_and_scratch_storage
 {
   private:
     using __sycl_buffer_t = sycl::buffer<_T, 1>;
+
+    _ExecutionPolicy __exec;
+    ::std::shared_ptr<_T> __scratch_buf;
```
SergeyKopienko, line 519: Are we going really to share this fields? Or simple manage memory by this way?
I think - the second case.
But this mean `std::unique_ptr' should be enough for us.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -554,26 +596,48 @@ struct __usm_host_or_buffer_storage
...
+    __get_value(size_t idx = 0) const
+    {
+        if (__use_USM_host && __supports_USM_device)
+        {
+            return *(__result_buf.get() + idx);
+        }
+        else if (__supports_USM_device)
+        {
+            _T __tmp;
+            __exec.queue().memcpy(&__tmp, __scratch_buf.get() + __scratch_n + idx, 1 * sizeof(_T)).wait();
```
SergeyKopienko, line 634: `wait` or `wait_and_throw` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
AidanBeltonS: I think you are pointing out an interesting choice between default assignment vs initializer list for this class with your comments. I agree it should be consistent, however I would instead remove the default assignment values and only use the initializer list. As all of the values change based upon the specific constructor called, if they did not change based on the constructor then I would use default assignment.

So I am proposing removing any default assignment values and only using initializer list. Would this address your comments, or due you still prefer the default assignment?

SergeyKopienko: Probably you will have the same initializations in different constructors in this case, as `__offset(0)` now.
I would like to propose a little bit different variant: add ` = 0` here and in constructors initialize only these fields which really required the different values, not as on declaration.

AidanBeltonS: Iv followed this approach

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
AidanBeltonS: I think this should stay as a `shared_ptr`. There is a situation where a variable of this class is copied and therefore two objects have ownership of the pointer. This happens when a `__future` is constructed to return the result value.

So due to the possibility of non-reference passing and copying of the class I think this should stay as is.

Error when using `unique_ptr`.
```
/home/aidanbelton/oneDPL/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h:147:41: error: call to implicitly-deleted copy constructor of 'oneapi::dpl::__par_backend_hetero::__result_and_scratch_storage<oneapi::dpl::execution::device_policy<> &, TestUtils::Sum>'
  147 |         return __future(__reduce_event, __scratch_container);
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
AidanBeltonS: I think `wait` is more appropriate. Unless you feel that this is the right place to handle any async errors that could have occurred in the kernel. The future class uses wait and throw, this makes sense here as this is likely the last wait so you would want to collect any unconsumed errors produced. So I think it is fine as is.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: The fix is very simple:
`return __future(__reduce_event, std::move(__scratch_container));

julianmi: Wouldn't `unique_ptr` limit what the user can do with the future-like object we return? I think `shared_ptr` is the more general approach here.

SergeyKopienko: @julianmi exactly, you are right.
But let's take a look at https://en.cppreference.com/w/cpp/thread/future/operator%3D
We haven't requirement that should be able to have copyable result.

julianmi: Thanks, @SergeyKopienko. I think we can switch to `unique_ptr` then. We should do so throughout the unified USM or buffer storage though.

SergeyKopienko: See my update: https://github.com/oneapi-src/oneDPL/pull/1354#discussion_r1562531934

AidanBeltonS: I think there is a discussion to be had regarding the exact usage of this class. However, I believe it is well beyond the scope of this PR, which is fundamentally a performance optimization. The usage of the shared_ptr existed prior to this change. I propose that this sort of architectural discussion and change should be placed into an issue or some other method of communication as it is not relevant to the proposed changes.

SergeyKopienko: Additional moment - we shouldn't broke existing behavior of this class : our users already have the code where it's copyable and moveable.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -526,19 +540,45 @@ struct __usm_host_or_buffer_storage
 #endif
     }
 
+    inline bool
+    __use_USM_allocations(sycl::queue __queue)
+    {
+#if _ONEDPL_SYCL_USM_HOST_PRESENT
+        auto __device = __queue.get_device();
```
SergeyKopienko, line 547: What about
```C++
return __queue.get_device().__device.has(sycl::aspect::usm_device_allocations);
```

AidanBeltonS: Done

SergeyKopienko (approved):
LGTM
But please wait CI and receive additional approves.

julianmi (approved):

PR #1357, opened 2024-01-18
Fixing comments from original DS integration

egfefey:
*Dropping std::decay where it's not needed
*Renaming offset to index in fixed policy
etc


AnuyaWelling2801 (dismissed):
LGTM

AnuyaWelling2801 (approved):
LGTM. Removing wrapped_resource for auto_tune_policy in another PR

PR #1360, opened 2024-01-19
Add ability to avoid group reduction operations for 64-bit arithmetic data types

SergeyKopienko:
The goal of this PR is to add ability to avoid group reduction operations for 64-bit arithmetic data types.

How user can use new macro:
```C++
// When ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION is defined as non-zero,
// we avoid using known identity for 64-bit arithmetic data types
#define ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION 1
```
before all includes of oneDPL library header files.

For implementation details please see changes in the file https://github.com/oneapi-src/oneDPL/pull/1360/files#diff-2af280a1d3fd082c617a16b60d169ef6d53092d3e48b0d347af87c56b8ce5ef3

Now we generate two targets for the test `reduce_by_segment.pass.cpp` :
- `reduce_by_segment.pass` - without any changes in defines (normal test state);
- `reduce_by_segment_workaround_64bit_reduction.pass` - with `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` defined to `1`.

**Attention**: this approach doesn't working now due error in `CMakeLists.txt` file which should be fixed in the PR https://github.com/oneapi-src/oneDPL/pull/1342


include/oneapi/dpl/internal/reduce_by_segment_impl.h
```diff
@@ -605,8 +605,8 @@ reduce_by_segment_impl(Policy&& policy, InputIterator1 first1, InputIterator1 la
     auto value_output_buf = keep_value_outputs(result2, result2 + n);
 
     using has_known_identity =
-        typename __dpl_sycl::__has_known_identity<BinaryOperator,
-                                                  typename ::std::iterator_traits<InputIterator2>::value_type>::type;
```
SergeyKopienko, line 609: during this implementation new issue were found: https://github.com/oneapi-src/oneDPL/issues/1353:
This issue should be fixed before the current PR.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -275,6 +275,9 @@
 #    if !(_ONEDPL_FPGA_DEVICE)
 #        define _USE_SUB_GROUPS 1
 #        define _USE_GROUP_ALGOS 1
+#        ifndef _ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES
```
danhoeflinger, line 278: I'm not sure the appropriate place to document the presence of such a PP define.  
How will a user know this feature is available?

I suppose we could add come CMake option, which then adds this to the cxx compiler flags.  On the other hand, perhaps this is too minor to include at that level, and users can just add their own flags.

danhoeflinger, line 278: For my education, what is the motivation for why this is an important feature to provide for the user?

Do these group reduction operations provide bad performance for 64 bit arithmetic types?

danhoeflinger (commented):
Before getting into details, some high level questions:

include/oneapi/dpl/pstl/onedpl_config.h
SergeyKopienko: They simple provides incorrect results for 64-bit data types.

SergeyKopienko: We will inform our users in documentation: which known issue we have and how to avoid it.

danhoeflinger: Would it be better then to make the default `0` to avoid using use known identity then, and update the default with a compiler version minimum when it is fixed?

SergeyKopienko: I think if we enable something optional by default, we able to write macros like here.
If we not disable something optional by default, we haven't reasons ang proper ways to show in the code that we have this ability. Also I have been implemented proposed by you variant early and it's look not very good when we write all `__has_known_identity` staff. But in any case, it's not the big deal to rewrite this macro by another (disabling) way.

danhoeflinger: Can we just turn off `__has_known_identity` for these types until it is fixed (and update with working compiler version in a future oneDPL release when it is fixed), rather than providing this workaround that requires the user to set this PP definition?

include/oneapi/dpl/internal/reduce_by_segment_impl.h
mmichel11: The fix has been merged, so I think you can rebase this branch with main.

include/oneapi/dpl/pstl/onedpl_config.h
SergeyKopienko: Technically we able to do this.
But I believe it's not the best solution: we will not be able to control in tests which functions for which types give errors.
As I know at this moment, for example we able to use `double` with `std::plus`, for example.

SergeyKopienko:
If we think that this PR is applicable for oneDPL, I propose also to extend set of new tests to all operations from https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_queue_interface - Table 120. Known identities :
- `sycl::bit_and`;
- `sycl::bit_or`;
- `sycl::bit_xor`;
- `sycl::logical_and`;
- `sycl::logical_or`;
- `sycl::minimum`;
- `sycl::maximum`
for the next test cases:
- for different types (if applicable);
- for `_ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES ` : 0, 1

include/oneapi/dpl/pstl/onedpl_config.h
danhoeflinger: Ah, I did not realize it was fine-grained and not affecting all 64bit type / operator combinations.  

I still believe the better option is to have the default be correct since we have a codepath which provides correct results, and give an option to "turn on" group reduction operations for those who are interested.

danhoeflinger: After our discussion in the meeting, I'll withdraw this objection to the strategy.  I was thinking this was a compiler issues rather than a driver issue, and there is not a good way to detect this a compile time.  It makes sense to be optimistic in this case and provide a documented workaround in this case.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -63,10 +64,30 @@ using __has_known_identity =
...
 
 template <typename _BinaryOp, typename _Tp>
-using __has_known_identity = std::false_type;
+using __has_known_identity_src = std::false_type;
 
 #endif //_USE_GROUP_ALGOS && defined(SYCL_IMPLEMENTATION_INTEL)
 
+template <typename _Tp>
+using __able_to_use_known_identity_src =
+#if !defined(_ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES) ||                                       \
```
mmichel11, line 73: I believe `!defined(_ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES)` can be removed since `_ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES` will evaluate to 0 if undefined.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -63,10 +64,30 @@ using __has_known_identity =
...
+    _ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES
+    typename ::std::true_type;
```
mmichel11, line 75: I think the `typename` is not needed here.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
SergeyKopienko: @mmichel11 disagree with you.

mmichel11:
Just a general question about this PR. Is splitting `reduce_by_segment` into 16 tests the best solution here or should we split them into a smaller number of groups such as the tests that set `#define _ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES 0` and those that do not?

SergeyKopienko:
> Just a general question about this PR. Is splitting `reduce_by_segment` into 16 tests the best solution here or should we split them into a smaller number of groups such as the tests that set `#define _ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES 0` and those that do not?

From my point of view, current state of this tests is more useful then before: we are able to have control for all situations (type + predicate) separately. Also we may see in test results some additional errors that we discussed offline.

mmichel11 (dismissed):
LGTM. I think it might be good to get a second opinion before merging.

danhoeflinger (dismissed):
LGTM, how do we want to deal with the (newly detected) failing CI tests?

SergeyKopienko:
> LGTM, how do we want to deal with the (newly detected) failing CI tests?
- discussed offline.

dmitriy-sobolev:
> > LGTM, how do we want to deal with the (newly detected) failing CI tests?
> 
> * discussed offline.

What was the conclusion? Did you agree to introduce a "broken" macro?

akukanov (commented):
I would propose a modified implementation that I think is simpler and easier to comprehend:

```
#if _USE_GROUP_ALGOS && defined(SYCL_IMPLEMENTATION_INTEL)

template <typename _Tp>
inline constexpr bool __can_use_known_identity =
#if MACRO
// When MACRO is defined as non-zero, we avoid using known identity for 64-bit arithmetic data types
    !(::std::is_arithmetic_v<_Tp> && sizeof(_Tp) == sizeof(::std::uint64_t));
#else
    true;
#endif

template <typename _BinaryOp, typename _Tp>
using __has_known_identity = ::std::conditional_t< __can_use_known_identity,
# if _ONEDPL_LIBSYCL_VERSION >= 50200
    /* the existing implementation (lines 43-60) goes here as-is, except for required small syntax changes */
# endif //_ONEDPL_LIBSYCL_VERSION >= 50200
    ::std::false_type>; // known identity cannot be used

#else //_USE_GROUP_ALGOS && defined(SYCL_IMPLEMENTATION_INTEL)
// No changes to the #else branch
```

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -275,6 +275,9 @@
...
+#            define _ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES 1
+#        endif // _ONEDPL_ICPX_USE_KNOWN_IDENTITY_FOR_ARITHMETIC_64BIT_DATA_TYPES
```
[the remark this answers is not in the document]
SergeyKopienko: Ok, will do.

SergeyKopienko:
@akukanov, your proposal has been applied at this PR, please take a look.

include/oneapi/dpl/pstl/onedpl_config.h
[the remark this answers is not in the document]
SergeyKopienko: Done

danhoeflinger (approved):
Still LGTM after changes from Alexey's comments.

danhoeflinger:
> > > LGTM, how do we want to deal with the (newly detected) failing CI tests?
> > 
> > 
> > 
> > * discussed offline.
> 
> What was the conclusion? Did you agree to introduce a "broken" macro?

@dmitriy-sobolev The proposal is to leave it as red CI, as incentive to fix the issue (with the added tests for reduce_by_segment).  However, if we find that it is due to errors outside of oneDPL, then I think that should be guarded accordingly.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -50,15 +60,16 @@ using __has_known_identity =
...
 #    else  //_ONEDPL_LIBSYCL_VERSION >= 50200
     typename ::std::conjunction<
         ::std::is_arithmetic<_Tp>,
         ::std::disjunction<::std::is_same<::std::decay_t<_BinaryOp>, ::std::plus<_Tp>>,
                            ::std::is_same<::std::decay_t<_BinaryOp>, ::std::plus<void>>,
                            ::std::is_same<::std::decay_t<_BinaryOp>, __dpl_sycl::__plus<_Tp>>,
-                           ::std::is_same<::std::decay_t<_BinaryOp>, __dpl_sycl::__plus<void>>>>;
+                           ::std::is_same<::std::decay_t<_BinaryOp>, __dpl_sycl::__plus<void>>>>,
 #    endif //_ONEDPL_LIBSYCL_VERSION >= 50200
+    ::std::false_type>;
```
akukanov, line 72: Please add a comment that this is for the case of __can_use_known_identity==false

akukanov:
For the tests: I will be fine with any design that others might prefer, including the proposed one, yet I would do it differently.

As far as I can see, all but one test follow the same pattern of invoking run_test<> with a certain set of template arguments, about half setting the workaround macro and the other half expecting it to be unset or 0. So, I would only have two separate compilation units, one combining all the tests (i.e. template argument combinations) for which the workaround is needed, and the other combining the rest. The source code could even be in a single file, with different implementations of ``main()`` depending on the value of the macro, which would be set outside of the test in CMake files. Besides reducing the number of files to maintain, another benefit is that all the tested combinations are easily seen in one or at most two places, providing a view on the overall coverage. The downsides are also known: it may be harder to find out which test combination fails, the test execution time increases, etc. Yet - unless some of the downsides becomes a critical issue - I would prefer the design with less source files.

SergeyKopienko:
@akukanov, FYI: I implemented you proposal from https://github.com/oneapi-src/oneDPL/pull/1360#issuecomment-1908833335.

test/CMakeLists.txt
```diff
@@ -177,9 +177,14 @@ macro(onedpl_add_test test_source_file switch_off_checked_iterators)
     string(REPLACE "\.cpp" "" _test_name ${_test_name})
 
     set(coal_tests "reduce.pass" "transform_reduce.pass" "count.pass" "sycl_iterator_reduce.pass" "minmax_element.pass")
+    set(workaround_for_igpu_64bit_reduction_tests "reduce_by_segment.pass")
     if (_test_name IN_LIST coal_tests)
         onedpl_construct_exec(${test_source_file} ${_test_name} ${switch_off_checked_iterators} "-D_ONEDPL_DETECT_SPIRV_COMPILATION=1")
         onedpl_construct_exec(${test_source_file} ${_test_name}.coal ${switch_off_checked_iterators} "-D_ONEDPL_DETECT_SPIRV_COMPILATION=0")
+    elseif (_test_name IN_LIST workaround_for_igpu_64bit_reduction_tests)
+        onedpl_construct_exec(${test_source_file} ${_test_name} ${switch_off_checked_iterators} "")
+        string(REPLACE "\.pass" "_avoid_64bit_reduction\.pass" _test_name ${_test_name})
```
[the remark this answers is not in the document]

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
```diff
@@ -330,31 +330,32 @@ main()
...
-#    else
-        test_algo_four_sequences<ValueType, test_reduce_by_segment<BinaryPredicate, BinaryOperation>>();
-#    endif // TEST_DPCPP_BACKEND_PRESENT
-#endif     // !_PSTL_ICC_TEST_SIMD_UDS_BROKEN && !_PSTL_ICPX_TEST_RED_BY_SEG_OPTIMIZER_CRASH
-    }
+    run_test<::std::uint64_t,       UserBinaryPredicate<::std::uint64_t>,       MaxFunctor<::std::uint64_t>>();
+    run_test<::std::complex<float>, UserBinaryPredicate<::std::complex<float>>, MaxFunctor<::std::complex<float>>>();
+
+    run_test<::std::equal_to, ::std::plus>();
+    run_test<::std::equal_to, ::std::multiplies>();
```
dmitriy-sobolev, line 358: Could you double-check that there is not overflow?

SergeyKopienko: Yes: error exist regardless of `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` state, in both variants.

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
```diff
@@ -330,31 +330,32 @@ main()
...
 
     return TestUtils::done();
-}
+}
```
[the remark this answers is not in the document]
SergeyKopienko: Done

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
dmitriy-sobolev: I am not totally sure, but my rude estimation shows that there is overflow with multiplication:
```python
from functools import reduce

# check if that sequence may overflow
# T keys[n] = { 1, 2, 3, 4, 1, 1, 2, 2, 3, 3, 4, 4, 1, 1, 1, ...};
# T vals[n] = { n random numbers between 0 and 4 };

n = 100000
val_number = 4

# number of occurrences of one unique value
single_val_all = n/val_number
# calculate a longest sequence length
single_val_max_seq_n = 1
while single_val_all > 0:
    single_val_all -= single_val_max_seq_n
    single_val_max_seq_n += 1

average_value = reduce(lambda x, y: x + y, range(1, val_number))/val_number
multiply_reduction = average_value ** single_val_max_seq_n
plus_reduction = average_value * single_val_max_seq_n

print(f"possible max single sequence lenght: {single_val_max_seq_n} when there are {val_number} unique keys and n is {n}")
print(f"possible max single sequence reduction result (multiply): {multiply_reduction}")
print(f"possible max single sequence reduction result (plus): {plus_reduction}")
```
output:
```
possible max single sequence lenght: 225 when there are 4 unique keys and n is 100000
possible max single sequence reduction result (multiply): 4.1738158843886507e+39
possible max single sequence reduction result (plus): 337.5
```

dmitriy-sobolev: I guess it would be safer to limit `n` for multiply predicate, unless my estimation above is entirely wrong.

dmitriy-sobolev: Sorry, I have noticed that the data is generated in the range of `[0, 4]` (instead of `[1,4]`). Perhaps, there is no overflow error with `multiplies` as there are high chances to get `0`.

However, I would still recommend to adapt the data generation for `multiplies` predicate:  start with `1` and limit `n`. It can be done in a separate PR in order not to delay the workaround, while removing (or commenting-out) a case with `multiplies` in this PR.

SergeyKopienko: I think that this value generation
```C++
host_vals[i] = rand() % 5;
```
not corelated with initial problem for which this PR was created. And for `std::multiplies` with `int` too.

dmitriy-sobolev: My point was that the data generation should be adapted for reduction with `multiplies` predicate, which is introduced by this PR.

The current generation presumably results in overflows.
It may result in issues or not. I do not know if it is UB either. However, it look like a case which we do not want to test. That's a caveat that the PR may add another issue into the testing.

Anyway, it is a minor point. It affects only tests and it is an estimation. I suggest temporary removing `multiplies` case, until it is investigated better, or at least to add a note that it may result in UB for further investigation/fixing.

SergeyKopienko: So the root issue was in CMakeLists.txt. @mmichel11 will fix it.
I propose to save data generation in this test as is.

dmitriy-sobolev: I guess we are talking about different problems. The PR adds a workaround and more checks. That's great, thanks. 
The tests now also pass (after a correction in CMakeLists.txt). This is good too.

But it also introduces a case which may potentially backfire somewhen.
This note may alleviate difficulties with searching in case of something goes wrong.

```suggestion
    // TODO: investigate possible overflow
    run_test<::std::equal_to, ::std::multiplies>();
```

CMakeLists.txt
```diff
@@ -29,6 +29,7 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} ${CMAKE_CURRENT_SOURCE_DIR}/cmake)
 option(ONEDPL_FPGA_STATIC_REPORT "Enable the static report generation for the FPGA device" OFF)
 option(ONEDPL_USE_AOT_COMPILATION "Enable the ahead of time compilation" OFF)
 option(ONEDPL_ENABLE_SIMD "Enable SIMD vectorization by passing an OpenMP SIMD flag to the compiler if supported" ON)
+option(ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION "Avoid using known identity for 64-bit arithmetic data types")
```
SergeyKopienko, line 32: By default, the macro `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` is undefined.
@dmitriy-sobolev, iIs it OK to declare this option without initial value?

cmake/README.md
```diff
@@ -18,6 +18,7 @@ The following variables are provided for oneDPL configuration:
 | ONEDPL_AOT_ARCH              | STRING | Architecture options for ahead-of-time compilation, supported values can be found [here](https://software.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/compilation/ahead-of-time-compilation.html)                                                                                            | "*" for GPU device and "avx" for CPU device |
 | ONEDPL_TEST_EXPLICIT_KERNEL_NAMES   | STRING | Control kernel naming. Affects only oneDPL test targets. Supported values: AUTO, ALWAYS. AUTO: rely on the compiler if "Unnamed SYCL lambda kernels" feature is on, otherwise provide kernel names explicitly; ALWAYS: provide kernel names explicitly | AUTO          |
 | ONEDPL_TEST_WIN_ICX_FIXES     | BOOL   | Affects only oneDPL test targets.  Enable icx, icx-cl workarounds to fix issues in CMake for Windows.                      | ON            |
+| ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION | BOOL | Avoid using known identity for 64-bit arithmetic data types                       | UNDEFINED     |
```
SergeyKopienko, line 21: @dmitriy-sobolev which proper type for this option `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` we should declare?

documentation/release_notes.rst
```diff
@@ -57,8 +57,10 @@ New in This Release
   with ``unseq`` or ``par_unseq`` policy when compiled by Intel® oneAPI DPC++/C++ Compiler
   with ``-fiopenmp``, ``-fiopenmp-simd``, ``-qopenmp``, ``-qopenmp-simd`` options on Linux.
   To avoid the issue, pass ``-fopenmp`` or ``-fopenmp-simd`` option instead.
-- Incorrect results may be produced by ``reduce`` and ``transform_reduce`` with 64-bit types and ``std::multiplies``,
-  ``sycl::multiplies`` operations when compiled by Intel® C++ Compiler 2021.3 and newer and executed on GPU devices. 
+- Incorrect results may be produced by ``reduce``, ``reduce_by_segment`` and ``transform_reduce`` with 64-bit types
```
SergeyKopienko, line 60: @dmitriy-sobolev, @dcbenito could you please check this text?

cmake/README.md
dmitriy-sobolev: "Known identity" looks like an implementation detail, it may confuse a user, because the consequences are not apparent.
What do you think about adding a line into the description which would highlight/imply a consequence of using this workaround and refer known issues section?

I suggest something like: "Fall back to more general, usually less optimized, implementations of reduction-based algorithms for 64-bit types. Refer to 'Known issues' section for more details".

cmake/README.md
```diff
@@ -18,6 +18,7 @@ The following variables are provided for oneDPL configuration:
 | ONEDPL_AOT_ARCH              | STRING | Architecture options for ahead-of-time compilation, supported values can be found [here](https://software.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/compilation/ahead-of-time-compilation.html)                                                                                            | "*" for GPU device and "avx" for CPU device |
 | ONEDPL_TEST_EXPLICIT_KERNEL_NAMES   | STRING | Control kernel naming. Affects only oneDPL test targets. Supported values: AUTO, ALWAYS. AUTO: rely on the compiler if "Unnamed SYCL lambda kernels" feature is on, otherwise provide kernel names explicitly; ALWAYS: provide kernel names explicitly | AUTO          |
 | ONEDPL_TEST_WIN_ICX_FIXES     | BOOL   | Affects only oneDPL test targets.  Enable icx, icx-cl workarounds to fix issues in CMake for Windows.                      | ON            |
+| ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION | BOOL | Avoid using known identity for 64-bit arithmetic data types                       |               |
```
akukanov, line 21: The description gives a developer perspective to the issue, which likely means little for users. It should be meaningful for those who might experience the issue.

documentation/release_notes.rst
```diff
@@ -57,8 +57,10 @@ New in This Release
...
+  and ``std::multiplies``, ``sycl::multiplies`` operations when compiled by Intel® C++ Compiler 2021.3 and newer
```
akukanov, line 61: If I remember correctly, you mentioned that the issue is not limited to multiplication but can be seen with other reduction operations as well. Did I misunderstand?

documentation/release_notes.rst
```diff
@@ -57,8 +57,10 @@ New in This Release
...
+  and executed on GPU devices. The macro ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION may be defined to 1
+  to void using known identity for 64-bit arithmetic data types in these cases.
```
akukanov, line 63: Same here - "avoid using known identity" is a piece of internal information which does not make much sense for readers. We just need to say that defining this macro solves the problem.

documentation/release_notes.rst
SergeyKopienko: No, you are correct. Fixed.

documentation/release_notes.rst
SergeyKopienko: Fixed

cmake/README.md
SergeyKopienko: Fixed

cmake/README.md
```diff
@@ -18,6 +18,7 @@ The following variables are provided for oneDPL configuration:
...
 | ONEDPL_TEST_EXPLICIT_KERNEL_NAMES   | STRING | Control kernel naming. Affects only oneDPL test targets. Supported values: AUTO, ALWAYS. AUTO: rely on the compiler if "Unnamed SYCL lambda kernels" feature is on, otherwise provide kernel names explicitly; ALWAYS: provide kernel names explicitly | AUTO          |
 | ONEDPL_TEST_WIN_ICX_FIXES     | BOOL   | Affects only oneDPL test targets.  Enable icx, icx-cl workarounds to fix issues in CMake for Windows.                      | ON            |
+| ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION | BOOL | Pass `-DONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION=1` for workaround incorrect results which may be produced by ``reduce``, ``reduce_by_segment`` and ``transform_reduce`` with 64-bit types when compiled by Intel� C++ Compiler 2021.3 and newer and executed on GPU devices. |               |
```
akukanov, line 21:
```suggestion
| ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION | BOOL | Use as a workaround for incorrect results which may be produced by reduction algorithms with 64-bit data types compiled by oneAPI DPC++ Compiler and executed on GPU devices. |               |
```

documentation/release_notes.rst
```diff
@@ -57,8 +57,9 @@ New in This Release
   with ``unseq`` or ``par_unseq`` policy when compiled by Intel® oneAPI DPC++/C++ Compiler
   with ``-fiopenmp``, ``-fiopenmp-simd``, ``-qopenmp``, ``-qopenmp-simd`` options on Linux.
   To avoid the issue, pass ``-fopenmp`` or ``-fopenmp-simd`` option instead.
-- Incorrect results may be produced by ``reduce`` and ``transform_reduce`` with 64-bit types and ``std::multiplies``,
-  ``sycl::multiplies`` operations when compiled by Intel® C++ Compiler 2021.3 and newer and executed on GPU devices. 
+- Incorrect results may be produced by ``reduce``, ``reduce_by_segment`` and ``transform_reduce`` with 64-bit types
+  when compiled by Intel® C++ Compiler 2021.3 and newer and executed on GPU devices.
+  The macro ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION should define to 1 for workaround of this issue.
```
akukanov, line 62:
```suggestion
- Incorrect results may be produced by ``reduce``, ``reduce_by_segment`` and ``transform_reduce``
  with 64-bit data types when compiled by Intel oneAPI DPC++/C++ Compiler 2021.3 and newer
  and executed on GPU devices.
  For a workaround, define the ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION macro to 1
  before including oneDPL header files.
```

akukanov (commented):
I propose some wording improvements.

SergeyKopienko:
@akukanov your suggestions has been applied.

akukanov:
I am fine with the implementation and the docs, but did not check the makefiles and the test.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
SergeyKopienko: Fixed

dmitriy-sobolev (approved):
The PR looks good enough to me. 

I think test generation can be implemented in a less convoluted way, but I do not have any solution right now. If I have enough time, I will investigate it separately. 
Debugging instructions can also be cleaned up (but it might still be useful, so I do not insist).

dmitriy-sobolev (approved):
Reapproving after clang-format. Please, wait until the whole CI passes before merging.

SergeyKopienko:
@mmichel11 after merge this PR into main branch, could you please create new one PR and add some broken macros in test `reduce_by_segment.pass.cpp` for case when `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` isn't defined?

documentation/release_notes.rst
```diff
@@ -57,8 +57,11 @@ New in This Release
   with ``unseq`` or ``par_unseq`` policy when compiled by Intel® oneAPI DPC++/C++ Compiler
   with ``-fiopenmp``, ``-fiopenmp-simd``, ``-qopenmp``, ``-qopenmp-simd`` options on Linux.
   To avoid the issue, pass ``-fopenmp`` or ``-fopenmp-simd`` option instead.
-- Incorrect results may be produced by ``reduce`` and ``transform_reduce`` with 64-bit types and ``std::multiplies``,
-  ``sycl::multiplies`` operations when compiled by Intel® C++ Compiler 2021.3 and newer and executed on GPU devices. 
+- Incorrect results may be produced by ``reduce``, ``reduce_by_segment`` and ``transform_reduce``
+  with 64-bit data types when compiled by Intel oneAPI DPC++/C++ Compiler 2021.3 and newer
+  and executed on GPU devices.
+  For a workaround, define the ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION macro to 1
+  before including oneDPL header files.
```
akukanov, line 64: Oops, I missed that this change is for the description in the "old" 2022.3 release, where the workaround is not available. Let's remove the last sentence; we will need to describe it in 2022.5 release notes. Sorry for confusion.

akukanov (changes requested):

SergeyKopienko:
@dmitriy-sobolev after merge this PR into main branch, could you please create broken macro  for these two tests to avoid failed test results like in https://github.com/oneapi-src/oneDPL/actions/runs/7670071255/job/20905400624?pr=1360 ?

documentation/release_notes.rst
SergeyKopienko: Fixed.

akukanov (approved):

akukanov (approved):
Restoring the approval after rebase - everything seems to be as before.

cmake/README.md
```diff
@@ -18,6 +18,7 @@ The following variables are provided for oneDPL configuration:
 | ONEDPL_AOT_ARCH              | STRING | Architecture options for ahead-of-time compilation, supported values can be found [here](https://software.intel.com/content/www/us/en/develop/documentation/oneapi-dpcpp-cpp-compiler-dev-guide-and-reference/top/compilation/ahead-of-time-compilation.html)                                                                                            | "*" for GPU device and "avx" for CPU device |
 | ONEDPL_TEST_EXPLICIT_KERNEL_NAMES   | STRING | Control kernel naming. Affects only oneDPL test targets. Supported values: AUTO, ALWAYS. AUTO: rely on the compiler if "Unnamed SYCL lambda kernels" feature is on, otherwise provide kernel names explicitly; ALWAYS: provide kernel names explicitly | AUTO          |
 | ONEDPL_TEST_WIN_ICX_FIXES     | BOOL   | Affects only oneDPL test targets.  Enable icx, icx-cl workarounds to fix issues in CMake for Windows.                      | ON            |
+| ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION | BOOL | Use as a workaround for incorrect results which may be produced by reduction algorithms with 64-bit data types compiled by oneAPI DPC++ Compiler and executed on GPU devices. |               |
```
dcbenito, line 21: My suggestion (please cite the compiler correctly, if you are referencing the open-source compiler, you can drop the Intel(R)) :

| ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION | BOOL | Use as a workaround for incorrect results, which may be produced by reduction algorithms with 64-bit data types compiled by the Intel® oneAPI DPC++/C++ Compiler and executed on GPU devices.

documentation/release_notes.rst
```diff
@@ -57,8 +57,9 @@ New in This Release
   with ``unseq`` or ``par_unseq`` policy when compiled by Intel® oneAPI DPC++/C++ Compiler
   with ``-fiopenmp``, ``-fiopenmp-simd``, ``-qopenmp``, ``-qopenmp-simd`` options on Linux.
   To avoid the issue, pass ``-fopenmp`` or ``-fopenmp-simd`` option instead.
-- Incorrect results may be produced by ``reduce`` and ``transform_reduce`` with 64-bit types and ``std::multiplies``,
-  ``sycl::multiplies`` operations when compiled by Intel® C++ Compiler 2021.3 and newer and executed on GPU devices. 
+- Incorrect results may be produced by ``reduce``, ``reduce_by_segment`` and ``transform_reduce``
```
dcbenito, line 60: My suggestion (adding commas, version, and correct Intel(R) citation):

- Incorrect results may be produced by ``reduce``, ``reduce_by_segment``, and ``transform_reduce`` with 64-bit data types when compiled with the Intel® oneAPI DPC++/C++ Compiler versions 2021.3 and newer and executed on GPU devices.

dcbenito (approved):

timmiesmith:
@dcbenito would you please confirm I've properly updated the files with your feedback and reapprove if I've done so?

cmake/README.md
akukanov: It's either compiler, and we do not want to name both here. Since Intel's compiler is derived from the open source one, I prefer to mention the upstream one. The name is taken from https://github.com/intel/llvm/blob/sycl/README.md.

documentation/release_notes.rst
akukanov: We do not have to put the (R) sign each time on the page, only the first time is enough, is not it? And this very trademark is already mentioned above on the page.

dcbenito: We should have the (R) at the first instance on each individual page. What I would do:

- Add the (R) to the first instance on the page (first bullet)
- Remove the (R) from the sixth and seventh bullet and the new entry

cmake/README.md
dcbenito: That is fine by me, but please cite it correctly: oneAPI DPC++/C++ Compiler

dcbenito (dismissed):
@timmiesmith changes LGTM.

CMakeLists.txt
SergeyKopienko: From my point of view, we are unable to do this because by default we we shouldn't define `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` macro to ay state.

cmake/README.md
SergeyKopienko: This proposal has been applied.
@akukanov Do you like this text?

CMakeLists.txt
dmitriy-sobolev: You are free to leave it as is. Note, that  `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` CMake option will be `OFF` whether its is explicitly set or omitted.

I believe you do not confuse `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` as a macro and a CMake option: these are different entities.

cmake/README.md
dmitriy-sobolev: Sorry for not answering the original question. I thought about a default value, rather than a type.

I am fine with `BOOL` since there is a precendent with `ONEDPL_USE_UNNAMED_LAMBDA`, which can also have "true", "false" and "non-defined" states. I am a bit confused since it can accept any value (it is not limited to "bool"), but I am not aware of a better alternative.

cmake/README.md
```diff
@@ -18,6 +18,7 @@ The following variables are provided for oneDPL configuration:
...
 | ONEDPL_TEST_EXPLICIT_KERNEL_NAMES   | STRING | Control kernel naming. Affects only oneDPL test targets. Supported values: AUTO, ALWAYS. AUTO: rely on the compiler if "Unnamed SYCL lambda kernels" feature is on, otherwise provide kernel names explicitly; ALWAYS: provide kernel names explicitly | AUTO          |
 | ONEDPL_TEST_WIN_ICX_FIXES     | BOOL   | Affects only oneDPL test targets.  Enable icx, icx-cl workarounds to fix issues in CMake for Windows.                      | ON            |
+| ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION | BOOL | Use as a workaround for incorrect results, which may be produced by reduction algorithms with 64-bit data types compiled by the Intel&reg oneAPI DPC++/C++ Compiler and executed on GPU devices. |               |
```
dmitriy-sobolev, line 21: "Intel&reg" does not render well at "View File" tab. It will likely remain so if merged as is. @SergeyKopienko, could you fix it?

dmitriy-sobolev:
```suggestion
| ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION | BOOL | Use as a workaround for incorrect results, which may be produced by reduction algorithms with 64-bit data types compiled by the Intel&reg; oneAPI DPC++/C++ Compiler and executed on GPU devices. |               |
```

akukanov (approved):

dcbenito (commented):
LGTM

PR #1365, opened 2024-01-22
Pass value and remove barrier between transform_reduce and reduce_over_group

AidanBeltonS:
This PR modifies how values are passed between `transform_reduce` and `reduce_over_group`. As opposed to moving data into local memory, syncing the work_group, then unloading we just move the value into a register and return it.

This allows us to save local memory loads and stores, and a barrier. We have seen improved performance on both Nvidia and Intel GPUs from this PR.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -363,7 +363,11 @@ struct __parallel_transform_reduce_impl
                         // 1. Initialization (transform part). Fill local memory
                         _Size __n_items;
 
-                        union __storage {_Tp __v; __storage(){} } __result;
+                        union __storage
+                        {
+                            _Tp __v;
+                            __storage() {}
+                        } __result;
                         if (__is_first)
                         {
                             __result.__v = __transform_pattern1(__item_id, __n, /*global_offset*/ (_Size)0, __rngs...);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -366,12 +366,12 @@ struct __parallel_transform_reduce_impl
                         _Tp __result = *std::launder(reinterpret_cast<_Tp*>(alloca(sizeof(_Tp))));
                         if (__is_first)
                         {
-                            __result = __transform_pattern1(__item_id, __n, /*global_offset*/ (_Size)0,  __rngs...);
+                            __result = __transform_pattern1(__item_id, __n, /*global_offset*/ (_Size)0, __rngs...);
                             __n_items = __transform_pattern1.output_size(__n, __work_group_size);
                         }
                         else
                         {
-                            __result = __transform_pattern2(__item_id, __n, __offset_2,  __temp_acc);
+                            __result = __transform_pattern2(__item_id, __n, __offset_2, __temp_acc);
                             __n_items = __transform_pattern2.output_size(__n, __work_group_size);
                         }
                         // 2. Reduce within work group using local memory
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -56,11 +56,10 @@ __work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, _Transfor
     auto __local_idx = __item_id.get_local_id(0);
     auto __group_size = __item_id.get_local_range().size();
     // 1. Initialization (transform part). Fill local memory
-    __transform_pattern(__item_id, __n, /*global_offset*/ (_Size)0, __local_mem, __acc...);
-    __dpl_sycl::__group_barrier(__item_id);
+    _Tp __result = __transform_pattern(__item_id, __n, /*global_offset*/ (_Size)0, __acc...);
     const _Size __n_items = __transform_pattern.output_size(__n, __group_size);
     // 2. Reduce within work group using local memory
-    _Tp __result = __reduce_pattern(__item_id, __n_items, __local_mem);
+    __result = __reduce_pattern(__item_id, __n_items, __result, __local_mem);
     if (__local_idx == 0)
     {
         __reduce_pattern.apply_init(__init, __result);
@@ -80,11 +79,10 @@ __device_reduce_kernel(const _NDItemId __item_id, const _Size __n, _TransformPat
     auto __group_idx = __item_id.get_group(0);
     auto __group_size = __item_id.get_local_range().size();
     // 1. Initialization (transform part). Fill local memory
-    __transform_pattern(__item_id, __n, /*global_offset*/ (_Size)0, __local_mem, __acc...);
-    __dpl_sycl::__group_barrier(__item_id);
+    _Tp __result = __transform_pattern(__item_id, __n, /*global_offset*/ (_Size)0, __acc...);
     const _Size __n_items = __transform_pattern.output_size(__n, __group_size);
     // 2. Reduce within work group using local memory
-    _Tp __result = __reduce_pattern(__item_id, __n_items, __local_mem);
+    __result = __reduce_pattern(__item_id, __n_items, __result, __local_mem);
     if (__local_idx == 0)
         __temp_acc[__group_idx] = __result;
 }
@@ -113,7 +111,7 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
                _InitType __init, _Ranges&&... __rngs) const
     {
         auto __transform_pattern =
-            unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _TransformOp,
+            unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _TransformOp, _Tp,
                                             _Commutative>{__reduce_op, __transform_op};
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
 
@@ -122,7 +120,8 @@ struct __parallel_transform_reduce_small_submitter<_Tp, __work_group_size, __ite
         sycl::event __reduce_event = __exec.queue().submit([&, __n](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...); // get an access to data under SYCL buffer
             auto __res_acc = __res_container.__get_acc(__cgh);
-            __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__work_group_size), __cgh);
+            ::std::size_t __local_mem_size = __reduce_pattern.local_mem_req(__work_group_size);
+            __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__local_mem_size), __cgh);
             __cgh.parallel_for<_Name...>(
                 sycl::nd_range<1>(sycl::range<1>(__work_group_size), sycl::range<1>(__work_group_size)),
                 [=](sycl::nd_item<1> __item_id) {
@@ -173,7 +172,7 @@ struct __parallel_transform_reduce_device_kernel_submitter<_Tp, __work_group_siz
                _InitType __init, sycl::buffer<_Tp>& __temp, _Ranges&&... __rngs) const
     {
         auto __transform_pattern =
-            unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _TransformOp,
+            unseq_backend::transform_reduce<_ExecutionPolicy, __iters_per_work_item, _ReduceOp, _TransformOp, _Tp,
                                             _Commutative>{__reduce_op, __transform_op};
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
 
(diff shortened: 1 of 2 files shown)


julianmi:
@AidanBeltonS Please rebase this PR off main to resolve the conflicts.

AidanBeltonS:
> @AidanBeltonS Please rebase this PR off main to resolve the conflicts.

I have rebased the PR

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -360,19 +362,20 @@ struct __parallel_transform_reduce_impl
                         auto __group_idx = __item_id.get_group(0);
                         // 1. Initialization (transform part). Fill local memory
                         _Size __n_items;
+
+                        _Tp __result = *std::launder(reinterpret_cast<_Tp*>(alloca(sizeof(_Tp))));
```
julianmi, line 366: Can we come up with a simpler approach and one that doesn't potentially lead to a stack overflow?

julianmi, line 366:
```suggestion
                        _Tp __result = *::std::launder(reinterpret_cast<_Tp*>(alloca(sizeof(_Tp))));
```

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -222,27 +221,27 @@ struct transform_reduce
...
             const _Size __items_to_process = __adjusted_n - __adjusted_global_id;
             // Keep these statements in the same scope to allow for better memory alignment
-            typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
+            _Tp __res = __unary_op(__adjusted_global_id, __acc...);
             for (_Size __i = 1; __i < __items_to_process; ++__i)
                 __res = __binary_op(__res, __unary_op(__adjusted_global_id + __i, __acc...));
-            __local_mem[__local_idx] = __res;
+            return __res;
         }
+        return *std::launder(reinterpret_cast<_Tp*>(alloca(sizeof(_Tp))));
```
julianmi, line 239:
```suggestion
        return *::std::launder(reinterpret_cast<_Tp*>(alloca(sizeof(_Tp))));
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
AidanBeltonS: This is something that is on the more risky side of this PR. I am using this approach to get around having no guarantees of `_Tp` having a default constructor. Essentially we are just allocating garbage memory on the stack. This should be safe so long as `_Tp` is not too large.  I believe we are already in a similar position as if `_Tp` is too large we cannot allocate enough local memory to accommodate the data type.

So far as I am aware this is the most suitable approach that does not use local memory, to get around this issue. Other approaches would require a default constructor.

The other alternative that I can think of might be do allocate an array of `char`s of equal size to `_Tp` then reinterpret cast this as the return value. I believe these two approaches would be essentially equivalent, however if you have a preference for the latter I can explore that option.

adamfidel: Could you update the PR to include your explanation as a comment? I think it would be helpful to understand the usage of `alloca` here.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -363,6 +358,15 @@ struct reduce_over_group
     {
         __init_processing<_Tp>{}(__init, __result, __bin_op1);
     }
+
+    inline ::std::size_t
+    local_mem_req(const ::std::uint16_t& __work_group_size) const
+    {
+        if (__has_known_identity<_BinaryOperation1, _Tp>{})
```
adamfidel, line 365: Not super important, but this can be an `if constexpr`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
adamfidel: Other algorithms in oneDPL have used unions to support types that are not default constructible. For example, radix sort: https://github.com/oneapi-src/oneDPL/blob/516984dbc8808f1a369816005af642ddfc6fb594/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h#L167

Could something like that be used here?

AidanBeltonS: Thanks for the suggestion. I have moved the code over to use them instead of alloca

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
AidanBeltonS: done

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -218,7 +218,11 @@ struct transform_reduce
         const _Size __adjusted_global_id = __global_offset + __iters_per_work_item * __global_idx;
         const _Size __adjusted_n = __global_offset + __n;
         // Sequential load and reduce from global memory
-        union __storage {_Tp __v; __storage(){} } __res;
+        union __storage
+        {
+            _Tp __v;
+            __storage() {}
```
SergeyKopienko, line 224: `__storage() = default;`   ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -363,7 +363,11 @@ struct __parallel_transform_reduce_impl
                         // 1. Initialization (transform part). Fill local memory
                         _Size __n_items;
 
-                        union __storage {_Tp __v; __storage(){} } __result;
+                        union __storage
+                        {
+                            _Tp __v;
+                            __storage() {}
```
SergeyKopienko, line 369: `__storage() = default;`   ?

SergeyKopienko: And is this constructor really needed?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
SergeyKopienko: And is this constructor really needed?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -250,7 +254,11 @@ struct transform_reduce
         const _Size __adjusted_n = __global_offset + __n;
 
         // Coalesced load and reduce from global memory
-        union __storage {_Tp __v; __storage(){} } __res;
+        union __storage
+        {
+            _Tp __v;
+            __storage() {}
```
SergeyKopienko, line 260: the same

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
AidanBeltonS: This approach is consistent with the existing code base. https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h#L169

If you wish to change this, I would recommend refactoring it in a separate PR

julianmi (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -360,29 +362,34 @@ struct __parallel_transform_reduce_impl
...
+
+                        union __storage
+                        {
+                            _Tp __v;
+                            __storage() {}
+                        } __result;
                         if (__is_first)
                         {
-                            __transform_pattern1(__item_id, __n, /*global_offset*/ (_Size)0, __temp_local, __rngs...);
+                            __result.__v = __transform_pattern1(__item_id, __n, /*global_offset*/ (_Size)0, __rngs...);
```
MikeDvorskiy, line 373: A `union `doesn't  call any constructor for type `_Tp`. 
So, `__v` is just a storage, not C++ created object. To be correct with C++ perspective, you should call a move/copy constructor (via a placement new call), like this:
`new (&__in_val.__v) _ValueT(std::move(...));`
https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h#L603

After usage of `__v`, you should call a destructor explicitly,  like this: https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h#L619

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -360,29 +362,34 @@ struct __parallel_transform_reduce_impl
...
                             __n_items = __transform_pattern1.output_size(__n, __work_group_size);
                         }
                         else
                         {
-                            __transform_pattern2(__item_id, __n, __offset_2, __temp_local, __temp_acc);
+                            __result.__v = __transform_pattern2(__item_id, __n, __offset_2, __temp_acc);
```
MikeDvorskiy, line 378: Please, see my comment above

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -360,29 +362,34 @@ struct __parallel_transform_reduce_impl
...
                             __n_items = __transform_pattern2.output_size(__n, __work_group_size);
                         }
-                        __dpl_sycl::__group_barrier(__item_id);
                         // 2. Reduce within work group using local memory
-                        _Tp __result = __reduce_pattern(__item_id, __n_items, __temp_local);
+                        __result.__v = __reduce_pattern(__item_id, __n_items, __result.__v, __temp_local);
```
MikeDvorskiy, line 382: Please, see my comment above

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -203,46 +203,48 @@ struct __init_processing
...
+        } __res;
         if (__adjusted_global_id + __iters_per_work_item < __adjusted_n)
         {
             // Keep these statements in the same scope to allow for better memory alignment
-            typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
+            __res.__v = __unary_op(__adjusted_global_id, __acc...);
```
MikeDvorskiy, line 229: Please, see my comment above (about explicit calls of a constructor and destructor)

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -203,46 +203,48 @@ struct __init_processing
...
-                __res = __binary_op(__res, __unary_op(__adjusted_global_id + __i, __acc...));
-            __local_mem[__local_idx] = __res;
+                __res.__v = __binary_op(__res.__v, __unary_op(__adjusted_global_id + __i, __acc...));
         }
         else if (__adjusted_global_id < __adjusted_n)
         {
             const _Size __items_to_process = __adjusted_n - __adjusted_global_id;
             // Keep these statements in the same scope to allow for better memory alignment
-            typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
+            __res.__v = __unary_op(__adjusted_global_id, __acc...);
```
MikeDvorskiy, line 238: Please, see my comment above (about explicit calls of a constructor and destructor)

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -252,38 +254,41 @@ struct transform_reduce
...
         // Coalesced load and reduce from global memory
+        union __storage
+        {
+            _Tp __v;
+            __storage() {}
+        } __res;
         if (__adjusted_global_id + __stride * __iters_per_work_item < __adjusted_n)
         {
-            typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
+            __res.__v = __unary_op(__adjusted_global_id, __acc...);
```
MikeDvorskiy, line 264: Please, see my comment above (about explicit calls of a constructor and destructor)

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -252,38 +254,41 @@ struct transform_reduce
...
-                __res = __binary_op(__res, __unary_op(__adjusted_global_id + __stride * __i, __acc...));
-            __local_mem[__local_idx] = __res;
+                __res.__v = __binary_op(__res.__v, __unary_op(__adjusted_global_id + __stride * __i, __acc...));
         }
         else if (__adjusted_global_id < __adjusted_n)
         {
             const _Size __items_to_process =
                 std::max(((__adjusted_n - __adjusted_global_id - 1) / __stride) + 1, static_cast<_Size>(0));
-            typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
+            __res.__v = __unary_op(__adjusted_global_id, __acc...);
```
MikeDvorskiy, line 273: Please, see my comment above (about explicit calls of a constructor and destructor)

MikeDvorskiy (commented):
Despite the changes have been merged already, I would recommend to fix the mentioned issues with  `union __storage`.

PR #1366, opened 2024-01-22
Testing Bugfix, histogram.pass, unique call number

danhoeflinger:
The tests as written were resulting in kernel name clashes with `-DONEDPL_USE_UNNAMED_LAMBDA=OFF`.  This is due to different types being used to call the histogram API with the same call number / kernel name.  

Kernel name generation assumes the the user is responsible for unique kernel names as input to oneDPL algorithms when `-DONEDPL_USE_UNNAMED_LAMBDA=OFF`.  To keep names reasonably sized, we do not embed ALL information about the kernel into the kernel name, just enough to distinguish between multiple kernels generated from the same external oneDPL API call.  Therefore, the fix is not to add type information to the kernel name, but rather to provide unique call numbers for each API call in the testing.

### DIFF
--- a/test/parallel_api/numeric/numeric.ops/histogram.pass.cpp
+++ b/test/parallel_api/numeric/numeric.ops/histogram.pass.cpp
@@ -61,7 +61,7 @@ struct test_histogram_range_bins
     }
 };
 
-template <typename Size, typename T>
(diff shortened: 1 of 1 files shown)


SergeyKopienko (approved):
LGTM

mmichel11 (approved):
LGTM

PR #1367, opened 2024-01-22
Histogram bugfixes for FPGA and icpx 2022.3 

danhoeflinger:
This PR addresses two bugfixes for histogram

1. FPGA handling for converting to device policy, bad rebase did not change with some refactoring to handling of policy (and correcting a universal reference to a const lvalue ref to be consistent with device policy)
2. Removing unnecessary `begin()` call for a local_accessor, which caused problems on icpx 2022.3.  We only need an accessor which supports bracket operators in the function we are calling, so we can just pass the `local_accessor` itself.
3. Removing unnecessary and problematic include of `async_impl_hetero.h`


mmichel11 (approved):
LGTM

PR #1368, opened 2024-01-22
Bugfix permutation iterator, recursion for fancy iterators composed of passed directly components

danhoeflinger:
This resolves an issue with `permutation_iterator` using a source iterator that is a fancy iterator fully composed of "passed directly" components.

This includes: 
`permutation_iterator`, `zip_iterator`, transform_iterator` and any combination of them using passed directly components (usm pointers, etc.)

Before this fix, this overload of the handling of `permutation_iterator` did not recurse, causing problems during the `_require_access` call for kernels, as these functions do not know how to fancy iterator typed directly.  Instead they know how to handle views which match each of these types, which are generated via recursion within the `__get_sycl_range` struct.

With permutation iterator we do not know the size of the source iterator which will be used.  However, since no real processing (transform, copy, require...) will occur for "passed directly" components, we can reasonably just use a length of one, and be confident this will not cause extra overhead, or missing data.  

This of course still assumes that the user's map indices fall within bounds of the source data for the `permutation_iterator`, but that requirement is always inherent with its usage.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -524,9 +524,16 @@ struct __get_sycl_range
         auto __n = __last - __first;
         assert(__n > 0);
 
-        // The size of the source range is unknown. So, just base iterator is passing to permutation_view
+        // The size of the source range is unknown. However, we know the base is passed directly,
+        // therefore we know that it is composed of purely passed directly components. To properly
+        // wrap the underlying ranges in views to be handled by require access properly, we must recurse.
+        // We will not require any real "processing" of the data, so the length of the range is
+        // unimportant. We use length one here as we don't have anything better.
+        auto base_src = __first.base();
+        auto res_src = __process_input_iter<_LocalAccMode>(base_src, base_src + 1);
+
         //_Map is handled by recursively calling __get_sycl_range() in __get_permutation_view.
-        auto rng = __get_permutation_view(__first.base(), __first.map(), __n);
+        auto rng = __get_permutation_view(res_src.all_view(), __first.map(), __n);
 
         return __range_holder<decltype(rng)>{rng};
     }


mmichel11:
I see the problem here and why your patch resolves the issue. Just to clarify my understanding: Is it safe for us to just pass a size of 1 `auto res_src = __process_input_iter<_LocalAccMode>(base_src, base_src + 1);` since there is never any error checking on the created view when we index that we are indexing passed the size?

If my understanding is true, is this something we can rely on long-term or is a temporary solution?

danhoeflinger:
> I see the problem here and why your patch resolves the issue. Just to clarify my understanding: Is it safe for us to just pass a size of 1 `auto res_src = __process_input_iter<_LocalAccMode>(base_src, base_src + 1);` since there is never any error checking on the created view when we index that we are indexing passed the size?
> 
> If my understanding is true, is this something we can rely on long-term or is a temporary solution?

If you look back at #1254, this is effectively the same approach we used to have: 

https://github.com/oneapi-src/oneDPL/pull/1254/files#diff-b240d6dc11eec9d0fe3a00e10f8506df39d3c09420d093c9c18a184f129a4438L487-L491

Of course, just because we did it before does not mean inherently it was correct or robust.  

The reality of the situation is we have no information about what part of the underlying base iterator will be used without traversing and keeping track of every index in the map (which is not a feasible solution performance-wise).  For some types like host_iterators, this is a big problem.  However, for "passed directly" types,  where nothing is done to prepare the data, and nothing is needed to `_require_access` for kernels using that data, this ends up being OK.  Technically, we create a `guard_view` of an incorrect size, but as you say this is not something which is checked in practice.  

There is another possible option for dealing with this which may be slightly better:
One could instead add infrastructure within `__require_access()` to detect and skip fancy iterators which are "passed directly".  This would provide effectively the same solution, without misrepresenting the size of the base range, but rather still relying upon a base iterator.  This will be more (and relatively redundant ) code in the form of overloads for `__require_access()`.  

I'd be open to this or other options as alternatives.

mmichel11:
> > I see the problem here and why your patch resolves the issue. Just to clarify my understanding: Is it safe for us to just pass a size of 1 `auto res_src = __process_input_iter<_LocalAccMode>(base_src, base_src + 1);` since there is never any error checking on the created view when we index that we are indexing passed the size?
> > If my understanding is true, is this something we can rely on long-term or is a temporary solution?
> 
> If you look back at #1254, this is effectively the same approach we used to have:
> 
> https://github.com/oneapi-src/oneDPL/pull/1254/files#diff-b240d6dc11eec9d0fe3a00e10f8506df39d3c09420d093c9c18a184f129a4438L487-L491 (you have to expand `utils_ranges_sycl.h` to see the highlighted section)
> 
> Of course, just because we did it before does not mean inherently it was correct or robust.
> 
> The reality of the situation is we have no information about what part of the underlying base iterator will be used without traversing and keeping track of every index in the map (which is not a feasible solution performance-wise). For some types like host_iterators, this is a big problem. However, for "passed directly" types, where nothing is done to prepare the data, and nothing is needed to `_require_access` for kernels using that data, this ends up being OK. Technically, we create a `guard_view` of an incorrect size, but as you say this is not something which is checked in practice.
> 
> There is another possible option for dealing with this which may be slightly better: One could instead add infrastructure within `__require_access()` to detect and skip fancy iterators which are "passed directly". This would provide effectively the same solution, without misrepresenting the size of the base range, but rather still relying upon a base iterator. This will be more (and relatively redundant ) code in the form of overloads for `__require_access()`.
> 
> I'd be open to this or other options as alternatives.

I do not have a strong stance on if this solution or an alternative is chosen. I think as long as this workaround is well-documented in the PR (which it is), then it is a sufficient fix.

danhoeflinger:
Closing this PR in favor of #1371.  
The design *should* accommodate `passed_directly` fancy iterators, but had some bugs.  That PR fixes those bugs instead of adding this workaround.  It seems to be a cleaner approach, and is a simplified version of what I mention in the discussion above.

PR #1371, opened 2024-01-23
Fix zip iterators to be device copyable, fix for require_access

danhoeflinger:
This PR contains two bugfixes, which both impact the handling of  `permutation_iterator<zip_iterator<usm_ptr,usm_ptr>,usm_ptr>` and similar types as input data to oneDPL APIs.  These bugs were originated in #1254.

1. Make the fancy iterators: `zip_iterator` and `zip_forward_iterator` trivially copyable, and therefore device copyable, along with tests to check this
2. Add forward declaration for `__require_access`, which is to be necessary for `_require_access_args` to find this overload.

During investigation of this error, we discovered that transform_iterator and therefore permutation_iterator were not officially SYCL 2020 device copyable (although the work in practice with icpx's implementation of SYCL as if they were device copyable).  An issue #1386 has been created to record this, and define the plan for resolving this soon.

### DIFF
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -411,14 +411,8 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
     transform_iterator() = default;
-    transform_iterator(_Iter __it)
-        : __my_it_(__it), __my_unary_func_()
-    {
-    }
-    transform_iterator(_Iter __it, _UnaryFunc __unary_func)
-        : __my_it_(__it), __my_unary_func_(__unary_func)
-    {
-    }
+    transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_() {}
+    transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(__it), __my_unary_func_(__unary_func) {}
     transform_iterator(const transform_iterator& __input) = default;
     transform_iterator&
     operator=(const transform_iterator& __input)
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -410,7 +410,10 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc()) : __my_it_(__it), __my_unary_func_(__unary_func) {}
+    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc())
+        : __my_it_(__it), __my_unary_func_(__unary_func)
+    {
+    }
     transform_iterator(const transform_iterator& __input) = default;
     transform_iterator&
     operator=(const transform_iterator& __input)
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -89,11 +89,8 @@ class zip_forward_iterator
     typedef ::std::tuple<typename ::std::iterator_traits<_Types>::pointer...> pointer;
     typedef ::std::forward_iterator_tag iterator_category;
 
-    zip_forward_iterator() : __my_it_() {}
+    zip_forward_iterator() = default;
     explicit zip_forward_iterator(_Types... __args) : __my_it_(::std::make_tuple(__args...)) {}
-    zip_forward_iterator(const zip_forward_iterator& __input) = default;
-    zip_forward_iterator&
-    operator=(const zip_forward_iterator& __input) = default;
 
     reference operator*() const
     {
@@ -156,11 +153,8 @@ class counting_iterator
     typedef _Ip reference;
     typedef ::std::random_access_iterator_tag iterator_category;
 
-    counting_iterator() : __my_counter_() {}
+    counting_iterator() = default;
     explicit counting_iterator(_Ip __init) : __my_counter_(__init) {}
-    counting_iterator(const counting_iterator& __input) = default;
-    counting_iterator&
-    operator=(const counting_iterator& __input) = default;
 
     reference operator*() const { return __my_counter_; }
     reference operator[](difference_type __i) const { return *(*this + __i); }
@@ -274,12 +268,8 @@ class zip_iterator
     typedef ::std::random_access_iterator_tag iterator_category;
     using is_zip = ::std::true_type;
 
-    zip_iterator() : __my_it_() {}
+    zip_iterator() = default;
     explicit zip_iterator(_Types... __args) : __my_it_(::std::make_tuple(__args...)) {}
-    explicit zip_iterator(std::tuple<_Types...> __arg) : __my_it_(__arg) {}
-    zip_iterator(const zip_iterator& __input) = default;
-    zip_iterator&
-    operator=(const zip_iterator& __input) = default;
 
     reference operator*() const
     {
@@ -419,7 +409,12 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc())
+    transform_iterator() = default;
+    transform_iterator(_Iter __it)
+        : __my_it_(__it)
+    {
+    }
+    transform_iterator(_Iter __it, _UnaryFunc __unary_func)
         : __my_it_(__it), __my_unary_func_(__unary_func)
     {
     }
@@ -594,9 +589,6 @@ class permutation_iterator
         : my_source_it(input1), my_index(counting_iterator<difference_type>(__idx), __f)
     {
     }
-    permutation_iterator(const permutation_iterator& __input) = default;
-    permutation_iterator&
-    operator=(const permutation_iterator& __input) = default;
 
   private:
     template <typename _T = _Permutation, ::std::enable_if_t<__internal::__is_functor<_T>, int> = 0>
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -430,7 +430,6 @@ class transform_iterator
         __my_it_ = __input.__my_it_;
         return *this;
     }
-
     reference operator*() const { return __my_unary_func_(*__my_it_); }
     reference operator[](difference_type __i) const { return *(*this + __i); }
     transform_iterator&
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -89,7 +89,7 @@ class zip_forward_iterator
     typedef ::std::tuple<typename ::std::iterator_traits<_Types>::pointer...> pointer;
     typedef ::std::forward_iterator_tag iterator_category;
 
-    zip_forward_iterator() = default;
+    zip_forward_iterator() : __my_it_() {}
     explicit zip_forward_iterator(_Types... __args) : __my_it_(::std::make_tuple(__args...)) {}
 
     reference operator*() const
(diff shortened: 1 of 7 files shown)


MikeDvorskiy (approved):
LGTM.

BTW, do whether we want to add a test here?

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -91,13 +91,9 @@ class zip_forward_iterator
 
     zip_forward_iterator() : __my_it_() {}
     explicit zip_forward_iterator(_Types... __args) : __my_it_(::std::make_tuple(__args...)) {}
-    zip_forward_iterator(const zip_forward_iterator& __input) : __my_it_(__input.__my_it_) {}
+    zip_forward_iterator(const zip_forward_iterator& __input) = default;
```
rarutyun, line 94: I would propose to remove the definition of copy constructor and copy assignment completely. Or add definitions of move assignment and move constructor. Please play with this https://godbolt.org/z/M19dbPhMP and let me know if you have questions.

SergeyKopienko: Probably also better to rewrite default constructor as
`zip_forward_iterator() = default;`

test/parallel_api/iterator/transform_iterator.pass.cpp
```diff
@@ -162,12 +162,32 @@ void test_multi_transform_copy(size_t buffer_size)
     TestUtils::invoke_on_all_hetero_policies<2>()(test, tr3_sycl_source_begin, tr3_sycl_source_end, sycl_result_begin, buffer_size, identity + 3);
 }
 
+void
+test_copyable()
```
SergeyKopienko, line 166: You call this function from the code under `#if TEST_DPCPP_BACKEND_PRESENT` macro.
Better to move it under `#if TEST_DPCPP_BACKEND_PRESENT` check too.

test/parallel_api/iterator/zip_iterator.pass.cpp
```diff
@@ -121,6 +121,10 @@ DEFINE_TEST(test_for_each)
 
         auto tuple_first1 = oneapi::dpl::make_zip_iterator(std::make_tuple(first1, first1));
         auto tuple_last1 = oneapi::dpl::make_zip_iterator(last1, last1);
+        if (!this->host_buffering_required())
```
SergeyKopienko, line 124: @danhoeflinger, could you please explain this check, why it's required?

SergeyKopienko:
Looks like there are a lot of changes in our fancy iterators, not all of them are really required from my point of view.
Probably it's enough to change only some things like in the PR https://github.com/oneapi-src/oneDPL/pull/1374

SergeyKopienko (dismissed):
Looks like there are a lot of changes in our fancy iterators, not all of them are really required from my point of view.
Probably it's enough to change only some things like in the PR https://github.com/oneapi-src/oneDPL/pull/1374, based on this code.

test/parallel_api/iterator/permutation_iterator.pass.cpp
```diff
@@ -35,6 +35,13 @@ main()
     const std::size_t perm_size_expected = kDefaultIndexStepOp.eval_items_count(countingItDistanceResult);
     auto permItBegin = dpl::make_permutation_iterator(countingItBegin, kDefaultIndexStepOp);
     auto permItEnd = permItBegin + perm_size_expected;
+
+    //transform_iterator is not trivially_copyable, as it defines a copy assignment operator which does not copy
```
MikeDvorskiy, line 39: But transform_iterator has a default copy constructor. Is it enough to be a trivial copyable type.

(https://godbolt.org/z/KheKPYq37)

akukanov:
> Note: I was unable to find a deprecation note about device_copyable in the sycl 2020 spec, but based this on: https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/types.hpp#L2581-L2588

@danhoeflinger I interpret that code as a checker for an old (pre-SYCL 2020) definition of device copyability. The lines above that contain predicates for checking the SYCL 2020 definition of device copyable types.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -91,13 +91,9 @@ class zip_forward_iterator
 
     zip_forward_iterator() : __my_it_() {}
```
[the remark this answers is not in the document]
SergeyKopienko: Please take a look at https://github.com/oneapi-src/oneDPL/pull/1374 for details.

danhoeflinger:
> > Note: I was unable to find a deprecation note about device_copyable in the sycl 2020 spec, but based this on: https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/types.hpp#L2581-L2588
> 
> @danhoeflinger I interpret that code as a checker for an old (pre-SYCL 2020) definition of device copyability. The lines above that contain predicates for checking the SYCL 2020 definition of device copyable types.

@akukanov 
Agreed.  What I mean to say is that there is no note about a previous / changed definition of device copyable in sycl 2020 spec, so as I understand, it is officially unsupported.  However, in practice with this SYCL implementation, we are able to copy and use `transform_iterator` in kernels as if they were device copyable, because the implementation checks `is_device_copyable || is_deprecated_device_copyable`.   

It seems we *should* be getting a deprecation warning, but you'll see in the code there is a comment that this warning does not work.

test/parallel_api/iterator/zip_iterator.pass.cpp
danhoeflinger: This is required to avoid fancy iterators which are composed of `sycl_iterator` types.  We don't expect a `zip_iterator` of `sycl_iterators` to be "device copyable", because  

"the copy constructor selected for every non-static class type (or array of class type) member of T is trivial;" 
is no longer true from https://en.cppreference.com/w/cpp/language/copy_constructor#Trivial_copy_constructor.  `zip_iterator` contains a tuple of `sycl_iterators` which are not "trivial".

According to https://en.cppreference.com/w/cpp/language/classes#Trivially_copyable_class
"each eligible copy constructor is trivial" must be true for a type to be trivially copyable.

This is fine, because with the design of `__get_sycl_range`, we don't expect such a type to be directly device_copyable.  Once such a type is processed by `__get_sycl_range` it ends up as a `zip_view<all_view, all_view>` where all_view contains a device_copyable `sycl::accessor`.  This resulting type is device copyable.

The current design only requires fancy iterators which are made up of fully "passed_directly" components to be device copyable.

test/parallel_api/iterator/permutation_iterator.pass.cpp
danhoeflinger: I believe the situation we have is more like this:

https://godbolt.org/z/excn3oYdo

The presence of the eligible copy assignment operator which is non-trivial breaks `is_trivially_copyable`.

https://en.cppreference.com/w/cpp/language/classes#Trivially_copyable_class
https://en.cppreference.com/w/cpp/language/copy_assignment#Trivial_copy_assignment_operator

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -432,9 +427,14 @@ class transform_iterator
     transform_iterator&
     operator=(const transform_iterator& __input)
     {
+        //TODO: Investigate making transform_iterator trivially_copyable. This custom copy assignment operator prevents
```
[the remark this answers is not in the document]
danhoeflinger: see above.

test/parallel_api/iterator/transform_iterator.pass.cpp
danhoeflinger: I believe it is, see the `#endif` below.

SergeyKopienko: Yes, it's correct.

test/parallel_api/iterator/zip_iterator.pass.cpp
danhoeflinger: Sure, we should be able to check `is_passed_directly` instead.  Thanks.

include/oneapi/dpl/pstl/iterator_impl.h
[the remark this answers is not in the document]
SergeyKopienko: Probably if we declare `__my_unary_func_` as `_UnaryFunc` (without const) we able to remove this `operator=` at all.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Thanks for this example.  It is helpful.

SergeyKopienko: Probably if we apply https://github.com/oneapi-src/oneDPL/pull/1374 it's wouldn't required anymore at all.

test/parallel_api/iterator/zip_iterator.pass.cpp
danhoeflinger: Actually, considering this further, we would be reaching into private API to check either `is_passed_directly` or `is_sycl_iterator`.  We could redundantly add these structs in testing utils, but we already have the same information in this function (the point of this function is to determine if the data is in a host data destined for a sycl::buffer or USM memory).

I prefer to add a comment and keep this call as is.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -424,14 +410,17 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc())
-        : __my_it_(__it), __my_unary_func_(__unary_func)
-    {
-    }
+    transform_iterator() = default;
+    transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_() {}
+    transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(__it), __my_unary_func_(__unary_func) {}
```
danhoeflinger, line 415: It seems that `transform_iterator() = default;` is inherently different from the previous default constructor with default param arguments, and it is causing failures in `iterators.pass`.

danhoeflinger: Reverting "extra" changes which are not toward this PR's goal of fixing device copyable for fancy iterators.  (default constructor changes, etc.  
This makes the above comment unrelated to this PR.

test/parallel_api/iterator/permutation_iterator.pass.cpp
MikeDvorskiy: Actually, I would merely remove 
`transform_iterator(const transform_iterator& __input) = default;`
and 
`transform_iterator& operator=(const transform_iterator& __input)`

It makes "transform_iterator" trivially copyable

test/support/utils_sycl.h
```diff
@@ -79,6 +79,11 @@ auto async_handler = [](sycl::exception_list ex_list) {
     }
 };
 
+//Check that type is device copyable including types which are deprecated in sycl 2020
+template <typename T>
+static constexpr bool check_device_copyable_w_deprecrated_sycl2020 =
```
akukanov, line 84:
```suggestion
static constexpr bool check_if_device_copyable_by_sycl2020_or_by_old_definition =
```

But really, we should stop relying on the old definition and testing for it, sooner rather than later.

danhoeflinger: I definitely agree that we should take care of this soon.  
I'll take this suggestion.

test/parallel_api/iterator/permutation_iterator.pass.cpp
danhoeflinger: Discussed offline.  This creates a new requirement that lambdas are copy_assignable, which is not the case for lambdas with value captures (because those are represented as const member variables, and the copy assignment operator is then deleted).  
Without copy assignable functors, we then lose the ability to copy assign `transform_functor`. Which is a bigger problem than it not being "trivially copyable".  
Note that trivially copyable does not imply "copy assignable".

MikeDvorskiy: In that case, as we discussed with Alexey, we should just define the trait sycl::is_device_copyable_v<T> to true for transform_iterator type.

danhoeflinger: Yes, but we also discussed delaying that change to be addressed after the release, as it is not currently required to resolve the issues we are seeing.  I see that as an independent (although related) issue.  

If you insist on including this to approve this PR, then I will work on getting that in.  However, this is something we have not explored yet, and we need to merge today.

MikeDvorskiy (approved):
LGTM

PR #1379, opened 2024-01-24
[Dynamic Selection] Addressed comments for auto tune policy

AnuyaWelling2801:
1. Added a static assert for submit - sizeof...(KeyArgs) == sizeof...(Args)
2. Removed std::decay for wrapped_resource_t


vossmjp (approved):
Looks ok to me.

PR #1380, opened 2024-01-24
Adding legacy is_hetero trait and supporting it more fully

danhoeflinger:
Re-add legacy trait `is_hetero`, but supporting it more fully by adding an additional requirement for the `is_hetero` type to provide `get_idx()` which returns the buffer offset this iterator has to its base buffer.

Adding this function to `sycl_iterator`, incorporating it into `__get_sycl_range` implementation.  
This allows us to support hetero types other than `sycl_iterator` if they support `get_buffer()` and `get_idx()`.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
@@ -120,7 +120,7 @@ struct sycl_iterator
     }
 
     Size
-    get_buffer_offset() const
+    get_idx() const
     {
         return idx;
     }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
@@ -118,6 +118,12 @@ struct sycl_iterator
     {
         return buffer;
     }
+
+    Size
+    get_buffer_offset() const
+    {
+        return idx;
+    }
 };
 
 // mode converter when property::noinit present
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -258,7 +258,12 @@ struct is_sycl_iterator<oneapi::dpl::__internal::sycl_iterator<Mode, Types...>> 
 };
 
 template <typename Iter>
-inline constexpr bool is_sycl_iterator_v = is_sycl_iterator<Iter>::value;
+struct is_hetero_legacy_trait<Iter, ::std::enable_if_t<Iter::is_hetero::value>> : ::std::true_type
+{
+};
+
+template <typename Iter>
+inline constexpr bool is_sycl_iterator_v = is_sycl_iterator<Iter>::value || is_hetero_legacy_trait<Iter>::value;
 
 //A trait for checking if it needs to create a temporary SYCL buffer or not
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -515,8 +515,10 @@ struct __get_sycl_range
         auto __n = __last - __first;
         assert(__n > 0);
 
-        auto res_src = __process_input_iter<_LocalAccMode>(oneapi::dpl::begin(__first.base().get_buffer()),
-                                                           oneapi::dpl::end(__first.base().get_buffer()));
+        auto __base_iter = __first.base();
+        auto __base_buffer = __base_iter.get_buffer();
+        auto res_src = __process_input_iter<_LocalAccMode>(
+            oneapi::dpl::begin(__base_buffer) + __base_iter.get_buffer_offset(), oneapi::dpl::end(__base_buffer));
 
         //_Map is handled by recursively calling __get_sycl_range() in __get_permutation_view.
         auto rng = __get_permutation_view(res_src.all_view(), __first.map(), __n);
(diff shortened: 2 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -118,6 +118,12 @@ struct sycl_iterator
     {
         return buffer;
     }
+
+    Size
+    get_idx() const
```
MikeDvorskiy, line 123: Do I understand correctly that "soclomatic"  uses such API?

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -505,8 +515,10 @@ struct __get_sycl_range
         auto __n = __last - __first;
         assert(__n > 0);
 
-        auto res_src =
-            __process_input_iter<_LocalAccMode>(__first.base(), oneapi::dpl::end(__first.base().get_buffer()));
+        auto __base_iter = __first.base();
```
MikeDvorskiy, line 518: I think that new 4 lines are equivalent to the couple of line that we had before... No?

`__process_input_iter<_LocalAccMode>(__first.base(), oneapi::dpl::end(__first.base().get_buffer()));`

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -581,7 +593,7 @@ struct __get_sycl_range
         assert(__first < __last);
         using value_type = val_t<_Iter>;
 
-        const auto __offset = __first - oneapi::dpl::begin(__first.get_buffer());
+        const auto __offset = __first.get_idx();
```
MikeDvorskiy, line 596: I think it is equivalent to code that we had before.

`const auto __offset = __first - oneapi::dpl::begin(__first.get_buffer());`

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
danhoeflinger: SYCLomatic does not use `sycl_iterator` at all, but rather creates a type similar to `sycl_iterator`  (`device_pointer` / `device_iterator`) which are similar.  They already provide `get_idx()` within those types.  

The idea is to require these `is_hetero` types to provide `get_idx()` and then use it within `__get_sycl_range` processing.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: It is, but only if `first.base()` is a `sycl_iterator`.  If `first.base()` is instead a `dpct::device_pointer` this fails because no overload exists with such combination of types for `__process_input_iter` (`dpct::device_pointer`,`sycl_iterator`).

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: Similar response to above, it is the same as long as `__first` is a `sycl_iterator`.  If it is instead a `dpct::device_pointer` then this subtraction is undefined.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
MikeDvorskiy: Don't quite understand. `__get_sycl_range `works fine without `get_idx()`.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
MikeDvorskiy: I see..

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
MikeDvorskiy: ... Actually, the answer the next comment  explain that. I see..

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
MikeDvorskiy: I see.
It looks we need to extend the contract for sycl_iterator type

MikeDvorskiy (approved):
Due to the necessity of these changes  is not so obviously, it makes sense to add comments why we need some changes, otherwise, some new developer may re-write (revert )that code as redundant.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -505,8 +515,10 @@ struct __get_sycl_range
         auto __n = __last - __first;
         assert(__n > 0);
 
-        auto res_src =
-            __process_input_iter<_LocalAccMode>(__first.base(), oneapi::dpl::end(__first.base().get_buffer()));
+        auto __base_iter = __first.base();
+        auto __base_buffer = __base_iter.get_buffer();
+        auto res_src = __process_input_iter<_LocalAccMode>(oneapi::dpl::begin(__base_buffer) + __base_iter.get_idx(),
```
adamfidel, line 520: From what I understand, the issue is that we cannot easily get the index of the iterator because we are assuming that the base iterator type is a `sycl_iterator`. I believe we can solve this by instead of using `oneapl::dpl::begin`, explicitly constructing a begin iterator of the base iterator type directly from the buffer.

```suggestion
        using _BaseIterator = ::std::decay_t<decltype(__base_iter)>;
        auto __base_begin = _BaseIterator(__base_buffer);
        auto __idx = __first.base() - __base_begin;
        auto res_src = __process_input_iter<_LocalAccMode>(__base_begin + __idx,
```
In this way, we do not have to introduce a `get_idx` method to the `sycl_iterator` class or introduce the `is_hetero_legacy_trait` struct.

One drawback to using the above proposed approach is that we rely on the base iterator to define a constructor that takes just a `sycl::buffer`. From what I can tell, this is true for both `sycl_iterator` and `dpct::device_ptr`.

danhoeflinger: If I understand your suggestion correctly, I think the problem here is that we would lose the information of the buffer offset if we use `oneapi::dpl::begin`.

danhoeflinger:
Closing this in favor of #1383,  I will add the requested comments there.

PR #1381, opened 2024-01-25
Make optional predicate assignment in transform_iterator

SergeyKopienko:


include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -396,36 +396,114 @@ make_zip_iterator(std::tuple<_Tp...> __arg)
...
+    operator=(UnaryFuncWrapperImplNotCopyable<_UnaryFunc>&&)
+    {
+        // do nothing
+        return *this;
+    }
+};
+
+template <typename _UnaryFunc>
+using UnaryFuncWrapper =
+    ::std::conditional_t<::std::is_trivially_copyable_v<_UnaryFunc>, UnaryFuncWrapperImplCopyable<_UnaryFunc>,
```
danhoeflinger, line 479: `::std::is_trivially_copyable_v<_UnaryFunc>` should instead be 
`::std::is_trivially_copy_assignable_v<_UnaryFunc>`  or maybe just `::std::is_copy_assignable_v<_UnaryFunc>` (needs more thought).
However, `::std::is_trivially_copyable_v` does not imply this explicitly, as discussed offline.

PR #1384, opened 2024-01-25
Dynamic selection CI fix

AnuyaWelling2801:


egfefey (approved):
This restores the missing guard.
Should be fine now.

PR #1389, opened 2024-01-30
[test] Fix Kernel names in tests

SergeyKopienko:
In this PR we fix compile errors in some tests for CMake option `ONEDPL_USE_UNNAMED_LAMBDA=OFF`

### DIFF
--- a/test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
+++ b/test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
@@ -93,8 +93,8 @@ struct test_one_policy
     }
 };
 
-template <::std::size_t CallNumber, typename In1, typename In2, typename Out, typename Predicate,
-          typename _IteratorAdapter = _Identity>
+template <typename In1, typename In2, typename Out, typename Predicate,
+          typename _IteratorAdapter = _Identity, ::std::size_t CallNumber = 0>
 void
 test(Predicate pred, _IteratorAdapter adap = {})
 {
@@ -107,7 +107,7 @@ test(Predicate pred, _IteratorAdapter adap = {})
 
         invoke_on_all_policies<CallNumber>()(test_one_policy(), adap(in1.begin()), adap(in1.end()), adap(in2.begin()),
                                              adap(in2.end()), adap(out.begin()), adap(out.end()), pred);
-        invoke_on_all_policies<CallNumber + 1>()(test_one_policy(), adap(in1.cbegin()), adap(in1.cend()),
+        invoke_on_all_policies<CallNumber + 10>()(test_one_policy(), adap(in1.cbegin()), adap(in1.cend()),
                                                  adap(in2.cbegin()), adap(in2.cend()), adap(out.begin()),
                                                  adap(out.end()), pred);
     }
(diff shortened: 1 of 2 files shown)


test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
```diff
@@ -93,7 +93,8 @@ struct test_one_policy
     }
 };
 
-template <typename In1, typename In2, typename Out, typename Predicate, typename _IteratorAdapter = _Identity>
+template <::std::size_t CallNumber, typename In1, typename In2, typename Out, typename Predicate,
```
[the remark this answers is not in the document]
SergeyKopienko: From first attempt your propose still not compiled.

danhoeflinger (approved):
LGTM

test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
```diff
@@ -107,7 +107,7 @@ test(Predicate pred, _IteratorAdapter adap = {})
 
         invoke_on_all_policies<CallNumber>()(test_one_policy(), adap(in1.begin()), adap(in1.end()), adap(in2.begin()),
                                              adap(in2.end()), adap(out.begin()), adap(out.end()), pred);
-        invoke_on_all_policies<CallNumber + 1>()(test_one_policy(), adap(in1.cbegin()), adap(in1.cend()),
+        invoke_on_all_policies<CallNumber + 10>()(test_one_policy(), adap(in1.cbegin()), adap(in1.cend()),
```
SergeyKopienko, line 110: Why exactly `+10` ?

MikeDvorskiy: 10 looks enough. Even if some body will add 1-5 additional calls (with the same other types) it works. No problem to write bigger 'capacity' - let's write 100, if we think that it is not ebough.

SergeyKopienko: If we are ready to discuss about this, so let's discuss.
From my point of view, better to manage this things on caller side, and have simple and clear solution as before.

MikeDvorskiy: I've just proposed 15 line to change instead of 24 lines in Sergey's variant.
Both variant are absolutely the same. You can choose any one. I have no objections.

MikeDvorskiy: ... OK, there are some issues with CI with the second variant. So, I would suggest dont waste time and chose the first proposed variant.

SergeyKopienko: In this case you are able simple to revert one top commit.

MikeDvorskiy (dismissed):
LGTM

danhoeflinger:
> There is an error in CI
I believe this is from tests added in #1360.  It sounds like there was a plan to add a broken macro to avoid these failures but it hasn't been merged yet. https://github.com/oneapi-src/oneDPL/pull/1360#issuecomment-1912328521

MikeDvorskiy (approved):
LGTM.
Regarding the CI error - it is another issue, not connected with current changes. It seems it will be fixed with https://github.com/oneapi-src/oneDPL/pull/1393

SergeyKopienko:
> > There is an error in CI
> 
> I believe this is from tests added in #1360. It sounds like there was a plan to add a broken macro to avoid these failures but it hasn't been merged yet. [#1360 (comment)](https://github.com/oneapi-src/oneDPL/pull/1360#issuecomment-1912328521)

@mmichel11 and @dmitriy-sobolev already going to create these broken macros by proper way.

PR #1392, opened 2024-01-30
Update jinja to eliminate xmlattr vulnerability.

timmiesmith:
Older Jinja versions are vulnerable to HTML attribute injection when passing user input as keys to xmlattr filter. The issue has been resolved in Jinja 3.1.3. Updating our doc build to use the newer version and eliminate exposure to the issue.


timmiesmith:
Updating sphinx to 4.4.0, which doesn't use the jinja2 APIs deprecated in 3.1.3 allows documentation to build successfully.

dcbenito (approved):
LGTM

PR #1393, opened 2024-01-30
[test] Introduce broken macro if workaround for reduce_by_segment is not used

mmichel11:
https://github.com/oneapi-src/oneDPL/pull/1360 introduced `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` which forces a different code path in `reduce_by_segment` to avoid a driver bug that causes incorrect results on certain hardware.

Currently, if this macro is not set, then `reduce_by_segment.pass` will fail with the DPCPP backend with 64-bit types on certain Intel GPUs. This PR excludes these tests when `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` is `0` or undefined to avoid these known failures.

The solution currently is implemented with a compile-time preprocessor check. However, when a driver fix is provided the long-term solution may involve a runtime check by checking the driver version of the `sycl::device`. I have marked this as a todo in the code.

### DIFF
--- a/test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
+++ b/test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
@@ -368,7 +368,7 @@ run_test()
     std::cout << "\t" << "run_test<float,  BinaryPredicate<float>,  BinaryOperation<float>>();" << std::endl;
 #endif
     run_test<float,  BinaryPredicate<float>,  BinaryOperation<float>>();
-#ifndef _PSTL_ICPX_TEST_RED_BY_SEG_BROKEN_64BIT_TYPES
+#if !_PSTL_ICPX_TEST_RED_BY_SEG_BROKEN_64BIT_TYPES || ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION
 #if LOG_TEST_INFO
     std::cout << "\t" << "run_test<double, BinaryPredicate<double>, BinaryOperation<double>>();" << std::endl;
 #endif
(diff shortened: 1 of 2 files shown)


test/support/test_config.h
```diff
@@ -151,4 +151,12 @@
 #   define _PSTL_ICPX_TEST_RED_BY_SEG_OPTIMIZER_CRASH 0
 #endif
 
+// If the workaround macro for the 64-bit type bug is not defined by the user, then exclude 64-bit type testing
+// in reduce_by_segment.pass.cpp.
+// TODO: When a driver fix is provided to resolve this issue, consider altering this macro or checking the driver version at runtime
+// of the underlying sycl::device to determine whether to include or exclude 64-bit type tests.
+#if !ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION && defined(__INTEL_LLVM_COMPILER)
```
SergeyKopienko, line 158: From my point of view this macro shouldn't depend on `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION`,
because probably we need use it only in the test where `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` is not defined.

mmichel11: For clarification, is your suggestion that if `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` is set to `0`, then we should still test with 64-bit type with DPCPP backend and that we should only exclude it in the case where it is not defined at all?

I see your point here and agree this may be better.

SergeyKopienko: ... when `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` is 0 or undefined.
Yes. I prefer in this case exclude only one problem test case for concrete type and operation.

If `#if ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` is `true`, when your broken macro shouldn't be used in test (avoided by preprocessor directives and so on).

test/support/test_config.h
```diff
@@ -151,4 +151,12 @@
 #   define _PSTL_ICPX_TEST_RED_BY_SEG_OPTIMIZER_CRASH 0
 #endif
 
+// If the workaround macro for the 64-bit type bug is not defined by the user, then exclude 64-bit type testing
+// in reduce_by_segment.pass.cpp.
+// TODO: When a driver fix is provided to resolve this issue, consider altering this macro or checking the driver version at runtime
+// of the underlying sycl::device to determine whether to include or exclude 64-bit type tests.
+#if !defined(ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION) && defined(__INTEL_LLVM_COMPILER)
```
SergeyKopienko, line 158: #if  defined(__INTEL_LLVM_COMPILER)
#    define _PSTL_ICPX_TEST_RED_BY_SEG_BROKEN_64BIT_TYPES
#endif

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
```diff
@@ -368,10 +368,12 @@ run_test()
     std::cout << "\t" << "run_test<float,  BinaryPredicate<float>,  BinaryOperation<float>>();" << std::endl;
 #endif
     run_test<float,  BinaryPredicate<float>,  BinaryOperation<float>>();
+#ifndef _PSTL_ICPX_TEST_RED_BY_SEG_BROKEN_64BIT_TYPES
```
SergeyKopienko, line 371: Replace
```C++
#ifndef _PSTL_ICPX_TEST_RED_BY_SEG_BROKEN_64BIT_TYPES
```
to
```C++
#if !_PSTL_ICPX_TEST_RED_BY_SEG_BROKEN_64BIT_TYPES || ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION
```

test/support/test_config.h
mmichel11: Done. Also, this bug is exclusive to release builds, so I have added an additional condition: `!PSTL_USE_DEBUG` to not undertest this configuration.

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
mmichel11: Done. I have also removed the macro exclusion of flag predicate testing. This test takes the safe codepath due to the predicate, so we do not need to exclude it here.

SergeyKopienko (approved):
LGTM

PR #1394, opened 2024-02-01
[oneDPL][sycl] sycl::is_device_copyable specialization of the SYCL trait for some oneDPL types

MikeDvorskiy:
[oneDPL][sycl] + sycl::is_device_copyable specialization of the SYCL trait for some oneDPL types.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
@@ -24,12 +24,6 @@
 #ifndef _ONEDPL_SYCL_TRAITS_H
 #define _ONEDPL_SYCL_TRAITS_H
 
-#include "../../../internal/function.h"
-#include "../../../binary_search_impl.h"
-#include "unseq_backend_sycl.h"
-#include "../../utils.h"
-#include "../../iterator_impl.h"
-
 #if __INTEL_LLVM_COMPILER && (__INTEL_LLVM_COMPILER < 20240100)
 
 #    define _ONEDPL_DEVICE_COPYABLE(TYPE)                                                                              \
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
@@ -24,6 +24,12 @@
 #ifndef _ONEDPL_SYCL_TRAITS_H
 #define _ONEDPL_SYCL_TRAITS_H
 
+#include "../../../internal/function.h"
+#include "../../../binary_search_impl.h"
+#include "unseq_backend_sycl.h"
+#include "../../utils.h"
+#include "../../iterator_impl.h"
+
 #if __INTEL_LLVM_COMPILER && (__INTEL_LLVM_COMPILER < 20240100)
 
 #    define _ONEDPL_DEVICE_COPYABLE(TYPE)                                                                              \
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
@@ -72,6 +72,9 @@ _ONEDPL_DEVICE_COPYABLE(__op_uninitialized_fill)
 
 using namespace oneapi::dpl::__par_backend_hetero;
 
+template <typename _ExecutionPolicy, typename _Pred>
+struct __early_exit_find_or;
+
 _ONEDPL_DEVICE_COPYABLE(__early_exit_find_or);
 
 using namespace oneapi::dpl::unseq_backend;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
@@ -20,9 +20,19 @@
 #ifndef _ONEDPL_SYCL_TRAITS_H
 #define _ONEDPL_SYCL_TRAITS_H
 
+#if __INTEL_LLVM_COMPILER && (__INTEL_LLVM_COMPILER < 20240100)
+
+#define _ONEDPL_DEVICE_COPYABLE(TYPE) \
+template<typename... Ts > \
+struct sycl::is_device_copyable<TYPE<Ts...>, ::std::enable_if_t<!std::is_trivially_copyable_v<TYPE<Ts...>>>>: ::std::conjunction<sycl::is_device_copyable<Ts>...> {};
+
+#else
+
 #define _ONEDPL_DEVICE_COPYABLE(TYPE) \
 template<typename... Ts > \
 struct sycl::is_device_copyable<TYPE<Ts...>>: ::std::conjunction<sycl::is_device_copyable<Ts>...> {};
+
+#endif
 
 using namespace oneapi::dpl::__internal;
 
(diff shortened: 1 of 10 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -0,0 +1,49 @@
...
+_ONEDPL_DEVICE_COPYABLE(__reorder_pred)
+_ONEDPL_DEVICE_COPYABLE(__equal_value_by_pred)
+_ONEDPL_DEVICE_COPYABLE(__equal_value)
+_ONEDPL_DEVICE_COPYABLE(__not_equal_value)
+_ONEDPL_DEVICE_COPYABLE(__transform_functor)
+_ONEDPL_DEVICE_COPYABLE(__transform_if_unary_functor)
+_ONEDPL_DEVICE_COPYABLE(__transform_if_binary_functor)
+_ONEDPL_DEVICE_COPYABLE(__replace_functor)
+_ONEDPL_DEVICE_COPYABLE(__replace_copy_functor)
+_ONEDPL_DEVICE_COPYABLE(zip_forward_iterator)
```
danhoeflinger, line 39: I think we need to add to this list. If I am correct, effectively any functor with template types used for member variables which could be or could contain a type like `std::tuple` (which includes the other functors on this list). std::tuple is device copyable but not trivially copyable, so such a specialization would be necessary.

Here is a list of functors which seem like they may need to be added.  It may not be exhaustive.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h:
 `fill_functor`
 `generate_functor`
`__brick_fill`
`__brick_fill_n`
`__search_n_unary_predicate`
`__is_heap_check`

include/oneapi/dpl/pstl/hetero/utils_hetero.h:
`equal_predicate`
`adjacent_find_fn`
`__create_mask_unique_copy`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
`__early_exit_find_or`

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h:
`walk_n`
`walk_adjacent_difference`
`transform_reduce`
`reduce_over_group`
`single_match_pred_by_idx`
`multiple_match_pred`
`n_elem_match_pred`
`first_match_pred`
`__create_mask`
`__copy_by_mask`
`__partition_by_mask`
 `__global_scan_functor`
`__scan`
`__brick_includes`
`__brick_set_op`
`__brick_reduce_idx`

include/oneapi/dpl/internal/binary_search_impl.h:
 `custom_brick` (binary_search), 

include/oneapi/dpl/internal/function.h:
 `replace_if_fun`, 
 `scan_by_key_fun` 
 `segmented_scan_fun`
 `scatter_and_accumulate_fun`
 `transform_if_stencil_fun`

include/oneapi/dpl/pstl/memory_impl.h:
 `__op_uninitialized_fill`

danhoeflinger (commented):
As mentioned offline, I think this is the right approach.  I think the list of types may need to be bigger (unless there is something I'm missing).  

I also think that we should look in a different PR at `transform_iterator` specifically, and attempt something related to #1381 to make copy/move assignment work fully when it is available for the functor.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
MikeDvorskiy: @danhoeflinger,
Would you prefer to add the rest functors written above step by step in the next PRs or into this one the all mentioned functors wholly?

danhoeflinger: I don't have a strong opinion if they go in here or in a separate PR. There is nothing we are changing in this PR which makes the status of the other functors I list worse than it is now. However I believe these functors fall into a similar category with the ones already in this PR.

Do you agree that these seem to be required?  
I believe the requirement for inclusion in the list is when a type may contain a member function of another type which is device copyable but not trivially copyable (like `std::tuple`, or any other type in this list).

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -0,0 +1,49 @@
...
+// This file contains some specialization SYCL traits for some oneDPL types
+//
+// Include this header before a kernel submit SYCL code
+
+#ifndef _ONEDPL_SYCL_TRAITS_H
+#define _ONEDPL_SYCL_TRAITS_H
+
+#define _ONEDPL_DEVICE_COPYABLE(TYPE) \
+template<typename... Ts > \
+struct sycl::is_device_copyable<TYPE<Ts...>>: ::std::conjunction<sycl::is_device_copyable<Ts>...> {};
```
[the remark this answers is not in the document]
MikeDvorskiy: yes... it makes sense

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
MikeDvorskiy: done

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -0,0 +1,104 @@
...
+
+#endif
+
+using namespace oneapi::dpl::__internal;
+
+_ONEDPL_DEVICE_COPYABLE(__not_pred)
```
danhoeflinger, line 39: Do we need to include the headers with all the definitions of these functors / iterators?

danhoeflinger (commented):
I still believe approach is still the right one.

However, I see an issue here with maintainability of this code.  It is not going to be clear that adding such a line to this header is required whenever we make a new internal helper functor.  

Would it be better to group the macro `_ONEDPL_DEVICE_COPYABLE` with the functor itself, and define it as a no-op for backends which are not dpcpp?  This way, when someone looks to create a new functor, they have a better chance of seeing this in an example they are working from?

Mixing the macro in with the functor definition clutters the non-sycl code with sycl-related things, but I worry about this being the source of future errors as times goes on.

MikeDvorskiy:
> I still believe approach is still the right one.
> 
> However, I see an issue here with maintainability of this code. It is not going to be clear that adding such a line to this header is required whenever we make a new internal helper functor.
> 
> Would it be better to group the macro `_ONEDPL_DEVICE_COPYABLE` with the functor itself, and define it as a no-op for backends which are not dpcpp? This way, when someone looks to create a new functor, they have a better chance of seeing this in an example they are working from?
> 
> Mixing the macro in with the functor definition clutters the non-sycl code with sycl-related things, but I worry about this being the source of future errors as times goes on.

In regard to "keep device copyable traits specialization with the functor itself". We discussed it (with Alexey or Ruslan) and tented to conclusion that it is worse comparing with to have one list in just sycl part code. In case of the host used side, the mentioned functor don't need any irrelevant code like, traits, const or macros. It pollutes  the code and makes harder LLVM synchronization.
Regarding the potential missed errors - in any case a user will see a compile time error because SYCL library has a static assert with device copyability check.

danhoeflinger:
> In regard to "keep device copyable traits specialization with the functor itself". We discussed it (with Alexey or Ruslan) and tented to conclusion that it is worse comparing with to have one list in just sycl part code. In case of the host used side, the mentioned functors don't need any irrelevant code like, traits, const or macros. It pollutes the code and makes harder LLVM synchronization. Regarding the potential missed errors - in any case a user will see a compile time error because SYCL library has a static assert with device copyability check.


Yes, I remember that discussion now.  I forgot the part about upstream LLVM synchronization.  What we have here is probably the best option, or at least I have not been able to come up with a better one.  

Perhaps we can at least add a comment similar to this in this header:

"Fancy iterators and internal functors which are device copyable when their template arguments are also device copyable should be explicitly specialized as such.  This is important when template argument member variables may be device copyable but not trivially copyable."

MikeDvorskiy:
> "Fancy iterators and internal functors which are device copyable when their template arguments are also device copyable should be explicitly specialized as such. This is important when template argument member variables may be device copyable but not trivially copyable."

Yes, it makes sense. I will add such comment.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
MikeDvorskiy: Actually, it supposed to be included in file where a submit is called, f.e.  `paralle_backend_sycl.h.`
it means the all mentioned types are known to a compiler.
BTW, I've really missed including that file.. I will fix it.

danhoeflinger: Do we need it to be included also in other submit locations like `parallel_backend_sycl_*.h`, (radix_sort, radix_sort_one_wg, reduce, histogram) ?

MikeDvorskiy: Probably, it makes sense to be included into "sycl_defs.h" once ?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -0,0 +1,152 @@
...
+_ONEDPL_DEVICE_COPYABLE(replace_if_fun)
+_ONEDPL_DEVICE_COPYABLE(scan_by_key_fun)
+_ONEDPL_DEVICE_COPYABLE(segmented_scan_fun)
+_ONEDPL_DEVICE_COPYABLE(scatter_and_accumulate_fun)
+_ONEDPL_DEVICE_COPYABLE(transform_if_stencil_fun)
+
+using namespace oneapi::dpl;
+
+_ONEDPL_DEVICE_COPYABLE(zip_iterator)
+_ONEDPL_DEVICE_COPYABLE(transform_iterator)
```
danhoeflinger, line 147: For `transform_iterator`, and `permutation_iterator`, it is possible we need to adjust our strategy slightly.  I think we may want to wait for our upcoming meeting before merging this PR.
Otherwise, I believe this LGTM.

danhoeflinger (dismissed):
LGTM
After our meeting and discussion, I think we can go forward with this PR as is.  I think no matter our plans for changes to `transform_iterator`, I think we can go ahead with specializing it as device_copyable.  

I think the clang format suggestions should be ignored.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -0,0 +1,152 @@
...
+_ONEDPL_DEVICE_COPYABLE(__reorder_pred)
+_ONEDPL_DEVICE_COPYABLE(__equal_value_by_pred)
+_ONEDPL_DEVICE_COPYABLE(__equal_value)
+_ONEDPL_DEVICE_COPYABLE(__not_equal_value)
+_ONEDPL_DEVICE_COPYABLE(__transform_functor)
+_ONEDPL_DEVICE_COPYABLE(__transform_if_unary_functor)
+_ONEDPL_DEVICE_COPYABLE(__transform_if_binary_functor)
+_ONEDPL_DEVICE_COPYABLE(__replace_functor)
+_ONEDPL_DEVICE_COPYABLE(__replace_copy_functor)
+_ONEDPL_DEVICE_COPYABLE(zip_forward_iterator)
```
danhoeflinger, line 58: It turns out `zip_forward_iterator` is only used on the host side (from `include/oneapi/dpl/pstl/algorithm_impl.h`), within functions which are enabled for host policies.  It is probably not necessary to specialize it as device copyable.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -0,0 +1,151 @@
...
+
+#    define _ONEDPL_DEVICE_COPYABLE(TYPE)                                                                              \
+        template <typename... Ts>                                                                                      \
+        struct sycl::is_device_copyable<TYPE<Ts...>> : ::std::conjunction<sycl::is_device_copyable<Ts>...>             \
+        {                                                                                                              \
+        };
+
+#endif
+
+using namespace oneapi::dpl::__internal;
```
adamfidel, line 46: @MikeDvorskiy By including this `using namespace` here, does this not pull in everything from our internal namespace to be visible by all user code now?

danhoeflinger: Good point... there are other instances of similar issues below as well.  We will need a PR to fix this I think.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
dmitriy-sobolev: > Actually, it supposed to be included in file where a submit is called, f.e. paralle_backend_sycl.h.

Sorry, but that's a hack... It is not reliable and does not work for kernel templates, which do not include `parallel_backend_sycl.h`.

PR #1400, opened 2024-02-09
Disallowing histogram with 64 bit output on FPGA

danhoeflinger:
FPGAs do not support 64 bit atomics.  The histogram pattern uses atomics of the value type of the output sequence.  
Adding static assertion for FPGAs to only allow sequences with a value type of 32bit or less.
Adjusting the tests to match.

Its rare that a 64 bit output type should be necessary for histogram.  Input sequences would only start to need 64 bit output with sizes greater than 4Billion (and that would only overflow if all elements map to a single bin).


timmiesmith (approved):
LGTM. Thank you.

PR #1406, opened 2024-02-13
[tests] Adding broken macro for cpp20 can_deref

danhoeflinger:
On Windows, MSVC concept for _Can_deref seems to return the wrong result, which leads to an incorrect implementation of move when dealing with sycl_iterator.  This results in a build error because sycl_iterator does not implement a dereference operator or bracket operator.
This broken macro avoids such examples in the tests for C++20 on windows.  This is a somewhat unsatisfying broken macro, as it has little to do with what we are trying to test, but it is a failure that we are not responsible for, and I see no easy way to avoid hitting while still testing the functionality in the test.


test/support/test_config.h
```diff
@@ -126,6 +126,9 @@
 
 #define _PSTL_MSVC_LESS_THAN_CPP20_COMPLEX_CONSTEXPR_BROKEN (_MSC_VER && __cplusplus < 202002L && _MSVC_LANG < 202002L)
 
+// In MSVC C++20, _Can_deref concept returns wrong result for sycl_iterator, resulting in errors moving sycl_iterator
+#define _PSTL_MSVC_GREATER_EQUAL_CPP20_CAN_DEREF_CONCEPT_BROKEN (_MSC_VER && __cplusplus >= 202002L && _MSVC_LANG >= 202002L)
```
SergeyKopienko, line 130: Should we check both `__cplusplus`  and `_MSVC_LANG` states here?

danhoeflinger: I'm following the example from the macro a few lines above it.  However, I suppose either just checking `_MSVC_LANG` only or using an `||` might be better here as I look a bit further (at test_macros.h for instance).  which do you suggest?

SergeyKopienko: May be
`#define _PSTL_MSVC_GREATER_EQUAL_CPP20_CAN_DEREF_CONCEPT_BROKEN (_MSVC_LANG >= 202002L)`
is enough?

danhoeflinger:
Marking as a draft for now, as more investigation is needed to understand why we are experiencing this issue and why `iter_move` is being called during this test body.

danhoeflinger:
Closing in favor of #1409

PR #1408, opened 2024-02-14
Enable zip_iterator to satisfy C++20 std::input_iterator constraint

mmichel11:
Currently with C++20, `oneapi::dpl::zip_iterator` cannot satisfy the C++20 [std::input_iterator](https://en.cppreference.com/w/cpp/iterator/input_iterator) constraint due to it's failure to satisfy [std::indirectly_readable](https://en.cppreference.com/w/cpp/iterator/indirectly_readable) which requires that the iterator's value and reference types to have a common reference.

To resolve this issue, we specialize `std::basic_common_reference` for `oneapi::dpl::__internal::tuple` to recursively fetch the common reference types of each corresponding tuple element in the input tuples.

Furthermore, the [tuple specialization of basic_common_reference](https://en.cppreference.com/w/cpp/utility/tuple/basic_common_reference) is not provided until C++23, so we must specialize this for C++20 to enable `zip_forward_iterator` support with these constraints which will meet compilation errors with the TBB backend if it is not provided.

### DIFF
--- a/include/oneapi/dpl/pstl/tuple_impl.h
+++ b/include/oneapi/dpl/pstl/tuple_impl.h
@@ -669,6 +669,25 @@ get(const oneapi::dpl::__internal::tuple<_Tp...>&& __a)
 {
     return ::std::move(__a).template get<_Idx>();
 }
+
+#if __cplusplus >= 202002L
+template <typename... TTypes, typename... UTypes, template <typename> typename TQual,
+          template <typename> typename UQual>
+struct basic_common_reference<oneapi::dpl::__internal::tuple<TTypes...>, oneapi::dpl::__internal::tuple<UTypes...>,
+                              TQual, UQual>
+{
+    using type = oneapi::dpl::__internal::tuple<::std::common_reference_t<TQual<TTypes>, UQual<UTypes>>...>;
+};
+#endif
+
+#if __cplusplus >= 202002L && __cplusplus < 202302L
+template <typename... TTypes, typename... UTypes, template <typename> typename TQual,
+          template <typename> typename UQual>
+struct basic_common_reference<::std::tuple<TTypes...>, ::std::tuple<UTypes...>, TQual, UQual>
+{
+    using type = ::std::tuple<::std::common_reference_t<TQual<TTypes>, UQual<UTypes>>...>;
+};
+#endif
 } // namespace std
 
 #endif // _ONEDPL_TUPLE_IMPL_H


include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -80,17 +80,17 @@ template <typename... _Types>
...
+    typedef typename oneapi::dpl::__internal::tuple<_Types...> __it_types;
 
   public:
     typedef ::std::make_signed_t<::std::size_t> difference_type;
-    typedef ::std::tuple<typename ::std::iterator_traits<_Types>::value_type...> value_type;
-    typedef ::std::tuple<typename ::std::iterator_traits<_Types>::reference...> reference;
-    typedef ::std::tuple<typename ::std::iterator_traits<_Types>::pointer...> pointer;
+    typedef oneapi::dpl::__internal::tuple<typename ::std::iterator_traits<_Types>::value_type...> value_type;
+    typedef oneapi::dpl::__internal::tuple<typename ::std::iterator_traits<_Types>::reference...> reference;
+    typedef oneapi::dpl::__internal::tuple<typename ::std::iterator_traits<_Types>::pointer...> pointer;
```
danhoeflinger, line 89: This isn't an argument against these changes, but to provide some context / history:

In #1432 (still unmerged), we removed a similar change because `zip_forward_iterator` was only used on the host side, not in hetero / dpcpp backend, and therefore did not need the tuple to be trivially copyable.

mmichel11:
For further context about the change to use `oneapi::dpl::__internal::tuple` in `zip_forward_iterator` here for reviewers:

I have tested with `C++23` with gcc 13 and using `std::tuple` builds fine due to the spec's update to include tuple-like common references. However, for `C++20` we have a gap since this is not provided in the standard. Using our own tuple resolves this gap as the standard allows us to define `std::basic_common_reference` for user-defined types with a very similar definition to what is provided in C++23.

Attempting to define `std::basic_common_reference` for `std::tuple` in this instance would be a spec violation and causes issues in practice, so I think just switching to our internal tuple is best.

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -669,6 +669,18 @@ get(const oneapi::dpl::__internal::tuple<_Tp...>&& __a)
...
     return ::std::move(__a).template get<_Idx>();
 }
+
+// To enable oneapi::dpl::zip_iterator to satisfy the requirements for the std::input_iterator concept in C++20,
+// we must provide a basic_common_reference specialization for our internal tuple implementation.
+#if _ONEDPL_CPP20_CONCEPTS_PRESENT
+template <typename... TTypes, typename... UTypes, template <typename> typename TQual,
+          template <typename> typename UQual>
+struct basic_common_reference<oneapi::dpl::__internal::tuple<TTypes...>, oneapi::dpl::__internal::tuple<UTypes...>,
+                              TQual, UQual>
```
danhoeflinger, line 679: I'm having trouble understanding `TQual` and `UQual`.  

Are they just something which is required for this function for it to be of use to the Standard library implementation? 
Perhaps it uses these to add type qualifiers like `const` to the types?

Its not very clear to me as it is written in the standard.

mmichel11: They are required for this function to be used by the standard library implementation. My understanding is that they add cvref-qualifiers to the types.

It is mentioned briefly in the explanation of [std::common_reference's](https://en.cppreference.com/w/cpp/types/common_reference) required behavior:

"Otherwise, if `std::basic_common_reference<std::remove_cvref_t<T1>, std::remove_cvref_t<T2>, T1Q, T2Q>::type` exists, where TiQ is a unary alias template such that TiQ<U> is U with the addition of Ti's cv- and reference qualifiers, then the member type type names that type;"

danhoeflinger:
One thing to note is that oneDPL's internal tuple may not support everything that `std::tuple` supports, we do have the ability to convert from one to the other, but that doesn't always allow full functionality.  One such example is #1472.  

I don't expect problems with this necessarily... however its worth acknowledging the potential risk of such a change.  

I don't think I like the following option, but I'll mention it anyway for discussion:
We could have a tuple type which resolves to std::tuple in most cases, and only for the specific situation we need it (TBB backend & C++ 20 only?) would it resolve to oneDPL internal tuple.  I think for C++23 and beyond, std::tuple should work just fine for these purposes because it will provide this specialization of `std::basic_common_reference` itself.

mmichel11:
@rarutyun @danhoeflinger 

I have looked a bit deeper to see if we can find a way to continue to use `std::tuple` in `zip_forward_iterator` with a `basic_common_reference` specialization. We run into the following compilation error in `std::common_reference_with` when using `std::tuple`:
```
/usr/include/c++/11/concepts:72:30: note: the expression ‘is_convertible_v<_From, _To> [with _From = std::tuple<double, double>&; _To = std::tuple<const double&, double&>]’ evaluated to ‘false’
   72 |     concept convertible_to = is_convertible_v<_From, _To>
```
I have looked into this error some more and it seems to be unavoidable with `std::tuple` in C++20 but is addressed with the changes to `std::tuple` in [P2321R2 (zip)](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2321r2.html) in C++23. The immediate compiler error we are hitting is that a tuple reference is not convertible to a tuple of references in C++20 but is in C++23. This is checked when finding a common reference between `zip_forward_iterator's` `value_type&` (tuple reference) and `reference` (tuple of references) in `std::indirectly_readable`.

The following godbolt example demonstrates the issue in C++20 and its resolution in C++23: https://godbolt.org/z/YzTje39K6.

With our internal tuple implementation, this is not an issue. My take is that we add some functionality to use `std::tuple` in `zip_forward_iterator` with C++17, switch to our internal tuple for C++20, and use `std::tuple` again with C++23 when we detect the appropriate feature macros. I am curious to hear your thoughts.

danhoeflinger:
> With our internal tuple implementation, this is not an issue. My take is that we add some functionality to use `std::tuple` in `zip_forward_iterator` with C++17, switch to our internal tuple for C++20, and use `std::tuple` again with C++23 when we detect the appropriate feature macros. I am curious to hear your thoughts.

I think this is a reasonable way to limit exposure of potential issues that could arise from switching to the internal tuple, and is probably the best path.  It could be possible to further limit exposure by limiting it only to the backends for which this concept is used (TBB).

mmichel11:
> > With our internal tuple implementation, this is not an issue. My take is that we add some functionality to use `std::tuple` in `zip_forward_iterator` with C++17, switch to our internal tuple for C++20, and use `std::tuple` again with C++23 when we detect the appropriate feature macros. I am curious to hear your thoughts.
> 
> I think this is a reasonable way to limit exposure of potential issues that could arise from switching to the internal tuple, and is probably the best path. It could be possible to further limit exposure by limiting it only to the backends for which this concept is used (TBB).
> 
> edit: However, without some known issue with our internal tuple to avoid, I don't think its necessary to limit only to the TBB backend, just raising it as an option.

I have added these changes and used the exact library features macros to dictate whether to use `std::tuple` or `oneapi::dpl::internal::__tuple`. If we are using C++20 and both concepts and concepts library support is present, then we use `oneapi::dpl::__internal::tuple`. Once we see `__cpp_lib_ranges_zip` and `__cpp_lib_ranges_zip`, then we can switch back to `std::tuple`. With C++17, these feature macros will evaluate to `0` and `std::tuple` will be used.

I have chosen not to limit to just the TBB backend in this commit. My reasoning is that `std::tuple` seems to not work well with proxy iterators until C++23. Currently, we only see issues with the TBB backend but since this is a general issue with `zip_forward_iterator` and C++20 concepts, I have chosen to disable it with other backends as well. Let me know if you think more discussion is needed.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -296,8 +300,29 @@
...
+#endif
+
+#if _ONEDPL___cplusplus >= 202302L && __cpp_lib_ranges_zip >= 202110L
+#    define _ONEDPL_CPP23_RANGES_ZIP_PRESENT 1
+#else
+#    define _ONEDPL_CPP23_RANGES_ZIP_PRESENT 0
+#endif
+
+// To use std::tuple as a proxy reference and satisfy iterator concepts for C++20 and beyond, we need
+// the changes to std::tuple in P2321R2 and the tuple-like basic_common_reference specialization in P2165R4.
```
danhoeflinger, line 322:
```
// When C++20 concepts are available, we must use std::tuple as a proxy reference to satisfy iterator concepts, which // requires the changes to std::tuple in P2321R2 and the tuple-like basic_common_reference specialization in P2165R4.
```
In my opinion, this or some similar phrasing makes the part of the expression before the `||` more clear.

danhoeflinger (commented):
Looks close to me, minor comments to address.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -79,18 +79,26 @@ struct __make_references
...
+    using __tuple_t =
+#if _ONEDPL_CAN_USE_STD_TUPLE_PROXY_ITERATOR
+        ::std::tuple<_Ts...>;
+#else
+        oneapi::dpl::__internal::tuple<_Ts...>;
+#endif
+
     static const ::std::size_t __num_types = sizeof...(_Types);
-    typedef typename oneapi::dpl::__internal::tuple<_Types...> __it_types;
+    typedef __tuple_t<_Types...> __it_types;
```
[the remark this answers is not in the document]
mmichel11: Since `__tuple_t` is not qualified, I believe it is incorrect to have `typename` here as there is no ambiguity it resolves. Doing so results in the following compilation error:
```
. . . oneDPL/include/oneapi/dpl/pstl/iterator_impl.h:91:22: error: expected nested-name-specifier
   91 |     typedef typename __tuple_t<_Types...> __it_types;
```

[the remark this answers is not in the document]
danhoeflinger: aha, yes.  Thanks for the explanation.  You are correct.

include/oneapi/dpl/pstl/onedpl_config.h
mmichel11: I agree this is clearer and have added this suggestion.

danhoeflinger (dismissed):
LGTM, but I'd recommend circling back with @rarutyun as well.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -16,6 +16,10 @@
 #ifndef _ONEDPL_CONFIG_H
 #define _ONEDPL_CONFIG_H
 
+#if _ONEDPL___cplusplus >= 202002L && __has_include(<version>)
+#    include <version>
+#endif
+
```
akukanov, line 22:
```suggestion
#if _ONEDPL___cplusplus >= 202002L && __has_include(<version>)
#    include <version>
#    define _ONEDPL_STD_FEATURE_MACROS_PRESENT 1
#endif

```

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -296,6 +300,30 @@
...
+#    define _ONEDPL_CPP23_TUPLE_LIKE_COMMON_REFERENCE_PRESENT 1
+#else
+#    define _ONEDPL_CPP23_TUPLE_LIKE_COMMON_REFERENCE_PRESENT 0
+#endif
+
+#if _ONEDPL___cplusplus >= 202302L && __cpp_lib_ranges_zip >= 202110L
+#    define _ONEDPL_CPP23_RANGES_ZIP_PRESENT 1
+#else
+#    define _ONEDPL_CPP23_RANGES_ZIP_PRESENT 0
+#endif
```
akukanov, line 319:
```suggestion
#if _ONEDPL_STD_FEATURE_MACROS_PRESENT
#    define _ONEDPL_CPP20_CONCEPTS_PRESENT (__cpp_concepts >= 201907L && __cpp_lib_concepts >= 202002L)
#    define _ONEDPL_CPP23_TUPLE_LIKE_COMMON_REFERENCE_PRESENT (_ONEDPL___cplusplus >= 202302L && __cpp_lib_tuple_like >= 202207L)
#    define _ONEDPL_CPP23_RANGES_ZIP_PRESENT (_ONEDPL___cplusplus >= 202302L && __cpp_lib_ranges_zip >= 202110L)
#else
#    define _ONEDPL_CPP20_CONCEPTS_PRESENT 0
#    define _ONEDPL_CPP23_TUPLE_LIKE_COMMON_REFERENCE_PRESENT 0
#    define _ONEDPL_CPP23_RANGES_ZIP_PRESENT 0
#endif
```

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -669,6 +669,18 @@ get(const oneapi::dpl::__internal::tuple<_Tp...>&& __a)
 {
     return ::std::move(__a).template get<_Idx>();
 }
+
+// To enable oneapi::dpl::zip_iterator to satisfy the requirements for the std::input_iterator concept in C++20,
+// we must provide a basic_common_reference specialization for our internal tuple implementation.
+#if _ONEDPL_CPP20_CONCEPTS_PRESENT
```
akukanov, line 675: It's not obvious to me why the presence of ``std::basic_common_reference`` depends on concepts. Is not it enough to check for C++20?

mmichel11: The reasoning here is that `std::basic_common_reference` is introduced in [P0898R3](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0898r3.pdf) which corresponds to `__cpp_lib_concepts`.

If we just check the C++ version and not the `__cpp_lib_concepts` feature macro, then we may be use a compiler that has C++20 support but the standard library version may not support library concepts (ex: libstdc++ 9).

include/oneapi/dpl/pstl/onedpl_config.h
mmichel11: Thanks, I have switched to defining them this way.

include/oneapi/dpl/pstl/onedpl_config.h
mmichel11: I have made this change.

include/oneapi/dpl/pstl/tuple_impl.h
akukanov: I see, thanks. Ok to keep it as you did, but a short comment would be useful.

mmichel11: A comment has been added clarifying why we guard the specialization with the concepts support macro. Also, I have kept the `__cpp_concepts` check in the macro since trying to use the concepts library without a compiler that supports concepts is problematic.

akukanov (dismissed):

danhoeflinger (approved):
I think the remaining concern is about changing the layout of `zip_forward_iterator` based upon the standard library version specified.  

`zip_forward_iterator` is in the `__internal` namespace, but we had discussed possibly adding a comment to its definition which points out that this should stay "internal", because of the layout ABI breaking changes which occur with different standard library implementations.  
It may not be clear to someone refactoring down the line that this would be problematic at first glance if they were wanting to expose it as a part of the public API.

I think such a comment is a good idea.  Otherwise, I think this is the best we can do for this issue.

mmichel11:
> I think the remaining concern is about changing the layout of `zip_forward_iterator` based upon the standard library version specified.
> 
> `zip_forward_iterator` is in the `__internal` namespace, but we had discussed possibly adding a comment to its definition which points out that this should stay "internal", because of the layout ABI breaking changes which occur with different standard library implementations. It may not be clear to someone refactoring down the line that this would be problematic if they were wanting to expose it as a part of the public API.
> 
> I think such a comment is a good idea. Otherwise, I think this is the best we can do for this issue.

Thanks, I have added a comment to address this in the zip_forward_iterator class definition.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -76,6 +76,10 @@ struct __make_references
 
 //zip_iterator version for forward iterator
 //== and != comparison is performed only on the first element of the tuple
+//
+//zip_forward_iterator is implemented as an internal class and should remain so. Users should never encounter
+//this class or be returned a type of its value_type, reference, etc as the tuple-like type internally
+//managed switches between C++ standard and library versions and could potentially break user code if exposed.
```
danhoeflinger, line 82:
``` 
//zip_forward_iterator is implemented as an internal class and should remain so. Users should never encounter
//this class or be returned a type of its value_type, reference, etc as the tuple-like type used internally
//is variable dependent on the C++ standard library version and could cause and inconsistent ABI due to 
//resulting layout changes of this class.

**Review by danhoeflinger — APPROVED** [review]
LGTM

File: include/oneapi/dpl/pstl/tuple_impl.h [hunk 9]
```diff
@@ -669,6 +669,19 @@ get(const oneapi::dpl::__internal::tuple<_Tp...>&& __a)
 {
     return ::std::move(__a).template get<_Idx>();
 }
+
+// To enable oneapi::dpl::zip_iterator to satisfy the requirements for the std::input_iterator concept in C++20,
+// we must provide a basic_common_reference specialization for our internal tuple implementation. std::basic_common_reference
+// is introduced in __cpp_lib_concepts, so we must ensure this feature is present.
+#if _ONEDPL_CPP20_CONCEPTS_PRESENT
```
rarutyun, line 676: Should it be
```suggestion
#if !_ONEDPL_CAN_USE_STD_TUPLE_PROXY_ITERATOR
```
If not, why?

rarutyun (commented):
After the discussion with @mmichel11 and going through the files it looks good to me except one thing

include/oneapi/dpl/pstl/tuple_impl.h
danhoeflinger: Why not always provide this feature for our internal tuple when we have the power to and parts of the standard library can make use of it?
(even if it is not used specifically by `zip_forward_iterator`)

mmichel11: With `!_ONEDPL_CAN_USE_STD_TUPLE_PROXY_ITERATOR` it limits this specialization to between C++20 and C++23. `_ONEDPL_CPP20_CONCEPTS_PRESENT` enables this for C++20 and beyond once library concepts are present.

The advantage of doing this is that `zip_iterator` can now satisfy `std::input_iterator` and other iterator traits as it always uses our internal tuple in its value and reference types and this change makes our tuple indirectly readable. If we were to use `!_ONEDPL_CAN_USE_STD_TUPLE_PROXY_ITERATOR`, then this behavior would break with C++23 and iterator concepts would not work with `zip_iterator`.

PR #1409, opened 2024-02-14
[tests] Remove std::distance calls for random access iters in permutation_iterator tests

danhoeflinger:
`std::distance` calls on random access iterator (`permutation_iterator<sycl_iterator, ...>`) typed variables cause compiler errors in MSVC C++20.
Somehow, `std::ranges::iter_move` is called within `std::distance`, resulting in an compiler error because `sycl_iterator` is not dereference-able.  

A simple subtraction will reliably suffice for these iterators to determine the size of the range between iterators, so this PR switches to this method, avoiding the compiler error.

This is a better alternative to https://github.com/oneapi-src/oneDPL/pull/1406, which simply turned off these tests, as we don't need to turn the tests off.


SergeyKopienko (approved):
LGTM

PR #1410, opened 2024-02-16
Extend compiler support for the unified USM and buffer storage

julianmi:
The existing unified USM or buffer storage used in the family of reduce algorithms uses the default constructor of SYCL accessors. This is a SYCL 2020 feature that was implemented here: https://github.com/intel/llvm/pull/7815. This limits the supported DPCPP compilers to 2023.1 and newer. This PR extends the compiler support by using a SYCL accessor with the `get_multi_ptr` routine on older (2023.0 and older) compilers instead of the `__usm_host_or_buffer_accessor` class. The `_ONEDPL_SYCL_USM_HOST_PRESENT` macro is further renamed to `_ONEDPL_SYCL_UNIFIED_USM_BUFFER_PRESENT` to highlight whether the compiler supports the unified USM and buffer implementation.

This issue was detected by @AidanBeltonS during the testing of https://github.com/oneapi-src/oneDPL/pull/1354. The proposed fix was tested with DPCPP 2022.0 through 2024.0.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -489,9 +489,8 @@ struct __usm_host_or_buffer_accessor
     // USM pointer
     __usm_host_or_buffer_accessor(sycl::handler& __cgh, _T* __usm_buf) : __ptr(__usm_buf), __usm(true) {}
 
-    template <sycl::access::decorated> // Mirror sycl::accessor::get_multi_ptr
     auto
-    get_multi_ptr() const // should be cached within a kernel
+    __get_pointer() const // should be cached within a kernel
     {
         return __usm ? __ptr : &__acc[0];
     }
@@ -564,6 +563,17 @@ struct __usm_host_or_buffer_storage
         return __usm;
     }
 };
+
+template <typename _Acc>
+auto
+__get_usm_host_or_buffer_accessor_ptr(const _Acc& __acc)
+{
+#if _ONEDPL_SYCL_UNIFIED_USM_BUFFER_PRESENT
+    return __acc.__get_pointer();
+#else
+    return &__acc[0];
+#endif
+}
 
 //A contract for future class: <sycl::event or other event, a value, sycl::buffers..., or __usm_host_or_buffer_storage>
 //Impl details: inheritance (private) instead of aggregation for enabling the empty base optimization.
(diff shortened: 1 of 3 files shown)


AidanBeltonS (dismissed):
LGTM!

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -82,12 +82,12 @@
 #    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE_IF_SUPPORTED(SIZE)
 #endif
 
-// The unified future supporting USM host memory and buffers is only supported after DPCPP 2023.1
+// The unified future supporting USM memory and buffers is only supported after DPCPP 2023.1
 // but not by 2023.2.
 #if (_ONEDPL_LIBSYCL_VERSION >= 60100 && _ONEDPL_LIBSYCL_VERSION != 60200)
-#    define _ONEDPL_SYCL_USM_HOST_PRESENT 1
+#    define _ONEDPL_SYCL_UNIFIED_USM_BUFFER_PRESENT 1
```
MikeDvorskiy, line 88: What if a device doesn't  support  USM?

julianmi: The existing code is unchanged where we check if the device supports USM and fall back to a buffer (https://github.com/oneapi-src/oneDPL/blob/main/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h#L528-L541). The main change here is that older compilers don't support the unified approach in `__usm_host_or_buffer_accessor` at all and we use a regular SYCL accessor instead.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -560,6 +564,17 @@ struct __usm_host_or_buffer_storage
     }
 };
 
+template <typename _Acc>
+auto
+__get_usm_host_or_buffer_accessor_ptr(const _Acc& __acc)
```
MikeDvorskiy, line 569: I would suggest to encapsulate this global function to `struct __usm_host_or_buffer_storage` scope (as a static method f.e)

MikeDvorskiy (commented):
Overall looks good, excepting one minor thing - see an in-place comment in the changes.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: I've made this change. One drawback of this I see is that it convolutes the call a bit.

MikeDvorskiy (approved):
LGTM

PR #1413, opened 2024-02-20
Bugfix for reverse_iterators as oneDPL input with device policy

danhoeflinger:
This PR switches the order of the iterators when processing the base iterator range for reverse_iterators.
If we do not do this, the next layer of handling for the input iterator will have incorrect ordering, and it will fail the assertion that `__first < __last` for the base.
This is a regression as of #1254, where a specialization was added specifically for reverse_iterators (However prior to that PR, it reverse_iterators may have been handled in a inefficient way).

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -481,7 +481,7 @@ struct __get_sycl_range
     {
         assert(__first < __last);
 
-        auto __res = __process_input_iter<_LocalAccMode>(__first.base(), __last.base());
+        auto __res = __process_input_iter<_LocalAccMode>(__last.base(), __first.base());
         auto __rng = oneapi::dpl::__ranges::reverse_view_simple<decltype(__res.all_view())>{__res.all_view()};
 
         return __range_holder<decltype(__rng)>{__rng};


danhoeflinger:
@SergeyKopienko Can you advise the best way to add some testing of reverse iterator in to dpcpp backend testing (without multiplying the runtime of all tests by a large factor)?

It seems we have the infrastructure with `invoke_on_all_iterator_types`, but it is only utilized for host policies.  This seems like a bug we should have caught in testing.

Perhaps it makes sense to have a unit test targeted specifically at handling different input types and covering `__get_sycl_range` combinations using a simple lightweight API call and a device policy.

mmichel11 (approved):
I confirmed this fixes the reproducer we have.

LGTM

MikeDvorskiy (approved):
LGTM

danhoeflinger:
> @SergeyKopienko Can you advise the best way to add some testing of reverse iterator in to dpcpp backend testing (without multiplying the runtime of all tests by a large factor)?
> 
> It seems we have the infrastructure with `invoke_on_all_iterator_types`, but it is only utilized for host policies. This seems like a bug we should have caught in testing.
> 
> Perhaps it makes sense to have a unit test targeted specifically at handling different input types and covering `__get_sycl_range` combinations using a simple lightweight API call and a device policy.

I've captured this in a new issue #1414, in preparation to merge this PR.

PR #1422, opened 2024-02-27
Lazy `sycl::queue` creation in oneDPL hetero policies

SergeyKopienko:
# Attention: ABI breaking change.

Yet another approach to lazy `sycl::queue` creation in oneDPL hetero policies:
- the source issue described at https://github.com/oneapi-src/oneDPL/issues/1060
- the previous approach has been implemented by @adamfidel in the PR https://github.com/oneapi-src/oneDPL/pull/1154

The main implementation detail is `std::once_flag` and `::std::call_once` usage packed into separate container
```C++
template <class TSYCLQueueFactory>
class sycl_queue_container
{
  public:
    template <typename... Args>
    sycl::queue
    get_queue(Args&&... args)
    {
        ::std::call_once(__is_created, [&]() {
            TSYCLQueueFactory factory;
            __queue.emplace(factory(::std::forward<Args>(args)...));
        });

        assert(__queue.has_value());
        return __queue.value();
    }

  private:
    ::std::once_flag __is_created;
    ::std::optional<sycl::queue> __queue;
};
```

This container shares between hetero policies instances:
```C++
template <typename TFactory>
using sycl_queue_container_ptr = ::std::shared_ptr<sycl_queue_container<TFactory>>;


// ...


template <typename KernelName = DefaultKernelName, class TSyclQueueFactory = TDefaultSYCLQueueFactory>
class device_policy
{
  public:
    // ...

  private:
    mutable sycl_queue_container_ptr<TSyclQueueFactory> q_container;       // <<<<<<-- shared sycl::queue container
};
```

From my point of view this approach is correct because it's simple repeat something like `sycl::queue` implementation ([pimpl-implementation](https://www.google.com/search?q=pimpl+implementation&rlz=1C1GCEA_en__1012__1012&oq=pimpl+implementation&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIJCAEQABgTGIAE0gEJNTAzOWowajE1qAIAsAIA&sourceid=chrome&ie=UTF-8)) and we don't introduce any new synchronization errors here.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -23,6 +23,8 @@
 #include "sycl_defs.h"
 
 #include <type_traits>
+#include <mutex>
+#include <optional>
 
 namespace oneapi
 {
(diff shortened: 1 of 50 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -34,29 +36,75 @@ inline namespace __dpl
...
+    }
+};
+
 //We can create device_policy object:
 // 1. from sycl::queue
 // 2. from sycl::device_selector (implicitly through sycl::queue)
 // 3. from sycl::device
 // 4. from other device_policy encapsulating the same queue type
-template <typename KernelName = DefaultKernelName>
+template <typename KernelName = DefaultKernelName, class TSyclQueueFactory = sycl_queue_factory_device>
```
kboyarinov, line 82: As far as I can see in [oneDPL specification](https://spec.oneapi.io/versions/latest/elements/oneDPL/source/parallel_api.html#dpc-execution-policy) - `device_policy` class is a public API and hence its template parameters are also documented. So do we think that it is necessary to have `TSyclQueueFactory` template parameter as part of public API?

SergeyKopienko: In this valiant - probably yes.

PR #1425, opened 2024-02-28
Fixing static_assert(false) bug

danhoeflinger:
This corrects a bug introduced by #1400. 
`static_assert(false,...` can break some compilers even if the code should be excluded at compile time via a constexpr. The static assertion should stop the build with an appropriate message, and removes the need for the constexpr avoidance of the __parallel_histogram (which can end up with a cryptic error message from the fpga compiler).

This fix was taken from #1239 , but separated as a bugfix from that large PR.


SergeyKopienko (approved):
LGTM

SergeyKopienko (approved):
LGTM

PR #1426, opened 2024-02-29
Sort sycl::half with radix sort

dmitriy-sobolev:
The PR adds:

-  A conversion to an ordered bit representation to sort `sycl::half` with radix sort algorithm.
- A test case for `sycl::half` elements.


danhoeflinger (approved):
LGTM, 
I don't see any issues. 
(I prefer your formatting to clang-format, so I agree the suggestion should be ignored)

julianmi (approved):
LGTM

PR #1383, opened 2024-01-25
Three fixes for permutation_iterator

danhoeflinger:
Three fixes for permutation_iterator:

1. Legacy `is_hetero` trait #1380 (for the sake of time, closing that PR in favor of this one combined)
2. Fix for `is_passed_directly` legacy trait to be usable recursively
3. Restore "general case" with recursion with size of 1 for base iterator of a `permutation_iterator`

PR #1429, opened 2024-03-01
[tests] Add Input Data Sweep tests

danhoeflinger:
Note: Should be merged after #1445 and #1438(merged), as errors can occur otherwise.

Summary: 
Adding test coverage to catch bugs which slipped through our previous testing by targeting a sweep of interesting combinations of input data. Would have detected the following bugs which were missed by our test coverage at the time:

#1413 (f69618b4e) reverse_iterator: detects as a runtime failure.
#1371 (286097b43) zip_iterator device copyable: detects as a build error
#1383 (a860b5d1b) 3 fixes for permutation iterator: detects as build error
#1341 (fef095835) "general case" with recursion of permutation iterator: detects as build error
#1325 (6d50dbad9) broken bracket operator permutation iterator: detects as build error

---
Details:
The test does not specifically target these bugs listed above, but rather identifies the target base data, wrappers, and recurses to generate a large number of possible iterator types.  It runs these resulting iterator types through a minimal kernel (copy of 10 elements) as both a read and write and checks the result.  Some types and wrappers disable tests or aspects of the resulting tests for deeper levels of recursion. 

Base Types:
1. `float` (recurse 0) 
2. `double` (recurse 0)
3. `uint64_t` (recurse 0)
4. `int32_t` (recurse 2 layers of wrappers)
 
Sequence types:
1. `host_iterator`
2. `sycl_iterator`
3. `counting_iterator` (for integral types)
4. `usm_shared`
5. `usm_device`
6. `std::vector<T,usm_allocator<T,sycl::usm::alloc::shared>>::iterator` 
7. `std::vector<T,usm_allocator<T,sycl::usm::alloc::host>>::iterator`

Wrappers:
1. `std::reverse_iterator(it)`
2. `transform_iterator(it, noop{})`
3. `permutation_iteartor(it,noop{})`
4. `permutation_iterator(it, counting_iter)`
5. `permutation_iterator(counting_iterator, it)`
8. `zip_iterator(counting_iterator, it)`
9. `zip_iterator(it, discard_iterator)`

It also explores a couple special cases each sent through 1 level of recursion:
1. `discard_iterators` 
2. `permutation_iterator( permutation_iterator( usm_shared<int>, counting_iterator), counting_iterator)` 

---
This PR catches 2 issues
1)  build error for `zip_iterator<it, transform_iterator<it,lambda>` inputs, which is fixed by #1445.

2) Inefficient and potentially incorrect handling of `std::vector<T,usm_allocator>::iterator`.  This has been addressed by #1438.
Note: Depending on implementation of `std::vector`, some `std::vector<T>::iterator` cannot be distinguished from `std::vector<T, usm_allocator>::iterator`.  For those cases, we must continue handling these types as if they were host_iterators.

Note:
I've disabled some combinations which don't make sense including `std::reverse_iterator(sycl_iterator based "iterators")` because `sycl_iterators` are not iterators.  In practice, those do seem to work on linux, but not windows.  There are build errors on windows when attempting to run the SFINAE trick for `__is_addressable`.


test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+#endif //TEST_DPCPP_BACKEND_PRESENT
+
+int
+main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+
+    constexpr size_t n = 10;
+
+    auto q = sycl::queue{sycl::default_selector_v};
```
SergeyKopienko, line 295: Please use `TestUtils::get_test_queue()`

test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+    //baseline with no wrapping
+    test<float, 0>(TestUtils::make_device_policy<class Kernel1>(q), -666.0f, n, "float");
+    test<double, 0>(TestUtils::make_device_policy<class Kernel2>(q), -666.0, n, "double");
```
SergeyKopienko, line 299: What if `double` type isn't supported on device?

test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+    oneapi::dpl::counting_iterator<int> counting(0);
+    oneapi::dpl::discard_iterator discard{};
+    wrap_recurse<1, 0, /*__read =*/false, /*__reset_read=*/false, /*__write=*/true,
+                 /*__check_write=*/false, /*__usable_as_perm_map=*/false, /*__usable_as_perm_src=*/true,
+                 /*__is_reversible=*/true>(TestUtils::make_device_policy<class Kernel5>(q), discard, discard + n,
+                                           counting, discard, discard, discard, discard, -666, "discard_iterator");
+
+    //recurse once on perm(perm(usm_shared<int>,count), count)
+    auto policy = TestUtils::make_device_policy<class Kernel6>(q);
+    auto copy_out = sycl::malloc_shared<int>(n, policy.queue());
```
SergeyKopienko, line 317: Who will delete this allocated memory?

test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+    //discard iterator
+    oneapi::dpl::counting_iterator<int> counting(0);
+    oneapi::dpl::discard_iterator discard{};
+    wrap_recurse<1, 0, /*__read =*/false, /*__reset_read=*/false, /*__write=*/true,
+                 /*__check_write=*/false, /*__usable_as_perm_map=*/false, /*__usable_as_perm_src=*/true,
+                 /*__is_reversible=*/true>(TestUtils::make_device_policy<class Kernel5>(q), discard, discard + n,
+                                           counting, discard, discard, discard, discard, -666, "discard_iterator");
+
+    //recurse once on perm(perm(usm_shared<int>,count), count)
+    auto policy = TestUtils::make_device_policy<class Kernel6>(q);
```
SergeyKopienko, line 316: What about fpga policy?

test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+            return exp;
+        }
+        else
+        {
+            return std::make_reverse_iterator(exp + n);
+        }
+    };
+
+#    if VERBOSE_TEST
+    std::cout << input_descr.c_str() << ":";
```
SergeyKopienko, line 71: `.c_str()` isn't required here I think,

test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+                                                std::string("sycl_iterator<") + type_text + std::string(">"));
+
+        //usm_shared
+        auto usm_shared = sycl::malloc_shared<T>(n, device_policy.queue());
+        //test all modes / wrappers
+        wrap_recurse<__recurse, 0>(device_policy4, usm_shared, usm_shared + n, counting, copy_out, usm_shared, copy_out,
+                                   counting, trash, std::string("usm_shared<") + type_text + std::string(">"));
+
+        //usm_device
+        auto usm_device = sycl::malloc_device<T>(n, device_policy.queue());
```
SergeyKopienko, line 275: The same

test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+
+// Enable for verbose list of tests run
+//#define VERBOSE_TEST 1
+
+//This test is written without indirection from invoke_on_all_hetero_policies to make clear exactly which types
+// are being tested, and to limit the number of types to be within reason.
+
+#if TEST_DPCPP_BACKEND_PRESENT
+template <int __recurse, int __reverses, bool __read = true, bool __reset_read = true, bool __write = true,
+          bool __check_write = true, bool __usable_as_perm_map = true, bool __usable_as_perm_src = true,
```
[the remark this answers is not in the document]
danhoeflinger: I believe you are allowed to treat template types and values independently, and can have defaults like this without issue.  
Yes, it would perhaps be better to collapse and pack these into a single value.  It has grown quite large, I'll see what I can come up with.

test/parallel_api/iterator/input_data_sweep.pass.cpp
danhoeflinger: done.

test/parallel_api/iterator/input_data_sweep.pass.cpp
danhoeflinger: I check for this within `test`, using `TestUtils::has_types_support`, and skip types which are unsupported.

test/parallel_api/iterator/input_data_sweep.pass.cpp
danhoeflinger: Good point, fixed.

test/parallel_api/iterator/input_data_sweep.pass.cpp
danhoeflinger: Fixed.  I created a new helper to use to get create a test policy based on backend, and use it here and in `invoke_on_all_hetero_policies`.

test/parallel_api/iterator/input_data_sweep.pass.cpp
danhoeflinger: Yes, thanks. Fixed.

test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,330 @@
...
+        sycl::buffer<T> buf(n);
+        //test all modes / wrappers
+        wrap_recurse<__recurse, 0, /*__read =*/true, /*__reset_read=*/true, /*__write=*/true,
+                     /*__check_write=*/true, /*__usable_as_perm_map=*/true, /*__usable_as_perm_src=*/false,
+                     /*__is_reversible=*/false>(device_policy3, oneapi::dpl::begin(buf), oneapi::dpl::end(buf),
+                                                counting, copy_out, oneapi::dpl::begin(buf), copy_out, counting, trash,
+                                                std::string("sycl_iterator<") + type_text + std::string(">"));
+
+        //usm_shared
+        auto usm_shared = sycl::malloc_shared<T>(n, device_policy.queue());
```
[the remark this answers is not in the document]
danhoeflinger: Fixed.

test/parallel_api/iterator/input_data_sweep.pass.cpp
danhoeflinger: Fixed

test/support/utils_invoke.h
```diff
@@ -157,6 +157,17 @@ inline void unsupported_types_notifier(const sycl::device& device)
     }
 }
 
+template<typename KernelName>
+auto
+create_test_policy(sycl::queue _queue)
```
SergeyKopienko, line 162: Are we able to  use already existing `make_new_policy` from `TestUtils` namespace?

test/parallel_api/iterator/input_data_sweep.pass.cpp
SergeyKopienko: ok

test/parallel_api/iterator/input_data_sweep.pass.cpp
SergeyKopienko: ok

test/parallel_api/iterator/input_data_sweep.pass.cpp
SergeyKopienko: ok

test/support/utils_invoke.h
danhoeflinger: I looked at that function, but it requires a policy to already exist, and just creates one from it.  The code which originally creates the policy was embedded within `TestUtils::invoke_on_all_hetero_policies`, but now it lives here.
We could consolidate these, and provide an overload for a queue perhaps.  Would you prefer that?

test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,348 @@
...
+    // discard iterator
+    oneapi::dpl::counting_iterator<int> counting(0);
+    oneapi::dpl::discard_iterator discard{};
+    wrap_recurse<1, 0, /*__read =*/false, /*__reset_read=*/false, /*__write=*/true,
+                 /*__check_write=*/false, /*__usable_as_perm_map=*/false, /*__usable_as_perm_src=*/true,
+                 /*__is_reversible=*/true>(policy5, discard, discard + n,
+                                           counting, discard, discard, discard, discard, -666, "discard_iterator");
+
+    // recurse once on perm(perm(usm_shared<int>,count), count)
+    auto copy_out = sycl::malloc_shared<int>(n, policy6.queue());
```
SergeyKopienko, line 333: You can try to use `usm_data_transfer` class - it's already has `alloc` and `free` calls

test/support/utils_invoke.h
SergeyKopienko: Up to you.

test/parallel_api/iterator/input_data_sweep.pass.cpp
danhoeflinger: I've switched to using that class for the data allocation / freeing.  Thanks.

SergeyKopienko: ok

test/parallel_api/iterator/input_data_sweep.pass.cpp
```diff
@@ -0,0 +1,348 @@
...
+
+        //Run test
+        oneapi::dpl::copy(exec3, first, last, copy_to_first);
+
+        //get expected sequence with proper number of reverses
+        auto expect = get_expect(expected_first);
+        std::string msg = std::string("wrong read effect from ") + input_descr;
+        //verify result using original unwrapped output
+        EXPECT_EQ_N(expect, orig_out_first, n, msg.c_str());
+#    if VERBOSE_TEST
```
SergeyKopienko, line 91: Probably you may use `_ONEDPL_DEBUG_SYCL` instead of this define.

test/support/utils_invoke.h
danhoeflinger: I've added an overload of `make_new_policy` for queue inputs, and relocated them so they can live side by side.

test/parallel_api/iterator/input_data_sweep.pass.cpp
danhoeflinger: done

SergeyKopienko: ok

test/support/utils_invoke.h
SergeyKopienko: ok

test/parallel_api/iterator/input_data_sweep.pass.cpp
SergeyKopienko: ok

test/parallel_api/iterator/input_data_sweep.pass.cpp
SergeyKopienko: ok

test/parallel_api/iterator/input_data_sweep.pass.cpp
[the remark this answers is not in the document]
danhoeflinger: Do you have a good suggestion for cleaning this up without over-engineering or obfuscating the tests? 

It needs to be compile time, the configuration is cumulative (more an more parts of the test may get turned of with layers of recursion), and its not necessarily easy to pull apart the final wrapped iterators to get to their fundamental sequence / important wrappers to determine these values from the types themselves.

I could group them into a single struct, but that doesn't change much.  Its ugly as is, but I don't think its confusing especially with the comments.  I'm inclined to leave it as is unless there is a more specific suggestion of how to improve it.

test/parallel_api/iterator/input_data_sweep_host_iter.pass.cpp
```diff
@@ -0,0 +1,82 @@
...
+void
+test(Policy&& policy, T trash, size_t n, const std::string& type_text)
+{
+    if (TestUtils::has_types_support<T>(policy.queue().get_device()))
+    {
+
+        TestUtils::usm_data_transfer<sycl::usm::alloc::shared, T> copy_out(policy.queue(), n);
+        oneapi::dpl::counting_iterator<int> counting(0);
+        // host iterator
+        std::vector<T> host_iter(n);
```
mmichel11, line 40: In addition to having a `std::vector` host iterator test, would it be useful you think to have a `std::vector` with USM allocator test as well?

test/parallel_api/iterator/input_data_sweep_host_iter.pass.cpp
```diff
@@ -0,0 +1,82 @@
...
+    auto q = TestUtils::get_test_queue();
+
+    auto policy = TestUtils::make_new_policy<class Kernel1>(q);
+
+    auto policy1 = TestUtils::create_new_policy_idx<decltype(policy), 0>(policy);
+    auto policy2 = TestUtils::create_new_policy_idx<decltype(policy), 1>(policy);
+    auto policy3 = TestUtils::create_new_policy_idx<decltype(policy), 2>(policy);
+    auto policy4 = TestUtils::create_new_policy_idx<decltype(policy), 3>(policy);
+
+    // baseline with no wrapping
```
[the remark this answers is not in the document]
danhoeflinger: Yeah, my original intention was to include this as well for the reason you describe, but the test would need to be specialized for this case (to not use counting_iterator as source & verification of sink). 
I'll see if it is reasonable to add without adding too much complexity.

test/parallel_api/iterator/input_data_sweep_host_iter.pass.cpp
danhoeflinger: As far as coverage of input data processing code, I believe this ends up being equivalent to USM pointers in practice. 
I will confirm this, and if so, I'd prefer not to add it to keep the number of permutations down.  If there is some different code invoked as a result of such a case, we should include it.

test/parallel_api/iterator/input_data_sweep_host_iter.pass.cpp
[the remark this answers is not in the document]
danhoeflinger: I've added boolean types for host iterators.  Thanks.

test/parallel_api/iterator/input_data_sweep_host_iter.pass.cpp
danhoeflinger: I've added tests to cover this, and in the process uncovered that `std::vectors` `usm_allocator` were being handled inefficiently by our input data processing.  This should be rectified with  #1438.

test/parallel_api/iterator/input_data_sweep_host_iter.pass.cpp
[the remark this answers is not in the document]
danhoeflinger: After discussion with @MikeDvorskiy , I have learned that `std::vector<bool>::iterator` for any allocator is undefined behavior as input of oneDPL due to the optimizations standard library implementations add to turn it into a bitfield.  This definitely doesn't work for host policies as the data is used directly and elements occupy an bit in a bitfield rather than a whole type.  In `icpx`, this does technically work with a device policy and `dpcpp` backend (due to the copy step into a `sycl::buffer`).  

However, since this depends on standard library implementation and policy, we determined it is best to just describe it as undefined behavior.  I've added #1443 to better document this.

test/parallel_api/iterator/input_data_sweep_usm_allocator.pass.cpp
```diff
@@ -0,0 +1,127 @@
...
+        run_tests = true;
+        test<std::uint64_t, 0>(policy3, 999, n, "uint64_t");
+    }
+    if (oneapi::dpl::usm_allocated_vector_iterators_supported_v<int32_t>)
+    {
+        run_tests = true;
+        // big recursion step: 1 and 2 layers of wrapping
+        test<std::int32_t, 2>(policy4, -666, n, "int32_t");
+    }
+    }
```
mmichel11, line 124: I think there is an extra `}` here.

danhoeflinger: It looks like you are right, so the more problematic thing is then, how are the tests passing...?

I'll take a look.

mmichel11: I was able to catch this when testing one of my branches with this PR. In the per-commit CI with the DPCPP backend, I don't think this test is ever ran which is why it is not caught.

test/parallel_api/iterator/input_data_sweep_usm_allocator.pass.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+#if TEST_DPCPP_BACKEND_PRESENT
+
+template <typename T, int __recurse, typename Policy>
+void
+test(Policy&& policy, T trash, size_t n, const std::string& type_text)
+{
+    if (TestUtils::has_types_support<T>(policy.queue().get_device()))
+    {
+        { //std::vector using usm shared allocator
```
mmichel11, line 36: I think we need to make sure the kernels for the std::vector with usm shared allocator test end up with a different kernel name than the kernels for usm host allocator test.

I see a duplicate kernel name issue with no unnamed lambda compilation.

test/parallel_api/iterator/input_data_sweep_usm_allocator.pass.cpp
danhoeflinger: Ah, i see.  we only run the sycl_iterator tests in the per commit CI for dpcpp.  OK, makes sense. 
Thanks for catching this.

test/parallel_api/iterator/input_data_sweep_usm_allocator.pass.cpp
danhoeflinger: fixed.  thanks.

test/parallel_api/iterator/input_data_sweep_usm_allocator.pass.cpp
danhoeflinger: fixed. Thanks.  I've also rebased after #1438, which enables this test to succeed on linux.  I've removed the checks of the public utility as that was separated out from the mentioned PR.

mmichel11:
Could you rebase this branch with main, so we can ensure everything works well after the merge of PR #1445? I'm ready to approve after.

danhoeflinger:
> Could you rebase this branch with main, so we can ensure everything works well after the merge of PR #1445? I'm ready to approve after.

Good point... done.

mmichel11 (approved):
LGTM

mmichel11 (approved):

mmichel11 (dismissed):
The latest changes LGTM.

PR #1433, opened 2024-03-04
Renaming licensing directory to follow OSSF best practices

timmiesmith:
A review of oneDPL using the OSSF scorecard showed that the oneDPL licenses are not in a standard location that makes it easily discoverable.  This PR resolves that issue by renaming licensing to LICENSES.
See https://github.com/ossf/scorecard/blob/main/docs/checks.md#license for details.


LICENSES/LICENSE.txt
danhoeflinger, line 1: Should we rename this file to `Apache-2.0.txt` or `Apache-2.0-LICENSE.txt` to go along with this blurb in the OSSF scorecard details?  

"Files in a LICENSES directory are typically named as their [SPDX](https://spdx.org/licenses/) license identifier followed by an appropriate file extension, as described in the [REUSE](https://reuse.software/spec/) Specification"

It seems like we need it to claim the 1 point for FSF license.
Is it appropriate to do that with the "LLVM Exceptions to the Apache 2.0 License" section?

danhoeflinger, line 1: It seems like we may only score 6/10 with this change since 3 points are based on the license being in top-level directory.  Its unclear if the alternative of the `LICENSES` directory being in the top level directory satisfies that.

danhoeflinger: Reading more, it looks like the right name for it is:
"Apache-2.0 WITH LLVM-exception.txt"

Although I hate introducing a name with spaces in the filename...

Its possible "Apache-2.0-WITH-LLVM-exception.txt" is also valid, but I can't really find the rules here.

dnmokhov (approved):

LICENSES/LICENSE.txt
timmiesmith: Good catch on the score.  A 6/10 isn't sufficient, and use of the LICENSES directory won't satisfy the top-level requirement. I'll close this PR and rework the licenses to what's required.

PR #1438, opened 2024-03-07
Adding std::vector<T, usm_allocator>::iterator to is_passed_directly types

danhoeflinger:
It seems that `std::vector` with `usm_allocator` with `sycl::usm::alloc::shared` data were being processed as if they were host_iterators when being passed to oneDPL with device policy and dpcpp backend.  

This means that they were being used to initialize a `sycl::buffer` which includes an extra copy to intermediate host memory.  This is unnecessary because iterators to such data should be passed directly, and directly usable within a kernel.  

This PR adds a specialization for the is_passed_directly trait, which enables such iterators to be properly passed directly.  This avoids the extra copy.  

This was uncovered in investigation #1429, and will be tested by those tests, after a suggestion from @mmichel11 .

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -78,11 +78,7 @@ class all_view
     {
         return begin() + size();
     }
-    __return_t&
-    operator[](__diff_type i) const
-    {
-        return begin()[i];
-    }
+    __return_t& operator[](__diff_type i) const { return begin()[i]; }
 
     __diff_type
     size() const
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -206,6 +206,7 @@ struct is_passed_directly<Iter, ::std::enable_if_t<Iter::is_passed_directly::val
 {
 };
 
+//support std::vector::iterator with usm shared allocator as passed directly
 template <class Iter>
 struct is_passed_directly<
     Iter,
@@ -213,6 +214,18 @@ struct is_passed_directly<
         Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
                                      typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
                                                                   sycl::usm::alloc::shared>::iterator>>>>
+    : ::std::true_type
+{
+};
+
+//support std::vector::iterator with usm host allocator as passed directly
+template <class Iter>
+struct is_passed_directly<
+    Iter,
+    ::std::enable_if_t<::std::is_same_v<
+        Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
+                                     typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
+                                                                  sycl::usm::alloc::host>::iterator>>>>
     : ::std::true_type
 {
 };
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -213,7 +213,7 @@ struct is_passed_directly<
     ::std::enable_if_t<::std::is_same_v<
         Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
                                      typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
-                                                                  sycl::usm::alloc::shared>::iterator>>>>
+                                                                  sycl::usm::alloc::shared>>::iterator>>>
     : ::std::true_type
 {
 };
@@ -225,7 +225,7 @@ struct is_passed_directly<
     ::std::enable_if_t<::std::is_same_v<
         Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
                                      typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
-                                                                  sycl::usm::alloc::host>::iterator>>>>
+                                                                  sycl::usm::alloc::host>>::iterator>>>
     : ::std::true_type
 {
 };
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -78,7 +78,11 @@ class all_view
     {
         return begin() + size();
     }
-    __return_t& operator[](__diff_type i) const { return begin()[i]; }
+    __return_t&
+    operator[](__diff_type i) const
+    {
+        return begin()[i];
+    }
 
     __diff_type
     size() const
@@ -206,26 +210,19 @@ struct is_passed_directly<Iter, ::std::enable_if_t<Iter::is_passed_directly::val
 {
 };
 
-//support std::vector::iterator with usm shared allocator as passed directly
+//support std::vector::iterator with usm host / shared allocator as passed directly
 template <class Iter>
 struct is_passed_directly<
     Iter,
-    ::std::enable_if_t<::std::is_same_v<
-        Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
-                                     typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
-                                                                  sycl::usm::alloc::shared>>::iterator>>>
-    : ::std::true_type
-{
-};
-
-//support std::vector::iterator with usm host allocator as passed directly
-template <class Iter>
-struct is_passed_directly<
-    Iter,
-    ::std::enable_if_t<::std::is_same_v<
-        Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
-                                     typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
-                                                                  sycl::usm::alloc::host>>::iterator>>>
+    ::std::enable_if_t<
+        ::std::is_same_v<
+            Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
+                                         typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
+                                                                      sycl::usm::alloc::shared>>::iterator> ||
+        ::std::is_same_v<
+            Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
+                                         typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
+                                                                      sycl::usm::alloc::host>>::iterator>>>
     : ::std::true_type
 {
 };
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -206,6 +206,17 @@ struct is_passed_directly<Iter, ::std::enable_if_t<Iter::is_passed_directly::val
 {
 };
 
+template <class Iter>
+struct is_passed_directly<
+    Iter,
+    ::std::enable_if_t<::std::is_same_v<
+        Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
+                                     typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
+                                                                  sycl::usm::alloc::shared>::iterator>>>>
+    : ::std::true_type
+{
+};
+
 template <typename Ip>
 struct is_passed_directly<oneapi::dpl::counting_iterator<Ip>> : ::std::true_type
 {


mmichel11:
Our documentation only makes mention of the case where we use a `sycl::usm_allocator` with `sycl::usm::alloc::shared`, but should we also provide a specialization for  `sycl::usm_allocator` with `sycl::usm::alloc::host` to be passed directly since SYCL host memory is device accessible?

danhoeflinger:
> Our documentation only makes mention of the case where we use a `sycl::usm_allocator` with `sycl::usm::alloc::shared`, but should we also provide a specialization for `sycl::usm_allocator` with `sycl::usm::alloc::host` to be passed directly since SYCL host memory is device accessible?

Its a good question, I will need to look into it some.  @MikeDvorskiy @rarutyun @akukanov do you have an opinion on how we should handle data allocated with a  `sycl::usm_allocator` with `sycl::usm::alloc::host` memory?

danhoeflinger:
> > Our documentation only makes mention of the case where we use a `sycl::usm_allocator` with `sycl::usm::alloc::shared`, but should we also provide a specialization for `sycl::usm_allocator` with `sycl::usm::alloc::host` to be passed directly since SYCL host memory is device accessible?
> 
> Its a good question, I will need to look into it some. @MikeDvorskiy @rarutyun @akukanov do you have an opinion on how we should handle `std::vector` input data allocated with a `sycl::usm_allocator` with `sycl::usm::alloc::host` memory?

After some thought, I believe we should support `sycl::usm::alloc::host` allocators as well as passed directly.  The alternative is them being treated as host iterators, and being used to initialize a `sycl::buffer`.  Of course, it is the user's responsibility to use host usm memory responsibly.  
I am interested in hearing other opinions on this as well.

danhoeflinger:
When run through the input data sweep tests from #1429, it seems that both `shared` and `host` `usm_allocator` iterators encounter issues when used as the source iterator for a `permutation_iterator`.  I don't yet understand why this causes issues, but should be understood before merging this PR.

danhoeflinger:
> When run through the input data sweep tests from #1429, it seems that both `shared` and `host` `usm_allocator` iterators encounter issues when used as the source iterator for a `permutation_iterator`. I don't yet understand why this causes issues, but should be understood before merging this PR.

There was a bug of miss-placed `>` which was causing these usm_allocator iterators to not be treated as "passed_directly".  It is now fixed, and the tests from #1429 now pass, and I've confirmed that they are handled via the correct path through the data input processing code.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -206,6 +206,23 @@ struct is_passed_directly<Iter, ::std::enable_if_t<Iter::is_passed_directly::val
...
+    Iter,
+    ::std::enable_if_t<
+        ::std::is_same_v<
+            Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
+                                         typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
+                                                                      sycl::usm::alloc::shared>>::iterator> ||
+        ::std::is_same_v<
+            Iter, typename ::std::vector<typename ::std::iterator_traits<Iter>::value_type,
+                                         typename sycl::usm_allocator<typename ::std::iterator_traits<Iter>::value_type,
+                                                                      sycl::usm::alloc::host>>::iterator>>>
```
MikeDvorskiy, line 221: Probably, should be a `sycl::usm::alloc::device` ?

danhoeflinger: This doesn't exist. sycl doesnt define a device usm_allocator.

danhoeflinger:
@MikeDvorskiy  @mmichel11   Actually there is a fundamental problem with this, and it will not work.  This breaks host_iterators because `std::vector::iterator` with a default allocator are indistinguishable from `std::vector_iterator` with a usm_allocator.

This is problematic though, because as it stands (without this PR), we have an extra copy to a host memory location, and also these iterators are unable to be used as the source iterator for permutation iterators (the same as host_iterators).  Sadly, I see no apparent solution to this issue.

danhoeflinger:
Update: 
Actually, for icpx, the `std::vector::iterator` **_does_** distinguish between host_iterator and usm_allocator,  except for the case of `std::vector<bool>::iterator` (which is always `std::_Bit_iterator`, regardless of the allocator it seems).

I've provided a solution which should hopefully allow support when it is possible to, without breaking host_iterators.

It first checks if we can distinguish between host_iterator and usm_allocator using a new struct `__vector_impl_distinguishes_usm_allocator_from_default`.  If we can distinguish between these types for the value type in question, we do and mark passed directly if we can detect usm_allocators.  If we cannot distinguish between the two, we do not change the passed directly status.  
This should not "spoil" any host_iterators, and it should allow better support for usm_allocators where it is possible.

@MikeDvorskiy Please take a look and see if you can find problems with this approach.

danhoeflinger:
Following up on this strategy of first checking if `usm allocator` is distinguishable vs the default allocator within the iterator type...

It is possible (though I believe improbable) for an implementation to have the same type of `iterator` for vector with `usm allocator` as it has for vector with some `custom host allocator`, while at the same time having different iterator types for `usm allocator` vs the default allocator.  
In my view, it is quite likely that if the `iterator` types are distinguishable, they are by including the allocator in the definition of the iterator type (as is done for icpx outside of `vector<bool>`).  

However, if we want to be most conservative, this is not a 100% conclusive check that the iterator points to data from a `usm allocator `.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -206,6 +225,24 @@ struct is_passed_directly<Iter, ::std::enable_if_t<Iter::is_passed_directly::val
 {
 };
 
+//support std::vector::iterator with usm host / shared allocator as passed directly
+template <class Iter>
+struct is_passed_directly<
```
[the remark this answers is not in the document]
danhoeflinger: Yes, this is a good idea.  I can make this change.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -45,6 +45,25 @@ __create_accessor(_BufferType& __buf, _DiffType __offset, _DiffType __n)
 
     return {__buf, sycl::range<1>(__n_acc), __offset};
 }
+
+// Evaluates to true if the provided type is an iterator with a value_type and if the implementation of a
+//  std::vector<value_type>::iterator can be distinguished from std::vector<value_type, usm_allocator>::iterator
```
MikeDvorskiy, line 50: It makes sense to add some explanation in the doc about that fact.
Lets create a reference issue.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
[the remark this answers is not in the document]
MikeDvorskiy: > (It will still require

I guess, no, will not require...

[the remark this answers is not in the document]
danhoeflinger: Ah, I see you accomplish this in the `enable_if_t` within `usm_vec_iterator`.  Let me think about it to see if this accomplishes the same result...

danhoeflinger:
@MikeDvorskiy After our offline discussion, I've updated the PR in 2 ways.  
1) Made it more readable by abstracting out definitions of std::vector::iterator of different types
2) Not only do we check if `std::vector<T>::iterator` is distinguishable from `std::vector<T, usm_shared_allocator>::iterator`, but we now check that all three are different: 
a)`std::vector<T>::iterator`
b)`std::vector<T, usm_shared_allocator>::iterator`
c)`std::vector<T, usm_host_allocator>::iterator`

With (2) it makes it extremely unlikely that the iterator does not include the type of the allocator.  

Ran this new version through the tests in #1429, and confirmed they are headed to the right processing.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -45,6 +45,36 @@ __create_accessor(_BufferType& __buf, _DiffType __offset, _DiffType __n)
...
+template <typename Iter, typename ValueType = typename ::std::iterator_traits<Iter>::value_type>
+using __usm_shared_alloc_vec_iter =
+    typename ::std::vector<ValueType, typename sycl::usm_allocator<ValueType, sycl::usm::alloc::shared>>::iterator;
+
+template <typename Iter, typename ValueType = typename ::std::iterator_traits<Iter>::value_type>
+using __usm_host_alloc_vec_iter =
+    typename ::std::vector<ValueType, typename sycl::usm_allocator<ValueType, sycl::usm::alloc::host>>::iterator;
+
+template <typename Iter>
+struct __vector_iter_distinguishes_by_allocator<
```
MikeDvorskiy, line 70: I guess we played it extra safe here, but let be so.

MikeDvorskiy (commented):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: Yes, but this provides more and better evidence that unique allocators (not defined by the standard) result in iterator types which are unique from each other.  All of it is done at compile time, so I don't see a downside.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,7 +81,19 @@ the buffer were created for the same queue. For example:
...
+Note: The ability to appropriately detect USM allocated ``std::vector::iterator`` depends
+on details of the C++ standard library implementation and what information about the
+allocator is included in the ``std::vector::iterator`` type definition. You can check
+if your C++ standard library implementation provides enough information by checking the
+value of ``oneapi::dpl::usm_allocated_vector_iterators_supported_v<ValueType>``. This will
+evaluate to ``true`` for C++ standard library implementations which allow oneDPL to support
+USM allocators with this ValueType and ``false`` otherwise. If usm_allocators are not
+supported with your C++ standard library, they will still function as inputs to oneDPL, but
+they will be treated as if they were host-side ``std::vector::iterator`` as described in
+the section below.
```
danhoeflinger, line 94: Its worth discussing if it is a good idea to expose this information to the end user, and if this is the best way to convey it.  

I don't think its possible to do better without forcing the user to mark the input data via some wrapper (which doesn't seem better or easier than using `std::vector::data()` to get a usm pointer instead.

danhoeflinger: Another necessary bit of information:  If we cannot detect USM allocated data, it is treated like a host iterator.  Host iterators are not supported as the source iterator for `permutation_iterator`, as it cannot be efficiently determined what range of data should be copied into a `sycl::buffer` for use on the device.  
Therefore, based on the C++ standard library implementation, some uses case may succeed or not succeed.  It seems problematic to not provide some programmatic way to detect which scenario the user is in (`usm_allocated_vector_iterators_supported_v`).  The other option would be to disallow any `std::vector::iterator` as source iterator for `permutation_iterator` with a device policy (rather than just "host-side iterator" as is stated in the docs currently).

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,7 +81,19 @@ the buffer were created for the same queue. For example:
     return 0;
   }
 
-Alternatively, use ``std::vector`` with a USM allocator. For example:
+Alternatively, use ``std::vector`` with a USM allocator.
+Note: The ability to appropriately detect USM allocated ``std::vector::iterator`` depends
+on details of the C++ standard library implementation and what information about the
+allocator is included in the ``std::vector::iterator`` type definition. You can check
+if your C++ standard library implementation provides enough information by checking the
+value of ``oneapi::dpl::usm_allocated_vector_iterators_supported_v<ValueType>``. This will
```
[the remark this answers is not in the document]
MikeDvorskiy: I would suggest the male another PR for that change, because introducing new piblic API stuff like `oneapi::dpl::usm_allocated_vector_iterators_supported_v` should be discussed and (approved) with oneDPL library architect.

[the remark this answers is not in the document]
danhoeflinger: Separated out to #1476, but kept some of the documentation to describe some nuance to the support of USM allocators.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,7 +81,16 @@ the buffer were created for the same queue. For example:
...
   }
 
-Alternatively, use ``std::vector`` with a USM allocator. For example:
+Alternatively, use ``std::vector`` with a USM allocator.
+
+Note: The ability to appropriately detect USM allocated ``std::vector::iterator`` depends
+on details of the C++ standard library implementation and what information about the
+allocator is included in the ``std::vector::iterator`` type definition. If USM allocated
+vector iterators are not detectable with your C++ standard library, they will still function
+as inputs to oneDPL, but they will be treated as if they were host-side
```
MikeDvorskiy, line 90: Probably, it makes sense to add the clarification that in case "USM allocated
vector iterators are not detectable with your C++ standard library" a user can use `vec.data()` instead of "vec.begin()". It guarantees data processing w/o host-device extra copying.

MikeDvorskiy (commented):
Looks good, with a couple of minor comments.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: Yes, this is a good idea.

danhoeflinger: After offline discussion, I've made these changes, but also separated out the documentation changes into another PR #1477.

MikeDvorskiy (approved):
LGTM

PR #1439, opened 2024-03-11
Out of Place Esimd Radix Sort and Tests

danhoeflinger:
- Adding out of place support to ESIMD radix sort [by key] (high level refactoring). 
- Adding tests to test out of place sort [by key].
- Removing some unnecessary extra kernel names (Odd vs Even should be the same kernel, just with different runtime params)

APIs are disambiguated between iterator and range interfaces using `enable_if` to check if a type is an iterator.  Documentation updates will be made in a separate PR.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
@@ -11,6 +11,7 @@
 #define _ONEDPL_KT_ESIMD_RADIX_SORT_H
 
 #include <cstdint>
+#include <type_traits>
 
 #include "../../pstl/hetero/dpcpp/utils_ranges_sycl.h"
 
--- a/include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
+++ b/include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
@@ -231,9 +231,6 @@ __onesweep_impl(sycl::queue __q, _RngPack1&& __input_pack, _RngPack2&& __virt_pa
             __has_values,
             __esimd_radix_sort_onesweep_by_key<std::decay_t<_RngPack2>, std::decay_t<_RngPack3>, _KernelName>,
             __esimd_radix_sort_onesweep<std::decay_t<_RngPack2>, std::decay_t<_RngPack3>, _KernelName>>>;
-    using _EsimdRadixSortCopyback = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-        ::std::conditional_t<__has_values, __esimd_radix_sort_onesweep_copyback_by_key<_KernelName>,
-                             __esimd_radix_sort_onesweep_copyback<_KernelName>>>;
 
     using _GlobalHistT = ::std::uint32_t;
     constexpr ::std::uint32_t __bin_count = 1 << __radix_bits;
@@ -376,7 +373,7 @@ sycl::event
 __radix_sort(sycl::queue __q, _RngPack1&& __pack_in, _RngPack2&& __pack_out, _KernelParam __param)
 {
     const auto __n = __pack_in.__keys_rng().size();
-    assert(__n > 1);
+    assert(__n > 0);
 
     // _PRINT_INFO_IN_DEBUG_MODE(__exec); TODO: extend the utility to work with queues
     constexpr auto __data_per_workitem = _KernelParam::data_per_workitem;
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -717,18 +717,18 @@ struct __spirv_target_conditional :
 // to determine SPIR-V targets.
 inline constexpr bool __is_spirv_target_v = __spirv_target_conditional<::std::true_type, ::std::false_type>::value;
 
-template <typename T, typename = void>
+template <typename _T, typename = void>
 struct __is_iterator_type : std::false_type
 {
 };
 
-template <typename T>
-struct __is_iterator_type<T, std::void_t<typename std::iterator_traits<T>::difference_type>> : std::true_type
+template <typename _T>
+struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::difference_type>> : std::true_type
 {
 };
 
-template <typename T>
-static constexpr bool __is_iterator_type_v = __is_iterator_type<T>::value;
+template <typename _T>
+static constexpr bool __is_iterator_type_v = __is_iterator_type<_T>::value;
 
 } // namespace __internal
 } // namespace dpl


test/kt/esimd_radix_sort_by_key_out_of_place.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+        oneapi::dpl::experimental::kt::esimd::radix_sort_by_key_out_of_place<isAscending, RadixBits>(
+            q, keys, values, keys_out, values_out, param)
+            .wait();
+    }
+
+    auto expected_first = oneapi::dpl::make_zip_iterator(std::begin(expected_keys), std::begin(expected_values));
+    std::stable_sort(expected_first, expected_first + size, CompareKey<isAscending>{});
+
+    std::string parameters_msg = ", n: " + std::to_string(size) + ", sizeof(key): " + std::to_string(sizeof(KeyT)) +
```
SergeyKopienko, line 61: If you use `std::strstream` you may to build this message without a lot of `std::to_string` calls.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -58,12 +52,25 @@ template <typename _KernelName, bool __is_ascending, ::std::uint8_t __radix_bits
...
+sycl::event
+__one_wg(sycl::queue __q, _RngPack1&& __pack_in, _RngPack2&& __pack_out, ::std::size_t __n)
+{
+    using _KeyT = typename ::std::decay_t<_RngPack1>::_KeyT;
     using _EsimRadixSortKernel =
         oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__esimd_radix_sort_one_wg<_KernelName>>;
 
     return __radix_sort_one_wg_submitter<__is_ascending, __radix_bits, __data_per_work_item, __work_group_size, _KeyT,
-                                         _EsimRadixSortKernel>()(__q, ::std::forward<_RngPack>(__pack), __n);
+                                         _EsimRadixSortKernel>()(__q, __pack_in, __pack_out, __n);
```
SergeyKopienko, line 73: Why you not use `std::forward` for pass `__pack_in` and `__pack_out` ?

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -266,33 +243,132 @@ __onesweep(sycl::queue __q, _RngPack&& __pack, ::std::size_t __n)
...
+    auto __tmp_pack = __create_temp_pack<_RngPack1>(__mem_holder, __n);
+    auto select_element = [] {
+        if constexpr (__in_place)
+            return 0;
+        else
+            return (__stage_count % 2);
+    };
+
+    auto __virt_pack1 = ::std::get<select_element()>(::std::tuple(__tmp_pack, __pack_out));
+    auto __virt_pack2 = ::std::get<select_element()>(::std::tuple(__pack_out, __tmp_pack));
```
adamfidel, line 366: This is an interesting way to statically select from one of two variables based on a constant expression. I did some testing on Godbolt and from what I can tell, it seems like these two lines would create 2 copies each of `__tmp_pack` and `__pack_out`: https://godbolt.org/z/nT1sMfMdv.

I'm not sure if these copies are important to avoid, but I thought it was an interesting experiment to test out.

danhoeflinger: See the response to Matt above, 
I'll experiment a bit in godbolt, and just make a constexpr selection function using const refs.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -266,33 +243,132 @@ __onesweep(sycl::queue __q, _RngPack&& __pack, ::std::size_t __n)
...
     __event_chain = __radix_sort_histogram_submitter<__is_ascending, __radix_bits, __hist_work_group_count,
                                                      __hist_work_group_size, _EsimdRadixSortHistogram>()(
-        __q, __pack.__keys_rng(), __mem_holder.__global_hist_ptr(), __n, __event_chain);
+        __q, __input_pack.__keys_rng(), __mem_holder.__global_hist_ptr(), __n, __event_chain);
 
     __event_chain = __radix_sort_onesweep_scan_submitter<__stage_count, __bin_count, _EsimdRadixSortScan>()(
         __q, __mem_holder.__global_hist_ptr(), __n, __event_chain);
 
-    for (::std::uint32_t __stage = 0; __stage < __stage_count; __stage++)
+    auto __submit_iteration = [&](auto __input_, auto __output_, auto __p_global_hist_, auto __p_group_hists_,
```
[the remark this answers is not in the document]
danhoeflinger: My intention is to reduce repetitive code to make more apparent what is actually different between these invocations of `__radix_sort_onesweep_submitter`, but I don't think it is required technically.  

Also, my changes unroll the first iteration of the submission before the loop, as its inputs / outputs are special cased so there are three redundant calls, rather than just two.

I can go back to something more similar to the previous way of doing it without the lambda.  I don't have a strong preference here.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -266,33 +243,132 @@ __onesweep(sycl::queue __q, _RngPack&& __pack, ::std::size_t __n)
...
     }
 
-    if constexpr (__stage_count % 2 != 0)
+    return __event_chain;
+}
+
+template <::std::uint8_t __radix_bits, ::std::uint16_t __data_per_work_item, ::std::uint16_t __work_group_size,
+          typename _RngPack, typename _MemHolder>
+void
+__allocate_temp(_MemHolder& __mem_holder, ::std::size_t __n)
```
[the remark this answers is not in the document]
danhoeflinger: I originally abstracted these two functions out to be separate functions to be able to reuse from multiple places, but then later unified out_of_place and in_place versions together anyway pulling that decision up the stack, so they don't even need to be separate functions.  I'll return them both to be within `__onesweep`

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -266,33 +243,132 @@ __onesweep(sycl::queue __q, _RngPack&& __pack, ::std::size_t __n)
...
+    constexpr ::std::uint32_t __bit_count = sizeof(_KeyT) * 8;
+    constexpr ::std::uint32_t __stage_count = oneapi::dpl::__internal::__dpl_ceiling_div(__bit_count, __radix_bits);
+
+    // Memory is not going to be allocated for void value type
+    // TODO: make this more explicit to reduce coupling between __onesweep_memory_holder and __rng_pack
+    __onesweep_memory_holder<_GlobalHistT, _KeyT, _ValT> __mem_holder(__q);
+
+    __allocate_temp<__radix_bits, __data_per_work_item, __work_group_size, _RngPack1>(__mem_holder, __n);
+    auto __tmp_pack = __create_temp_pack<_RngPack1>(__mem_holder, __n);
+    auto select_element = [] {
```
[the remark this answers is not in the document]
mmichel11: I think that sounds like the best solution.

Interestingly, the assignment to a constexpr variable with a ternary operator seems to work and evaluate at compile time even though there is no constexpr ternary operator (at least with gcc): https://godbolt.org/z/aGbjaE1KK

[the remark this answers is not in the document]
danhoeflinger: Interesting... 
In any case, I've updated it to be simpler and confirmed via that this works with no copies.  https://godbolt.org/z/Pe45WeYKT

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
danhoeflinger: I've simplified it away from `tuple` and confirmed this works without copies https://godbolt.org/z/Pe45WeYKT

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
danhoeflinger: fixed, thanks.

test/kt/esimd_radix_sort_by_key_out_of_place.cpp
danhoeflinger: fixed, thanks.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -81,7 +81,92 @@ radix_sort_by_key(sycl::queue __q, _KeysIterator __keys_first, _KeysIterator __k
...
+    return __impl::__radix_sort<__is_ascending, __radix_bits, /*__in_place=*/true>(__q, __pack, __pack, __param);
+}
+
+template <bool __is_ascending = true, ::std::uint8_t __radix_bits = 8, typename _KernelParam, typename _KeysRng1,
+          typename _KeysRng2>
+sycl::event
+radix_sort_out_of_place(sycl::queue __q, _KeysRng1&& __keys_rng, _KeysRng2&& __keys_rng_out, _KernelParam __param = {})
+{
+    __impl::__check_esimd_sort_params<__radix_bits, _KernelParam::data_per_workitem, _KernelParam::workgroup_size>();
+    if (__keys_rng.size() <= 0)
```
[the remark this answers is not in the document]
danhoeflinger: fixed, thanks.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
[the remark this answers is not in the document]
danhoeflinger: I've removed the lambda `__submit_iteration`.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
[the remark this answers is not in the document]
danhoeflinger: I've restored this to be within the `__onesweep` function.

dmitriy-sobolev (commented):
The PR looks good to me besides `_out_of_place` function name part, which is mentioned in the description.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -227,9 +234,80 @@ __onesweep(sycl::queue __q, _RngPack&& __pack, ::std::size_t __n)
...
+                __q, __virt_pack1, __virt_pack2, __p_global_hist, __p_group_hists, __sweep_work_group_count, __n,
+                __stage, __event_chain);
+        }
+        else
+        {
+            __event_chain = __radix_sort_onesweep_submitter<__is_ascending, __radix_bits, __data_per_work_item,
+                                                            __work_group_size, _EsimdRadixSortSweep>()(
+                __q, __virt_pack2, __virt_pack1, __p_global_hist, __p_group_hists, __sweep_work_group_count, __n,
+                __stage, __event_chain);
+        }
```
akukanov, line 276: Unless I miss something obvious, these two kernel submissions will not generate identical kernels, and so should really have distinct names (like with the Even/Odd suffixes before the patch).

danhoeflinger: Now in hindsight, I think you are correct, because the types of `__virt_pack1` and `__virt_pack2` may not match.

However, if they do match, do you agree that these would generate an identical kernel?

My proposal would be to merely include the types of the input & output in the kernel name, that way they aren't required to generate two identical kernels with different names when the types match.

danhoeflinger: I'm doing some testing with no unnamed lambda, and seeing the issue.  I'll provide an update when it is fixed.

danhoeflinger: I've fixed the issue as proposed by including the decayed RngPack typenames in the kernel name.  
I've also fixed kernel name clashing in the tests for this and the in-place version of esimd radix sort, and added a test utility to generate a new kernel_params struct with a new  kernelname.

Thanks for catching this.

danhoeflinger:
> The PR looks good to me besides `_out_of_place` function name part, which is mentioned in the description.
@dmitriy-sobolev 
I've updated the APIs to use `enable_if` to disambiguate the range and iterator versions.  If we decide to remove the range interfaces, it will be an easy change from the current state of the PR.  Please take a look at the `__is_iterator_type_v` check.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -717,6 +717,19 @@ struct __spirv_target_conditional :
 // to determine SPIR-V targets.
 inline constexpr bool __is_spirv_target_v = __spirv_target_conditional<::std::true_type, ::std::false_type>::value;
 
+template <typename T, typename = void>
```
dmitriy-sobolev, line 720: Uglification: `T` -> `_T`.

dmitriy-sobolev (commented):
The approach to get rid of `out_of_place` suffix looks good to me. I've left some minor comments, though.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -216,9 +216,15 @@ __onesweep_impl(sycl::queue __q, _RngPack1&& __input_pack, _RngPack2&& __virt_pa
         __esimd_radix_sort_onesweep_histogram<_KernelName>>;
     using _EsimdRadixSortScan = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
         __esimd_radix_sort_onesweep_scan<_KernelName>>;
-    using _EsimdRadixSortSweep =
-        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<::std::conditional_t<
-            __has_values, __esimd_radix_sort_onesweep_by_key<_KernelName>, __esimd_radix_sort_onesweep<_KernelName>>>;
+    using _EsimdRadixSortSweepInitial = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
+        ::std::conditional_t<__has_values, __esimd_radix_sort_onesweep_by_key<std::decay_t<_RngPack1>, std::decay_t<_RngPack2>, _KernelName>,
+                             __esimd_radix_sort_onesweep<std::decay_t<_RngPack1>, std::decay_t<_RngPack2>, _KernelName>>>;
```
dmitriy-sobolev: I'm not strictly opposed of that approach, but there is a drawback: kernel names will become longer (I guess, much longer).

The motivation of the change is to separate kernel names for the potential arrangements of `_RngPack1`, `_RngPack2`, `_RngPack3`, isn't it? If so, what do you think about using some constant instead, e.g.:

- ` __esimd_radix_sort_onesweep<0, ...>` instead of `__esimd_radix_sort_onesweep<std::decay_t<_RngPack1>, std::decay_t<_RngPack2>, ...>`,  
- `__esimd_radix_sort_onesweep<1, ...>` instead of `__esimd_radix_sort_onesweep<std::decay_t<_RngPack3>, std::decay_t<_RngPack2>, ...>`?

danhoeflinger: The goal is to have a unique kernel name if and only if we have a unique kernel.  We want to avoid same mangled name error, but also we don't want to add extra kernel compilation only to support a differently named kernel when the types match exactly.  

Lets say `_RngPack1` is the exact same as `_RngPack3`, we then want `__esimd_radix_sort_onesweep<std::decay_t<_RngPack1>, std::decay_t<_RngPack2>, ...>`
and `__esimd_radix_sort_onesweep<std::decay_t<_RngPack3>, std::decay_t<_RngPack2>, ...>` to be the exact same as well.  If not, I think we require an extra kernel compilation only to satisfy having a different name.

If all three exactly match, we only want one kernel and one kernel name.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -194,24 +202,35 @@ class __onesweep_memory_holder
...
+            __esimd_radix_sort_onesweep_by_key<std::decay_t<_RngPack3>, std::decay_t<_RngPack2>, _KernelName>,
+            __esimd_radix_sort_onesweep<std::decay_t<_RngPack3>, std::decay_t<_RngPack2>, _KernelName>>>;
+    using _EsimdRadixSortSweepOdd =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<::std::conditional_t<
+            __has_values,
+            __esimd_radix_sort_onesweep_by_key<std::decay_t<_RngPack2>, std::decay_t<_RngPack3>, _KernelName>,
+            __esimd_radix_sort_onesweep<std::decay_t<_RngPack2>, std::decay_t<_RngPack3>, _KernelName>>>;
     using _EsimdRadixSortCopyback = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
         ::std::conditional_t<__has_values, __esimd_radix_sort_onesweep_copyback_by_key<_KernelName>,
                              __esimd_radix_sort_onesweep_copyback<_KernelName>>>;
```
dmitriy-sobolev, line 236: It is not used, let's remove it.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
dmitriy-sobolev: That's reasonable, thanks for describing. 

Probably, it is still possible to use shorter kernel names. I will think about it separately as it is rather a "nice to have" thing.

Currently, I see the following options:

Rng1 type | Rng2 type | Rng3 type | Kernels | Constants
-- | -- | -- | -- | --
A | B | C | A->B, B->C, C->B | 0,1, 2
A | A | A | A->A, A->A, A->A | 3, 3, 3
A | A | B | A->A, A->B, B->A | 3, 0, 4
A | B | A | A->B, B->A, A->B | 0, 4, 0
B | A | A | B->A, A->A, A->A | 4, 3, 3

danhoeflinger: I'm interested in the idea, but to create such constants, you would need to basically check `is_same<decay_t<_RngPack1>, decay_t<_RngPack2>>` for all three combos and assign constants appropriately.  
The code is ugly either way.  Perhaps the shorter kernel name is worth it.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
```diff
@@ -81,7 +82,91 @@ radix_sort_by_key(sycl::queue __q, _KeysIterator __keys_first, _KeysIterator __k
...
+    return __impl::__radix_sort<__is_ascending, __radix_bits, /*__in_place=*/true>(__q, __pack, __pack, __param);
+}
+
+template <bool __is_ascending = true, ::std::uint8_t __radix_bits = 8, typename _KernelParam, typename _KeysRng1,
+          typename _KeysRng2>
+std::enable_if_t<!oneapi::dpl::__internal::__is_iterator_type_v<_KeysRng1>, sycl::event>
+radix_sort(sycl::queue __q, _KeysRng1&& __keys_rng, _KeysRng2&& __keys_rng_out, _KernelParam __param = {})
+{
+    __impl::__check_esimd_sort_params<__radix_bits, _KernelParam::data_per_workitem, _KernelParam::workgroup_size>();
+    if (__keys_rng.size() == 0)
```
dmitriy-sobolev, line 94: Let's change the check to `... == 1` here and in other places. We do not want to sort 1 element.

danhoeflinger: If we do this, then we need to have a separate mechanism to copy 1 element from the input to the output.  
I suppose we can do this as an optimization, but I was hoping to avoid having to specially handle such a case for kernel templates.  (We cant just noop because the output needs to be populated).
While `n` may be a runtime parameter, I think that a user who is calling kernel templates has more responsibility to avoid doing something like sorting a set of `1` element on the device.  For me, within KT there is less responsibility to provide special casing like this on behalf of the user.

dmitriy-sobolev: I agree with you. Thanks for describing one more subtle (at least for me) detail. 

What about removing the assert below?
 https://github.com/oneapi-src/oneDPL/blob/f371fb91935e9bc1076f7881853d200f894a7bbc/include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h#L310

dmitriy-sobolev (commented):
I've checked the code once again. I'm ready to approve once minor issues noted above have been solved.

include/oneapi/dpl/experimental/kt/esimd_radix_sort.h
danhoeflinger: Aha! good catch.  Yes I'll make the change.

danhoeflinger: I adjusted this to `__n > 0` because we still enforce that.

dmitriy-sobolev (approved):
LGTM

PR #1440, opened 2024-03-11
[Dynamic Selection] Skipping auto_tune and dynamic load tests when device=FPGA_EMU

AnuyaWelling2801:
Tests with default device as FPGA_EMU have a runtime error of - terminate called after throwing an instance of 'sycl::_V1::runtime_error'.
Skipping since we do not use default device in our tests.


timmiesmith (approved):
LGTM

PR #1443, opened 2024-03-12
[Doc] Adding note for std::vector<bool>

danhoeflinger:
`std::vector<bool>` has an optimization to a bitfield rather than individual booleans.  For host policies, this will simply not work for parallel algorithms (without some extra processing step / copy, which will kill performance).

On `icpx` and with device policies it works:
- With the default allocator, `std::vector<bool>::iterator` works as input to algorithms using a device policy, but only because of the copy to initialize the `sycl::buffer`. 
- With `usm_allocator`, `std::vector<bool>::iterator` also works, but only because it is treated as if it were a host_iterator, and copied into a `sycl::buffer`.  After #1438, this should still work by treating it as a host_iterator and copying to a `sycl::buffer`, because the implementation does not distinguish iterator types based on allocator.  

Since this depends on the implementation details of `std::vector<bool>`, it is safest to just label it undefined behavior, even if it happens to currently work for device policies.

### DIFF
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -1,15 +1,14 @@
 Pass Data to Algorithms
 #######################
 
-When using C++ Standard Execution Policies, oneDPL supports data being passed to its algorithms as specified in the
-ISO/IEC 14882:2017 standard (commonly called C++17). According to this standard, it is the user's responsibility to
-avoid data races when using these parallel execution policies.
+When using the C++ standard execution policies, oneDPL supports data being passed to its algorithms as specified
+in the ISO/IEC 14882:2017 standard (commonly called C++17). According to the standard, the calling code
+must prevent data races when using algorithms with parallel execution policies.
 
-Note: ``std::vector<bool>`` implementations are not required to avoid data races when the contents of the contained
-object in different elements in the same sequence are modified concurrently. Some implementations of
-``std::vector<bool>`` may optimize multiple ``bool`` elements into a bitfield, causing data races with parallel
-execution policies. For this reason, it is recommended to avoid ``std::vector<bool>`` for anything but a read-only
-input with C++ Standard Execution Policies.
+Note: Implementations of ``std::vector<bool>`` are not required to avoid data races for concurrent modifications
+of vector elements. Some implementations may optimize multiple ``bool`` elements into a bitfield, making it unsafe
+for multithreading. For this reason, it is recommended to avoid ``std::vector<bool>`` for anything but a read-only
+input with the C++ standard execution policies.
 
 When using a device execution policy, you can use one of the following ways to pass data to an algorithm:
 
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -1,7 +1,17 @@
 Pass Data to Algorithms
 #######################
 
-You can use one of the following ways to pass data to an algorithm executed with a device policy:
+When using C++ Standard Execution Policies, oneDPL supports data being passed to its algorithms as specified in the
+ISO/IEC 14882:2017 standard (commonly called C++17). According to this standard, it is the user's responsibility to
+avoid data races when using these parallel execution policies.
+
+Note: ``std::vector<bool>`` implementations are not required to avoid data races when the contents of the contained
+object in different elements in the same sequence are modified concurrently. Some implementations of
+``std::vector<bool>`` may optimize multiple ``bool`` elements into a bitfield, causing data races with parallel
+execution policies. For this reason, it is recommended to avoid ``std::vector<bool>`` for anything but a read-only
+input with C++ Standard Execution Policies.
+
+When using a device execution policy, you can use one of the following ways to pass data to an algorithm:
 
 * ``oneapi:dpl::begin`` and ``oneapi::dpl::end`` functions
 * Unified shared memory (USM) pointers and ``std::vector`` with USM allocators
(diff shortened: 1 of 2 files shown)


akukanov:
I have tried to find out if the standard says anything about using parallel algorithms with `vector<bool>`. There is nothing directly said in this regard, and the only indirect restriction I found is that `vector<bool>` is not required to avoid data races (see [container.requirements.dataraces]) while using parallel execution policies requires the caller to avoid data races (see [algorithms.parallel.exec]). So the standard prohibits modifying `vector<bool>` in parallel algorithms, but seemingly does not prevent using it for read-only input data.

So, instead of calling it a known limitation, I would add some note to the subsection "Difference with Standard C++ Parallel Algorithms". Also it makes sense to clarify in https://github.com/oneapi-src/oneDPL/edit/main/documentation/library_guide/parallel_api/pass_data_algorithms.rst (subsections "Use Unified Shared Memory" and "Use Host-Side std::vector") that `vector<bool>` will not work.

In the future, we can perhaps consider checking where our implementation breaks with `vector<bool>` and see if it can be improved.

danhoeflinger:
> I have tried to find out if the standard says anything about using parallel algorithms with `vector<bool>`. There is nothing directly said in this regard, and the only indirect restriction I found is that `vector<bool>` is not required to avoid data races (see [container.requirements.dataraces]) while using parallel execution policies requires the caller to avoid data races (see [algorithms.parallel.exec]). So the standard prohibits modifying `vector<bool>` in parallel algorithms, but seemingly does not prevent using it for read-only input data.

Thanks for these pointers to the specification, this makes sense.  I haven't done testing to confirm this, but I believe our implementation likely then complies to the standard. 
 
> So, instead of calling it a known limitation, I would add some note to the subsection "Difference with Standard C++ Parallel Algorithms". Also it makes sense to clarify in https://github.com/oneapi-src/oneDPL/edit/main/documentation/library_guide/parallel_api/pass_data_algorithms.rst (subsections "Use Unified Shared Memory" and "Use Host-Side std::vector") that `vector<bool>` will not work.

With the spec sections you point to, I believe it is neither a "known limitation" nor a "difference from the standard".  We can do some testing to confirm this though.  

Despite this, I do I think it would still be helpful to users to draw attention to the fact that `std::vector<bool>::iterator` should not be used as output data in oneDPL (at least for for  C++ Standard Execution Policies) as it is not protected against dataraces, as that may not be clear to those who are less familiar with the details of `std::vector<bool>`.  

I think it probably doesn't belong on the page you mention though, because that is specific to device execution policies, where `std::vector<bool>::iterator` will actually work both as input and output due to the copy into a `sycl::buffer`.  

I will try to find a better place to put such a note, and update this PR.
 
> In the future, we can perhaps consider checking where our implementation breaks with `vector<bool>` and see if it can be improved.

Agreed.

danhoeflinger:
@akukanov I've updated the PR to instead add to the input data page of the guide, with a note about `std::vector<bool>` (and remove it as a known limitation).  Lets wait to merge until I can confirm some things through testing about host-side support of `std::vector<bool>` as read-only input data.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -1,7 +1,17 @@
...
+Note: ``std::vector<bool>`` implementations are not required to avoid data races when the contents of the contained
+object in different elements in the same sequence are modified concurrently. Some implementations of
+``std::vector<bool>`` may optimize multiple ``bool`` elements into a bitfield, causing data races with parallel
+execution policies. For this reason, it is recommended to avoid ``std::vector<bool>`` for anything but a read-only
+input with C++ Standard Execution Policies.
+
+When using a device execution policy, you can use one of the following ways to pass data to an algorithm:
 
 * ``oneapi:dpl::begin`` and ``oneapi::dpl::end`` functions
 * Unified shared memory (USM) pointers and ``std::vector`` with USM allocators
```
MikeDvorskiy, line 17: In regard to ```std::vector`` with USM allocators`..
In spite of we have a comment 
`// for raw pointers and direct pass objects (for example, counting_iterator, iterator of USM-containers)`
I an afraid we don't (and cannot) differ  an iterator of the host data allocated container and an iterator of C++ container with USM allocator.

MikeDvorskiy (approved):
The new wording looks good to me.
But I have a concern regarding the sentence which we had  before and keep now:
`std::vector with USM allocators`

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: In many cases I believe we can distinguish a iterator from a `std::vector` with `USM allocator`, however it is implementation dependent.
In #1438 I add some code which does to the extent that it is possible.  
Please see this comment for explanation:  https://github.com/oneapi-src/oneDPL/pull/1438#issuecomment-1989312899

I think it better to continue the discussion of this topic in that PR.  We could improve the documentation language there perhaps to be more conservative in our claims for USM allocators.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -1,7 +1,17 @@
 Pass Data to Algorithms
 #######################
 
-You can use one of the following ways to pass data to an algorithm executed with a device policy:
+When using C++ Standard Execution Policies, oneDPL supports data being passed to its algorithms as specified in the
+ISO/IEC 14882:2017 standard (commonly called C++17). According to this standard, it is the user's responsibility to
+avoid data races when using these parallel execution policies.
```
akukanov, line 6:
```suggestion
When using the C++ standard execution policies, oneDPL supports data being passed to its algorithms as specified
in the ISO/IEC 14882:2017 standard (commonly called C++17). According to the standard, the calling code
must prevent data races when using algorithms with parallel execution policies.
```

akukanov (approved):

PR #1445, opened 2024-03-13
`transform_iterator` improvements, constructors, copy assignment, docs, tests

danhoeflinger:
The `transform_iterator` constructor has the potential to cause problems for checks of default constructability (and different results for different compilers `clang` vs `gcc`).

This PR makes clear to the compiler exactly when the `transform_iterator` is default constructible, where other simpler versions with default arguments or without the `enable_if` may result in build errors or the wrong value being returned when `std::is_default_constructible` is called on the `transform_iterator`.

This PR changes `transform_iterator` in that it only enables the constructors which are valid in the instantiation by separating out the constructors which rely upon default constructors of the template types, and only enabling them under the correct circumstances of default constructability of template arguments.  

This is a better fix for the issue which motivated #1432.  However, I think both are changes for the better.

---
Update:  The PR has grown to include:

1. Changes to constructors of `transform_iterator` to enable default constructability checks
2. removal of `const` from the unary functor within `transform_iterator`
3. testing changes to match changes in code and test fixed functionality
4. documentation changes to be more explicit about eligible functors for use with `transform_iterator`
5. Adding of copy assignment of functor when functor is copy assignable from copy assignment operator of `transform_iterator`
6. Tests for copy_assignable of `transform_iterator` and checking that functor is copied when it is possible

### DIFF
--- a/documentation/library_guide/parallel_api/iterators.rst
+++ b/documentation/library_guide/parallel_api/iterators.rst
@@ -72,7 +72,7 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
   the ``transform_iterator`` class. 
 
   The unary functor provided to a transform iterator should have a const call operator that takes a single argument
-  whose paramter type can be converted to from the reference type of the original iterator. The functor's call operator
+  whose parameter type can be converted to from the reference type of the original iterator. The functor's call operator
   should not have any side effects and should not modify the state of the functor object.
   
   The ``transform_iterator`` class provides three constructors:
--- a/documentation/library_guide/parallel_api/iterators.rst
+++ b/documentation/library_guide/parallel_api/iterators.rst
@@ -75,7 +75,7 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
   the reference type of the base iterator as argument. The functor's call operator should not have any side effects and
   should not modify the state of the functor object.
   
-  The ``transform_iterator`` class provides three constructors:
+  The ``transform_iterator`` class provides the following constructors:
 
   * ``transform_iterator()``: instantiates the iterator using a default constructed base iterator and unary functor.
   This constructor participates in overload resolution only if the base iterator and unary functor are both default constructible.
--- a/documentation/library_guide/parallel_api/iterators.rst
+++ b/documentation/library_guide/parallel_api/iterators.rst
@@ -71,16 +71,16 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
   iterator and the unary function applied during dereference operations are required template parameters of
   the ``transform_iterator`` class. 
 
-  The unary functor provided to a transform iterator should have a const call operator which is callable with an
-  argument of the reference type of the original iterator. The functor's call operator should not have any side
-  effects and should not modify the state of the functor object.
+  The unary functor provided to a ``transform_iterator`` should have a ``const``-qualified call operator which can
+  be called using the reference type of the original iterator as argument. The functor's call operator should not
+  have any side effects and should not modify the state of the functor object.
   
   The ``transform_iterator`` class provides three constructors:
 
   * ``transform_iterator()``: instantiates the iterator using a default constructed base iterator and unary functor.
-  This constructor is only provided if the base iterator and unary functor are both default constructible.
+  This constructor participates in overload resolution only if the base iterator and unary functor are both default constructible.
   * ``transform_iterator(iter)``: instantiates the iterator using the base iterator provided and a default constructed
-  unary functor. This constructor is only provided if the unary functor is default constructible.
+  unary functor. This constructor participates in overload resolution only if the unary functor is default constructible.
   * ``transform_iterator(iter, func)``: instantiates the iterator using the base iterator and unary functor provided.
 
   To simplify the construction of the iterator, ``oneapi::dpl::make_transform_iterator`` is provided. The
--- a/documentation/library_guide/parallel_api/iterators.rst
+++ b/documentation/library_guide/parallel_api/iterators.rst
@@ -71,9 +71,9 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
   iterator and the unary function applied during dereference operations are required template parameters of
   the ``transform_iterator`` class. 
 
-  The unary functor provided to a transform iterator should have a const call operator that takes a single argument
-  whose parameter type can be converted to from the reference type of the original iterator. The functor's call operator
-  should not have any side effects and should not modify the state of the functor object.
+  The unary functor provided to a transform iterator should have a const call operator which is callable using the
+  reference type of the original iterator. The functor's call operator should not have any side effects and should not
+  modify the state of the functor object.
   
   The ``transform_iterator`` class provides three constructors:
 
--- a/documentation/library_guide/parallel_api/iterators.rst
+++ b/documentation/library_guide/parallel_api/iterators.rst
@@ -67,13 +67,13 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
                  [](auto t){return get<1>(t) == 1;}
 
 * ``transform_iterator``: an iterator defined over another iterator whose dereferenced value is the result
-  of a function applied to the corresponding element of the original iterator. Both the type of the original
+  of a function applied to the corresponding element of the base iterator. Both the type of the base
   iterator and the unary function applied during dereference operations are required template parameters of
   the ``transform_iterator`` class. 
 
-  The unary functor provided to a ``transform_iterator`` should have a ``const``-qualified call operator which can
-  be called using the reference type of the original iterator as argument. The functor's call operator should not
-  have any side effects and should not modify the state of the functor object.
+  The unary functor provided to a ``transform_iterator`` should have a ``const``-qualified call operator which accepts
+  the reference type of the base iterator as argument. The functor's call operator should not have any side effects and
+  should not modify the state of the functor object.
   
   The ``transform_iterator`` class provides three constructors:
 
@@ -84,7 +84,7 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
   * ``transform_iterator(iter, func)``: instantiates the iterator using the base iterator and unary functor provided.
 
   To simplify the construction of the iterator, ``oneapi::dpl::make_transform_iterator`` is provided. The
-  function receives the original iterator and transform operation instance as arguments, and constructs the
+  function receives the base iterator and transform operation instance as arguments, and constructs the
   ``transform_iterator`` instance::
 
     using namespace oneapi;
--- a/documentation/library_guide/parallel_api/iterators.rst
+++ b/documentation/library_guide/parallel_api/iterators.rst
@@ -69,10 +69,18 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
 * ``transform_iterator``: an iterator defined over another iterator whose dereferenced value is the result
   of a function applied to the corresponding element of the original iterator. Both the type of the original
   iterator and the unary function applied during dereference operations are required template parameters of
-  the ``transform_iterator`` class. The ``transform_iterator`` class provides three constructors:
+  the ``transform_iterator`` class. 
+
+  The unary functor provided to a transform iterator should have a const call operator that takes a single argument
+  whose paramter type can be converted to from the reference type of the original iterator. The functor's call operator
+  should not have any side effects and should not modify the state of the functor object.
+  
+  The ``transform_iterator`` class provides three constructors:
 
   * ``transform_iterator()``: instantiates the iterator using a default constructed base iterator and unary functor.
-  * ``transform_iterator(iter)``: instantiates the iterator using the base iterator provided and a default constructed unary functor.
+  This constructor is only provided if the base iterator and unary functor are both default constructible.
+  * ``transform_iterator(iter)``: instantiates the iterator using the base iterator provided and a default constructed
+  unary functor. This constructor is only provided if the unary functor is default constructible.
   * ``transform_iterator(iter, func)``: instantiates the iterator using the base iterator and unary functor provided.
 
   To simplify the construction of the iterator, ``oneapi::dpl::make_transform_iterator`` is provided. The
--- a/documentation/library_guide/parallel_api/iterators.rst
+++ b/documentation/library_guide/parallel_api/iterators.rst
@@ -71,9 +71,9 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
   iterator and the unary function applied during dereference operations are required template parameters of
   the ``transform_iterator`` class. 
 
-  The unary functor provided to a transform iterator should have a const call operator which is callable using the
-  reference type of the original iterator. The functor's call operator should not have any side effects and should not
-  modify the state of the functor object.
+  The unary functor provided to a transform iterator should have a const call operator which is callable with an
+  argument of the reference type of the original iterator. The functor's call operator should not have any side
+  effects and should not modify the state of the functor object.
   
   The ``transform_iterator`` class provides three constructors:
 
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -401,7 +401,7 @@ class transform_iterator
 {
   private:
     _Iter __my_it_;
-    const _UnaryFunc __my_unary_func_;
+    _UnaryFunc __my_unary_func_;
 
   public:
     typedef typename ::std::iterator_traits<_Iter>::difference_type difference_type;
@@ -410,19 +410,16 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    //only enable this constructor if both template types are default constructible
-    template <typename _IterLocal = _Iter, typename _UnaryFuncLocal = _UnaryFunc,
-              typename = ::std::enable_if_t<
-                  ::std::is_default_constructible_v<_IterLocal> && ::std::is_default_constructible_v<_UnaryFuncLocal>>>
-    transform_iterator() : __my_it_(), __my_unary_func_()
-    {
-    }
+    //default constructor will only be present if both the unary functor and iterator are default constructible
+    transform_iterator() = default;
+
     //only enable this constructor if the unary functor is default constructible
     template <typename _UnaryFuncLocal = _UnaryFunc,
               ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFuncLocal>, int> = 0>
     transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_()
     {
     }
+
     transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(__it), __my_unary_func_(__unary_func) {}
 
     transform_iterator(const transform_iterator& __input) = default;
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -403,8 +403,7 @@ class transform_iterator
     _Iter __my_it_;
     _UnaryFunc __my_unary_func_;
 
-    static_assert(oneapi::dpl::__internal::__is_const_callable_with_arg_v<
-                      _UnaryFunc, typename std::iterator_traits<_Iter>::reference>,
+    static_assert(std::is_invocable_v<const _UnaryFunc, typename std::iterator_traits<_Iter>::reference>,
                   "Unary Functor is not const callable with the reference type of base iterator.");
 
   public:
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -403,6 +403,10 @@ class transform_iterator
     _Iter __my_it_;
     _UnaryFunc __my_unary_func_;
 
+    static_assert(oneapi::dpl::__internal::__is_const_callable_with_arg_v<
+                      _UnaryFunc, typename std::iterator_traits<_Iter>::reference>,
+                  "Unary Functor is not const callable with the reference type of base iterator.");
+
   public:
     typedef typename ::std::iterator_traits<_Iter>::difference_type difference_type;
     typedef decltype(__my_unary_func_(::std::declval<typename ::std::iterator_traits<_Iter>::reference>())) reference;
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -410,7 +410,13 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator() = default;
+    //only enable this constructor if both template types are default constructible
+    template <typename _IterLocal = _Iter, typename _UnaryFuncLocal = _UnaryFunc,
+              typename = ::std::enable_if_t<
+                  ::std::is_default_constructible_v<_IterLocal> && ::std::is_default_constructible_v<_UnaryFuncLocal>>>
+    transform_iterator() : __my_it_(), __my_unary_func_()
+    {
+    }
     //only enable this constructor if the unary functor is default constructible
     template <typename _UnaryFuncLocal = _UnaryFunc,
               ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFuncLocal>, int> = 0>
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -404,7 +404,8 @@ class transform_iterator
     _UnaryFunc __my_unary_func_;
 
     static_assert(std::is_invocable_v<const std::decay_t<_UnaryFunc>, typename std::iterator_traits<_Iter>::reference>,
-                  "_UnaryFunc does not have a const-qualified call operator which accepts the reference type of the base iterator as argument.");
+                  "_UnaryFunc does not have a const-qualified call operator which accepts the reference type of the "
+                  "base iterator as argument.");
 
   public:
     typedef typename ::std::iterator_traits<_Iter>::difference_type difference_type;
@@ -433,7 +434,7 @@ class transform_iterator
     operator=(const transform_iterator& __input)
     {
         __my_it_ = __input.__my_it_;
-        
+
         // If copy assignment is available, copy the functor, otherwise skip it.
         // For non-copy assignable functors, this copy assignment operator departs from the sycl 2020 specification
         // requirement of device copyable types for copy assignment to be the same as a bitwise copy of the object.
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -419,7 +419,7 @@ class transform_iterator
 
     //only enable this constructor if the unary functor is default constructible
     template <typename _UnaryFuncLocal = _UnaryFunc,
-              ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFuncLocal>, int> = 0>
+              std::enable_if_t<std::is_default_constructible_v<_UnaryFuncLocal>, int> = 0>
     transform_iterator(_Iter __it) : __my_it_(std::move(__it))
     {
     }
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -410,8 +410,19 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator() : __my_it_(), __my_unary_func_() {}
-    transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_() {}
+    //only enable this constructor if both template types are default constructible
+    template <typename _Iter_ = _Iter, typename _UnaryFunc_ = _UnaryFunc,
+              typename = ::std::enable_if_t<
+                  ::std::is_default_constructible_v<_Iter_>&& ::std::is_default_constructible_v<_UnaryFunc_>>>
+    transform_iterator() : __my_it_(), __my_unary_func_()
+    {
+    }
+    //only enable this constructor if the unary functor is default constructible
+    template <typename _UnaryFunc_ = _UnaryFunc,
+              typename = ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFunc_>>>
+    transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_()
+    {
+    }
     transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(__it), __my_unary_func_(__unary_func) {}
 
     transform_iterator(const transform_iterator& __input) = default;
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -410,10 +410,10 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc())
-        : __my_it_(__it), __my_unary_func_(__unary_func)
-    {
-    }
+    transform_iterator() : __my_it_(), __my_unary_func_() {}
+    transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_() {}
+    transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(__it), __my_unary_func_(__unary_func) {}
+
     transform_iterator(const transform_iterator& __input) = default;
     transform_iterator&
     operator=(const transform_iterator& __input)
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -432,11 +432,17 @@ class transform_iterator
     transform_iterator&
     operator=(const transform_iterator& __input)
     {
-        //TODO: Investigate making transform_iterator trivially_copyable. This custom copy assignment operator prevents
-        // transform_iterator, and therefore permutation_iterator from being trivially_copyable.  Not being trivially
-        // copyable makes their device_copyable trait deprecated in SYCL2020. However, defaulting this function implies
-        // an extra requirement that __my_unary_func_ implements a copy assignment operator.
         __my_it_ = __input.__my_it_;
+        
+        // If copy assignment is available, copy the functor, otherwise skip it.
+        // For non-copy assignable functors, this copy assignment operator departs from the sycl 2020 specification
+        // requirement of device copyable types for copy assignment to be the same as a bitwise copy of the object.
+        // TODO: Explore (ABI breaking) change to use std::optional or similar and using copy constructor to implement
+        //       copy assignment to better comply with SYCL 2020 specification.
+        if constexpr (std::is_copy_assignable_v<_UnaryFunc>)
+        {
+            __my_unary_func_ = __input.__my_unary_func_;
+        }
         return *this;
     }
     reference operator*() const { return __my_unary_func_(*__my_it_); }
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -419,11 +419,14 @@ class transform_iterator
     //only enable this constructor if the unary functor is default constructible
     template <typename _UnaryFuncLocal = _UnaryFunc,
               ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFuncLocal>, int> = 0>
-    transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_()
+    transform_iterator(_Iter __it) : __my_it_(std::move(__it))
     {
     }
 
-    transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(__it), __my_unary_func_(__unary_func) {}
+    transform_iterator(_Iter __it, _UnaryFunc __unary_func)
+        : __my_it_(std::move(__it)), __my_unary_func_(std::move(__unary_func))
+    {
+    }
 
     transform_iterator(const transform_iterator&) = default;
     transform_iterator&
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -404,7 +404,7 @@ class transform_iterator
     _UnaryFunc __my_unary_func_;
 
     static_assert(std::is_invocable_v<const std::decay_t<_UnaryFunc>, typename std::iterator_traits<_Iter>::reference>,
-                  "_UnaryFunc is not const callable with the reference type of base iterator.");
+                  "_UnaryFunc does not have a const-qualified call operator which accepts the reference type of the base iterator as argument.");
 
   public:
     typedef typename ::std::iterator_traits<_Iter>::difference_type difference_type;
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -412,8 +412,8 @@ class transform_iterator
 
     transform_iterator() = default;
     //only enable this constructor if the unary functor is default constructible
-    template <typename _UnaryFunc_ = _UnaryFunc,
-              ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFunc_>, int> = 0>
+    template <typename _UnaryFuncLocal = _UnaryFunc,
+              ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFuncLocal>, int> = 0>
     transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_()
     {
     }
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -404,7 +404,7 @@ class transform_iterator
     _UnaryFunc __my_unary_func_;
 
     static_assert(std::is_invocable_v<const _UnaryFunc, typename std::iterator_traits<_Iter>::reference>,
-                  "Unary Functor is not const callable with the reference type of base iterator.");
+                  "_UnaryFunc is not const callable with the reference type of base iterator.");
 
   public:
     typedef typename ::std::iterator_traits<_Iter>::difference_type difference_type;
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -403,7 +403,7 @@ class transform_iterator
     _Iter __my_it_;
     _UnaryFunc __my_unary_func_;
 
-    static_assert(std::is_invocable_v<const _UnaryFunc, typename std::iterator_traits<_Iter>::reference>,
+    static_assert(std::is_invocable_v<const std::decay_t<_UnaryFunc>, typename std::iterator_traits<_Iter>::reference>,
                   "_UnaryFunc is not const callable with the reference type of base iterator.");
 
   public:
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -425,7 +425,7 @@ class transform_iterator
 
     transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(__it), __my_unary_func_(__unary_func) {}
 
-    transform_iterator(const transform_iterator& __input) = default;
+    transform_iterator(const transform_iterator&) = default;
     transform_iterator&
     operator=(const transform_iterator& __input)
     {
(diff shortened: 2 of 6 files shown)


danhoeflinger:
So, this change has an consequence... 

[LegacyForwardIterator](https://en.cppreference.com/w/cpp/named_req/ForwardIterator) and therefore [LegacyBidirectionalIterator](https://en.cppreference.com/w/cpp/named_req/BidirectionalIterator) and [LegacyRandomAccessIterator](https://en.cppreference.com/w/cpp/named_req/RandomAccessIterator) requires default constructible.  

The way `transform_iterator` was written, it was returning `true` for `is_default_constructible` even when it was not the case (lambdas are not default constructible in C++17).  Now it accurately returns `false` when the unary functor is not default constructible.

The [random_access_iterator](https://en.cppreference.com/w/cpp/iterator/random_access_iterator) concept is added in C++20.  Practically, we may not have an issue here, other with our tests which are `static_assert` default constructible.  We need to consider the consequences of this change though.  We also merely carry forward the iterator category from the source iterator, regardless of the functor.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,10 +410,21 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc())
-        : __my_it_(__it), __my_unary_func_(__unary_func)
+    //only enable this constructor if both template types are default constructible
```
SergeyKopienko, line 413: What if we will write something like
```C++
 transform_iterator() = default;

 transform_iterator(_Iter __it)
        : __my_it_(__it)
    {
    }

 transform_iterator(_Iter __it, _UnaryFunc __unary_func)
        : __my_it_(__it), __my_unary_func_(__unary_func)
    {
    }
```

?

test/parallel_api/iterator/transform_iterator.pass.cpp
```diff
@@ -182,9 +182,58 @@ test_copyable()
...
+    EXPECT_TRUE(
+        !::std::is_default_constructible_v<decltype(trans4)>,
+        "transform_iterator with non-default constructible iterator source is seen to be default constructible");
+}
+
 std::int32_t
 main()
 {
+    test_default_constructible();
 #if TEST_DPCPP_BACKEND_PRESENT
```
SergeyKopienko, line 237: May be better to separate this test?
One for `test_default_constructible` check + one for `test_copyable` (with `#if TEST_DPCPP_BACKEND_PRESENT`) ?

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,10 +410,21 @@ class transform_iterator
...
+    template <typename _Iter_ = _Iter, typename _UnaryFunc_ = _UnaryFunc,
```
SergeyKopienko, line 414: Should we really have here `typename _Iter_ = _Iter, typename _UnaryFunc_ = _UnaryFunc,` ?

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,10 +410,21 @@ class transform_iterator
...
+              typename = ::std::enable_if_t<
+                  ::std::is_default_constructible_v<_Iter_>&& ::std::is_default_constructible_v<_UnaryFunc_>>>
+    transform_iterator() : __my_it_(), __my_unary_func_()
     {
     }
+    //only enable this constructor if the unary functor is default constructible
+    template <typename _UnaryFunc_ = _UnaryFunc,
```
SergeyKopienko, line 421: Should we really have here `typename _UnaryFunc_ = _UnaryFunc,` ?

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Sadly it is not enough.  With these formulations `std::is_default_constructible_v<transform_iterator<it,non_default_constructible_functor>>` returns `true`.  It seems that only the immediate context it evaluated during this check, which does not include the "body" of the constructor in question.  We need to explicitly disable the ctor, so that it is not included in the instantiation.

SergeyKopienko: OK

test/parallel_api/iterator/transform_iterator.pass.cpp
```diff
@@ -182,9 +182,58 @@ test_copyable()
...
+    oneapi::dpl::transform_iterator<decltype(trans3), noop> trans4{trans3, noop{}};
+    EXPECT_TRUE(
+        !::std::is_default_constructible_v<decltype(trans4)>,
+        "transform_iterator with non-default constructible iterator source is seen to be default constructible");
+}
+
 std::int32_t
 main()
 {
+    test_default_constructible();
```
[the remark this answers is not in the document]
danhoeflinger: Do you mean add `transform_iterator_traits.pass.cpp` or something like that to run the `test_default_constructible` and `test_copyable` ?

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: We can use different names for the types, but yes I believe we need these rather than just relying on the structure's templates.  If we make these not dependent on template parameters specific to the function, it will cause problems for the structure as a whole I believe.

test/parallel_api/iterator/transform_iterator.pass.cpp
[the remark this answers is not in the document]
SergeyKopienko: transform_iterator_constructible.pass.cpp
transform_iterator_copyable.pass.cpp with skip if `TEST_DPCPP_BACKEND_PRESENT` is undefined

include/oneapi/dpl/pstl/iterator_impl.h
SergeyKopienko: ok

test/parallel_api/iterator/transform_iterator.pass.cpp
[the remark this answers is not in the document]
danhoeflinger: OK.  with the rest of transform_iterator.pass.cpp also depending on `TEST_DPCPP_BACKEND_PRESENT`, do you think its necessary to pull `test_copyable()` out in addition to `transform_iterator_constructible.pass.cpp`?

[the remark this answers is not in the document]
SergeyKopienko: Yes, but I mean when this test passed you may don't know have you checked `test_copyable()` or no.
Because you calls `return TestUtils::done();` in all cases.
But up to you.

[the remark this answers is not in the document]
danhoeflinger: If we pull out `transform_iterator_constructible`, then I will restore the previous done() call here.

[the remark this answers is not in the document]
danhoeflinger: I've pulled out `transform_iterator_constructible.pass`, now the PR doesn't touch `transform_iterator.pass`.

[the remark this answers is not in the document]
SergeyKopienko: ok

test/parallel_api/iterator/transform_iterator_constructible.pass.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+    int* ptr = nullptr;
+    oneapi::dpl::transform_iterator<int*, decltype(transformation)> trans1{ptr, transformation};
+    //default constructibility of lambdas depends on c++ standard, we want transform iterator to match its template args
+    EXPECT_TRUE((std::is_default_constructible_v<decltype(transformation)> ==
+                 ::std::is_default_constructible_v<decltype(trans1)>),
+                "transform_iterator with lambda does not match default constructibility status of the lambda itself");
+
+    //both types are default constructible
+    oneapi::dpl::transform_iterator<int*, noop> trans2{ptr, noop{}};
```
rarutyun, line 59: what this test is missing is the checking for std::is_constructible<transform_iterator, int*>. Also why asserts, not static_asserts?

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,10 +410,21 @@ class transform_iterator
     typedef typename ::std::iterator_traits<_Iter>::pointer pointer;
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc())
-        : __my_it_(__it), __my_unary_func_(__unary_func)
+    //only enable this constructor if both template types are default constructible
+    template <typename _Iter_ = _Iter, typename _UnaryFunc_ = _UnaryFunc,
+              typename = ::std::enable_if_t<
+                  ::std::is_default_constructible_v<_Iter_>&& ::std::is_default_constructible_v<_UnaryFunc_>>>
+    transform_iterator() : __my_it_(), __my_unary_func_()
```
rarutyun, line 417: `= default` should be enough

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,10 +410,21 @@ class transform_iterator
...
     {
     }
+    //only enable this constructor if the unary functor is default constructible
+    template <typename _UnaryFunc_ = _UnaryFunc,
+              typename = ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFunc_>>>
```
rarutyun:
```suggestion
              ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFunc_>, int> = 0>
```
because we don't create the precedence of overload existence. I mean, I understand that this is perhaps unclear :) But I don't know how to make it short and self-explanatory :)

include/oneapi/dpl/pstl/iterator_impl.h
rarutyun, line 415: Do we want to have less similar name of Dummy template parameter? It's hard to mess it with the original one and check the wrong one in `enable_if` condition?

test/parallel_api/iterator/transform_iterator_constructible.pass.cpp
danhoeflinger: Good idea.  Added the test and switched to static asserts.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Yes, I agree it is better for many reasons.

danhoeflinger: Actually, `= default` causes problems in combination with the `const` unary functor.  It implicitly deletes the default constructor because of this const member variable.  In the future, we may be removing `const` from the unary functor, but lets have that discussion separately, in the context of potentially removing the custom assignment operator. 

For now, we can use this version with `enable_if`.  We lose default constructability and the code is ugly, but I think its the best option for now.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Yes, this is better.  Thanks also for the explanation offline.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: I've improved it.

danhoeflinger:
The clang format suggestion is a bug in clang format where it mistakes a binary AND for a rvalue reference.  It is intentionally ignored.

include/oneapi/dpl/pstl/iterator_impl.h
rarutyun: It looks weird that for `const` member it actually deletes the default constructor...

rarutyun: If we want to have our code cleaner and keep `const` (purpose of which I don't understand) we can to like that:

```cpp
const _UnaryFunc __my_unary_func_ = _UnaryFunc();

transform_iterator() = default;
```

We lose default constructor triviality anyway. So, seems like for `const` member trivial default constructor may not exist. That makes it more evil than value

danhoeflinger: My best understanding of why the functor is `const` here is to convey to users that the functor in a `transform_iterator` should be stateless, and should not be relied upon to collect some data as it operates to affect the transformation, or for later retrieval.    

In the long run, I do believe we should remove `const`.  

Sadly, I think your suggestion to use `const _UnaryFunc __my_unary_func_ = _UnaryFunc();` results in something similar to the original issue, a build error when `is_default_constructible` is called on clang when `_UnaryFunc` is non-default constructible.

rarutyun: Ok, I see. My experiments were not quite correct. What is we remove `const` right now? Does it break somebody's code?

test/parallel_api/iterator/iterators.pass.cpp
```diff
@@ -32,7 +32,7 @@
 using namespace TestUtils;
 
 //common checks of a random access iterator functionality
-template <typename RandomIt>
+template <bool check_default_constructible = true, typename RandomIt>
```
rarutyun, line 35: This bool variable looks a little bit hacky to me. Can we do something better?

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: I don't think it can break any code to remove it, but I have not investigated this in detail.  I know @akukanov had mentioned trying to find some other way to convey the same message to users of `transform_iterator` if we decide to remove `const`.  

Perhaps that just comes in the form of documentation. I would be open to removing `const` and adding some language to the documentation.

test/parallel_api/iterator/iterators.pass.cpp
danhoeflinger: Perhaps you have a suggestion?  

Basically, we are conveying to the random iterator check that we are willing to skip checking the default constructible check for specific cases like if the lambda is not default constructible.  To me, this seems like what we want to do, maybe it would be slightly better to have `skip_default_constructible_check` and default to `false`.

Otherwise, I suppose we could separate the default constructible check out and replicate it for each iterator?  I'd prefer this `bool` to replicated code or creating another function.  As mentioned in my other comment, `LegacyRandomAccessIterator` does include default constructible, so it seems to belong in this check.

include/oneapi/dpl/pstl/iterator_impl.h
rarutyun: The thing is our `operator*` is const anyway. So having some state within lambda and trying to call mutable `operator()` ends up with [compile-time error](https://godbolt.org/z/M17obYqh3). So, I don't know how much value we bring to the users by marking the member of transform iterator as `const`. Perhaps there is a value in `operator=` but so far I am hearing that it rather creates problems for us (if my understanding is correct)

test/parallel_api/iterator/iterators.pass.cpp
rarutyun: > Perhaps you have a suggestion?

Nothing is particular. Perhaps just making those checks separately... But I have not though deeply about that. I kind of would like to see simpler code. It took me some time to understand what exactly you are trying to test for. If everybody but me feels that the simplification is not needed I am fine with that.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: I'm in agreement that `const` should be removed.  We can go ahead and do it here, then use the defaulted ctor.  I'll look at the documentation to see if there is a good way to improve it to make the user understand how transform iterator should be used, and what functors are eligible.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -401,7 +401,7 @@ class transform_iterator
 {
   private:
     _Iter __my_it_;
-    const _UnaryFunc __my_unary_func_;
+    _UnaryFunc __my_unary_func_;
```
akukanov, line 404: What do you think about adding a static_assert to enforce the callable to be non-mutable?

danhoeflinger: We already require that the call operator is const with the `const` in the dereference operator of `transform_iterator`.
`reference operator*() const { return __my_unary_func_(*__my_it_); }` 

Are you meaning to check with a `static_assert` for the availability of an appropriate const call operator of the proper signature?  I'll have to experiment with how this can be done (without sacrificing our default ctor).

I don't think there is a way to add a compile time check that the functor has no `mutable` fields.

danhoeflinger: Looking at `oneapi::dpl::__internal::__is_const_callable_object_v`, it does not work properly whenever there are multiple overloads of `operator()` (including if there is a templated `operator()`).  I don't want to use this here, and we may want to examine it's usage in `replace_copy_if`. Adding a more robust version of this trait for arbitrary signatures is probably beyond the scope of this PR.

However, I've added a check that a const version of the functor is callable with a specific argument type, and added a `static_assert` to the structure definition.  This could probably be expanded to arbitrary argument list rather than a single argument with more effort.

The benefit this provides would be a better error message for the user. It also checks upon instantiation of the type, and doesn't require attempted dereference of the `transform_iterator` within the code.

What do you think?

test/parallel_api/iterator/iterators.pass.cpp
danhoeflinger: I've separated out the check for default constructible and replaced it with a note on random access iterators in oneDPL.
  
I've confirmed that a default constructible check is present for the iterator types I think we care about being default constructible, and made some minor changes to `iterator.pass`.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: I've removed `const` and switched back to `transform_iterator() = default` .

I've added a check `__is_const_callable_with_arg_v` to the definition of `transform_iterator` to ensure that a const version of the functor is callable with the reference type of the base iterator.  This check needs review, because if it is more aggressive than the intention, then this could cause problems for user code.

I've also added some to the documentation of `tranform_iterator`.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -403,6 +403,10 @@ class transform_iterator
     _Iter __my_it_;
     _UnaryFunc __my_unary_func_;
 
+    static_assert(oneapi::dpl::__internal::__is_const_callable_with_arg_v<
+                      _UnaryFunc, typename std::iterator_traits<_Iter>::reference>,
+                  "Unary Functor is not const callable with the reference type of base iterator.");
```
akukanov, line 408: I think `__is_const_callable_with_arg_v` reinvents `std::is_invocable_v`; i.e. it should be `static_assert(std::is_invocable_v<const _UnaryFunc, typename std::iterator_traits<_Iter>::reference>,"...");`

include/oneapi/dpl/pstl/iterator_impl.h
akukanov: > Adding a more robust version of this trait for arbitrary signatures is probably beyond the scope of this PR.

I guess it is called `std::is_invocable_v` :) But yes. switching to that is beyond the scope of this PR.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Good point.  I'll make this switch.  Thanks for pointing me to that.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: `is_invocable_v` is probably what we want to be checking within `replace_copy_if` with a specific argument type.  However, it is probably not well suited to implement the general helpers `__is_callable_object`, `__is_const_callable_object` as those aren't taking into account the arguments we want to pass in (those can probably be removed though).  

I'll create an issue for this.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -401,7 +401,10 @@ class transform_iterator
 {
   private:
     _Iter __my_it_;
-    const _UnaryFunc __my_unary_func_;
+    _UnaryFunc __my_unary_func_;
+
+    static_assert(std::is_invocable_v<const _UnaryFunc, typename std::iterator_traits<_Iter>::reference>,
+                  "Unary Functor is not const callable with the reference type of base iterator.");
```
SergeyKopienko, line 407: Ma be better to use type name in this message?
`"_UnaryFunc is not const callable with the reference type of base iterator."`

SergeyKopienko: Why you are using `const _UnaryFunc` instead of `_UnaryFunc` here?
Because you already declared it as
`_UnaryFunc __my_unary_func_;` ...

danhoeflinger: It is because we want to confirm that `_UnaryFunc` is callable from within `operator*` of `transform_iterator` which is marked as `const` (meaning that member fields cannot be modified within the body).  This check is to ensure that a `_UnaryFunc` has a `const` call operator which would be eligible to be called during a dereference.

SergeyKopienko: If it's needed, we able to avoid this restriction, I think...

danhoeflinger: > May be better to use `decltype(*__my_it_)` here instead of typename std::iterator_traits<_Iter>::reference` ?

Are you suggesting this just to shorten the number of characters?  It seems like these should be the same as long as `reference` is defined correctly.  I suppose using `decltype(*__my_it_)` removes any question of that, but it seems perhaps less clear / readable.

danhoeflinger: > If it's needed, we able to avoid this restriction, I think...

I don't think I understand, are you saying that this `static_assert` is unnecessary because we already would get a compile time error from it not being there?  
This is true, but as mentioned in this thread, the difference is:
1) better error message for the user 
2) check on instantiation, without requiring an attempted dereference.  (This is perhaps not an advantage, but it is a difference)

danhoeflinger: > Ma be better to use type name in this message? `"_UnaryFunc is not const callable with the reference type of base iterator."`

Done.

SergeyKopienko: > > If it's needed, we able to avoid this restriction, I think...
> 
> I don't think I understand, are you saying that this `static_assert` is unnecessary because we already would get a compile time error from it not being there? This is true, but as mentioned in this thread, the difference is:
> 
> 1. better error message for the user
> 2. check on instantiation, without requiring an attempted dereference.  (This is perhaps not an advantage, but it is a difference)

I mean that we can make local copy of this `__my_unary_func_ ` if `std::is_invocable_v<const _UnaryFunc, typename std::iterator_traits<_Iter>::reference>` condition isn't true.

danhoeflinger: > I mean that we can make local copy of `__my_unary_func_ ` if `std::is_invocable_v<const _UnaryFunc, typename std::iterator_traits<_Iter>::reference>` condition isn't true.

I would argue that we specifically do not want to give the impression that we support such functors which are not const callable.  My understanding of our initial reason for making the functor itself `const` was to portray this idea that the functor should be stateless and without side effects.  I see this `static_assert` as helping to help maintain clarity in our portrayal of this expectation for the functors.

Practically, we could support such functors in the way you describe, but it wouldn't *actually* support any extra functionality which is allowed by having a non-const `operator()`.  (It also would add some requirements on copy constructability of the functors.)

SergeyKopienko: I understood, ok.

documentation/library_guide/parallel_api/iterators.rst
```diff
@@ -69,10 +69,18 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
 * ``transform_iterator``: an iterator defined over another iterator whose dereferenced value is the result
   of a function applied to the corresponding element of the original iterator. Both the type of the original
   iterator and the unary function applied during dereference operations are required template parameters of
-  the ``transform_iterator`` class. The ``transform_iterator`` class provides three constructors:
+  the ``transform_iterator`` class. 
+
+  The unary functor provided to a transform iterator should have a const call operator that takes a single argument
+  whose parameter type can be converted to from the reference type of the original iterator. The functor's call operator
```
mmichel11, line 75: Should the `can be converted to from the reference type` here be just `to`, `from`, or `to and from`?

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,10 +413,18 @@ class transform_iterator
...
+    //only enable this constructor if the unary functor is default constructible
+    template <typename _UnaryFuncLocal = _UnaryFunc,
+              ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFuncLocal>, int> = 0>
+    transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_()
     {
     }
+
+    transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(__it), __my_unary_func_(__unary_func) {}
+
     transform_iterator(const transform_iterator& __input) = default;
```
SergeyKopienko, line 428: Looks like `__input` param isn't required here.

danhoeflinger: removed.

documentation/library_guide/parallel_api/iterators.rst
danhoeflinger: reworded for clarity to :
"The unary functor provided to a transform iterator should have a const call operator which is callable with an
  argument of the reference type of the original iterator."

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -401,7 +401,10 @@ class transform_iterator
 {
   private:
     _Iter __my_it_;
-    const _UnaryFunc __my_unary_func_;
+    _UnaryFunc __my_unary_func_;
+
+    static_assert(std::is_invocable_v<const _UnaryFunc, typename std::iterator_traits<_Iter>::reference>,
```
rarutyun, line 406: We don't support `std::invoke` semantics (though, we could) for `operator*`. So, this `static_assert` may return you `true` for wider amount of scenarios that we actually support. I am not sure if it's a problem or not.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,11 +413,19 @@ class transform_iterator
...
 
-    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc())
-        : __my_it_(__it), __my_unary_func_(__unary_func)
+    //default constructor will only be present if both the unary functor and iterator are default constructible
+    transform_iterator() = default;
+
+    //only enable this constructor if the unary functor is default constructible
+    template <typename _UnaryFuncLocal = _UnaryFunc,
+              ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFuncLocal>, int> = 0>
+    transform_iterator(_Iter __it) : __my_it_(__it), __my_unary_func_()
```
rarutyun, line 422: Technically, we can do `std::move`. And you don't actually need `__my_unary_func_()`

```suggestion
    transform_iterator(_Iter __it) : __my_it_(::std::move(__it))
```

include/oneapi/dpl/pstl/iterator_impl.h
rarutyun, line 406: Do we support function as the ref?

In other words: Does `transform_iterator<whatever_it, whatever_callable&>` work? If yes, my understanding is that the `static_assert` is broken.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,11 +413,19 @@ class transform_iterator
...
     {
     }
-    transform_iterator(const transform_iterator& __input) = default;
+
+    transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(__it), __my_unary_func_(__unary_func) {}
```
rarutyun, line 426:
```suggestion
    transform_iterator(_Iter __it, _UnaryFunc __unary_func) : __my_it_(::std::move__it)), __my_unary_func_(::std::move(__unary_func)) {}
```

include/oneapi/dpl/pstl/iterator_impl.h
rarutyun: > oneapi::dpl::__internal::__is_const_callable_object_v, does not work properly

I agree. It's broken. I forgot how exactly broken, but broken

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: This is an interesting question... 

Before the addition of the `std::move` to the ctors, it seems that this "works" in practice to use `whatever_callable&`, at least on the host.  Further, this circumvents the `const` on the `operator*` call, because if we just have a ref as the type, the ref wont change even if its data inside does change, so `const` is no longer required for its `operator()` by our code.  

If anyone was previously using a ref type as the functor, adding `std::move` to the constructors seems to be a breaking change. 

Looking at the documentation and spec, there is nothing that specifically disallows using a `whatever_callable&`, but we do talk about it as a "unary functor/function".  It seems to be at least a stretch, and probably not something we want to support.  I would expect it is problematic for use on the device. 

Whether or not we add `std::move`, I think its best to add `std::decay_t<_UnaryFunc>` to the `static_assert` to make sure that our intended restriction is enforced.  

Note: 
For someone to be using a ref type they would have to do so explicitly by defining the `transform_iterator` template argument as a ref.  Sending a ref in to the constructor with deduction or to `make_transform_iterator` results in the decayed type being used.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Added, but see my note about this possibly being a breaking change:  
https://github.com/oneapi-src/oneDPL/pull/1445/files#r1535775332

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Added, but see my note about this possibly being a breaking change:  
https://github.com/oneapi-src/oneDPL/pull/1445/files#r1535775332

documentation/library_guide/parallel_api/iterators.rst
```diff
@@ -69,10 +69,18 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
 * ``transform_iterator``: an iterator defined over another iterator whose dereferenced value is the result
   of a function applied to the corresponding element of the original iterator. Both the type of the original
   iterator and the unary function applied during dereference operations are required template parameters of
-  the ``transform_iterator`` class. The ``transform_iterator`` class provides three constructors:
+  the ``transform_iterator`` class. 
+
+  The unary functor provided to a transform iterator should have a const call operator which is callable with an
```
rarutyun, line 74:
```suggestion
  The unary functor provided to a ``transform_iterator`` should have a ``const``-qualified call operator which is callable with an
```

documentation/library_guide/parallel_api/iterators.rst
```diff
@@ -69,10 +69,18 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
...
+  argument of the reference type of the original iterator. The functor's call operator should not have any side
```
rarutyun, line 75: > an argument of the reference type of the original iterator...

That is something I don't understand

documentation/library_guide/parallel_api/iterators.rst
```diff
@@ -69,10 +69,18 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
...
+  effects and should not modify the state of the functor object.
+  
+  The ``transform_iterator`` class provides three constructors:
 
   * ``transform_iterator()``: instantiates the iterator using a default constructed base iterator and unary functor.
-  * ``transform_iterator(iter)``: instantiates the iterator using the base iterator provided and a default constructed unary functor.
+  This constructor is only provided if the base iterator and unary functor are both default constructible.
```
rarutyun, line 81:
```suggestion
  This constructor participates in overload resolution only if the base iterator and unary functor are both default constructible.
```

documentation/library_guide/parallel_api/iterators.rst
```diff
@@ -69,10 +69,18 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
...
+  * ``transform_iterator(iter)``: instantiates the iterator using the base iterator provided and a default constructed
+  unary functor. This constructor is only provided if the unary functor is default constructible.
```
rarutyun, line 83:
```suggestion
  unary functor. This constructor participates in overload resolution only if the unary functor is default constructible.
```

test/parallel_api/iterator/iterators.pass.cpp
```diff
@@ -43,7 +43,9 @@ void test_random_iterator(const RandomIt& it) {
         [[maybe_unused]] auto t4 = typename RandomIt::iterator_category{};
     }
 
-    static_assert(::std::is_default_constructible_v<RandomIt>, "iterator is not default constructible");
+    //Skips check of default constructible which is required by LegacyRandomAccessIterator, but not by oneDPL.
```
rarutyun, line 46:
```suggestion
    // Skips check of default constructible which is required by LegacyRandomAccessIterator, but not by oneDPL.
```

rarutyun (commented):
Overall, I like the current state of the PR much more that previously. I should be able to review it more carefully next week if necessary, but I believe it's in a pretty good shape to be merged to main.

test/parallel_api/iterator/transform_iterator_constructible.pass.cpp
```diff
@@ -0,0 +1,87 @@
...
+    oneapi::dpl::transform_iterator<int*, decltype(transformation)> trans1{ptr, transformation};
+    //default constructibility of lambdas depends on c++ standard, we want transform iterator to match its template args
+    static_assert((std::is_default_constructible_v<decltype(transformation)> ==
+                   ::std::is_default_constructible_v<decltype(trans1)>),
+                  "transform_iterator with lambda does not match default constructibility status of the lambda itself");
+
+    //both types are default constructible
+    oneapi::dpl::transform_iterator<int*, noop> trans2{ptr, noop{}};
+    static_assert(::std::is_default_constructible_v<decltype(trans2)>,
+                  "transform_iterator with default constructible lambda is seen to be non-default constructible");
```
danhoeflinger, line 61:
```suggestion
                  "transform_iterator with default constructible unary functor is seen to be non-default constructible");
``
Note to self: fix this and similar mistakes in assert messages below.

File: documentation/library_guide/parallel_api/iterators.rst (hunk 19 above)
Comment (danhoeflinger) [reply documentation/library_guide/parallel_api/iterators.rst:75]: Basically, I mean to say that we want `__my_unary_func_(*__my_it_)` to be a valid call.  
`*__my_it_` returns a reference to an element of the original iterator, we want to be able to use that to call the callable (directly or via some conversion sequence).

Perhaps the sentence should instead read: 
"The unary functor provided to a ``transform_iterator`` should have a ``const``-qualified call operator that can be called using the reference type of the original iterator as argument."

File: test/parallel_api/iterator/transform_iterator_constructible.pass.cpp (hunk 23 above)
Comment (danhoeflinger) [reply test/parallel_api/iterator/transform_iterator_constructible.pass.cpp]: fixed

File: test/parallel_api/iterator/iterators.pass.cpp (hunk 22 above)
Comment (danhoeflinger) [reply test/parallel_api/iterator/iterators.pass.cpp]: fixed

File: documentation/library_guide/parallel_api/iterators.rst (hunk 21 above)
Comment (danhoeflinger) [reply documentation/library_guide/parallel_api/iterators.rst]: fixed

File: documentation/library_guide/parallel_api/iterators.rst (hunk 20 above)
Comment (danhoeflinger) [reply documentation/library_guide/parallel_api/iterators.rst]: fixed

File: documentation/library_guide/parallel_api/iterators.rst (hunk 18 above)
Comment (danhoeflinger) [reply documentation/library_guide/parallel_api/iterators.rst]: fixed

File: documentation/library_guide/parallel_api/iterators.rst (hunk 19 above)
Comment (danhoeflinger) [reply documentation/library_guide/parallel_api/iterators.rst]: I adjusted it as mentioned above, let me know if it still requires clarification.

**Review by danhoeflinger — COMMENTED** [review]
from offline review with @rarutyun

File: test/parallel_api/iterator/iterators.pass.cpp [hunk 24]
```diff
@@ -43,7 +43,9 @@ void test_random_iterator(const RandomIt& it) {
         [[maybe_unused]] auto t4 = typename RandomIt::iterator_category{};
     }
 
-    static_assert(::std::is_default_constructible_v<RandomIt>, "iterator is not default constructible");
```
rarutyun, line 46: I am not 100% sure this is a safe change to do.

test/parallel_api/iterator/transform_iterator_constructible.pass.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+    oneapi::dpl::transform_iterator<oneapi::dpl::counting_iterator<int>, noop> trans2{count, noop{}};
+    static_assert(::std::is_copy_assignable_v<decltype(trans2)>,
+                  "transform_iterator with noop functor is not copy assignable");
+
+    oneapi::dpl::transform_iterator<oneapi::dpl::counting_iterator<int>, stateful_functor> trans3{count, stateful_functor{1}};
+    static_assert(::std::is_copy_assignable_v<decltype(trans3)>,
+                  "transform_iterator with stateful functor is not copy assignable");
+
+    oneapi::dpl::transform_iterator<oneapi::dpl::counting_iterator<int>, stateful_functor> trans4{count, stateful_functor{0}};
```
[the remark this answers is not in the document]
danhoeflinger: Good point, adjusted to distinguish from default constructed value for state.

test/parallel_api/iterator/iterators.pass.cpp
danhoeflinger: As per our offline discussion, I've adjusted the code to have a pair of functions, one which wraps the other:
`test_random_iterator` - does a full check of random access iterator
`test_random_iterator_skip_default_ctor_check` - skips default constructor check

For a `permutation_iterator` based on a lambda index:
  C++17 and earlier, we skip default ctor check. For greater than C++17, we check default ctor.

documentation/library_guide/parallel_api/iterators.rst
```diff
@@ -67,16 +67,24 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
...
+  of a function applied to the corresponding element of the base iterator. Both the type of the base
   iterator and the unary function applied during dereference operations are required template parameters of
-  the ``transform_iterator`` class. The ``transform_iterator`` class provides three constructors:
+  the ``transform_iterator`` class. 
+
+  The unary functor provided to a ``transform_iterator`` should have a ``const``-qualified call operator which accepts
+  the reference type of the base iterator as argument. The functor's call operator should not have any side effects and
+  should not modify the state of the functor object.
+  
+  The ``transform_iterator`` class provides three constructors:
```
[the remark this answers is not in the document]
danhoeflinger: incorporated this change.

test/parallel_api/iterator/iterators.pass.cpp
```diff
@@ -319,7 +325,13 @@ struct test_permutation_iterator
         auto perm_it_fun_rev = oneapi::dpl::make_permutation_iterator(in1.begin(), [n] (auto i) { return n - i - 1;}, 1);
         EXPECT_TRUE(*++perm_it_fun_rev == *(in1.end()-3), "wrong result from permutation_iterator(base_iterator, functor)");
 
+#if TEST_STD_VER <= 17
+        // c++17 lambdas are not default constructible, therefore this permutation_iterator is not either
+        test_random_iterator_skip_default_ctor_check(perm_it_fun_rev);
+#else // c++20 or later
+        // starting at c++20, lambdas are default constructible, therefore this permutation_iterator should be as well
         test_random_iterator(perm_it_fun_rev);
```
rarutyun, line 333: Sorry, I don't get it. How stateful lambda can be default constructible? I might misunderstand something but it's likely that we miss the failure of our test because we don't have C++20 in our CI

danhoeflinger: Thanks, my mistake.  Fixed to always skip default ctor check for stateful lambda and added a check for stateless lambda which checks only if lambda is default constructible.

rarutyun (approved):
It was a long journey but now the PR is in a very good shape to me. I really enjoyed polishing the code to the state where we are right now. I hope you also happy with the result.

I put some small comments but those are not necessary to be fixed.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,19 +414,36 @@ class transform_iterator
...
-        // copyable makes their device_copyable trait deprecated in SYCL2020. However, defaulting this function implies
-        // an extra requirement that __my_unary_func_ implements a copy assignment operator.
         __my_it_ = __input.__my_it_;
+
+        // If copy assignment is available, copy the functor, otherwise skip it.
+        // For non-copy assignable functors, this copy assignment operator departs from the sycl 2020 specification
+        // requirement of device copyable types for copy assignment to be the same as a bitwise copy of the object.
+        // TODO: Explore (ABI breaking) change to use std::optional or similar and using copy constructor to implement
+        //       copy assignment to better comply with SYCL 2020 specification.
+        if constexpr (std::is_copy_assignable_v<_UnaryFunc>)
```
SergeyKopienko, line 443: In the opposite case, should we initialize `__my_unary_func_`  ?

danhoeflinger: Maybe I'm missing something, but I'm not sure how we would do that.  We are in the copy assignment operator, and it is not copy assignable.  
We can't construct it without the `std::optional` -like mechanism, which we have decided not to pursue right now.

Do you have another method in mind?

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -410,19 +414,36 @@ class transform_iterator
...
     typedef typename ::std::iterator_traits<_Iter>::iterator_category iterator_category;
 
-    transform_iterator(_Iter __it = _Iter(), _UnaryFunc __unary_func = _UnaryFunc())
-        : __my_it_(__it), __my_unary_func_(__unary_func)
+    //default constructor will only be present if both the unary functor and iterator are default constructible
+    transform_iterator() = default;
+
+    //only enable this constructor if the unary functor is default constructible
+    template <typename _UnaryFuncLocal = _UnaryFunc,
+              ::std::enable_if_t<::std::is_default_constructible_v<_UnaryFuncLocal>, int> = 0>
```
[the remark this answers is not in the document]
danhoeflinger: Yes, I've tried to switch to `std` without `::` generally in my new code, but missed some here.  I'll adjust this one quickly to follow that.

test/parallel_api/iterator/iterators.pass.cpp
```diff
@@ -310,10 +322,22 @@ struct test_permutation_iterator
         test_random_iterator(perm_begin);
 
         auto n = in1.size();
-        auto perm_it_fun_rev = oneapi::dpl::make_permutation_iterator(in1.begin(), [n] (auto i) { return n - i - 1;}, 1);
+        auto stateful_lambda = [n] (auto i) { return n - i - 1;};
```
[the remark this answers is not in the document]
danhoeflinger: Yes, I changed it for symmetry and for documentation that the lambda is "stateful", and to show it is the novel thing we are testing.

include/oneapi/dpl/pstl/iterator_impl.h
SergeyKopienko: I think that we can try to do something like this:
```C++
        if constexpr (std::is_copy_assignable_v<_UnaryFunc>)
        {
            __my_unary_func_ = __input.__my_unary_func_;
        }
        else
        {
            __my_unary_func_ = _UnaryFunc{};
        }
```

danhoeflinger: Isn't that using the copy assignment?

https://godbolt.org/z/nGTGe4rqE

danhoeflinger: Technically, we could do something like in the following link (check default constructibility and move assignability, and use them if avail).  However, this seems like it only serves a very unlikely scenario with an unexpected result (default construction rather than copy within the copy assignment operator).  I think that we are better off just doing nothing when we are missing copy assignment.

https://godbolt.org/z/vbzhGK9T1

SergeyKopienko: @dan, of course, it's up to you, Bit I think that implementation of `operator=` where something assigned and other part sometimes save their state looks like a little bit incorrect because you allow through constructor to have state for this "unassigned" part.

danhoeflinger: @rarutyun and I discussed this in some detail offline, and I perhaps didn't fully describe that here in the PR comments.  I will try to state some of that here.

Some other options:
1) leave as is  (worse for copy assignable functors with regard to device copyable sycl spec)
2) delete copy assignment when functor has no copy assignment (potential for user code breaking with c++17 lambdas, see #1437)
3) std::optional (ABI breaking)

We are not adding this operator as new functionality in this PR, its previous state was to *always* skip copying of the functor. In fact, we are shrinking the number of cases which have this skipped copy of the functor (prior to the PR, all functors skip this copy).  The PR remedies many cases where copy assignment is available, and makes them in alignment with the SYCL specification for device copyable.

This change also does not break ABI, or break any existing user code (the one exception could be users which depend upon the functor *not* being copied even when it is copy assignable, but that seems far-fetched).

We are not preventing future changes as indicated in the "TODO" to use something like a `std::optional` and using copy construction to implement copy assignment where it is possible.  However, it seems that such a change would require more thought and is ABI breaking.

I agree with your point that this unbalanced copy assignment is not ideal, but it seems the best option at the current time with the context we have.

SergeyKopienko: ok. Thank you for this clarification.

SergeyKopienko (approved):
LGTM

PR #1446, opened 2024-03-13
Tune vectorized binary_search, lower_bound, and upper_bound algorithms

mmichel11:
This PR introduces some additional tuning for the vectorized search algorithms (binary_search, lower_bound, and upper_bound) where a series of searches across a provided set of search keys are performed on a single input buffer.

The improvements are explained below:

1. In the current implementations, `__pstl_lower_bound`  / `__pstl_upper_bound` have been passed a 64-bit size type to compute indicies in its binary search. For inputs with a large number of search keys, this is a bottleneck in performance and rooflines have shown that this kernel is bound by 64-bit arithmetic. This PR introduces a 2-kernel solution where a 32-bit unsigned type is used for index calculation if there are less than ``2^32`` search keys with a fallback to 64-bit offset calculation otherwise. This results in a performance improvement across these algorithms of `~1.3x` on GPU Series Max 1550 for large inputs.
2. The second improvement is in the implementation of `__pstl_lower_bound`. Simplifying the loop body by checking descending powers-of-2 as opposed to continuously dividing the input size by `2` reduces the amount of total arithmetic in the compiled code and results in a performance improvement of `10-15%` for large inputs on top of the changes in `1`. This improvement requires the size type to be unsigned and is implemented as a new utility function: `__shars_lower_bound`.

### DIFF
--- a/include/oneapi/dpl/internal/binary_search_impl.h
+++ b/include/oneapi/dpl/internal/binary_search_impl.h
@@ -230,7 +230,7 @@ binary_search_impl(__internal::__hetero_tag<_BackendTag>, Policy&& policy, Input
     auto keep_result = oneapi::dpl::__ranges::__get_sycl_range<__bknd::access_mode::read_write, OutputIterator>();
     auto result_buf = keep_result(result, result + value_size);
     auto zip_vw = make_zip_view(input_buf.all_view(), value_buf.all_view(), result_buf.all_view());
-    auto run_binary_search = [&comp, &value_size, &zip_vw](auto&& policy, auto size) {
+    auto run_binary_search = [comp, value_size, zip_vw](auto&& policy, auto size) {
         __bknd::__parallel_for(
             _BackendTag{}, ::std::forward<decltype(policy)>(policy),
             custom_brick<StrictWeakOrdering, decltype(size), search_algorithm::binary_search>{comp, size}, value_size,
--- a/include/oneapi/dpl/internal/binary_search_impl.h
+++ b/include/oneapi/dpl/internal/binary_search_impl.h
@@ -76,7 +76,7 @@ struct custom_brick
         if (use_32bit_indexing)
             search_impl<std::uint32_t>(idx, acc);
         else
-            search_impl<std::size_t>(idx, acc);
+            search_impl<std::uint64_t>(idx, acc);
     }
 };
 
--- a/include/oneapi/dpl/internal/binary_search_impl.h
+++ b/include/oneapi/dpl/internal/binary_search_impl.h
@@ -51,18 +51,18 @@ struct custom_brick
         using ::std::get;
         if constexpr (func == search_algorithm::lower_bound)
         {
-            get<2>(acc[idx]) = oneapi::dpl::internal::lower_bound_fun(get<0>(acc.tuple()), start_orig, end_orig,
-                                                                      get<1>(acc[idx]), comp);
+            get<2>(acc[idx]) = oneapi::dpl::__internal::__pstl_lower_bound(get<0>(acc.tuple()), start_orig, end_orig,
+                                                                           get<1>(acc[idx]), comp);
         }
-        else if (func == search_algorithm::upper_bound)
+        else if constexpr (func == search_algorithm::upper_bound)
         {
-            get<2>(acc[idx]) = oneapi::dpl::internal::upper_bound_fun(get<0>(acc.tuple()), start_orig, end_orig,
-                                                                      get<1>(acc[idx]), comp);
+            get<2>(acc[idx]) = oneapi::dpl::__internal::__pstl_upper_bound(get<0>(acc.tuple()), start_orig, end_orig,
+                                                                           get<1>(acc[idx]), comp);
         }
         else
         {
-            auto value = oneapi::dpl::internal::lower_bound_fun(get<0>(acc.tuple()), start_orig, end_orig,
-                                                                get<1>(acc[idx]), comp);
+            auto value = oneapi::dpl::__internal::__pstl_lower_bound(get<0>(acc.tuple()), start_orig, end_orig,
+                                                                     get<1>(acc[idx]), comp);
             get<2>(acc[idx]) = (value != end_orig) && (get<1>(acc[idx]) == get<0>(acc[value]));
         }
     }
--- a/include/oneapi/dpl/internal/binary_search_impl.h
+++ b/include/oneapi/dpl/internal/binary_search_impl.h
@@ -149,8 +149,8 @@ lower_bound_impl(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIt
     const bool use_32bit_indexing = size <= std::numeric_limits<std::uint32_t>::max();
     __bknd::__parallel_for(
         _BackendTag{}, ::std::forward<decltype(policy)>(policy),
-        custom_brick<StrictWeakOrdering, decltype(size), search_algorithm::lower_bound>{comp, size, use_32bit_indexing}, value_size,
-        zip_vw)
+        custom_brick<StrictWeakOrdering, decltype(size), search_algorithm::lower_bound>{comp, size, use_32bit_indexing},
+        value_size, zip_vw)
         .wait();
     return result + value_size;
 }
@@ -181,8 +181,8 @@ upper_bound_impl(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIt
     const bool use_32bit_indexing = size <= std::numeric_limits<std::uint32_t>::max();
     __bknd::__parallel_for(
         _BackendTag{}, std::forward<decltype(policy)>(policy),
-        custom_brick<StrictWeakOrdering, decltype(size), search_algorithm::upper_bound>{comp, size, use_32bit_indexing}, value_size,
-        zip_vw)
+        custom_brick<StrictWeakOrdering, decltype(size), search_algorithm::upper_bound>{comp, size, use_32bit_indexing},
+        value_size, zip_vw)
         .wait();
     return result + value_size;
 }
@@ -211,10 +211,10 @@ binary_search_impl(__internal::__hetero_tag<_BackendTag>, Policy&& policy, Input
     auto result_buf = keep_result(result, result + value_size);
     auto zip_vw = make_zip_view(input_buf.all_view(), value_buf.all_view(), result_buf.all_view());
     const bool use_32bit_indexing = size <= std::numeric_limits<std::uint32_t>::max();
-    __bknd::__parallel_for(
-        _BackendTag{}, std::forward<decltype(policy)>(policy),
-        custom_brick<StrictWeakOrdering, decltype(size), search_algorithm::binary_search>{comp, size, use_32bit_indexing}, value_size,
-        zip_vw)
+    __bknd::__parallel_for(_BackendTag{}, std::forward<decltype(policy)>(policy),
+                           custom_brick<StrictWeakOrdering, decltype(size), search_algorithm::binary_search>{
+                               comp, size, use_32bit_indexing},
+                           value_size, zip_vw)
         .wait();
     return result + value_size;
 }
--- a/include/oneapi/dpl/internal/binary_search_impl.h
+++ b/include/oneapi/dpl/internal/binary_search_impl.h
@@ -137,20 +137,22 @@ lower_bound_impl(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIt
     auto keep_result = oneapi::dpl::__ranges::__get_sycl_range<__bknd::access_mode::read_write, OutputIterator>();
     auto result_buf = keep_result(result, result + value_size);
     auto zip_vw = make_zip_view(input_buf.all_view(), value_buf.all_view(), result_buf.all_view());
-    // Enable index calculation to proceed with uint32_t if input range is small enough. 
+    // Enable index calculation to proceed with uint32_t if input range is small enough.
     if (size <= ::std::numeric_limits<::std::uint32_t>::max())
     {
         __bknd::__parallel_for(_BackendTag{}, ::std::forward<Policy>(policy),
-                               custom_brick<StrictWeakOrdering, ::std::uint32_t, search_algorithm::lower_bound>{comp, static_cast<::std::uint32_t>(size)}, value_size,
-                               zip_vw)
+                               custom_brick<StrictWeakOrdering, ::std::uint32_t, search_algorithm::lower_bound>{
+                                   comp, static_cast<::std::uint32_t>(size)},
+                               value_size, zip_vw)
             .wait();
     }
     else
     {
         auto fallback_policy = __bknd::make_wrapped_policy<lower_bound_fallback>(::std::forward<Policy>(policy));
         __bknd::__parallel_for(_BackendTag{}, ::std::move(fallback_policy),
-                               custom_brick<StrictWeakOrdering, ::std::size_t, search_algorithm::lower_bound>{comp, static_cast<::std::size_t>(size)}, value_size,
-                               zip_vw)
+                               custom_brick<StrictWeakOrdering, ::std::size_t, search_algorithm::lower_bound>{
+                                   comp, static_cast<::std::size_t>(size)},
+                               value_size, zip_vw)
             .wait();
     }
     return result + value_size;
@@ -181,20 +183,22 @@ upper_bound_impl(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIt
     auto keep_result = oneapi::dpl::__ranges::__get_sycl_range<__bknd::access_mode::read_write, OutputIterator>();
     auto result_buf = keep_result(result, result + value_size);
     auto zip_vw = make_zip_view(input_buf.all_view(), value_buf.all_view(), result_buf.all_view());
-    // Enable index calculation to proceed with uint32_t if input range is small enough. 
+    // Enable index calculation to proceed with uint32_t if input range is small enough.
     if (size <= ::std::numeric_limits<::std::uint32_t>::max())
     {
         __bknd::__parallel_for(_BackendTag{}, ::std::forward<Policy>(policy),
-                               custom_brick<StrictWeakOrdering, ::std::uint32_t, search_algorithm::upper_bound>{comp, static_cast<::std::uint32_t>(size)}, value_size,
-                               zip_vw)
+                               custom_brick<StrictWeakOrdering, ::std::uint32_t, search_algorithm::upper_bound>{
+                                   comp, static_cast<::std::uint32_t>(size)},
+                               value_size, zip_vw)
             .wait();
     }
     else
     {
         auto fallback_policy = __bknd::make_wrapped_policy<upper_bound_fallback>(::std::forward<Policy>(policy));
         __bknd::__parallel_for(_BackendTag{}, ::std::move(fallback_policy),
-                               custom_brick<StrictWeakOrdering, ::std::size_t, search_algorithm::upper_bound>{comp, static_cast<::std::size_t>(size)}, value_size,
-                               zip_vw)
+                               custom_brick<StrictWeakOrdering, ::std::size_t, search_algorithm::upper_bound>{
+                                   comp, static_cast<::std::size_t>(size)},
+                               value_size, zip_vw)
             .wait();
     }
     return result + value_size;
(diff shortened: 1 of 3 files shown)


SergeyKopienko:
@mmichel11, let's merge this PR only after https://github.com/oneapi-src/oneDPL/pull/1239, ok?

mmichel11:
> @mmichel11, let's merge this PR only after #1239, ok?

Sounds good. I will not merge it until after the tag dispatching PR and rebase.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -717,6 +717,41 @@ struct __spirv_target_conditional :
 // to determine SPIR-V targets.
 inline constexpr bool __is_spirv_target_v = __spirv_target_conditional<::std::true_type, ::std::false_type>::value;
 
+// Lower bound implementation based on Shar's algorithm for binary search.
+template <typename _Acc, typename _Size1, typename _Value, typename _Compare>
+_Size1
+__shars_lower_bound(_Acc __acc, _Size1 __first, _Size1 __last, const _Value& __value, _Compare __comp)
+{
+    static_assert(::std::is_unsigned_v<_Size1>, "__shars_lower_bound requires an unsigned size type");
+    _Size1 __n = __last - __first;
```
SergeyKopienko, line 726: const ?

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -188,10 +226,25 @@ binary_search_impl(Policy&& policy, InputIterator1 start, InputIterator1 end, In
...
+    {
+        __bknd::__parallel_for(::std::forward<Policy>(policy),
+                               custom_brick<StrictWeakOrdering, ::std::uint32_t, search_algorithm::binary_search>{
+                                   comp, static_cast<::std::uint32_t>(size)},
+                               value_size, zip_vw)
+            .wait();
+    }
+    else
+    {
+        auto fallback_policy = __bknd::make_wrapped_policy<binary_search_fallback>(::std::forward<Policy>(policy));
```
SergeyKopienko, line 241: Do we really need to have `fallback_policy` as local variable?

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -128,13 +132,30 @@ lower_bound_impl(Policy&& policy, InputIterator1 start, InputIterator1 end, Inpu
...
     auto result_buf = keep_result(result, result + value_size);
     auto zip_vw = make_zip_view(input_buf.all_view(), value_buf.all_view(), result_buf.all_view());
-    __bknd::__parallel_for(::std::forward<Policy>(policy),
-                           custom_brick<StrictWeakOrdering, decltype(size), lower_bound>{comp, size}, value_size,
-                           zip_vw)
-        .wait();
+    // Enable index calculation to proceed with uint32_t if input range is small enough.
+    if (size <= ::std::numeric_limits<::std::uint32_t>::max())
+    {
+        __bknd::__parallel_for(::std::forward<Policy>(policy),
```
SergeyKopienko, line 138: The code in these two branches looks like the same excepting used policy.
May we combine it into one branch?

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -158,13 +179,30 @@ upper_bound_impl(Policy&& policy, InputIterator1 start, InputIterator1 end, Inpu
...
     auto result_buf = keep_result(result, result + value_size);
     auto zip_vw = make_zip_view(input_buf.all_view(), value_buf.all_view(), result_buf.all_view());
-    __bknd::__parallel_for(::std::forward<Policy>(policy),
-                           custom_brick<StrictWeakOrdering, decltype(size), upper_bound>{comp, size}, value_size,
-                           zip_vw)
-        .wait();
+    // Enable index calculation to proceed with uint32_t if input range is small enough.
+    if (size <= ::std::numeric_limits<::std::uint32_t>::max())
+    {
+        __bknd::__parallel_for(::std::forward<Policy>(policy),
```
SergeyKopienko, line 185: The same

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -188,10 +226,25 @@ binary_search_impl(Policy&& policy, InputIterator1 start, InputIterator1 end, In
...
     auto zip_vw = make_zip_view(input_buf.all_view(), value_buf.all_view(), result_buf.all_view());
-    __bknd::__parallel_for(::std::forward<Policy>(policy),
-                           custom_brick<StrictWeakOrdering, decltype(size), binary_search>{comp, size}, value_size,
-                           zip_vw)
-        .wait();
+
+    // Enable index calculation to proceed with uint32_t if input range is small enough.
+    if (size <= ::std::numeric_limits<::std::uint32_t>::max())
+    {
+        __bknd::__parallel_for(::std::forward<Policy>(policy),
```
SergeyKopienko, line 233: The same

include/oneapi/dpl/internal/binary_search_impl.h
mmichel11: I don't think it is possible here. In the first branch `custom_brick` uses a `::std::uint32_t` index type and a `::std::size_t` in the second branch.  We have to make the dispatch decision at runtime to choose the correct kernel.

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -164,13 +184,30 @@ upper_bound_impl(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIt
...
     auto result_buf = keep_result(result, result + value_size);
     auto zip_vw = make_zip_view(input_buf.all_view(), value_buf.all_view(), result_buf.all_view());
-    __bknd::__parallel_for(_BackendTag{}, ::std::forward<Policy>(policy),
-                           custom_brick<StrictWeakOrdering, decltype(size), upper_bound>{comp, size}, value_size,
-                           zip_vw)
-        .wait();
+    // Enable index calculation to proceed with uint32_t if input range is small enough.
+    if (size <= ::std::numeric_limits<::std::uint32_t>::max())
+    {
+        __bknd::__parallel_for(_BackendTag{}, ::std::forward<Policy>(policy),
```
danhoeflinger, line 190: Seems like we can reduce this to a lambda and make the branches just provide `policy` and `size` of the appropriate type.

It adds the complexity of the lambda to reduce the code replication, so its not a clear win, but I think it would be better.
(same comment for all three instances)

include/oneapi/dpl/pstl/utils.h
```diff
@@ -717,6 +717,41 @@ struct __spirv_target_conditional :
...
+            __search_offset |= __cur_pow2;
+    }
+    return __shifted_first + __search_offset;
+}
+
+template <typename _Acc, typename _Size1, typename _Value, typename _Compare>
+_Size1
+__shars_upper_bound(_Acc __acc, _Size1 __first, _Size1 __last, const _Value& __value, _Compare __comp)
+{
+    return __pstl_lower_bound(__acc, __first, __last, __value,
```
danhoeflinger, line 750: should this use `__shars_lower_bound` rather than `__pstl_lower_bound`?

julianmi (commented):
These changes make sense to me. There might be potential benefits of utilizing 32-bit addressing on other algorithms as well. This is beyond this PR but maybe something to keep in mind.

danhoeflinger:
> There might be potential benefits of utilizing 32-bit addressing on other algorithms as well. 

I agree with this, but I also feel like this may be something to take up with the `sycl` runtime folks.  Its not a good user experience to need to have this sort of logic to achieve good performance (and double the number of compiled kernels).  Of course, we do not want to have the runtime automatically double the number of kernels for us, but they may be able to improve the performance of the 64-bit version somehow.

Another option @mmichel11 and I were brainstorming offline:

Adding some opt-in define which would use 32-bit addressing everywhere, where the user guarantees they wont be calling the algorithms for `n > max(uint32_t)`, and in return gets a performance boost across the board.

julianmi:
> > There might be potential benefits of utilizing 32-bit addressing on other algorithms as well.
> 
> I agree with this, but I also feel like this may be something to take up with the `sycl` runtime folks. Its not a good user experience to need to have this sort of logic to achieve good performance (and double the number of compiled kernels). Of course, we do not want to have the runtime automatically double the number of kernels for us, but they may be able to improve the performance of the 64-bit version somehow.
> 
> Another option @mmichel11 and I were brainstorming offline:
> 
> Adding some opt-in define which would use 32-bit addressing everywhere, where the user guarantees they wont be calling the algorithms for `n > max(uint32_t)`, and in return gets a performance boost across the board.

I agree with this and especially the opt-in define. Many of our size-related types are derived from SYCL's `size()` or previously `get_count()` which seem to use 64-bits. We could cast them down if the user permits. We should probably investigate the performance benefits in more detail.

include/oneapi/dpl/pstl/utils.h
mmichel11: Yes, I have made this change.

include/oneapi/dpl/internal/binary_search_impl.h
mmichel11: Changed for all three instances to use a lambda.

include/oneapi/dpl/internal/binary_search_impl.h
mmichel11: No, I have changed this.

include/oneapi/dpl/internal/binary_search_impl.h
mmichel11: Changed

include/oneapi/dpl/pstl/utils.h
```diff
@@ -717,6 +717,41 @@ struct __spirv_target_conditional :
 // to determine SPIR-V targets.
 inline constexpr bool __is_spirv_target_v = __spirv_target_conditional<::std::true_type, ::std::false_type>::value;
 
+// Lower bound implementation based on Shar's algorithm for binary search.
+template <typename _Acc, typename _Size1, typename _Value, typename _Compare>
+_Size1
```
adamfidel, line 722: Is there a reason to use the name `_Size1` instead of `_Size`? I don't see a `_Size2`.

mmichel11: No, I just followed what was being done in `__pstl_lower_bound`. It is confusing, so I have changed it to `_Size`.

mmichel11:
> > > There might be potential benefits of utilizing 32-bit addressing on other algorithms as well.
> > 
> > 
> > I agree with this, but I also feel like this may be something to take up with the `sycl` runtime folks. Its not a good user experience to need to have this sort of logic to achieve good performance (and double the number of compiled kernels). Of course, we do not want to have the runtime automatically double the number of kernels for us, but they may be able to improve the performance of the 64-bit version somehow.
> > Another option @mmichel11 and I were brainstorming offline:
> > Adding some opt-in define which would use 32-bit addressing everywhere, where the user guarantees they wont be calling the algorithms for `n > max(uint32_t)`, and in return gets a performance boost across the board.
> 
> I agree with this and especially the opt-in define. Many of our size-related types are derived from SYCL's `size()` or previously `get_count()` which seem to use 64-bits. We could cast them down if the user permits. We should probably investigate the performance benefits in more detail.

This is an investigation I plan to do in the future. There are several places now where we define separate kernels to take advantage of 32-bit index types when possible. I think an opt-in define would be the best solution instead of doubling the number of kernels per algorithm.

mmichel11:
After offline discussion with @danhoeflinger, we agreed it would be best to group the upper / lower bound implementations together. I have moved them after the bit algorithms as they are required to implement `__shars_lower_bound`.

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -134,13 +138,29 @@ lower_bound_impl(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIt
...
+    // Enable index calculation to proceed with uint32_t if input range is small enough.
+    if (size <= ::std::numeric_limits<::std::uint32_t>::max())
+    {
+        run_lower_bound(::std::forward<Policy>(policy), static_cast<::std::uint32_t>(size));
+    }
+    else
+    {
+        run_lower_bound(__bknd::make_wrapped_policy<lower_bound_fallback>(::std::forward<Policy>(policy)),
+                        static_cast<::std::size_t>(size));
+    }
```
akukanov, line 157: Have you tried to move the runtime decision to `custom_brick` instead? As far as I can see, the size value to make the decision is the same on the host and in the brick, so that should be technically possible. How would that perform?

mmichel11: I did some experimenting and there seems to be a small performance penalty to performing the check within the kernel itself (<1% for large n and large number of search keys). However, this allows us to have a compiled single kernel which is likely worth it.

I will do a bit more investigation of doing this runtime check in the kernel.

include/oneapi/dpl/pstl/utils.h
mmichel11: Missed responding to this. I made this variable `const` as well as others where applicable.

mmichel11:
@julianmi @SergeyKopienko @akukanov @danhoeflinger @adamfidel

Are there any additional comments / concerns you have for this PR, or are you ready to approve?

include/oneapi/dpl/pstl/utils.h
```diff
@@ -669,6 +559,151 @@ __dpl_ceiling_div(_T1 __number, _T2 __divisor)
...
+    // 2^(bit_floor(__n)) - 1 elements.
+    const _Size __shifted_first = __comp(__acc[__midpoint], __value) ? __n + 1 - __cur_pow2 : __first;
+    // Check descending powers of two. If __comp(__acc[__search_idx], __pow) holds for a __cur_pow2, then its
+    // bit must be set in the result.
+    _Size __search_offset{0};
+    for (__cur_pow2 /= 2; __cur_pow2 > 0; __cur_pow2 /= 2)
+    {
+        const _Size __search_idx = __shifted_first + (__search_offset | __cur_pow2) - 1;
+        if (__comp(__acc[__search_idx], __value))
+            __search_offset |= __cur_pow2;
```
SergeyKopienko, line 693: 1) Are we able to rewrite this like `__search_offset += __cur_pow2;` ?
2) What about remove this `if` ? 
```C++
__search_offset  += __comp(__acc[__search_idx], __value) ? __cur_pow2 : 0;
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -669,6 +559,151 @@ __dpl_ceiling_div(_T1 __number, _T2 __divisor)
...
+    }
+#endif
+};
+
+// Reverse searching for the first element strongly less than a passed value - left bound
+template <typename _Buffer, typename _Index, typename _Value, typename _Compare>
+_Index
+__pstl_left_bound(_Buffer& __a, _Index __first, _Index __last, const _Value& __val, _Compare __comp)
+{
+    auto __beg = _ReverseCounter<_Index, _Buffer>{__last - 1};
```
SergeyKopienko, line 666: Revere counter iterator looks like a little bit strange for me,
As far as I know we already have `oneapi::dpl::counting_iterator`. Is it not enough?
Also when I see the API like `_Index __first, _Index __last` I assume that the element before `__first` isn't exist and the element iterated by `__last` isn't exist too. From this point of view what does it mean `__first - 1` below?

@danhoeflinger what do you think about this?

include/oneapi/dpl/internal/binary_search_impl.h
SergeyKopienko: As described at https://en.cppreference.com/w/cpp/types/size_t, "The bit width of std::size_t is not less than 16.(since C++11)"
So does this solution quite correct when `size > ::std::numeric_limits<::std::uint32_t>::max()` ?
Is OK that n this case you simple using `static_cast<::std::size_t>(size)` ?
May be we should use some another type instead of `std::size_t` ?

include/oneapi/dpl/pstl/utils.h
mmichel11: For 1, using bitwise or proved to be slightly faster, so I think sticking with that is best.

I should be able to use your suggestion from 2 still. Let me try check performance with that change.

include/oneapi/dpl/internal/binary_search_impl.h
mmichel11: I see your point here. I had used `size_t` for the opposite case where it could holder larger values than `uint64_t`. In other places in the code where we do this, I see we use `std::uint64_t` as the fallback. 

Does switching to `std::uint64_t` seem like a good alternative to you?

include/oneapi/dpl/pstl/utils.h
mmichel11: I tried the approach you suggested with the ternary operator and a "no-op" bitwise or with 0 in the case the predicate is not satisfied. Interestingly, this ended up being slightly slower. It seems that the compiler is able to better optimize with the way it is currently written, so my thoughts are we leave it as is.

include/oneapi/dpl/pstl/utils.h
mmichel11: @SergeyKopienko and I discussed offline and agreed. This code is just being moved so that we can group these similar utilities together.

If there is an issue or room for improvement in `__pstl_left_bound`, then we can open a GH issue to address in a future PR.

include/oneapi/dpl/internal/binary_search_impl.h
SergeyKopienko: Probably yes.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: So, @mmichel11, could you please create the issue for this point?

SergeyKopienko:
@mmichel11 no any comments or concerns from my side. I am ready to approve this PR because I can't find something incorrect anymore. But please receive additional approve from somebody too,

include/oneapi/dpl/pstl/utils.h
```diff
@@ -669,6 +559,151 @@ __dpl_ceiling_div(_T1 __number, _T2 __divisor)
...
+        return __first;
+    _Size __cur_pow2 = __dpl_bit_floor(__n);
+    const _Size __midpoint = __n / 2;
+    // Check the middle element to determine if we should search the first or last
+    // 2^(bit_floor(__n)) - 1 elements.
+    const _Size __shifted_first = __comp(__acc[__midpoint], __value) ? __n + 1 - __cur_pow2 : __first;
+    // Check descending powers of two. If __comp(__acc[__search_idx], __pow) holds for a __cur_pow2, then its
+    // bit must be set in the result.
+    _Size __search_offset{0};
+    for (__cur_pow2 /= 2; __cur_pow2 > 0; __cur_pow2 /= 2)
```
danhoeflinger, line 689: My assumption is that the compiler will always replace these divisions by two with shift right by one.  You've probably checked performance comparison, if not it could be worth a look.

So if they are the same performance, then we have to decide which is more clear to read.  My preference for the ones in this loop would be to use `>> 1`, since we are effectively setting each bit of the index with each loop iteration, and we use bitwise ops elsewhere in the loop.

Just my preference, feel free to leave as is.

danhoeflinger (approved):
LGTM, optional change if you want to apply I'll reapprove.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: I read that `operator>>' isn't so clear as we thinking for the signed types with the value `< 0`. The interesting moment is the value of sign bit and highest bit(s) which state should be assigned during the `operator>>' implementation. See https://stackoverflow.com/questions/141525/what-are-bitwise-shift-bit-shift-operators-and-how-do-they-work for details. So would prefer `/2` operation when the type is unknown.

danhoeflinger: I agree with you for unknown types, but this function has a static_assert enforcing that `_Size` is an unsigned type.

SergeyKopienko: So it's OK in this case.

mmichel11: Performance is identical between these two versions, and unsigned types are required for the bit floor call.

I wasn't sure which one is clearer when writing this, but I think `>>=` makes the most sense since we use other bitwise operations in the loop.

I have went ahead and made this change.

danhoeflinger (approved):
LGTM

julianmi (approved):
LGTM

PR #1460, opened 2024-03-21
Fix `__parallel_partial_sort` call without wait/get

SergeyKopienko:
In this PR we add some comments why we shouldn't call .wat() for `__parallel_partial_sort` result in concrete case.

Another variant (in the last commit) - to add `IsSync` type into `__buffer` implementation based on `sycl::buffer`

See also https://github.com/oneapi-src/oneDPL/pull/1261

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1514,11 +1514,15 @@ __pattern_partial_sort_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& 
 
         auto __buf_mid = __buf_first + __out_size;
 
+        // The wait() call on result of __parallel_partial_sort isn't required here
+        // because our temporary buffer __buf (__par_backend_hetero::__buffer) is implemented on sycl::buffer
+        // and all required syncronization will be done on sycl::accessors level
+        // inside the next __pattern_walk2 call.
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1514,11 +1514,13 @@ __pattern_partial_sort_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&&
 
         auto __buf_mid = __buf_first + __out_size;
 
-        __par_backend_hetero::__parallel_partial_sort(
+        auto __future_obj = __par_backend_hetero::__parallel_partial_sort(
             _BackendTag{}, __par_backend_hetero::make_wrapped_policy<__partial_sort_2>(__exec),
             __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read_write>(__buf_first),
             __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read_write>(__buf_mid),
             __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read_write>(__buf_last), __comp);
+        if constexpr (!decltype(__buf).IsSync{})
```
SergeyKopienko, line 1522: One more way is to add `.wait()` call here without any conditions and de-facto existing synchronizations inside `__parallel_partial_sort` (`__parallel_partial_sort_impl`) due `sycl::accessor` syncrhonizations.

PR #1462, opened 2024-03-21
Removing noexcept to resolve issue reported by static analysis.

timmiesmith:
The pattern is used by uninitialized_fill, which uses placement new to created objects, and new may throw a length_error exception.  Issue was reported by static analysis tools.

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_fwd.h
+++ b/include/oneapi/dpl/pstl/algorithm_fwd.h
@@ -102,7 +102,7 @@ _RandomAccessIterator __brick_walk1_n(_RandomAccessIterator, _DifferenceType, _F
 
 template <class _Tag, class _ExecutionPolicy, class _ForwardIterator, class _Size, class _Function>
 _ForwardIterator
-__pattern_walk1_n(_Tag, _ExecutionPolicy&&, _ForwardIterator, _Size, _Function) noexcept;
+__pattern_walk1_n(_Tag, _ExecutionPolicy&&, _ForwardIterator, _Size, _Function);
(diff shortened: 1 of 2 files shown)


mmichel11 (dismissed):
The equivalent `noexcept` for the declaration in `algorithm_fwd.h` should also be removed. This should resolve the CI failures.

SergeyKopienko (changes requested):
What about `noexcept` from the files for some patterns:
- include/oneapi/dpl/pstl/numeric_impl.h
- include/oneapi/dpl/pstl/experimental/internal/for_loop_impl.h

Of course, forward declarations should be fixed too.

SergeyKopienko:
> The equivalent `noexcept` for the declaration in `algorithm_fwd.h` should also be removed. This should resolve the CI failures.

I have fixed this issue now.
But my comments from https://github.com/oneapi-src/oneDPL/pull/1462#pullrequestreview-1954262686 still valid.
@timmiesmith, do you need my help here?

timmiesmith:
> What about `noexcept` from the files for some patterns:
> 
> * include/oneapi/dpl/pstl/numeric_impl.h
> * include/oneapi/dpl/pstl/experimental/internal/for_loop_impl.h
> 
> Of course, forward declarations should be fixed too.

If the patterns aren't used with functors that perform memory allocation then the noexcept is valid.  I'm not looking for a wide removal of the keyword, just removing it where we identify that an exception may indeed be thrown in the use cases we're testing.

SergeyKopienko (approved):
LGTM

PR #1464, opened 2024-03-22
[Dynamic Selection] Adding sycl profiling for auto tune policy

AnuyaWelling2801:
Changes made : 
1. Adding an async_waiter_base to ensure type erasure for the selection type and enable lazy reporting
2. Adding a choice to use use_event_profiling = q.template has_property<sycl::property::queue::enable_profiling>() 
3. Using lazy report to report time instead of host task for auto_tune policy
4. Added more tests in auto_tune_policy_sycl for devices with event profiling enabled.
5. Changing the report method for auto_tune policy


include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
```diff
@@ -229,6 +229,9 @@ class auto_tune_policy
             }
             else
             {
+                if constexpr(lazy_report_v<Backend>){
+                        backend_->lazy_report();
```
vossmjp, line 233: You are holding a lock on state_->m_ when calling the lazy_report function. Seems unnecessary and perhaps dangerous as a general pattern.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -142,6 +195,15 @@ class sycl_backend
         return global_rank_;
     }
 
+    void lazy_report(){
```
vossmjp, line 198: This function is not thread safe. You might have concurrent accesses to the async_waiter array that change its size -- calls to lazy_report and submit (which calls add_waiter).

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -36,24 +36,59 @@ class sycl_backend
...
+        public:
+            virtual void wait() = 0;
+            virtual void report() = 0;
+            virtual bool is_complete() = 0;
+    };
+
+    std::vector<async_waiter_base*> async_waiter_arr;
+
+    template<typename T>
+    void add_waiter(T *t){
```
vossmjp, line 49: Are you protecting this?  Is it thread safe?

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -106,27 +141,45 @@ class sycl_backend
...
+                auto e2 = q.submit([=](sycl::handler& h){
+                    h.depends_on(e1);
+                    h.host_task([=](){
                         s.report(execution_info::task_completion);
-                    }
+                    });
                 });
-            });
-            return async_waiter{e2};
+                return async_waiter{e2, new SelectionHandle(s)};
```
vossmjp, line 160: I know we don't currently have a policy that requires reporting of both task completion and time, but this implementation would not support it because of this return. If we assumed they were mutually exclusive, the next condition would be an "else if", but its just an "if" so that both might be supported.

include/oneapi/dpl/internal/dynamic_selection_traits.h
```diff
@@ -299,6 +311,14 @@ struct report_value
 template <typename S, typename Info>
 inline constexpr bool report_value_v = report_value<S, Info>::value;
 
+template <typename S>
+struct lazy_report_value
```
vossmjp, line 315: I believe this is the first trait for a backend -- i.e. that a backend requires a call to lazy report. Backends are not as visible as policies so likely don't want a trait in this header. End users cannot access the backend, only the policy implementation can.

vossmjp (commented):
You need to take a closer look at thread safety and also find a non-trait-based approach to checking for lazy_report.

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_traits.h
AnuyaWelling2801: Not sure how to check if the backend has certain functions without traits. Created a new namespace called backend_traits for checking presence of lazy_report and has_enable_profiling in the backend from the policy

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +35,79 @@ class sycl_backend
     using execution_resource_t = resource_type;
     using resource_container_t = std::vector<execution_resource_t>;
 
+    bool has_enable_profiling = false;
   private:
-    class async_waiter
+    class async_waiter_base{
```
SergeyKopienko, line 40: What about virtual destructor for classes which will implement this interface?

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +35,79 @@ class sycl_backend
...
+        is_complete() override{
+            return e_.get_info<sycl::info::event::command_execution_status>() == sycl::info::event_command_status::complete;
+        }
+
     };
 
+    struct async_waiter_list_t{
+
+        std::mutex m_;
+        std::vector<async_waiter_base*> async_waiters;
```
SergeyKopienko, line 88: What about `std::unique_ptr` or `std::shared_ptr` instead of `async_waiter_base*` ?
Now it's looks like you will have memory leaks here...

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +35,79 @@ class sycl_backend
...
+            async_waiters.push_back(t);
+        }
+
+        void lazy_report(){
+            std::lock_guard<std::mutex> l(m_);
+            int size = async_waiters.size();
+            for(auto i = async_waiters.begin(); i!=async_waiters.begin()+size; i++){
+                if((*i)->is_complete()){
+                    (*i)->report();
+                    async_waiters.erase(i);
```
SergeyKopienko, line 102: ...because you simple delete item from vector without `delete` call.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -106,27 +163,45 @@ class sycl_backend
...
                 });
-            });
-            return async_waiter{e2};
+                return async_waiter{e2, new SelectionHandle(s)};
+            }
+            else if constexpr(report_value_v<SelectionHandle, execution_info::task_time_t>){
+                if (use_event_profiling)
+                {
+                    auto waiter = async_waiter{e1, new SelectionHandle(s)};
+                    async_waiter_list.add_waiter(new async_waiter(waiter));
```
SergeyKopienko, line 188: The same.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Added a virtual destructor.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Used a std::unique_ptr

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed. Used a std::unique_ptr

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -42,6 +42,7 @@ class sycl_backend
             virtual void wait() = 0;
             virtual void report() = 0;
             virtual bool is_complete() = 0;
+            virtual ~async_waiter_base(){}
```
SergeyKopienko, line 45: virtual ~async_waiter_base() = default;

SergeyKopienko:
@AnuyaWelling2801 I think that we still have memory leaks in this PR. Could you please check all `new` operations in this PR and how we will deallocate this memory?

SergeyKopienko (approved):
LGTM
But please wait additional approves and green CI before merge.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -106,27 +171,45 @@ class sycl_backend
...
+                    h.host_task([=](){
                         s.report(execution_info::task_completion);
-                    }
+                    });
                 });
-            });
-            return async_waiter{e2};
+                return async_waiter{e2, std::make_shared<SelectionHandle>(s)};
+            }
+            else if constexpr(report_value_v<SelectionHandle, execution_info::task_time_t>){
```
akukanov, line 192: Why `else if` here? Cannot both task time and task completion be reported?

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -89,6 +151,9 @@ class sycl_backend
         for (auto e : v)
         {
             global_rank_.push_back(e);
+            if(e.template has_property<sycl::property::queue::enable_profiling>()){
+                has_enable_profiling = true;
+            }
```
akukanov, line 156: The default constructor does not have similar runtime checks and therefore always disables profiling. Is that by design or an oversight?

akukanov (commented):
See some questions below.

Also, I do not quite understand the need for `lazy_report`. It's clear that you want to take kernel execution time from the event instead of measuring it on the host. But why cannot that be done in the host task following the kernel?

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +36,86 @@ class sycl_backend
...
+
+        void
+        report() override{
+            if constexpr (report_value_v<Selection, execution_info::task_time_t>){
+                cl_ulong time_start = e_.template get_profiling_info<sycl::info::event_profiling::command_start>();
+                cl_ulong time_end = e_.template get_profiling_info<sycl::info::event_profiling::command_end>();
+                s->report(execution_info::task_time, time_end-time_start);
+            }
+
+        }
```
akukanov, line 84: This implementation is very problematic.
A small problem that it uses `cl_ulong` for the timestamps, but `event::get_profiling_info<>` returns `uint64_t` according to the spec.

The big problem is that the timestamps are essentially implementation-defined in SYCL:
> The possible values for each info parameter and any restrictions are defined in the specification of the SYCL backend associated with the event.

Moreover, the implementation of the policy allows using queues with and without event profiling; in the latter case, the timestamps are taken on the host. All that means that the difference between two timestamps cannot be used to compare execution time for different queues without first being somehow converted to the same units. 

I think this is the showstopper for the patch.

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
```diff
@@ -217,6 +218,11 @@ class auto_tune_policy
     select(Function&& f, Args&&... args)
     {
         static_assert(sizeof...(KeyArgs) == sizeof...(Args));
+        if constexpr(backend_traits::lazy_report_v<Backend> && backend_traits::enable_profiling_v<Backend>){
+            if(backend_->has_enable_profiling == true){
+                backend_->lazy_report();
+            }
+        }
```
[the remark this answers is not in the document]
AnuyaWelling2801: Yes, I considered this. However, this would mean adding lazy_report to every new backend. And the usage of lazy_report for now is limited to only sycl::queue resource that have sycl_profiling enabled.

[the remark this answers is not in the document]
akukanov: You can check existence of `lazy_report` with a compile-time backend trait, that's still OK. I am only asking to hide the other trait, which seems more an implementation detail.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
akukanov: Looks like I am mistaken here. Reading the SYCL spec further, I see that these timestamps are required to be in nanoseconds and have the same time base per platform.
But let's make it well documented in the code, at least.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +36,86 @@ class sycl_backend
...
+
+    struct async_waiter_list_t{
+
+        std::mutex m_;
+        std::vector<std::shared_ptr<async_waiter_base>> async_waiters;
+
+        template<typename T>
+        void add_waiter(T *t){
+            std::lock_guard<std::mutex> l(m_);
+            async_waiters.push_back(std::unique_ptr<T>(t));
```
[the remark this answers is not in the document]
SergeyKopienko: According to https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr
```C++
template< class Y, class Deleter >
shared_ptr( [std::unique_ptr](http://en.cppreference.com/w/cpp/memory/unique_ptr)<Y, Deleter>&& r );
```
technically it's correct, but I agree there is no reasons to do this.
Probably better solution simple to change this declaration
```C++
std::vector<std::shared_ptr<async_waiter_base>> async_waiters;
```
to
```C++
std::vector<std::unique_ptr<async_waiter_base>> async_waiters;
```

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
SergeyKopienko: [`https://en.cppreference.com/w/cpp/header/chrono](https://en.cppreference.com/w/cpp/header/chrono) provide good solutions for that.
For example we may use one of `std::chrono::nanoseconds`, `std::chrono::microseconds` or `std::chrono::milliseconds` in all this internal API.

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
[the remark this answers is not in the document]
vossmjp: I agree that lazy_report is enough. No need for the other enable_profiling trait.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
vossmjp: There shouldn't be this restriction. I think it should an if and not an if-else.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
vossmjp: I think we should have three consistent behaviors: (1) has_enable_profiling should be true only if all queues have profiling enabled, currently it is true if one or more has it, (2) the default constructor should try to create queues with profiling enabled (but only if all default devices support profiling), and (3) given the previous 2 points, this single has_enable_profiling bool can be checked in submit instead of querying on the queue. We want this approach consistent across all queues in the universe because the timings will be incompatible and no meaningful comparison can be made if the approach differs between queues,

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/backend_traits.h
```diff
@@ -0,0 +1,66 @@
...
+// Copyright (C) 2023 Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _ONEDPL_INTERNAL_BACKEND_TRAITS_H
+#define _ONEDPL_INTERNAL_BACKEND_TRAITS_H
+
+#include "oneapi/dpl/internal/dynamic_selection_traits.h"
```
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/backend_traits.h
```diff
@@ -0,0 +1,66 @@
...
+}//namespace internal
+
+namespace backend_traits {
+    template <typename S>
+    struct lazy_report_value
+    {
+        static constexpr bool value = ::oneapi::dpl::experimental::internal::has_lazy_report<S>::value;
+    };
+    template <typename S>
+    inline constexpr bool lazy_report_v = lazy_report_value<S>::value;
```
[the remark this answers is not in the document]
AnuyaWelling2801: Staying consistent with the other traits of dynamic_selection and https://en.cppreference.com/w/cpp/types/is_floating_point

include/oneapi/dpl/internal/dynamic_selection_impl/backend_traits.h
```diff
@@ -0,0 +1,66 @@
...
+
+    template <typename S>
+    struct enable_profiling_value
+    {
+        static constexpr bool value = ::oneapi::dpl::experimental::internal::has_enable_profiling<S>::value;
+    };
+    template <typename S>
+    inline constexpr bool enable_profiling_v = enable_profiling_value<S>::value;
```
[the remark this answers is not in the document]
AnuyaWelling2801: Staying consistent with the other traits of dynamic_selection and https://en.cppreference.com/w/cpp/types/is_floating_point

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed. Changed to "if"

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +36,86 @@ class sycl_backend
...
-        async_waiter(const sycl::event& e) : e_(e) {}
+        async_waiter(sycl::event e) : e_(e) {}
+        async_waiter(sycl::event e, std::shared_ptr<Selection> selection) : e_(e), s(selection) {}
+
+        async_waiter(async_waiter &w) : e_(w.e_), s(w.s) {}
+        async_waiter& operator=(async_waiter &w){
+            s = w.s;
+            e_ = w.e_;
+            return *this;
+        }
```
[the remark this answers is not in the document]
AnuyaWelling2801: fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +36,86 @@ class sycl_backend
...
+
+    template<typename Selection>
+    class async_waiter : public async_waiter_base
     {
         sycl::event e_;
-
+        std::shared_ptr<Selection> s;
       public:
-        async_waiter(const sycl::event& e) : e_(e) {}
+        async_waiter(sycl::event e) : e_(e) {}
```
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed. Thank you @SergeyKopienko

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
```diff
@@ -47,6 +49,9 @@ class auto_tune_policy
     using size_type = typename std::vector<typename Backend::resource_type>::size_type;
     using timing_t = uint64_t;
 
+    using report_clock_type = std::chrono::steady_clock;
+    using report_duration = std::chrono::duration<double, std::milli>;
```
SergeyKopienko, line 53:
```suggestion
    using report_duration = std::chrono::milliseconds;
```

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +37,89 @@ class sycl_backend
     using execution_resource_t = resource_type;
     using resource_container_t = std::vector<execution_resource_t>;
 
+    using report_clock_type = std::chrono::steady_clock;
+    using report_duration = std::chrono::duration<double, std::milli>;
```
SergeyKopienko, line 41:
```suggestion
    using report_duration = std::chrono::milliseconds;
```

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
```diff
@@ -100,8 +105,8 @@ class auto_tune_policy
             }
             else
             {
-                auto now = std::chrono::steady_clock::now();
-                auto ms = std::chrono::duration_cast<std::chrono::milliseconds>(now - t0_).count();
+                auto now = report_clock_type::now();
+                auto ms = report_duration(now - t0_).count();
```
SergeyKopienko, line 109:
```suggestion
                const auto now = report_clock_type::now();
                const auto ms = std::chrono::duration_cast<report_duration>(now - t0_).count();
```

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +37,89 @@ class sycl_backend
...
+            if constexpr (report_value_v<Selection, execution_info::task_time_t>){
+                cl_ulong time_start = e_.template get_profiling_info<sycl::info::event_profiling::command_start>();
+                cl_ulong time_end = e_.template get_profiling_info<sycl::info::event_profiling::command_end>();
+                if(s!=nullptr){
+                    s->report(execution_info::task_time, report_duration(time_end-time_start).count());
+                }
+            }
+
+
+        }
```
SergeyKopienko, line 86:
```suggestion
        void
        report() override
        {
            if constexpr (report_value_v<Selection, execution_info::task_time_t>)
            {
                const auto time_start = e_.template get_profiling_info<sycl::info::event_profiling::command_start>();
                const auto time_end = e_.template get_profiling_info<sycl::info::event_profiling::command_end>();
                if (s != nullptr)
                {
                    // According to https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_event_information_and_profiling_descriptors
                    // each profiling descriptor returns a 64-bit timestamp that represents the number of nanoseconds that have elapsed since some implementation-defined timebase.
                    const auto duration_in_ns = std::chrono::nanoseconds(time_end - time_start);

                    s->report(execution_info::task_time, std::chrono::duration_cast<report_duration>(duration_in_ns).count());
                }
            }
        }
```

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -105,28 +176,45 @@ class sycl_backend
...
+                {
+                    auto waiter = async_waiter{e1,std::make_shared<SelectionHandle>(s)};
+                    async_waiter_list.add_waiter(new async_waiter(waiter));
+                    return waiter;
+                }
+                else{
+                    auto e2 = q.submit([=](sycl::handler& h){
+                        h.depends_on(e1);
+                        h.host_task([=](){
+                            s.report(execution_info::task_time, (report_clock_type::now() - t0).count());
```
SergeyKopienko, line 208:
```suggestion
                            const auto tp_now = report_clock_type::now();
                            s.report(execution_info::task_time, std::chrono::duration_cast<report_duration>(tp_now - t0).count());
```

SergeyKopienko (changes requested):

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
SergeyKopienko, line 53:
```suggestion

    using report_clock_type = std::chrono::steady_clock;
    using report_duration = std::chrono::milliseconds;
    using timing_t = typename report_duration::rep;
```

SergeyKopienko, line 53:
```suggestion
    using report_duration = std::chrono::milliseconds;
```

AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: fixed

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +37,90 @@ class sycl_backend
...
+
+        void
+        report() override{
+            if constexpr (report_value_v<Selection, execution_info::task_time_t>){
+                cl_ulong time_start = e_.template get_profiling_info<sycl::info::event_profiling::command_start>();
+                cl_ulong time_end = e_.template get_profiling_info<sycl::info::event_profiling::command_end>();
+                if(s!=nullptr){
+                    const auto duration_in_ns = std::chrono::nanoseconds(time_end-time_start);
+                    s->report(execution_info::task_time, std::chrono::duration_cast<report_duration>(duration_in_ns).count());
+                }
```
akukanov, line 83: It does not make much sense to obtain timestamps when reporting is skipped.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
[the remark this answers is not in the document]
akukanov: You removed the assignment, but the "copy" ctor is still explicit and takes an lvalue reference.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +37,90 @@ class sycl_backend
...
 
+    using report_clock_type = std::chrono::steady_clock;
+    using report_duration = std::chrono::milliseconds;
+
+    static inline bool is_profiling_enabled = false;
   private:
-    class async_waiter
+    class async_waiter_base{
+        public:
+            virtual void wait() = 0;
```
akukanov, line 47: Why is `wait()` a virtual method? I do not see it used via the base class pointer; do I miss something?

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +37,90 @@ class sycl_backend
...
+        }
+
+    };
+
+    struct async_waiter_list_t{
+
+        std::mutex m_;
+        std::vector<std::unique_ptr<async_waiter_base>> async_waiters;
+
+        template<typename T>
```
akukanov, line 101: I do not see why `add_waiter` should be a function template, instead of a member function taking `async_waiter_base*`. You would need to know the derived type for creating an object of that type, but this function creates no objects. Do I miss something?

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -105,28 +177,46 @@ class sycl_backend
...
+                auto e2 = q.submit([=](sycl::handler& h){
+                    h.depends_on(e1);
+                    h.host_task([=](){
                         s.report(execution_info::task_completion);
-                    }
+                    });
                 });
-            });
-            return async_waiter{e2};
+                return async_waiter{e2, std::make_shared<SelectionHandle>(s)};
```
akukanov, line 196: After changing `else if` to `if`, you have a problem with this return - it prevents the rest of the function code to execute. Therefore, supporting both task time and task completion reports still require more work.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
akukanov: This is insufficient; more details in another comment.

akukanov (changes requested):

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -142,17 +230,34 @@ class sycl_backend
...
+        bool profiling = true;
+        auto prop_list = sycl::property_list{};
         auto devices = sycl::device::get_devices();
         for (auto x : devices)
         {
             global_rank_.push_back(sycl::queue{x});
+            if(!x.has(sycl::aspect::queue_profiling)){
+                profiling = false;
+            }
+        }
```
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -142,17 +230,34 @@ class sycl_backend
...
+        is_profiling_enabled = profiling;
+        if(is_profiling_enabled){
+            prop_list = sycl::property_list{sycl::property::queue::enable_profiling()};
+        }
+        for (auto x : devices)
```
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Yes, it is an object of type async_waiter_base*. Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed. Although the async_waiter is over-written since we don't know what to do when a policy both satisfies execution_info::task_time_t and execution_info::task_completion_t

SergeyKopienko:
@AnuyaWelling2801 please take a look at https://github.com/oneapi-src/oneDPL/pull/1537, there two moments in the test which better to fix.

include/oneapi/dpl/internal/dynamic_selection_impl/backend_traits.h
```diff
@@ -0,0 +1,55 @@
...
+}//namespace internal
+
+namespace backend_traits {
+    template <typename S>
+    struct lazy_report_value
+    {
+        static constexpr bool value = ::oneapi::dpl::experimental::internal::has_lazy_report<S>::value;
+    };
+    template <typename S>
+    inline constexpr bool lazy_report_v = lazy_report_value<S>::value;
```
akukanov, line 47: @vossmjp should `backend_traits`, `lazy_report`, `lazy_report_v` etc. names be considered a (future) public API of DS backends, or there is no such intent?
I am asking because `backend_traits` is put into `oneapi::dpl[::experimental]` and looks more like a potential public name. Still experimental and so OK to change; nevertheless, should we have a naming discussion for those?

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -105,28 +174,47 @@ class sycl_backend
...
+                    });
+                    waiter = async_waiter{e2, std::make_shared<SelectionHandle>(s)};
+                }
+            }
+            return waiter;
         }
         else
         {
-            return async_waiter{f(unwrap(s), std::forward<Args>(args)...)};
+            return async_waiter{f(unwrap(s), std::forward<Args>(args)...), std::make_shared<SelectionHandle>(s)};
```
akukanov, line 217:
```suggestion
            return async_waiter{f(q, std::forward<Args>(args)...), std::make_shared<SelectionHandle>(s)};
```

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +37,87 @@ class sycl_backend
...
         {
             e_.wait();
         }
+
+        void
+        report() override{
+            if constexpr (report_value_v<Selection, execution_info::task_time_t>){
+                if(s!=nullptr){
+                    cl_ulong time_start = e_.template get_profiling_info<sycl::info::event_profiling::command_start>();
+                    cl_ulong time_end = e_.template get_profiling_info<sycl::info::event_profiling::command_end>();
```
SergeyKopienko, line 78:
```suggestion
                    const auto time_start = e_.template get_profiling_info<sycl::info::event_profiling::command_start>();
                    const auto time_end = e_.template get_profiling_info<sycl::info::event_profiling::command_end>();
```

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
akukanov: I think the following code should address that. It only creates a host task once.
```
            auto e1 = f(q, std::forward<Args>(args)...);
            async_waiter<SelectionHandle> waiter{e1,std::make_shared<SelectionHandle>(s)};

            if constexpr(report_value_v<SelectionHandle, execution_info::task_time_t>) {
                if (is_profiling_enabled)
                {
                    async_waiter_list.add_waiter(new async_waiter(waiter));
                }
            }

            bool is_host_task_needed = report_value_v<SelectionHandle, execution_info::task_time_t> && !is_profiling_enabled
                                       || report_info_v<SelectionHandle, execution_info::task_completion_t>;
            if (is_host_task_needed) {
                auto e2 = q.submit([=](sycl::handler& h){
                    h.depends_on(e1);
                    h.host_task([=](){
                        if constexpr(report_value_v<SelectionHandle, execution_info::task_time_t>){
                            if (!is_profiling_enabled) {
                                const auto tp_now = report_clock_type::now();
                                s.report(execution_info::task_time, std::chrono::duration_cast<report_duration>(tp_now - t0).count());
                            }
                        }
                        if constexpr(report_info_v<SelectionHandle, execution_info::task_completion_t>){
                            s.report(execution_info::task_completion);
                        }
                    });
                });
                waiter = async_waiter{e2, std::make_shared<SelectionHandle>(s)};
            }
            return waiter;
```

akukanov: @vossmjp I have a question though, whether in the case of a host task the returned event should point to that, or still to the submitted kernel. Given the "lazy" reporting, we cannot guarantee that waiting for the returned event ensures all associated "telemetry" is reported; but we can do that for task completion. Is it a sufficient motivation to return the host task event instead of the original kernel event?

include/oneapi/dpl/internal/dynamic_selection_impl/backend_traits.h
AnuyaWelling2801: `backend_traits `has traits that are meant at the backend level, not the policy level like rest of the traits. The reason for creating another namespace inside `oneapi::dpl::experimental` was to differentiate between those two and since the backend_traits.h file is inside the include/oneapi/dpl/internal/dynamic_selection_impl folder it is assumed that it is an implementation detail. 
The backend is never exposed to the user and thus `backend_traits`,` lazy_report`,` lazy_report_v` etc. names may not be  a public API of DS backends. 
Let us know your opinions, @vossmjp

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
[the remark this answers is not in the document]
AnuyaWelling2801: fixed. Changed to const auto

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed

SergeyKopienko:
@AnuyaWelling2801 please review and merge the PR https://github.com/oneapi-src/oneDPL/pull/1539

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
vossmjp: I think there is a tradeoff in either direction. The host task might take a while to run. If we wait for the telemetry-reporting host task to complete, this may delay the main application thread that waits on it, even though the work of the kernel is done. However, if another DS selection is made by the main thread and we do not wait for the telemetry-reporting host task, that decision is based on stale information. My preference would be to have less accurate selection instead of delaying the main thread from using a kernel result. @AnuyaWelling2801 are there measurements or tests that justify one approach over another?

include/oneapi/dpl/internal/dynamic_selection_impl/backend_traits.h
vossmjp: @AnuyaWelling2801 and @akukanov, it is expected that we will eventually document how to create custom backends. This will be a feature for policy writers and not end users of dynamic selection. So, it will become a documented set of APIs but for a much more limited audience. Previously, there was only an undocumented contract between the current concrete policies and the single current concrete backend. What we might do is include the internal/dynamic_selection_traits.h (which includes policy_traits.h) in the dynamic_selection header but not the backend_traits.h.

include/oneapi/dpl/internal/dynamic_selection_impl/dynamic_load_policy.h
```diff
@@ -150,6 +150,10 @@ struct dynamic_load_policy
     selection_type
     select(Args&&...)
     {
+        if constexpr (backend_traits::lazy_report_v<Backend>)
```
vossmjp, line 153: @AnuyaWelling2801 Shouldn't this file need to include the backend_traits.h?  Is it somehow including indirectly?

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: I tried waiting on the submitted kernel instead of the host task on our existing tests and they don't show much difference in timing. We should try this on existing examples to see if waiting on the existing kernel still makes it correct.

include/oneapi/dpl/internal/dynamic_selection_impl/dynamic_load_policy.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Also, fixed the ordering

include/oneapi/dpl/internal/dynamic_selection_impl/dynamic_load_policy.h
AnuyaWelling2801: The backend_traits.h file is being included indirectly since the tests include the file oneapi/dpl/dynamic_selection. This file includes auto_tune_policy,h (which contains the file backend_traits.h) before the file dynamic_load_policy.h

include/oneapi/dpl/internal/dynamic_selection_impl/backend_traits.h
```diff
@@ -0,0 +1,56 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _ONEDPL_INTERNAL_BACKEND_TRAITS_H
+#define _ONEDPL_INTERNAL_BACKEND_TRAITS_H
+
+#include <utility>
+#include <cstdint>
```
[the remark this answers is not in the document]
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +37,103 @@ class sycl_backend
...
+        {
+            std::lock_guard<std::mutex> l(m_);
+            async_waiters.push_back(std::unique_ptr<async_waiter_base>(t));
+        }
+
+        void
+        lazy_report()
+        {
+            if (is_profiling_enabled)
+            {
```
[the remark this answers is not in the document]
AnuyaWelling2801: Making the variable `is_profiling_enabled` private.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -35,25 +37,103 @@ class sycl_backend
     using execution_resource_t = resource_type;
     using resource_container_t = std::vector<execution_resource_t>;
 
+    using report_clock_type = std::chrono::steady_clock;
+    using report_duration = std::chrono::milliseconds;
```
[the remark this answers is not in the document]
AnuyaWelling2801: No, moved to private. Fixed.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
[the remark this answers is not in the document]
akukanov: Thanks. To clarify, by "move the is_profiling_enabled check to the caller" I meant deleting it from the method here and adding around the call to the method.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -180,16 +180,19 @@ class sycl_backend
     auto
     submit(SelectionHandle s, Function&& f, Args&&... args)
     {
+        constexpr bool report_task_completion = report_info_v<SelectionHandle, execution_info::task_completion_t>;
+        constexpr bool report_task_submission = report_info_v<SelectionHandle, execution_info::task_submission_t>;
+        constexpr bool  report_task_time = report_value_v<SelectionHandle, execution_info::task_time_t, report_duration>;
```
SergeyKopienko: Please remove extra space

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -199,32 +202,30 @@ class sycl_backend
...
                 if (is_profiling_enabled)
                 {
                     async_waiter_list.add_waiter(new async_waiter(waiter));
                 }
             }
 
-            bool is_host_task_needed = report_value_v<SelectionHandle, execution_info::task_time_t, report_duration> &&
-                                           !is_profiling_enabled ||
-                                       report_info_v<SelectionHandle, execution_info::task_completion_t>;
+            bool is_host_task_needed = report_task_time && !is_profiling_enabled || report_task_completion;
```
SergeyKopienko, line 212: const?

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -85,48 +161,82 @@ class sycl_backend
...
+                profiling = false;
+            }
         }
+        is_profiling_enabled = profiling;
         sgroup_ptr_ = std::make_unique<submission_group>(global_rank_);
     }
 
     template <typename SelectionHandle, typename Function, typename... Args>
     auto
     submit(SelectionHandle s, Function&& f, Args&&... args)
```
SergeyKopienko, line 180: @AnuyaWelling2801 I think we are able to make this function a little bit shorter: please take a loot at PR https://github.com/oneapi-src/oneDPL/pull/1547

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -36,24 +38,98 @@ class sycl_backend
...
+
+    class async_waiter_base
+    {
+      public:
+        virtual void
+        report() = 0;
+        virtual bool
+        is_complete() = 0;
+        virtual ~async_waiter_base() = default;
+    };
```
SergeyKopienko, line 53: I propose to skip clang format for this part and rewrite it as
```suggestion
    class async_waiter_base
    {
      public:
        virtual void report() = 0;
        virtual bool is_complete() = 0;
        virtual ~async_waiter_base() = default;
    };
```
because it's interface declaration and should be readable.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -36,24 +38,98 @@ class sycl_backend
...
+    using report_clock_type = std::chrono::steady_clock;
+    using report_duration = std::chrono::milliseconds;
+
+    class async_waiter_base
+    {
+      public:
+        virtual void
+        report() = 0;
+        virtual bool
+        is_complete() = 0;
```
SergeyKopienko, line 51: I think we are able and should declare this function as `const` :
```suggestion
        is_complete() const = 0;
```
Here and in all required places.

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
```diff
@@ -47,6 +49,9 @@ class auto_tune_policy
     using size_type = typename std::vector<typename Backend::resource_type>::size_type;
     using timing_t = uint64_t;
 
+    using report_clock_type = std::chrono::steady_clock;
+    using report_duration = std::chrono::milliseconds;
+
```
SergeyKopienko, line 54: Looks like safe to initialize
```C++
    struct resource_with_index_t
    {
        wrapped_resource_t r_;
        size_type index_ = 0;
    };
```

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Merged

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Overwritten by another fix

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: Fixed for `is_complete` and` report`

include/oneapi/dpl/internal/dynamic_selection_impl/auto_tune_policy.h
AnuyaWelling2801: Fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
AnuyaWelling2801: fixed

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -85,49 +159,67 @@ class sycl_backend
...
             report(s, execution_info::task_submission);
-        }
-        if constexpr (report_info_v<SelectionHandle, execution_info::task_completion_t> ||
-                      report_value_v<SelectionHandle, execution_info::task_time_t>)
+
+        if constexpr (report_task_completion || report_task_time)
         {
-            std::chrono::steady_clock::time_point t0;
-            if constexpr (report_value_v<SelectionHandle, execution_info::task_time_t>)
+            const auto t0 = report_clock_type::now();
```
akukanov, line 191: I am not sure I agree with the change Sergey proposed to remove `if constexpr` around taking the initial timing snapshot. It makes the code concise, but is that better in this case? Performance-wise it might be OK if we assume the overhead of `std::chrono::steady_clock::now()` is negligible, but it provokes questions like "why are the conditions to set and to use the variable different, including even constexpr checks; is it by purpose?".

Maybe using the ternary operator `?:` is a good compromise here; it is not constexpr but hopefully the compiler can optimize it anyway. But if I am the only one concerned and others are not, then fine to keep the code as now.

include/oneapi/dpl/internal/dynamic_selection_impl/dynamic_load_policy.h
vossmjp: Ok.

include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
vossmjp: I think it's ok as is. Let's keep as is. We can always revisit it if we see some performance impact later.

vossmjp (approved):
All issues are resolved, So looks good to me now.

PR #1457, opened 2024-03-20
[onedpl][ranges][productization] Range API productization, MVP plan, Part 1

MikeDvorskiy:
[onedpl][ranges][productization] Range API productization, MVP plan

oneDPL Range API productization:

Support the minimal algorithm set, according to  MVP plan. Part 1 (this patch): for_each, transform (unary), find, find_if, find_if_not, any_of, all_of, none_of, adjacent_find, search, search_n.
A common test (functor, projection, vector, usm_vector, std::subrange, std::span, sycl::buffer)

(A reference to the previous PR to refresh the conversation if it needs: https://github.com/oneapi-src/oneDPL/pull/1405)

A reference to the proposal for C++ parallel ranges algorithms, which this PR relies on: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3179r1.html

PR #1465, opened 2024-03-22
[CI] Add c++20 configurations with tests for ranges

dmitriy-sobolev:
The change aims to cover the basic functionality from #1457.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,6 +172,24 @@ jobs:
             build_type: release
             backend: serial
             device_type: HOST
+          - os: ubuntu-latest
+            cxx_compiler: icpx
+            std: 20
+            build_type: release
+            backend: dpcpp
+            device_type: CPU
+          - os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


dmitriy-sobolev:
Testing: https://github.com/oneapi-src/oneDPL/actions/runs/8395675455

.github/workflows/ci.yml
```diff
@@ -207,8 +225,8 @@ jobs:
           echo "::warning::Compiler: $(${{ matrix.cxx_compiler }} --version)"
 
           if [[ "${{ matrix.backend }}" == "dpcpp" ]]; then
-            make_targets="build-onedpl-sycl_iterator-tests"
-            ctest_flags="-R sycl_iterator_.*\.pass"
+            make_targets="build-onedpl-sycl_iterator-tests std_ranges.pass"
```
MikeDvorskiy, line 228: I think it would be an additional job which tests only std_ranges.pass and enables C++20.

PR #1470, opened 2024-03-27
Enable vectorized global loads for the reduction algorithms

julianmi:
Vectorization is performance critical on SIMD architectures. This patch enables vectorization by unrolling vector size wide loop iterations on both coalesced (commutative algorithms) and consecutive (non-commutative algorithms) loads. Coalesced loads will then load vectors of consecutive elements. This change improves the coalesced loads on Intel SIMD GPUs without decreasing the throughput on SIMT GPUs. Coalesced loads are therefore enabled on SPIR-V backends as well. `min_element` and `max_element` continue using consecutive loads on SPIR-V backends due to the performance penalty of the required index check when using coalesced global loads.

Secondly, the vectorization enables dynamic number of elements to be processed per work-item. Launch parameter tuning with compile time constants is therefore not needed anymore. This reduces the number of template instantiations from 13 to 3, which improves the compile times significantly (e.g., half the time for `sycl_iterator_reduce.pass`).

Thirdly, branch divergence is minimized by adding a flag showing whether the work-group can process full sequences of the input array. If so, branching withing the inner kernel can be removed. If not, all work-items in a group follow the same boundary-checked implementation.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -61,20 +61,21 @@ __adjust_iters_per_work_item(_Size& __iters_per_work_item)
 template <typename _Tp, typename _NDItemId, typename _Size, typename _TransformPattern, typename _ReducePattern,
           typename _InitType, typename _AccLocal, typename _Res, typename... _Acc>
 void
-__work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, ::std::uint8_t __iters_per_work_item,
+__work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Size __iters_per_work_item,
                            const bool __is_full, _TransformPattern __transform_pattern, _ReducePattern __reduce_pattern,
                            _InitType __init, const _AccLocal& __local_mem, const _Res& __res_acc, const _Acc&... __acc)
 {
     auto __local_idx = __item_id.get_local_id(0);
-    auto __group_size = __item_id.get_local_range().size();
+    const _Size __group_size = __item_id.get_local_range().size();
     union __storage
     {
         _Tp __v;
         __storage() {}
     } __result;
     // 1. Initialization (transform part). Fill local memory
-    __transform_pattern(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0, __local_mem, __acc...);
+    __transform_pattern(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0, __is_full,
                         /*__n_groups*/ (_Size)1, __local_mem, __result, __acc...);
+
     const _Size __n_items = __transform_pattern.output_size(__n, __group_size, __iters_per_work_item);
     // 2. Reduce within work group using local memory
     __result.__v = __reduce_pattern(__item_id, __n_items, __result.__v, __local_mem);
@@ -90,22 +91,23 @@ __work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, ::std::ui
 template <typename _Tp, typename _NDItemId, typename _Size, typename _TransformPattern, typename _ReducePattern,
           typename _AccLocal, typename _Tmp, typename... _Acc>
 void
-__device_reduce_kernel(const _NDItemId __item_id, const _Size __n, ::std::uint8_t __iters_per_work_item,
+__device_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Size __iters_per_work_item,
                        const bool __is_full, const _Size __n_groups, _TransformPattern __transform_pattern,
                        _ReducePattern __reduce_pattern, const _AccLocal& __local_mem, const _Tmp& __temp_acc,
                        const _Acc&... __acc)
 {
     auto __local_idx = __item_id.get_local_id(0);
     auto __group_idx = __item_id.get_group(0);
-    auto __group_size = __item_id.get_local_range().size();
+    const _Size __group_size = __item_id.get_local_range().size();
     union __storage
     {
         _Tp __v;
         __storage() {}
     } __result;
     // 1. Initialization (transform part). Fill local memory
-    __transform_pattern(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0, __local_mem, __acc...);
+    __transform_pattern(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0, __is_full, __n_groups,
                         __local_mem, __result, __acc...);
+
     const _Size __n_items = __transform_pattern.output_size(__n, __group_size, __iters_per_work_item);
     // 2. Reduce within work group using local memory
     __result.__v = __reduce_pattern(__item_id, __n_items, __result.__v, __local_mem);
@@ -132,8 +134,8 @@ struct __parallel_transform_reduce_small_submitter<_Tp, _Commutative, _VecSize,
               typename... _Ranges>
     auto
     operator()(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, const _Size __n,
-               ::std::uint16_t __work_group_size, ::std::uint8_t __iters_per_work_item, _ReduceOp __reduce_op,
-			   _TransformOp __transform_op, _InitType __init, _Ranges&&... __rngs) const
+               const _Size __work_group_size,const _Size __iters_per_work_item, _ReduceOp __reduce_op, _TransformOp __transform_op,
+               _InitType __init, _Ranges&&... __rngs) const
     {
         auto __transform_pattern =
             unseq_backend::transform_reduce<_ExecutionPolicy, _ReduceOp, _TransformOp, _Tp, _Commutative, _VecSize>{
@@ -168,9 +170,9 @@ template <typename _Tp, typename _Commutative, int _VecSize, typename _Execution
           typename _ReduceOp, typename _TransformOp, typename _InitType,
           oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0, typename... _Ranges>
 auto
-__parallel_transform_reduce_small_impl(oneapi::dpl::__internal::__device_backend_tag __backend_tag,
+__parallel_transform_reduce_small_impl((oneapi::dpl::__internal::__device_backend_tag __backend_tag,
                                        _ExecutionPolicy&& __exec, const _Size __n, ::std::uint16_t __work_group_size,
-                                       ::std::uint8_t __iters_per_work_item, _ReduceOp __reduce_op,
+                                       const _Size __iters_per_work_item, _ReduceOp __reduce_op,
                                        _TransformOp __transform_op, _InitType __init, _Ranges&&... __rngs)
 {
     using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
@@ -196,7 +198,7 @@ struct __parallel_transform_reduce_device_kernel_submitter<_Tp, _Commutative, _V
               typename _ExecutionPolicy2, typename... _Ranges>
     auto
     operator()(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Size __n,
-	           ::std::uint16_t __work_group_size, ::std::uint8_t __iters_per_work_item, _ReduceOp __reduce_op,
+	           const _Size __work_group_size, const _Size __iters_per_work_item, _ReduceOp __reduce_op,
 			   _TransformOp __transform_op, __result_and_scratch_storage<_ExecutionPolicy2, _Tp> __scratch_container,
 			   _Ranges&&... __rngs) const
     {
@@ -206,7 +208,7 @@ struct __parallel_transform_reduce_device_kernel_submitter<_Tp, _Commutative, _V
         auto __reduce_pattern = unseq_backend::reduce_over_group<_ExecutionPolicy, _ReduceOp, _Tp>{__reduce_op};
 
         // number of buffer elements processed within workgroup
-        _Size __size_per_work_group = __iters_per_work_item * __work_group_size;
+        const _Size __size_per_work_group = __iters_per_work_item * __work_group_size;
         const _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
         const bool __is_full = __n == __size_per_work_group * __n_groups;
 
@@ -242,7 +244,7 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<_Tp, _Commutative
     template <typename _ExecutionPolicy, typename _Size, typename _ReduceOp, typename _TransformOp, typename _InitType>
     auto
     operator()(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, sycl::event& __reduce_event, _Size __n,
-               ::std::uint16_t __work_group_size, ::std::uint8_t __iters_per_work_item, _ReduceOp __reduce_op, _InitType __init,
+               const _Size __work_group_size, const _Size __iters_per_work_item, _ReduceOp __reduce_op, _InitType __init,
                __result_and_scratch_storage<_ExecutionPolicy2, _Tp> __scratch_container) const
     {
         using _NoOpFunctor = unseq_backend::walk_n<_ExecutionPolicy, oneapi::dpl::__internal::__no_op>;
@@ -295,9 +297,9 @@ template <typename _Tp, typename _Commutative, int _VecSize, typename _Execution
           oneapi::dpl::__internal::__enable_if_device_execution_policy<_ExecutionPolicy, int> = 0, typename... _Ranges>
 auto
 __parallel_transform_reduce_mid_impl(oneapi::dpl::__internal::__device_backend_tag __backend_tag,
-                                     _ExecutionPolicy&& __exec, _Size __n, ::std::uint16_t __work_group_size,
-                                     ::std::uint8_t __iters_per_work_item_device_kernel,
-                                     ::std::uint8_t __iters_per_work_item_work_group_kernel, _ReduceOp __reduce_op,
+                                     _ExecutionPolicy&& __exec, _Size __n, const _Size __work_group_size,
+                                     const _Size __iters_per_work_item_device_kernel,
+                                     const _Size __iters_per_work_item_work_group_kernel, _ReduceOp __reduce_op,
                                      _TransformOp __transform_op, _InitType __init, _Ranges&&... __rngs)
 {
     using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
@@ -307,7 +309,7 @@ __parallel_transform_reduce_mid_impl(oneapi::dpl::__internal::__device_backend_t
         __reduce_mid_work_group_kernel<_CustomName>>;
 
     // number of buffer elements processed within workgroup
-    _Size __size_per_work_group = __iters_per_work_item_device_kernel * __work_group_size;
+    const _Size __size_per_work_group = __iters_per_work_item_device_kernel * __work_group_size;
     const _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
     __result_and_scratch_storage<_ExecutionPolicy, _Tp> __scratch_container(__exec, __n_groups);
 
@@ -316,10 +318,10 @@ __parallel_transform_reduce_mid_impl(oneapi::dpl::__internal::__device_backend_t
             __backend_tag, __exec, __n, __reduce_op, __transform_op, __init, __scratch_container,
             ::std::forward<_Ranges>(__rngs)...);
 
-    __n = __n_groups; // Number of preliminary results from the device kernel.
+    // __n_groups preliminary results from the device kernel.
     return __parallel_transform_reduce_work_group_kernel_submitter<_Tp, _Commutative, _VecSize,
                                                                    _ReduceWorkGroupKernel>()(
-        __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __reduce_event, __n, __work_group_size,
+        __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __reduce_event, __n_groups, __work_group_size,
         __iters_per_work_item_work_group_kernel, __reduce_op, __init, __temp);
         __init, __scratch_container);
 }
@@ -331,9 +333,9 @@ struct __parallel_transform_reduce_impl
     template <typename _ExecutionPolicy, typename _Size, typename _ReduceOp, typename _TransformOp, typename _InitType,
               typename... _Ranges>
     static auto
-    submit(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Size __n, 
-	       ::std::uint16_t __work_group_size, ::std::uint8_t __iters_per_work_item, _ReduceOp __reduce_op,
-		   _TransformOp __transform_op, _InitType __init, _Ranges&&... __rngs)
+    submit(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Size __n, _Size __work_group_size,
+           const _Size __iters_per_work_item, _ReduceOp __reduce_op, _TransformOp __transform_op, _InitType __init,
+		   _Ranges&&... __rngs)
     {
         using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
         using _NoOpFunctor = unseq_backend::walk_n<_ExecutionPolicy, oneapi::dpl::__internal::__no_op>;
@@ -350,11 +352,11 @@ struct __parallel_transform_reduce_impl
 
 #if _ONEDPL_COMPILE_KERNEL
         auto __kernel = __internal::__kernel_compiler<_ReduceKernel>::__compile(__exec);
-        __work_group_size = ::std::min(
-            __work_group_size, (::std::uint16_t)oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel));
+        __work_group_size =
+            ::std::min(__work_group_size, (_Size)oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel));
 #endif
 
-        _Size __size_per_work_group =
+        const _Size __size_per_work_group =
             __iters_per_work_item * __work_group_size; // number of buffer elements processed within workgroup
         _Size __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_work_group);
 
@@ -484,41 +486,45 @@ __parallel_transform_reduce(oneapi::dpl::__internal::__device_backend_tag __back
     ::std::size_t __iters_per_work_item = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __work_group_size);
     __adjust_iters_per_work_item<_VecSize>(__iters_per_work_item);
     const ::std::size_t __max_elements_per_wg = __work_group_size * 32;
+    const bool __is_n_32_bits = __n <= ::std::numeric_limits<::std::uint32_t>::max();
 
     // Use single work group implementation.
-    if (__iters_per_work_item <= 32)
+    if (__iters_per_work_item <= 32 && __is_n_32_bits)
     {
         return __parallel_transform_reduce_small_impl<_Tp, _Commutative, _VecSize>(
-            __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __n, __work_group_size, __iters_per_work_item,
-            __reduce_op, __transform_op, __init, ::std::forward<_Ranges>(__rngs)...);
+            __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), (::std::uint32_t)__n, (::std::uint32_t)__work_group_size,
+            (::std::uint32_t)__iters_per_work_item, __reduce_op, __transform_op, __init,
+            ::std::forward<_Ranges>(__rngs)...);
     }
     // Use two-step tree reduction.
     // First step reduces __work_group_size * __iters_per_work_item_device_kernel elements.
     // Second step reduces __work_group_size * __iters_per_work_item_work_group_kernel elements.
-    else if (__iters_per_work_item <= __max_elements_per_wg * __max_elements_per_wg)
+    else if (__iters_per_work_item <= __max_elements_per_wg * __max_elements_per_wg && __is_n_32_bits)
     {
-        ::std::size_t __iters_per_work_item_device_kernel =
-            oneapi::dpl::__internal::__dpl_ceiling_div(__n, __max_elements_per_wg);
+        const ::std::uint32_t __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
+        // Add a factor more work-groups than compute units to fully utilizes the device and hide latencies.
+        // The value is empirically tested.
+        constexpr ::std::uint32_t __oversubscription = 2;
+        const ::std::uint32_t __max_wg = __max_cu * __oversubscription;
+
+        ::std::uint32_t __iters_per_work_item_device_kernel =
+            oneapi::dpl::__internal::__dpl_ceiling_div(__n, __max_wg * __work_group_size);
         __adjust_iters_per_work_item<_VecSize>(__iters_per_work_item_device_kernel);
-        ::std::size_t __iters_per_work_item_work_group_kernel = 1;
-        if (__iters_per_work_item_device_kernel > 32)
-        {
-            __iters_per_work_item_work_group_kernel =
-                oneapi::dpl::__internal::__dpl_ceiling_div(__iters_per_work_item_device_kernel, 32);
-            __adjust_iters_per_work_item<_VecSize>(__iters_per_work_item_work_group_kernel);
-            __iters_per_work_item_device_kernel = 32;
-        }
+        ::std::uint32_t __iters_per_work_item_work_group_kernel =
+            oneapi::dpl::__internal::__dpl_ceiling_div(__max_wg, __work_group_size);
+        __adjust_iters_per_work_item<_VecSize>(__iters_per_work_item_work_group_kernel);
         return __parallel_transform_reduce_mid_impl<_Tp, _Commutative, _VecSize>(
-            __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __n, __work_group_size, __iters_per_work_item_device_kernel,
-            __iters_per_work_item_work_group_kernel, __reduce_op, __transform_op, __init,
+            __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), (::std::uint32_t)__n, (::std::uint32_t)__work_group_size,
+            (::std::uint32_t)__iters_per_work_item_device_kernel,
+            (::std::uint32_t)__iters_per_work_item_work_group_kernel, __reduce_op, __transform_op, __init,
             ::std::forward<_Ranges>(__rngs)...);
     }
     // Otherwise use a recursive tree reduction with 32 __iters_per_work_item.
     else
     {
         return __parallel_transform_reduce_impl<_Tp, _Commutative, _VecSize>::submit(
-            __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __n, __work_group_size, 32, __reduce_op, __transform_op, __init,
-            ::std::forward<_Ranges>(__rngs)...);
+            __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __n, (decltype(__n))__work_group_size, (decltype(__n))32,
+            __reduce_op, __transform_op, __init, ::std::forward<_Ranges>(__rngs)...);
     }
 }
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -235,11 +235,12 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<_Tp, _Commutative
 
         // Lower the work group size of the second kernel to the next power of 2 if __n < __work_group_size.
         auto __work_group_size2 = __work_group_size;
-        if (__iters_per_work_item == 1)
+        if (__iters_per_work_item == 4)
         {
-            if (__n < __work_group_size)
+            const _Size __no_wis = oneapi::dpl::__internal::__dpl_ceiling_div(__n, 4);
+            if (__no_wis < __work_group_size)
             {
-                __work_group_size2 = __n;
+                __work_group_size2 = __no_wis;
                 if ((__work_group_size2 & (__work_group_size2 - 1)) != 0)
                     __work_group_size2 = oneapi::dpl::__internal::__dpl_bit_floor(__work_group_size2) << 1;
             }
@@ -250,8 +251,7 @@ struct __parallel_transform_reduce_work_group_kernel_submitter<_Tp, _Commutative
 
             auto __temp_acc = __scratch_container.__get_scratch_acc(__cgh);
             auto __res_acc = __scratch_container.__get_result_acc(__cgh);
-            ::std::size_t __local_mem_size = __reduce_pattern.local_mem_req(__work_group_size2);
-            __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__local_mem_size), __cgh);
+            __dpl_sycl::__local_accessor<_Tp> __temp_local(sycl::range<1>(__work_group_size), __cgh);
 
             __cgh.parallel_for<_KernelName...>(
                 sycl::nd_range<1>(sycl::range<1>(__work_group_size2), sycl::range<1>(__work_group_size2)),
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -50,11 +50,12 @@ class __reduce_kernel;
 // Adjust number of sequential operations per work-item based on the vector size. Single elements are kept to
 // improve performance of small arrays or remainder loops.
 template <::std::uint8_t _VecSize, typename _Size>
-inline void
-__adjust_iters_per_work_item(_Size& __iters_per_work_item)
+inline _Size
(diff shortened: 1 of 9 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -426,96 +424,76 @@ __parallel_transform_reduce(oneapi::dpl::__internal::__device_backend_tag __back
 {
     auto __n = oneapi::dpl::__ranges::__get_first_range_size(__rngs...);
     assert(__n > 0);
+    using _Size = decltype(__n);
+
+    // Empirically found tuning parameters for typical devices.
+    constexpr _Size __max_iters_per_work_item = 32;
+    constexpr ::std::size_t __max_work_group_size = 256;
+    constexpr ::std::uint8_t __vector_size = 4;
```
mmichel11, line 432: Does a `__vector_size` of 4 generally perform well for different types or should this be based on the size of the type?

Ex: If `__vector_size = 4` performs best when reducing 32-byte buffers, would `__vector_size = 2` perform best for 64-byte types and `__vector_size = 8` for 16-byte types?

julianmi: This is an interesting question. I will gather some data.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -426,96 +424,76 @@ __parallel_transform_reduce(oneapi::dpl::__internal::__device_backend_tag __back
 {
     auto __n = oneapi::dpl::__ranges::__get_first_range_size(__rngs...);
     assert(__n > 0);
+    using _Size = decltype(__n);
+
+    // Empirically found tuning parameters for typical devices.
+    constexpr _Size __max_iters_per_work_item = 32;
+    constexpr ::std::size_t __max_work_group_size = 256;
```
MikeDvorskiy, line 431: Probably, is better to make request to a device, like `oneapi::dpl::__internal::__max_work_group_size(...)` and `oneapi::dpl::__internal::__max_sub_group_size(...)` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: 4-wide vectors seem to perform best across 16, 32, and 64-bit types on data center GPUs.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,101 +202,141 @@ struct __init_processing
...
-    template <typename _NDItemId, typename _Size, typename _AccLocal, typename... _Acc>
-    inline void
-    operator()(const _NDItemId& __item_id, const _Size& __n, const _Size& __global_offset, const _AccLocal& __local_mem,
-               const _Acc&... __acc) const
-    {
-        if constexpr (__use_nonseq_impl)
-            return nonseq_impl(__item_id, __n, __global_offset, __local_mem, __acc...);
-        else
-            return seq_impl(__item_id, __n, __global_offset, __local_mem, __acc...);
+        return;
```
mmichel11, line 323: Can the `return` be removed? We are at the end of the function.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,101 +202,141 @@ struct __init_processing
...
+            _Res __res = __unary_op(__adjusted_global_id, __acc...);
+            for (_Size __i = 1; __i < __vec_size_minus_one; ++__i)
+            {
+                const _Size __base_idx = __adjusted_global_id + __i;
+                if (__base_idx >= __adjusted_n)
+                    break;
+                __res = __binary_op(__res, __unary_op(__base_idx, __acc...));
+            }
             __local_mem[__local_idx] = __res;
+            return;
```
mmichel11, line 321: Can this `return` be removed? Once we leave the if statement we will exit the function.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,101 +202,141 @@ struct __init_processing
...
+                        if (__base_idx + __idx >= __adjusted_n)
+                            break;
+                        __res = __binary_op(__res, __unary_op(__base_idx + __idx, __acc...));
+                    }
+                }
+                else
+                    break;
+            }
             __local_mem[__local_idx] = __res;
+            return;
```
mmichel11, line 307: Can we remove this return and make the subsequent `if` an `else if` or is there a performance penalty otherwise? There's a few other places we do this as well.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,101 +202,141 @@ struct __init_processing
...
+        // Scalar remainder
+        if (__adjusted_global_id < __adjusted_n)
         {
-            const _Size __items_to_process =
-                std::max(((__adjusted_n - __adjusted_global_id - 1) / __stride) + 1, static_cast<_Size>(0));
-            typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
-            for (_Size __i = 1; __i < __items_to_process; ++__i)
-                __res = __binary_op(__res, __unary_op(__adjusted_global_id + __stride * __i, __acc...));
+            _Res __res = __unary_op(__adjusted_global_id, __acc...);
+            for (_Size __i = 1; __i < __vec_size_minus_one; ++__i)
```
mmichel11, line 313: Does it make sense to put this loop into a helper function similar to `vectorized_reduction_remainder` but for a nonvectorized equivalent?

mmichel11 (commented):
The comments I have are mostly stylistic to better understand the flow of the code.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,101 +202,141 @@ struct __init_processing
...
+            for (_Size __i = 1; __i < __no_vec_ops; ++__i)
+            {
+                const _Size __base_idx = __adjusted_global_id + __i * __stride;
+                if (__base_idx + __vec_size_minus_one < __adjusted_n)
+                    vectorized_reduction_remainder(__base_idx, __res, __acc...);
+                else if (__base_idx < __adjusted_n)
+                {
+                    for (_Size __idx = 0; __idx < __vec_size_minus_one; ++__idx)
+                    {
+                        if (__base_idx + __idx >= __adjusted_n)
```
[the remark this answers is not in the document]
julianmi: Good idea. I've changed this.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,101 +202,141 @@ struct __init_processing
...
+            return;
+        }
+        const _Size __local_range = __item_id.get_local_range(0);
+        const _Size __no_vec_ops = __iters_per_work_item / _VecSize;
         const _Size __adjusted_n = __global_offset + __n;
-        // Sequential load and reduce from global memory
-        if (__adjusted_global_id + __iters_per_work_item < __adjusted_n)
+
+        _Size __stride = _VecSize; // sequential loads with _VecSize-wide vectors
+        if constexpr (_Commutative{})
```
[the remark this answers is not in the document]
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -45,20 +45,31 @@ class __reduce_mid_work_group_kernel;
...
 class __reduce_kernel;
 
+// Adjust number of sequential operations per work-item based on the vector size. Single elements are kept to
+// improve performance of small arrays or remainder loops.
+template <::std::uint8_t _VecSize, typename _Size>
+inline void
+__adjust_iters_per_work_item(_Size& __iters_per_work_item)
+{
+    if (__iters_per_work_item > 1)
+        __iters_per_work_item = ((__iters_per_work_item + _VecSize - 1) / _VecSize) * _VecSize;
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,101 +202,141 @@ struct __init_processing
...
-            const _Size __items_to_process =
-                std::max(((__adjusted_n - __adjusted_global_id - 1) / __stride) + 1, static_cast<_Size>(0));
-            typename _AccLocal::value_type __res = __unary_op(__adjusted_global_id, __acc...);
-            for (_Size __i = 1; __i < __items_to_process; ++__i)
-                __res = __binary_op(__res, __unary_op(__adjusted_global_id + __stride * __i, __acc...));
+            _Res __res = __unary_op(__adjusted_global_id, __acc...);
+            for (_Size __i = 1; __i < __vec_size_minus_one; ++__i)
+            {
+                const _Size __base_idx = __adjusted_global_id + __i;
+                if (__base_idx >= __adjusted_n)
```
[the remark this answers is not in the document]
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -47,24 +47,37 @@ class __reduce_mid_work_group_kernel;
 template <typename... _Name>
 class __reduce_kernel;
 
+// Adjust number of sequential operations per work-item based on the vector size. Single elements are kept to
+// improve performance of small arrays or remainder loops.
+template <::std::uint8_t _VecSize, typename _Size>
+inline _Size
```
SergeyKopienko, line 53: I think `inline` isn't required for template functions

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,99 +202,131 @@ struct __init_processing
...
+        else if (__adjusted_global_id + __vec_size_minus_one < __adjusted_n)
+        {
+            vectorized_reduction_first(__adjusted_global_id, __res, __acc...);
+            for (_Size __i = 1; __i < __no_vec_ops; ++__i)
+            {
+                const _Size __base_idx = __adjusted_global_id + __i * __stride;
+                if (__base_idx + __vec_size_minus_one < __adjusted_n)
+                    vectorized_reduction_remainder(__base_idx, __res, __acc...);
+                else if (__base_idx < __adjusted_n)
+                {
```
SergeyKopienko, line 300: Here you place one statement into `{}` brackets, but in other cases you are using one statement lines without  `{}` brackets.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,99 +202,131 @@ struct __init_processing
...
             _Size __full_group_contrib = (__n / __items_per_work_group) * __work_group_size;
             _Size __last_wg_remainder = __n % __items_per_work_group;
-
-            _Size __last_wg_contrib = ::std::min(__last_wg_remainder, static_cast<_Size>(__work_group_size));
+            // Adjust remainder and wg size for vector size
+            _Size __last_wg_vec = oneapi::dpl::__internal::__dpl_ceiling_div(__last_wg_remainder, _VecSize);
+            _Size __last_wg_contrib = ::std::min(__last_wg_vec, static_cast<_Size>(__work_group_size * _VecSize));
             return __full_group_contrib + __last_wg_contrib;
         }
         else
```
SergeyKopienko, line 332: I think this `else` isn't required.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: I've removed the `inline` statement.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: I've removed the brackets.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: I've removed the `else`.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -61,15 +61,15 @@ using __has_known_identity = ::std::conditional_t<
                                               ::std::is_same<::std::decay_t<_BinaryOp>, __dpl_sycl::__minimum<void>>,
                                               ::std::is_same<::std::decay_t<_BinaryOp>, __dpl_sycl::__maximum<_Tp>>,
                                               ::std::is_same<::std::decay_t<_BinaryOp>, __dpl_sycl::__maximum<void>>>>>,
-#    else  //_ONEDPL_LIBSYCL_VERSION >= 50200
+#    else               //_ONEDPL_LIBSYCL_VERSION >= 50200
```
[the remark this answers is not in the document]
julianmi: Thanks for catching that. I've removed these unintentional formatting changes.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,132 @@ struct __init_processing
...
-        if constexpr (__use_nonseq_impl)
-            return nonseq_impl(__item_id, __n, __global_offset, __res, __acc...);
-        else
-            return seq_impl(__item_id, __n, __global_offset, __res, __acc...);
     }
 
     template <typename _Size>
     _Size
-    output_size(const _Size& __n, const ::std::uint16_t& __work_group_size) const
+    output_size(const _Size& __n, const _Size& __work_group_size, const _Size& __iters_per_work_item) const
```
SergeyKopienko, line 316: Small comment: should we really pass `_Size` data types by reference? May be simple by value?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,132 @@ struct __init_processing
...
+        for (_Size __i = 0; __i < _VecSize; ++__i)
+            __res.__v = __binary_op(__res.__v, __unary_op(__start_idx + __i, __acc...));
+    }
+
+    template <typename _Size, typename _Res, typename... _Acc>
+    void
+    scalar_reduction_remainder(const _Size __start_idx, const _Size __adjusted_n, const _Size __max_iters, _Res& __res,
+                               const _Acc&... __acc) const
+    {
+        const _Size __no_iters = std::min(static_cast<_Size>(__adjusted_n - __start_idx), __max_iters);
```
MikeDvorskiy, line 236: `static_cast<_Size>` here and everywhere looks very suspicious and bulky...
Probably, we can pass a "right" integer type as _Size? Or/and use `auto` when it is applicable and doesn't break correctness?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,132 @@ struct __init_processing
...
+    void
+    operator()(const _NDItemId& __item_id, const _Size& __n, const _Size& __iters_per_work_item,
+               const _Size& __global_offset, const bool __is_full, const _Size __n_groups, _Res& __res,
+               const _Acc&... __acc) const
+    {
+        const _Size __local_idx = __item_id.get_local_id(0);
+        const _Size __global_idx = __item_id.get_global_id(0);
+        if (__iters_per_work_item == 1)
+        {
+            new (&__res.__v) _Tp(__unary_op(__global_idx, __acc...));
```
MikeDvorskiy, line 251: Due to `__local_idx` is not used within this scope, the definition of `__local_idx` (line 247) may be moved down after `if `operator.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: While this was not introduced in this PR, I agree that we should pass by value here and changed that.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,138 @@ struct __init_processing
...
+            _Size __local_idx = __item_id.get_local_id(0);
+            _Size __group_idx = __item_id.get_group_linear_id();
+            __adjusted_global_id += __group_idx * __local_range * __iters_per_work_item + __local_idx * _VecSize;
+        }
+        else
+            __adjusted_global_id += __iters_per_work_item * __global_idx;
+
+        // If n is not evenly divisible by the number of elements processed per work-group, the last work-group might
+        // need to process less elements than __iters_per_work_item.
+        bool __is_full_wg = __is_full;
```
mmichel11, line 273: Can we simplify this to something like:
```
bool __is_full_wg = __is_full ||
    (static_cast<_Size>( __item_id.get_group(0)) < static_cast<_Size>(__n_groups - 1));
```
I think the `__is_multi_group` check is not needed since if there is one group and it is is full, then `__is_full` will be true already.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,138 @@ struct __init_processing
...
+                vectorized_reduction_remainder(__adjusted_global_id + __i * __stride, __res, __acc...);
         }
+        // At least one vector operation
+        else if (__adjusted_global_id + __vec_size_minus_one < __adjusted_n)
+        {
+            vectorized_reduction_first(__adjusted_global_id, __res, __acc...);
+            for (_Size __i = 1; __i < __no_vec_ops; ++__i)
+            {
+                const _Size __base_idx = __adjusted_global_id + __i * __stride;
+                if (__base_idx + __vec_size_minus_one < __adjusted_n)
```
mmichel11, line 300: Can we calculate the number of loop iterations outside of the loop? I think that the `scalar_reduction_remainder` call can be pulled after the loop to remove most the branching if we reduce the iteration count by 1 since it should be called on the last iteration if I understand correctly.

mmichel11 (commented):
I have a few small comments and am ready to approve once these are considered.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -47,24 +47,37 @@ class __reduce_mid_work_group_kernel;
...
 class __reduce_kernel;
 
+// Adjust number of sequential operations per work-item based on the vector size. Single elements are kept to
+// improve performance of small arrays or remainder loops.
+template <std::uint8_t _VecSize, typename _Size>
+auto
+__adjust_iters_per_work_item(_Size __iters_per_work_item) -> _Size
+{
+    if (__iters_per_work_item > 1)
+        return ((__iters_per_work_item + _VecSize - 1) / _VecSize) * _VecSize;
```
adamfidel, line 57: I think this can be written with `__dpl_ceiling_div` like:

```c++
return __dpl_ceiling_div(__iters_per_work_item, _VecSize) * _VecSize;
```

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,138 @@ struct __init_processing
 
 // Load elements consecutively from global memory, transform them, and apply a local reduction. Each local result is
 // stored in local memory.
-template <typename _ExecutionPolicy, ::std::uint8_t __iters_per_work_item, typename _Operation1, typename _Operation2,
-          typename _Tp, typename _Commutative>
+template <typename _ExecutionPolicy, typename _Operation1, typename _Operation2, typename _Tp, typename _Commutative,
+          std::uint8_t _VecSize>
 struct transform_reduce
```
danhoeflinger, line 207: The changes here must be propagated to `sycl_traits.h` to update the forward declaration there (I believe this is causing the red CI).  Also, depending on which PR is merged first, #1521 the definition for `transform_reduce` will need to be updated accordingly to specify the template parameters which need to be checked to check device copyability.  
I can help make this adjustment if it is not clear to you, let me know.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,138 @@ struct __init_processing
...
         {
             _Size __items_per_work_group = __work_group_size * __iters_per_work_item;
             _Size __full_group_contrib = (__n / __items_per_work_group) * __work_group_size;
             _Size __last_wg_remainder = __n % __items_per_work_group;
-
-            _Size __last_wg_contrib = ::std::min(__last_wg_remainder, static_cast<_Size>(__work_group_size));
+            // Adjust remainder and wg size for vector size
+            _Size __last_wg_vec = oneapi::dpl::__internal::__dpl_ceiling_div(__last_wg_remainder, _VecSize);
+            _Size __wg_vec_size = __work_group_size * _VecSize;
+            _Size __last_wg_contrib = std::min(__last_wg_vec, __wg_vec_size);
```
danhoeflinger, line 333: `__last_wg_contrib` seems incorrect, we are dividing the remainder by the vector size, and then comparing that vs a full number of elements processed by a full group.  It seems like one half of the min operation is in "vector" space, and the other half is in "element" space.  

I haven't fully wrapped my head around whether we should be multiplying again by `_VecSize` and this is a rounding up to the nearest multiple of the vector size, or if we shouldn't be dividing in the first place. I think perhaps we shouldn't be dividing the remainder by the vector size?

I'm probably missing something.

danhoeflinger (commented):
apologies for this late review.  Mostly minor things.  
I'm still wrapping my head around things fully so I probably shouldn't be the approver.  However hopefully some of these comments can help.  Continuing to look as time permits as well.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: I've changed the check and tried to clarify the comment.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -456,96 +457,75 @@ __parallel_transform_reduce(oneapi::dpl::__internal::__device_backend_tag __back
...
-                __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __n, __reduce_op, __transform_op, __init,
-                ::std::forward<_Ranges>(__rngs)...);
-        }
-        else if (__n <= 67108864)
+    // Use single work group implementation if less than __max_iters_per_work_item elements per work-group.
+    // We can use 16-bit addressing since we have at most __max_work_group_size * __max_iters_per_work_item elements.
+    if (__n <= __max_elements_per_wg)
+    {
+        const auto __n_short = static_cast<std::uint16_t>(__n);
+        const auto __work_group_size_short = static_cast<std::uint16_t>(__work_group_size);
```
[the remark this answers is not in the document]
julianmi: Just for brevity. I can change this if you have a strong opinion.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: Good catch. I believe `__work_group_size` is the right minimum value to check here.

The general idea is to find how many work-items reduce at least a single element in the coalesced access. Each participating work-item does vector width operation. If each work-item in the last group participates, we have work-grup size partial results. If less than that, we need to divide and round up by vector size to get the busy work-items.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,138 @@ struct __init_processing
...
             return __full_group_contrib + __last_wg_contrib;
         }
```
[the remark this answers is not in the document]
julianmi: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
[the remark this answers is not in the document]
danhoeflinger: No strong opinion, just preference,  you can leave as is.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
danhoeflinger: I believe it looks correct to me.  It should merely be adjusting the fwd declaration for now to match what is currently in the definition.  Once this or #1521 gets merged, we will need to adjust another line slightly to match the types in the specialization itself.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
danhoeflinger: That makes sense, and I think it is the appropriate fix.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,133 @@ struct __init_processing
...
+        // At least one vector operation
+        else if (__adjusted_global_id + __vec_size_minus_one < __adjusted_n)
+        {
+            vectorized_reduction_first(__adjusted_global_id, __res, __acc...);
+            if (__no_vec_ops > 1)
+            {
+                _Size __n_diff = __adjusted_n - __adjusted_global_id - _VecSize;
+                _Size __no_iters = __n_diff / __stride;
+                _Size __no_vec_ops_minus_one = __no_vec_ops - 1;
+                __no_iters = std::min(__no_iters, __no_vec_ops_minus_one);
```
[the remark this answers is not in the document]
julianmi: I might be misunderstanding your comment, but `__n_diff` can get very large. E.g., work-item 0 might start at index 0 (`__adjusted_global_id = 0`), which might lead to a large `__n_diff` depending on the input size. In the scalar case, the stride is the vector size and `__no_iters` will be large as well. The stride is much bigger for the coalesced case but could still be many times smaller than `__n_diff`. So, I think limiting this is required for all work-groups except for the last one.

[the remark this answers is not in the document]
mmichel11: The `if` branch above is:
`(__is_full || (__group_idx < __n_groups_minus_one))`

and this code is in the subsequent else if. It seems to me that only the last work group reaches this point which is why I think the min call can be removed.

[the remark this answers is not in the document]
julianmi: I agree that only the last work-group can reach this but I think the issue is with the scalar version. E.g., the work-item 0 of the last work-group is supposed to process 8 elements with a vector size of 2 (2 vector operations). If the remainder to be process with the last work-group is larger than 16 (`__n_diff = 16 - 4 = 12, __no_iters = 12 / 4 = 3`), work-item 0 will process too many elements.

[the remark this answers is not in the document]
mmichel11: Thanks for explaining. I understand why this is necessary now.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
julianmi: @danhoeflinger #1521 was merged and I've rebased this PR. Please have a look if the updated forward declarations are fine. Thanks.

danhoeflinger: I believe you will need to also update `implementation_details/device_copyable.pass`.  That should also be a good test of if we are still in good shape.  Otherwise the change looks good.

julianmi: Thanks, I've made the test change.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -294,15 +284,20 @@ struct transform_reduce
...
+                _Size __base_idx = __adjusted_global_id + __stride;
+                for (_Size __i = 1; __i <= __no_iters; ++__i)
+                {
                     vectorized_reduction_remainder(__base_idx, __res, __acc...);
-                else if (__base_idx < __adjusted_n)
+                    __base_idx += __stride;
+                }
+                if (__no_iters < __no_vec_ops_minus_one && __base_idx < __adjusted_n)
```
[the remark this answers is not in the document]
julianmi: Thanks, I tend to agree. One issue I see is that `_Size` might be unsigned and the length of the remainder (`_Size __remainder = __adjusted_n - __start_idx;`) might then be falsely positive. So I believe we need to do the bounds check anyways. I've moved it to `scalar_reduction_remainder` such that we don't have to duplicate it. I've also added your proposed `__excess_scalar_elements` change below for clarity. This one is specific to the code path where we do at least one vector operation and I'm not sure how we would include this in the general `scalar_reduction_remainder`.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -294,15 +284,20 @@ struct transform_reduce
...
             vectorized_reduction_first(__adjusted_global_id, __res, __acc...);
-            for (_Size __i = 1; __i < __no_vec_ops; ++__i)
+            if (__no_vec_ops > 1)
             {
-                const _Size __base_idx = __adjusted_global_id + __i * __stride;
-                if (__base_idx + __vec_size_minus_one < __adjusted_n)
+                _Size __n_diff = __adjusted_n - __adjusted_global_id - _VecSize;
+                _Size __no_iters = __n_diff / __stride;
+                _Size __no_vec_ops_minus_one = __no_vec_ops - 1;
+                __no_iters = std::min(__no_iters, __no_vec_ops_minus_one);
```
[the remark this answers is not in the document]
julianmi: Done (see comment above).

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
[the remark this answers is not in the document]
danhoeflinger: Thanks, see my other comment, I think we can further simplify.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,144 @@ struct __init_processing
...
+    {
+        if (__start_idx < __adjusted_n)
         {
-            const _Size __items_to_process = __adjusted_n - __adjusted_global_id;
-            // Keep these statements in the same scope to allow for better memory alignment
-            new (&__res.__v) _Tp(__unary_op(__adjusted_global_id, __acc...));
-            for (_Size __i = 1; __i < __items_to_process; ++__i)
-                __res.__v = __binary_op(__res.__v, __unary_op(__adjusted_global_id + __i, __acc...));
+            const _Size __remainder = __adjusted_n - __start_idx;
+            const _Size __no_iters = std::min(__remainder, __max_iters);
```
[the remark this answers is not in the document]
danhoeflinger: It may of course be nice to replace these checks with some comment saying bounds checking isn't required because of previous checks.

julianmi:
Given the complexity of the indexing changes and the limited user benefits, I think it's best to postpone this to the 2022.7 release.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -456,96 +457,75 @@ __parallel_transform_reduce(oneapi::dpl::__internal::__device_backend_tag __back
...
-            return __parallel_transform_reduce_mid_impl<_Tp, 256, 32, 16, _Commutative>(
-                __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __n, __reduce_op, __transform_op, __init,
-                ::std::forward<_Ranges>(__rngs)...);
-        }
-        else if (__n <= 67108864)
+    // Use single work group implementation if less than __max_iters_per_work_item elements per work-group.
+    // We can use 16-bit addressing since we have at most __max_work_group_size * __max_iters_per_work_item elements.
+    if (__n <= __max_elements_per_wg)
+    {
+        const auto __n_short = static_cast<std::uint16_t>(__n);
```
danhoeflinger, line 482: While I don't disagree with this statement in the comments, it is dependent on the numbers above.  Can we do more to protect from future bugs?  
A couple options:
1) Add an assert that `__max_work_group_size * __max_iters_per_work_item`  is less than numerical limits of uint16_t. 
2) Add a comment to the empirically found tuning parameters, mentioning that index types might need to be adjusted if there are any changes to these values.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -456,96 +457,75 @@ __parallel_transform_reduce(oneapi::dpl::__internal::__device_backend_tag __back
...
+            __iters_per_work_item, __reduce_op, __transform_op, __init, std::forward<_Ranges>(__rngs)...);
+    }
+    // Use two-step tree reduction.
+    // First step reduces __work_group_size * __iters_per_work_item_device_kernel elements.
+    // Second step reduces __work_group_size * __iters_per_work_item_work_group_kernel elements.
+    // We can use 32-bit addressing since we have at most (__max_work_group_size * __max_iters_per_work_item) ^ 2
+    // elements.
+    if (__n <= __max_elements_per_wg * __max_elements_per_wg)
+    {
+        const auto __n_short = static_cast<std::uint32_t>(__n);
```
danhoeflinger, line 497: Same comment about types as above.

danhoeflinger (commented):
Only a minor comment.  Otherwise, I think this looks good, especially now since we have some time to react to any issues which may arise before a release.  

I went through the PR with fresh eyes and couldn't find any real issues.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
julianmi: I've added a comment and an assert.

danhoeflinger: I think a `static_assert` of the "max" constants vs the numerical max is the better option, because with this assert we wont know about the issue until a person runs an example in debug mode which has more elements than the type can hold, which may not be at the same time as the tuning parameters are tweaked.
  
If we use a static assert, we can know as soon as we compile new tuning parameters which breaks our type rules.

julianmi: Good point. I've changed it.

danhoeflinger (dismissed):
LGTM.  Probably good to at least check for objections from others who have reviewed this PR before merging. 
(and wait for green CI)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -47,24 +47,37 @@ class __reduce_mid_work_group_kernel;
 template <typename... _Name>
 class __reduce_kernel;
 
+// Adjust number of sequential operations per work-item based on the vector size. Single elements are kept to
+// improve performance of small arrays or remainder loops.
+template <std::uint8_t _VecSize, typename _Size>
+auto
+__adjust_iters_per_work_item(_Size __iters_per_work_item) -> _Size
```
SergeyKopienko, line 54:
```suggestion
_Size
__adjust_iters_per_work_item(_Size __iters_per_work_item)
```

julianmi: Done

julianmi:
> @julianmi, how do you think, should we introduce some type for the `union`
> 
> ```c++
>     union __storage
>     {
>         _Tp __v;
>         __storage() {}
>     };
> ```
> 
> ?

I've added a union type to reduce the code duplication.

SergeyKopienko (dismissed):
LGTM

mmichel11 (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -47,11 +47,19 @@ class __reduce_mid_work_group_kernel;
 template <typename... _Name>
 class __reduce_kernel;
 
+// Storage helper since _Tp may not have a default constructor.
+template <typename _Tp>
+union __storage
```
danhoeflinger, line 52: If we are going to lift this type definition out, we probably need to rename it as well.

SergeyKopienko: Additionally if we are going to move list this type definition out, we may cover the case when we have array of elements too.

danhoeflinger: I'm not sure about the array of elements, perhaps that reaches too far beyond the scope of this PR, but maybe something like  `__delayed_ctor_storage`?  

I think we need something which describes its purpose.

SergeyKopienko: `__optional_ctor_storage` ?

danhoeflinger: I suppose the first is the array case Sergey was referring to,  I'd be fine with leaving that one out for now to limit the scope of the PR if it makes it significantly more complicated.

SergeyKopienko: I propose to make additional changes with it in some separate PR.

danhoeflinger: Sure, for this PR, lets just rename it, we can unify, etc. in a separate PR.
My vote is for `__lazy_ctor_storage` because I think optional advertises more functionality than is provided here.

julianmi: Thanks for this discussion. I agree that larger changes are outside the scope of this PR and change the naming to `__lazy_ctor_storage`.

SergeyKopienko (approved):
LGTM

mmichel11 (approved):

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -202,103 +202,146 @@ struct __init_processing
...
+    template <typename _NDItemId, typename _Size, typename _Res, typename... _Acc>
+    void
+    operator()(const _NDItemId& __item_id, const _Size& __n, const _Size& __iters_per_work_item,
+               const _Size& __global_offset, const bool __is_full, const _Size __n_groups, _Res& __res,
+               const _Acc&... __acc) const
+    {
+        const _Size __global_idx = __item_id.get_global_id(0);
+        // Check if there is any work to do
+        if (__global_idx >= __n)
+            return;
```
SergeyKopienko, line 251: Now we have an error here: https://github.com/uxlfoundation/oneDPL/pull/2214

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -47,24 +47,41 @@ class __reduce_mid_work_group_kernel;
...
-    {
-        _Tp __v;
-        __storage() {}
-    } __result;
+    const _Size __group_size = __item_id.get_local_range().size();
+    __lazy_ctor_storage<_Tp> __result;
     // 1. Initialization (transform part). Fill local memory
-    __transform_pattern(__item_id, __n, /*global_offset*/ (_Size)0, __result, __acc...);
-    const _Size __n_items = __transform_pattern.output_size(__n, __group_size);
+    __transform_pattern(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0, __is_full,
```
SergeyKopienko, line 81: Now we have problem here: https://github.com/uxlfoundation/oneDPL/pull/2214

PR #1472, opened 2024-03-28
Enable binary comparison operations for internal tuple 

danhoeflinger:
Summary
---
This PR enables comparison `<`, `>`,`<=`,`>=` of our internal tuple `oneapi::dpl::__internal::tuple<T...>` with another `oneapi::dpl::__internal::tuple<U...>`  or `std::tuple<U...>` with equal number of elements where corresponding elements from rhs and lhs are comparable.

We use the std::tuple implementation for these operations by static casting and repeating the operation.

We extend the same to `==` and `!=`, where we previously implemented the operation directly.

Details
---
We already provide an implicit conversion operator to a `std::tuple`, and also are able to construct from a `std::tuple`.  However, when a template function is considered for overload resolution it does type deduction and will only use exact matches.  It will not use any implicit conversion to build the candidate list. Therefore, this implicit conversion is not used when searching for resolution for `operator<`, and we cannot take advantage of `std::tuple`'s implementation of these operators (at least without extra effort).

We use a macro to unify implementation of all binary operators in the same fashion and reduce code to a minimum.


tomflinda (approved):
LGTM

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -424,16 +424,34 @@ struct tuple<T1, T...>
...
+    {                                                                                                                  \
+        return __lhs __OPERATOR static_cast<std::tuple<T1, T...>>(__rhs);                                              \
+    }
+
+    BINARY_OPERATOR_OVERLOAD(==)
+    BINARY_OPERATOR_OVERLOAD(!=)
+    BINARY_OPERATOR_OVERLOAD(<)
+    BINARY_OPERATOR_OVERLOAD(<=)
+    BINARY_OPERATOR_OVERLOAD(>)
+    BINARY_OPERATOR_OVERLOAD(>=)
```
adamfidel, line 454: Can we undefine the `BINARY_OPERATOR_OVERLOAD` macro after we create the overloads? Otherwise, this macro would be present in all downstream user code.

danhoeflinger: Yes, good point.  Fixed.  I've also fixed the naming of the macro.

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -424,16 +424,36 @@ struct tuple<T1, T...>
...
+// This macro is used to define binary operators for comparison between oneAPI::dpl::__internal::tuple and
+//  std::tuple when they have the same number of template arguments and the individual elements which are invdividually
+//  comparable. These overloads are required because for overload resolution of template arguments, exact match is
+//  required and the compiler does not consider implicit conversions. We also allow std::tuple implementation to
+//  implement the operators themselves, which guarantees a match with oneAPI::dpl::__internal::tuple
+#define _TUPLE_BINARY_OPERATOR_OVERLOAD(__OPERATOR)                                                                    \
+    template <typename _U1, typename... _U, typename = ::std::enable_if_t<(sizeof...(_U) == sizeof...(T))>>            \
+    friend bool operator __OPERATOR(const tuple& __lhs, const oneapi::dpl::__internal::tuple<_U1, _U...>& __rhs)       \
+    {                                                                                                                  \
+        return static_cast<std::tuple<T1, T...>>(__lhs) __OPERATOR static_cast<std::tuple<_U1, _U...>>(__rhs);         \
```
adamfidel, line 436: Slightly pedantic, but `std::` should be `::std::` to match the rest of the methods in this class.

danhoeflinger: So, I think our mindset has changed on `::std` vs `std`.  Tagging @MikeDvorskiy.  We had talked about adding this to some sort of a style document or at least making an announcement, but that hasn't happened yet.
 
I was told that it is unnecessary to use `::std` vs `std` and that new code should be the later.  Similar to formatting though, we don't want to ruin the "blame" by changing it aggressively everywhere.

danhoeflinger:
I'm working on adding some unit testing for tuple to the test suite.  Once I add that, I will probably revert the changes to `zip_iterator.pass` .

include/oneapi/dpl/pstl/tuple_impl.h
danhoeflinger: Yes, we need to propagate that decision out and also codify it somewhere here in the repo in some sort of a style guide, I think.  We could even simply add a section to `CONTRIBUTING.md` with some general rules and best practices.

adamfidel:
I was curious if the three provided functions in the macro were enough to cover all of the cases to compare. But I enumerated them and it does seem like it is sufficient.

There are three tuple types we would like to compare:
1. `tuple` (self with `T1, T...`)
2. `tuple<_U1, _U...>` (self with potentially different types)
3. `std::tuple<_U1, _U...>` (standard tuple with potentially different types


| lhs | rhs | matching candidate |
|--|--|--|
| 1 | 1 | `__OPERATOR(const tuple& __lhs, const oneapi::dpl::__internal::tuple<_U1, _U...>& __rhs)`
| 1 | 2 | `__OPERATOR(const tuple& __lhs, const oneapi::dpl::__internal::tuple<_U1, _U...>& __rhs)`
| 1 | 3 | `__OPERATOR(const tuple& __lhs, const std::tuple<_U1, _U...>& __rhs)` 
| 2 | 1 | `__OPERATOR(const tuple& __lhs, const oneapi::dpl::__internal::tuple<_U1, _U...>& __rhs)`
| 2 | 2 | `__OPERATOR(const tuple& __lhs, const oneapi::dpl::__internal::tuple<_U1, _U...>& __rhs)`
| 2 | 3 | `__OPERATOR(const tuple& __lhs, const std::tuple<_U1, _U...>& __rhs)` 
| 3 | 1 | `__OPERATOR(const std::tuple<_U1, _U...>& __lhs, const tuple& __rhs)`
| 3 | 2 | `__OPERATOR(const std::tuple<_U1, _U...>& __lhs, const tuple& __rhs)`
| 3 | 3 | defined by standard

Based on this, it seems like the three functions that are defined are sufficient.

danhoeflinger:
> Based on this, it seems like the three functions that are defined are sufficient.

Yes, thanks for doing this enumeration of the possibilities.  Working on a test which should prove that out as well.

danhoeflinger:
OK, now all those options you mention should be covered by the unit test.  I've reverted the `zip_iterator` test changes in favor of unit testing.  

I've also rebased to have 2 commits so that we can separate the unit test addition in the main git log from the tuple changes.

tomflinda (approved):

SergeyKopienko (commented):
I think better if we will support all restrictions for these operators like in https://en.cppreference.com/w/cpp/utility/tuple : all these operators will be removed in C++20.

test/parallel_api/iterator/tuple_unit.pass.cpp
```diff
@@ -0,0 +1,100 @@
...
+    swap(onedpl_t5, onedpl_t6);
+    EXPECT_TRUE(((onedpl_t1 == onedpl_t5) && (onedpl_t6 == onedpl_t2)),
+                "swap of oneDPL tuple provides incorrect results");
+}
+
+int
+main()
+{
+
+    test_tuple(std::tuple<int, int, int>{1, 2, 3}, std::tuple<int, int, int>{1, 2, 3});
```
SergeyKopienko, line 93: Does it make sense to add tests for Kernels too in this test?

danhoeflinger: Do you mean calling sycl kernels or oneDPL algos with oneDPL internal tuples?  Checking `is_trivially_copyable` should check that this is possible.  Also, we have other tests which utilize `zip_iterator` with the algorithms which should provide system level testing.

My goal when adding this new test is to unit test tuple for some specific attributes we want it to have.

SergeyKopienko:
> > I think better if we will support all restrictions for these operators like in https://en.cppreference.com/w/cpp/utility/tuple : all these operators will be removed in C++20.
> 
> @SergeyKopienko Can you be more specific about what you mean by this comment? Do you mean we should provide a different implementation for C++20 guarded by preprocessor logic?
> 
> Also, these operators are removed in C++20, but only in favor of the `operator<=>` which supports all of them with a single operator. From the outside looking in, there shouldn't be a difference.

For me it's looks like we should implement all these operators only for C++17 version and implement `<=>` operator for C++20.

danhoeflinger:
> For me it's looks like we should implement all these operators only for C++17 version and implement `<=>` operator for C++20.

The way it is currently, it works for both C++17 and C++20 and offers the support and comparison with `std::tuple`.

If we change it to be `operator<=>` when C++20 is defined (removing the individual operators), we need to be in lock step with when `std::tuple::operator<=>` was included in the implementation of C++20 features, otherwise it will be broken.  If we include both at the same time, it is technically ambiguous and will at least provide warnings.

I think the only thing we miss out on in c++20 with the current implementation is if someone explicitly uses the `<=>` operator (which seems both unlikely and not a case we care much about).

SergeyKopienko:
> > For me it's looks like we should implement all these operators only for C++17 version and implement `<=>` operator for C++20.
> 
> I'm wondering if it is a good idea after some exploration...
> 
> The way it is currently, it works for both C++17 and C++20 and offers the support and comparison with `std::tuple`.
> 
> If we change it to be `operator<=>` when C++20 is defined (removing the individual operators), we need to be in lock step with when `std::tuple::operator<=>` was included in the implementation of C++20 features, otherwise it will be broken. If we include both individual operators and `<=>` at the same time, it is technically ambiguous and will at least provide warnings.
> 
> I think the only thing we miss out on in c++20 with the current implementation is if someone explicitly uses the `<=>` operator (which seems both unlikely and not a case we care much about).

Understood, ok.

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -424,16 +424,36 @@ struct tuple<T1, T...>
...
+    template <typename... _U, typename = std::enable_if_t<sizeof...(_U) == (sizeof...(T) + 1)>>                        \
+    friend constexpr bool operator __OPERATOR(const tuple& __lhs, const std::tuple<_U...>& __rhs)                      \
+    {                                                                                                                  \
+        return static_cast<std::tuple<T1, T...>>(__lhs) __OPERATOR __rhs;                                              \
+    }                                                                                                                  \
+    template <typename... _U, typename = std::enable_if_t<sizeof...(_U) == (sizeof...(T) + 1)>>                        \
+    friend constexpr bool operator __OPERATOR(const std::tuple<_U...>& __lhs, const tuple& __rhs)                      \
+    {                                                                                                                  \
+        return __lhs __OPERATOR static_cast<std::tuple<T1, T...>>(__rhs);                                              \
+    }
```
danhoeflinger, line 447: Question for the reviewers:

Does anyone know if there a way to enable forwarding (universal) references and perfect forwarding for something like this without enumerating all combinations of lvalue and rvalue refs for the two input parameters for each function call?

We still need to requires the type is a `std::tuple<U...>` / `oneapi::dpl::__internal::tuple<U...>` respectively, but we would like to use forwarding (universal) references.

danhoeflinger: I've created #1481 as an exploration of another approach.  Please take a look there.

MikeDvorskiy (commented):
Let me ask: what's  a motivation for that internal::tuple functionality extending?
Other words, where in OneDPL impl do we need such functionality?

danhoeflinger:
> Let me ask: what's a motivation for that internal::tuple functionality extending? Other words, where in OneDPL impl do we need such functionality?

Example:
Someone has a `zip_iterator` of two sequences of integers zipped together.   They call `oneapi::dpl::sort` with no comparator.  This will compare each element (onedpl tuple) using `operator<`.  There is a well defined semantic for comparing `std::tuple` with `operator<`.  

Then, if we are covering this case, why should it work with one such operator and not others?  

Similarly, since we return this tuple type from `zip_iterator` where the user may then use it, why not allow it to have the same semantics (comparable with different element types) and be interoperable with `std::tuple`?

I'd say the most important use case is to (1) allow binary comparison between two same element-typed internal tuples as that is what we may encounter within oneDPL algorithm calls.  This allows a much simpler implementation.  

Second most interesting would be (2) comparisons between onedpl tuples with different element types as users could contrive examples which this is necessary.

Least interesting is (3) interoperability with `std::tuple`.  

I'd be willing to drop (3) from this PR, it causes the most complication in implementation and has the least utility.

include/oneapi/dpl/pstl/tuple_impl.h
danhoeflinger: closed that other approach in favor of the one here to avoid any copies.

danhoeflinger:
> I'd be willing to drop (3) from this PR, it causes the most complication in implementation and has the least utility.

I've re-organized the code to make clear which code is to support which functionality.  The first set of operators covers 1+2, the second and third sets of operators are required to enable (3).

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -424,16 +424,128 @@ struct tuple<T1, T...>
...
+        return !(__lhs < __rhs);
+    }
+
+    // The following operators enable support of binary comparison of oneDPL internal tuples with std::tuple as rhs
+    // with the same number of types where individual elements are binary comparable
+    template <typename... _U, typename = std::enable_if_t<sizeof...(_U) == (sizeof...(T) + 1)>>
+    friend constexpr bool
+    operator==(const tuple& __lhs, const std::tuple<_U...>& __rhs)
+    {
+        return __lhs.holder.value == std::get<0>(__rhs) && __lhs.next == oneapi::dpl::__internal::get_tuple_tail(__rhs);
```
SergeyKopienko, line 476: Looks like here we will make copying of the rest of our tuple too...

danhoeflinger: Yes, you are correct.  Working on a fix...

danhoeflinger: OK I believe we should be free from copies now.

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -289,6 +306,36 @@ struct __copy_assignable_holder<_Tp, false> : oneapi::dpl::__internal::__value_h
     operator=(__copy_assignable_holder&& other) = default;
 };
 
+template <typename _Tuple1, typename _Tuple2, int I = 0>
+constexpr bool
+__equal(const _Tuple1& __lhs, const _Tuple2& __rhs)
+{
+    if constexpr (I < std::tuple_size_v<_Tuple1>)
```
SergeyKopienko, line 313: May be I didn't quite understood the logic here.
But as for me it's looks like an logical error due
https://en.cppreference.com/w/cpp/utility/tuple/operator_cmp
`If sizeof...(TTypes) does not equal sizeof...(UTypes), or [std::get](http://en.cppreference.com/w/cpp/utility/variant/get)<i>(lhs) == [std::get](http://en.cppreference.com/w/cpp/utility/variant/get)<i>(rhs) is not a valid expression returning a type that is convertible to bool(until C++23) for any i in [​0​, sizeof...(Types)), the program is ill-formed.`

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -289,6 +306,36 @@ struct __copy_assignable_holder<_Tp, false> : oneapi::dpl::__internal::__value_h
...
+    else
+    {
+        return true;
+    }
+}
+template <typename _Tuple1, typename _Tuple2, int I = 0>
+constexpr bool
+__less(const _Tuple1& __lhs, const _Tuple2& __rhs)
+{
+    if constexpr (I < std::tuple_size_v<_Tuple1>)
```
SergeyKopienko, line 327: The same

include/oneapi/dpl/pstl/tuple_impl.h
SergeyKopienko: And what about the case when `N == 0` ?

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -289,6 +306,36 @@ struct __copy_assignable_holder<_Tp, false> : oneapi::dpl::__internal::__value_h
...
+    {
+        return std::get<I>(__lhs) < std::get<I>(__rhs) ||
```
SergeyKopienko, line 329: Is it an error?
I mean that if `std::get<I>(__lhs) > std::get<I>(__rhs)` then we shouldn't call anything else, I think:
https://en.cppreference.com/w/cpp/utility/tuple/operator_cmp 
```C++
template< class... TTypes, class... UTypes >
constexpr bool operator<( const [std::tuple](http://en.cppreference.com/w/cpp/utility/tuple)<TTypes...>& lhs,
                          const [std::tuple](http://en.cppreference.com/w/cpp/utility/tuple)<UTypes...>& rhs );
(3)
`true` if the first non-equivalent element in `lhs` is less than the one in `rhs`, `false` if the first non-equivalent element in `rhs` is less than the one in `lhs` or there is no non-equivalent element. For two empty tuples, returns `false`.
```

SergeyKopienko (approved):
LGTM

danhoeflinger:
> @danhoeflinger It's not the result of you PR, but for `operator=(const tuple<U1, U...>& other)` what about the case when `this == &other' ? I mean probably we should avoid self-assignment...

I believe the case you speak of cannot happen here because there is a defaulted copy assignment operator which should handle the cases where the type of the `other` matches `this` (when the types don't match, the object instance itself wont either).  The implementation you are referring to without the self-assignment check is for a tuple with arbitrary (but convertible element types).

SergeyKopienko:
> > @danhoeflinger It's not the result of you PR, but for `operator=(const tuple<U1, U...>& other)` what about the case when `this == &other' ? I mean probably we should avoid self-assignment...
> 
> I believe the case you speak of cannot happen here because there is a defaulted copy assignment operator which should handle the cases where the type of the `other` matches `this` (when the types don't match, the object instance itself wont either). The implementation you are referring to without the self-assignment check is for a tuple with arbitrary (but convertible element types).

OK, np.

danhoeflinger:
After an offline discussion with @rarutyun , we have decided not to target the coming release with this PR, and take some time to investigate some context around it.  Specifically, we want to take a look at `std::sort` with a vector of `std::tuple`, and with `zip_iterator`, to understand if there are any other limitations which hold back such a case.

SergeyKopienko:
@danhoeflinger How do you think, should we support three-way comparison `<=>` too (https://en.cppreference.com/w/cpp/language/operator_comparison)?

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -456,15 +612,41 @@ struct tuple<>
...
     operator=(const tuple&) = default;
     tuple&
     operator=(const ::std::tuple<>& /*other*/)
     {
         return *this;
     }
-    friend bool
-    operator==(const tuple& /*__lhs*/, const tuple& /*__rhs*/)
+    friend constexpr bool
+    operator==(const tuple& __lhs, const tuple& __rhs)
```
akukanov, line 624: Unused function parameters should better be unnamed, to avoid possible compilation warnings.

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -37,6 +37,23 @@ struct tuple;
 
 namespace std
 {
+//forward declare new std::get<I>() functions so they can be used in impl
```
akukanov, line 40: Move these declarations down, below the specializations of `tuple_element` etc.? They use `std::tuple_element_t`, and perhaps it is even OK for forward declarations but why not avoid this concern.

akukanov:
Regarding `operator<=>` - since there are internal helpers `__equal` and `__less` anyway, why not implement a three-way comparison helper instead and use it in all operators? Then if the feature macro is set, we can define just `==` and `<=>` and let the C++ implementation do the rest, otherwise we define everything,

danhoeflinger:
> Regarding `operator<=>` - since there are internal helpers `__equal` and `__less` anyway, why not implement a three-way comparison helper instead and use it in all operators? Then if the feature macro is set, we can define just `==` and `<=>` and let the C++ implementation do the rest, otherwise we define everything,

I suppose we could do this, but we would want to implement a three-way comparison helper in different ways different ways depending on if `__cpp_lib_three_way_comparison` is set, because we rely upon the building block of comparing the elements themselves (and we wont have access to `<=>` for the elements without that feature macro).

I may be misunderstanding, but I think it may be better to just use `__equal` and `__less` directly when `__cpp_lib_three_way_comparison` is not set, rather than building `<=>` and then decoding it back into individual operators. If we don't do it this way, I think we may be adding unnecessary instructions in some cases.

akukanov:
> we would want to implement a three-way comparison helper in different ways depending on if `__cpp_lib_three_way_comparison` is set

Not sure about that. I mean, we can, but that does not mean we should. We can as well do something like this:
```
template <typename _Tuple1, typename _Tuple2, int I = 0>
constexpr int
__three_way_comp(const _Tuple1& __lhs, const _Tuple2& __rhs)
{
    if constexpr (I < std::tuple_size_v<_Tuple1>)
    {
        auto __left = std::get<I>(__lhs);
        auto __right = std::get<I>(__rhs);
        if (__left == __right)
            return oneapi::dpl::__internal::__less<_Tuple1, _Tuple2, I + 1>(__lhs, __rhs));
        else if (__left < __right)
            return -1;
        else
            return 1;
    else
    {
        return 0;
    }
}
```
The key difference perhaps is that both `==` and `<` are always required for tuple parameters.

But I am not that bothered about supporting the spaceship operator. Duplicating essentially the same code three times bothers me much more in this patch.

danhoeflinger:
> The key difference perhaps is that both `==` and `<` are always required for tuple parameters.

Yes, these are the extra instructions I was referring to.

> But I am not that bothered about supporting the spaceship operator. Duplicating essentially the same code three times bothers me much more in this patch.

Fair enough.  Yes, I spent some time trying to figure out how to avoid replication this but was unable to come up with any satisfactory solution.  We could decline to support all combinations, but it seems like if we are adding this support we should allow it to work with `std::tuple` fully and across non-matching but comparable element types.  We could possibly consolidate the code some with preprocessor macros, but I think that may be worse in the end.

The helpers at least allow us to only repeat the very simple portion of this.

danhoeflinger:
@SergeyKopienko @akukanov 
After exploring this a little bit, and running into a few complexities of the spaceship operator, including when to use strong ordering / partial ordering / weak ordering for different element types in the inputs.  I propose to merge this PR with the current level of functionality (no spaceship), and if we deem it important to support in the future, we can return to this and add support later.  

My reasoning:

1) It does not seem trivial to add, as it brings additional complexities. For instance, I believe the type of return value should depend upon the types compared (strong_ordering, partial_ordering, weak_ordering).
2) No one is asking for the spaceship, **but they are asking for the C++17 comparision operators**.
3) It doesn't seem to solve the frustrating things about this PR (meaning: repetitions of implementations).  We have to support when spaceship doesn't exist as well, so it will be at least as complicated.

All considered, it seems to me to be not worth the time and effort it would take to accomplish.

I am open to suggestion on how to reduce the redundancy in the current functionality.

include/oneapi/dpl/pstl/tuple_impl.h
danhoeflinger: done, thanks.

include/oneapi/dpl/pstl/tuple_impl.h
danhoeflinger: removed, thanks

danhoeflinger (commented):
After some help from @akukanov, I was able to consolidate 3 the different implementations of the comparators using some trait definitions.  

Its a little tricky because we have to avoid ambiguous definitions of the same comparator from two separate instantiations of the internal tuple.  The logic in the current set of `enable_if` is able to work by only defining comparisons between two different internal tuples in the instantiation matching the `lhs` type.

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -289,6 +306,71 @@ struct __copy_assignable_holder<_Tp, false> : oneapi::dpl::__internal::__value_h
...
+template <typename... Ts>
+struct __is_std_tuple<std::tuple<Ts...>> : std::true_type
+{
+};
+
+template <typename T>
+inline constexpr bool __is_std_tuple_v = __is_std_tuple<T>::value;
+
+template <typename T>
+inline constexpr bool __is_tuple_comparable_v = __is_std_tuple_v<T> || __is_internal_tuple_v<T>;
```
akukanov, line 372: Since the enable conditions for all 6 operators are the same, it makes sense to have a helper for these, like `__are_comparable_tuples_v<U,V>`. In that case, `__is_tuple_comparable_v` would only be used once, and I would not even create it then - particularly because spelling it explicitly would make the asymmetry of `(this tuple U, internal or standard tuple V) or (standard tuple U, this tuple V)` more visible. And the comment explaining this asymmetry would also be useful I think.

danhoeflinger: Yes, I was thinking about how to consolidate that logic.  It needs knowledge of what "my internal tuple" type is, so it either needs to be created within the tuple struct, or accept it as a template argument as well.  I'll consolidate this and add a comment explaining the logic.

danhoeflinger: I've consolidated the logic to one place and added a comment.

akukanov (commented):
The implementation looks good to me. The only question is whether `oneapi::dpl::__internal` is justified when used within the same namespace; but I do not mind it either way.
I have not reviewed the test.

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -49,6 +49,23 @@ struct tuple_element<0, oneapi::dpl::__internal::tuple<T, Rest...>>
...
+constexpr ::std::tuple_element_t<_Idx, oneapi::dpl::__internal::tuple<_Tp...>> const&
+get(const oneapi::dpl::__internal::tuple<_Tp...>&);
+
+template <size_t _Idx, typename... _Tp>
+constexpr ::std::tuple_element_t<_Idx, oneapi::dpl::__internal::tuple<_Tp...>>&&
+get(oneapi::dpl::__internal::tuple<_Tp...>&&);
+
+template <size_t _Idx, typename... _Tp>
+constexpr ::std::tuple_element_t<_Idx, oneapi::dpl::__internal::tuple<_Tp...>> const&&
+get(const oneapi::dpl::__internal::tuple<_Tp...>&&);
```
SergeyKopienko, line 67:
```suggestion
template <size_t _Idx, typename... _Tp>
constexpr std::tuple_element_t<_Idx, oneapi::dpl::__internal::tuple<_Tp...>>&
get(oneapi::dpl::__internal::tuple<_Tp...>&);

template <size_t _Idx, typename... _Tp>
constexpr std::tuple_element_t<_Idx, oneapi::dpl::__internal::tuple<_Tp...>> const&
get(const oneapi::dpl::__internal::tuple<_Tp...>&);

template <size_t _Idx, typename... _Tp>
constexpr std::tuple_element_t<_Idx, oneapi::dpl::__internal::tuple<_Tp...>>&&
get(oneapi::dpl::__internal::tuple<_Tp...>&&);

template <size_t _Idx, typename... _Tp>
constexpr std::tuple_element_t<_Idx, oneapi::dpl::__internal::tuple<_Tp...>> const&&
get(const oneapi::dpl::__internal::tuple<_Tp...>&&);
```

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -289,6 +306,82 @@ struct __copy_assignable_holder<_Tp, false> : oneapi::dpl::__internal::__value_h
...
+inline constexpr bool __is_std_tuple_v = __is_std_tuple<T>::value;
+
+// Either LHS = this internal tuple type + RHS = any compatible tuple type
+// or     LHS = std::tuple               + RHS = this internal tuple type
+// Note: We do not allow LHS to be "any compatible type" with a RHS as this internal tuple type because then we would
+//       have ambiguity comaparing two different compatible internal tuple types, because both instantiations would
+//       create the necessary comparison. As written, the type instatation of the type in the LHS is always
+//       responsible for creating this comparison operator.
+template <typename _ThisTuple, typename _U, typename _V>
+inline constexpr bool __enable_comparison_op_v =
```
SergeyKopienko, line 378: Will we have any benefits here if use [std::disjunction](https://en.cppreference.com/w/cpp/types/disjunction) and then define `__enable_comparison_op_v` as `...::value` ?

include/oneapi/dpl/pstl/tuple_impl.h
danhoeflinger: fixed.

include/oneapi/dpl/pstl/tuple_impl.h
danhoeflinger: I suppose it would allow us to omit defining `__is_std_tuple_v` and `__is_internal_tuple_v`.  I'm not aware of any other real advantages here.  
I think we may get some very slight compile time improvement, because I think that more might be able to be "short circuited" here and not evaluated, but I can't imagine that is significant.

I prefer the current formulation because I find it makes the most clear structure for the logic below.

For a similar reason, I might be in favor of removing `oneapi::dpl::__internal::` as @akukanov  mentions.  However, I think it has been our most recent convention to use the fully qualified name at all times.

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -289,6 +306,82 @@ struct __copy_assignable_holder<_Tp, false> : oneapi::dpl::__internal::__value_h
...
+
+template <typename T>
+inline constexpr bool __is_std_tuple_v = __is_std_tuple<T>::value;
+
+// Either LHS = this internal tuple type + RHS = any compatible tuple type
+// or     LHS = std::tuple               + RHS = this internal tuple type
+// Note: We do not allow LHS to be "any compatible type" with a RHS as this internal tuple type because then we would
+//       have ambiguity comaparing two different compatible internal tuple types, because both instantiations would
+//       create the necessary comparison. As written, the type instatation of the type in the LHS is always
+//       responsible for creating this comparison operator.
```
SergeyKopienko, line 376:
```suggestion
// Either LHS = this internal tuple type + RHS = any compatible tuple type
// or     LHS = std::tuple               + RHS = this internal tuple type
// Note: We do not allow LHS to be "any compatible type" with a RHS as this internal tuple type because then we would
//       have ambiguity comparing two different compatible internal tuple types, because both instantiations would
//       create the necessary comparison. As written, the type instantiation of the type in the LHS is always
//       responsible for creating this comparison operator.
```

include/oneapi/dpl/pstl/tuple_impl.h
SergeyKopienko: ok.

include/oneapi/dpl/pstl/tuple_impl.h
danhoeflinger: Thanks, fixed.  

Not sure why that didn't get caught by codespell...

test/parallel_api/iterator/tuple_unit.pass.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+    EXPECT_EQ((t1 == t2), (onedpl_t1 == t2), "equality comparison does not match std::tuple");
+    EXPECT_EQ((t1 != t2), (onedpl_t1 != t2), "inquality comparison does not match std::tuple");
+    EXPECT_EQ((t1 < t2), (onedpl_t1 < t2), "less than comparison does not match std::tuple");
+    EXPECT_EQ((t1 <= t2), (onedpl_t1 <= t2), "less than or equal to comparison does not match std::tuple");
+    EXPECT_EQ((t1 > t2), (onedpl_t1 > t2), "greater than comparison does not match std::tuple");
+    EXPECT_EQ((t1 >= t2), (onedpl_t1 >= t2), "greater than or equal to comparison does not match std::tuple");
+
+    auto onedpl_t3 = to_onedpl_tuple(t1);
+    auto onedpl_t4 = to_onedpl_tuple(t2);
```
akukanov, line 62: Maybe `decltype(onedpl_t1) onedpl_t3{};` etc. - to ensure that the assignment really does the work, and not just "duplicates" construction?

test/parallel_api/iterator/tuple_unit.pass.cpp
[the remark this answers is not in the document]
danhoeflinger: Ah yes, I must've created this before we made that directory with the device copyable test.  "Unit" is for unit testing of tuple, no need if it is within "implementation_details`, we can perhaps call it `internal_tuple.pass`.

test/parallel_api/iterator/tuple_unit.pass.cpp
```diff
@@ -0,0 +1,191 @@
...
+                "swap of oneDPL tuple provides incorrect results");
+}
+
+constexpr void
+test_compile_time()
+{
+    constexpr std::tuple<int, int, uint64_t> std_tup1{1, 2, 3};
+    constexpr std::tuple<int, uint32_t, int> std_tup2{99, 2, 3};
+    constexpr oneapi::dpl::tuple<int, int, uint64_t> dpl_tup1{1, 2, 3};
+    constexpr oneapi::dpl::tuple<int, uint32_t, int> dpl_tup2{99, 2, 3};
```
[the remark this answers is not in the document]
danhoeflinger: Yes, this is a sizeable mistake, good catch.  Changing to internal tuple reveals many missing elements to handle `constexpr` properly.  None are complicated (adding `constexpr` to methods), but they need to be fixed to support this functionality.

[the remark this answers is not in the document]
danhoeflinger: Resolved the `constexpr` additions.  If we don't want to add these in this PR, we can strip out the constexpr changes, and this part of the testing.  I don't have a strong opinion here, it is not needed for the use case being requested.

test/parallel_api/iterator/tuple_unit.pass.cpp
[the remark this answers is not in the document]
danhoeflinger: renamed to `internal_tuple.pass` and moved to `implementation_details`.

test/parallel_api/iterator/tuple_unit.pass.cpp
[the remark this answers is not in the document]
danhoeflinger: fixed

test/parallel_api/iterator/tuple_unit.pass.cpp
danhoeflinger: Maybe I'm missing the point of your comment, but the highlighted lines are not the assignment being tested below.  These lines construct two onedpl internal tuples, and the subsequent two lines construct two std::tuples.  Then the later lines immediately preceding the `EXPECT_TRUE` calls are the assignment.

test/parallel_api/iterator/tuple_unit.pass.cpp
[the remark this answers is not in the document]
akukanov: The additions are not huge, but they seem kind of strange comparing to other code. I do not think we have / will ever have use cases for constexpr operations with the internal tuple. Do you have some scenario in mind where it could be useful? Doing it "just in case" does not sound justified.

test/parallel_api/iterator/tuple_unit.pass.cpp
akukanov: Currently, this is what happens with `onedpl_t3`:

```
    auto onedpl_t3 = to_onedpl_tuple(t1);
...
    onedpl_t3 = t1;
```
So the assignment sets essentially the same value as the constructor. That seems redundant at the very least. Since you intend to test the assignment, the value at construction should not matter, so for me it would be best to set no specific value. That would additionally test that the internal tuple is default-constructible/

test/parallel_api/iterator/tuple_unit.pass.cpp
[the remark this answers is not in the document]
danhoeflinger: My intention was just to get our internal tuple and its support for the binary comparison operators closer to `std::tuple` specification.  I don't have a specific use case in mind, and such a case probably does not exist. A user should only encounter these as a result of `zip_iterator` which don't have the ability to create tuples at compile time.

After consideration, I think it is probably best to remove support and the testing of `constexpr` part of this.  We can always add it in the future if we have a reason to do so.

test/parallel_api/iterator/tuple_unit.pass.cpp
danhoeflinger: Sure, that makes sense.  
The intention here was to do 
    auto onedpl_t3 = to_onedpl_tuple(t1);
...
    onedpl_t3 = t1;

danhoeflinger: done.

akukanov (approved):

PR #1475, opened 2024-04-04
Distributed-ranges namespace

mateuszpn:
Use cmake -DCMAKE_CXX_STANDARD=20 to include DR to your project


BenBrock:
This PR currently applies `clang-format` to a bunch of oneDPL files that are not part of distributed ranges.  We need to revert those changes, as formatting pre-existing code should not be part of this PR.  We should limit this PR to only include changes to distributed ranges.

BenBrock:
Please ignore my previous comment; I see now that this is a PR against the `distributed-ranges` branch, not against `main`, and it looks like the `distributed-ranges` branch is in good shape in terms of not applying these kinds of formatting changes.  I assume you'll fix this in this PR before merging with `distributed-ranges`.

I went ahead and opened a draft PR for the [whole `distributed-ranges` branch](https://github.com/oneapi-src/oneDPL/pull/1479), just so I can easily look at what we're adding.

PR #1476, opened 2024-04-04
Adding public utility for users to check if usm_allocators are supported

danhoeflinger:
Support for USM allocators depends on the implementation details of the C++ standard library used after #1438, and after doc changes of #1477.

This PR adds a public utility (`usm_allocated_vector_iterators_supported_v<ValueType>`) which users can use at compile time to check if usm_allocators are available to them as efficient input to oneDPL algorithms.  

- If `usm_allocated_vector_iterators_supported_v<ValueType>` evaluates to `true`, `std::vector<ValueType, usm_allocator>::iterator` input types are passed directly to sycl kernels, and are an efficient way of using oneDPL.  

- If `usm_allocated_vector_iterators_supported_v<ValueType>` evaluates to `false`, `std::vector<ValueType, usm_allocator>::iterator` input types are treated as if they were host-side iterators, and wrapped in `sycl::buffer` before use in sycl kernels.  This includes all that comes with that, an extra host-side copy in and out.  Also these cannot then be used as the source iterator for `permutation_iterator`.

### DIFF
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -88,10 +88,10 @@ on details of the C++ standard library implementation and what information about
 allocator is included in the ``std::vector::iterator`` type definition. If USM allocated
 vector iterators are not detectable with your C++ standard library, they will still function
 as inputs to oneDPL, but they will be treated as if they were host-side
-``std::vector::iterator`` as described in the section below. To guarantee no additional
-host-side copies, you can use ``std::vector::data()`` in combination with ``std::vector::size()``
-with USM allocated vectors to obtain a USM pointers. This will avoid extra host-side copies
-regardless of the C++ standard library implementation.
+``std::vector::iterator`` as described in the `Use Host-Side std::vector`_ section. To guarantee
+no additional host-side copies, you can use ``std::vector::data()`` in combination with
+``std::vector::size()`` with USM allocated vectors to obtain a USM pointer. This will avoid
+extra host-side copies regardless of the C++ standard library implementation.
 
 An example of ``std::vector`` with a USM allocator:
 
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -81,7 +81,19 @@ the buffer were created for the same queue. For example:
     return 0;
   }
 
-Alternatively, use ``std::vector`` with a USM allocator. For example:
+Alternatively, use ``std::vector`` with a USM allocator.
+
+Note: The ability to appropriately detect USM allocated ``std::vector::iterator`` depends
+on details of the C++ standard library implementation and what information about the
+allocator is included in the ``std::vector::iterator`` type definition. If USM allocated
+vector iterators are not detectable with your C++ standard library, they will still function
+as inputs to oneDPL, but they will be treated as if they were host-side
+``std::vector::iterator`` as described in the section below. To guarantee no additional
+host-side copies, you can use ``std::vector::data()`` in combination with ``std::vector::size()``
+with USM allocated vectors to obtain a USM pointers. This will avoid extra host-side copies
+regardless of the C++ standard library implementation.
+
+An example of ``std::vector`` with a USM allocator:
 
 .. code:: cpp
 
@@ -95,6 +107,9 @@ Alternatively, use ``std::vector`` with a USM allocator. For example:
     std::vector<int, decltype(alloc)> vec(n, alloc);
 
     std::fill(policy, vec.begin(), vec.end(), 42);
+
+    //alternative to use USM pointers:
+    // std::fill(policy, vec.data(), vec.data() + vec.size(), 42);
 
     return 0;
   }
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -90,8 +90,9 @@ vector iterators are not detectable with your C++ standard library, they will st
 as inputs to oneDPL, but they will be treated as if they were host-side
 ``std::vector::iterator`` as described in the `Use Host-Side std::vector`_ section. To guarantee
 no additional host-side copies, you can use ``std::vector::data()`` in combination with
-``std::vector::size()`` with USM allocated vectors to obtain a USM pointer. This will avoid
-extra host-side copies regardless of the C++ standard library implementation.
+``std::vector::size()`` with USM allocated vectors to obtain USM pointers which can be used as
+substitutes for ``std::vector::begin()`` and ``std::vector::end()``. This will avoid extra
+host-side copies regardless of the C++ standard library implementation.
 
 An example of ``std::vector`` with a USM allocator:
 
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -85,17 +85,21 @@ Alternatively, use ``std::vector`` with a USM allocator.
 
 Note: The ability to appropriately detect USM allocated ``std::vector::iterator`` depends
 on details of the C++ standard library implementation and what information about the
-allocator is included in the ``std::vector::iterator`` type definition. You can check
-if your C++ standard library implementation provides enough information by checking the
-value of ``oneapi::dpl::usm_allocated_vector_iterators_supported_v<ValueType>``. This will
-evaluate to ``true`` for C++ standard library implementations which allow oneDPL to support
-USM allocators with this ValueType and ``false`` otherwise. If usm_allocators are not
-supported with your C++ standard library, they will still function as inputs to oneDPL, but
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -233,6 +233,11 @@ __internal::sycl_iterator<access_mode::discard_read_write, T, Allocator> end(syc
     return __internal::sycl_iterator<access_mode::discard_read_write, T, Allocator>{buf,
                                                                                     __dpl_sycl::__get_buffer_size(buf)};
 }
+
+template <typename ValueType>
+constexpr static bool usm_allocated_vector_iterators_supported_v =
```
MikeDvorskiy, line 238: 1) I guess the traits doesn't depend on `ValueType`.
2) As an idea - to make such public trait more general, for example,
```
template <typename Container>
constexpr static bool is_usm_container_supported = ...
```

danhoeflinger: Sadly, I believe it can depend on the `ValueType`, for example `std::vector<bool>`.  GNU c++ standard library implementation normally includes allocator type in the iterator, but not for `std::vector<bool>` with the bitfield "optimization".

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -233,6 +233,11 @@ __internal::sycl_iterator<access_mode::discard_read_write, T, Allocator> end(syc
...
+    oneapi::dpl::__internal::__vector_iter_distinguishes_by_allocator<ValueType*>::value;
```
MikeDvorskiy, line 239: BTW, what if `ValueType` is a pointer to the host data?
It is rather a kind of general question connected to "Pass Data to algorithms" topic.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -149,6 +149,33 @@ struct _ModeConverter<access_mode::write>
...
+// std::vector<value_type, Alloc>::iterator can be distinguished between three different allocators, the
+// default, usm_shared, and usm_host. If all are distinct, it is very unlikely any non-usm based allocator
+// could be confused with a usm allocator.
+template <typename Iter, typename Void = void>
+struct __vector_iter_distinguishes_by_allocator : std::false_type
+{
+};
+template <typename Iter>
+struct __vector_iter_distinguishes_by_allocator<
+    Iter, std::enable_if_t<!std::is_same_v<__default_alloc_vec_iter<Iter>, __usm_shared_alloc_vec_iter<Iter>> &&
```
SergeyKopienko, line 172: Should we apply here `std::decay_t` for `Iter` template param?

danhoeflinger: Actually, this is already merged code from #1438 .  Working to rebase, to make this focus on just what is added by this PR.  
If we need to go back and fix something here, it will needs its own PR I think.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -206,6 +207,16 @@ struct is_passed_directly<Iter, ::std::enable_if_t<Iter::is_passed_directly::val
 {
 };
 
+//support std::vector::iterator with usm host / shared allocator as passed directly
+template <typename Iter>
+struct is_passed_directly<
+    Iter, std::enable_if_t<(std::is_same_v<Iter, oneapi::dpl::__internal::__usm_shared_alloc_vec_iter<Iter>> ||
```
SergeyKopienko, line 213: Should we apply here `std::decay_t` for `Iter` template param?

danhoeflinger: Same as above, this is merged code from #1438.  The duplicated conflict code has been fixed via rebase.  Any changes here will need a new PR.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
danhoeflinger: If I understand correctly, you suggest the following type:

`std::vector<int*, usm_allocator<int, ...>>`

In theory, a vector of host pointers will be very similar to a vector of integers I think.  

In principle, as long as they are not dereferencing the pointers on the device, there is nothing wrong with this.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -85,13 +85,17 @@ Alternatively, use ``std::vector`` with a USM allocator.
...
+allocator is included in the ``std::vector::iterator`` type definition. You can check
+if your C++ standard library implementation provides enough information by checking the
+value of ``oneapi::dpl::usm_allocated_vector_iterators_supported_v<ValueType>``. This will
+evaluate to ``true`` for C++ standard library implementations which allow oneDPL to support
+USM allocators with this ValueType and ``false`` otherwise. If usm_allocators are not
+supported with your C++ standard library, they will still function as inputs to oneDPL, but
+they will be treated as if they were host-side ``std::vector::iterator`` as described in
+the section below. To guarantee no additional host-side copies, you can use
+``std::vector::data()`` in combination with ``std::vector::size()`` with USM allocated
+vectors to obtain a USM pointers. This will avoid extra host-side copies regardless of the
```
dcbenito, line 97: The plurals don't match in this "...USM allocated vectors to obtain a USM pointers."

I suggest either, depending on if you are talking about a single pointer or multiple:

USM allocated vectors to obtain a USM pointer.
USM allocated vectors to obtain USM pointers.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -85,13 +85,17 @@ Alternatively, use ``std::vector`` with a USM allocator.
...
-vector iterators are not detectable with your C++ standard library, they will still function
-as inputs to oneDPL, but they will be treated as if they were host-side
-``std::vector::iterator`` as described in the section below. To guarantee no additional
-host-side copies, you can use ``std::vector::data()`` in combination with ``std::vector::size()``
-with USM allocated vectors to obtain a USM pointers. This will avoid extra host-side copies
-regardless of the C++ standard library implementation.
+allocator is included in the ``std::vector::iterator`` type definition. You can check
+if your C++ standard library implementation provides enough information by checking the
+value of ``oneapi::dpl::usm_allocated_vector_iterators_supported_v<ValueType>``. This will
+evaluate to ``true`` for C++ standard library implementations which allow oneDPL to support
```
dcbenito, line 91: Please add a comma between "implementations" and "which" and add an "s" to "allow". The final sentence should read: 

This will evaluate to ``true`` for C++ standard library implementations, which allows oneDPL to support USM allocators with this ValueType and ``false`` otherwise.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -85,13 +85,17 @@ Alternatively, use ``std::vector`` with a USM allocator.
...
+USM allocators with this ValueType and ``false`` otherwise. If usm_allocators are not
```
dcbenito, line 92: Should usm_allocators be in monospace font? (``usm_allocators`` )

dcbenito (changes requested):

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: I've properly rebased this PR so that this no longer repeated both here and in #1477.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: It should be USM allocators, thanks.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: I think there was some confusion by my wording.  I've expanded this some to hopefully  make things more clear.

dcbenito: Thanks for the update. My suggestions on the new content (ignore spacing; I broke things up for this review). 

This evaluates to ``true`` when using a C++ standard library implementation, which allows oneDPL to detect and
support USM allocators for ``ValueType`` and ``false``. 

When ``oneapi::dpl::usm_allocated_vector_iterators_supported_v<ValueType>`` evaluates to ``false``,
you should rely on USM pointers directly instead of using iterators to vectors with USM allocators.

danhoeflinger:
I think that with the direction we are going with #1477, this does not make sense to continue with such a public API.  
I prefer our decision to recommend against using USM allocated vector iterators in favor of USM pointers to data within them.  
Exposing a public API like this is probably too "in the weeds" for the user about the details of the C++ standard library implementations.  
Therefore, I'm closing this PR.

PR #1477, opened 2024-04-04
[Doc] Additional details for USM allocators with std::vector

danhoeflinger:
Separated out from #1438.  
Documentation changes to describe the details of usm allocators after that PR is merged.  
It also describes a more guaranteed way for proper handling of USM allocated data by directly obtaining USM pointers from a `std::vector`, as this does not rely on any implementation details of the C++ standard library implementation.

### DIFF
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -88,10 +88,10 @@ on details of the C++ standard library implementation and what information about
 allocator is included in the ``std::vector::iterator`` type definition. If USM allocated
 vector iterators are not detectable with your C++ standard library, they will still function
 as inputs to oneDPL, but they will be treated as if they were host-side
-``std::vector::iterator`` as described in the section below. To guarantee no additional
-host-side copies, you can use ``std::vector::data()`` in combination with ``std::vector::size()``
-with USM allocated vectors to obtain a USM pointers. This will avoid extra host-side copies
-regardless of the C++ standard library implementation.
+``std::vector::iterator`` as described in the `Use Host-Side std::vector`_ section. To guarantee
+no additional host-side copies, you can use ``std::vector::data()`` in combination with
+``std::vector::size()`` with USM allocated vectors to obtain a USM pointer. This will avoid
+extra host-side copies regardless of the C++ standard library implementation.
 
 An example of ``std::vector`` with a USM allocator:
 
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -81,7 +81,19 @@ the buffer were created for the same queue. For example:
     return 0;
   }
 
-Alternatively, use ``std::vector`` with a USM allocator. For example:
+Alternatively, use ``std::vector`` with a USM allocator.
+
+Note: The ability to appropriately detect USM allocated ``std::vector::iterator`` depends
+on details of the C++ standard library implementation and what information about the
+allocator is included in the ``std::vector::iterator`` type definition. If USM allocated
+vector iterators are not detectable with your C++ standard library, they will still function
+as inputs to oneDPL, but they will be treated as if they were host-side
+``std::vector::iterator`` as described in the section below. To guarantee no additional
+host-side copies, you can use ``std::vector::data()`` in combination with ``std::vector::size()``
+with USM allocated vectors to obtain a USM pointers. This will avoid extra host-side copies
+regardless of the C++ standard library implementation.
+
+An example of ``std::vector`` with a USM allocator:
 
 .. code:: cpp
 
@@ -95,6 +107,9 @@ Alternatively, use ``std::vector`` with a USM allocator. For example:
     std::vector<int, decltype(alloc)> vec(n, alloc);
 
     std::fill(policy, vec.begin(), vec.end(), 42);
+
+    //alternative to use USM pointers:
+    // std::fill(policy, vec.data(), vec.data() + vec.size(), 42);
 
     return 0;
   }
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -4,8 +4,8 @@ Pass Data to Algorithms
 You can use one of the following ways to pass data to an algorithm executed with a device policy:
 
 * ``oneapi:dpl::begin`` and ``oneapi::dpl::end`` functions
-* Unified shared memory (USM) pointers and ``std::vector`` with USM allocators
-* Iterators of host-side ``std::vector``
+* Unified shared memory (USM) pointers
+* Iterators or USM pointers to ``std::vector`` data
 
 .. _use-buffer-wrappers:
 
@@ -56,14 +56,9 @@ To use the functions, add ``#include <oneapi/dpl/iterator>`` to your code. For e
 Use Unified Shared Memory
 -------------------------
 
-The following examples demonstrate two ways to use the parallel algorithms with USM:
-
-* USM pointers
-* USM allocators
-
-If you have a USM-allocated buffer, pass the pointers to the start and past the end
-of the buffer to a parallel algorithm. Make sure that the execution policy and
-the buffer were created for the same queue. For example:
+If you have USM-allocated memory, pass the pointers to the start and past the end
+of the sequence to a parallel algorithm. Make sure that the execution policy and
+the USM-allocated memory were created for the same queue. For example:
 
 .. code:: cpp
 
@@ -81,52 +76,19 @@ the buffer were created for the same queue. For example:
     return 0;
   }
 
-Alternatively, use ``std::vector`` with a USM allocator.
-
-Note: The ability to appropriately detect USM allocated ``std::vector::iterator`` depends
-on details of the C++ standard library implementation and what information about the
-allocator is included in the ``std::vector::iterator`` type definition. If USM allocated
-vector iterators are not detectable with your C++ standard library, they will still function
-as inputs to oneDPL, but they will be treated as if they were host-side
-``std::vector::iterator`` as described in the `Use Host-Side std::vector`_ section. To guarantee
-no additional host-side copies, you can use ``std::vector::data()`` in combination with
-``std::vector::size()`` with USM allocated vectors to obtain USM pointers, which can be used as
-substitutes for ``std::vector::begin()`` and ``std::vector::end()``. This will avoid extra
-host-side copies regardless of the C++ standard library implementation.
-
-An example of ``std::vector`` with a USM allocator:
-
-.. code:: cpp
-
-  #include <oneapi/dpl/execution>
-  #include <oneapi/dpl/algorithm>
-  #include <sycl/sycl.hpp>
-  int main(){
-    const int n = 1000;
-    auto policy = oneapi::dpl::execution::dpcpp_default;
-    sycl::usm_allocator<int, sycl::usm::alloc::shared> alloc(policy.queue());
-    std::vector<int, decltype(alloc)> vec(n, alloc);
-
-    std::fill(policy, vec.begin(), vec.end(), 42);
-
-    //alternative to use USM pointers:
-    // std::fill(policy, vec.data(), vec.data() + vec.size(), 42);
-
-    return 0;
-  }
-
 When using device USM, such as allocated by ``malloc_device``, manually copy data to this memory
 before calling oneDPL algorithms, and copy it back once the algorithms have finished execution.
 
-Use Host-Side std::vector
+Use std::vector
 -----------------------------
 
-|onedpl_long| parallel algorithms can be called with ordinary (host-side) iterators, as seen in the
-example below.
-In this case, a temporary SYCL buffer is created, and the data is copied to this buffer.
-After processing on a device is complete, the modified data is copied from the temporary buffer back
-to the host container.
-For example:
+The following examples demonstrate two ways to use the parallel algorithms with ``std::vector``:
+
+* host allocators
+* USM allocators
+
+You can use iterators to host allocated ``std::vector`` data
+as shown in the following example:
 
 .. code:: cpp
 
@@ -140,4 +102,40 @@ For example:
     return 0;
   }
 
-Working with SYCL buffers is recommended to reduce data copying between the host and device.
+When using iterators to host allocated data, a temporary SYCL buffer is created, and the data
+is copied to this buffer. After processing on a device is complete, the modified data is copied
+from the temporary buffer back to the host container. While convenient, using host allocated
+data can lead to unintended copying between host and device. We recommend working with SYCL buffers
+or USM memory to reduce data copying between the host and device. 
+
+You can also use ``std::vector`` with a USM allocator, as shown in the following example:
+
+.. code:: cpp
+
+  #include <oneapi/dpl/execution>
+  #include <oneapi/dpl/algorithm>
+  #include <sycl/sycl.hpp>
+  int main(){
+    const int n = 1000;
+    auto policy = oneapi::dpl::execution::dpcpp_default;
+    sycl::usm_allocator<int, sycl::usm::alloc::shared> alloc(policy.queue());
+    std::vector<int, decltype(alloc)> vec(n, alloc);
+
+    // Recommended to use USM pointers:
+    std::fill(policy, vec.data(), vec.data() + vec.size(), 42);
+
+    // Iterators for USM allocators might require extra copying
+    // std::fill(policy, vec.begin(), vec.end(), 42);
+
+    return 0;
+  }
+
+Make sure that the execution policy and the USM-allocated memory were created for the same queue.
+When using USM allocators with ``std::vector``, it is recommended to use USM pointers as shown in
+the example above, rather than with iterators to ``std::vector``. In some implementations of the C++
+Standard Library, it is not possible to detect that iterators are pointing to USM-allocated data.
+In those cases, data will be treated as if it were host-side data, and will require extra copying
+between host and device. |onedpl_short| will avoid copies whenever it is possible with the C++
+Standard Library implementation.  However, to guarantee no unintended copying, use
+``std::vector::data()`` in combination with ``std::vector::size()`` when using USM-allocated data
+to retrieve USM pointers for the sequence as shown.
--- a/documentation/library_guide/parallel_api/pass_data_algorithms.rst
+++ b/documentation/library_guide/parallel_api/pass_data_algorithms.rst
@@ -44,10 +44,19 @@ To use the functions, add ``#include <oneapi/dpl/iterator>`` to your code. For e
   #include <oneapi/dpl/iterator>
   #include <sycl/sycl.hpp>
   int main(){
-    sycl::buffer<int> buf { 1000 };
+    std::vector<int> vec(1000);
+    // generate random data on host for simplicity
+    std::random_device r;
+    std::default_random_engine e1(r());
+    std::uniform_int_distribution<int> uniform_dist(-2000, 2000);
+    std::generate(vec.begin(), vec.end(), [&](){ return uniform_dist(e1); });
+
+    //create a buffer from host memory
+    sycl::buffer<int> buf { vec.data(), vec.size() };
     auto buf_begin = oneapi::dpl::begin(buf);
     auto buf_end   = oneapi::dpl::end(buf);
-    std::fill(oneapi::dpl::execution::dpcpp_default, buf_begin, buf_end, 42);
+
+    std::sort(oneapi::dpl::execution::dpcpp_default, buf_begin, buf_end);
     return 0;
   }
 
@@ -69,8 +78,13 @@ the USM-allocated memory were created for the same queue. For example:
     sycl::queue q;
     const int n = 1000;
     int* d_head = sycl::malloc_shared<int>(n, q);
+    // generate random data on host for simplicity
+    std::random_device r;
+    std::default_random_engine e1(r());
+    std::uniform_int_distribution<int> uniform_dist(-2000, 2000);
+    std::generate(d_head, d_head + n, [&](){ return uniform_dist(e1); });
 
-    std::fill(oneapi::dpl::execution::make_device_policy(q), d_head, d_head + n, 42);
+    std::sort(oneapi::dpl::execution::make_device_policy(q), d_head, d_head + n);
 
     sycl::free(d_head, q);
     return 0;
@@ -98,8 +112,12 @@ as shown in the following example:
   #include <vector>
   int main(){
     std::vector<int> vec( 1000 );
-    std::fill(oneapi::dpl::execution::dpcpp_default, vec.begin(), vec.end(), 42);
-    // each element of vec equals to 42
+    std::random_device r;
+    std::default_random_engine e1(r());
+    std::uniform_int_distribution<int> uniform_dist(-2000, 2000);
+    std::generate(vec.begin(), vec.end(), [&](){ return uniform_dist(e1); });
+
+    std::sort(oneapi::dpl::execution::dpcpp_default, vec.begin(), vec.end();
     return 0;
   }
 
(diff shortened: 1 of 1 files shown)


MikeDvorskiy:
it looks the code changes are caught into the PR....

danhoeflinger:
> it looks the code changes are caught into the PR....

Fixed with a rebase.  Thanks.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,7 +81,19 @@ the buffer were created for the same queue. For example:
...
 
-Alternatively, use ``std::vector`` with a USM allocator. For example:
+Alternatively, use ``std::vector`` with a USM allocator.
+
+Note: The ability to appropriately detect USM allocated ``std::vector::iterator`` depends
+on details of the C++ standard library implementation and what information about the
+allocator is included in the ``std::vector::iterator`` type definition. If USM allocated
+vector iterators are not detectable with your C++ standard library, they will still function
+as inputs to oneDPL, but they will be treated as if they were host-side
+``std::vector::iterator`` as described in the section below. To guarantee no additional
```
dcbenito, line 91: Can we change to "... as described in the Use Host-Side std::vector section." with a link to the section?

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,7 +81,19 @@ the buffer were created for the same queue. For example:
...
+host-side copies, you can use ``std::vector::data()`` in combination with ``std::vector::size()``
+with USM allocated vectors to obtain a USM pointers. This will avoid extra host-side copies
```
dcbenito, line 93: The tenses don't match in this "...USM allocated vectors to obtain a USM pointers."

I suggest either, depending on if you are talking about a single pointer or multiple:

- USM allocated vectors to obtain a USM pointer.
- USM allocated vectors to obtain USM pointers.

dcbenito (changes requested):

dcbenito (approved):
LGTM

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: Thanks, I've made the change.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: I added a little extra to be more specific about what is being suggested, but it should resolve this issue.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,7 +81,20 @@ the buffer were created for the same queue. For example:
...
+Alternatively, use ``std::vector`` with a USM allocator.
+
+Note: The ability to appropriately detect USM allocated ``std::vector::iterator`` depends
+on details of the C++ standard library implementation and what information about the
+allocator is included in the ``std::vector::iterator`` type definition. If USM allocated
+vector iterators are not detectable with your C++ standard library, they will still function
+as inputs to oneDPL, but they will be treated as if they were host-side
+``std::vector::iterator`` as described in the `Use Host-Side std::vector`_ section. To guarantee
+no additional host-side copies, you can use ``std::vector::data()`` in combination with
+``std::vector::size()`` with USM allocated vectors to obtain USM pointers which can be used as
```
dcbenito, line 93: Please add a comma between USM pointers and which.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
dcbenito: LGTM, aside from a comma instance (captured in a separate comment).

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
akukanov, line 1: Looks like changes in this file are accidental and not related to the purpose of the PR.

danhoeflinger: Oh, strange, yes.  I'm not quite sure how that happened, but that must've been from a dirty working copy or something.  I'll fix it momentarily...

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: thanks, fixed.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov, line 1: Instead of adding a long note to the USM part, I would suggest to rewrite the std::vector part following next. Rename it to remove the notion of host-side, and explain there the differences between using the vector with USM allocator vs. the standard allocator, describe the caveat (perhaps in a bit more concise way), make the recommendation to use data() and size(), etc. This will also require some changes at the beginning of the page where various ways of data passing are listed.

danhoeflinger: OK, I can do this.

danhoeflinger: Alright, hopefully that is more clear.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,47 +76,66 @@ the buffer were created for the same queue. For example:
...
\ No newline at end of file
+Make sure that the execution policy and the USM-allocated memory were created for the same queue.
+When using USM allocators with ``std::vector``, it is recommended to use USM pointers as shown in
+the example above, rather than with iterators to ``std::vector``. In some implementations of the C++
+Standard Library, it is not possible to detect that iterators are pointing to USM-allocated data.
+In those cases, data will be treated as if it were host-side data, and will require extra copying
+between host and device. |onedpl_short| will avoid copies whenever it is possible with the C++
+Standard Library implementation.  However, to guarantee no unintended copying, use
+``std::vector::data()`` in combination with ``std::vector::size()`` when using USM-allocated data
+to retrieve USM pointers for the sequence as shown.
```
akukanov, line 141: Hopefully the following makes all the same points while being somewhat more concise.

```suggestion
For ``std::vector`` with a USM allocator we recommend to use ``std::vector::data()`` in
combination with ``std::vector::size()`` as shown in the example above, rather than iterators to
``std::vector``. That is because for some implementations of the C++ Standard Library it might not
be possible for |onedpl_short| to detect that iterators are pointing to USM-allocated data. In that
case the data will be treated as if it were host-allocated, with an extra copy made to a SYCL buffer.
Retrieving USM pointers from ``std::vector`` as shown guarantees no unintended copying.
```

akukanov (commented):
@MikeDvorskiy please take a look as well.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,47 +76,66 @@ the buffer were created for the same queue. For example:
...
+    // Iterators for USM allocators might require extra copying
+    // std::fill(policy, vec.begin(), vec.end(), 42);
+
     return 0;
   }
 
-Working with SYCL buffers is recommended to reduce data copying between the host and device.
\ No newline at end of file
+Make sure that the execution policy and the USM-allocated memory were created for the same queue.
+When using USM allocators with ``std::vector``, it is recommended to use USM pointers as shown in
```
danhoeflinger, line 134:
```suggestion
Make sure that the execution policy and the USM-allocated memory were created for the same queue.

When using USM allocators with ``std::vector``, it is recommended to use USM pointers as shown in
```

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: Yes, this gets to the point, and doesn't bog the user down with unnecessary details.  Thanks, suggestion taken.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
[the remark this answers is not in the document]
danhoeflinger: Sure, done.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -4,8 +4,8 @@ Pass Data to Algorithms
 You can use one of the following ways to pass data to an algorithm executed with a device policy:
 
 * ``oneapi:dpl::begin`` and ``oneapi::dpl::end`` functions
-* Unified shared memory (USM) pointers and ``std::vector`` with USM allocators
-* Iterators of host-side ``std::vector``
+* Unified shared memory (USM) pointers
+* Iterators or USM pointers to ``std::vector`` data
```
MikeDvorskiy, line 8: I guess the phrase `Iterators or USM pointers to ``std::vector`` data` has a double meaning...
I think it is better to write it more clearly, in two bullets:

- Iterators of host-side ``std::vector``
- USM pointers obtained as ``std::vector::data()`` and ``std::vector::data() + std::vector::size()`` in case of USM allocator is passed into ``std::vector`` constructor.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,47 +76,65 @@ the buffer were created for the same queue. For example:
...
-to the host container.
-For example:
+You can also use ``std::vector`` with a USM allocator, as shown in the following example:
 
 .. code:: cpp
 
   #include <oneapi/dpl/execution>
   #include <oneapi/dpl/algorithm>
-  #include <vector>
+  #include <sycl/sycl.hpp>
```
MikeDvorskiy, line 117: I believe we don't want to have underscored lines in the code examples.
code_example

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,47 +76,65 @@ the buffer were created for the same queue. For example:
     return 0;
   }
 
-Alternatively, use ``std::vector`` with a USM allocator. For example:
+When using device USM, such as allocated by ``malloc_device``, manually copy data to this memory
```
MikeDvorskiy, line 79: It seems the sentence might lead to a false judgment - to do manually copy to device and copy back, which is not entirely true for arbitrary use case. And even incorrect  for the case above (it is not  needed to copy data to device allocated memory for `std::fill` algorithm).

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,47 +76,65 @@ the buffer were created for the same queue. For example:
...
     return 0;
   }
 
-When using device USM, such as allocated by ``malloc_device``, manually copy data to this memory
-before calling oneDPL algorithms, and copy it back once the algorithms have finished execution.
-
-Use Host-Side std::vector
------------------------------
+When using iterators to host allocated data, a temporary SYCL buffer is created, and the data
+is copied to this buffer. After processing on a device is complete, the modified data is copied
```
MikeDvorskiy, line 106: I have a couple of comments here:
1) Unfortunately, that example is not quite correct from very beginning. `std::fill` algo has just output data and doesn't  need to have in the buffer the copied data from the host, moreover, uninitialized data :)
2) Also the wording below  is not  quite correct too (I mean `the data is copied to this buffer`), because we have just the "output data" here, so that algo doesn't  need to have in the buffer the copied data from the host, moreover, uninitialized data :)

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,47 +76,65 @@ the buffer were created for the same queue. For example:
...
-    // each element of vec equals to 42
+    const int n = 1000;
+    auto policy = oneapi::dpl::execution::dpcpp_default;
+    sycl::usm_allocator<int, sycl::usm::alloc::shared> alloc(policy.queue());
+    std::vector<int, decltype(alloc)> vec(n, alloc);
+
+    // Recommended to use USM pointers:
+    std::fill(policy, vec.data(), vec.data() + vec.size(), 42);
+
+    // Iterators for USM allocators might require extra copying
```
MikeDvorskiy, line 127: 1) I would add clarification  `not recommended way`:
`// Iterators for USM allocators might require extra copying - not recommended way`
2) BTW, the overloads of  `dpl::begin`, `dpl::end` for `std::vector` (see my proposal above) could always return pointer type and encapsulate memory type RT check and create a memory buffer if is necessary.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -81,47 +76,65 @@ the buffer were created for the same queue. For example:
...
   }
 
-Working with SYCL buffers is recommended to reduce data copying between the host and device.
\ No newline at end of file
+Make sure that the execution policy and the USM-allocated memory were created for the same queue.
+
+For ``std::vector`` with a USM allocator we recommend to use ``std::vector::data()`` in
+combination with ``std::vector::size()`` as shown in the example above, rather than iterators to
+``std::vector``. That is because for some implementations of the C++ Standard Library it might not
+be possible for |onedpl_short| to detect that iterators are pointing to USM-allocated data. In that
```
MikeDvorskiy, line 138: I am not sure that we should write the any internal implementation issues in details.
Other thought on that?

MikeDvorskiy (commented):
Left many comments and ideas.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: I think that is just a rendering quirk for github and rst.  I don't think we are doing anything to have the code underlined.  Perhaps @dcbenito  can confirm.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: This is a very high level view of the issue, and accurately describes the behavior of oneDPL, I think it is reasonable to include. 
If we omit mention of the standard library limitation, it will be difficult to fully explain what is going on here.  As we have discussed previously, it is more than just extra copies because of the limitations of host iterators (for instance as source data for `permutation_iterator`).  

Here we can accurately state that in some situations, we have no choice but to treat them as host iterators, with a reasonable explanation that this is the best we can do (rather than making it look like we just have a sloppy implementation).

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: I really like the idea of adding `dpl::begin(std::vector<...>)`, `dpl::end(std::vector<...>)` overloads, and recommending that over iterators from the standard library.  However, that obviously won't happen within the scope of this PR.  

I'm also in favor of extending the `no_init` tags to our `__get_sycl_range` implementations and aligning with sycl on that (and possibly exposing it to the user as described above).

I'll address the other parts of this comment soon.

danhoeflinger: My thought with the `begin`, `end` overloads would be that we suggest to use them for ALL vectors, not only for USM allocated ones.  Since we have full access to allocator information in the vector type, we can make the correct decision for the user about how to treat their data.  With the iterators only, we aren't guaranteed to have allocator type info.
With this information at our disposal, we can copy to buffers in the case of host allocated data, and pass directly to kernels in the case of USM allocated data.  It allows us to have a single recommendation for use of oneDPL with `std::vector`, without caveats.  
I do understand though that it adds steps for users coming from normal parallel policies, and we would need to weigh that tradeoff.

Also, this is not to say we deprecate using bare iterators, just that we recommend using our `begin` and `end` functions instead for best performance.

--

As for `no_init`, you are likely correct that all the information we need is in the algorithm semantics, and this is not necessary to add to the public API.  There is a related change I was thinking of with regard to our internal implementation where our input data "processing" for the dpcpp backend does not align with the SYCL interface for buffer modes & properties exactly, so our internal interface is less expressive than the SYCL interface.  [Here is an example](https://github.com/oneapi-src/oneDPL/blob/d442b291c86b1d44539f2e472bb96a4067ef6e82/include/oneapi/dpl/pstl/hetero/histogram_impl_hetero.h#L133-L136) where we don't have the expressivity that we want.  I

danhoeflinger: If we take `no_init` out of it for now, my initial thought would be to implement them as follows:
`dpl::begin(vec)`:
check the allocator type and either `vec.data()` or `vec.begin()` depending on USM allocator or not.

`dpl::end(vec)`:
check the allocator type and either `vec.data() + vec.size()` or `vec.end()` depending on USM allocator or not.

The rest can be left to our existing implementation.

akukanov: > My thought with the begin, end overloads would be that we suggest to use them for ALL vectors, not only for USM allocated ones. ... Also, this is not to say we deprecate using bare iterators, just that we recommend using our begin and end functions instead for best performance.

I see, thanks for explaining. That makes more sense but needs to be further elaborated. Let's discuss separately from this patch.

danhoeflinger: Agreed, also @MikeDvorskiy will have to weigh in if his intentions are different, I hijacked his comment.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
dcbenito: I ran a test build of the documentation and there are no underlines there. I am confirming this is a rendering quirk.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: fixed (1).

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: Applied the rewrite I mention above.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: Thanks, I've applied this change.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: > I'll address the other parts of this comment soon.

The other part, I think, is that the example that uses only `std::fill` is not sufficient to illustrate the aspects of host-side data handling explained in the text. So, I guess, a different example is desirable.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: > As we have discussed previously, it is more than just extra copies because of the limitations of host iterators (for instance as source data for permutation_iterator).

I have not thought of the `permutation_iterator` problems when I coined the proposed wording. Dan, do you think it's sufficient for that case as well?

danhoeflinger: I think without a lot of details which I don't think we want in this setting,  I don't think this will be clear to users directly.  

I think the important thing is that we recommend using USM pointers rather than iterators, and have some mention that they are treated as if they are host allocated iterators.  

With that, if users have trouble, we can point them to some additional information / known limitations to connect the dots to explain in more detail why they experience the issue they are experiencing and suggest they use USM pointers instead.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: I think we can replace `fill` with `generate`, and then add `sort`.

danhoeflinger: This works for me.  However, for the example with host-side iterators, I would prefer to call generate with a `par_unseq` policy.  I don't want to encourage the use of host iterators as input to multiple APIs successively.  This is a bad pattern of usage which results in lots of data transfers to and from the device.  
We can have a comment to explain why, and it may be useful to users to show that with USM allocated data they can get performance gains by not thrashing data transfers.  I'll try to put this together.

danhoeflinger: I've made the change to use `generate` and `sort`.  I've just generated the data sequentially on the host, because to do it efficiently on the device (or even with `par[_unseq]` would take more complicated code than should live in an example which is meant to show how to pass data, and I don't think it makes sense to write example code device code we wouldn't recommend people use.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -44,10 +44,19 @@ To use the functions, add ``#include <oneapi/dpl/iterator>`` to your code. For e
   #include <oneapi/dpl/iterator>
   #include <sycl/sycl.hpp>
   int main(){
-    sycl::buffer<int> buf { 1000 };
+    std::vector<int> vec(1000);
+
+    std::random_device r;
+    std::default_random_engine e1(r());
+    std::uniform_int_distribution<int> uniform_dist(-2000, 2000);
+    std::generate(vec.begin(), vec.end(), [&](){ return uniform_dist(e1); });
```
akukanov, line 52: I think this is quite overcomplicated for the examples. I would simplify it to just
```suggestion
    std::generate(vec.begin(), vec.end(), std::minstd_rand{});
```
Also `<random>` should be included.

akukanov (commented):
Looks fine to me, though I would simplify the generation part of the examples - see below.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: Thanks, done.

akukanov (dismissed):

danhoeflinger:
@dcbenito When you get a chance, could you take another look at this now that we have solidified changes to the content?

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -56,72 +62,91 @@ To use the functions, add ``#include <oneapi/dpl/iterator>`` to your code. For e
...
-    std::fill(oneapi::dpl::execution::make_device_policy(q), d_head, d_head + n, 42);
+    std::sort(oneapi::dpl::execution::make_device_policy(q), d_head, d_head + n);
 
     sycl::free(d_head, q);
     return 0;
   }
 
-Alternatively, use ``std::vector`` with a USM allocator. For example:
+When using device USM, such as allocated by malloc_device, the user is responsible for data
+transfers to and from the device to ensure that input data is device accessible during oneDPL
```
dcbenito, line 88: My suggestion:

When using device USM, such as allocated by ``malloc_device``, you are responsible for data
transfers to and from the device to ensure that the input data is device accessible during oneDPL
algorithm execution and that the result is available to the operation.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -56,72 +62,91 @@ To use the functions, add ``#include <oneapi/dpl/iterator>`` to your code. For e
...
+algorithm execution and that the result is available to the operation which follows.
+
+Use std::vector
+-----------------------------
+
+The following examples demonstrate two ways to use the parallel algorithms with ``std::vector``:
+
+* host allocators
```
dcbenito, line 96: Please capitalize host.

dcbenito (dismissed):
Minor issues to fix, otherwise LGTM.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: Taken, but pluralizing "operations" at the end.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: done, thanks.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
dcbenito: Looks good! Should malloc_device be in monospace font? It is not in the original text.

danhoeflinger: good catch, fixed.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -84,16 +84,16 @@ the USM-allocated memory were created for the same queue. For example:
     return 0;
   }
 
-When using device USM, such as allocated by malloc_device, the user is responsible for data
+When using device USM, such as allocated by ``malloc_device``, you are responsible for data
 transfers to and from the device to ensure that input data is device accessible during oneDPL
-algorithm execution and that the result is available to the operation which follows.
+algorithm execution and that the result is available to the operations.
```
akukanov, line 89:
```suggestion
algorithm execution and that the result is available to the subsequent operations.
```

akukanov (approved):

PR #1481, opened 2024-04-05
Enable binary comparison operations for internal tuple (second approach)

danhoeflinger:
This takes an alternative approach to #1472 to enable binary comparison using SFINAE.  

It only needs a single overload per operator, as it uses SFINAE create the overloads to match the incoming types rather than partial specialization. 

In this version we are able to use forwarding (universal references) for the inputs, forwarding into the `static_cast`.
Here we `static_cast` all inputs (which may be `std::tuple` or `oneapi::dpl::__internal::tuple`) to `std::tuple` with the same element types.  I believe this cast should be converted to a noop at compile time in the case where the `static_cast` is the existing type of the input.

### DIFF
--- a/include/oneapi/dpl/pstl/tuple_impl.h
+++ b/include/oneapi/dpl/pstl/tuple_impl.h
@@ -485,6 +485,21 @@ using __is_tuple_type = std::negation<std::is_void<__get_std_tuple_t<T>>>;
 template <typename T>
 using __is_std_tuple_type = std::is_same<__get_std_tuple_t<T>, std::decay_t<T>>;
 
+//TODO: fix forwarding, this isn't quite right to minimize copies
+template <typename T, std::enable_if_t<!__is_std_tuple_type<T>::value, int> = 0>
+__get_std_tuple_t<T>
+__to_std_type(T&& tup)
+{
+    return static_cast<__get_std_tuple_t<T>>(std::forward<T>(tup));
+}
+
+template <typename T, std::enable_if_t<__is_std_tuple_type<T>::value, int> = 0>
+T&&
+__to_std_type(T&& tup)
+{
+    return std::forward<T>(tup);
+}
+
 // Require that both are tuple types, but at least one type is a oneDPL internal tuple, and that the number of tuple
 // elements are the same
 template <typename Tuple1, typename Tuple2>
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -487,38 +456,100 @@ struct tuple<>
...
+// These binary comparison overloads required because the template arguments require exact matches to participate in
+// overload resolution. Implicit conversion to std::tuple does not provide these for free. The static cast should be
+// without performance penalty if the casting to its own type.
+template <typename Tuple1, typename Tuple2,
+          typename = std::enable_if_t<__comparable_tuple_types<Tuple1, Tuple2>::value>>
+constexpr bool
+operator==(Tuple1&& __lhs, Tuple2&& __rhs)
+{
+    return static_cast<__get_std_tuple_t<Tuple1>>(std::forward<Tuple1>(__lhs)) ==
+           static_cast<__get_std_tuple_t<Tuple2>>(std::forward<Tuple2>(__rhs));
```
adamfidel, line 505: I double checked on [Godbolt](https://godbolt.org/z/ccc3n7q8s) and this does not introduce a copy for the cast of a `std::tuple` to a `std::tuple`. I don't have anything to add; I just wanted to verify that there wasn't an unnecessary copy being introduced.

danhoeflinger: I'm still working it out.  I think it may inject an extra copy for lvalue refs... still a work in progress.

SergeyKopienko:
@danhoeflinger you wrote this type conversion operator:
```C++
    operator tuple_type() const
    {
        return tuple_type{};
    }
```
Looks like a little bit strange... Why you return new instance?

SergeyKopienko:
From my pointy of view, when we call
```C++
    template <typename U1, typename... U, ::std::size_t... _Ip>
    static ::std::tuple<U1, U...>
    to_std_tuple(const oneapi::dpl::__internal::tuple<U1, U...>& __t, ::std::index_sequence<_Ip...>)
    {
        return ::std::tuple<U1, U...>(oneapi::dpl::__internal::get_impl<_Ip>()(__t)...);
    }
```
from `operator==(Tuple1&& __lhs, Tuple2&& __rhs)` we will copy data. Am I correct?

danhoeflinger:
> From my pointy of view, when we call
> 
> ```c++
>     template <typename U1, typename... U, ::std::size_t... _Ip>
>     static ::std::tuple<U1, U...>
>     to_std_tuple(const oneapi::dpl::__internal::tuple<U1, U...>& __t, ::std::index_sequence<_Ip...>)
>     {
>         return ::std::tuple<U1, U...>(oneapi::dpl::__internal::get_impl<_Ip>()(__t)...);
>     }
> ```
> 
> from `operator==(Tuple1&& __lhs, Tuple2&& __rhs)` we will copy data. Am I correct? If so, this approach probably should have not very good performance...

Yes, it seems I've missed a big issue with this PR.  I was trying to avoid re-implementing the functionality of `std::tuple` but now I think there is no good way to take advantage of their implementation without copies.  

I guess the appropriate thing to do is to just reimplement the functionality of the operators here...  then no casting is necessary.

danhoeflinger:
Closing this in light of the issue discussed above in favor of changes in the main branch.

PR #1483, opened 2024-04-08
Fix linker error LNK2005 for tag dispatching staff

SergeyKopienko:
In this PR we fix linker error when we have more then one translation unit for one project:
```C++
error LNK2005: "class oneapi::dpl::execution::v1::parallel_policy const & __cdecl oneapi::dpl::get_unvectorized_policy(class oneapi::dpl::execution::v1::parallel_unsequenced_policy const &)" (?get_unvectorized_policy@dpl@oneapi@@YAAEBVparallel_policy@v1@execution@12@AEBVparallel_unsequenced_policy@4512@@Z) already defined in hello-1-5be217.obj
```

The problem is that our functions `get_unvectorized_policy` are not `inline`.

### DIFF
--- a/include/oneapi/dpl/pstl/glue_memory_impl.h
+++ b/include/oneapi/dpl/pstl/glue_memory_impl.h
@@ -199,7 +199,7 @@ get_unvectorized_policy(const oneapi::dpl::execution::unsequenced_policy&)
     return oneapi::dpl::execution::seq;
 }
 
-inline template <typename _ExecutionPolicy>
+template <typename _ExecutionPolicy>
 const _ExecutionPolicy&
 get_unvectorized_policy(const _ExecutionPolicy& __exec)
 {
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/glue_memory_impl.h
```diff
@@ -187,13 +187,13 @@ uninitialized_fill_n(_ExecutionPolicy&& __exec, _ForwardIterator __first, _Size
 
 #if (_PSTL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN || _ONEDPL_ICPX_OMP_SIMD_DESTROY_WINDOWS_BROKEN)
 
-const oneapi::dpl::execution::parallel_policy&
+inline const oneapi::dpl::execution::parallel_policy&
```
rarutyun, line 190: I hope we don't have a similar situation with `__select_backend`, do we?

SergeyKopienko: The `__select_backend` functions is template functions so there are no problems for them.

Alexandr-Konovalov (approved):
LGTM for the fix, but I don't now current policy wrt the regression testing.

include/oneapi/dpl/pstl/glue_memory_impl.h
rarutyun: That's what I was asking for. Thanks!

rarutyun (approved):

PR #1485, opened 2024-04-08
Merge of distributed-ranges into oneDPL - namespaces and cmake files. 

mateuszpn:


lslusarczyk:
Thank you @mateuszpn . You did a lot of work beyond namespace (removal of matrices, source_location). Great!

In case of namespaces, what I see is that onedpl puts things in oneapi::dpl::experimental
e.g. here:
`documentation/library_guide/dynamic_selection_api/dynamic_load_policy.rst:  namespace ex = oneapi::dpl::experimental;`

while in distrubuted ranges "experimental" namespace is top level namespace

Could you please add "oneapi::dpl::" before "experimental"?

PR #1487, opened 2024-04-09
Handle potential sycl::malloc issues

julianmi:
The SYCL spec on `sycl::malloc` reads: " If there are not enough resources to allocate the requested memory, these functions return nullptr." This patch handles the null pointer case by throwing `std::bad_alloc`.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -424,7 +424,11 @@ struct __sycl_usm_alloc
     operator()(::std::size_t __elements) const
     {
         const auto& __queue = __exec.queue();
-        return (_T*)sycl::malloc(sizeof(_T) * __elements, __queue.get_device(), __queue.get_context(), __alloc_t);
+        auto __buf = static_cast<_T*>(
+            sycl::malloc(sizeof(_T) * __elements, __queue.get_device(), __queue.get_context(), __alloc_t));
+        if (!__buf)
+            throw std::bad_alloc();
+        return __buf;
     }
 };
 


dmitriy-sobolev (approved):
LGTM.

It is aligned with the C++ [spec](https://eel.is/c++draft/algorithms.parallel.exceptions#1.sentence-1):
> During the execution of a parallel algorithm, if temporary memory resources are required for parallelization and none are available, the algorithm throws a bad_alloc exception[.]

@MikeDvorskiy, could you also check the PR?

MikeDvorskiy (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -424,7 +424,11 @@ struct __sycl_usm_alloc
     operator()(::std::size_t __elements) const
     {
         const auto& __queue = __exec.queue();
-        return (_T*)sycl::malloc(sizeof(_T) * __elements, __queue.get_device(), __queue.get_context(), __alloc_t);
+        auto __buf = static_cast<_T*>(
```
SergeyKopienko, line 427: What if we a little bit rewrite this?

```C++
    _T*
    operator()(::std::size_t __elements) const
    {
        if (auto __buf = static_cast<_T*>(
                sycl::malloc(sizeof(_T) * __elements, __exec.queue().get_device(), __queue.get_context(), __alloc_t)))
            return __buf;

        throw std::bad_alloc();
    }
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -424,7 +424,11 @@ struct __sycl_usm_alloc
...
+            sycl::malloc(sizeof(_T) * __elements, __queue.get_device(), __queue.get_context(), __alloc_t));
+        if (!__buf)
+            throw std::bad_alloc();
```
SergeyKopienko, line 430: Should we include `<new>` header file ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: This is mostly a preference of style, isn't it? Or do you feel strongly about this needing to be changed post merging?

SergeyKopienko: Not really, I think: most preferable to handle under `if (true)`  most frequently case. That's what I mean.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: > Should we include `<new>` header file ?

This should be included by `utils.h`.

> And the second question: if we are unable to allocate memory, should we use `sycl::buffer` instead?

I think we should instead check if USM is supported before using this routine. I think the only place where we currently use it is for reduce with heterogeneous backends and there we check it. This patch is mostly to handle out of resource errors.

SergeyKopienko: > I think we should instead check if USM is supported before using this routine. I think the only place where we currently use it is for reduce with heterogeneous backends and there we check it. This patch is mostly to handle out of resource errors.

OK.

PR #1491, opened 2024-04-10
Common implementation of hosts backends buffer

SergeyKopienko:
The goal of this PR - to reduce the amount of duplicated code in our host backends implementations.

In this PR we implement and use the common implementation of host backends buffer, specialized by different memory allocators.
Previously we had three different implementations of hosts backends buffer: they were absolutely the same but have used different memory allocators.

### DIFF
--- a/include/oneapi/dpl/pstl/omp/util.h
+++ b/include/oneapi/dpl/pstl/omp/util.h
@@ -58,33 +58,7 @@ __cancel_execution(oneapi::dpl::__internal::__omp_backend_tag)
 //------------------------------------------------------------------------
 
 template <typename _ExecutionPolicy, typename _Tp>
-class __buffer_impl
-{
-    std::allocator<_Tp> __allocator_;
-    _Tp* __ptr_;
-    const std::size_t __buf_size_;
-    __buffer_impl(const __buffer_impl&) = delete;
-    void
-    operator=(const __buffer_impl&) = delete;
-
-  public:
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -58,33 +58,7 @@ __cancel_execution(oneapi::dpl::__internal::__omp_backend_tag)
...
-    get() const
-    {
-        return __ptr_;
-    }
-    ~__buffer_impl() { __allocator_.deallocate(__ptr_, __buf_size_); }
-};
-
-template <typename _ExecutionPolicy, typename _Tp>
-using __buffer = __buffer_impl<::std::decay_t<_ExecutionPolicy>, _Tp>;
+using __buffer = oneapi::dpl::__utils::__buffer_impl<::std::decay_t<_ExecutionPolicy>, _Tp, std::allocator>;
```
julianmi, line 61: Nit: should we unify `std` vs. `::std` here and in the following changes?

julianmi (commented):
LGTM

include/oneapi/dpl/pstl/omp/util.h
SergeyKopienko: Will fix, thanks.

SergeyKopienko: Done.

julianmi (approved):

MikeDvorskiy (approved):
LGTM

PR #1496, opened 2024-04-11
Remove using namespace directives in header

adamfidel:
PR #1394 introduced a few `using namespace` directives in a header, which would pull everything from our internal namespaces to be visible in user code.

This PR removes the `using` directives by explicitly writing the namespaces and wrapping the forward declarations in their appropriate namespace.

### DIFF
--- a/include/oneapi/dpl/internal/binary_search_extension_defs.h
+++ b/include/oneapi/dpl/internal/binary_search_extension_defs.h
@@ -25,6 +25,17 @@ namespace oneapi
 {
 namespace dpl
 {
+namespace internal
+{
+
+enum class search_algorithm
+{
+    lower_bound,
+    upper_bound,
+    binary_search
+};
+
+} // namespace internal
 
 template <typename Policy, typename InputIterator1, typename InputIterator2, typename OutputIterator>
 oneapi::dpl::__internal::__enable_if_execution_policy<Policy, OutputIterator>
--- a/include/oneapi/dpl/internal/binary_search_extension_defs.h
+++ b/include/oneapi/dpl/internal/binary_search_extension_defs.h
@@ -25,17 +25,6 @@ namespace oneapi
 {
 namespace dpl
 {
-namespace internal
-{
-
-enum class search_algorithm
-{
-    lower_bound,
-    upper_bound,
-    binary_search
-};
-
-} // namespace internal
 
 template <typename Policy, typename InputIterator1, typename InputIterator2, typename OutputIterator>
 oneapi::dpl::__internal::__enable_if_execution_policy<Policy, OutputIterator>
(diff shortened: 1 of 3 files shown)


danhoeflinger (approved):
LGTM.  Thanks for putting this PR up.
@MikeDvorskiy  should probably weigh in before merging.

include/oneapi/dpl/internal/binary_search_extension_defs.h
```diff
@@ -25,6 +25,17 @@ namespace oneapi
 {
 namespace dpl
 {
+namespace internal
+{
+
+enum class search_algorithm
```
SergeyKopienko, line 31: Does it really needed to have only one brick `struct custom_brick` + user visible `enum class search_algorithm` + parametrizations by their enum values? What about three bricks without  `enum class search_algorithm` at all?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -24,6 +24,8 @@
 #ifndef _ONEDPL_SYCL_TRAITS_H
 #define _ONEDPL_SYCL_TRAITS_H
 
+#include <oneapi/dpl/internal/binary_search_extension_defs.h>
```
MikeDvorskiy, line 27: I believe it should be a relative patch, via ""

include/oneapi/dpl/internal/binary_search_extension_defs.h
MikeDvorskiy, line 31: Why only a forward declaration is not enough?
Like this?
`enum class search_algorithm;`

SergeyKopienko: I am not sure that this `enum` required at all: see https://github.com/oneapi-src/oneDPL/pull/1496#discussion_r1562193572

adamfidel: I did not realize forward declaring enum classes were possible. Thanks for pointing that out. I have removed the changes to the binary search files and just forward declared the enum class.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
adamfidel: I have removed this include.

include/oneapi/dpl/internal/binary_search_extension_defs.h
adamfidel: I'm sorry, but I don't quite follow the question. I have updated the forward declaration of `custom_brick` to now use a forward declared `search_algorithm` enum class. Are you suggesting a different approach?

SergeyKopienko: OK: forward declaration is enough too.
My idea was to split this brick into three separate bricks and remove this `enum`.
But I am fine with forward declaration too.

SergeyKopienko (approved):
LGTM

PR #1502, opened 2024-04-15
Fix incorrect _ONEDPL_SYCL_USM_HOST_PRESENT usage in USM allocation helpers

mmichel11:
`__use_USM_allocations` introduced in PR #1354 uses `_ONEDPL_SYCL_USM_HOST_PRESENT` to check if host memory support is present. However, this macro was renamed to `_ONEDPL_SYCL_UNIFIED_USM_BUFFER_PRESENT` in PR #1410, so the function always returns false. This causes incompatible scratch and result buffer paths to be taken in the code causing incorrect results and hangs.

This PR switches to use the correct macro here.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -567,7 +567,6 @@ struct __result_and_scratch_storage
         }
         else if (__supports_USM_device)
         {
-            std::cout << "device" << std::endl;
             // If we don't use host memory, malloc only a single unified device allocation
             __scratch_buf = ::std::shared_ptr<_T>(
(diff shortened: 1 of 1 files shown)


julianmi:
@mmichel11 Thanks for raising this. I noticed several issues as well and am working on fixes for them. Namely:

1. Outdated macro as you suggested.
2. Empty scratch space allocation is unguarded when using USM host memory. This leads to runtime issues or `bad_alloc`.
3. The guards to support older compilers (#1410) are missing.
4. `__get_usm_host_or_buffer_accessor_ptr` is not needed anymore.

I have a WIP branch here `dev/julianmi/fix_reduce_1354` that will eventually address all four issues. So I think we can close this one.

SergeyKopienko (approved):
LGTM

julianmi:
I believe #1503 supersedes this PR by including this macro fix and several other fixes.

PR #1506, opened 2024-04-16
[oneDPL][hetero] A blocking behavior fix for some algo patterns (sync API)

MikeDvorskiy:
[oneDPL][hetero] + a blocking behavior fix for some algo patterns (sync API)

According to 4.7.2.3. Buffer synchronization rules, rule number (1), a destructor of such buffer does not need to block. So, RAII instance of such buffer doesn't guarantee implicit wait. The PR fixes the behavior for oneDPL sync API.

(https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:buf-sync-rules)

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1031,8 +1031,8 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, 
     //TODO: To optimize copy back depending on Iterator, i.e. set_final_data for host iterator/pointer
     // __pattern_copy_if above may be async due to there is implicit synchronization on sycl::buffer and the accessors
 
-    //An explicit wait isn't required here because we have implicit synchronization on sycl::buffer destructor.
-    return __pattern_walk2</*_IsSync=*/std::false_type>(
+    //A buffer is constructed from a range, the destructor does not need to block.
+    return __pattern_walk2(
         __tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper>(::std::forward<_ExecutionPolicy>(__exec)),
         __copy_first, __copy_last, __first, __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
 }
@@ -1053,8 +1053,8 @@ __pattern_unique(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _It
 
     //TODO: optimize copy back depending on Iterator, i.e. set_final_data for host iterator/pointer
 
-    //An explicit wait isn't required here because we have implicit synchronization on sycl::buffer destructor.
-    return __pattern_walk2</*_IsSync=*/std::false_type, __par_backend_hetero::access_mode::read_write,
+    //A buffer is constructed from a range, the destructor does not need to block.
+    return __pattern_walk2</*_IsSync=*/std::true_type, __par_backend_hetero::access_mode::read_write,
                            __par_backend_hetero::access_mode::read_write>(
         __tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper>(::std::forward<_ExecutionPolicy>(__exec)),
         __copy_first, __copy_last, __first, __brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
@@ -1236,8 +1236,8 @@ __pattern_inplace_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __ex
 
     //TODO: optimize copy back depending on Iterator, i.e. set_final_data for host iterator/pointer
 
-    //An explicit wait isn't required here because we have implicit synchronization on sycl::buffer destructor.
-    __pattern_walk2</*_IsSync=*/std::false_type>(
+    //A buffer is constructed from a range, the destructor does not need to block.
+    __pattern_walk2(
         __tag, __par_backend_hetero::make_wrapped_policy<copy_back_wrapper>(::std::forward<_ExecutionPolicy>(__exec)),
         __copy_first, __copy_last, __first, __brick_move<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
 }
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1031,8 +1031,8 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec,
     //TODO: To optimize copy back depending on Iterator, i.e. set_final_data for host iterator/pointer
     // __pattern_copy_if above may be async due to there is implicit synchronization on sycl::buffer and the accessors
 
-    //An explicit wait isn't required here because we have implicit synchronization on sycl::buffer destructor.
-    return __pattern_walk2</*_IsSync=*/std::false_type>(
+    //A buffer is constructed from a range, the destructor does not need to block.
```
danhoeflinger, line 1034: I don't think any comment is required here anymore.  If we want some comment here, perhaps: 

//__pattern_walk2 is a blocking call

Similar comment for the repetitions of this later.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1662,9 +1662,9 @@ __pattern_rotate(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator
     auto __temp_rng_rw =
         oneapi::dpl::__ranges::all_view<_Tp, __par_backend_hetero::access_mode::read_write>(__temp_buf.get_buffer());
     oneapi::dpl::__par_backend_hetero::__parallel_for(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec), __brick,
-                                                      __n, __temp_rng_rw, __buf.all_view());
+                                                      __n, __temp_rng_rw, __buf.all_view()).wait();
 
-    //An explicit wait doesn't need here because we have implicit synchronization (and wait) on sycl::buffer destructor.
+    //A buffer is constructed from a range, the destructor does not need to block.
```
danhoeflinger, line 1667: I dont think a comment is necessary here.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1662,9 +1662,9 @@ __pattern_rotate(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator
     auto __temp_rng_rw =
         oneapi::dpl::__ranges::all_view<_Tp, __par_backend_hetero::access_mode::read_write>(__temp_buf.get_buffer());
     oneapi::dpl::__par_backend_hetero::__parallel_for(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec), __brick,
-                                                      __n, __temp_rng_rw, __buf.all_view());
+                                                      __n, __temp_rng_rw, __buf.all_view()).wait();
```
danhoeflinger, line 1665: I believe I agree with the clang-format suggestion to move `.wait()` to the next line, but it is your call.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1328,12 +1328,12 @@ __pattern_stable_partition(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& _
```
danhoeflinger, line 1328: Do we have any guarantee that the first `__pattern_walk2()` call above returns?
I think these are independent from each other, and were both depending on the destructors for synchronization (or so we thought...).

I think it may need to be made synchronous as well.  It would be nice to have it return an event which could be waited on so that these functions could run concurrently, but that is likely beyond the scope of this PR.

SergeyKopienko (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
MikeDvorskiy: 1) These patterns (pattern_merge and pattern_walk2) have  an implicit dependence via SYCL dependency graph.
2) In other pattern (__pattern_stable_partition) we have dependent working patterns, and those can be launched asynchrony and we have to wait two events by `wait(e1, e2)`. There is a comment about that in the code.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
MikeDvorskiy: I think to keep such comment is very important , because the situation (with misunderstanding what is going on)  might be repeated again by looking at the someone? not me, you or Sergey.. And that person also may decide that __pattern_walk2 may be called in async way "because" we have sycl::buffer above on the stack.

danhoeflinger: Fair point, perhaps we just need to expand it a little more then:
```
// The temporary buffer is constructed from a range, therefore it's destructor will not block, therefore 
// we must call __pattern_walk2 in a way which provides blocking synchronization for this pattern.
```

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: For `__pattern_stable_partition`, line 1325, we call `__pattern_walk2` without synchronization.  It has no dependencies with the blocking call on line 1331 as far as I can see.  

Therefore, we have no guarantee that the call from line 1325 completes before we return from `__pattern_stable_partition`.  Ideally `__pattern_walk2` would return a future or event we could wait on, but it does not, so I think we just need to call it in a blocking fashion, unless we want more refactoring.

I'm missing where you mention we wait on 2 events as you describe: `wait(e1, e2)`.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: A similar adjustment to the comment as mentioned above would work instead.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: yes, I agree with that.

danhoeflinger (dismissed):
I'd recommend fixing the clang-format issues (there are some for comments and others for code).  Otherwise this LGTM.

Also lets wait for green CI.

MikeDvorskiy:
> I'd recommend fixing the clang-format issues (there are some for comments and others for code). Otherwise this LGTM.
> 
> Also lets wait for green CI.

I've just applied of the most of clang format recommendations, excepting one case with ".wait()"

danhoeflinger (approved):
LGTM, wait for green CI to merge please.

PR #1511, opened 2024-04-17
Avoid extra constructor and destructor calls in oneDPL code

SergeyKopienko:
The goal of this PR - to avoid extra constructor and destructor calls in oneDPL code.
For that we using the next additional compile-time checks:
- [std::is_trivial_v](https://en.cppreference.com/w/cpp/types/is_trivial) check to avoid extra **constructor** calls;
- [std::is_trivially_destructible_v](https://en.cppreference.com/w/cpp/types/is_destructible) check to avoid extra **destructor** calls.

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -1497,7 +1497,7 @@ __remove_elements(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
                 __internal::__brick_copy_by_mask(
                     __first + __i, __first + __i + __len, __result + __initial, __mask + __i,
                     [](_RandomAccessIterator __x, _Tp* __z) {
-                        if constexpr (std::is_trivially_move_constructible_v<_Tp>)
+                        if constexpr (std::is_trivial_v<_Tp>)
                             *__z = std::move(*__x);
                         else
                             ::new (std::addressof(*__z)) _Tp(std::move(*__x));
@@ -2605,7 +2605,7 @@ __pattern_partial_sort_copy(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec
                     // 1. Copy elements from input to raw memory
                     for (_T1* __k = __i; __k != __j; ++__k, ++__it)
                     {
-                        if constexpr (std::is_trivially_copy_constructible_v<_T2>)
+                        if constexpr (std::is_trivial_v<_T2>)
                             *__k = *__it;
                         else
                             ::new (__k) _T2(*__it);
@@ -3097,7 +3097,7 @@ __pattern_inplace_merge(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _R
     _Tp* __r = __buf.get();
     __internal::__except_handler([&]() {
         auto __move_values = [](_RandomAccessIterator __x, _Tp* __z) {
-            if constexpr (std::is_trivially_move_constructible_v<_Tp>)
+            if constexpr (std::is_trivial_v<_Tp>)
                 *__z = std::move(*__x);
             else
                 ::new (std::addressof(*__z)) _Tp(std::move(*__x));
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -1497,10 +1497,10 @@ __remove_elements(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
                 __internal::__brick_copy_by_mask(
                     __first + __i, __first + __i + __len, __result + __initial, __mask + __i,
                     [](_RandomAccessIterator __x, _Tp* __z) {
-                        if constexpr (::std::is_trivial_v<_Tp>)
-                            *__z = ::std::move(*__x);
+                        if constexpr (std::is_trivially_move_constructible_v<_Tp>)
+                            *__z = std::move(*__x);
                         else
-                            ::new (::std::addressof(*__z)) _Tp(::std::move(*__x));
+                            ::new (std::addressof(*__z)) _Tp(std::move(*__x));
                     },
                     _IsVector{});
             },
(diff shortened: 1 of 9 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -231,7 +234,10 @@ struct transform_reduce
         {
             const _Size __items_to_process = __adjusted_n - __adjusted_global_id;
             // Keep these statements in the same scope to allow for better memory alignment
-            new (&__res.__v) _Tp(std::move(__unary_op(__adjusted_global_id, __acc...)));
```
SergeyKopienko, line 234: I believe that `std::move` call isn't required in this and the same places in this file because the result of `__unary_op` call already is [pr-value](https://en.cppreference.com/w/cpp/language/value_category).

julianmi: I do. Thanks for catching this.

SergeyKopienko: I have fixed this point here.

julianmi (commented):
I think these changes are beneficial and LGTM in general. Does it make sense to introduce helpers to minimize the code duplication?

SergeyKopienko:
> I think these changes are beneficial and LGTM in general. Does it make sense to introduce helpers to minimize the code duplication?

I am not sure should we have this helper or no. At least it's easy to review without it.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -2603,7 +2605,10 @@ __pattern_partial_sort_copy(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec
                     // 1. Copy elements from input to raw memory
                     for (_T1* __k = __i; __k != __j; ++__k, ++__it)
                     {
-                        ::new (__k) _T2(*__it);
+                        if constexpr (std::is_trivially_copy_constructible_v<_T2>)
+                            *__k = *__it;
+                        else
+                            ::new (__k) _T2(*__it);
```
kboyarinov, line 2611: The value of this change is not clear to me. As far as I understand, if the type is trivially copyable, it can be copied bitwise (e.g. with `std::memcpy`) instead of calling the copy constructor directly. In this particular case, we just replace the copy constructor call with the copy assignment call. May be I simply don't understand why it is better. Could you please explain in details if you find it useful?

SergeyKopienko: As I described at this PR, I simple trying to avoid constructor/destructor calls when it's not required.
In this case we simple make copy assignment call without constructor call - I hope it's should work faster then constructor call.

kboyarinov: It seems like this change introduced more problems then benefits. According to the standard, the class with deleted copy assignment operator still considered trivially copyable. Consider this example: https://godbolt.org/z/ofPYaEMW9.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -106,7 +106,12 @@ struct __subgroup_radix_sort
         {
             const uint16_t __idx = __wi * __block_size + __i;
             if (__idx < __n)
-                new (&__values[__i]) _ValueT(::std::move(__src[__idx]));
+            {
+                if constexpr (std::is_trivially_move_constructible_v<_ValueT>)
+                    __values[__i] = std::move(__src[__idx]);
```
rarutyun, line 111: This check conceptually says that if something is `trivially_move_constructible`, call move assignment, while there is no guarantee that the move assignment is available. This is like check `foo()` for availability and then call `bar()`, which is nonsense

SergeyKopienko: Fixed

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
kboyarinov: I believe the comment raised by Ruslan and myself is not about the type of check made.
The question is why should we call move assignment where move constructor is expected. The only thing that is guaranteed if the type is `trivially_move_constructible` is that we can do a bitwise move (e.g. `std::memmove`) instead of calling the move constructor. Move assignment is not guaranteed to be presented even if the type is trivial (not only trivially movable.
And it is also should be checked if using bitwise move instead of directly calling move constructor is beneficial.

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: We already have this approach in our code: https://github.com/oneapi-src/oneDPL/blob/8b9a2dd9a347c0255e4759fa93f6738e5d103340/include/oneapi/dpl/pstl/algorithm_impl.h#L1498

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
SergeyKopienko: We already have this approach in our code: https://github.com/oneapi-src/oneDPL/blob/8b9a2dd9a347c0255e4759fa93f6738e5d103340/include/oneapi/dpl/pstl/algorithm_impl.h#L1498

kboyarinov: I understand, and at the same time it seems like we have issues caused by this approach:
https://godbolt.org/z/6Pad9M8bK.
I guess we need to discuss it within the team and either do it correctly in all places or remove this unclear optimizations.

SergeyKopienko:
I f we need some common copy / move helper, separate PR https://github.com/oneapi-src/oneDPL/pull/1514 prepared.

PR #1495, opened 2024-04-10
Call constructor of reduce storage

julianmi:
@SergeyKopienko and @MikeDvorskiy pointed out that the changes in #1365 lack the constructor and destructor call to `_Tp`.

PR #1512, opened 2024-04-17
Remove unnecessary move statements added in #1495

julianmi:
#1495 added move statements which are not needed since the result of `__unary_op` is already an [prvalue](https://en.cppreference.com/w/cpp/language/value_category#prvalue).

See original comment by @SergeyKopienko: https://github.com/oneapi-src/oneDPL/pull/1511#discussion_r1569022127


SergeyKopienko (approved):
LGTM

MikeDvorskiy (approved):
LGTM

PR #1513, opened 2024-04-18
[oneDPL][hetero] Re-woring sycl::is_device_copyable trait specializations for oneDPL types

MikeDvorskiy:
[oneDPL][hetero] Re-woring sycl::is_device_copyable trait specializations for oneDPL types.

1) After offline discussed internally in oneDPL team, we decide to keep  sycl::is_device_copyable trait specialization next to each oneDPL type definition.
2) Also we have to support the previous  oneAPI compiler version via special macro
3) Also we have to support a case when SYCL backend is not active.

This is a prototype for 3 different oneDPL types: `transform_reduce`, `__brick_copy_n` and `zip_iterator`
(The total number of such  oneDPL types is about 50.)
Also the prototype introduces a common code stuff to define the is_device_copyable trait specializations.

Based on GoDBolt prototype: https://godbolt.org/z/qx8nTaKsK

UPDATE!!!
The prototype was modified: via a special trait in each device copyable oneDPL type

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -1066,6 +1066,8 @@ struct __brick_copy_n<_Tag, _ExecutionPolicy,
     {
         return ::std::copy_n(__first, __n, __result);
     }
+
+    _ONEDPL_IS_DEVICE_COPYABLE()
 };
 
 //------------------------------------------------------------------------
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -375,6 +375,8 @@ struct __brick_copy_n<__hetero_tag<_BackendTag>, _ExecutionPolicy>
     {
         __target = ::std::forward<_SourceT>(__source);
     }
+
+    _ONEDPL_IS_DEVICE_COPYABLE()
 };
 
 template <typename _BackendTag, typename _ExecutionPolicy>
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -48,6 +48,30 @@ namespace dpl
 namespace __internal
 {
 
+#if _ONEDPL_BACKEND_SYCL
+
+template <typename... _Ts>
+struct __is_types_device_copyable: std::conjunction<sycl::is_device_copyable<_Ts>...> {};
+
+#if __INTEL_LLVM_COMPILER && (__INTEL_LLVM_COMPILER < 20240100)
+#   define _ONEDPL_IS_DEVICE_COPYABLE(TYPE, ...) sycl::is_device_copyable<TYPE<__VA_ARGS__>, \
+    std::enable_if_t<!std::is_trivially_copyable_v<TYPE<__VA_ARGS__>>>>
+#else
+#   define _ONEDPL_IS_DEVICE_COPYABLE(TYPE, ...) sycl::is_device_copyable<TYPE<__VA_ARGS__>>
+#endif //__INTEL_LLVM_COMPILER && (__INTEL_LLVM_COMPILER < 20240100)
+
+#else
+
+template <typename...>
+struct __is_types_device_copyable: std::false_type {};
+
+template <typename>
+struct __is_device_copyable_ignore: std::false_type {};
+
+# define _ONEDPL_IS_DEVICE_COPYABLE(TYPE, ...) oneapi::dpl::__internal::__is_device_copyable_ignore<TYPE<__VA_ARGS__>>
+
+#endif //_ONEDPL_BACKEND_SYCL
+
 template <typename Iterator>
 using is_const_iterator =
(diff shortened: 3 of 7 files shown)


include/oneapi/dpl/pstl/utils.h
```diff
@@ -48,6 +48,30 @@ namespace dpl
 namespace __internal
 {
 
+#if _ONEDPL_BACKEND_SYCL
+
+template <typename... _Ts>
```
danhoeflinger, line 53: In the full PR, we need comments for how this is to be used:
"__is_types_device_copyable is used to define which template parameters are required to be sycl device copyable to make the type being specialized device copyable.  Generally, this means the template arguments which are used as member variables of the type being specialized."

include/oneapi/dpl/pstl/utils.h
```diff
@@ -48,6 +48,30 @@ namespace dpl
...
+struct __is_types_device_copyable: std::conjunction<sycl::is_device_copyable<_Ts>...> {};
+
+#if __INTEL_LLVM_COMPILER && (__INTEL_LLVM_COMPILER < 20240100)
```
danhoeflinger, line 56: Proposed comment:
```
// _ONEDPL_IS_DEVICE_COPYABLE is used to declare the type we are specializing to be device copyable, with the
// template type name listed first, and then each template argument listed next.  The following is an example for 
// a structure A with two template parameters, where only the second template parameter type is used as a 
// member variable:
//
// template <typename T1, typename T2> struct A{ T2 b; };
// 
// template<typename T1, typename T2>
// struct _ONEDPL_IS_DEVICE_COPYABLE(A, T1, T2):
//    oneapi::dpl::__internal::__is_types_device_copyable<T2> {};
//

File: include/oneapi/dpl/pstl/utils.h [hunk 3]
```diff
@@ -48,6 +48,30 @@ namespace dpl
...
+#   define _ONEDPL_IS_DEVICE_COPYABLE(TYPE, ...) sycl::is_device_copyable<TYPE<__VA_ARGS__>, \
+    std::enable_if_t<!std::is_trivially_copyable_v<TYPE<__VA_ARGS__>>>>
+#else
+#   define _ONEDPL_IS_DEVICE_COPYABLE(TYPE, ...) sycl::is_device_copyable<TYPE<__VA_ARGS__>>
+#endif //__INTEL_LLVM_COMPILER && (__INTEL_LLVM_COMPILER < 20240100)
+
+#else
```
danhoeflinger, line 63:
```suggestion
#else // _ONEDPL_BACKEND_SYCL
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -48,6 +48,30 @@ namespace dpl
 namespace __internal
 {
 
+#if _ONEDPL_BACKEND_SYCL
+
+template <typename... _Ts>
+struct __is_types_device_copyable: std::conjunction<sycl::is_device_copyable<_Ts>...> {};
```
danhoeflinger, line 54:
```suggestion
struct __are_types_device_copyable: std::conjunction<sycl::is_device_copyable<_Ts>...> {};
```

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1066,6 +1066,8 @@ struct __brick_copy_n<_Tag, _ExecutionPolicy,
     {
         return ::std::copy_n(__first, __n, __result);
     }
+
+    _ONEDPL_IS_DEVICE_COPYABLE()
```
danhoeflinger, line 1070: I think this is unnecessary, and will be so whenever there are no types to provide to this macro.
The type is already "trivially copyable", so it is inherently "device copyable".

Additionally, this is in the non-hetero impl so this will always result in a noop.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -375,6 +375,8 @@ struct __brick_copy_n<__hetero_tag<_BackendTag>, _ExecutionPolicy>
     {
         __target = ::std::forward<_SourceT>(__source);
     }
+
+    _ONEDPL_IS_DEVICE_COPYABLE()
```
danhoeflinger, line 379: Unnecessary, as mentioned above.  I think `__brick_fill_n` may be the one you intended to add to.

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: I want to walk that back slightly.  It could be conceivable that there could exist a type which we just want to always claim is device copyable, with no arguments here.  

However, most (if not all) cases will be either "trivially copyable", or have some fixed type we would want to send in to this macro.  For example, if a type always had a `std::tuple<int,int>` member field unrelated to template arguments, we would need to "mark" it as "device copyable". We could use `_ONEDPL_IS_DEVICE_COPYABLE()`, but it would perhaps be more clear to instead use `_ONEDPL_IS_DEVICE_COPYABLE(std::tuple<int, int>)` to better document the reason it is required in the first place.

MikeDvorskiy:
Close the draft because the other solution (right solution)  was merged. See https://github.com/oneapi-src/oneDPL/pull/1521 for details.

PR #1514, opened 2024-04-18
Introduce smart constructor and destructor calls common helpers

SergeyKopienko:
Example: https://godbolt.org/z/3MYz3EaKb

## The main idea
The main idea - to unify constructor calls, constructors with assignment/move calls and destructor calls in oneDPL code.
This approach give us ability to make same things by similar way and avoid some errors:
- assignment without constructor call and etc.;
- destructor calls for data types where it's not mandatory required and some loops for that;
- give us ability to clarify in the future the conditions of constructor/destructor calls in one place of oneDPL code.

### `struct __op_smart_ctor` - constructor call :
```C++
template <typename _ValueType>
struct __op_smart_ctor
{
    using _ValueTypeDecayed = std::decay_t<_ValueType>;

    template <typename... Args>
    inline void
    operator()(_ValueTypeDecayed* __p_value_to, Args&&... args) const
    {
        if constexpr (std::is_trivial_v<_ValueTypeDecayed> && sizeof...(Args) <= 1)
        {
            if constexpr (sizeof...(Args) == 1)
            {
                __assign_impl(__p_value_to, std::forward<Args>(args)...);
            }
        }
        else
        {
            ::new (__p_value_to) _ValueTypeDecayed(std::forward<Args>(args)...);
        }
    }

  private:
    template <typename Arg>
    inline void
    __assign_impl(_ValueTypeDecayed* __p_value_to, Arg&& arg) const
    {
        *__p_value_to = std::forward<Arg>(arg);
    }
};
```

### `struct __op_smart_dtor` - destructor call :
```C++
template <typename _ValueType>
struct __op_smart_dtor
{
    static_assert(std::is_same_v<_ValueType, std::decay_t<_ValueType>>);

    static constexpr bool required = !std::is_trivially_destructible_v<_ValueType>;

    inline void
    operator()(_ValueType& __value) const
    {
        if constexpr (required)
            __value.~_ValueType();
    }
};
```

[std::is_scalar_v](https://en.cppreference.com/w/cpp/types/is_scalar) - [ScalarType](https://en.cppreference.com/w/cpp/named_req/ScalarType)

[std::is_trivial_v](https://en.cppreference.com/w/cpp/types/is_trivial) - [Trivial types](https://en.cppreference.com/w/cpp/named_req/TrivialType)

[std::is_trivially_destructible_v](https://en.cppreference.com/w/cpp/types/is_destructible)

### DIFF
--- a/include/oneapi/dpl/pstl/parallel_backend_utils.h
+++ b/include/oneapi/dpl/pstl/parallel_backend_utils.h
@@ -28,12 +28,6 @@ namespace dpl
 {
 namespace __utils
 {
-//------------------------------------------------------------------------
-// Uninitialized / initialized copy / move for trivial / non-trivial types
-//------------------------------------------------------------------------
-
-// struct __op_smart_ctor - helper struct for uninitialized/initialized copy/move operations.
-// It can be used to perform copy/move operations for scalar and non-scalar types.
 template <typename _ValueType>
 struct __op_smart_ctor
 {
@@ -65,15 +59,10 @@ struct __op_smart_ctor
     }
 };
 
-// struct __op_smart_dtor - helper struct for destructor calls.
-// It can be used to perform destructor calls for types,
-// which are not trivially-destructible (we call destructor)
-// and for all other types (we doesn't call destructor).
 template <typename _ValueType>
 struct __op_smart_dtor
 {
     static_assert(std::is_same_v<_ValueType, std::decay_t<_ValueType>>);
-
     static constexpr bool required = !std::is_trivially_destructible_v<_ValueType>;
 
     inline void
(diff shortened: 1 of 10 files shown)


include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -28,6 +28,72 @@ namespace dpl
...
+};
+
+// struct __op_smart_dtor - helper struct for destructor calls.
+// It can be used to perform destructor calls for types,
+// which are not trivially-destructible (we call destructor)
+// and for all other types (we doesn't call destructor).
+template <typename _ValueType>
+struct __op_smart_dtor
+{
+    static_assert(std::is_same_v<_ValueType, std::decay_t<_ValueType>>);
```
adamfidel, line 86: Is there a reason why this helper requires the input type to be decayed but the assignment helper decays it internally?

SergeyKopienko: I think because (as in example at https://godbolt.org/z/vMT36djf3, see `sa1_ref`) for reference types we should avoid destructor calls. So I think we should require `decayed` types on caller side.

PR #1515, opened 2024-04-18
Fix errors in some uninitalized_xxx algorithms

SergeyKopienko:
During the review of https://github.com/oneapi-src/oneDPL/pull/1511 @rarutyun and @kboyarinov found some incorrect implementations in our code.

To clarify this moment I extend tests for the set of algorithms


include/oneapi/dpl/pstl/memory_impl.h
```diff
@@ -119,10 +119,7 @@ struct __op_uninitialized_copy<_ExecutionPolicy>
     operator()(_SourceT&& __source, _TargetT& __target) const
     {
         using _TargetValueType = std::decay_t<_TargetT>;
-        if constexpr (std::is_trivial_v<_TargetValueType>)
```
SergeyKopienko, line 122: This code were introduced in https://github.com/oneapi-src/oneDPL/pull/746

kboyarinov:
> We doesn't call destructors in these algorithms for already created objects if some constructor throw any exception.
For example, implementation at https://en.cppreference.com/w/cpp/memory/uninitialized_fill doing destructor calls int this case.

Regarding calling destructors on created objects - as far as I understand, our implementation is correct, since the standard declares that any algorithm with `ExecutionPolicy` calls `std::terminate` on exceptions and may throw only `std::bad_alloc` if "the algorithm fails to allocate memory"

include/oneapi/dpl/pstl/memory_impl.h
SergeyKopienko: @kboyarinov, @rarutyun  Here we simple remove the `if constexpr` branch. But also we have variant to use `std::is_arithmetic_v` instead of `std::is_trivial_v`.  What do you think about this?

SergeyKopienko: Looking at https://stackoverflow.com/questions/57432081/is-placement-new-required-when-using-allocated-memory-with-fundamental-data-type I think the approach in this PR is correct.

kboyarinov: My opinion is the following. I don't understand why assigning the arithmetical type is more beneficial then calling the constructor of the same type. For cases like the code above, where know exactly that our type is arithmetical, writing assignment may be more clear to understand comparing with placement new call. We are a library and have to write generic code where the exact type is unknown. Hence we will need to support 2 variants that makes our code more complex with doubtful benefit.

kboyarinov: An other thing is that the code above implements an assignment, not a constructor and hence assigning the inner integer element is expected operation.

SergeyKopienko: You a re ton tight - please take a look to the line below:
```C++
 if (type == Tag::text) s.~string(); // explicit destroy
```

test/parallel_api/memory/specialized.algorithms/uninitialized_copy_move.pass.cpp
```diff
@@ -183,28 +198,42 @@ test_uninitialized_copy_move_by_type()
         // TODO: "memory objects" should be created in another abstraction level
         //       to avoid multiple enable/disable macro for different backends
 #endif
+        if constexpr (std::is_trivially_copy_constructible_v<T>)
```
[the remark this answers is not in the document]
SergeyKopienko: Let's take a look how they are implemented at cppreference:
https://en.cppreference.com/w/cpp/memory/uninitialized_copy
```C++
template<class InputIt, class NoThrowForwardIt>
NoThrowForwardIt uninitialized_copy(InputIt first, InputIt last, NoThrowForwardIt d_first)
{
    using T = typename [std::iterator_traits](http://en.cppreference.com/w/cpp/iterator/iterator_traits)<NoThrowForwardIt>::value_type;
    NoThrowForwardIt current = d_first;
    try
    {
        for (; first != last; ++first, (void) ++current)
            ::new (static_cast<void*>(std::addressof(*current))) T(*first); // <<< It's required copy constructor
        return current;
    }
    catch (...)
    {
        // This part of code isn't interested for us
    }
}```
where `T(*first)` required copy constructor.

File: test/parallel_api/memory/specialized.algorithms/uninitialized_custom_data_struct.h [hunk 3]
```diff
@@ -0,0 +1,65 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _UNINITIALIZED_CUSTOM_DATA_STRUCT_H
+#define _UNINITIALIZED_CUSTOM_DATA_STRUCT_H
+
+struct StructTriviallyCopyConstructible
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/parallel_api/memory/specialized.algorithms/uninitialized_custom_data_struct.h
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return true;
+    }
+    bool
+    operator!=(const StructTriviallyCopyConstructible& other) const
+    {
+        return false;
+    }
+};
+
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/parallel_api/memory/specialized.algorithms/uninitialized_construct.pass.cpp
```diff
@@ -197,6 +197,13 @@ test_uninit_construct_by_type()
 int
 main()
 {
+    static_assert(std::is_trivial_v<StructTriviallyCopyConstructible>);
+    static_assert(std::is_trivially_copy_constructible_v<StructTriviallyCopyConstructible>);
+    test_uninit_construct_by_type<StructTriviallyCopyConstructible>(1);
+
+    static_assert(std::is_trivial_v<StructTriviallyMoveConstructible>);
+    static_assert(std::is_trivially_move_constructible_v<StructTriviallyMoveConstructible>);
+    test_uninit_construct_by_type<StructTriviallyMoveConstructible>(1);
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/parallel_api/memory/specialized.algorithms/uninitialized_construct.pass.cpp
```diff
@@ -208,5 +207,10 @@ main()
     test_uninit_construct_by_type<std::int8_t>();
     test_uninit_construct_by_type<float64_t>();
 
+    // Regression tests to check compatibility of oneDPL implementation wit
```
kboyarinov, line 210: Typo (should be applied here and in other files):
```suggestion
    // Regression tests to check compatibility of oneDPL implementation with
```

test/parallel_api/memory/specialized.algorithms/uninitialized_copy_move.pass.cpp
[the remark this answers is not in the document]
SergeyKopienko: Fixed

akukanov:
This file https://github.com/oneapi-src/oneDPL/blob/7547d7dc21f1ccac583e2815595c5047cb931eb9/include/pstl/memory contains the initial implementation of uninitialized algorithms for CPU. You can see here the same generic "pattern" of using a combination of `pattern_walk*` with `brick_copy` for trivial types, and `pattern_it_walk` (iterator based) with placement new for other types. My conclusion is that this was made for supporting SIMD.

Therefore, we need to make sure that any changes we make right now work well (that is, generate efficient SIMD code) with `unseq` and `par_unseq` policies.

test/parallel_api/memory/specialized.algorithms/uninitialized_custom_data_struct.h
```diff
@@ -0,0 +1,69 @@
...
+    {
+        return true;
+    }
+    bool
+    operator!=(const NonAssignableCopyConstructible& other) const
+    {
+        return false;
+    }
+};
+static_assert(std::is_copy_constructible_v<NonAssignableCopyConstructible>);
```
kboyarinov, line 43: I guess triviality checks should not be removed, since it is important for the issue regression testing.

test/parallel_api/memory/specialized.algorithms/uninitialized_fill_destroy.pass.cpp
```diff
@@ -220,5 +220,10 @@ main()
     test_uninitialized_fill_destroy_by_type<Wrapper<std::int8_t*>>();
 #endif
 
+    // Regression tests to check compatibility of oneDPL implementation wit
```
kboyarinov, line 223:
```suggestion
    // Regression tests to check compatibility of oneDPL implementation with
```

PR #1517, opened 2024-04-18
[Tests] Fix for set_intersection kernel names

danhoeflinger:
Fixes set intersection kernel names.
Removes some unnecessary parts from a utility for generating new kernel names

Bug introduced in #983


SergeyKopienko (approved):
LGTM

PR #1520, opened 2024-04-19
Add missing fwd decl

danhoeflinger:
Adding the remaining fwd declarations.  Also reordering a few to match order of macros.

If these were omitted for some reason, let me know.  Otherwise, this should provide all the remaining ones.

@MikeDvorskiy We still need to convert to the new version of the macro / explicit specialization which allows using a subset of template parameters as originally proposed in #1513 for which require it.


adamfidel (approved):
LGTM. Thanks for adding these extra missing declarations!

danhoeflinger:
OK, merged into @adamfidel 's branch.

PR #1521, opened 2024-04-19
[oneDPL][hetero] Reworking 2 for sycl::is_device_copyable trait specializations for oneDPL types

MikeDvorskiy:
[oneDPL][hetero] Reworking 2 for sycl::is_device_copyable trait specializations for oneDPL types.

1) Adding ability to check a subset of types.
2) Adding explicit SYCL trait specialization for each (device copyable) oneDPL type. Because general general solution doesn't work at all, See example https://godbolt.org/z/1f7T3EefM
3) adds tests to check device copyability of these types with both positive and negative checks.


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -24,24 +24,18 @@
...
-            : ::std::conjunction<sycl::is_device_copyable<Ts>...>                                                      \
-        {                                                                                                              \
-        };
+namespace oneapi::dpl::__internal
+{
+template <typename... _Ts>
+struct __is_device_copyable: std::conjunction<sycl::is_device_copyable<_Ts>...> {};
+}
 
+#if __INTEL_LLVM_COMPILER && (__INTEL_LLVM_COMPILER < 20240100)
```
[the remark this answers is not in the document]
danhoeflinger: I believe this is not a change in the SYCL spec, but rather a change in the implementation within the intel compiler.  The initial implementation relied upon this extra template parameter to remove ambiguity with the trivially copyable definition.  I don't think this coincided with a change to the spec.

https://github.com/intel/llvm/commit/6c8dbf4b6fc88dd1358cc8d6defaf16c25b5f39c

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -24,24 +24,18 @@
...
-        template <typename... Ts>                                                                                      \
-        struct sycl::is_device_copyable<TYPE<Ts...>, ::std::enable_if_t<!std::is_trivially_copyable_v<TYPE<Ts...>>>>   \
-            : ::std::conjunction<sycl::is_device_copyable<Ts>...>                                                      \
-        {                                                                                                              \
-        };
+namespace oneapi::dpl::__internal
+{
+template <typename... _Ts>
+struct __is_device_copyable: std::conjunction<sycl::is_device_copyable<_Ts>...> {};
+}
```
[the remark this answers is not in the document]
MikeDvorskiy: done

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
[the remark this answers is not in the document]
akukanov: `_ONEDPL_LIBSYCL_VERSION` has nothing to do with the spec, it's the version of the SYCL headers library wrapped into our macro. The details we care of in this patch, as far as I can tell, come from the headers, not the compiler.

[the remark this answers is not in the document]
danhoeflinger: Aha, yes.  My mistake.  I was thinking that `__LIBSYCL_MAJOR_VERSION` was not specific to the dpc++ compiler, but I am wrong.  As long as we can tie the change in our code basically to the that specific commit linked (or the corresponding release), that is the important thing. 

It looks like the appropriate version is `_ONEDPL_LIBSYCL_VERSION < 70100`  https://github.com/intel/llvm/pull/11882  but please double check my work.

danhoeflinger:
As mentioned offline,  I believe this warrants a new test file which goes through at least some of these functors (if not all) and checks that it returns the appropriate response from `is_device_copyable<>` for both a device copyable template argument type (in a the subset which is checked) and for a non-device copyable template argument type.

danhoeflinger:
I've added a new test to our suite which specifically does static assertion checks to check on device copyability of all the types we are specializing here.  We check both that with device copyable template parameters types, our type is device copyable, and that with non device copyable template parameters, we have a non device copyable type.  

I use a few types defined in the test code created to not be trivially copyable, but explicitly are specialized as device copyable.  
Using these types instead of trivially copyable types for the positive checks removes the ability to rely upon the trivial copyability of types to provide device copyable, and we therefore must rely upon our explicit specializations for device copyable.

In the process of adding the tests, I caught a bug that our `__init_value` type was not included in this list of device copyable types.  I've added it, and it is tested by testing `__scan`.

I think I am done with my changes.  I will take another look at the rest of the patch, but this should be ready for review now.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -420,6 +421,9 @@ __get_accessor_ptr(const _Acc& __acc)
 #endif
 }
 
+template <typename... _Ts>
+struct __is_device_copyable: std::conjunction<sycl::is_device_copyable<_Ts>...> {};
```
rarutyun, line 425: I don't feel this belong to this file and to current namespace. It seems like the better place is `sycl_traits.h`. The reason for that is here we check the some compiler/ SYCL_LIBRARY versions, while this trait exist unconditionally. It's not the super strong preference, though, so if you have counter arguments I am happy to hear them.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -24,24 +24,15 @@
...
-        struct sycl::is_device_copyable<TYPE<Ts...>, ::std::enable_if_t<!std::is_trivially_copyable_v<TYPE<Ts...>>>>   \
-            : ::std::conjunction<sycl::is_device_copyable<Ts>...>                                                      \
-        {                                                                                                              \
-        };
-
+#if _ONEDPL_SYCL_DEVICE_COPYABLE_SPECIALIZATION_BROKEN
+// Prior to this version within DPC++ compiler, sycl::is_device_copyable relied upon a second template parameter to
+// resolve ambiguity with the general is_trivially_copyable specialization. This does not follow the sycl specification,
+// but is required to to support versions of the compiler prior to this version. Without it, the user will face a
+// compile time error of ambiguous template specialization.
```
rarutyun, line 31:
```suggestion
// Prior to the particular version of SYCL library implementation, sycl::is_device_copyable relied upon a second
// template parameter to resolve ambiguity with the general is_trivially_copyable trait. This does not follow the SYCL
// specification, but is required to support older versions of SYCL library. Without this workaround, the user will face a
// compile time error of ambiguous partial template specialization.
```

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -176,6 +224,9 @@ struct __partition_by_mask;
 template <typename _Inclusive, typename _BinaryOp, typename _InitType>
 struct __global_scan_functor;
 
+template <typename _InitType>
+ struct __init_value;
```
rarutyun, line 228: Is it just a forward declaration of some struct we should specialize for?

rarutyun, line 228:
```suggestion
struct __init_value;
```

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -191,22 +242,78 @@ struct __brick_reduce_idx;
...
-_ONEDPL_DEVICE_COPYABLE(oneapi::dpl::unseq_backend::__brick_reduce_idx)
+template <typename _ExecutionPolicy, typename _F>
+struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::unseq_backend::walk_n, _ExecutionPolicy, _F)>:
+    __dpl_sycl::__is_device_copyable<_F> {};
+
+template <typename _ExecutionPolicy, typename _F>
+struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::unseq_backend::walk_adjacent_difference, _ExecutionPolicy, _F)>:
+    __dpl_sycl::__is_device_copyable<_F> {};
+
+
```
rarutyun, line 253:
```suggestion
```

test/parallel_api/iterator/device_copyable.pass.cpp
```diff
@@ -0,0 +1,627 @@
...
+    static_assert(
+        sycl::is_device_copyable_v<oneapi::dpl::internal::replace_if_fun<int_device_copyable, noop_device_copyable>>,
+        "replace_if_fun is not device copyable with device copyable types");
+    //scan_by_key_fun
+    static_assert(
+        sycl::is_device_copyable_v<
+            oneapi::dpl::internal::scan_by_key_fun<int_device_copyable, int_device_copyable, noop_device_copyable>>,
+        "scan_by_key_fun is not device copyable with device copyable types");
+    //segmented_scan_fun
+    static_assert(sycl::is_device_copyable_v<oneapi::dpl::internal::segmented_scan_fun<
```
rarutyun, line 219: I am not sure how else we could do the same, but do we really want to use our internals in this test?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
danhoeflinger: I agree.  I'm not sure why it was put there in the first place.  I don't think we will have use for it outside `sycl_traits.h`. I put it into the `sycl_traits.h`, into a unnamed namespace.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
danhoeflinger: Incorporated this suggestion, thanks.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
danhoeflinger: Yes.  It was a type that was missed in the first round.  It was made clear in the test for `__scan` that this was necessary.  It is specialized for device copyable below.

danhoeflinger: incorporated this suggestion, thanks.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
danhoeflinger: done.

test/parallel_api/iterator/device_copyable.pass.cpp
danhoeflinger: I thought about this when adding these, but I don't have a better way of doing this.  I know we have had a general decision against doing this.  However, I think this situation deserves targeted "unit testing" and that requires reaching into the internals in this case.  

We could try to do it all via setting up contrived examples with system level tests, but that would take a lot more thought and effort for the same coverage, and result in a more computationally wasteful test with more redundancy with the rest of the test suite.  

I would be interested in establishing some "unit test" namespace, or suite of tests, where reaching into the internals was not frowned upon.  It can be quite useful for some situations like this.

danhoeflinger: Basically, I agree that "using" internals in a system level test should not be done.  
However, "testing" internals in a unit level test should be a tool we are allowed to use toward the goal of well tested software.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
akukanov: It was put there because I suggested it :) (As a general recommendation, I think it makes sense to skim through the comments of others, e.g. most recent ones, to see which questions have been discussed).

Essentially, the idea of `namespace __dpl_sycl` is to replace/complement/handle version differences for the SYCL routines, keeping their key semantics intact. This one is not much different really. But I am OK with the current code either.

test/parallel_api/iterator/device_copyable.pass.cpp
akukanov: The problem is, we do not clearly separate and not even have clear definitions of "system level" vs "unit level" tests. Perhaps I would prefer to name it "API conformance" and "implementation detail" tests.

If it's an "implementation detail" test, I prefer this to be clearly marked as such in the name, or even better put to a separate directory.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -24,24 +24,23 @@
...
-
+#if _ONEDPL_SYCL_DEVICE_COPYABLE_SPECIALIZATION_BROKEN
+// Prior to the particular version of SYCL library implementation, sycl::is_device_copyable relied upon a second
+// template parameter to resolve ambiguity with the general is_trivially_copyable trait. This does not follow the SYCL
+// specification, but is required to support older versions of SYCL library. Without this workaround, the user will face a
+// compile time error of ambiguous partial template specialization.
+#   define _ONEDPL_SPECIALIZE_FOR(TYPE, ...) TYPE<__VA_ARGS__>, std::enable_if_t<!std::is_trivially_copyable_v<TYPE<__VA_ARGS__>>>
 #else
+#   define _ONEDPL_SPECIALIZE_FOR(TYPE, ...) TYPE<__VA_ARGS__>
+#endif //__INTEL_LLVM_COMPILER && (__INTEL_LLVM_COMPILER < 20240100)
```
akukanov, line 35: The #endif comment does not match the #if condition.
Actually, for only 8 lines without nested #ifs the closing comment has very little value, in my opinion.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
danhoeflinger: Ah, I missed that discussion above, apologies.  
I understand.  Yes, I suppose essentially we are enhancing the existing `sycl::is_device_copyable` trait with this struct by allowing it to accept a pack of types, rather than a single one at a time.  I was merely looking at this as a helper for our device_copyable specializations.  

Do you see us wanting to use `__is_device_copyable` outside of `sycl_traits.h`?  If so I think that is reason to move it back.  My understanding was that we wanted to keep all the device copyable specializations here in `sycl_traits.h`, so having the helper here doesn't encourage usage elsewhere.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -24,24 +24,23 @@
...
+namespace
+{
 
-#    define _ONEDPL_DEVICE_COPYABLE(TYPE)                                                                              \
-        template <typename... Ts>                                                                                      \
-        struct sycl::is_device_copyable<TYPE<Ts...>> : ::std::conjunction<sycl::is_device_copyable<Ts>...>             \
-        {                                                                                                              \
-        };
+template <typename... _Ts>
+struct __is_device_copyable: std::conjunction<sycl::is_device_copyable<_Ts>...> {};
```
akukanov, line 41: Maybe rename to `__device_copyable`, which would better read as both "singular" and "plural"? It's purely subjective, feel free to ignore.

test/parallel_api/iterator/device_copyable.pass.cpp
danhoeflinger: Sure, we can make that change.  Do you think that we should then separate into two files the testing of the custom iterator types from the testing of internal helper types?
Or should we move all testing of device copyable trait into an "implementation_detail" folder?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
danhoeflinger: I agree that such a name change is useful.
another option I was thinking could be `__are_all_device_copyable`.
I'm not sure which I like better.  `__device_copyable` makes singular and plural more understandable, but we perhaps lose the idea that we are checking a trait.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
danhoeflinger: removed the comment.

test/parallel_api/iterator/device_copyable.pass.cpp
akukanov: Testing the oneDPL iterator types for copyability makes most sense in their respective API tests - essentially, it's part of their expected semantics (which may need to be better documented/specified, but anyway).

danhoeflinger: Ah, that is a good plan.  I'll need to move the helpers to some central location so they can be used in both tests.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
danhoeflinger: I don't see a use case where we would want this either (other than use within these specializations in `sycl_traits.h`).  My preference is to keep it as is.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
danhoeflinger: I've updated to `__are_all_device_copyable`, though I'm not tied to it, willing to change if someone has an argument.

test/parallel_api/iterator/device_copyable.pass.cpp
danhoeflinger: This is done.  I moved the tests to a folder called "implementation_details", and moved the custom iterator tests to their respective public API tests.  
I also replaced the old check for device copyable or legacy device copyable with this new static assert., and removed the utility as we should not need or rely on it anymore.

test/support/utils_sycl.h
```diff
@@ -397,6 +392,152 @@ test4buffers(int mult = kDefaultMultValue)
 {
     test4buffers<alloc_type, typename TestName::UsedValueType, TestName>(mult);
 }
+
+
+// Device copyable noop functor used in testing as surrogate for predicates, binary ops, unary functors
```
rarutyun, line 397: Do we want to keep it here? Or do we want to create some `test/device_copyable_utils.h`? I don't have strong preference to either and I don't want to hold this PR any longer but I am a little bit worried about manageability of our files. Think about this comment as of asking your opinion. I won't push for any of those.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -53,6 +53,7 @@
 #define _ONEDPL_SYCL2023_ATOMIC_REF_PRESENT (_ONEDPL_LIBSYCL_VERSION >= 50500)
 #define _ONEDPL_SYCL_SUB_GROUP_MASK_PRESENT (SYCL_EXT_ONEAPI_SUB_GROUP_MASK >= 1) && (_ONEDPL_LIBSYCL_VERSION >= 50700)
 #define _ONEDPL_SYCL_PLACEHOLDER_HOST_ACCESSOR_DEPRECATED (_ONEDPL_LIBSYCL_VERSION >= 60200)
+#define _ONEDPL_SYCL_DEVICE_COPYABLE_SPECIALIZATION_BROKEN (_ONEDPL_LIBSYCL_VERSION < 70100)
```
rarutyun, line 56: Probably the most important remaining question. Is this macro for Intel specific SYCL implementation? If no, should we also check that if it's Intel compiler or not?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -191,22 +250,77 @@ struct __brick_reduce_idx;
...
-_ONEDPL_DEVICE_COPYABLE(oneapi::dpl::unseq_backend::__brick_reduce_idx)
+template <typename _ExecutionPolicy, typename _F>
+struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::unseq_backend::walk_n, _ExecutionPolicy, _F)>:
+    __are_all_device_copyable<_F> {};
+
+template <typename _ExecutionPolicy, typename _F>
+struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::unseq_backend::walk_adjacent_difference, _ExecutionPolicy, _F)>:
+    __are_all_device_copyable<_F> {};
+
+template <typename _ExecutionPolicy, ::std::uint8_t __iters_per_work_item, typename _Operation1, typename _Operation2,
```
rarutyun, line 256: Do we want to have a little bit of clean up of `::std` in favor of just `std`? there are several places in this PR

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
danhoeflinger: Yes, this is specific to the intel compiler implementation of sycl.  However, it was my understanding that  `__LIBSYCL_MINOR_VERSION`, `__LIBSYCL_MINOR_VERSION` , `__LIBSYCL_PATCH_VERSION` were specific to the intel compiler's implementation of the sycl runtime.  

Are these macros populated by other sycl runtime implementations?

If not, then I think there does not need to be any check.  The `else` should be the version which follows the spec, and where we end up if `_ONEDPL_LIBSYCL_VERSION` is undefined.

Is your understanding different on this?

rarutyun: Good.

test/support/utils_sycl.h
danhoeflinger: These helpers are very specific and not useful for other purposes, so perhaps it is better to create a new file..  To keep with our naming conventions, perhaps `test/utils_device_copyable.h` would be better.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
danhoeflinger: sure I'll sweep through.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
danhoeflinger: Ah, ok.  They are definitely not defined by the spec.  I am not aware of any other place they would be coordinated between implementors.  I think they are merely updated with intel implementation of sycl runtime within intel/llvm 

https://intel.github.io/llvm-docs/PreprocessorMacros.html
I think what we have here is correct.  There was an earlier discussion of this here https://github.com/oneapi-src/oneDPL/pull/1521#discussion_r1572752995

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
danhoeflinger: done.

test/support/utils_sycl.h
danhoeflinger: I've separated them out.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
danhoeflinger: Ive realized I was partially wrong.  If the version macros are not populated then `_ONEDPL_LIBSYCL_VERSION` is set to `0`.  This would put non-intel sycl runtime implementations in the wrong branch.  I've adjusted the macro to exclude `0`.  We could change it to explicitly check for the intel compiler, but this more directly checks for the presence of the sycl runtime library headers.

rarutyun: Yeah, after investigation I agree. `!= 0` is a good addition. I grepped the code and there are two other places where we check `_ONEDPL_LIBSYCL_VERSION` less than something. I don't propose to fix it in this PR, though.

rarutyun (approved):
I don't see the reason to hold this PR any longer

PR #1523, opened 2024-04-23
[test] Fix dpl::copy call where we attempt to access host memory on the device

mmichel11:
In `permutation_iterator_parallel_find.pass`, we need to copy a value from the host to the device. Currently, we do this by passing a host pointer to the device but this is problematic since the host pointer is not device accessible and results in segfaults on the device.

I have resolved this by passing an iterator to a single element `std::vector` to the `copy` call so the copy to host memory is properly handled.


danhoeflinger (approved):
LGTM, thanks for the fix.

PR #1524, opened 2024-04-23
Improve kernel name generation for unnamed lambda (kernel templates)

dmitriy-sobolev:
Let's test if we allow the compiler to generate unique kernel names for a default kernel name case.

### DIFF
--- a/test/support/utils.h
+++ b/test/support/utils.h
@@ -48,6 +48,7 @@
 #if TEST_DPCPP_BACKEND_PRESENT
 #    include "utils_sycl.h"
 #    include "oneapi/dpl/experimental/kt/kernel_param.h"
+#    include "oneapi/dpl/execution" // for oneapi::dpl::execution::DefaultKernelName
 #endif
 
 namespace TestUtils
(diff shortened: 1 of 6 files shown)


test/support/utils.h
```diff
@@ -967,11 +968,19 @@ struct __kernel_name_with_idx
 
 template <int idx, typename KernelParams>
 constexpr auto
-get_new_kernel_params(KernelParams)
+get_new_kernel_params([[maybe_unused]] KernelParams params)
```
SergeyKopienko, line 971: I am not sure that `[[maybe_unused]]` is the best way here: I think usually this approach applied to return value and local variables.
Instead that we may use some self-made `PARAM_REF` macro to avoid compiler warnings.

test/support/utils.h
```diff
@@ -967,11 +968,19 @@ struct __kernel_name_with_idx
...
 {
-    return oneapi::dpl::experimental::kt::kernel_param<
+    auto new_params = oneapi::dpl::experimental::kt::kernel_param<
```
SergeyKopienko, line 973: This local variable doesn't required in come condition check variants...

test/support/utils.h
dmitriy-sobolev: [maybe_unused](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) page lists that it can be applied to a function argument.
Anyway, the argument seems unnecessary, because `KernelParams` can be reconstructed. I've removed it.

test/support/utils.h
dmitriy-sobolev: It's been removed.

test/support/utils.h
dmitriy-sobolev: > [maybe_unused](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) page lists that it can be applied to a function argument.

It looks tricky after a closer look. Function argument use-case is shown only as an example. Thanks for pointing at it.

SergeyKopienko (approved):
LGTM

PR #1527, opened 2024-04-24
Fix for iterators with const-qualified value type

danhoeflinger:
Iterator inputs to parallel APIs with a device policy which have a const-qualified value type are experiencing build errors on with MSVC or GCC standard library implementation because of static assertions in the class definition for std::vector.  

Our check for USM allocated vector iterators checks all input iterators, and within the check, instantiates a `std::vector` with a value type which matches the input iterator's value type. Prior to this PR, we do not protect against "ill formed" vector instantiations, like const-qualified, reference, or function types.  When the value type of the input iterator are any of these, the build hits `std::vector` [static assertions ](https://github.com/microsoft/STL/blob/0515a05b394596de92d08cb0f352614479a2a883/stl/inc/xmemory#L898C1-L903C80) while attempting to determine if the iterator is a USM allocator std::vector iterator.
This bug was introduced in #1438.

This PR adds a short circuit to avoid such ill-formed std::vector instantiations for specific value types (reference, const, and function types).  These will not be USM allocated std::vector iterators, so it should not effect our ability to detect those types.

This fixes the bug reproducer as reported.  I have also confirmed that input data processing still works as intended using #1429.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
@@ -177,6 +177,36 @@ struct __vector_iter_distinguishes_by_allocator<
 {
 };
 
+template <typename Iter, typename Void = void>
+struct __is_known_usm_vector_iter_impl : std::false_type
+{
+};
+
+template <typename Iter>
+struct __is_known_usm_vector_iter_impl<
+    Iter, std::enable_if_t<std::conjunction<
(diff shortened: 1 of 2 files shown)


mmichel11 (approved):
LGTM

danhoeflinger:
This also affects GCC standard library implementation in addition to MSVC.  The PR fixes both.  Libc++ does not have a static assertion to check constness of value type.

timmiesmith (approved):
LGTM. Thank you.

PR #1531, opened 2024-04-25
[Tests] Fix scan KT test: validate with internal serial scan

dmitriy-sobolev:
Use internal serial scan for validation, since not all standard libraries have it.


timmiesmith (approved):
LGTM.

PR #1532, opened 2024-04-25
Simplify constructor calls

SergeyKopienko:
In this PR we simplify constructor calls in oneDPL code:
- before:
```C++
        if constexpr (std::is_trivial_v<_TargetValueType>)
            __target = std::forward<_SourceT>(__source);
        else
            ::new (std::addressof(__target)) _TargetValueType(std::forward<_SourceT>(__source));
```
- after:
```C++
        ::new (std::addressof(__target)) _TargetValueType(std::forward<_SourceT>(__source));
```
We able to make this change because C++ compiler will optimize constructor call for trivial data types.

### DIFF
--- a/include/oneapi/dpl/pstl/memory_impl.h
+++ b/include/oneapi/dpl/pstl/memory_impl.h
@@ -41,7 +41,7 @@ __brick_uninitialized_move(_ForwardIterator __first, _ForwardIterator __last, _O
     using _ValueType = typename ::std::iterator_traits<_OutputIterator>::value_type;
     for (; __first != __last; ++__first, ++__result)
     {
-        ::new (::std::addressof(*__result)) _ValueType(::std::move(*__first));
+        ::new (std::addressof(*__result)) _ValueType(std::move(*__first));
     }
     return __result;
 }
@@ -57,7 +57,7 @@ __brick_uninitialized_move(_RandomAccessIterator __first, _RandomAccessIterator 
 
     return __unseq_backend::__simd_walk_2(
         __first, __last - __first, __result,
-        [](_ReferenceType1 __x, _ReferenceType2 __y) { ::new (::std::addressof(__y)) __ValueType(::std::move(__x)); });
+        [](_ReferenceType1 __x, _ReferenceType2 __y) { ::new (std::addressof(__y)) __ValueType(std::move(__x)); });
 }
 
 template <typename _Iterator>
@@ -92,7 +92,7 @@ __brick_uninitialized_copy(_ForwardIterator __first, _ForwardIterator __last, _O
     using _ValueType = typename ::std::iterator_traits<_OutputIterator>::value_type;
     for (; __first != __last; ++__first, ++__result)
     {
-        ::new (::std::addressof(*__result)) _ValueType(*__first);
+        ::new (std::addressof(*__result)) _ValueType(*__first);
     }
     return __result;
 }
@@ -108,7 +108,7 @@ __brick_uninitialized_copy(_RandomAccessIterator __first, _RandomAccessIterator 
 
     return __unseq_backend::__simd_walk_2(
         __first, __last - __first, __result,
-        [](_ReferenceType1 __x, _ReferenceType2 __y) { ::new (::std::addressof(__y)) __ValueType(__x); });
+        [](_ReferenceType1 __x, _ReferenceType2 __y) { ::new (std::addressof(__y)) __ValueType(__x); });
 }
 
 template <typename _ExecutionPolicy>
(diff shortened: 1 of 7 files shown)


include/oneapi/dpl/pstl/memory_impl.h
```diff
@@ -41,7 +41,7 @@ __brick_uninitialized_move(_ForwardIterator __first, _ForwardIterator __last, _O
     using _ValueType = typename ::std::iterator_traits<_OutputIterator>::value_type;
     for (; __first != __last; ++__first, ++__result)
     {
-        ::new (::std::addressof(*__result)) _ValueType(::std::move(*__first));
+        ::new (std::addressof(*__result)) _ValueType(std::move(*__first));
```
julianmi, line 44: Should we really touch this and other unrelated parts of this patch only to remove the leading `::` from `std`?

SergeyKopienko: Yes, I think so. I hope our colleagues will also write their opinion about this.
@rarutyun ?

rarutyun: I would only touch the pieces of the code that have "true" modifications. Probably I would consider to change something that is physically close in the code but also has some logical connection. In other words, I agree with Julian's concern and would not modify these pieces that only remove `::` before `std` (

rarutyun (commented):
Overall, I think the changes are correct. However, the interesting question is what is the plan with `glue_memory_impl.h?

I understand that it perhaps requires deeper investigation but still we have some nonsense checks there in my opinion.

include/oneapi/dpl/pstl/memory_impl.h
SergeyKopienko: OK, these extra changes has been removed.

SergeyKopienko:
> Overall, I think the changes are correct. However, the interesting question is what is the plan with `glue_memory_impl.h?
> 
> I understand that it perhaps requires deeper investigation but still we have some nonsense checks there in my opinion.

OK, these extra changes has been removed.

SergeyKopienko:
> Overall, I think the changes are correct. However, the interesting question is what is the plan with `glue_memory_impl.h?
> 
> I understand that it perhaps requires deeper investigation but still we have some nonsense checks there in my opinion.

After restores `std` to `::std` as before, there are no changes in this PR in the file `glue_memory_impl.h` anymore.
"we have some nonsense checks there" - @rarutyun, could you please explain, what you mean?

MikeDvorskiy (approved):
LGTM

julianmi (approved):

rarutyun:
> > Overall, I think the changes are correct. However, the interesting question is what is the plan with `glue_memory_impl.h?
> > I understand that it perhaps requires deeper investigation but still we have some nonsense checks there in my opinion.
> 
> After restores `std` to `::std` as before, there are no changes in this PR in the file `glue_memory_impl.h` anymore. "we have some nonsense checks there" - @rarutyun, could you please explain, what you mean?

I mean that this PR is removing `is_trivial_v` check. There several checks in `glue_memory_impl.h` like that and they are erroneous to me (at least at first glance). I am asking what is the plan for the similar checks in `glue_memory_impl.h`?

SergeyKopienko:
> > > Overall, I think the changes are correct. However, the interesting question is what is the plan with `glue_memory_impl.h?
> > > I understand that it perhaps requires deeper investigation but still we have some nonsense checks there in my opinion.
> > 
> > 
> > After restores `std` to `::std` as before, there are no changes in this PR in the file `glue_memory_impl.h` anymore. "we have some nonsense checks there" - @rarutyun, could you please explain, what you mean?
> 
> I mean that this PR is removing `is_trivial_v` check. There several checks in `glue_memory_impl.h` like that and they are erroneous to me (at least at first glance). I am asking what is the plan for the similar checks in `glue_memory_impl.h`?

@rarutyun, It is out of scope of this PR.
Could you please create separate issue to these checks if you think that they are erroneous to you?

rarutyun:
> > > > Overall, I think the changes are correct. However, the interesting question is what is the plan with `glue_memory_impl.h?
> > > > I understand that it perhaps requires deeper investigation but still we have some nonsense checks there in my opinion.
> > > 
> > > 
> > > After restores `std` to `::std` as before, there are no changes in this PR in the file `glue_memory_impl.h` anymore. "we have some nonsense checks there" - @rarutyun, could you please explain, what you mean?
> > 
> > 
> > I mean that this PR is removing `is_trivial_v` check. There several checks in `glue_memory_impl.h` like that and they are erroneous to me (at least at first glance). I am asking what is the plan for the similar checks in `glue_memory_impl.h`?
> 
> @rarutyun, It is out of scope of this PR. Could you please create separate issue to these checks if you think that they are erroneous to you?

I am 100% sure somebody else can create an issue. Even you @SergeyKopienko :)

PR #1533, opened 2024-04-25
Fix device_copyable for internal tuple with non trivially copyable elements

danhoeflinger:
Fix for device copyability for oneDPL's internal tuple for non-trivially copyable element types which are device copyable. (like std::tuple)

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
@@ -100,6 +100,9 @@ struct __brick_fill;
 template <class _Tag, typename _ExecutionPolicy, typename _Tp, typename>
 struct __brick_fill_n;
 
+template <typename... T>
+struct tuple;
+
 } // namespace oneapi::dpl::__internal
 
 template <typename _Pred>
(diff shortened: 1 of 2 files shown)


timmiesmith (approved):
LGTM.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -221,6 +224,12 @@ struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::
 {
 };
 
+template <typename... _Types>
+struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::tuple, _Types...)>
```
rarutyun, line 228: Sorry, but this fix seems very weird to me. Why do we need this specialization at all, if `internal::tuple` is trivially copyable if all the fields are trivially_copyable?

danhoeflinger: This is how we missed it the first time around.  Consider the following example
`internal::tuple<std::tuple<int>>`

Is it trivially copyable?  it depends on `std::tuple` implementation, implementations we use often, no.  Should it be device copyable? yes.

This is why we need the specialization, for non-trivially copyable but still device copyable types.

danhoeflinger: `std::tuple` is not unique here also, all other types in this file may also be in the same situation, if some type they depend upon is not trivially copyable.

rarutyun: ok, got it, thanks. I would prefer to `enable_if` this specialization only in case `_Types` are not trivially copyable, but unfortunately it's impossible to the best of our knowledge (prior C++20)

danhoeflinger: Ah ok.  I'm not sure what exactly that buys us.  The result effectively is the same, but just a different compile time path to the same `std::true_type` result.  

I too wish this specialization was not required, most of the types we specialize in this file are trivially copyable with trivially copyable template params.

PR #1536, opened 2024-04-25
Simplify check for usm allocated vector iterator

danhoeflinger:
After a hint from @akukanov, it seems there is a simpler approach to the fix provided in #1527.
Instead of specifically avoiding instantiating `std::vector<T>` for types `T` which are not valid, we can merely use `std::decay_t` prior to our USM iterator checks.  
This will prevent any ill-formed `std::vector` from being instantiated, and also should not damage our ability to detect any valid USM allocated `std::vector::iterator`.  
Any `std::vector::iterator` type created from a decayed value type should not match the original iterator type if `std::decay` has an effect on the value type.  If `std::decay` has an effect on the value type, then it is not a valid `std::vector`.

This should fix the problem same issue with simpler code, and not change our behavior.


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -178,35 +179,19 @@ struct __vector_iter_distinguishes_by_allocator<
...
 
 template <typename Iter>
-struct __is_known_usm_vector_iter_impl<
+struct __is_known_usm_vector_iter<
     Iter, std::enable_if_t<std::conjunction<
               std::disjunction<std::is_same<Iter, oneapi::dpl::__internal::__usm_shared_alloc_vec_iter<Iter>>,
                                std::is_same<Iter, oneapi::dpl::__internal::__usm_host_alloc_vec_iter<Iter>>>,
               oneapi::dpl::__internal::__vector_iter_distinguishes_by_allocator<Iter>>::value>> : std::true_type
 {
 };
```
akukanov, line 193: I wonder if the same result would be get by simply declaring `__is_known_usm_vector_iter` as an alias to the logical expression in the specialization, starting from `std::conjunction`. Also maybe the same technic can be used with `__vector_iter_distinguishes_by_allocator`? And the last, do we need those as types, or constexpr bool variables would be enough?

danhoeflinger: Good point.  I had this originally but ran into some errors with permutation iterator index maps.  However, this is a problem with the way we define `permutation_iterators` with relation to `is_passed_directly`, rather than anything else.  

I've resolved that problem, and now simplified these to `constexpr bool`.

akukanov (approved):
Looks good to me and clearly improves the code, thanks.

PR #1538, opened 2024-04-26
[Tests] Fix header include for device_copyable.pass

danhoeflinger:
Resolves incorrect header inclusion for sycl.hpp.  
This PR uses our local `util_sycl_defs.h` in place of `sycl.hpp` to ensure we have access to sycl device_copyable checks.


mmichel11 (approved):
LGTM

PR #1540, opened 2024-04-26
[Tests] Fix broken macro so it can be used as expression

danhoeflinger:
We can only count on undefined macros to be treated as `0` within `#if` checks.  Within `tuple_get_const_rvalue.pass`, this macro is used as an expression to determine if the test was skipped or not, and this fails when `_GLIBCXX_RELEASE` is undefined.

Here is the usage 
https://github.com/oneapi-src/oneDPL/blob/3d762ceb57721b6fd6059086ccc757e2d2466f08/test/xpu_api/tuple/tuple.tuple/tuple.elem/tuple_get_const_rvalue.pass.cpp#L60

This changes the macro to work both in `#if` and also as an expression, allowing the test to build.

C++ Standard 16.1.4 describes when 0 is used in replacement of undefined identifiers.  (but is only relevant within `#if` checks).


mmichel11 (approved):
LGTM. Thanks for the fix

SergeyKopienko (commented):
LGTM

PR #1543, opened 2024-04-29
Fix broken test `tuple_get_const_rvalue.pass` under GCC 7

SergeyKopienko:
In this PR we fix broken test `tuple_get_const_rvalue.pass` under GCC 7.

In the file `test/support/test_config.h` we taking a look at `_GLIBCXX_RELEASE` state :
```C++
// GCC7 std::get doesn't return const rvalue reference from const rvalue reference of tuple
#if defined(_GLIBCXX_RELEASE)
#   define _PSTL_TEST_GCC7_RVALUE_TUPLE_GET_BROKEN (_GLIBCXX_RELEASE < 8)
#else
#   define _PSTL_TEST_GCC7_RVALUE_TUPLE_GET_BROKEN 0
#endif
```

But in concrete case it's not enough, because not files from standard library has been included before and `_GLIBCXX_RELEASE` is undefined: so we have
`#   define _PSTL_TEST_GCC7_RVALUE_TUPLE_GET_BROKEN 0` 
as result.
In this case the test `test/xpu_api/tuple/tuple.tuple/tuple.elem/tuple_get_const_rvalue.pass.cpp` has compile errors due https://gcc.gnu.org/onlinedocs/gcc-7.5.0/cpp/Defined.html.

### DIFF
--- a/test/support/test_config.h
+++ b/test/support/test_config.h
@@ -16,13 +16,6 @@
 #ifndef _TEST_CONFIG_H
 #define _TEST_CONFIG_H
 
-// Any include from standard library required to have correct state of _GLIBCXX_RELEASE
-#if __has_include(<version>)
-#   include <version>
-#else
-#   include <ciso646>
-#endif
-
 #define _PSTL_TEST_STRING(X) _PSTL_TEST_STRING_AUX(oneapi/dpl/X)
 #define _PSTL_TEST_STRING_AUX(X) #X
 //to support the optional including: <algorithm>, <memory>, <numeric> or <pstl/algorithm>, <pstl/memory>, <pstl/numeric>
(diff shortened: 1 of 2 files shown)


test/support/test_config.h
```diff
@@ -16,6 +16,13 @@
 #ifndef _TEST_CONFIG_H
 #define _TEST_CONFIG_H
 
+// Any include from standard library required to have correct state of _GLIBCXX_RELEASE
+#if __cplusplus >= 202002L || _MSVC_LANG >= 202002L
```
akukanov: Better `#if __has_include(<version>)`; `__has_include` is a standard feature since C++17 so can be used freely.

See also https://github.com/oneapi-src/oneDPL/blob/5c8bb63e2d1288c4c2c71bc740ab1a4f80644fcd/test/support/test_macros.h#L19-L30 - I think it makes sense to eliminate duplicated code (but please, do not introduce yet another header just for these 5 lines) or at least have the same code if the headers should remain independent.

danhoeflinger (dismissed):
LGTM, please wait for green CI to merge.  Thanks for the fix!

test/support/test_config.h
SergeyKopienko: Fixed

SergeyKopienko:
I create temporary PR https://github.com/oneapi-src/oneDPL/pull/1544 to check that this failed CI task is not linked with my change together with PR https://github.com/oneapi-src/oneDPL/pull/1250

test/support/test_config.h
```diff
@@ -52,11 +52,7 @@
 // GCC10 produces wrong answer calling exclusive_scan using vectorized polices
 #define TEST_GCC10_EXCLUSIVE_SCAN_BROKEN (_GLIBCXX_RELEASE == 10)
 // GCC7 std::get doesn't return const rvalue reference from const rvalue reference of tuple
-#if defined(_GLIBCXX_RELEASE)
-#   define _PSTL_TEST_GCC7_RVALUE_TUPLE_GET_BROKEN (_GLIBCXX_RELEASE < 8)
-#else
-#   define _PSTL_TEST_GCC7_RVALUE_TUPLE_GET_BROKEN 0
-#endif
+#define _PSTL_TEST_GCC7_RVALUE_TUPLE_GET_BROKEN (_GLIBCXX_RELEASE > 0 && _GLIBCXX_RELEASE < 8)
```
SergeyKopienko, line 55: This part `_GLIBCXX_RELEASE > 0` is mandatory required for the case when `_GLIBCXX_RELEASE` is undefined - for example on Windows compilers and etc.

dmitriy-sobolev (approved):
LGTM

danhoeflinger (approved):
LGTM

PR #1545, opened 2024-04-29
[Tests] Speedup permutation iterator partial sort 

danhoeflinger:
Speedup the tests for permutation iterator partial sort by running the verification in parallel on the host (instead of a transfer back to the device), and most importantly cutting down the number of iterations run of the "partial" part of the sort.


timmiesmith (dismissed):
LGTM. Thank you.

mmichel11 (approved):
LGTM

SergeyKopienko (approved):
LGTM

PR #1546, opened 2024-04-30
[test] Speed up the reduce_by_segment.pass test

mmichel11:
`reduce_by_segment.pass` has become a bottleneck in CI runtime and is taking excessively long compared to similar tests. I have made the following adjustments to speed it up:

- Previously for custom binary predicates and / or operators, we would first test the case with the default predicate and default operator (`std::plus<T>` and `std::equal_to<T>`) on the underlying data and then the default predicate and custom operator before testing with the custom predicate and custom operator. In my opinion, this is overtesting `std::equal_to` and `std::plus`. which is ultimately leading to long compilation and run times. I have replaced this to only test with the custom predicate and operator and use the default predicate / operator APIs if possible with a compile time check.
- We had several tests with same the same operator and predicate but different datatype. I have reduced this to save time.
- Lastly, I switched `run_test<::std::uint64_t, UserBinaryPredicate<::std::uint64_t>, MaxFunctor<::std::uint64_t>>();` to use `uint32_t` We already test with an 8-byte type elsewhere and realized this test is commonly skipped due to lack of 64-bit int support on some devices.

### DIFF
--- a/test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
+++ b/test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
@@ -148,51 +148,35 @@ DEFINE_TEST_2(test_reduce_by_segment, BinaryPredicate, BinaryOperation)
         TestDataTransfer<UDTKind::eRes2, Size> host_res(*this, n);
 
         typedef typename ::std::iterator_traits<Iterator1>::value_type KeyT;
+        typedef typename ::std::iterator_traits<Iterator2>::value_type ValT;
 
-        // call algorithm with no optional arguments
-        initialize_data(host_keys.get(), host_vals.get(), host_res_keys.get(), host_res.get(), n);
-        update_data(host_keys, host_vals, host_res_keys, host_res);
-
-        auto new_policy = make_new_policy<new_kernel_name<Policy, 0>>(exec);
-        auto res1 =
-            oneapi::dpl::reduce_by_segment(new_policy, keys_first, keys_last, vals_first, key_res_first, val_res_first);
+        std::pair<Iterator3, Iterator4> res;
+        if constexpr (std::is_same_v<std::equal_to<KeyT>, std::decay_t<BinaryPredicate>> &&
+                      std::is_same_v<std::plus<ValT>, std::decay_t<BinaryOperation>>)
+        {
+            auto new_policy = make_new_policy<new_kernel_name<Policy, 0>>(exec);
+            res = oneapi::dpl::reduce_by_segment(std::move(new_policy), keys_first, keys_last, vals_first,
+                                                 key_res_first, val_res_first);
+        }
+        else if constexpr (std::is_same_v<std::equal_to<KeyT>, std::decay_t<BinaryPredicate>>)
+        {
+            auto new_policy = make_new_policy<new_kernel_name<Policy, 1>>(exec);
+            res = oneapi::dpl::reduce_by_segment(std::move(new_policy), keys_first, keys_last, vals_first,
+                                                 key_res_first, val_res_first, BinaryPredicate());
+        }
+        else
+        {
+            auto new_policy = make_new_policy<new_kernel_name<Policy, 2>>(exec);
+            res = oneapi::dpl::reduce_by_segment(std::move(new_policy), keys_first, keys_last, vals_first,
+                                                 key_res_first, val_res_first, BinaryPredicate(), BinaryOperation());
+        }
         exec.queue().wait_and_throw();
 
         retrieve_data(host_keys, host_vals, host_res_keys, host_res);
-        size_t segments_key_ret1 = ::std::distance(key_res_first, res1.first);
-        size_t segments_val_ret1 = ::std::distance(val_res_first, res1.second);
-        check_values(host_keys.get(), host_vals.get(), host_res_keys.get(), host_res.get(), n, segments_key_ret1,
-                     segments_val_ret1);
-
-        // call algorithm with predicate
-        initialize_data(host_keys.get(), host_vals.get(), host_res_keys.get(), host_res.get(), n);
-        update_data(host_keys, host_vals, host_res_keys, host_res);
-
-        auto new_policy2 = make_new_policy<new_kernel_name<Policy, 1>>(exec);
-        auto res2 = oneapi::dpl::reduce_by_segment(new_policy2, keys_first, keys_last, vals_first, key_res_first,
-                                                   val_res_first, BinaryPredicate());
-        exec.queue().wait_and_throw();
-
-        retrieve_data(host_keys, host_vals, host_res_keys, host_res);
-        size_t segments_key_ret2 = ::std::distance(key_res_first, res2.first);
-        size_t segments_val_ret2 = ::std::distance(val_res_first, res2.second);
-        check_values(host_keys.get(), host_vals.get(), host_res_keys.get(), host_res.get(), n, segments_key_ret2,
-                     segments_val_ret2, BinaryPredicate());
-
-        // call algorithm with predicate and operator
-        initialize_data(host_keys.get(), host_vals.get(), host_res_keys.get(), host_res.get(), n);
-        update_data(host_keys, host_vals, host_res_keys, host_res);
-
-        auto new_policy3 = make_new_policy<new_kernel_name<Policy, 2>>(exec);
-        auto res3 = oneapi::dpl::reduce_by_segment(new_policy3, keys_first, keys_last, vals_first, key_res_first,
-                                                   val_res_first, BinaryPredicate(), BinaryOperation());
-        exec.queue().wait_and_throw();
-
-        retrieve_data(host_keys, host_vals, host_res_keys, host_res);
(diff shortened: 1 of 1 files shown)


SergeyKopienko:
@mmichel11 I believe alternative approach available: not change functional but split to separate 2-3 tests. What do you think?

danhoeflinger:
> @mmichel11 I believe alternative approach available: not change functional but split to separate 2-3 tests. What do you think?

I think the goal of this is to reduce overall test time for the full suite, as we are running over.

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
```diff
@@ -368,16 +350,13 @@ main()
     test_flag_pred<sycl::usm::alloc::device, class KernelName2, dpl::complex<float>>();
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
-    run_test<::std::uint64_t, UserBinaryPredicate<::std::uint64_t>, MaxFunctor<::std::uint64_t>>();
+    run_test<::std::uint32_t, UserBinaryPredicate<::std::uint32_t>, MaxFunctor<::std::uint32_t>>();
     run_test<::std::complex<float>, UserBinaryPredicate<::std::complex<float>>, MaxFunctor<::std::complex<float>>>();
 
     run_test<int, ::std::equal_to<int>, ::std::plus<int>>();
-    run_test<float, ::std::equal_to<float>, ::std::plus<float>>();
```
SergeyKopienko, line 375: @mmichel11, you exclude test for `float` type: but float type available on each device, and `double` type only on some devices.
So why do you think that this approach is correct?

SergeyKopienko:
> > @mmichel11 I believe alternative approach available: not change functional but split to separate 2-3 tests. What do you think?
> 
> I think the goal of this is to reduce overall test time for the full suite, as we are running over.

Disagree with you: please see my comment at https://github.com/oneapi-src/oneDPL/pull/1546/files#r1584720076
I believe the reduce execution time of each test is real task because overwise test will be cancelled by CI system.
But reduce testing functional looks like not so correct task...

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
mmichel11: That is a good point. I could add back the `float` test and remove the `double`. I don't think there is much of a need to test both `float` and `double` with the same operator and predicate. If one passes, I suspect the other one will likely be fine as well.

mmichel11:
> > > @mmichel11 I believe alternative approach available: not change functional but split to separate 2-3 tests. What do you think?
> > 
> > 
> > I think the goal of this is to reduce overall test time for the full suite, as we are running over.
> 
> Disagree with you: please see my comment at https://github.com/oneapi-src/oneDPL/pull/1546/files#r1584720076 I believe the reduce execution time of each test is real task because overwise test will be cancelled by CI system. But reduce testing functional looks like not so correct task...

We are seeing timeouts in some of the configurations and this is one of the longer running tests. My approach has been to not remove any of the operators and predicates we are testing but reduce the datatypes we are testing each operator / predicate with.

Do you think testing `int`, `float`, and `double`, with each operator / predicate is necessary or is removing in some instances acceptable?

SergeyKopienko:
> > > > @mmichel11 I believe alternative approach available: not change functional but split to separate 2-3 tests. What do you think?
> > > 
> > > 
> > > I think the goal of this is to reduce overall test time for the full suite, as we are running over.
> > 
> > 
> > Disagree with you: please see my comment at https://github.com/oneapi-src/oneDPL/pull/1546/files#r1584720076 I believe the reduce execution time of each test is real task because overwise test will be cancelled by CI system. But reduce testing functional looks like not so correct task...
> 
> We are seeing timeouts in some of the configurations and this is one of the longer running tests. My approach has been to not remove any of the operators and predicates we are testing but reduce the datatypes we are testing each operator / predicate with.
> 
> Do you think testing `int`, `float`, and `double`, with each operator / predicate is necessary or is removing in some instances acceptable?

I seen a lot of different errors for these combinations of types / operations. So I would like to not remove them.
It was about `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION`, `__can_use_known_identity`, `__has_known_identity` and group operations for different types.

mmichel11:
> > > > > @mmichel11 I believe alternative approach available: not change functional but split to separate 2-3 tests. What do you think?
> > > > 
> > > > 
> > > > I think the goal of this is to reduce overall test time for the full suite, as we are running over.
> > > 
> > > 
> > > Disagree with you: please see my comment at https://github.com/oneapi-src/oneDPL/pull/1546/files#r1584720076 I believe the reduce execution time of each test is real task because overwise test will be cancelled by CI system. But reduce testing functional looks like not so correct task...
> > 
> > 
> > We are seeing timeouts in some of the configurations and this is one of the longer running tests. My approach has been to not remove any of the operators and predicates we are testing but reduce the datatypes we are testing each operator / predicate with.
> > Do you think testing `int`, `float`, and `double`, with each operator / predicate is necessary or is removing in some instances acceptable?
> 
> I seen a lot of different errors for these combinations of types / operations. So I would like to not remove them. It was about `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION`, `__can_use_known_identity`, `__has_known_identity` and group operations for different types.

Understandable, I have added back the removed cases from main. The biggest change is not running `3` separate `reduce_by_segment` calls per test but only one which was resulting in duplicate testing in many instances.

SergeyKopienko:
@mmichel11 I think that `auto new_policy = make_new_policy<new_kernel_name<Policy, 0>>(exec);` and etc. new policies not required anymore as far as now you have only one operation under `if constexpr` ?

mmichel11:
> @mmichel11 I think that `auto new_policy = make_new_policy<new_kernel_name<Policy, 0>>(exec);` and etc. new policies not required anymore as far as now you have only one operation under `if constexpr` ?

I have addressed this. Additionally, I fixed some bugs I had introduced in the PR and verified the tests pass.

My goal was to cut test time down by factor of two for FPGA_EMU + debug for CI purposes. I removed one test to do so: `run_test<::std::uint64_t, UserBinaryPredicate<::std::uint64_t>, MaxFunctor<::std::uint64_t>>();`. We are already testing with the identical predicate / operator with `std::complex<float>`, so I think this test is not adding any value. Let me know if you object.

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
```diff
@@ -150,24 +150,24 @@ DEFINE_TEST_2(test_reduce_by_segment, BinaryPredicate, BinaryOperation)
...
+        initialize_data(host_keys.get(), host_vals.get(), host_res_keys.get(), host_res.get(), n);
+        update_data(host_keys, host_vals, host_res_keys, host_res);
+
         std::pair<Iterator3, Iterator4> res;
         if constexpr (std::is_same_v<std::equal_to<KeyT>, std::decay_t<BinaryPredicate>> &&
                       std::is_same_v<std::plus<ValT>, std::decay_t<BinaryOperation>>)
         {
-            auto new_policy = make_new_policy<new_kernel_name<Policy, 0>>(exec);
-            res = oneapi::dpl::reduce_by_segment(std::move(new_policy), keys_first, keys_last, vals_first,
+            res = oneapi::dpl::reduce_by_segment(std::forward<Policy>(exec), keys_first, keys_last, vals_first,
```
SergeyKopienko, line 160: I made issue here: `std::forward<Policy>(exec)`
Because later you have `exec.queue().wait_and_throw();` call

mmichel11: Thanks I have fixed this for this case. For the host case, we do not use exec later on so it can be forwarded.

SergeyKopienko (approved):
LGTM

danhoeflinger (approved):
LGTM,  
The only question I have is how much the removal of the `uint64_t` case gets us on the slowest configuration / environment.  If this is one of the tests which is skipped by that case anyway, I'd be inclined to keep it.  Otherwise, I'm fine with dropping it.

mmichel11:
> LGTM, The only question I have is how much the removal of the `uint64_t` case gets us on the slowest configuration / environment. If this is one of the tests which is skipped by that case anyway, I'd be inclined to keep it. Otherwise, I'm fine with dropping it.

With FPGA EMU in debug mode on my laptop, compilation time is roughly the same but runtime is reduced from 50 sec to 42 sec. These times are still a lot faster than what we have seen on systems with lower specs, so I do not know how it translates there.

danhoeflinger:
> With FPGA EMU in debug mode on my laptop, compilation time is roughly the same but runtime is reduced from 50 sec to 42 sec. These times are still a lot faster than what we have seen on systems on Windows with lower specs, so I do not know how it translates there.

I think this answers my question that FPGA_EMU debug does support the `uint64_t` case, and it is therefore beneficial to cut it.
I'm fine with dropping it.  
LGTM

PR #1547, opened 2024-04-30
Simplify sycl_backend::submit(SelectionHandle s, Function&& f, Args&&... args) function

SergeyKopienko:


include/oneapi/dpl/internal/dynamic_selection_impl/sycl_backend.h
```diff
@@ -184,60 +184,44 @@ class sycl_backend
...
         {
-            report_clock_type::time_point t0;
-            if constexpr (report_task_time)
-            {
-                if (!is_profiling_enabled)
-                {
-                    t0 = report_clock_type::now();
-                }
-            }
+            const auto t0 = report_clock_type::now();
```
AnuyaWelling2801, line 193: We only need to calculate t0 when report_task_time is true and is_profile_enabled is called

SergeyKopienko: I believe we are able to calculate it every once, because it's very lightweight operation.
But use it when it's really required.

PR #1548, opened 2024-04-30
Additional includes into test/support/test_config.h

SergeyKopienko:
In this PR we have 3 big changes:

1) In the file `test/support/test_config.h` we add inclusion of some files from standard library to have correct definitions of used defines:
```C++
// Any include from standard library required to have correct state of _GLIBCXX_RELEASE
#if __has_include(<version>)
#   include <version>
#else
#   include <ciso646>
#endif
```
2) In all tests now we include `test/support/test_config.h` at first place.
No additional includes required anymore if they are really not required for the test: [example1](https://github.com/oneapi-src/oneDPL/pull/1548/files#diff-d6f212e5d19f8d89f8e4da0e0289b29c43a44037d21c70dda8240c9d96a5ea0f), [example2](https://github.com/oneapi-src/oneDPL/pull/1548/files#diff-4e77151a0523c3322ea0d9e095266b06653bac4e259a6688665ccd3cb71da8c9)
3) For CI errors after (1) and (2) @kboyarinov prepared the fix for the failed CI tasks at `include/oneapi/dpl/internal/common_config.h` : see [changes](https://github.com/oneapi-src/oneDPL/pull/1548/files#diff-4c32e30dceaeba2ff83a44fecd56fb4bfe504e60b3f0e2bed77a74abc0b54176).
Also this fix resolves a lot of errors in the tests compiled with `g++` compiler.

Also we have the alternative approach at https://github.com/oneapi-src/oneDPL/pull/1555

### DIFF
--- a/include/oneapi/dpl/internal/common_config.h
+++ b/include/oneapi/dpl/internal/common_config.h
@@ -61,8 +61,8 @@
 #            define __PSTL_USE_PAR_POLICIES 0
 #        endif
 #        ifdef __PSTL_PAR_BACKEND_TBB
-#             undef __PSTL_PAR_BACKEND_TBB
-#             define __PSTL_PAR_BACKEND_TBB 0
+#            undef __PSTL_PAR_BACKEND_TBB
+#            define __PSTL_PAR_BACKEND_TBB 0
 #        endif
 #        define PSTL_USE_PARALLEL_POLICIES (_GLIBCXX_RELEASE != 9)
 #    endif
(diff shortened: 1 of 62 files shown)


danhoeflinger (approved):
LGTM,
I think before merging (especially for inclusion with 2022.6.0), we should:
1) Create an issue to add testing for include order which includes oneDPL headers *before* stdlib headers, since this changes all of our tests to include a stdlib header first ,as discussed offline.
2) Get approval from another person.
3) Run a full nightly CI with this branch to see the effect of the changes in `common_config.h` (that there are no unforeseen consequences)  I believe that this is merely applying the same treatment for the missing TBB headers as we do for the mismatched tbb install above, so I think it should be OK.

danhoeflinger:
Discussed offline, with another approval, it seems reasonable to merge and check nightly CI due to contention on runners rather than generating an extra full run.  We just need to follow up to confirm we haven't caused trouble after the next run.

timmiesmith (approved):
LGTM

PR #1549, opened 2024-04-30
Redefine `timing_t` as `std::chrono::milliseconds`

SergeyKopienko:
In this PR we redefine `timing_t` as `std::chrono::milliseconds` and remove `count()` function call at all, so we always operates by value with physical sence.


akukanov:
As far as I understand, this change might affect the public API of dynamic selection, namely, the reporting requirements of the auto-tune policy. I suggest to take it for discussion, but not apply right now.

PR #1551, opened 2024-05-01
Change usage of _PSTL_UD[R/S]_PRESENT macros

adamfidel:
On some platforms, specifically Fedora 40 with GCC 14, the included `pstl_config.h` is such that the macros  `_PSTL_UDR_PRESENT` and `_PSTL_UDS_PRESENT` are present but defined as an empty macro. This is causing compilation errors.

In this PR, we work around the empty macros by first testing `_PSTL_UDR_PRESENT + 0`, which evaluates to `+0` (i.e., false) if the macro is undefined and the correct value if it is defined.

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3882,7 +3882,7 @@ _RandomAccessIterator
 __brick_min_element(_RandomAccessIterator __first, _RandomAccessIterator __last, _Compare __comp,
                     /* __is_vector = */ ::std::true_type) noexcept
 {
-#if _ONEDPL_IS_ONE_OR_EMPTY(_PSTL_UDR_PRESENT) || _ONEDPL_UDR_PRESENT
+#if /*_PSTL_UDR_PRESENT ||*/ _ONEDPL_UDR_PRESENT
     return __unseq_backend::__simd_min_element(__first, __last - __first, __comp);
 #else
     return ::std::min_element(__first, __last, __comp);
@@ -3949,7 +3949,7 @@ template <typename _RandomAccessIterator, typename _Compare>
 __brick_minmax_element(_RandomAccessIterator __first, _RandomAccessIterator __last, _Compare __comp,
                        /* __is_vector = */ ::std::true_type) noexcept
 {
-#if _ONEDPL_IS_ONE_OR_EMPTY(_PSTL_UDR_PRESENT) || _ONEDPL_UDR_PRESENT
+#if /*_PSTL_UDR_PRESENT ||*/ _ONEDPL_UDR_PRESENT
     return __unseq_backend::__simd_minmax_element(__first, __last - __first, __comp);
 #else
     return ::std::minmax_element(__first, __last, __comp);
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3882,7 +3882,7 @@ _RandomAccessIterator
 __brick_min_element(_RandomAccessIterator __first, _RandomAccessIterator __last, _Compare __comp,
                     /* __is_vector = */ ::std::true_type) noexcept
 {
-#if _ONEDPL_DEFINED_AND_NOT_ZERO(_PSTL_UDR_PRESENT) || _ONEDPL_UDR_PRESENT
+#if _ONEDPL_DEFINED_AND_IS_ONE(_PSTL_UDR_PRESENT) || _ONEDPL_UDR_PRESENT
     return __unseq_backend::__simd_min_element(__first, __last - __first, __comp);
 #else
     return ::std::min_element(__first, __last, __comp);
@@ -3949,7 +3949,7 @@ template <typename _RandomAccessIterator, typename _Compare>
 __brick_minmax_element(_RandomAccessIterator __first, _RandomAccessIterator __last, _Compare __comp,
                        /* __is_vector = */ ::std::true_type) noexcept
 {
-#if _ONEDPL_DEFINED_AND_NOT_ZERO(_PSTL_UDR_PRESENT) || _ONEDPL_UDR_PRESENT
+#if _ONEDPL_DEFINED_AND_IS_ONE(_PSTL_UDR_PRESENT) || _ONEDPL_UDR_PRESENT
     return __unseq_backend::__simd_minmax_element(__first, __last - __first, __comp);
 #else
     return ::std::minmax_element(__first, __last, __comp);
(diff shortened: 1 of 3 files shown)


danhoeflinger (commented):
It seems this is a place it is defined like this:

https://github.com/gcc-mirror/gcc/blob/0695aba3e987f4bb06c95f29ff90a8a3234e1507/libstdc%2B%2B-v3/include/pstl/pstl_config.h#L127

Should we instead align our _PSTL_UDR_PRESENT with how others treat it (defined on undefined) and then check `defined(_PSTL_UDR_PRESENT)`?

danhoeflinger:
Alternatively, perhaps we can instead "process" an empty but defined `_PSTL_UDR_PRESENT` into some local version that we then use so that we can have this workaround only in one place.

adamfidel:
> @danhoeflinger: Should we instead align our _PSTL_UDR_PRESENT with how others treat it (defined on undefined) and then check `defined(_PSTL_UDR_PRESENT)`?

 Are you suggesting that we change oneDPL's `pstl_config.h` to make this macro either undefined or empty defined, instead of always being defined 0 or 1 as it is currently? I'm not sure what the "correct" version is or how the upstream PSTL manages their `pstl_config.h`. I believe @rarutyun might have to weigh in on that.

> @danhoeflinger: Alternatively, perhaps we can instead "process" an empty but defined _PSTL_UDR_PRESENT into some local version that we then use so that we can have this workaround only in one place.

That was my original thought, but I couldn't quite place where to put it.

It doesn't make sense to put it in `pstl_config.h` since that is where the issue arises (and there's the question of whether it will be included at all instead of the system's `pstl_config.h`). It also doesn't makes sense to put it into `onedpl_config.h` because there we define our own  `_ONEDPL_UDR_PRESENT` which raises the question, why do we have two macros for the same thing? I don't have the historical context for that.

danhoeflinger:
We do not include `pstl_config.h` anywhere in our own code.  I think it is here to easier sync with upstream, but someone with more knowledge of the sync process can correct me there.  This wouldn't be the place to create our own version, but I also don't know the appropriate place.

I'm realizing that my original suggestion to align with their handling doesn't work, because we may encounter either way of defining `_PSTL_UDR_PRESENT` from an external source.  Looking at the blame, https://github.com/gcc-mirror/gcc/commit/3162ca09dbdc2ea3ff1d4b752edf76955b6d94f9 is where they made the switch away from our current approach to this one.

It seems we will have the same problem with `_PSTL_UDS_PRESENT` as well.  

One option is to make a macro: 
`_ONEDPL_DEFINED_AND_NOT_ZERO(FLAG)` which could be reused for something like this and have the workaround in one place.  However, again I don't know where to put that.

adamfidel:
> @danhoeflinger One option is to make a macro: `_ONEDPL_DEFINED_AND_NOT_ZERO(FLAG)` which could be reused for something like this and have the workaround in one place. However, again I don't know where to put that.

I changed this PR to create a new `_ONEDPL_DEFINED_AND_NOT_ZERO` macro and use that instead. I think it made it much simpler.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -41,6 +41,8 @@
 #    include <cstring> // memcpy
 #endif
 
+#define _ONEDPL_DEFINED_AND_NOT_ZERO(X) (X + 0)
```
danhoeflinger, line 44: I think this doesn't do what we are advertising by the name (and we want the semantics of the name I think).  
`_PSTL_UDS_PRESENT` is defined to empty in the external std lib def when it is "true" and left undefined when it is "false".  

This macro will convert that empty string to a "false" rather than a "true".

I think we want it to be "true" / "1" any time it is defined and not explicitly set to "0".

include/oneapi/dpl/pstl/numeric_impl.h
```diff
@@ -213,13 +213,14 @@ __brick_transform_scan(_RandomAccessIterator __first, _RandomAccessIterator __la
...
 {
-#if (_PSTL_UDS_PRESENT || _ONEDPL_UDS_PRESENT)
+#if _ONEDPL_DEFINED_AND_NOT_ZERO(_PSTL_UDS_PRESENT) || _ONEDPL_UDS_PRESENT
     if (_Inclusive() || !oneapi::dpl::__internal::__iterators_possibly_equal(__first, __result))
     {
         return __unseq_backend::__simd_scan(__first, __last - __first, __result, __unary_op, __init, __binary_op,
                                             _Inclusive());
     }
 #endif
+
```
danhoeflinger, line 223: extra space

danhoeflinger:
> I changed this PR to create a new `_ONEDPL_DEFINED_AND_NOT_ZERO` macro and use that instead. I think it made it much simpler.

I like it a lot better now after this switch.  One real comment (above) to address from my perspective, but I also think this deserves a look from @rarutyun or someone else with better understanding of the upstream process before going forward.

include/oneapi/dpl/pstl/utils.h
adamfidel: Yes, you are correct. This macro should be truthy if the input is defined and empty, but that is not the case here.

I made this [godbolt](https://godbolt.org/z/z1xK4Gd7c) to try to write a macro that has all of the properties that we want, but I cannot come up with a macro that satisfies everything. So perhaps the original version of the PR is the solution that we should go with for now.

danhoeflinger: Yeah, this is kind of an annoying case.

How about this:
https://godbolt.org/z/GY1qTjb5e

It would require basically converting the public pstl defines to our own local copy of the pstl defines (which would be "0" or "1"), then we could use our own copy as "normal".  

Its possible that we could work it out so that our local `_ONEDPL_UDS_PRESENT` is our "local" copy, where we only attempt to define it ourselves if it is undefined.

adamfidel: After some offline discussion, @danhoeflinger and I arrived at a version of the macro that now correctly determines if the input define is either 1 or defined but empty.

include/oneapi/dpl/pstl/numeric_impl.h
adamfidel: Fixed.

danhoeflinger (dismissed):
LGTM, but I agree with the clang format suggestions, so I'd prefer to take those before merging.
Also, I still think it deserves a look from someone more familiar with upstream sync.

danhoeflinger (dismissed):

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3882,7 +3882,7 @@ _RandomAccessIterator
 __brick_min_element(_RandomAccessIterator __first, _RandomAccessIterator __last, _Compare __comp,
                     /* __is_vector = */ ::std::true_type) noexcept
 {
-#if (_PSTL_UDR_PRESENT || _ONEDPL_UDR_PRESENT)
```
danhoeflinger, line 3885: Just my opinion, but I prefer. 
`#if _ONEDPL_UDR_PRESENT  //_PSTL_UDR_PRESENT`

The `||` is our creation, I think upstream is just `#if _PSTL_UDR_PRESENT`

danhoeflinger (approved):
LGTM

PR #1553, opened 2024-05-01
[Tests] Adjusting std lib macros to lazily evaluation

danhoeflinger:
Following the lead of https://github.com/oneapi-src/oneDPL/pull/1543, we have a similar issue in another place.

This PR adjusts the standard library implementation specific macros withing `test_config.h` to be lazily evaluated later when they are used, and we can be confident that a standard library header will have been included.  

During `test_config.h`, generally this is the first include, and we do not have knowledge of the standard library, so `#if` checks in macro definitions here can lead to mismatched results to if they are lazily evaluated.  

For `transform2_ranges_factory_sycl.pass`, this can cause compiler errors in situations where `_GLIBCXX_RELEASE < 8`.

I've also added comments to be more explicit about this to be able to avoid similar errors in the future.

In the serial backends, we see compiler errors when including some standard library header from `test_config.h` to populate these macros, instead of relying upon this lazy evaluation. This should be investigated, but for now lets apply this workaround.

### DIFF
--- a/test/support/test_config.h
+++ b/test/support/test_config.h
@@ -52,6 +52,8 @@
 // GCC10 produces wrong answer calling exclusive_scan using vectorized polices
 #define TEST_GCC10_EXCLUSIVE_SCAN_BROKEN (_GLIBCXX_RELEASE == 10)
 // GCC7 std::get doesn't return const rvalue reference from const rvalue reference of tuple
+// Defining this all on one line to delay evaluation of `_GLIBCXX_RELEASE` until later, when
+// _PSTL_TEST_GCC7_RVALUE_TUPLE_GET_BROKEN is being used, and we can be confident that std lib macros are defined
 #define _PSTL_TEST_GCC7_RVALUE_TUPLE_GET_BROKEN (_GLIBCXX_RELEASE > 0 && _GLIBCXX_RELEASE < 8)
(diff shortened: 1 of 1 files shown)


SergeyKopienko:
I think alternative approach available too: https://github.com/oneapi-src/oneDPL/pull/1548
My assumption that better to include all required headers to have all required and used definition.
Because lazy define initialization you correctly rewrite but we able always to have them again in our tests.

danhoeflinger:
> I think alternative approach available too: #1548 My assumption that better to include all required headers to have all required and used definition. Because lazy define initialization you correctly rewrite but we able always to have them again in our tests.

I also don't understand why the original approach we were pursuing with including the STL headers in test_config breaks the serial backend #1555.  This is the way I'd like to fix it, but we must address the compilation errors with the failing CI.  It is way fewer changes than #1548.

SergeyKopienko:
> > I think alternative approach available too: #1548 My assumption that better to include all required headers to have all required and used definition. Because lazy define initialization you correctly rewrite but we able always to have them again in our tests.
> 
> I also don't understand why the original approach we were pursuing with including the STL headers in test_config breaks the serial backend #1555. This is the way I'd like to fix it, but we must address the compilation errors with the failing CI. It is way fewer changes than #1548.

OK, up to you.

SergeyKopienko (approved):
LGTM

PR #1554, opened 2024-05-01
[Dynamic Selection] Added volatile to tests to prevent optimization for release builds

AnuyaWelling2801:


timmiesmith (approved):
LGTM

danhoeflinger (approved):
After offline discussion, I think adding volatile here to convince the compiler to not optimize out the "long kernel" makes sense.
LGTM.

A couple extra endlines, if you wanna clear I'll reapprove.

SergeyKopienko:
@danhoeflinger, could you please explain for me, why exactly in one this test `volatile` required?
Do you think that without `volatile` we will use `dt_helper_v` inside Kernel after compiler optimizations? I think it's simple not usable there...

PR #1559, opened 2024-05-03
Documentation for scan kernel template

adamfidel:
This PR adds documentation for the scan kernel template.

### DIFF
--- a/documentation/library_guide/kernel_templates/single_pass_scan.rst
+++ b/documentation/library_guide/kernel_templates/single_pass_scan.rst
@@ -51,6 +51,9 @@ Parameters
 |                                               |   <use-buffer-wrappers>` (1).                                       |
 |                                               |                                                                     |
 +-----------------------------------------------+---------------------------------------------------------------------+
+|  ``binary_op``                                | A function object that will applied to the elements of the input    |
+|                                               |                                                                     |
++-----------------------------------------------+---------------------------------------------------------------------+
 |  ``param``                                    | A :doc:`kernel_param <../kernel_configuration>` object.             |
 |                                               |                                                                     |
 +-----------------------------------------------+---------------------------------------------------------------------+
@@ -60,6 +63,9 @@ Parameters
 
 - The element type of sequence(s) to scan must be a C++ integral or floating-point type
   other than ``bool`` with a width of up to 64 bits.
+- The binary operator must have a known identity. That is, ``sycl::has_known_identity<Op>::value`` must
+  evaluate to true.
+- The result is non-deterministic if the binary operator is non-associative (such as in floating-point addition).
 
 .. note::
 
--- a/documentation/library_guide/kernel_templates/single_pass_scan.rst
+++ b/documentation/library_guide/kernel_templates/single_pass_scan.rst
@@ -1,5 +1,5 @@
 Single-Pass Scan
-##########
+################
 
 -----------------------------------------------------------
 ``inclusive_scan`` Function Template
@@ -8,6 +8,10 @@ Single-Pass Scan
 The ``inclusive_scan`` function computes the inclusive prefix sum using a given binary operation.
 The function implements a single-pass algorithm, where each input element is read exactly once from
 global memory and each output element is written to exactly once in global memory.
+
+As this algorithm uses inter-work-group communication, it must be run on a device that supports
+strong forward progress guarantees, such as the `Intel® Data Center GPU Max
+<https://www.intel.com/content/www/us/en/products/details/discrete-gpus/data-center-gpu/max-series/products.html>`_.
 
 A synopsis of the ``inclusive_scan`` function is provided below:
 
@@ -63,16 +67,16 @@ Parameters
 
 - The element type of sequence to scan must be a C++ integral or floating-point type
   other than ``bool`` with a width of up to 64 bits.
-- The binary operator must have a known identity. That is, ``sycl::has_known_identity<Op>::value`` must
-  evaluate to true.
 - The result is non-deterministic if the binary operator is non-associative (such as in floating-point addition).
 
-.. note::
 
-   Current limitations:
+**Current limitations**:
 
-   - The function will internally block until the issued kernels have completed execution
-   - The SYCL device associated with the provided queue must support 64-bit atomic operations
+- The function will internally block until the issued kernels have completed execution.
+  Although intended in the future to be an asynchronous call, the algorithm is currently synchronous.
+- The SYCL device associated with the provided queue must support 64-bit atomic operations
+- The binary operator must have a known identity. That is, ``sycl::has_known_identity<Op>::value`` must
+  evaluate to true. Such operators are listed in the `SYCL 2020 specification <https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.identities>`_.
 
 Return Value
 ------------
@@ -154,7 +158,7 @@ where V is the number of bytes needed to store the input value type and F is the
 Currently, F is hard-coded to 4 as the flag type is 32-bits.
 
 The value of N\ :sub:`flags` represents the number of work-groups and depends on ``param.data_per_workitem`` and ``param.workgroup_size``.
-It can be approxmiated by dividing the number of input elements N by the product of ``param.data_per_workitem`` and ``param.workgroup_size``
+It can be approximated by dividing the number of input elements N by the product of ``param.data_per_workitem`` and ``param.workgroup_size``
 and adding 33 for padding.
 
 .. note::
@@ -182,14 +186,15 @@ The general advice is to choose kernel parameters based on performance measureme
 The initial configuration may be selected according to these high-level guidelines:
 
 
-- When the number of elements to scan (N) is small (~8K or less),
-  generally it is more efficient to process the algorithm by a single work-group.
-  In this case, the choice of kernel parameters is irrelevant, as we internally dispatch to a single
-  work-group implementation.
+- When the number of elements is small enough to fit within single work-group, the algorithm will ignore kernels
+  parameters and instead dispatch to a single workgroup version, where it is generally more efficient.
 
 - Generally, utilizing all available
-  compute cores is key for better performance. Allow creating enough work chunks to feed all
-  X\ :sup:`e`-cores [#fnote1]_ on a GPU: ``param.data_per_workitem * param.workgroup_size ≈ N / xe_core_count``.
+  compute cores is key for better performance. To allow sufficient work to satisfy all
+  X\ :sup:`e`-cores [#fnote1]_ on a GPU, use ``param.data_per_workitem * param.workgroup_size ≈ N / xe_core_count``.
+
+- On devices with mutiple tiles, it may prove beneficial to experiment with different tile hierarchies as described
+  in https://www.intel.com/content/www/us/en/developer/articles/technical/flattening-gpu-tile-hierarchy.html.
 
 
 .. warning::
--- a/documentation/library_guide/kernel_templates/single_pass_scan.rst
+++ b/documentation/library_guide/kernel_templates/single_pass_scan.rst
@@ -172,7 +172,7 @@ Local memory is used for storing elements of the input that are to be scanned by
 The used amount is denoted as N\ :sub:`elems_per_workgroup`, which equals to ``sizeof(key_type) * param.data_per_workitem * param.workgroup_size``.
 
 Some amount of local memory is also used by the calls to SYCL's group reduction and group scan. The amount of memory used particularly
-for these calls is implementation dependant.
+for these calls is implementation dependent.
 
 -----------------------------------------
 Recommended Settings for Best Performance
--- a/documentation/library_guide/kernel_templates/single_pass_scan.rst
+++ b/documentation/library_guide/kernel_templates/single_pass_scan.rst
@@ -2,7 +2,7 @@ Single-Pass Scan
 ################
 
 -----------------------------------------------------------
-``inclusive_scan`` Function Template
+inclusive_scan Function Template
 -----------------------------------------------------------
 
 The ``inclusive_scan`` function computes the inclusive prefix sum using a given binary operation.
@@ -10,7 +10,7 @@ The function implements a single-pass algorithm, where each input element is rea
 global memory and each output element is written to exactly once in global memory.
 
 As this algorithm uses inter-work-group communication, it must be run on a device that supports
-strong forward progress guarantees, such as the `Intel® Data Center GPU Max
+strong forward progress guarantees, such as the `Intel® Data Center GPU Max Series
 <https://www.intel.com/content/www/us/en/products/details/discrete-gpus/data-center-gpu/max-series/products.html>`_.
 
 A synopsis of the ``inclusive_scan`` function is provided below:
@@ -40,9 +40,9 @@ Parameters
 +-----------------------------------------------+---------------------------------------------------------------------+
 | Name                                          | Description                                                         |
 +===============================================+=====================================================================+
-|  ``q``                                        | SYCL* queue to submit the kernels to.                               |
+|  ``q``                                        | The SYCL* queue where kernels are submitted.                        |
 +-----------------------------------------------+---------------------------------------------------------------------+
-|                                               | The sequences of elements to apply the algorithm to.                |
+|                                               | The sequences to apply the algorithm to.                            |
 |  - ``in_begin``, ``in_end``, ``out_begin`` (1)| Supported sequence types:                                           |
 |  - ``in_rng``, ``out_rng`` (2)                |                                                                     |
 |                                               | - ``sycl::buffer`` (2),                                             |
@@ -55,7 +55,7 @@ Parameters
 |                                               |   <use-buffer-wrappers>` (1).                                       |
 |                                               |                                                                     |
 +-----------------------------------------------+---------------------------------------------------------------------+
-|  ``binary_op``                                | A function object that will applied to the elements of the input    |
+|  ``binary_op``                                | A function object that is applied to the elements of the input.     |
 |                                               |                                                                     |
 +-----------------------------------------------+---------------------------------------------------------------------+
 |  ``param``                                    | A :doc:`kernel_param <../kernel_configuration>` object.             |
@@ -70,7 +70,7 @@ Parameters
 - The result is non-deterministic if the binary operator is non-associative (such as in floating-point addition).
 
 
-**Current limitations**:
+**Current Limitations**:
 
 - The function will internally block until the issued kernels have completed execution.
   Although intended in the future to be an asynchronous call, the algorithm is currently synchronous.
@@ -88,7 +88,7 @@ Usage Examples
 --------------
 
 
-``inclusive_scan`` Example
+inclusive_scan Example
 ----------------------
 
 .. code:: cpp
@@ -193,8 +193,8 @@ The initial configuration may be selected according to these high-level guidelin
   compute cores is key for better performance. To allow sufficient work to satisfy all
   X\ :sup:`e`-cores [#fnote1]_ on a GPU, use ``param.data_per_workitem * param.workgroup_size ≈ N / xe_core_count``.
 
-- On devices with mutiple tiles, it may prove beneficial to experiment with different tile hierarchies as described
-  in https://www.intel.com/content/www/us/en/developer/articles/technical/flattening-gpu-tile-hierarchy.html.
+- On devices with multiple tiles, it may prove beneficial to experiment with different tile hierarchies as described
+  in `Options for using a GPU Tile Hierarchy <https://www.intel.com/content/www/us/en/developer/articles/technical/flattening-gpu-tile-hierarchy.html>`_.
 
 
 .. warning::
--- a/documentation/library_guide/kernel_templates/single_pass_scan.rst
+++ b/documentation/library_guide/kernel_templates/single_pass_scan.rst
@@ -61,7 +61,7 @@ Parameters
 
 **Type Requirements**:
 
-- The element type of sequence(s) to scan must be a C++ integral or floating-point type
+- The element type of sequence to scan must be a C++ integral or floating-point type
   other than ``bool`` with a width of up to 64 bits.
 - The binary operator must have a known identity. That is, ``sycl::has_known_identity<Op>::value`` must
   evaluate to true.
--- a/documentation/library_guide/kernel_templates/single_pass_scan.rst
+++ b/documentation/library_guide/kernel_templates/single_pass_scan.rst
@@ -1,0 +1,197 @@
+Single-Pass Scan
+##########
+
+-----------------------------------------------------------
+``inclusive_scan`` Function Template
+-----------------------------------------------------------
+
+The ``inclusive_scan`` function computes the inclusive prefix sum using a given binary operation.
+The function implements a single-pass algorithm, where each input element is read exactly once from
+global memory and each output element is written to exactly once in global memory.
+
+A synopsis of the ``inclusive_scan`` function is provided below:
+
+.. code:: cpp
+
+   // defined in <oneapi/dpl/experimental/kernel_templates>
+
+   namespace oneapi::dpl::experimental::kt::gpu {
+
+   template <typename InIterator, typename OutIterator, typename BinaryOp>
+   sycl::event
+   inclusive_scan (sycl::queue q, InIterator in_begin, InIterator in_end, OutIterator out_begin,
+                   BinaryOp binary_op, KernelParam param); // (1)
+
+   template <typename InRng, typename OutRng, typename BinaryOp>
+   sycl::event
+   inclusive_scan (sycl::queue q, InRng in_rng, OutRng out_rng, BinaryOp binary_op,
+                   KernelParam param) // (2)
+
+   }
+
+
+Parameters
+----------
+
++-----------------------------------------------+---------------------------------------------------------------------+
+| Name                                          | Description                                                         |
++===============================================+=====================================================================+
+|  ``q``                                        | SYCL* queue to submit the kernels to.                               |
++-----------------------------------------------+---------------------------------------------------------------------+
+|                                               | The sequences of elements to apply the algorithm to.                |
+|  - ``in_begin``, ``in_end``, ``out_begin`` (1)| Supported sequence types:                                           |
+|  - ``in_rng``, ``out_rng`` (2)                |                                                                     |
+|                                               | - ``sycl::buffer`` (2),                                             |
+|                                               | - :ref:`oneapi::dpl::experimental::ranges::views::all               |
+|                                               |   <viewable-ranges>` (2),                                           |
+|                                               | - :ref:`oneapi::dpl::experimental::ranges::views::subrange          |
+|                                               |   <viewable-ranges>` (2),                                           |
+|                                               | - :ref:`USM pointers <use-usm>` (1),                                |
+|                                               | - :ref:`oneapi::dpl::begin and oneapi::dpl::end                     |
+|                                               |   <use-buffer-wrappers>` (1).                                       |
+|                                               |                                                                     |
++-----------------------------------------------+---------------------------------------------------------------------+
+|  ``param``                                    | A :doc:`kernel_param <../kernel_configuration>` object.             |
+|                                               |                                                                     |
++-----------------------------------------------+---------------------------------------------------------------------+
+
+
+**Type Requirements**:
+
+- The element type of sequence(s) to scan must be a C++ integral or floating-point type
+  other than ``bool`` with a width of up to 64 bits.
+
+.. note::
+
+   Current limitations:
+
+   - The function will internally block until the issued kernels have completed execution
+   - The SYCL device associated with the provided queue must support 64-bit atomic operations
+
+Return Value
+------------
+
+A ``sycl::event`` object representing the status of the algorithm execution.
+
+--------------
+Usage Examples
+--------------
+
+
+``inclusive_scan`` Example
+----------------------
+
+.. code:: cpp
+
+   // possible build and run commands:
+   //    icpx -fsycl inclusive_scan.cpp -o inclusive_scan -I /path/to/oneDPL/include && ./inclusive_scan
+
+   #include <cstdint>
+   #include <iostream>
+   #include <sycl/sycl.hpp>
+
+   #include <oneapi/dpl/experimental/kernel_templates>
+
+   namespace kt = oneapi::dpl::experimental::kt;
+
+   int main()
+   {
+      std::size_t n = 6;
+      sycl::queue q{sycl::gpu_selector_v};
+      std::uint32_t* arr = sycl::malloc_shared<std::uint32_t>(n, q);
+      std::uint32_t* out = sycl::malloc_shared<std::uint32_t>(n, q);
+
+      // initialize
+      arr[0] = 1, arr[1] = 2, arr[2] = 1, arr[3] = 3, arr[4] = 1, arr[5] = 2;
+
+      // scan
+      auto e = kt::gpu::inclusive_scan(q, arr, arr + n, out, std::plus<std::uint32_t>{}, kt::kernel_param<256, 8>{});
+      e.wait();
+
+      // print
+      for(std::size_t i = 0; i < n; ++i)
+         std::cout << out[i] << ' ';
+      std::cout << '\n';
+
+      sycl::free(arr, q);
+      sycl::free(out, q);
+      return 0;
+   }
+
+**Output:**
+
+.. code:: none
+
+   1 3 4 7 8 10
+
+.. _memory-requirements:
+
+-------------------
+Memory Requirements
+-------------------
+
+The algorithms use global and local device memory (see `SYCL 2020 Specification
+<https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_sycl_device_memory_model>`_)
+for intermediate data storage. For the algorithms to operate correctly, there must be enough memory
+on the device; otherwise, the behavior is undefined. The amount of memory that is required
+depends on input data and configuration parameters, as described below.
+
+Global Memory Requirements
+--------------------------
+
+Global memory is used for copying the input sequence and storing internal data such as status flags.
+The used amount depends on many parameters; below is an upper bound approximation:
+
+2 * V * N \ :sub:`flags` + F * N \ :sub:`flags`
+
+where V is the number of bytes needed to store the input value type and F is the number of bytes needed to store the flags.
+Currently, F is hard-coded to 4 as the flag type is 32-bits.
+
+The value of N\ :sub:`flags` represents the number of work-groups and depends on ``param.data_per_workitem`` and ``param.workgroup_size``.
+It can be approxmiated by dividing the number of input elements N by the product of ``param.data_per_workitem`` and ``param.workgroup_size``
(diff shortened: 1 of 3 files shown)


documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+Recommended Settings for Best Performance
+-----------------------------------------
+
+The general advice is to choose kernel parameters based on performance measurements and profiling information.
+The initial configuration may be selected according to these high-level guidelines:
+
+
+- When the number of elements to scan (N) is small (~8K or less),
+  generally it is more efficient to process the algorithm by a single work-group.
+  In this case, the choice of kernel parameters is irrelevant, as we internally dispatch to a single
```
timmiesmith, line 187:
```suggestion
  In this case, the choice of kernel parameters is irrelevant, as the implementation will internally dispatch to a single
```

timmiesmith (commented):
One small suggestion. Otherwise it looks good to me.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+-----------------------------------------
+Recommended Settings for Best Performance
+-----------------------------------------
+
+The general advice is to choose kernel parameters based on performance measurements and profiling information.
+The initial configuration may be selected according to these high-level guidelines:
+
+
+- When the number of elements to scan (N) is small (~8K or less),
+  generally it is more efficient to process the algorithm by a single work-group.
```
danhoeflinger, line 186: Does this number (~8k) depend upon type, and hardware?  Can we say more about this?

Also, does the user have any control here about if a single workgroup algorithm is used?  Perhaps we may want to reword this to say something like:

"When the number of elements is small enough to fit within single work-group, the algorithm will ignore kernel parameters and instead dispatch to a single workgroup version, where it is generally more efficient."  

(Sidenote for the algorithm in general: we may want to consider if this fits within the mindset of kernel templates, or if we would instead want to provide a separate overload or merely documentation guidance about how to set kernel params to achieve a single workgroup impl)

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+  In this case, the choice of kernel parameters is irrelevant, as the implementation will internally dispatch to a single
+  work-group implementation.
+
+- Generally, utilizing all available
+  compute cores is key for better performance. Allow creating enough work chunks to feed all
+  X\ :sup:`e`-cores [#fnote1]_ on a GPU: ``param.data_per_workitem * param.workgroup_size ≈ N / xe_core_count``.
```
danhoeflinger, line 192: I think this needs a little re-wording:
"To allow sufficient work to satisfy all X_e cores. use ..." perhaps?

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
+Single-Pass Scan
```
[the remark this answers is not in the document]
adamfidel: Fixed.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+##########
+
+-----------------------------------------------------------
+``inclusive_scan`` Function Template
+-----------------------------------------------------------
+
+The ``inclusive_scan`` function computes the inclusive prefix sum using a given binary operation.
+The function implements a single-pass algorithm, where each input element is read exactly once from
+global memory and each output element is written to exactly once in global memory.
+
```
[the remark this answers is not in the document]
adamfidel: I added a few words here about forward progress guarantees and an example device that supports it.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+.. warning::
+
+   Avoid setting too large ``param.data_per_workitem`` and ``param.workgroup_size`` values.
+   Make sure that :ref:`Memory requirements <memory-requirements>` are satisfied.
+
+.. [#fnote1] The X\ :sup:`e`-core term is described in the `oneAPI GPU Optimization Guide
+   <https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2024-0/intel-xe-gpu-architecture.html#XE-CORE>`_.
+   Check the number of cores in the device specification, such as `Intel® Data Center GPU Max specification
+   <https://www.intel.com/content/www/us/en/products/details/discrete-gpus/data-center-gpu/max-series/products.html>`_.
```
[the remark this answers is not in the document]
adamfidel: I added a link to this article and a suggestion to utilize it when tuning for performance.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+
+- The element type of sequence to scan must be a C++ integral or floating-point type
+  other than ``bool`` with a width of up to 64 bits.
+- The binary operator must have a known identity. That is, ``sycl::has_known_identity<Op>::value`` must
+  evaluate to true.
+- The result is non-deterministic if the binary operator is non-associative (such as in floating-point addition).
+
+.. note::
+
+   Current limitations:
```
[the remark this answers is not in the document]
adamfidel: Yes, I changed it.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+
+**Type Requirements**:
+
+- The element type of sequence to scan must be a C++ integral or floating-point type
+  other than ``bool`` with a width of up to 64 bits.
+- The binary operator must have a known identity. That is, ``sycl::has_known_identity<Op>::value`` must
+  evaluate to true.
+- The result is non-deterministic if the binary operator is non-associative (such as in floating-point addition).
+
+.. note::
```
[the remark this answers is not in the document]
adamfidel: I removed the note directive.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+Global memory is used for copying the input sequence and storing internal data such as status flags.
+The used amount depends on many parameters; below is an upper bound approximation:
+
+2 * V * N \ :sub:`flags` + F * N \ :sub:`flags`
+
+where V is the number of bytes needed to store the input value type and F is the number of bytes needed to store the flags.
+Currently, F is hard-coded to 4 as the flag type is 32-bits.
+
+The value of N\ :sub:`flags` represents the number of work-groups and depends on ``param.data_per_workitem`` and ``param.workgroup_size``.
+It can be approxmiated by dividing the number of input elements N by the product of ``param.data_per_workitem`` and ``param.workgroup_size``
```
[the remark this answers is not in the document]
adamfidel: Changed, thanks.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+|                                               |                                                                     |
++-----------------------------------------------+---------------------------------------------------------------------+
+
+
+**Type Requirements**:
+
+- The element type of sequence to scan must be a C++ integral or floating-point type
+  other than ``bool`` with a width of up to 64 bits.
+- The binary operator must have a known identity. That is, ``sycl::has_known_identity<Op>::value`` must
+  evaluate to true.
```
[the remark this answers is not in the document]
adamfidel: I suppose this is more of a current limitation since we do plan to support other operators in the future. I moved it around and added a link to the table.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+- The result is non-deterministic if the binary operator is non-associative (such as in floating-point addition).
+
+.. note::
+
+   Current limitations:
+
+   - The function will internally block until the issued kernels have completed execution
```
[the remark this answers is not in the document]
adamfidel: I added another sentence about how this makes the algorithm synchronous instead of the intent of being asynchronous.

In the future, I'm not sure what the best way is to change the API to show that is now asynchronous. Perhaps it would be better if we change the algorithm now to return `void` and in the future, it will return a `sycl::event`?

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: I added the suggested statements. I think it can be argued to not include specific cutoffs (e.g., 8k) because they are dependent on the device's local memory size and is not exactly something that can be controlled currently.

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: I changed the wording to hopefully be a little clearer.

documentation/library_guide/kernel_templates/single_pass_scan.rst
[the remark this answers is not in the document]
danhoeflinger: This is probably good enough for now, and when we add more algorithms, we can maybe group these and say more about this.  Are there any good resources to allow users to understand what hardware does or doesn't provide these strong forward progress guarantees?

I know there is the proposed sycl extension, but until then, is there some resource we can point users to? maybe there isn't.

documentation/library_guide/kernel_templates/single_pass_scan.rst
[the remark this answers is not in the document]
danhoeflinger: Its an interesting question about a general mindset for KT, one probably worth discussing, but we can leave it alone in the context of this doc PR.
Thanks for the clarification in the text.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
+Single-Pass Scan
+################
+
+-----------------------------------------------------------
+``inclusive_scan`` Function Template
```
dcbenito, line 5: We don't use monospace font in headers, please remove the ```` from inclusive_scan.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+-----------------------------------------------------------
+
+The ``inclusive_scan`` function computes the inclusive prefix sum using a given binary operation.
+The function implements a single-pass algorithm, where each input element is read exactly once from
+global memory and each output element is written to exactly once in global memory.
+
+As this algorithm uses inter-work-group communication, it must be run on a device that supports
+strong forward progress guarantees, such as the `Intel® Data Center GPU Max
```
dcbenito, line 13: Please cite the full name of the link: Intel® Data Center GPU Max Series

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+|                                               |   <viewable-ranges>` (2),                                           |
+|                                               | - :ref:`oneapi::dpl::experimental::ranges::views::subrange          |
+|                                               |   <viewable-ranges>` (2),                                           |
+|                                               | - :ref:`USM pointers <use-usm>` (1),                                |
+|                                               | - :ref:`oneapi::dpl::begin and oneapi::dpl::end                     |
+|                                               |   <use-buffer-wrappers>` (1).                                       |
+|                                               |                                                                     |
++-----------------------------------------------+---------------------------------------------------------------------+
+|  ``binary_op``                                | A function object that will applied to the elements of the input    |
```
dcbenito, line 58: Please change to:  A function object that is applied to the elements of the input.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+   }
+
+
+Parameters
+----------
+
++-----------------------------------------------+---------------------------------------------------------------------+
+| Name                                          | Description                                                         |
++===============================================+=====================================================================+
+|  ``q``                                        | SYCL* queue to submit the kernels to.                               |
```
dcbenito, line 43: Please change to: The SYCL* queue where kernels are submitted.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
++-----------------------------------------------+---------------------------------------------------------------------+
+|                                               | The sequences of elements to apply the algorithm to.                |
```
dcbenito, line 45: Please change to: The algorithm where the sequences of elements are applied.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+
+
+**Type Requirements**:
+
+- The element type of sequence to scan must be a C++ integral or floating-point type
+  other than ``bool`` with a width of up to 64 bits.
+- The result is non-deterministic if the binary operator is non-associative (such as in floating-point addition).
+
+
+**Current limitations**:
```
dcbenito, line 73: Please capitalize limitations.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+
+A ``sycl::event`` object representing the status of the algorithm execution.
+
+--------------
+Usage Examples
+--------------
+
+
+``inclusive_scan`` Example
+----------------------
```
dcbenito, line 92: We don't use monospace font in headers, please remove the ```` from inclusive_scan. The header underline needs to go the full length of the header, otherwise it renders incorrectly. Please fix the length.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+
+- When the number of elements is small enough to fit within single work-group, the algorithm will ignore kernel
+  parameters and instead dispatch to a single workgroup version, where it is generally more efficient.
+
+- Generally, utilizing all available
+  compute cores is key for better performance. To allow sufficient work to satisfy all
+  X\ :sup:`e`-cores [#fnote1]_ on a GPU, use ``param.data_per_workitem * param.workgroup_size ≈ N / xe_core_count``.
+
+- On devices with mutiple tiles, it may prove beneficial to experiment with different tile hierarchies as described
+  in https://www.intel.com/content/www/us/en/developer/articles/technical/flattening-gpu-tile-hierarchy.html.
```
dcbenito, line 197: Please change this to a live link: `Options for using a GPU Tile Hierarchy <https://www.intel.com/content/www/us/en/developer/articles/technical/flattening-gpu-tile-hierarchy.html>`

dcbenito (changes requested):
Minor changes requested, otherwise, LGTM.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+inclusive_scan Function Template
+-----------------------------------------------------------
+
+The ``inclusive_scan`` function computes the inclusive prefix sum using a given binary operation.
+The function implements a single-pass algorithm, where each input element is read exactly once from
+global memory and each output element is written to exactly once in global memory.
+
+As this algorithm uses inter-work-group communication, it must be run on a device that supports
+strong forward progress guarantees, such as the `Intel® Data Center GPU Max Series
+<https://www.intel.com/content/www/us/en/products/details/discrete-gpus/data-center-gpu/max-series/products.html>`_.
```
dmitriy-sobolev, line 14: Forward progress guarantees is a complex topic for GPUs. We should be careful while doing statements about it.

As far as I know (not totally sure), Data Center GPU Max guarantees at least parallel forward progress between work-groups for  [cooperative launches](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_root_group.asciidoc). Normal launches provide weakly parallel forward guarantee (essentially, no guarantees).
At this point, I suppose, one should thoroughly investigate GPU architecture specifics to understand whether this synchronization pattern would work or not. That's why I recommend simply specifying the list of the supported GPUs.

danhoeflinger: This makes sense.  I was hoping there was some better existing reference to point to, but perhaps its best to just list "known" supported GPUs for now.

dmitriy-sobolev: Unfortunately, it's like folklore. At least, I am not aware of such references.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,203 @@
...
+
+   template <typename InIterator, typename OutIterator, typename BinaryOp>
+   sycl::event
+   inclusive_scan (sycl::queue q, InIterator in_begin, InIterator in_end, OutIterator out_begin,
+                   BinaryOp binary_op, KernelParam param); // (1)
+
+   template <typename InRng, typename OutRng, typename BinaryOp>
+   sycl::event
+   inclusive_scan (sycl::queue q, InRng in_rng, OutRng out_rng, BinaryOp binary_op,
+                   KernelParam param) // (2)
```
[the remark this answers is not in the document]
dmitriy-sobolev: I will swap the APIs for the radix sort.

documentation/library_guide/kernel_templates/single_pass_scan.rst
dcbenito: LGTM.

dcbenito (dismissed):

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
++-----------------------------------------------+---------------------------------------------------------------------+
+|  ``param``                                    | A :doc:`kernel_param <../kernel_configuration>` object.             |
+|                                               |                                                                     |
++-----------------------------------------------+---------------------------------------------------------------------+
+
+
+**Type Requirements**:
+
+- The element type of sequence to scan must be a C++ integral or floating-point type
+  other than ``bool`` with a width of up to 64 bits.
```
dmitriy-sobolev, line 69: Is it correct? As I recall we test only 32-bit and 64-bit arithmetic types.

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: Would it be sufficient to remove the phrase "must be run on a device that supports strong forward progress guarantees" and simply state that the algorithm has been designed specifically to run on Intel® Data Center GPU Max Series?

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+- The result is non-deterministic if the binary operator is non-associative (such as in floating-point addition).
```
dmitriy-sobolev, line 70: I suppose there should be commutativity as well.

The implementation uses reduction group operation, which requires both commutativity and associativity:
https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#:~:text=The%20result%20of%20a%20call%20to%20these%20functions%20is%20non%2Ddeterministic%20if%20the%20binary%20operator%20is%20not%20commutative%20and%20associative.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+
+
+**Current Limitations**:
+
+- The function will internally block until the issued kernels have completed execution.
+  Although intended in the future to be an asynchronous call, the algorithm is currently synchronous.
+- The SYCL device associated with the provided queue must support 64-bit atomic operations
```
dmitriy-sobolev, line 77: Is it true when we scan 32-bit elements?

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+
+
+**Type Requirements**:
+
+- The element type of sequence to scan must be a C++ integral or floating-point type
+  other than ``bool`` with a width of up to 64 bits.
+- The result is non-deterministic if the binary operator is non-associative (such as in floating-point addition).
+
+
+**Current Limitations**:
```
dmitriy-sobolev, line 73: Could you turn that section into a note? It would stylistically highlight that it's a temporary thing and will align with radix sort page.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+.. _memory-requirements:
+
+-------------------
+Memory Requirements
+-------------------
+
+The algorithms use global and local device memory (see `SYCL 2020 Specification
+<https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_sycl_device_memory_model>`_)
+for intermediate data storage. For the algorithms to operate correctly, there must be enough memory
+on the device; otherwise, the behavior is undefined. The amount of memory that is required
```
dmitriy-sobolev, line 146: Contrary to radix sort, scan throws `bad_alloc` (I will do the same for radix sort), so it is not undefined behavior. Could you adjust the description?

dmitriy-sobolev (commented):
Here are some my input. I am still looking at the PR.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
+Single-Pass Scan
```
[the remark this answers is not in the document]
adamfidel: I'm not sure how much value there is to having a new page for GPU kernel templates vs just linking to this document in the top-level `kernel_templates_main.rst` page. The ESIMD page includes information about ESIMD itself, but I don't know if there is much to say in a generic GPU kernel template page that is not already said in `kernel_templates_main.rst`.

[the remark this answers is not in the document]
adamfidel: Good point. I renamed it to "Inclusive Scan."

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,208 @@
...
+################
+
+-----------------------------------------------------------
+inclusive_scan Function Template
+-----------------------------------------------------------
```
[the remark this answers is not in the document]
adamfidel: Fixed.

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: You are right; we do not test 16-bit or 8-bit types so I updated this to make it clearer.

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: Thanks, added.

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: I reworded this to clarify that this only holds when the element type is 64-bits.

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: It was a note, but a previous reviewer suggested to remove it as a note. I suppose it makes sense to align it with the radix sort page so I turned it back into a note.

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: Fixed.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,210 @@
...
+|                                               |   <use-buffer-wrappers>` (1).                                       |
+|                                               |                                                                     |
++-----------------------------------------------+---------------------------------------------------------------------+
+|  ``binary_op``                                | A function object that is applied to the elements of the input.     |
+|                                               |                                                                     |
++-----------------------------------------------+---------------------------------------------------------------------+
+|  ``param``                                    | A :doc:`kernel_param <../kernel_configuration>` object.             |
+|                                               |                                                                     |
++-----------------------------------------------+---------------------------------------------------------------------+
```
dmitriy-sobolev, line 63: That should fix rendering of the table. I've also swapped (1) and (2) on the column on the right.

```suggestion
+------------------------------------------------+---------------------------------------------------------------------+
| Name                                           | Description                                                         |
+================================================+=====================================================================+
| ``q``                                          | The SYCL* queue where kernels are submitted.                        |
+------------------------------------------------+---------------------------------------------------------------------+
|                                                |                                                                     |
|                                                | The sequences to apply the algorithm to.                            |
| - ``in_begin``, ``in_end``, ``out_begin`` (1), | Supported sequence types:                                           |
| - ``in_rng``, ``out_rng`` (2).                 |                                                                     |
|                                                | - :ref:`USM pointers <use-usm>` (1),                                |
|                                                | - :ref:`oneapi::dpl::begin and oneapi::dpl::end                     |
|                                                |   <use-buffer-wrappers>` (1),                                       |
|                                                | - ``sycl::buffer`` (2),                                             |
|                                                | - :ref:`oneapi::dpl::experimental::ranges::views::subrange          |
|                                                | - :ref:`USM pointers <use-usm>` (1),                                |
|                                                | - :ref:`oneapi::dpl::begin and oneapi::dpl::end                     |
|                                                |   <use-buffer-wrappers>` (1).                                       |
|                                                |                                                                     |
+------------------------------------------------+---------------------------------------------------------------------+
| ``binary_op``                                  | A function object that is applied to the elements of the input.     |
|                                                |                                                                     |
+------------------------------------------------+---------------------------------------------------------------------+
| ``param``                                      | A :doc:`kernel_param <../kernel_configuration>` object.             |
|                                                |                                                                     |
+------------------------------------------------+---------------------------------------------------------------------+
```

dmitriy-sobolev, line 63: That should fix rendering of the table. I've also swapped (1) and (2) on the column of the right. 

```suggestion
+------------------------------------------------+---------------------------------------------------------------------+
| Name                                           | Description                                                         |
+================================================+=====================================================================+
| ``q``                                          | The SYCL* queue where kernels are submitted.                        |
+------------------------------------------------+---------------------------------------------------------------------+
|                                                |                                                                     |
|                                                | The sequences to apply the algorithm to.                            |
| - ``in_begin``, ``in_end``, ``out_begin`` (1), | Supported sequence types:                                           |
| - ``in_rng``, ``out_rng`` (2).                 |                                                                     |
|                                                | - :ref:`USM pointers <use-usm>` (1),                                |
|                                                | - :ref:`oneapi::dpl::begin and oneapi::dpl::end                     |
|                                                |   <use-buffer-wrappers>` (1),                                       |
|                                                | - ``sycl::buffer`` (2),                                             |
|                                                | - :ref:`oneapi::dpl::experimental::ranges::views::all               |
|                                                |   <viewable-ranges>` (2),                                           |
|                                                | - :ref:`oneapi::dpl::experimental::ranges::views::subrange          |
|                                                |   <viewable-ranges>` (2).                                           |
|                                                |                                                                     |
+------------------------------------------------+---------------------------------------------------------------------+
| ``binary_op``                                  | A function object that is applied to the elements of the input.     |
|                                                |                                                                     |
+------------------------------------------------+---------------------------------------------------------------------+
| ``param``                                      | A :doc:`kernel_param <../kernel_configuration>` object.             |
|                                                |                                                                     |
+------------------------------------------------+---------------------------------------------------------------------+
```

documentation/library_guide/kernel_templates/single_pass_scan.rst
dmitriy-sobolev: I've been thinking about the previous Dan's suggestion and the flexibility of it is a good thing.
Adam, what do you think about that slight correction to your note, which avoids making that general statement about the GPU?

I also mentioned device USM as it is an optional SYCL feature. 

>The algorithm is designed to be compatible with a variety of devices that provide at least parallel forward guarantees between work-groups, due to cross-work-group communication. Additionally, it requires support for device USM. It has been verified to be compatible with Intel® Data Center GPU Max Series.

adamfidel: I like your wording a lot. I updated the paragraph accordingly.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,213 @@
...
+
+.. note::
+
+  Current limitations:
+
+  - The function will internally block until the issued kernels have completed execution.
+    Although intended in the future to be an asynchronous call, the algorithm is currently synchronous.
+  - The SYCL device associated with the provided queue must support 64-bit atomic operations if the element type is 64-bits.
+  - The binary operator must have a known identity. That is, ``sycl::has_known_identity<Op>::value`` must
+    evaluate to true. Such operators are listed in the `SYCL 2020 specification <https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.identities>`_.
```
dmitriy-sobolev, line 84: `has_known_identity` requires accumulator type and operation type. Let's adjust the description accordingly: 

```suggestion
  - There must be a known identity value for the provided combination of the element type and the binary operation. That is, ``sycl::has_known_identity_v`` must evaluate to true. Such operators are listed in the `SYCL 2020 specification <https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.identities>`_.
```

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,213 @@
...
+
+-------------------
+Memory Requirements
+-------------------
+
+The algorithms use global and local device memory (see `SYCL 2020 Specification
+<https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_sycl_device_memory_model>`_)
+for intermediate data storage. For the algorithms to operate correctly, there must be enough memory
+on the device; otherwise, the algorithm throws a ``std::bad_alloc`` exception. The amount of memory that is required
+depends on input data and configuration parameters, as described below.
```
dmitriy-sobolev, line 152: Suggestion to highlight that std::bad_alloc is applicable only to the global device memory. Plus, some minor corrections. 

```suggestion
The algorithm uses global and local device memory (see `SYCL 2020 Specification
<https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_sycl_device_memory_model>`_)
for intermediate data storage. For the algorithm to operate correctly, there must be enough memory
on the device. It throws a ``std::bad_alloc`` exception if there is not enough global device memory. The behavior is undefined if there is not enough local memory. The amount of memory that is required
depends on input data and configuration parameters, as described below.
```

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,213 @@
...
+
+Global Memory Requirements
```
dmitriy-sobolev, line 154: I suppose there is no need in a such elaborate excursion into the global memory requirements as we have ``bad_alloc`` for the user as a tool to handle the corresponding issue.

What about limiting the description of the whole section to: 
> The amount of the internally allocated memory does not exceed the amount needed to store the input sequence and usually it is a small fracture of it.

documentation/library_guide/kernel_templates_main.rst
```diff
@@ -15,6 +15,7 @@ The primary API namespace is ``oneapi::dpl::experimental::kt``, and nested names
 
 * :doc:`Kernel Configuration <kernel_templates/kernel_configuration>`. Generic structure for configuring a kernel template.
 * :doc:`ESIMD-based kernel templates <kernel_templates/esimd_main>`. Algorithms implemented with the Explicit SIMD SYCL extension.
+* :doc:`Scan kernel template algorithm <kernel_templates/single_pass_scan>`. Scan algorithm using a single-pass approach.
```
dmitriy-sobolev, line 18:
```suggestion
* :doc:`Inclusive scan algorithm <kernel_templates/single_pass_scan>`. Inclusive scan kernel template algorithm using a single-pass approach.
```

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: I believe there is still some value in informing the user about how much global memory is required. I agree that throwing `bad_alloc` will tell the user that their current configuration cannot be satisfied, but it will probably still be helpful to let users know what type of global memory usage to expect and how that usage depends on their kernel parameters.

dmitriy-sobolev: I still believe that it is too detailed that may confuse a user. One would need to understand that section before starting using that functionality and it is not simple.

What do you think about removing "upper bound" and "adding 33 for padding" parts to reduce verbosity?

dmitriy-sobolev: Or alternatively, accompany the verbose estimation with a rough one.

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,212 @@
...
+   If the number of input elements can be efficiently processed by a single work-group,
+   the kernel template is executed by a single work-group and does not use any global memory.
+
+
+.. _local-memory:
+
+Local Memory Requirements
+-------------------------
+
+Local memory is used for storing elements of the input that are to be scanned by a single work group.
```
dmitriy-sobolev, line 179:
```suggestion
Local memory is used for storing elements of the input that are to be scanned by a single work-group.
```

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -0,0 +1,212 @@
...
+-----------------------------------------
+Recommended Settings for Best Performance
+-----------------------------------------
+
+The general advice is to choose kernel parameters based on performance measurements and profiling information.
+The initial configuration may be selected according to these high-level guidelines:
+
+
+- When the number of elements is small enough to fit within single work-group, the algorithm will ignore kernel
+  parameters and instead dispatch to a single workgroup version, where it is generally more efficient.
```
dmitriy-sobolev, line 194:
```suggestion
  parameters and instead dispatch to a single work-group version, where it is generally more efficient.
```

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: I agree there was a lot of verbosity there. I removed a lot of the unnecessary details such as the padding and the variable for the flag length.

dmitriy-sobolev: It looks much simpler now, thanks

dmitriy-sobolev (approved):
LGTM

dmitriy-sobolev (dismissed):
Reapproving after minor corrections.

dmitriy-sobolev (approved):
Reapproving after fixing a typo.

PR #766, opened 2023-01-27
Testing workaround to avoid unsupported calls in kernel

danhoeflinger:
Temporary testing workaround to avoid currently unsupported calls to vector algorithms from inside sycl kernels when calling `find()` or initializer list overloads of `min()` and `max()` on Windows at runtime.

Define must be set prior to include of the standard library headers. There is no natural unified location, given the need to precede the standard library headers.  It seems best to add the workaround in each test individually.  This also provides more apparent documentation of the known limitation.

PR #1560, opened 2024-05-03
[Tests] Adding windows vector alg workaround for swap

danhoeflinger:
Similar to #766, swap and swap1 tests have issues with precompiled vector algorithms being used, and require a workaround for use in SYCL kernels on windows.


timmiesmith (approved):
LGTM. Thank you.

PR #1561, opened 2024-05-03
[Test] Replacing bad type to check for device_copyable

danhoeflinger:
It doesn't make sense to check a std::vector::iterator allocated on the host for device_copyability.  

We don't want to copy this to the device anyway.  A smarter check is to check with a pointer to usm shared memory.
Generally, this check passes anyway, however in some configurations on windows, std::vector:iterator is non device_copyable, resulting in a trigger for this static assertion.


timmiesmith (approved):
LGTM after offline discussion with @danhoeflinger

PR #1562, opened 2024-05-06
Fix fp64 aspect error in tests

SergeyKopienko:
In this PR we fix exceptions `Required aspect fp64 is not supported on the device` in three tests:
- `test/xpu_api/numerics/complex.number/cmplx.over/imag.pass.cpp`;
- `test/xpu_api/numerics/complex.number/cmplx.over/real.pass.cpp`;
- `test/xpu_api/numerics/c.math/cmath.pass.cpp`.

### DIFF
--- a/test/xpu_api/numerics/c.math/cmath.pass.cpp
+++ b/test/xpu_api/numerics/c.math/cmath.pass.cpp
@@ -101,7 +101,8 @@ void test_abs()
 #pragma clang diagnostic pop
 #endif
 
-    assert(dpl::abs(-1.) == 1);
+    assert(dpl::abs(-1.f) == 1.f);
+    IF_DOUBLE_SUPPORT(assert(dpl::abs(-1.) == 1));
 }
 
 ONEDPL_TEST_DECLARE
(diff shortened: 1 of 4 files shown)


test/xpu_api/numerics/c.math/cmath.pass.cpp
```diff
@@ -101,7 +101,8 @@ void test_abs()
 #pragma clang diagnostic pop
 #endif
 
-    assert(dpl::abs(-1.) == 1);
+    assert(dpl::abs(-1.f) == 1.f);
+    IF_DOUBLE_SUPPORT(assert(dpl::abs(-1.) == 1));
```
danhoeflinger, line 105: We check `-1.f` vs `1.f`, but then `1.` vs `1`.
I think either the first should compare to an integer or the second should compare to a double.  I don't quite understand which is the intended check here though.

SergeyKopienko: Fixed

danhoeflinger (approved):
LGTM

PR #1563, opened 2024-05-06
Skip minmax_element.pass with icx 2022.2

julianmi:
The `minmax_element.pass` test shows an internal compiler error in the DPCPP backend with the icx 2022.2 compiler. This patch skips the affected part of this test.


test/support/test_config.h
```diff
@@ -166,4 +166,11 @@
 // of the underlying sycl::device to determine whether to include or exclude 64-bit type tests.
 #define _PSTL_GROUP_REDUCTION_MULT_INT64_BROKEN 1
 
+// oneAPI DPC++ compiler in 2022.2 shows an internal compiler error during the backend code generation of minmax_element.pass.cpp
+#if TEST_DPCPP_BACKEND_PRESENT && (__INTEL_LLVM_COMPILER == 20220200 || __INTEL_LLVM_COMPILER == 20220201)
```
danhoeflinger, line 170: Does this only fail specifically 2022.2.0 and 2022.2.1? (what about 2022.1 and before?)

Would we be better off checking `(__INTEL_LLVM_COMPILER >= 20220200 &__INTEL_LLVM_COMPILER < 20220300)` to cover future possible hotfixes to 2022.2.x?  I'd expect they wouldn't fix this specific issue in a hotfix, but I'm not sure what the right approach is here.

julianmi: It looks like older compilers are affected as well. I've changed this to check for releases before 2022.3.

test/support/test_config.h
```diff
@@ -166,4 +166,8 @@
 // of the underlying sycl::device to determine whether to include or exclude 64-bit type tests.
 #define _PSTL_GROUP_REDUCTION_MULT_INT64_BROKEN 1
 
+// oneAPI DPC++ compiler 2022.2 an below show an internal compiler error during the backend code generation of
+// minmax_element.pass.cpp
+#define _PSTL_ICPX_TEST_MINMAX_ELEMENT_BROKEN (TEST_DPCPP_BACKEND_PRESENT && __INTEL_LLVM_COMPILER < 20220300)
+
```
danhoeflinger, line 172: What if we are using a different compiler and `__INTEL_LLVM_COMPILER` is undefined?

julianmi: I've added a check if `__INTEL_LLVM_COMPILER` is defined.

danhoeflinger (approved):
LGTM

PR #1568, opened 2024-05-07
[Tests] Skipping c++20 MSVC comparison checks on sycl

danhoeflinger:
MSVC STL implementation a consteval trick to allow comparisons with literal zero but not other integers which involves eliminated a dead code function call.  However, within a sycl kernel, it seems this dead function call still causes an error complaining that the undefined function call is not `SYCL_EXTERNAL`.

This was included in https://github.com/microsoft/STL/releases/tag/vs-2022-17.7, specifically this commit https://github.com/microsoft/STL/commit/581b4f089c303123071a173a16ca2f2260a3a891 which corresponds to `_MSVC_STL_UPDATE` `202303L`, `202304L`, `202305L`.  Therefore, we encounter issues on the device after that update when using C++20.

I've added a check for compiler version, and we will follow up with the compiler team to see future versions of intel llvm compiler can compile this code.


test/support/test_config.h
```diff
@@ -154,6 +154,9 @@
 #define _PSTL_TEST_COMPLEX_SINH_BROKEN _PSTL_TEST_COMPLEX_OP_BROKEN
 #define _PSTL_TEST_COMPLEX_TANH_BROKEN _PSTL_TEST_COMPLEX_OP_BROKEN
 
+#define _PSTL_TEST_COMPARISON_BROKEN                                                                                   \
```
mmichel11, line 157: Since we are defining this macro and are not first checking if `_MSVC_LANG` and other macros are present, then I believe this will cause a compile error on Linux due to the following usages:

`return TestUtils::done(!_PSTL_TEST_COMPARISON_BROKEN);`

danhoeflinger: Good point.  Why did this not trip our per commit CI...?

danhoeflinger: Ah, its a blind spot for our per commit CI because only the sycl_iterator tests run for dpcpp backend.
I'll make a fix.  Thanks.

danhoeflinger:
> Some additional information here: [microsoft/STL#4359](https://github.com/microsoft/STL/issues/4359) It's not a precompiled routine, it's a deliberately undefined routine which is named to give a hint to what's wrong with the code that does not link. The idea is that the type `_Literal_zero` has an implicit `consteval` constructor (see https://en.cppreference.com/w/cpp/language/consteval) which argument may only have the value of `0`. Due to consteval the only allowed context is constant expressions, so compilers can throw away the function call when the code is valid - but it looks like icpx does not do that for device code.

Aha, yes that makes sense now.  Thanks for pointing this out.  Seems to be an issue with `icx` not being able to eliminate the function call.  We will need to include a icx version number and follow up with the icx team.

test/xpu_api/containers/sequences/array/array.comparison/greater_g.pass.cpp
```diff
@@ -42,12 +43,18 @@ kernel_test()
...
+    bool tests_run = false;
+
+#if !_PSTL_TEST_COMPARISON_BROKEN
     auto ret = kernel_test();
     EXPECT_TRUE(ret, "Wrong result of work with dpl::array and '>' in kernel_test");
+    tests_run = true;
+#endif // !_PSTL_TEST_COMPARISON_BROKEN
 
-    return TestUtils::done();
+    return TestUtils::done(tests_run);
```
SergeyKopienko, line 59: I propose to change a little bit this code:
- option 1: rename `tests_run` to `bProcessed` like in other tests or wise versa;
- option 2: not introduce this local variable (here and in all other places and use `_PSTL_TEST_COMPARISON_BROKEN` here: return `TestUtils::done(_PSTL_TEST_COMPARISON_BROKEN);`
But redefine `_PSTL_TEST_COMPARISON_BROKEN` by another way:
```C++
// oneAPI DPC++ compiler 2024.2.0 and earlier is unable to eliminate a "dead" function call to an undefined function
// within a sycl kernel which MSVC uses to allow comparisons with literal zero without warning
#if __cplusplus >= 202002L && _MSVC_LANG >= 202002L && _MSVC_STL_VERSION >= 143 && _MSVC_STL_UPDATE >= 202303L && __INTEL_LLVM_COMPILER <= 20240200
#    define _PSTL_TEST_COMPARISON_BROKEN 1
#else
#    define _PSTL_TEST_COMPARISON_BROKEN 0
#endif
```

danhoeflinger: Sure, I can do that.  I don't have a strong preference, I'll do option 1 if there are no objections.

test/support/test_config.h
```diff
@@ -154,6 +154,12 @@
 #define _PSTL_TEST_COMPLEX_SINH_BROKEN _PSTL_TEST_COMPLEX_OP_BROKEN
 #define _PSTL_TEST_COMPLEX_TANH_BROKEN _PSTL_TEST_COMPLEX_OP_BROKEN
 
+// oneAPI DPC++ compiler 2024.2.0 and earlier is unable to eliminate a "dead" function call to an undefined function
+// within a sycl kernel which MSVC uses to allow comparisons with literal zero without warning
+#define _PSTL_TEST_COMPARISON_BROKEN                                                                                   \
+    (__cplusplus >= 202002L && _MSVC_LANG >= 202002L && _MSVC_STL_VERSION >= 143 && _MSVC_STL_UPDATE >= 202303L &&     \
+    __INTEL_LLVM_COMPILER <= 20240200)
```
akukanov, line 161: Since you compare for less, you also need to check that the macro is non-zero (or defined)

danhoeflinger: Yes thanks, fixed.

test/xpu_api/containers/sequences/array/array.comparison/greater_g.pass.cpp
danhoeflinger: done (option 1)

test/support/test_config.h
```diff
@@ -154,6 +154,12 @@
 #define _PSTL_TEST_COMPLEX_SINH_BROKEN _PSTL_TEST_COMPLEX_OP_BROKEN
 #define _PSTL_TEST_COMPLEX_TANH_BROKEN _PSTL_TEST_COMPLEX_OP_BROKEN
 
+// oneAPI DPC++ compiler 2024.2.0 and earlier is unable to eliminate a "dead" function call to an undefined function
+// within a sycl kernel which MSVC uses to allow comparisons with literal zero without warning
+#define _PSTL_TEST_COMPARISON_BROKEN                                                                                   \
+    (__cplusplus >= 202002L && _MSVC_LANG >= 202002L && _MSVC_STL_VERSION >= 143 && _MSVC_STL_UPDATE >= 202303L &&     \
```
akukanov, line 160: The rules how Microsoft defines their language version macros: `__cplusplus` https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=msvc-170, `_MSVC_LANG` https://learn.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-170#microsoft-specific-predefined-macros. Unfortunately, our compiler does not follow these rules closely - in particular, if I remember correctly, their values are not necessarily the same with `/Zc:__cplusplus`.

Nevertheless, **both** macros are set to e.g. `C++20` only with that extra option. Therefore, generally we should check **any** of the two (i.e., ``||``) when looking for a certain version or greater, and none of the two (i.e. ``&&``) when looking for less than a certain version.

See also the definitions of ``_ONEDPL___cplusplus`` https://github.com/oneapi-src/oneDPL/blob/456bbe553b5b4fcc523875dd507f4a75d529b494/include/oneapi/dpl/internal/common_config.h#L19-L25 and of ``TEST_STD_VER`` https://github.com/oneapi-src/oneDPL/blob/456bbe553b5b4fcc523875dd507f4a75d529b494/test/support/test_macros.h#L95-L108

akukanov, line 160: In principle, checking ``_MSVC_STL_UPDATE`` makes the check of ``_MSVC_STL_VERSION`` redundant; though I do not mind if you prefer checking both.

Citing https://devblogs.microsoft.com/cppblog/c17-progress-in-vs-2017-15-5-and-15-6/ :

> The STL now identifies itself with two macros, defined by including any STL header (e.g. <ciso646>). _MSVC_STL_VERSION is defined as 141 and will remain that way for the VS 2017 v141 toolset series. _MSVC_STL_UPDATE is defined as 201709 in VS 2017 15.5, and will be increased as features are added in future updates. (The year-month typically won’t correspond to an update’s release date. Our process for updating this macro is “we’re checking in std::meow() and it’s now October, so we need to change the value”, and depending on branch/merge timing, it might take a month or more to ship.) The value for VS 2017 15.6 is not yet set in stone, but it will be 201711 or greater.

See also https://github.com/microsoft/STL/wiki/Macro-_MSVC_STL_UPDATE ; it does not mention ``_MSVC_STL_VERSION`` directly, but in fact its values are only updated with major Visual Studio versions: `141` for VS 2017, `142` for VS 2019, and `143` for VS 2022.

danhoeflinger: Thanks for this information, I was not aware of this.  I thought I was following the example in this file, but when I flipped the directionality of the sign I did not flip the `&&` to an `||`.  Fixed.

A broader question:  Is there a reason we cannot include `test_macros.h` in this file, and make use of the unified `TEST_STD_VER` macro?  (or combine the two headers?)

akukanov: > A broader question: Is there a reason we cannot include test_macros.h in this file, and make use of the unified TEST_STD_VER macro? (or combine the two headers?)

I do not know such a reason. As far as I understand, these two files come from different places. i.e. evolved independently till recent time. I think there is a sufficient overlap to consider unification.

Setting the problem even broader, our testing harness contains a wohle lot of poorly documented macros, routines etc. which could be reused if developers knew of their existence.

SergeyKopienko (approved):
LGTM

PR #1571, opened 2024-05-08
[onedpl][ranges][test] Added std ranges to oneDPL Tested API

MikeDvorskiy:
[onedpl][ranges][test] Added std ranges to oneDPL Tested API (for parallel oneDPL algorithms):

C++ standard sized RA views (factories and adaptors), released with C++20:

1. views::all
2. subrange
3. single_view
4. iota_view
5. take_view
6. drop_view
7. reverse_view
8. transform_view


test/parallel_api/ranges/xpu_api/xpu_std_ranges_test.h
```diff
@@ -0,0 +1,32 @@
...
+//
+// This file incorporates work covered by the following copyright and permission
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+template<typename KernelName, typename Test>
```
SergeyKopienko, line 16: Looks like some includes required here.

test/support/test_config.h
```diff
@@ -117,6 +117,8 @@
 #endif
 #endif //!defined(_ENABLE_RANGES_TESTING)
 
+#define _ENABLE_STD_RANGES_TESTING (_ONEDPL___cplusplus >= 202002L)
```
akukanov, line 120: `_ONEDPL___cplusplus` is an implementation detail. Please use `__cplusplus` and `_MSVC_LANG` here. See also https://github.com/oneapi-src/oneDPL/pull/1568#discussion_r1593947643

akukanov (commented):
For other tested APIs we preferred to adjust libc++ tests, Have you looked at https://github.com/llvm/llvm-project/tree/main/libcxx/test/std/ranges? Does it make sense to use these for testing, and if not - why?

MikeDvorskiy:
> For other tested APIs we preferred to adjust libc++ tests, Have you looked at https://github.com/llvm/llvm-project/tree/main/libcxx/test/std/ranges? Does it make sense to use these for testing, and if not - why?

No, I have not looked at ".../libcxx/test/std/ranges" before...
I've done it right now.... 
The first concern, that I have - a std test contains many RT assserts in a test body...  It doesnt work in a kernel, as far as I understand. 
As variant we can "invent" some macro which "converts" each assert condition to "b_res &&= condition" and write RT assert after kernel on the host side with global variable "b_res" check.
The second concern -  there are many files on each std range. Each file checks just one part of functionality of the range. As far as we have some restrictions on std views/ranges (like sized RA views), probably we we need just the test files which test juts "sized RA views/ranges"

MikeDvorskiy:
> RT asserts
Update: it seems that works, it leads to std::abort call... (https://intel.github.io/llvm-docs/design/Assert.html)

MikeDvorskiy:
> The second concern - there are many files on each std range. Each file checks just one part of functionality of the range.

as an example, I have adapted a test for "all_view" (as is with minimal changes).
https://github.com/oneapi-src/oneDPL/pull/1571/files#diff-e827a6e11d65e046ef5e9e92d7c4302e71a32b3ac4557dd16f8a423762221701

Ok, for "all_view" there is one (two but we can take juts one) test file (from LLVM). 
BUT, for the others views LLVM has about 10 test files for each testing view! Each file checks juts one method. I am not sure that we want to have so bulk tests for our needs in oneDPL.
In my opinion, checking  the base functionality in 1-3 lines (see below) in one lambda is enough for oneDPL.
```
auto test = [](){
        auto res = std::ranges::views::iota(0, 4);
        return res.size() == 4 && res[0] == 0 && res[1] == 1 && res[2] == 2 && res[3] == 3 
                  && *(res.begin() + 2) == 2 && (res.end() - res.begin()) == 4;
    };
```

SergeyKopienko:
> > The second concern - there are many files on each std range. Each file checks just one part of functionality of the range.
> 
> as an example, I have adapted a test for "all_view" (as is with minimal changes). https://github.com/oneapi-src/oneDPL/pull/1571/files#diff-e827a6e11d65e046ef5e9e92d7c4302e71a32b3ac4557dd16f8a423762221701
> 
> Ok, for "all_view" there is one (two but we can take juts one) test file (from LLVM). BUT, for the others views LLVM has about 10 test files for each testing view! Each file checks juts one method. I am not sure that we want to have so bulk tests for our needs in oneDPL. In my opinion, checking the base functionality in 1-3 lines (see below) in one lambda is enough for oneDPL.
> 
> ```
> auto test = [](){
>         auto res = std::ranges::views::iota(0, 4);
>         return res.size() == 4 && res[0] == 0 && res[1] == 1 && res[2] == 2 && res[3] == 3 
>                   && *(res.begin() + 2) == 2 && (res.end() - res.begin()) == 4;
>     };
> ```

From point of view of other our test implementations, it's good way to have one test for one function.
We always able to se exactly what was broken.

test/support/test_config.h
MikeDvorskiy: Probbaly, does it make sense to use the following definition in `onedpl_config.h `??
`#define _ONEDPL_CPP20_RANGES_PRESENT (__cplusplus >= 202002L || _MSVC_LANG >= 202002L) && __cpp_lib_ranges >= 201911L `
?
(https://en.cppreference.com/w/cpp/feature_test)

test/parallel_api/ranges/xpu_api/xpu_std_ranges_test.h
MikeDvorskiy: + `#include "support/utils_sycl.h"`

test/xpu_api/ranges/xpu_std_single_view.pass.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include "xpu_std_ranges_test.h"
+#endif //_ENABLE_STD_RANGES_TESTING
+
+int
+main()
+{
+#if _ENABLE_STD_RANGES_TESTING
+    auto test = [](){
+        auto res = std::ranges::views::single(1);
+        return res.size() == 1 && res[0] == 1 && *res.begin() == 1 && res.end() - res.begin() == 1;
```
dmitriy-sobolev, line 31: Could you also check `data()`?

test/xpu_api/ranges/xpu_std_subrange.pass.cpp
```diff
@@ -0,0 +1,40 @@
...
+int
+main()
+{
+#if _ENABLE_STD_RANGES_TESTING
+    auto test = [](){
+        auto v = std::ranges::views::iota(0, 4);
+        auto res = std::ranges::subrange(v.begin() + 1, v.end());
+        return res.size() == 3 && res[0] == 0 && res[1] == 1 && res[2] == 2 &&
+               (*res.begin() + 2) == 2 && res.end() - res.begin() == 3;
+    };
```
dmitriy-sobolev, line 34: Could you also check `prev()`, `next()` and `advance()`?

dmitriy-sobolev (commented):
Overall, the PR looks good, but I've left some comments regarding the added test.

The PR covers only a portion of the c++20 ranges functionality and does not cover free range functions like `ranges::size()` or more advanced adaptors like `join_view`. Do you plan adding it in consequent PRs?


Functions such as `std::ranges::size()` and more adaptors such as `ranges::join_view`?

test/xpu_api/ranges/xpu_std_ranges_test.h
```diff
@@ -0,0 +1,34 @@
...
+    sycl::range<1> numOfItems{1};
+    sycl::buffer<bool, 1> buffer1(&ret, numOfItems);
+    deviceQueue.submit([&](sycl::handler& cgh) {
+        auto ret_access = buffer1.get_access<sycl::access::mode::write>(cgh);
+        cgh.single_task<KernelName>([=]() {
+        ret_access[0] = test(); //run test
+        });
+    });
+
+    return buffer1.get_host_access(sycl::read_only)[0];
```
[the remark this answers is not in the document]
MikeDvorskiy: According to the SYCL spec  `get_host_access` is a blocking call and returns RAII SYCL `accessor` type. A lifetime of an `accessor` instance is limited by "line number 33", where a subscription operator returns a `bool` value. So, I don't see any issues here.

test/xpu_api/ranges/xpu_std_subrange.pass.cpp
MikeDvorskiy: due to within a kernel any range/view must be a random access view, testing of `advance` doesn't have much sense, I guess...

test/xpu_api/ranges/xpu_std_ranges_test.h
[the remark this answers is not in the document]
dmitriy-sobolev: > where a subscription operator returns a bool value

As far as I understand, it returns a reference.

The synopsis of `host_accessor::operator[]` according to https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_interface_for_buffer_host_accessors:
`reference operator[](size_t index) const;`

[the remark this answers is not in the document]
dmitriy-sobolev: I experimented more, and I agree there is no UB: a copy is created upon returning.

test/xpu_api/ranges/xpu_std_drop_view.pass.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include "xpu_std_ranges_test.h"
+#endif //_ENABLE_STD_RANGES_TESTING
+
+int
+main()
+{
+#if _ENABLE_STD_RANGES_TESTING
+    auto test = [](){
+        auto res = std::ranges::views::iota(0, 4) | std::ranges::views::drop(2);
+        return res.size() == 2 && res[0] == 2 && res[1] == 3 && *res.begin() == 2 && res.end() - res.begin() == 2;
```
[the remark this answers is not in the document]
MikeDvorskiy: Actually, the case is a little bit tricky..
There is a "pipe fusion"  here and `res` type is `iota_view`, which doesn't have `base()` method.

test/xpu_api/ranges/xpu_std_reverse_view.pass.cpp
```diff
@@ -0,0 +1,39 @@
...
+#endif //_ENABLE_STD_RANGES_TESTING
+
+int
+main()
+{
+#if _ENABLE_STD_RANGES_TESTING
+    auto test = [](){
+        auto res = std::ranges::views::iota(0, 4) | std::ranges::views::reverse;
+        return res.size() == 4 && res[0] == 3 && res[1] == 2 && res[2] == 1 && res[3] == 0 &&
+               (*res.begin() + 2) == 1 && res.end() - res.begin() == 4;
```
dmitriy-sobolev, line 32:
```suggestion
              *(res.begin() + 2) == 1 && res.end() - res.begin() == 4;
```

dmitriy-sobolev (approved):
LGTM

SergeyKopienko (approved):
LGTM

PR #1575, opened 2024-05-09
Remove ranges-v3 from experimental::dr

mateuszpn:
Replace all the ranges-v3 with std::ranges. Requires C++23.


include/oneapi/dpl/internal/distributed_ranges_impl/detail/enumerate.hpp
```diff
@@ -40,10 +40,10 @@ class enumerate_adapter_closure {
     }
   }
 
-  template <rng::viewable_range R>
-  friend auto operator|(R &&r, const enumerate_adapter_closure &closure) {
-    return closure(std::forward<R>(r));
-  }
+   template <rng::viewable_range R>
```
lslusarczyk, line 43: remove just formatting change

PR #1577, opened 2024-05-10
[test] Resolve UB in _PSTL_ICPX_TEST_MINMAX_ELEMENT_BROKEN macro

mmichel11:
The following compiler warning is produced when compiling `minmax_element.pass` with `icpx`:
```
warning: macro expansion producing 'defined' has undefined behavior [-Wexpansion-to-defined]
   28 |      !defined(_PSTL_TEST_MINMAX_ELEMENT) && !_PSTL_ICPX_TEST_MINMAX_ELEMENT_BROKEN
```
This PR changes how we define this macro to avoid the UB.


julianmi (dismissed):
Thanks for catching this. LGTM.

test/support/test_config.h
```diff
@@ -192,7 +192,8 @@
 
 // oneAPI DPC++ compiler 2022.2 an below show an internal compiler error during the backend code generation of
 // minmax_element.pass.cpp
-#define _PSTL_ICPX_TEST_MINMAX_ELEMENT_BROKEN                                                                         \
-    (TEST_DPCPP_BACKEND_PRESENT && defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER < 20220300)
+#if TEST_DPCPP_BACKEND_PRESENT && defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER < 20220300
+#   define _PSTL_ICPX_TEST_MINMAX_ELEMENT_BROKEN 1
```
MikeDvorskiy, line 196: It is not quite correct, due to _PSTL_ICPX_TEST_MINMAX_ELEMENT_BROKEN might be undefined when condition on line 195 is false

MikeDvorskiy (changes requested):

test/support/test_config.h
danhoeflinger: One option would be to just change `defined(__INTEL_LLVM_COMPILER)` to `__INTEL_LLVM_COMPILER > 0`, and have no other changes. 

As it is written, the check is if `!defined(_PSTL_ICPX_TEST_MINMAX_ELEMENT_BROKEN)` so I think this does work, but its a little odd to set to `1` then check if defined rather than if `true`

MikeDvorskiy (dismissed):

mmichel11:
I've applied @danhoeflinger 's suggestion on just checking of `__INTEL_LLVM_COMPILER > 0`. If it is undefined, it will be treated as 0 in the preprocessor if. Most of our broken macros evaluate to being `0` or `1` and the value is then checked, so I think it's best to keep this behavior.

test/support/test_config.h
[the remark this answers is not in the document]
mmichel11: Made this change.

danhoeflinger (approved):
LGTM

julianmi (approved):

PR #1578, opened 2024-05-10
[Tests] Avoid usm-allocated vector iterators wrapped in custom iterators for FPGA

danhoeflinger:
In the Intel LLVM compiler when building for FPGA, usm-allocated vector iterators fail to build when summed with an integer.  
This means that using usm-allocated vector iterators cannot be used in a `zip_iterator`, `transform_iterator` or `permutation_iterator` as a map and then dereferenced with the bracket operator.

This PR skips testing these input types for intel llvm compiler 2024.2 or earlier.

(This has been reported to the intel llvm compiler. Depending on when the fix is made, we may need to update the macro with the appropriate version)


mmichel11 (approved):
LGTM

SergeyKopienko (approved):
LGTM

PR #1579, opened 2024-05-13
Fix broken tests of `std::complex` - part 1/2

SergeyKopienko:
In this PR we fix two broken tests of `std::complex` :
- complex_times_complex.pass.cpp
- complex_divide_complex.pass.cpp


test/xpu_api/numerics/complex.number/complex.ops/complex_divide_complex.pass.cpp
```diff
@@ -69,15 +69,17 @@ void test_edges()
                 switch (classify(testcases[j]))
                 {
                 case zero:
+#if !_PSTL_TEST_COMPLEX_DIV_COMPLEX_BROKEN && !_PSTL_TEST_COMPLEX_DIV_COMPLEX_BROKEN_GLIBCXX
```
danhoeflinger, line 72: I see here you've added both `!_PSTL_TEST_COMPLEX_DIV_COMPLEX_BROKEN && !_PSTL_TEST_COMPLEX_DIV_COMPLEX_BROKEN_GLIBCXX`.  Is that intentional?

This is the only place `!_PSTL_TEST_COMPLEX_DIV_COMPLEX_BROKEN` is added to an assert within this PR

SergeyKopienko: Fixed

danhoeflinger (approved):
LGTM

PR #1580, opened 2024-05-13
Fix broken tests of `std::complex` - part 2/2

SergeyKopienko:
In this PR we fix two broken tests of std::complex :
- acos.pass.cpp
- exp.pass.cpp
- tanh.pass.cpp

Also in this PR we move assertion check from `is_about` function to caller side to have ability view line code with broken test case error.


test/support/test_config.h
```diff
@@ -144,11 +144,17 @@
 #define _PSTL_TEST_COMPLEX_OP_BROKEN (_MSVC_STL_VERSION && _MSVC_STL_VERSION <= 143)
 
 #define _PSTL_TEST_COMPLEX_ACOS_BROKEN _PSTL_TEST_COMPLEX_OP_BROKEN
+#define _PSTL_TEST_COMPLEX_ACOS_BROKEN_GLIBCXX                                                                         \
+    (__SYCL_DEVICE_ONLY__ && __INTEL_LLVM_COMPILER > 0 && __INTEL_LLVM_COMPILER <= 20240200 && __GLIBCXX__ > 0 &&      \
```
danhoeflinger, line 148: I've not seen us protect this with `__SYCL_DEVICE_ONLY__` before.  Is this necessary?
If so, should we be doing this for many more of these macros?

danhoeflinger, line 148: Just to confirm, this is specifically broken only for the combination of Intel LLVM and gcc standard library implementation?

danhoeflinger, line 148: Does it make sense for us to consolidate all "broken" macros which are meant to mean "up to the current version of X" where X is:
 intel llvm compiler, gcc, libstdc++, libc++, MSVC STL, etc?

This should of course be considered in a different PR, but this could help us to update these for releases.  If we do something like this, we need to be careful to remove issues which have since been fixed from the "up to the current version of X" macros.  (Which means there should be some sort of formal process more than incrementing the consolidated number blindly.)

Just a thought for the future, not for this PR.

PR #1581, opened 2024-05-13
Remove extra broken defines from `std::complex` tests

SergeyKopienko:
This extra broken define has been introduced in https://github.com/oneapi-src/oneDPL/pull/1579 :
```C++
_PSTL_TEST_COMPLEX_TIMES_COMPLEX_BROKEN_GLIBCXX (__GLIBCXX__ > 0 && __GLIBCXX__ <= 20210601)
```

But as far as `_PSTL_TEST_COMPLEX_TIMES_COMPLEX_BROKEN_GLIBCXX` always used together with `_PSTL_TEST_COMPLEX_TIMES_COMPLEX_BROKEN` and vice versa, better to combine them in one macro:
```C++
#define _PSTL_TEST_COMPLEX_TIMES_COMPLEX_BROKEN (_PSTL_TEST_COMPLEX_OP_BROKEN || (__GLIBCXX__ > 0 && __GLIBCXX__ <= 20210601))
```


test/xpu_api/numerics/complex.number/complex.ops/complex_divide_complex.pass.cpp
```diff
@@ -71,7 +71,7 @@ void test_edges()
                 case zero:
 #if !_PSTL_TEST_COMPLEX_DIV_COMPLEX_BROKEN_GLIBCXX
                     assert(classify(r) == inf);
-#endif
+#endif // _PSTL_TEST_COMPLEX_DIV_COMPLEX_BROKEN_GLIBCXX
```
danhoeflinger, line 74: Is this necessary?  the `#if` is 2 lines away.  This is also somewhat unrelated to this PR. (same for the others in this file)

If we add this, there are some other ones like line 82 and 103 we should also add for consistency (but I'd prefer just removing them).

PR #1582, opened 2024-05-13
Pinning dependencies per OSSF security practices and coverting to ASCII

timmiesmith:
https://github.com/ossf/scorecard/blob/main/docs/checks.md#pinned-dependencies recommends explicitly pinning dependencies to reduce several security risks.

Line endings in the file were inconsistent so I converted them all to ASCII instead of a mix of CR and CRLF line endings.

### DIFF
--- a/documentation/library_guide/requirements.txt
+++ b/documentation/library_guide/requirements.txt
@@ -10,8 +10,8 @@ sphinxcontrib_htmlhelp<2.0.4
 sphinxcontrib_qthelp<1.0.6
 sphinxcontrib-serializinghtml<1.1.10
 
-sphinx_book_theme
+sphinx_book_theme==1.1.2
 sphinx-intl==2.0.0
-sphinx-tabs
-sphinx-prompt
-sphinx_substitution_extensions
+sphinx-tabs==3.4.5
+sphinx-prompt==1.8.0
+sphinx_substitution_extensions==2024.2.25
--- a/documentation/library_guide/requirements.txt
+++ b/documentation/library_guide/requirements.txt
@@ -10,8 +10,8 @@ sphinxcontrib_htmlhelp<2.0.4
 sphinxcontrib_qthelp<1.0.6
 sphinxcontrib-serializinghtml<1.1.10
 
-sphinx_book_theme==1.1.2
+sphinx_book_theme==1.0.1
 sphinx-intl==2.0.0
 sphinx-tabs==3.4.5
-sphinx-prompt==1.8.0
-sphinx_substitution_extensions==2024.2.25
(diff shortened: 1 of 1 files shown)


danhoeflinger (commented):
LGTM,  
I checked that these version numbers are the current versions of these libraries.

danhoeflinger (dismissed):

mmichel11 (approved):
LGTM

Edit: See below

mmichel11 (changes requested):
These dependencies actually seem impossible to fulfill based on the CI failure:

```
    The user requested sphinx==4.4.0
    breathe 4.9.1 depends on Sphinx>=1.4
    sphinx-book-theme 1.1.2 depends on sphinx>=5
```
Would increasing the required sphinx version resolve this?

danhoeflinger:
> These dependencies actually seem impossible to fulfill based on the CI failure:
> 
> ```
>     The user requested sphinx==4.4.0
>     breathe 4.9.1 depends on Sphinx>=1.4
>     sphinx-book-theme 1.1.2 depends on sphinx>=5
> ```
> 
> Would increasing the required sphinx version resolve this?

That may have unintended consequences.  May be best to figure out which version of these was currently being used prior to this PR and just use those version numbers.

documentation/library_guide/requirements.txt
```diff
@@ -1,17 +1,17 @@
...
+# https://github.com/sphinx-doc/sphinx/issues/9923
+docutils==0.15
+# Pin versions to avoid a compatibility issue: "The sphinxcontrib.<extension_name> extension used by this project needs at least Sphinx v5.0"
+sphinxcontrib_applehelp<1.0.8
+sphinxcontrib_devhelp<1.0.5
+sphinxcontrib_htmlhelp<2.0.4
+sphinxcontrib_qthelp<1.0.6
+sphinxcontrib-serializinghtml<1.1.10
+
+sphinx_book_theme==1.1.2
```
danhoeflinger, line 13: Looking at another PR, it looks like this is correct:
sphinx_book_theme==1.0.1
sphinx-tabs==3.4.5
sphinx-prompt==1.5.0
sphinx_substitution_extensions==2022.2.16

timmiesmith: Thanks.  The version numbers I selected had been pulled from a previous PR, but I'm not able to find that PR now.

danhoeflinger: I think it initially tries the numbers you originally had, but then backs them off later in the script due to dependencies.

danhoeflinger (approved):
LGTM

mmichel11 (approved):

PR #1589, opened 2024-05-15
[oneDPL][tbb] + memory leaks fix  (sort/stable sort, tbb backend)

MikeDvorskiy:
[oneDPL][tbb] + memory leaks fix  (sort/stable sort, tbb backend)

### DIFF
--- a/include/oneapi/dpl/pstl/parallel_backend_tbb.h
+++ b/include/oneapi/dpl/pstl/parallel_backend_tbb.h
@@ -510,7 +510,7 @@ class __root_task
     friend class __func_task<_Func>;
 };
 
-#else  // TBB_INTERFACE_VERSION <= 12000
+#else  // TBB_INTERFACE_VERSION > 12000
 class __task : public tbb::detail::d1::task
 {
   protected:
@@ -644,10 +644,13 @@ class __func_task : public __task
 
         assert(__parent != nullptr);
         assert(__parent->_M_refcount.load(::std::memory_order_relaxed) > 0);
-        if (--__parent->_M_refcount == 0)
+
+        auto __refcount = --__parent->_M_refcount;
+        __alloc.deallocate(this, *__ed);
+
+        if (__refcount == 0)
         {
             assert(__next == nullptr);
-            __alloc.deallocate(this, *__ed);
             return __parent;
         }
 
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -644,10 +644,13 @@ class __func_task : public __task
 
         assert(__parent != nullptr);
         assert(__parent->_M_refcount.load(::std::memory_order_relaxed) > 0);
-        if (--__parent->_M_refcount == 0)
+
+        auto __refcount = --__parent->_M_refcount;
+        __alloc.deallocate(this, *__ed);
```
danhoeflinger, line 649: Is there a reason we have to decrement the refcount of the parent *before* we deallocate the child memory?

If not, couldn't we just move the deallocation call outside the `if` and simplify this code?

MikeDvorskiy: @pavelkumbrasev,
Could you please answer that question?

pavelkumbrasev: I believe it is some sort of an optimization (Used in several TBB patterns) where deallocation can be done concurrently  (the reference reached 0 and the thread can proceed with tree folding without waiting for task deallocation on another thread).
I think the benefit is really small so we can amend the optimization and use @danhoeflinger suggestion if it helps with readability.

danhoeflinger: If it is faster to do it this way I wont object.  It may be worth adding a comment to make that clear going forward.  

The difference in readability is minor.  However, from someone unfamiliar with the code to start with, it makes it seem like there are more hidden connections which require this pattern.

MikeDvorskiy: I've added a comment.

danhoeflinger (approved):
LGTM

include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -644,10 +644,16 @@ class __func_task : public __task
 
         assert(__parent != nullptr);
         assert(__parent->_M_refcount.load(::std::memory_order_relaxed) > 0);
-        if (--__parent->_M_refcount == 0)
+
+        auto __refcount = --__parent->_M_refcount;
+        
+        //The deallocation can be done concurrently (the reference reached 0 and the thread can proceed with tree
+        //folding without waiting for task deallocation on another thread)
```
akukanov, line 651:
```suggestion
        // Placing the deallocation after the refcount decrement allows another thread to proceed with tree
        // folding concurrently with this task cleanup.
```

danhoeflinger (approved):
LGTM

PR #1590, opened 2024-05-16
Optional avoid switch off MS STL checked iterators in tests

SergeyKopienko:
Looks like in modern versions of MS STL the iterator debug level switching off isn't required anymore.
It was introduced here: https://github.com/oneapi-src/oneDPL/pull/915

In this PR we give the opportunity to not change default debug iterator level at all.
For that we introduce new CMake option `ONEDPL_TESTS_DO_NOT_SWITCH_OFF_ITERATOR_DEBUG_LEVEL` :
```C++
option(ONEDPL_TESTS_DO_NOT_SWITCH_OFF_ITERATOR_DEBUG_LEVEL "Do not switch off MS STL iterator debug level" OFF)
```

Their usage in cmake command line:
`-DONEDPL_TESTS_DO_NOT_SWITCH_OFF_ITERATOR_DEBUG_LEVEL=ON`


test/CMakeLists.txt
```diff
@@ -17,6 +17,8 @@ add_subdirectory(kt)
 set (ranlux_24_48_test.pass_timeout_debug "900") # 15min
 set (ranlux_24_48_test.pass_timeout_release "720") # 12min
 
+option(ONEDPL_TESTS_DO_NOT_SWITCH_OFF_ITERATOR_DEBUG_LEVEL "Do not switch off MS STL iterator debug level in tests" OFF)
+
```
dmitriy-sobolev, line 21: Is there a way to infer this, e.g. basing on MSVC library version? It might be better comparing to manually setting it as no actions are needed externally.

Have you tried [MSVC_VERSION](https://cmake.org/cmake/help/v3.15/variable/MSVC_VERSION.html#variable:MSVC_VERSION) macro?

SergeyKopienko: I am already displaying their state in `CMakeLists.txt`.
But from which versions of `MSVC_VERSION` we able to not switch off checking iterators for Kernels code - this question still open.
```C++
# Display version of used MS STL library
if (${MSVC_VERSION})
    # Their value described at https://cmake.org/cmake/help/latest/variable/MSVC_VERSION.html
    message(STATUS "MSVC_VERSION: ${MSVC_VERSION}")
endif()
```

SergeyKopienko:
Looks like this PR doesn't required anymore.

PR #1592, opened 2024-05-16
test rfc, ignore

rscohn2:


rscohn2:
The notification job has been queued for more than 1 hour: https://github.com/oneapi-src/oneDPL/actions/runs/9112412107/job/25051604985?pr=1592
This is because oneapi-src can only have 20 runners across all projects.

After the move to onedpl-project org, you won't have to share with other projects. We are also looking at getting runners with more than 1 core.

rscohn2:
In all my previous testing I was the author of the PR and did the labelling. I see that it is using the id of the labeller, not the PR author.
https://uxlfoundation.slack.com/archives/C06HLTMF8QK/p1715870164231359

I will look into that.

PR #1593, opened 2024-05-17
Remove assertions in minmax_element binary operator

julianmi:
The current `minmax_element` implementation requires in-order processing since its binary operator assumes that the index of the first input is lower than that of the second input. This is a performance improvement since we can skip the case when both input values are equal. The in-order processing is guaranteed by following the non-commutative reduce pattern. We currently check the index order with assertions with lead to high overheads when compiled without `NDEBUG`. This may lead to test timeouts or user code looking to be hanging. I think the assertions are not needed since we enforce them in the in-order reduce pattern and we don't use them in the similar `min_element` algorithm.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -522,9 +522,6 @@ __pattern_minmax_element(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _
         auto __chosen_for_min = __a;
         auto __chosen_for_max = __b;
 
-        assert(get<0>(__a) < get<0>(__b));
-        assert(get<1>(__a) < get<1>(__b));
-
         if (__comp(get<2>(__b), get<2>(__a)))
             __chosen_for_min = ::std::move(__b);
         if (__comp(get<3>(__b), get<3>(__a)))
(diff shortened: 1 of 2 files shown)


dmitriy-sobolev:
@julianmi, could you remove the assertions in `include\oneapi\dpl\pstl\hetero\algorithm_ranges_impl_hetero.h`?

Meanwhile, I found no mention if assert can be used in SYCL code. 
Given that there is a corresponding extension in [oneAPI DPC++ Compiler](https://github.com/intel/llvm/blob/5d02eaedcba58d1139dd69a88e8e372bf582bbda/sycl/doc/extensions/supported/sycl_ext_oneapi_assert.asciidoc), I assume that we should avoid using asserts in kernels, or at least use them conditionally.

julianmi:
> @julianmi, could you remove the assertions in `include\oneapi\dpl\pstl\hetero\algorithm_ranges_impl_hetero.h`?


Done.

> Meanwhile, I found no mention if assert can be used in SYCL device code reading its specification. Given that there is a corresponding extension in [oneAPI DPC++ Compiler](https://github.com/intel/llvm/blob/5d02eaedcba58d1139dd69a88e8e372bf582bbda/sycl/doc/extensions/supported/sycl_ext_oneapi_assert.asciidoc), I assume that we should avoid using asserts in kernels, or at least use them conditionally.

I agree with your statement. I looks like assert seems to work for now in kernel code though. Anyways, I think our tests should capture such semantic issue in our kernels (and they do for this specific case).

dmitriy-sobolev (approved):
LGTM

danhoeflinger (approved):
LGTM

PR #1603, opened 2024-05-27
Fix errors in `__pattern_sort` implementations and calls

SergeyKopienko:
In this PR we implement another approach for the problem found in https://github.com/oneapi-src/oneDPL/pull/1599
Also approach from this PR is compatible with implementation of `__pattern_sort` from our up-stream:
https://github.com/search?q=repo%3Allvm%2Fllvm-project%20path%3A%2F%5Epstl%5C%2Finclude%5C%2Fpstl%5C%2F%2F%20__pattern_sort(&type=code

What we doing in this PR:
- remove `static_assert' check of tag type from  `__pattern_sort` + tag implemetation.
- pass real sign of `is_move_constructible` value into `__pattern_sort` from `__pattern_partial_sort_copy` + hetero impl.


SergeyKopienko:
PR has been created by error

PR #1605, opened 2024-05-28
Only declaration required as make_wrapped_policy template params

SergeyKopienko:
In this PR we doing the next things:
1) Only declaration required as `make_wrapped_policy` template param
2) For aligning we declare these things always as `struct`.

The goal - to align these structures: their body isn't required.
And we have a lot of the same structures without body.
For example, 
```C++
template <typename Name>
class ExclusiveScan1;
template <typename Name>
class ExclusiveScan2;
```
and so far and so on.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -168,9 +168,7 @@ __pattern_walk3(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _ForwardIt
 //------------------------------------------------------------------------
 
 template <typename _Name>
-struct __walk_brick_wrapper
-{
-};
+struct __walk_brick_wrapper;
 
 template <typename _BackendTag, typename _ExecutionPolicy, typename _ForwardIterator, typename _Function>
 void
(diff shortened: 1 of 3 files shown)


SergeyKopienko:
@rarutyun, how do you think, does it make sense to apply this PR ?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1002,13 +990,10 @@ __pattern_unique_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec
...
-class copy_back_wrapper
-{
-};
+struct copy_back_wrapper;
+
 template <typename _Name>
-class copy_back_wrapper2
-{
-};
+struct copy_back_wrapper2;
```
danhoeflinger, line 996: id recommend changing this to `copy_back2_wrapper` for consistency, or encoding more meaningful information in the name instead of `2`.

danhoeflinger (commented):
I don't believe there is any downside to making this change to be declaration only, since these types are merely used as indicator types to make unique and identifiable policy names.  However, I am not 100% sure that making the change to declaration only is without negative consequences in all compilers / environments.  

I don't see much benefit to the change, as it only removes a handful of braces / lines. I suppose it may prevent someone from doing something with the type for some other purpose beyond how it is currently used as a way to make a unique policy name.

Alignment is the most interesting part to me, but I think the most important part of an alignment would be the name, because the name is effectively the only purpose of these types.  I've pointed out a few outliers and places we can improve.

I don't see a benefit to aligning class vs struct, but I don't object to it either.

danhoeflinger (approved):
After offline discussion, I realize I've missed that we have many such instances of declaration only objects used in this fashion throughout oneDPL (I think I've written some).  
With this in mind, I think this is safe to do.  I wont insist on naming changes, because that would make this PR much larger than its original intentions, but I do think it could be a benefit.  

I see no problem with this PR merging.

MikeDvorskiy (approved):
LGTM

PR #1606, opened 2024-05-29
Exclude sycl::half testing on FPGA devices in sort.pass

mmichel11:
`sort.pass` currently fails on FPGA_EMU devices with the following error message:

```
 what():  Required aspect fp16 is not supported on the device
```

This PR removes `sycl::half` testing with FPGA devices to resolve this issue.

### DIFF
--- a/test/parallel_api/algorithm/alg.sorting/sort.pass.cpp
+++ b/test/parallel_api/algorithm/alg.sorting/sort.pass.cpp
@@ -451,7 +451,7 @@ main()
             [](size_t k, size_t val) {
             return std::int16_t(val) * (k % 2 ? 1 : -1); });
 
-#if TEST_DPCPP_BACKEND_PRESENT
+#if TEST_DPCPP_BACKEND_PRESENT && !ONEDPL_FPGA_DEVICE
         auto convert = [](size_t k, size_t val) {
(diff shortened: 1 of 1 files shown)


test/parallel_api/algorithm/alg.sorting/sort.pass.cpp
```diff
@@ -451,7 +451,7 @@ main()
             [](size_t k, size_t val) {
             return std::int16_t(val) * (k % 2 ? 1 : -1); });
 
-#if TEST_DPCPP_BACKEND_PRESENT
+#if TEST_DPCPP_BACKEND_PRESENT && !ONEDPL_FPGA_DEVICE
```
SergeyKopienko, line 454: Probably only some FPGA-devices hasn't `sycl::half` support.
If so, required to check device aspect about this.

mmichel11:
I have closed this pull request. The root of the issue is that `FPGA_EMU` is using AOT compilation which currently avoids our checks for device aspect support for fp16 with the icpx compiler. @dmitriy-sobolev and I have discussed and the proper solution is to add `-fsycl-device-code-split=per_kernel` to `FPGA` compilation targets similarly to what we do for GPUs.

PR #1607, opened 2024-05-30
[KT] Implement reduce then scan optimization in ESIMD scan prototype

mmichel11:
This PR is targeting the ESIMD scan development branch.

This PR implements the reduce-then-scan approach in the ESIMD scan implementation reducing the ~4N global memory accesses to ~3N with a tradeoff of extra compute in the second kernel.

The following changes have been made:

-  `MAX_INPUTS_PER_BLOCK` has been adjusted to an empirically determined value. The prior value is based on what would be used in a single-pass implementation where we first load data into SLM and are limited by the SLM available per XE-core. We do not have this limit with the two-kernel implementation and processing `2^24` elements per block has yielded the best results in my experiments.
- `N` stores of partial prefix sums to global memory have been removed from the first kernel and an additional `prefix_sum` computation is added to the second kernel.
- The carry in from the previous block is propagated in the second kernel instead of the first.
- I have adjusted the header comment.

Testing has been performed on powers-of-2 from 2^17 to 2^28.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/esimd_scan.h
+++ b/include/oneapi/dpl/experimental/kt/esimd_scan.h
@@ -171,14 +171,8 @@ void inclusive_scan(sycl::queue q, InputIterator first, InputIterator last, Outp
 
         // propogate carry in from previous block
         ValueType carry_in = 0;
-        if (b > 0) {
-        #if 0
-        ci = scalar_load<uint>(result,offset+(b*blockSize-1)*sizeof(uint32_t));
-        #else
-        // No scalar_load pointer API. Use a gather with a constant offset
-        carry_in = gather<ValueType, 1>(result, simd<uint32_t, 1>(offset+(b*blockSize-1)*sizeof(ValueType)))[0];
-        #endif
-        }
+        if (b > 0)
+            carry_in = gather<ValueType, 1>(result, simd<uint32_t, 1>(offset+(b*blockSize-1)*sizeof(ValueType)))[0];
 
         // on each Xe T0: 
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/experimental/kt/esimd_scan.h
```diff
@@ -179,6 +169,17 @@ void inclusive_scan(sycl::queue q, InputIterator first, InputIterator last, Outp
         simd<ValueType,VL> v;
         ValueType carry31;
 
+        // propogate carry in from previous block
+        ValueType carry_in = 0;
+        if (b > 0) {
+        #if 0
+        ci = scalar_load<uint>(result,offset+(b*blockSize-1)*sizeof(uint32_t));
+        #else
```
akukanov, line 177: There should be no `#if 0` blocks of code; please either remove or specify a reasonable preprocessor condition. This code will not even compile, because of `ci` being not declared.

mmichel11: The line in the `#if 0` is no longer needed, so I have removed it. I will prepare a PR for the other places this has been done when working on the prototype.

danhoeflinger: Its worth noting that this PR is targeting a branch for shared development of a prototype, rather than oneDPL main.

adamfidel (approved):
LGTM.

mmichel11:
This PR will serve as the baseline for prototyping a SYCL version of the same algorithm.

PR #1614, opened 2024-06-05
update of dr::shp::sort()

lslusarczyk:
propagation of change https://github.com/oneapi-src/distributed-ranges/pull/740 and https://github.com/oneapi-src/distributed-ranges/pull/759


lslusarczyk:
@mateuszpn ,

1. format test is failing, both in this PR and in your https://github.com/oneapi-src/oneDPL/pull/1608 - please make it passing. But do it automatically, not manually. Look at oneDPL CI what command do they invoke to check format and invoke the same or similar one to fix the format of your change.
2. you have added `dr` namespace where it is not needed, e.g. changed `distributed_range` into `dr::distributed_range`. Please revert. Don't use namespace if you are already in this namespace. (`dr::ranges::rang` -> `ranges::rank`, `dr::shp::__detail::local` -> `__detail::local`). Shorter code increases readability. 
3. Don't use internal bug tracker IDs like `[GSD-3893]` in oneDPL
4. `ONEDPL_USE_DISTRIBUTED_RANGES` is not used in the code - remove from cmake.
5. Is `-ftemplate-backtrace-limit=0` really needed by this change. I don't think so because it was not needed before your change and the change does not add more templates.

include/oneapi/dpl/internal/distributed_ranges_impl/shp/algorithms/sort.hpp
```diff
@@ -137,118 +139,125 @@ sort(R&& r, Compare comp = Compare())
...
-        std::size_t* splitter_i = sycl::malloc_shared<std::size_t>(n_splitters, q.get_device(), context());
+        // splitter_i = [ index in local_segment of first element greater or equal
+        // 1st global median, index ... 2nd global median, ..., size of
+        // local_segment]
+        std::size_t* splitter_i = sycl::malloc_shared<std::size_t>(n_segments, q.get_device(), shp::context());
         splitter_indices.push_back(splitter_i);
 
-        // Local copy `medians_l` necessary due to bug in Level Zero.
-        T* medians_l = sycl::malloc_device<T>(n_splitters, q.get_device(), context());
+        // Local copy `medians_l` necessary due to [GSD-3893]
```
lslusarczyk, line 166: We can not use jira ID in the oneDPL code. Please restore previous version of this line.

test/CMakeLists.txt
```diff
@@ -249,6 +249,14 @@ foreach (_file IN LISTS UNIT_TESTS)
 
 endforeach()
 
+if (ONEDPL_USE_DR)
```
lslusarczyk, line 252: remove this

PR #1617, opened 2024-06-07
Performance improvements of `__parallel_find_or` + `__device_backend_tag`

SergeyKopienko:
In this PR we made some performance improvements:
- some performance improvements for `__parallel_find_or` + `__device_backend_tag` for the usage with the `__parallel_or_tag`.

------------------
The changes for
- `__pattern_any_of(__hetero_tag` has been implemented on `__parallel_transform_reduce`;
- `__pattern_find_if(__hetero_tag` has been implemented on `__parallel_transform_reduce`;
has been extracted to the separate PR https://github.com/oneapi-src/oneDPL/pull/1622
and rolled back in this PR.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -682,42 +682,19 @@ bool
 __pattern_any_of(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator __first, _Iterator __last,
                  _Pred __pred)
 {
-    using _difference_type = typename ::std::iterator_traits<_Iterator>::difference_type;
-
-    const _difference_type __n = __last - __first;
-    if (__n == 0)
+    if (__first == __last)
         return false;
 
-    using _result_type = oneapi::dpl::__internal::tuple<bool, _difference_type>;
-    const auto __init = _result_type{false, __n};
+    using _Predicate = oneapi::dpl::unseq_backend::single_match_pred<_ExecutionPolicy, _Pred>;
 
-    // __counting_iterator_t - iterate position (index) in source data
-    using __counting_iterator_t = oneapi::dpl::counting_iterator<_difference_type>;
+    auto __keep = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
+    auto __buf = __keep(__first, __last);
 
-    using _zipped_data_type = typename std::iterator_traits<decltype(oneapi::dpl::make_zip_iterator(__first, __counting_iterator_t{0}))>::value_type;
-
-    __find_if_binary_reduce_op<_zipped_data_type, /*_IsFirst*/ std::true_type> __reduce_op;
-    __find_if_unary_transform_op<_zipped_data_type, _Pred> __transform_op{__pred};
-
-    using _Functor = unseq_backend::walk_n<_ExecutionPolicy, decltype(__transform_op)>;
-    using _RepackedTp = __par_backend_hetero::__repacked_tuple_t<_result_type>;
-
-    auto __keep_src_data = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
-    auto __buf_src_data = __keep_src_data(__first, __last);
-
-    const __counting_iterator_t __counting_it_first{0}, __counting_it_last{__n};
-    auto __keep_counting_it = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, __counting_iterator_t>();
-    auto __buf_counting_it = __keep_counting_it(__counting_it_first, __counting_it_last);
-
-    auto res =
-        oneapi::dpl::__par_backend_hetero::__parallel_transform_reduce<_RepackedTp, std::true_type /*is_commutative*/>(
-            _BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
-            __reduce_op, _Functor{__transform_op},
-            unseq_backend::__init_value<_RepackedTp>{__init}, // initial value
-            oneapi::dpl::__ranges::make_zip_view(__buf_src_data.all_view(), __buf_counting_it.all_view()))
-            .get();
-
-    return std::get<0>(res);
+    return oneapi::dpl::__par_backend_hetero::__parallel_find_or(
+        _BackendTag{},
+        __par_backend_hetero::make_wrapped_policy<__par_backend_hetero::__or_policy_wrapper>(
+            ::std::forward<_ExecutionPolicy>(__exec)),
+        _Predicate{__pred}, __par_backend_hetero::__parallel_or_tag{}, __buf.all_view());
 }
 
 //------------------------------------------------------------------------
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -646,6 +646,37 @@ __pattern_count(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator 
 // any_of
 //------------------------------------------------------------------------
 
+template <typename _Typle, typename _UnaryTransformOp>
+struct __find_if_unary_transform_op
+{
+    _UnaryTransformOp __transform_op;
+
+    template <typename Arg>
+    _Typle
+    operator()(const Arg& arg) const
+    {
+        return {__transform_op(std::get<0>(arg)), std::get<1>(arg)};
+    }
+};
+
+template <typename _Typle, typename _IsFirst>
+struct __find_if_binary_reduce_op
+{
+    _Typle
+    operator()(const _Typle& op1, const _Typle& op2) const
+    {
+        if (std::get<0>(op1) && std::get<0>(op2))
+        {
+            if constexpr (_IsFirst{})
+                return {true, std::min(std::get<1>(op1), std::get<1>(op2))};
+            else
+                return {true, std::max(std::get<1>(op1), std::get<1>(op2))};
+        }
+
+        return std::get<0>(op1) ? op1 : op2;
+    }
+};
+
 template <typename _BackendTag, typename _ExecutionPolicy, typename _Iterator, typename _Pred>
 bool
 __pattern_any_of(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator __first, _Iterator __last,
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -572,9 +572,9 @@ __pattern_adjacent_find(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _I
 
     // TODO: in case of conflicting names
     // __par_backend_hetero::make_wrapped_policy<__par_backend_hetero::__or_policy_wrapper>()
-    bool result = __par_backend_hetero::__parallel_find_or(
-        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
-        _Predicate{adjacent_find_fn<_BinaryPredicate>{__predicate}}, __par_backend_hetero::__parallel_or_tag{},
+    bool result = __par_backend_hetero::__parallel_find_any(
+        _BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
+        _Predicate{adjacent_find_fn<_BinaryPredicate>{__predicate}},
         oneapi::dpl::__ranges::make_zip_view(__buf1.all_view(), __buf2.all_view()));
 
     // inverted conditional because of
@@ -659,11 +659,11 @@ __pattern_any_of(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator
     auto __keep = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
     auto __buf = __keep(__first, __last);
 
-    return oneapi::dpl::__par_backend_hetero::__parallel_find_or(
+    return oneapi::dpl::__par_backend_hetero::__parallel_find_any(
         _BackendTag{},
         __par_backend_hetero::make_wrapped_policy<__par_backend_hetero::__or_policy_wrapper>(
-            ::std::forward<_ExecutionPolicy>(__exec)),
-        _Predicate{__pred}, __par_backend_hetero::__parallel_or_tag{}, __buf.all_view());
+            std::forward<_ExecutionPolicy>(__exec)),
+        _Predicate{__pred}, __buf.all_view());
 }
 
 //------------------------------------------------------------------------
@@ -687,9 +687,8 @@ __pattern_equal(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator1
 
     // TODO: in case of conflicting names
     // __par_backend_hetero::make_wrapped_policy<__par_backend_hetero::__or_policy_wrapper>()
-    return !__par_backend_hetero::__parallel_find_or(
-        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec), _Predicate{equal_predicate<_Pred>{__pred}},
-        __par_backend_hetero::__parallel_or_tag{},
+    return !__par_backend_hetero::__parallel_find_any(
+        _BackendTag{}, std::forward<_ExecutionPolicy>(__exec), _Predicate{equal_predicate<_Pred>{__pred}},
         oneapi::dpl::__ranges::make_zip_view(__buf1.all_view(), __buf2.all_view()));
 }
 
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -694,7 +694,8 @@ __pattern_any_of(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator
     // __counting_iterator_t - iterate position (index) in source data
     using __counting_iterator_t = oneapi::dpl::counting_iterator<_difference_type>;
 
-    using _zipped_data_type = typename std::iterator_traits<decltype(oneapi::dpl::make_zip_iterator(__first, __counting_iterator_t{0}))>::value_type;
+    using _zipped_data_type = typename std::iterator_traits<decltype(oneapi::dpl::make_zip_iterator(
+        __first, __counting_iterator_t{0}))>::value_type;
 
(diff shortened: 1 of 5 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1084,13 +1123,98 @@ struct __early_exit_find_or
...
+    _PRINT_INFO_IN_DEBUG_MODE(__exec, __wgroup_size, __max_cu);
+
+    const _AtomicType __init_value = _BrickTag::__init_value(__rng_n);
+    _AtomicType __result = __init_value;
+
+    const auto __pred = oneapi::dpl::__par_backend_hetero::__early_exit_find_or<_ExecutionPolicy, _Brick>{__f};
+
+    // scope is to copy data back to __result after destruction of temporary sycl:buffer
+    {
+        auto __temp = sycl::buffer<_AtomicType, 1>(&__result, 1); // temporary storage for global atomic
```
SergeyKopienko, line 1166: @julianmi, how do you think is it better to use `__result_and_scratch_storage` onsteam of this buffer?

julianmi: If I understand correctly, `__temp` is a global atomic used only during the kernel execution. Moving it to the host seems like an unnecessary performance penalty to me. Further, you don't need scratch space, right? It could make sense to use USM device memory to reduce some overheads using a smart pointer with `__sycl_usm_alloc` and `__sycl_usm_free` like it is done in `__result_and_scratch_storage `. You might want to benchmark this and analyze the performance benefits though.

SergeyKopienko: Unfortunately, the state under global atomic `__temp` used on host side to eval the return value:
```C++
template <typename _ExecutionPolicy, typename _Brick, typename... _Ranges>
bool
__parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Brick __f,
                   __parallel_or_tag /*__brick_tag*/, _Ranges&&... __rngs)
{
    // ...
    _AtomicType __result = __init_value;
    // ...

    {
        // ...
        auto __temp = sycl::buffer<_AtomicType, 1>(&__result, 1); // temporary storage for global atomic
        // ...
    }

    return __result != __init_value;
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,47 +1267,45 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
-                    __pred(__item_id, __n_iter, __wgroup_size, __comp, __found_local, __brick_tag, __rngs...);
-                    __dpl_sycl::__group_barrier(__item_id);
+                    __pred(__nd_item, __n_iter, __wgroup_size, __comp, __found_local, _BrickTag{}, __rngs...);
+                    // Point #A3.1
+                    __dpl_sycl::__group_barrier(__nd_item);
 
+                    // Point #A4
                     // Set local atomic value to global atomic
-                    if (__local_idx == 0 && __comp(__found_local.load(), __found.load()))
+                    if (__local_idx == 0)
```
SergeyKopienko, line 1295: I believe we are able to remove this extra `__comp(__found_local.load(), __found.load())` call
due it's already implemented in `for`-loop below.

SergeyKopienko:
I am going to remove all comments like ```// Point #``` after receive some reviews.
It may help to compare differences between some specializations for ```__parallel_or_tag``` and ```__parallel_find_forward_tag```, ```__parallel_find_backward_tag```

SergeyKopienko:
@MikeDvorskiy, how do you think, does it make sense to apply this changes for range's implementations too?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -646,24 +646,80 @@ __pattern_count(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator
...
+    {
+        return {__transform_op(std::get<0>(arg)), std::get<1>(arg)};
+    }
+};
+
+template <typename _Typle, typename _IsFirst>
+struct __find_if_binary_reduce_op
+{
+    _Typle
+    operator()(const _Typle& op1, const _Typle& op2) const
```
MikeDvorskiy, line 666: Probably, `_Typle `=> `_Tuple` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1028,52 +1030,91 @@ struct __early_exit_find_or
...
+            __group_idx * __wg_size * __n_iter + __leader * __n_iter + __local_idx % __shift;
+
+        // if our "line" is out of work group size, reduce the line to the number of the rest elements
+        if (__wg_size - __leader < __shift)
+            __shift = __wg_size - __leader;
+        for (_IterSize __i = 0; __i < __n_iter; ++__i)
+        {
+            // Point #B1 - not required to have _ShiftedIdxType
+
+            _IterSize __current_iter = __i;
```
adamfidel, line 1058: Could you use the loop induction variable `__i` directly in the following lines instead of introducing this new variable `__current_iter`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1084,11 +1124,93 @@ struct __early_exit_find_or
...
+    const auto __pred = oneapi::dpl::__par_backend_hetero::__early_exit_find_or<_ExecutionPolicy, _Brick>{__f};
+
+    // scope is to copy data back to __result after destruction of temporary sycl:buffer
+    {
+        sycl::buffer<_AtomicType, 1> __temp(&__result, 1); // temporary storage for global atomic
+
+        // main parallel_for
+        __exec.queue().submit([&](sycl::handler& __cgh) {
+            oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
+            auto __temp_acc = __temp.template get_access<access_mode::read_write>(__cgh);
```
SergeyKopienko, line 1172: Have anybody opinion, is it correct to use there ```write``` instead of ```read_write``` ?
It's compiled with ```write``` at this place but I am not sure, will it correct or no...
We apply for this atomic only one operation - ```fetch_or```.
Thanks.

julianmi: I believe `read_write` is correct since `fetch_or` will read the value of `__temp_acc` first, then perform the bitwise or with `__found_local`, and finally assigns the result to `__temp_acc`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1084,11 +1124,93 @@ struct __early_exit_find_or
 // parallel_find_or - sync pattern
 //------------------------------------------------------------------------
 
-// Base pattern for __parallel_or and __parallel_find. The execution depends on tag type _BrickTag.
+// Specialization for __parallel_or_tag
+template <typename _ExecutionPolicy, typename _Brick, typename... _Ranges>
+bool
+__parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Brick __f,
```
MikeDvorskiy, line 1130: Essentially, did you "split" `__parallel_find_or` pattern into two with "OR search logic" and "first/last search logic".
In that cause I would propose to have different names for patterns: `__parallel_find_or `and `__parallel_find_first`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1084,11 +1124,93 @@ struct __early_exit_find_or
...
+#endif
+                sycl::nd_range</*dim=*/1>(sycl::range</*dim=*/1>(__n_groups * __wgroup_size),
+                                          sycl::range</*dim=*/1>(__wgroup_size)),
+                [=](sycl::nd_item</*dim=*/1> __item_id) {
+                    __dpl_sycl::__atomic_ref<_AtomicType, sycl::access::address_space::global_space> __found(
+                        *__dpl_sycl::__get_accessor_ptr(__temp_acc));
+                    // Point #A1 - not required
+
+                    // Point #A2 - rewritten
+                    _AtomicType __found_local = __init_value;
```
MikeDvorskiy, line 1189: As far as I see `__found_local` variable is created and accessible just for the current "thread" (work item), it is not placed in SLM or global memory and is not shared between  work items. So, there is no concurrent access here and we dont nee use any atomic, just simple bool type is enough

SergeyKopienko: In global atomic type `bool` is absent.
So really it's one of integer types here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1084,11 +1124,93 @@ struct __early_exit_find_or
...
+                    // Point #A2.1 - not required
+
+                    // Point #A3 - rewritten
+                    constexpr auto __comp = typename _BrickTag::_Compare{};
+                    __pred(__item_id, __n_iter, __wgroup_size, __comp, __found_local, __brick_tag, __rngs...);
```
MikeDvorskiy, line 1194: if a solution is found, we the current work item may don't compute `__pred` any more with "find_or" algorithm perspective. Did you try to skip `__pred` computation and estimate performance change? I understand that in case of simple predicate such checks might lead to overheads... But in any case, can we get a performance gain here?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: @MikeDvorskiy, how about `__parallel_find_or` and `__parallel_find_entry` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1084,11 +1124,93 @@ struct __early_exit_find_or
...
+
+    auto __rng_n = oneapi::dpl::__ranges::__get_first_range_size(__rngs...);
+    assert(__rng_n > 0);
+
+    // TODO: find a way to generalize getting of reliable work-group size
+    auto __wgroup_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
+#if _ONEDPL_COMPILE_KERNEL
+    auto __kernel = __internal::__kernel_compiler<_FindOrKernel>::__compile(__exec);
+    __wgroup_size = ::std::min(__wgroup_size, oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel));
+#endif
```
danhoeflinger, line 1149: I think that this workaround `_ONEDPL_COMPILE_KERNEL` is unnecessary here. My understanding of this kernel compilation bundle step is that it is a workaround to prevent very large allocations from occurring on CPU targets which may return very large `__max_work_group_size` if left unchecked.  Since we are not allocating based upon workgroup size, I don't think this is needed in the first place.  

If we want some reasonable limit on workgroup size, this can be replaced by a simple "reasonable" maximum workgroup size like `1024` as is done in histogram.
```
    ::std::size_t __max_wgroup_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
    ::std::uint16_t __work_group_size = ::std::min(::std::size_t(1024), __max_wgroup_size);
```
I intend to investigate if we can move away from this workaround altogether in oneDPL, but I have not done so yet.  For the time being though I'd prefer not to propagate it unless we have some specific motivation for doing so which I am unaware of.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -997,25 +997,27 @@ struct __parallel_find_backward_tag
...
+    static constexpr _AtomicType __found_state = 1;
+    static constexpr _AtomicType __not_found_state = 0;
+
+    struct __compare_state
     {
-      public:
-        template <typename _LocalAtomic, typename _GlobalAtomic>
         bool
-        operator()(const _LocalAtomic& __found_local, const _GlobalAtomic& __found) const
+        operator()(const _AtomicType __found_local, const _AtomicType __found) const
```
danhoeflinger, line 1008: Am I correct that this is actually never called in the current PR?

the `__parallel_or_tag` overload takes `__comp` as a parameter but discards it without use.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1084,11 +1124,93 @@ struct __early_exit_find_or
...
+                    // Point #A3 - rewritten
+                    constexpr auto __comp = typename _BrickTag::_Compare{};
+                    __pred(__item_id, __n_iter, __wgroup_size, __comp, __found_local, __brick_tag, __rngs...);
+                    // Point #A3.1 - not required
+
+                    // Point #A4 - rewritten
+                    // Set found state result to global atomic
+                    if (__found_local != __init_value)
+                    {
+                        __found.fetch_or(__found_local);
```
MikeDvorskiy, line 1201: As far as I can see, here `__found_local` is not atomic, just a bool/int type and  each work item computes predicate and "touches" (updates) the global atomic only if the current predicate is true. I guess such approach may be effective if an input sequence has "few" solutions. 
I think if an input sequence has many solution, each time work item will "touch" global atomic to updated. An access to global atomic is very "expensive. I guess  we might a big performance penalty in that case.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: You are correct, fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: the standard and cppreference uses the "first" word. 
As  variant
`__parallel_find_first`
and 
`__parallel_find_any`

danhoeflinger:
As written, it seems like the `parallel_find_or` case seems to be doing a `||` reduction using global atomic variables to combine the results from different work items.  I would expect that for this to be better than the transform reduce pattern, we would need a real "early exit" check at the beginning of a workitem's work (or every `m` iterations) which could exit if the global atomic was `true`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Fixed: `__parallel_find_or` has been renamed to `__parallel_find_first` and `__parallel_find_any`
And some staff too.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Assuming we keep this strategy, I'd be in favor of using a different name for the type `_AtomicType` is a very confusing typename for something which is not being used as an atomic.  
I'd prefer to define two aliases for the same type for clarity here if you wanted to use `_AtomicType` for the global atomic, and have it share a type with the `_LocalStatusType`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: @MikeDvorskiy, what do you think about `_ONEDPL_COMPILE_KERNEL` and etc.?

MikeDvorskiy:
> a workitem's work (or every `m` iterations) which could exit if the global atomic was `true`.

Theoretically, it might be true. But, practically, access to a global atomic is very "expensive"..
image

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: My understanding was (is) that we want to know "right" maximus work group size, so the all processed values by WG will be fit into SLM memory. It leads to conclusion that it needs just for those kernels where we use SLM.

danhoeflinger:
> Theoretically, it might be true. But, practically, access to a global atomic is very "expensive".. 

I agree.  I'm not sure without running benchmarks what the best option is.  I'll elaborate more to explain the point of my comment. 

TLDR:
I think that we should call the transform reduce pattern here, and possibly make a specialization of reduce specific to `std::logical_or` / `std::logical_and` if that will help from a performance perspective which might include local and/or global "early exits". Also, I think the implementation as written makes the assumption that the data is "sparse" or mostly the identity for the operation, and we should not assume this. My guess is that if a global atomic is used in this specialization for an early exit, we will need to take more steps to reduce contention.

---
More details:

Right now, this implementation looks like a `std::logical_or` specialized atomic-based reduction kernel for commutative operations with a couple augmentations:
1) It checks for the identity (`false`), and opts out of touching the global atomic when it finds the identity.
2) It short circuits (or early exits) an individual workitems's work when it finds a `true`.

For (1)
Lets define an input data sequence to be "**sparse**" if each element has a very high likelihood to be the identity of the binary operation in a reduction operation. This reduction implementation should perform well with **sparse** data as it would limit congestion on the global atomic variable by opting out of touching the global variable when the reduction of all an individual workitem's reductions results in the identity (all data in a workitem is `false`). However, I don't know of a reason that we should assume input sequences to this pattern would be **sparse**. I would expect our normal (commutative) reduction algorithm to perform better for **non-sparse** data due to congestion on the atomic. In any case, I would still think we would want to use the commutative reduce pattern here.  Then, if necessary, improve the performance of the reduce pattern itself.

For (2)
Similarly to above, it could be interesting to consider if it is worthwhile to have some specialization for our reduction algorithm for `std::logical_and`, `std::logical_or` which takes advantage of an early exit (which is relevant for these specific reduce calls as well). The early exit at the work-item level (as done here) should be possible with our without atomics, but we should check that it helps. We could consider "early exit" at the subgroup or workgroup level, using `sycl::any_of_group` (or local atomics), this is also orthogonal to the question of using global atomics though.

One possible justification for the global atomic beyond an assumption of **sparse** data could be to utilize it as a global "early exit". However, that would need to be proven generally beneficial on a variety of input data. There are different approaches we could try here as well, for instance you could involve the global atomic only to implement the early exit, and not involve it in the return value. A single workitem could update this global atomic at the end of a workgroup's work, and workitems could check (read-only) this global atomic on startup, or periodically. I think there is a way we can reduce the contention impact of the "atomic reads" by using a relaxed memory order, as we don't really care **when** the read happens, but only that it is valid at a single point in time when it does happen. This is merely a signal to exit early, so if it is missed, it is not a correctness issue.

PR #1618, opened 2024-06-07
Work around the global policy issue in a backward-compatible manner

akukanov:
This patch proposes a different approach to address the issue #1060.

Previously we tried to re-implement the `device_policy` class in a way that would avoid eager initialization of a SYCL queue in the constructor, and instead would do it on the first use. That cannot be done however without having layout changes for the class, which we prefer to avoid for sake of preserving backward compatibility.

However, the issue #1060 does not really require lazy initialization; it asks for absence of exceptions or any other runtime errors in case a global SYCL queue object cannot be initialized. We use SYCL's default selector to create a queue used by the `dpcpp_default` predefined policy, which is supposed to **always** return a valid device because a SYCL implementation **must** provide at least one valid device. So our current implementation is formally correct, while the problem is observed in a strange environment where SYCL cannot find/initialize any device, even a CPU device; that seems to be a territory of undefined behavior.

Therefore a workaround can be implemented that detects during the construction of `dpcpp_default` that there are no SYCL devices and does not even try to initialize the queue, instead leaving the policy in an unusable state. According to the SYCL specification, the set of available platforms and devices does not change during the program lifetime, therefore our device policies, including `dpcpp_default`, cannot be used anyway (again, it's practically a UB territory) - so just silently ignoring the issue at global policy construction and letting it fire on the actual use seems OK.

The patch here attempts to implement this approach without breaking the class layout. For that, use of `sycl::queue` in the policies is replaced with a specially crafter `__queue_holder`, which has the same size and alignment as `sycl::queue` and uses "placement new" to construct a real queue in the space taken by the holder. In most cases, the queue is constructed unconditionally, but a special constructor for predefined policy instances checks if there is at least one device available. As an indication that the construction was omitted, the first `sizeof(void*)` bytes of the holder are nullified; the rationale is that a SYCL queue is likely implemented as a `shared_ptr` (that's certainly the case for DPC++), which in turn typically holds several pointers to the actual object, a service block or a reference counter, etc. It is highly unlikely therefore that the first bytes in a properly constructed queue object will be equivalent to a null pointer. That validity check, however, is only used in the destructor; all other methods assume that the queue was properly constructed, which is asserted in the `__queue_ref` method that returns a reference to the queue kept in the holder.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -24,6 +24,9 @@
 
 #include <type_traits>
 
+#include <cstddef> // std::byte, std::nullptr_t
+#include <cassert>
+
 namespace oneapi
 {
 namespace dpl
@@ -32,6 +35,81 @@ namespace execution
 {
 inline namespace __dpl
 {
+
+namespace __internal {
+
+struct __lazy_ctor_tag {};
+
+struct alignas(sycl::queue) __lazy_queue
+{
+    static_assert(sizeof(sycl::queue) >= sizeof(void*));
+    static_assert(alignof(sycl::queue) >= alignof(void*));
+
+    std::byte __buf[sizeof(sycl::queue)];
+
+    template <typename... _Args>
+    __lazy_queue(_Args... __args)
+    {
+        new(__buf) sycl::queue(__args...);
+    }
+
+    __lazy_queue(__lazy_ctor_tag)
+    {
+        new(__buf) std::nullptr_t;
+    }
+
+    ~__lazy_queue()
+    {
+        if (__has_queue())
+            __get_queue_ref().~queue();
+    }
+
+    bool __has_queue() const
+    {
+        return nullptr!=*reinterpret_cast<void* const*>(__buf);
+    }
+
+    const sycl::queue& __get_queue_ref() const
+    {
+        assert(__has_queue());
+        return *reinterpret_cast<sycl::queue const*>(__buf);
+    }
+    sycl::queue& __get_queue_ref()
+    {
+        assert(__has_queue());
+        return *reinterpret_cast<sycl::queue*>(__buf);
+    }
+
+    sycl::queue __get_queue_with_fallback(sycl::queue(*__f)()) const
+    {
+        return __has_queue()? __get_queue_ref() : __f();
+    }
+};
+
+inline sycl::queue
+__get_default_queue()
+{
+    static sycl::queue __q(sycl::default_selector_v);
+    return __q;
+}
+
+#if _ONEDPL_FPGA_DEVICE
+using __fpga_default_selector = 
+#if _ONEDPL_FPGA_EMU
+    __dpl_sycl::__fpga_emulator_selector;
+#else
+    __dpl_sycl::__fpga_selector;
+#endif
+
+inline sycl::queue
+__get_fpga_default_queue()
+{
+    static sycl::queue __q(__fpga_default_selector());
+    return __q;
+}
+#endif // _ONEDPL_FPGA_DEVICE
+
+} // namespace __internal
 
 struct DefaultKernelName;
 
@@ -53,15 +131,22 @@ class device_policy
     }
     explicit device_policy(sycl::queue q_) : q(q_) {}
     explicit device_policy(sycl::device d_) : q(d_) {}
-    operator sycl::queue() const { return q; }
+    explicit device_policy(__internal::__lazy_ctor_tag t_) : q(t_) {}
+
+    operator sycl::queue() const { return queue(); }
     sycl::queue
     queue() const
     {
-        return q;
+        return __get_queue_with_fallback(__internal::__get_default_queue);
     }
 
+  protected:
+    sycl::queue __get_queue_with_fallback(sycl::queue(*__f)())
+    {
+        return q.__get_queue_with_fallback(__f);
+    }
   private:
-    sycl::queue q;
+    __internal::__lazy_queue q;
 };
 
 #if _ONEDPL_FPGA_DEVICE
@@ -74,14 +159,7 @@ class fpga_policy : public device_policy<KernelName>
   public:
     static constexpr unsigned int unroll_factor = factor;
 
-    fpga_policy()
-        : base(sycl::queue(
-#    if _ONEDPL_FPGA_EMU
-              __dpl_sycl::__fpga_emulator_selector()
-#    else
-              __dpl_sycl::__fpga_selector()
-#    endif // _ONEDPL_FPGA_EMU
-                  ))
+    fpga_policy() : base(sycl::queue(__internal::__fpga_default_selector()))
     {
     }
 
@@ -89,8 +167,15 @@ class fpga_policy : public device_policy<KernelName>
     fpga_policy(const fpga_policy<other_factor, OtherName>& other) : base(other.queue()){};
     explicit fpga_policy(sycl::queue q) : base(q) {}
     explicit fpga_policy(sycl::device d) : base(d) {}
-};
+    explicit fpga_policy(__internal::__lazy_ctor_tag t) : base(t) {}
 
+    operator sycl::queue() const { return queue(); }
+    sycl::queue
+    queue() const
+    {
+        return base::__get_queue_with_fallback(__internal::__get_fpga_default_queue);
+    }
+};
 #endif // _ONEDPL_FPGA_DEVICE
 
 // 2.8, Execution policy objects
@@ -101,9 +186,9 @@ class fpga_policy : public device_policy<KernelName>
 // Starting with c++17 we can simply define sycl as inline variable.
 #    if _ONEDPL___cplusplus >= 201703L
 
-inline device_policy<> dpcpp_default{};
+inline device_policy<> dpcpp_default{__internal::__lazy_ctor_tag{}};
 #        if _ONEDPL_FPGA_DEVICE
-inline fpga_policy<> dpcpp_fpga{};
+inline fpga_policy<> dpcpp_fpga{__internal::__lazy_ctor_tag{}};
 #        endif // _ONEDPL_FPGA_DEVICE
 
 #    endif // _ONEDPL___cplusplus >= 201703L
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -24,8 +24,7 @@
 
 #include <type_traits>
 
-#include <cstddef> // std::byte
-#include <cstring> // memset
+#include <cstdint> // std::uintptr_t
 #include <atomic>  // atomic_signal_fence
 #include <cassert>
 
@@ -47,15 +46,16 @@ struct __global_instance_tag {};
 
 class alignas(sycl::queue) __queue_holder
 {
-    static_assert(sizeof(sycl::queue) >= sizeof(void*));
-    static_assert(alignof(sycl::queue) >= alignof(void*));
+    static_assert(sizeof(sycl::queue) >= sizeof(std::uintptr_t));
+    static_assert(sizeof(sycl::queue) % sizeof(std::uintptr_t) == 0);
+    static_assert(alignof(sycl::queue) >= alignof(std::uintptr_t));
 
-    std::byte __buf[sizeof(sycl::queue)];
+    std::uintptr_t __buf[sizeof(sycl::queue)/sizeof(std::uintptr_t)];
 
     bool __has_queue() const
     {
-        bool res = (nullptr != reinterpret_cast<void* const&>(*this));
-        std::atomic_signal_fence(std::memory_order_acq_rel); // mitigate possible reordering due to type punning
+        bool res = (__buf[0] == 0); // If the first size-of-pointer bytes are zeros, we consider there is no valid queue
+        std::atomic_signal_fence(std::memory_order_acq_rel); // to prevent possible reordering due to type punning
         return res;
     }
 
@@ -74,7 +74,7 @@ class alignas(sycl::queue) __queue_holder
         else
         {
             // an "impossible" case of SYCL providing no devices, which we however must handle
-            std::memset(this, 0, sizeof(void*));
+            __buf[0] = 0; // nullify the first size-of-pointer bytes of the holder
             // Since the queue holder does not have a valid queue in this case,
             // the predefined device policies are de-facto unusable.
             // That seems OK though, because there is no device to run SYCL anyway.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -160,17 +160,19 @@ class device_policy
 #if _ONEDPL_FPGA_DEVICE
 struct DefaultKernelNameFPGA;
 
-using __fpga_default_selector =
-#if _ONEDPL_FPGA_EMU
-    __dpl_sycl::__fpga_emulator_selector;
-#else
-    __dpl_sycl::__fpga_selector;
-#endif
-
 template <unsigned int factor = 1, typename KernelName = DefaultKernelNameFPGA>
 class fpga_policy : public device_policy<KernelName>
 {
     using base = device_policy<KernelName>;
+
+    static inline auto __fpga_default_selector()
+    {
+#if _ONEDPL_FPGA_EMU
+        return __dpl_sycl::__fpga_emulator_selector();
+#else
+        return __dpl_sycl::__fpga_selector();
+#endif
+    }
 
   public:
     static constexpr unsigned int unroll_factor = factor;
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +38,93 @@ namespace execution
...
+        bool res = nullptr!=reinterpret_cast<void* const&>(*this);
+        std::atomic_signal_fence(std::memory_order_acq_rel); // mitigate possible reordering due to type punning
+        return res;
+    }
+
+  public:
+    template <typename... _Args>
+    __queue_holder(_Args... __args)
+    {
+        new (this) sycl::queue(__args...);
```
SergeyKopienko, line 65: May be more proper way to use `&__buf[0]` or `std::addressof` instead of `this` in this place and in other places below?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +38,93 @@ namespace execution
...
+class alignas(sycl::queue) __queue_holder
+{
+    static_assert(sizeof(sycl::queue) >= sizeof(void*));
+    static_assert(alignof(sycl::queue) >= alignof(void*));
+
+    std::byte __buf[sizeof(sycl::queue)];
+
+    bool __has_queue() const
+    {
+        bool res = nullptr!=reinterpret_cast<void* const&>(*this);
```
SergeyKopienko, line 56: Looks like big trick! :-)

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: I believe `this`, `__buf` and `&__buf[0]` are all interchangeable as the address for placement new.
To my taste, using `this` rather than `__buf` in constructors and casts better conveys the idea that the holder instance is really interpreted as the queue; but it's subjective.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -101,9 +204,9 @@ class fpga_policy : public device_policy<KernelName>
 // Starting with c++17 we can simply define sycl as inline variable.
 #    if _ONEDPL___cplusplus >= 201703L
 
-inline device_policy<> dpcpp_default{};
+inline device_policy<> dpcpp_default{__internal::__global_instance_tag{}};
```
adamfidel, line 207: If `_ONEDPL_PREDEFINED_POLICIES` is not set, then the `__global_instance_tag` struct is not declared/defined. Would that be a problem for this line here and the fpga one below?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -53,44 +145,55 @@ class device_policy
     }
     explicit device_policy(sycl::queue q_) : q(q_) {}
     explicit device_policy(sycl::device d_) : q(d_) {}
-    operator sycl::queue() const { return q; }
+#if _ONEDPL_PREDEFINED_POLICIES
+    explicit device_policy(__internal::__global_instance_tag t_) : q(t_) {}
+#endl
```
SergeyKopienko, line 150: error : **invalid preprocessing directive**

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: The macro is checked at the line 97 above, just not shown in the diff by default.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +38,88 @@ namespace execution
...
+        if (this != &__h)
+            __queue_ref() = std::move(__h.__queue_ref());
+        return *this;
+    }
+
+    const sycl::queue& __queue_ref() const
+    {
+        assert(__has_queue());
+        return reinterpret_cast<const sycl::queue&>(*this);
+    }
```
danhoeflinger, line 113: With the assertion in `__queue_ref()`, and the use of `__queue_ref()` within assignment operators, copy and move constructors, we are basically saying if no device is available, you cannot move or copy a `__queue_holder` (with runtime assertions enabled) and therefore the same would go for the defaulted policies themselves.  
In principle, we could handle this case by moving / copying the zeroed out bytes.  

Is the idea that if no device is available, we shouldn't be doing anything the default policy at all? 
I suppose there is a small benefit in avoiding a branch as written by not supporting the "no device" case.

However, a user may believe that as long as they don't do something like call an algorithm with a policy, they are free to copy/move it around even without a device present.

akukanov: In order to copy or move a policy, it would first need to be created. But the creation will fail, except for the predefined policies. Therefore I think caring of copy/move operations for invalid policies is not worth the trouble.

danhoeflinger:
I ran a test similar to the one described here https://github.com/oneapi-src/oneDPL/pull/1154/files#r1317837347 using this branch.  The test is a check that we can include oneDPL headers without crashing when no device is available.  

On `origin/main`, when running the test, it passes when a sycl device is available, but crashes when `ONEAPI_DEVICE_SELECTOR='!*:*'` is specified with `PI_ERROR_DEVICE_NOT_FOUND`.  

On this branch, the test passes with no error with or without `ONEAPI_DEVICE_SELECTOR='!*:*'`.  It seems this resolves the issue successfully, if we are satisfied with the approach and if it succeeds in preserving the layout.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
danhoeflinger: So... how about a case like this: 
``` 
#include <sycl/sycl.hpp>

template <typename Policy>
void
function(Policy policy)
{
	std::vector<int> v{10,0};

	if (!sycl::device::get_devices().empty())
	{
		std::sort(policy, v.begin(), v.end());
	}
	else 
	{
		std::sort(v.begin(), v.end());
	}
}

int main()
{
	function(oneapi::dpl::execution::dpcpp_default);

    return 0;
}
```

Admittedly, I'm not sure this use case has much merit, but this is a case I could conceive of.  

A user has some code which can determine if its reasonable to call a parallel API based on if valid devices exist, but it is buried in their code somewhere.  At runtime, this would result in a valid serial sort being called when no device is available if no assertion was triggered.  With this branch, the assertion is triggered on copy of the policy into the function.  

Of course it could be argued that a user should just use `oneapi::dpl::execution::dpcpp_default` directly rather than passing it around, but I just want to make sure we are fully solving the problem we want to be solving with this fix.

If in these places we check `__has_queue()` first and `memset` to `0` in the `else`, we could run this code without issue.

akukanov: Passing the policy by reference, as in `template <typename Policy> void function(Policy& policy)`, would do the thing equally well, right? And still no other policy besides the default one can be used with `function`.

At some extra cost, we can have a more elaborated and perhaps more robust solution; e.g. the initial version of this patch tried to be somewhat more generic: https://github.com/oneapi-src/oneDPL/compare/main...b83ab2d
But I am not sure it makes much sense to do that way. If we want/need to have a better solution, I think we should bite the bullet and reimplement the class with real lazy-init behavior for all instances, accepting and properly handling backward incompatibility.

danhoeflinger: Yes, passing by reference resolves the issue in the example, and that is the norm for passing around policies.
I'm probably making an issue where one doesn't exist in practice, or is at least easily worked around. 

Once we are ready to merge, it might be good to reach out to the reporter of #1060 to confirm that this branch resolves the issue fully for their case, as it is an active example of someone experiencing this problem.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +38,88 @@ namespace execution
...
+class alignas(sycl::queue) __queue_holder
+{
+    static_assert(sizeof(sycl::queue) >= sizeof(void*));
+    static_assert(alignof(sycl::queue) >= alignof(void*));
+
+    std::byte __buf[sizeof(sycl::queue)];
+
+    bool __has_queue() const
+    {
+        bool res = (nullptr != reinterpret_cast<void* const&>(*this));
```
SergeyKopienko, line 57: What about using something like
```C++
union Data
{
    void* raw;
    std::byte __buf[sizeof(sycl::queue)];
};
```
Does this data structure able to help us simplify your line above?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +38,88 @@ namespace execution
...
+    }
+
+    const sycl::queue& __queue_ref() const
+    {
+        assert(__has_queue());
+        return reinterpret_cast<const sycl::queue&>(*this);
+    }
+    sycl::queue& __queue_ref()
+    {
+        assert(__has_queue());
```
SergeyKopienko, line 116: What this assertions will doing in release mode? Looks like as crash in release ma happen later...

SergeyKopienko: May be better to throw some exception instead of `assert` ?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,89 @@ namespace execution
...
+{
+    static_assert(sizeof(sycl::queue) >= sizeof(std::uintptr_t));
+    static_assert(sizeof(sycl::queue) % sizeof(std::uintptr_t) == 0);
+    static_assert(alignof(sycl::queue) >= alignof(std::uintptr_t));
+
+    std::uintptr_t __buf[sizeof(sycl::queue)/sizeof(std::uintptr_t)];
+
+    bool __has_queue() const
+    {
+        bool res = (__buf[0] == 0); // If the first size-of-pointer bytes are zeros, we consider there is no valid queue
```
danhoeflinger, line 57:
```suggestion
        bool res = (__buf[0] != 0); // If the first size-of-pointer bytes are zeros, we consider there is no valid queue
```
In the changes, this conditional got flipped.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: One of the design goals for the patch was to make it in a "pay-as-you-go" way - i.e. avoid having any impact from the workaround to the normal usage of valid device policies. I have to be convinced that throwing an exception for the sake of better diagnostic for a strange corner case is worth even that small extra check on the regular code path.

Moreover, if that extra check is deemed worthwhile, we might as well not throw anything but try creating/using a valid queue - again, see https://github.com/oneapi-src/oneDPL/compare/main...b83ab2d for a possible implementation.

danhoeflinger: Sure, I understand.  This is certainly an edge case, and to have all of oneDPL pay some penalty (even if it quite small) would require some real justification.  One could argue that it would not take much time to diagnose an error which is only present when no valid device is available, even if this worsens the message as compared to the original experience.

adamfidel: I often run into the situation where I am attempting to run a oneDPL benchmark / test when there is not a valid SYCL device available. This is due to testing multiple drivers / platforms and something not being installed or configured correctly on the system. Seeing the current error of `PI_ERROR_DEVICE_NOT_FOUND` makes it clear what the issue is and I believe that changing this behavior to instead be a segmentation fault could result in a worse experience.

If I look at it from the perspective of someone that is trying oneDPL for the first time and their system happens to be not configured correctly to make a SYCL device available, encountering a segmentation fault instead of a clearer error message might end up scaring that person away.

rarutyun: > If I look at it from the perspective of someone that is trying oneDPL for the first time and their system happens to be not configured correctly to make a SYCL device available, encountering a segmentation fault instead of a clearer error message might end up scaring that person away.

I think it can be mitigated by building it in DEBUG mode. Not super convenient but also not the end of the world. Anyway, if `default_selector_v` doesn't return a device we cannot rely on the SYCL implementation throwing an exception because strictly speaking, it's Undefined Behavior from the SYCL spec standpoint. We, of course, can invent our own exception but "should we?" is a big question to me because it increases the size of binary code and it's not zero-cost.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,89 @@ namespace execution
...
+    static_assert(sizeof(sycl::queue) >= sizeof(std::uintptr_t));
+    static_assert(sizeof(sycl::queue) % sizeof(std::uintptr_t) == 0);
+    static_assert(alignof(sycl::queue) >= alignof(std::uintptr_t));
+
+    std::uintptr_t __buf[sizeof(sycl::queue)/sizeof(std::uintptr_t)];
+
+    bool __has_queue() const
+    {
+        bool res = (__buf[0] != 0); // If the first size-of-pointer bytes are zeros, we consider there is no valid queue
+        std::atomic_signal_fence(std::memory_order_acq_rel); // to prevent possible reordering due to type punning
```
danhoeflinger, line 58: For my own understanding, can you explain why we need this protection against reordering here?

Could the compiler not realize that `__buf[0]` and the return of `__queue_ref()` are in fact the same and lose some inherent protections against reordering?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
danhoeflinger: Yes, I agree.  This is clearly outside the realm of the SYCL spec, and we can have a clear solution for users if they still encounter problems:  
If there is no device, you must avoid touching (copying, moving, calling APIs, ...)  `dpcpp_default` or `default_selector_v`.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: > Could the compiler not realize that __buf[0] and the return of __queue_ref() are in fact the same and lose some inherent protections against reordering?

Exactly. The C++ standard claims program behavior undefined if an object is accessed through an lvalue of an unrelated type (see [basic.lval] for details). That essentially allows compilers to assume "such access does not happen" and, in particular, do not even attempt any dependency analysis.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: I've attempted to improve the assertion diagnostics with a more descriptive predicate name and an attached message.
@SergeyKopienko @adamfidel @danhoeflinger please take a look again.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -201,7 +198,7 @@ class fpga_policy : public device_policy<KernelName>
 
 inline device_policy<> dpcpp_default{__internal::__global_instance_tag{}};
 #        if _ONEDPL_FPGA_DEVICE
-inline fpga_policy<> dpcpp_fpga{__internal::__global_instance_tag{}};
+inline fpga_policy<> dpcpp_fpga{};
```
akukanov, line 201: I have dropped the workaround from the predefined FPGA policy. The implementation would incorrectly create a default queue instead of an FPGA queue. Fixing it would require `__queue_holder` to search for FPGA devices, likely enabled by a special construction tag, etc. Given that the ONEDPL_FPGA_DEVICE mode is off by default and requires special environment to work, it seems unlikely that somebody faces the same issue with `dpcpp_fpga` as with `dpcpp_default`.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +36,92 @@ namespace execution
...
+        std::atomic_signal_fence(std::memory_order_acq_rel); // to prevent possible reordering due to type punning
+#endif
+        return res;
+    }
+
+  public:
+    template <typename... _Args>
+    __queue_holder(_Args... __args)
+    {
+        new (this) sycl::queue{__args...};
```
SergeyKopienko, line 68: std::forward ?

akukanov:
In order to address #1631, the approach proposed in the patch is not enough and the predefined policies should never initialize SYCL queues before the first use. I think it might still be doable in backward-compatible manner, but it makes sense to create a new PR for that.

PR #1619, opened 2024-06-07
Fixed a bullet list in the iterators topic.

dcbenito:
A sub-bullet list didn't have the correct white space for a sub-bullet list, I have fixed and tested the fix.


danhoeflinger (approved):
LGTM, thanks for the fix!

PR #1620, opened 2024-06-10
[test][Dynamic Selection] Refactor tests to account for offsets

AnuyaWelling2801:
Realized that the fixed resource tests for both sycl and inline backends did not account for offsets.
Fixed it by creating two new files in test/support for tests with and without offsets


SergeyKopienko:
Looks like we have a lot of duplicated code in the files
- test/support/test_dynamic_selection_offset_utils.h
- test/support/test_dynamic_selection_utils.h

And one significant diff between them that I found is:
- ```my_policy_t p{u};``` <-> ```my_policy_t p(u, offset);```
- ```auto test_resource = f(i);``` <-> ```auto test_resource = f(i, offset);```
- may by something else...

But I believe we have a lot of patterns to avoid this code duplications and to have only one copy of this code.

test/support/test_dynamic_selection_no_offset_utils.h
```diff
@@ -0,0 +1,228 @@
...
+    {
+        std::cout << "ERROR: did not select expected resources\n";
+        return 1;
+    }
+    std::cout << "submit_and_wait: OK\n";
+    return 0;
+}
+
+
+#endif /* _ONEDPL_TEST_DYNAMIC_SELECTION_UTILS_H */
```
SergeyKopienko, line 228: This comment doesn't correspond to file name.

SergeyKopienko:
@AnuyaWelling2801 Looks like we have some type which declared but never tested:
```C++
template <typename Policy>
using wait_t = typename policy_traits<Policy>::wait_type;
```
I mean `wait_t` type.

Could you please use this type in tests too?

timmiesmith:
Tests are outdated for the reworked dynamic selection API introduced in oneDPL 2022.10.0

PR #1621, opened 2024-06-11
[oneDPL] + oneapi::dpl::__internal::__compare

MikeDvorskiy:
This PR introduces oneapi::dpl::__internal::__compare which replace a lambda in hetero __parallel_stable_sort pattern

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1821,9 +1821,8 @@ auto
 __parallel_stable_sort(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _ExecutionPolicy&& __exec,
                        _Range&& __rng, _Compare __comp, _Proj __proj)
 {
-    auto __cmp_f = oneapi::dpl::__internal::__compare<_Compare, _Proj>{__comp, __proj};
     return __parallel_sort_impl(__backend_tag, ::std::forward<_ExecutionPolicy>(__exec), ::std::forward<_Range>(__rng),
-                                __cmp_f);
+                                oneapi::dpl::__internal::__compare<_Compare, _Proj>{__comp, __proj});
 }
 
 //------------------------------------------------------------------------
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1821,9 +1821,7 @@ auto
 __parallel_stable_sort(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _ExecutionPolicy&& __exec,
                        _Range&& __rng, _Compare __comp, _Proj __proj)
 {
-    auto __cmp_f = [__comp, __proj](const auto& __a, const auto& __b) mutable {
-        return __comp(__proj(__a), __proj(__b));
-    };
+    auto __cmp_f = oneapi::dpl::__internal::__compare<_Compare, _Proj>{__comp, __proj};
     return __parallel_sort_impl(__backend_tag, ::std::forward<_ExecutionPolicy>(__exec), ::std::forward<_Range>(__rng),
                                 __cmp_f);
 }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
@@ -58,6 +58,9 @@ class __equal_value;
 template <typename _Tp>
 class __not_equal_value;
 
+template <typename _Comp, typename _Proj>
+class __compare;
+
 template <typename _Pred>
 class __transform_functor;
 
@@ -132,6 +135,12 @@ struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::
 template <typename _Tp>
 struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::__not_equal_value, _Tp)>
     : oneapi::dpl::__internal::__are_all_device_copyable<_Tp>
+{
+};
+
+template <typename _Comp, typename _Proj>
+struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::__compare, _Comp, _Proj)>
+    : oneapi::dpl::__internal::__are_all_device_copyable<_Comp, _Proj>
 {
 };
 
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -129,12 +129,11 @@ class __pstl_assign
 };
 
 template <typename _Comp, typename _Proj>
-class __compare
+struct __compare
 {
     _Comp __comp;
     _Proj __proj;
 
-  public:
     template <typename _Xp, typename _Yp>
     bool
     operator()(_Xp&& __x, _Yp&& __y)
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -128,6 +128,21 @@ class __pstl_assign
     }
 };
 
+template <typename _Comp, typename _Proj>
+class __compare
+{
+    _Comp __comp;
+    _Proj __proj;
+
+  public:
+    template <typename _Xp, typename _Yp>
+    bool
+    operator()(_Xp&& __x, _Yp&& __y)
+    {
+        return __comp(__proj(std::forward<_Xp>(__x)), __proj(std::forward<_Yp>(__y)));
+    }
+};
+
 //! "==" comparison.
 /** Not called "equal" to avoid (possibly unfounded) concerns about accidental invocation via
     argument-dependent name lookup by code expecting to find the usual ::std::equal. */
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -131,14 +131,15 @@ class __pstl_assign
 template <typename _Comp, typename _Proj>
 struct __compare
 {
-    _Comp __comp;
-    _Proj __proj;
+    //'mutable' is to relax the requirements for a user comparator or/and projection type operator() may be non-const
+    mutable _Comp __comp;
+    mutable _Proj __proj;
 
     template <typename _Xp, typename _Yp>
     bool
-    operator()(_Xp&& __x, _Yp&& __y)
+    operator()(const _Xp& __x, const _Yp& __y) const
     {
-        return __comp(__proj(std::forward<_Xp>(__x)), __proj(std::forward<_Yp>(__y)));
+        return __comp(__x, __y);
     }
 };
 
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -139,7 +139,7 @@ struct __compare
     bool
     operator()(const _Xp& __x, const _Yp& __y) const
     {
-        return __comp(__x, __y);
+        return __comp(__proj(__x), __proj(__y));
     }
 };
 


include/oneapi/dpl/pstl/utils.h
```diff
@@ -128,6 +128,21 @@ class __pstl_assign
     }
 };
 
+template <typename _Comp, typename _Proj>
+class __compare
+{
+    _Comp __comp;
+    _Proj __proj;
```
danhoeflinger, line 135: I believe you need a constructor with this as a class and these as private member variables.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -128,6 +128,21 @@ class __pstl_assign
...
+
+  public:
+    template <typename _Xp, typename _Yp>
+    bool
+    operator()(_Xp&& __x, _Yp&& __y)
+    {
+        return __comp(__proj(std::forward<_Xp>(__x)), __proj(std::forward<_Yp>(__y)));
+    }
```
akukanov, line 143: Why not keep `const&` parameters as in the original lambda? I believe neither comparators nor projections are supposed to modify their arguments.

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: renaned to struct

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: good question. I intended keep const&, but made approach at least  consistency with the other similar predicates (__pstl_equal, __pstl_less, etc).
The next question why the other similar predicates accept T&&, not const T&... Probably, we had a compile issue with some tricky types (a kind proxy types etc). I will try to figure out it.

MikeDvorskiy: Actually, I've remembered a case when we had a compilation issue when we had changed `T&&` by const `T&`. It was the binary_search.pass test. That code tries to call a lambda with T& argument, which is const T& in its turn.

So, keeping `const T&` for argument of `__compare` doesn't matche to operator `__proj`, which can be implemented by user as`operator()(T&) {}`. 

So, keeping `T&` we make oneDPL more robust relative to a user projection type definition.

danhoeflinger: If I follow correctly, you are saying that using forwarding references here allows us to accommodate the case where a user provides a projection operator which takes as input a **non-const lvalue ref** (`T&`).  

The question then follows: is this a good thing?  

Our previous code did not enable this, it would be a compile time error.  If a user's projection operator actually modifies the input reference, I believe that would result in UB in our sort routine (modified inputs every time the projection is applied).  Conceptually projection operators shouldn't modify the input.  So, what we are enabling is "convenience" for the user of not adding a `const` to their input params, and possibly letting a compile time error become a runtime error if their projection actually does modify the input.

akukanov: C++17 states that predicates (unary and binary) as well as comparators "shall/will not apply any non-constant function through the dereferenced iterator". For parallel algorithms, it additionally states that unary and binary operations also "shall not directly or indirectly modify objects via their arguments".

In C++20, projections, comparators, predicates etc. for range-based algorithms are required to satisfy the `regular_invocable` concept. That concept states that the function invocation "shall not modify the
function object or the arguments."

That's what I meant above saying that "neither comparators nor projections are supposed to modify their arguments." Even if user-provided function objects can technically have non-constant call operators/parameters, its only benefit seems to allow users "forget" const qualifiers - at the risk of having runtime errors, like Dan said above.

Unfortunately, both std libraries and Thrust are tolerant to function objects omitting `const`, and this is the main reason why we may need to do the same.

MikeDvorskiy: So, if we want to keep `T&&`, I think it should be consistency with the similar predicates in the same file, at least. I mean file `utils.h` and `__pstl_equal`, `__pstl_less`, `__pstl_greater` etc.

MikeDvorskiy: One one hand, in "narrow sight" I understand: We had a customer issue with  "device copyable", found a reason as lambda-comparator with the captured user comparator projection and "`const T&` call convention". Looking just only that, no considering the situation in general, yes we can just keep `const T&`, as we had in the mentioned "lambda" version. It can be one commit (one PR).
On the other hand,
1) I don't like to make "the inconsistency" bigger  and to "seed" a potential oneDPL issue when a user  will pass "the comparators / projection operators to allow non-const qualified lvalue ref parameters". I understand that it was before, with lambda,  but any any case - if we change that place  - lets looking in the future.
2) Moreover, having that inconsistency in the similar predicates in this file confuses the others oneDPL contributors. Somebody might add a new predicate with `T&` someone  - with `const T&`. The mess will be increased.
3) Also that inconsistency will confuse the reviewers when we merge it to LLVM upstream. 

As an option, that I would propose  - 1) to merge the "device copyable" fix for `__parallel_stable_sort` with `const T&`, as it was in lambda. 2) right after that - the next change: to eliminate the mentioned inconsistency for the all predicates, at least "comparing" type.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -128,6 +128,20 @@ class __pstl_assign
     }
 };
 
+template <typename _Comp, typename _Proj>
+struct __compare
```
danhoeflinger, line 132: I wonder if this should be named `__project_and_compare` or `__compare_with_projection`, or `__projected_compare`
`__compare` seems too general, and doesn't actually describe the purpose behind this structure.  The reason one would use this is to apply the projection as input to a comparison.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: I'm OK with this plan.  I'll note that if we make this change and release, it becomes much more difficult to revert this change since it could then "break" working user code. 

Do you intend to make changes for more alignment, or do you think all that is required is contained in the existing PR?

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: Actually, I thought about that... about the similar names that you mention...
But came to conclusion:
1) any of them is too long.
2) consistency reasons:  we have many  other functor with additional fields like predicate or/and value and we don't  mention that fields in the functor name. We dont write  `__replace_functor` `_with_predicate_and_value`...
3) as for me I think even `__replace_functor `or `__transform_functor `or `__pstl_less` is too long and "functor" (or pstl) word doesn't add any sense:) it's the same if we write `transform_struct `or `transform_class` or `sort_function` :) The fact that `__transform_functor`  is a functor so is clear from context call transform{pred}() or class definition. So, in my opinion `__transform`, `__compare` etc is absolutely enough. Moreover we have ones in the special internal namespace, if you scare about a name conflict.
4) The other libraries also don't  use a kind of additional suffix word for the predicates: sycl::plus, sycl::bit_or etc.

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: 1) done
2) It seems we tend to relax the requirement  - to use `T&&` instead for `const T&` for input parameter - so, it doesn't  break working user code.

danhoeflinger: I think you misunderstood what I'm saying.  
I agree this PR does not break working user code.  The purpose of my comment is to state that it is "easy" to go in this direction.  However, if we later were to change our minds and want to go back to `const T&`, from `T&&`, that could break working user code, so we would not make that change. 
I'm just mentioning it to make it clear that we should be sure about our decision because it would be difficult to revert.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: I'm not worried about a name conflict.  I just think it is better for understanding the code later, but its just my preference.  Especially with this being internal, I prefer more descriptive names to concise ones. 

I wont insist, not a blocker.

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: I see your point, Dan....
Alexey (@akukanov),  It seems we need a decision maker here.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -128,6 +128,21 @@ class __pstl_assign
...
+{
+    //'mutable' is to relax the requirements for a user comparator or/and projection type operator() may be non-const
+    mutable _Comp __comp;
+    mutable _Proj __proj;
+
+    template <typename _Xp, typename _Yp>
+    bool
+    operator()(const _Xp& __x, const _Yp& __y) const
+    {
+        return __comp(__x, __y);
```
adamfidel, line 142: Should we call the projection operator on `__x` and `__y` before passing to the comparator?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1821,9 +1821,7 @@ auto
 __parallel_stable_sort(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _ExecutionPolicy&& __exec,
                        _Range&& __rng, _Compare __comp, _Proj __proj)
 {
-    auto __cmp_f = [__comp, __proj](const auto& __a, const auto& __b) mutable {
-        return __comp(__proj(__a), __proj(__b));
-    };
+    auto __cmp_f = oneapi::dpl::__internal::__compare<_Compare, _Proj>{__comp, __proj};
     return __parallel_sort_impl(__backend_tag, ::std::forward<_ExecutionPolicy>(__exec), ::std::forward<_Range>(__rng),
                                 __cmp_f);
```
adamfidel, line 1826: Could we pass the object by move here (i.e., `std::move(__cmp_f)`) or is there a downside to that?

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: sure.. Good catch, thank you, Adam!

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: good question...
In my opinion,  passing an object by move makes sense just for the stateful functors. Due to in the parallel calls the stateless functors are used, usage of `std::move` just makes the code more complicated and less readable... But, I open to discuss it.

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: One more argument in favor of `T&&` - std::invoke accepts arguments as `Args&&... args`

akukanov: I am OK with the current patch. As for the plan - I do not object to revisiting the content of `pstl/utils/h` but I do not think it should be done for the same milestone.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: I would instead create `__internal::__compare` directly as a temporary to pass to the function (i.e. without assigning to a variable).

danhoeflinger: Ready to improve with this suggested change.

mmichel11:
I've looked through this PR as well as the parameter discussion and the current version of the patch looks good to me (assuming the minor outstanding comment is addressed).

There is a CI failure but it looks to be from an outdated mac configuration with the no longer supported `icpc` compiler that we have replaced to test with `clang++`. I think we should just make sure to get a clean per-commit CI run before merging.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger, line 1826:
```suggestion
    return __parallel_sort_impl(__backend_tag, ::std::forward<_ExecutionPolicy>(__exec), ::std::forward<_Range>(__rng),
                                oneapi::dpl::__internal::__compare<_Compare, _Proj>{__comp, __proj});
```

I believe this is the change that @akukanov  is suggesting.  If we commit it, we also get the rerun of the per-commit CI that Matt described below.

danhoeflinger (approved):
LGTM, 
I still prefer the version that adds the change suggested in [this comment ](https://github.com/oneapi-src/oneDPL/pull/1621/files#r1711317435) from @akukanov (I suggest it [as a commit here](https://github.com/oneapi-src/oneDPL/pull/1621/files#r1736874752) for convenience), but I don't see it as a blocker, so I wont withhold my approval.

MikeDvorskiy:
@danhoeflinger , could you please re-approve (and merge) after your suggestion was committed?

danhoeflinger (approved):

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -58,6 +58,9 @@ class __equal_value;
 template <typename _Tp>
 class __not_equal_value;
 
+template <typename _Comp, typename _Proj>
+class __compare;
```
dmitriy-sobolev, line 62: class -> struct to align with the definition.

PR #1622, opened 2024-06-11
Performance improvements of `__pattern_any_of`, `__pattern_find_if`

SergeyKopienko:
In this PR we made some performance improvements:
- `__pattern_any_of(__hetero_tag` has been implemented on `__parallel_transform_reduce`;
- `__pattern_find_if(__hetero_tag` has been implemented on `__parallel_transform_reduce` (gave us the same results as without this change but with https://github.com/oneapi-src/oneDPL/pull/1617).

My experiments shows significant performance boost for subset of algorithms after these changes.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -670,19 +670,44 @@ bool
 __pattern_any_of(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator __first, _Iterator __last,
                  _Pred __pred)
 {
-    if (__first == __last)
+    using _difference_type = typename ::std::iterator_traits<_Iterator>::difference_type;
+
+    const _difference_type __n = __last - __first;
+    if (__n == 0)
         return false;
 
-    using _Predicate = oneapi::dpl::unseq_backend::single_match_pred<_ExecutionPolicy, _Pred>;
+    using _result_type = oneapi::dpl::__internal::tuple<bool, _difference_type>;
+    const auto __init = _result_type{false, __n};
 
-    auto __keep = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
-    auto __buf = __keep(__first, __last);
+    // __counting_iterator_t - iterate position (index) in source data
+    using __counting_iterator_t = oneapi::dpl::counting_iterator<_difference_type>;
 
-    return oneapi::dpl::__par_backend_hetero::__parallel_find_or(
-        _BackendTag{},
-        __par_backend_hetero::make_wrapped_policy<__par_backend_hetero::__or_policy_wrapper>(
-            ::std::forward<_ExecutionPolicy>(__exec)),
-        _Predicate{__pred}, __par_backend_hetero::__parallel_or_tag{}, __buf.all_view());
+    using _zipped_data_type = typename std::iterator_traits<decltype(oneapi::dpl::make_zip_iterator(
+        __first, __counting_iterator_t{0}))>::value_type;
+
+    __find_if_binary_reduce_op<_zipped_data_type, /*_IsFirst*/ std::true_type> __reduce_op;
+    __find_if_unary_transform_op<_zipped_data_type, _Pred> __transform_op{__pred};
+
+    using _Functor = unseq_backend::walk_n<_ExecutionPolicy, decltype(__transform_op)>;
+    using _RepackedTp = __par_backend_hetero::__repacked_tuple_t<_result_type>;
+
+    auto __keep_src_data =
+        oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -715,16 +771,44 @@ _Iterator
 __pattern_find_if(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator __first, _Iterator __last,
```
SergeyKopienko, line 771: __pattern_find_if impl on __parallel_transform_reduce

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -646,24 +646,80 @@ __pattern_count(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator
...
+                return {true, std::max(std::get<1>(op1), std::get<1>(op2))};
+        }
+
+        return std::get<0>(op1) ? op1 : op2;
+    }
+};
+
 template <typename _BackendTag, typename _ExecutionPolicy, typename _Iterator, typename _Pred>
 bool
 __pattern_any_of(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator __first, _Iterator __last,
```
MikeDvorskiy, line 682: @SergeyKopienko  did you check a performance gain after your previous perf improvement for `oneapi::dpl::__par_backend_hetero::__parallel_find_or` pattern?
(https://github.com/oneapi-src/oneDPL/pull/1617)

My question here is: which base pattern is better (with performance perspective) for `__pattern_any_of`?
The improved `__parallel_find_or` or `__parallel_transform_reduce`?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -646,24 +646,80 @@ __pattern_count(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator
...
                  _Pred __pred)
 {
```
MikeDvorskiy, line 684: With code design perspective I would introduce some new (oneDPL middle patterns level) range based pattern, which would recall range based `oneapi::dpl::__par_backend_hetero::__parallel_transform_reduce`, and be shared between iterator based `__pattern_any_of` and range based `__pattern_any_of`.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: I have measured this case.
The overall performance is the same for these two approaches.
But in details there are some differences.

MikeDvorskiy, line 682: BTW, if we have an intention to use "reduction" we can use a functional approach, shorter, readable and easer to further support.
For, example "any_of"  can be expressed via 
`reduce(data | transform([pred](auto x) { return pred(x);}), false, std::logical_or<bool>{});`
Please, have a look: https://godbolt.org/z/csMfaE54h

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -646,24 +646,80 @@ __pattern_count(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator
...
         return false;
 
-    using _Predicate = oneapi::dpl::unseq_backend::single_match_pred<_ExecutionPolicy, _Pred>;
+    using _result_type = oneapi::dpl::__internal::tuple<bool, _difference_type>;
+    const auto __init = _result_type{false, __n};
 
-    auto __keep = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
-    auto __buf = __keep(__first, __last);
+    // __counting_iterator_t - iterate position (index) in source data
+    using __counting_iterator_t = oneapi::dpl::counting_iterator<_difference_type>;
```
danhoeflinger, line 695: I'm not following the reason to include the counting iterator and the rest of this that tracks the index of the output for `__pattern_any_of`, which only returns the `true` or `false` information, and discards the index information.  It seems we are adding lots of complexity and computation.
If we want to handle this with a transform_reduce, it seems like we can avoid the counting iterator, zipping, tuples, etc. 
 and just have a reduction operation `std::logical_and`.  

I should note that I'm not sure a reduction is what we want here necessarily.  There may be some better "early exit" routine which checks for `true` and short circuits remaining work if it encounters one.  As discussed though, this would need to balance contention on some synchronization mechanism with the potential savings.

SergeyKopienko: Good, point, fixed.

SergeyKopienko:
Our `parallel_find_or`-based algorithms implementation is more perform now then `transform_reduce`-based implementation.
So this PR is not required anymore.

PR #1624, opened 2024-06-17
Performance improvements: insert `break` into `__early_exit_find_or`

SergeyKopienko:
In this error we fixing performance issue in the `__early_exit_find_or::operator()` :
- we should `break` `for`-loop after call `__found_local.store(1);`

This performance issue has been found during preparation of the PR https://github.com/oneapi-src/oneDPL/pull/1617


MikeDvorskiy (approved):
LGTM

danhoeflinger (approved):
Very good improvement.  
I didn't understand at first, but because don't use SLM, and we limit the number of groups to the number of compute units, this allows `__n_iter` to grow with `N`, basically unbounded.  This means that even though we likely need *ALL* work-items in a subgroup to see the early exit condition to get a benefit here (due to implementation as masked vector operations), we might have a very large `__n_iter`, and with data that has a reasonable chance to find the criteria throughout, this can still provide big savings.

PR #1627, opened 2024-06-18
[KT][Tests] Remove non-supported cases from testing (ESIMD sort)

dmitriy-sobolev:
`oneapi::dpl::experimental::kt::gpu::esimd_radix_sort*` currently supports only `workgroup_size` equal to `64`. There are no short-term plans to extend it with `32`. Let's remove `32` from testing.


danhoeflinger (approved):
LGTM

PR #1586, opened 2024-05-15
[Doc] Split radix_sort doc page into key and key-value API

dmitriy-sobolev:
Separate `radix_sort` and `radix_sort_by_key`.

Pros:

- Improved readability: no branching for a specific case, less number of interfaces per page.
- Examples do not take too much space.

Cons:
- Duplication, more to maintain. 

I assume that benefits outweigh drawbacks specifically from a user perspective.

This PR was motivated by this discussion: https://github.com/oneapi-src/oneDPL/pull/1583#discussion_r1601293197

PR #1629, opened 2024-06-20
[KT] Rename out-of-place esimd radix_sort as radix_sort_copy

danhoeflinger:
It is the convention of the standard library when both an in-place and out-of-place algorithm are provided to to label out-of-place versions of APIs as `*_copy`.  See `partial_sort[_copy]`, `remove[_copy]`, `partition[_copy]`, etc..

This PR deprecates the existing out-of-place overloads for `radix_sort` and `radix_sort_by_key` in favor of `radix_sort_copy` and `radix_sort_by_key_copy`.  We originally avoided this to limit the line length with the large KT namespace, but the convention is strong and consistent enough to warrant this change.

This updates the code, the testing, and the documentation.

This PR is based on #1586 , and should be merged into main after that PR is merged. For now I will point it at that PR, to properly show the diff.

### DIFF
--- a/documentation/library_guide/kernel_templates/esimd/radix_sort_by_key.rst
+++ b/documentation/library_guide/kernel_templates/esimd/radix_sort_by_key.rst
@@ -5,13 +5,14 @@ Radix Sort By Key
 radix_sort_by_key Function Templates
 ------------------------------------
 
-The ``radix_sort_by_key`` function sorts keys using the radix sort algorithm, applying the same order to the corresponding values.
+The ``radix_sort_by_key`` and ``radix_sort_by_key_copy`` functions sort keys using the radix sort algorithm,
+applying the same order to the corresponding values.
 The sorting is stable, preserving the relative order of elements with equal keys.
-Both in-place and out-of-place overloads are provided. Out-of-place overloads do not alter the input sequences.
+Both in-place and out-of-place APIs are provided. Out-of-place APIs do not alter the input sequences.
 
 The functions implement a Onesweep* [#fnote1]_ algorithm variant.
 
-A synopsis of the ``radix_sort_by_key`` function is provided below:
+A synopsis of the ``radix_sort_by_key`` and ``radix_sort_by_key_copy`` functions are provided below:
 
 .. code:: cpp
 
@@ -39,7 +40,7 @@ A synopsis of the ``radix_sort_by_key`` function is provided below:
              typename ValuesIterator1, typename KeysIterator2,
              typename ValuesIterator2>
    sycl::event
-   radix_sort_by_key (sycl::queue q, KeysIterator1 keys_first,
+   radix_sort_by_key_copy (sycl::queue q, KeysIterator1 keys_first,
                       KeysIterator1 keys_last, ValuesIterator1 values_first,
                       KeysIterator2 keys_out_first, ValuesIterator2 values_out_first,
                       KernelParam param); // (3)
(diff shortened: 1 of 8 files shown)


documentation/library_guide/kernel_templates/esimd/radix_sort_by_key.rst
```diff
@@ -39,7 +40,7 @@ A synopsis of the ``radix_sort_by_key`` function is provided below:
              typename ValuesIterator1, typename KeysIterator2,
              typename ValuesIterator2>
    sycl::event
-   radix_sort_by_key (sycl::queue q, KeysIterator1 keys_first,
+   radix_sort_by_key_copy (sycl::queue q, KeysIterator1 keys_first,
```
akukanov, line 43: Should it be `radix_sort_copy_by_key`? I feel it reads slightly more natural.
In the standard, a similar composite suffix is `_copy_if`.

danhoeflinger: I prefer `radix_sort_by_key_copy`, but I'm open for discussion on it.
 
With the current formulation, the name describes the semantics as:  
(radix sort by key) copy          or "sort the data by key, then copy the result"

With the proposed formulation, our intention would be for it to mean: 
(radix sort copy) ... by key

However it also could be read by a user as :
(radix sort) (copy by key) or "radix sort the data, then modify the copy of the data with the key somehow"  possibly it could mean that it copies the sorted data to unique spots based on the key, or masked by key somehow. 

The meaning seems more ambiguous to me based on the punctuation the reader inserts in the second formulation.  Of course, you could also say that a user could read for the first one:  (radix sort) and (by key copy), but that seems much more of a stretch to me.

akukanov: Let's try to make some analogy for existing C++ algorithms with the "copy" flavor. I'll start with `remove` (for which I could imagine `remove_by_key`).

There is `remove(first, last, value);` that removes all elements equal to the value, and `remove_if(first, last, predicate)` that removes all elements for which the predicate evaluates to true. And, there is `remove_copy` and `remove_copy_if` which have an extra parameter for the output sequence where the not-to-be-removed elements are copied.

Now let's imagine `remove_by_key(key_first, key_last, val_first, key)` that removes from both sequences all items with the matching key. In principle, `_by_key` here indicates a condition, similar to how `_if` does. So following the same principle of ordering as for `_if`, I would name a hypothetical out-of-place variant of that as `remove_copy_by_key`.

We can also imagine `unique[_copy]_by_key` that would remove all duplicate keys and corresponding values, and even `copy_by_key` that would copy values which keys match to the given one. And I think in all cases the other order - `remove_by_key_copy`, `unique_by_key_copy`, and especially `by_key_copy` :) - would be worse.

danhoeflinger: Yeah its a fair argument.  `remove_copy_if` is the best example here to convince me.  I think its more straightforward with that API though because `remove_copy` is a much simpler single operation than `radix_sort_copy`.  With the wording of `radix_sort_copy`, is is difficult for me to think of it (semantically) as a single operation to be modified then `by_key`, but rather a pair of operations which happen in sequence.  You are correct though that `radix_sort_copy_by_key` is still the most consistent with the current standard.

I'll make the change.

danhoeflinger: Alright, it should be fixed now.

PR #1630, opened 2024-06-20
[CI] Fix FPGA emulator installation

dmitriy-sobolev:
The issue started happening after switching to oneAPI 2024.2. The root cause is currently unclear.

Upd. (June 21):
The issue is fixed by replacing: `intel-oneapi-runtime-dpcpp-sycl-fpga-emul` -> `intel-oneapi-compiler-fpga`.
I communicated with the maintainers of the packages and this is expected behaviour. FPGA emulator is not a part of oneAPI since 2024.2 and should be installed through `intel-oneapi-compiler-fpga`.


timmiesmith (approved):
LGTM

PR #1632, opened 2024-06-21
[PROTOTYPE] Add support for iterators / buffers / ranges to two-pass scan

adamfidel:
This PR adds support for iterators / buffers / ranges to the prototype two-pass scan algorithm.


mmichel11 (approved):
LGTM

PR #1634, opened 2024-06-23
Removing requirement that all commits be signed

timmiesmith:
This is being done to be consistent with LLVM and UXL Foundation components. Contributions will be under the Apache License regardless of signature.  This addresses https://github.com/uxlfoundation/open-source-working-group/issues/44


ValentinaKats (approved):
LGTM

PR #1637, opened 2024-06-24
[PROTOTYPE] Improve tests for two-pass scan

adamfidel:
This PR adds tests for all of the numeric types for the two-pass scan prototype.


mmichel11 (approved):
LGTM

PR #1638, opened 2024-06-25
[PROTOTYPE] Two-pass scan - Generalize work-group size support

mmichel11:
This PR is targeted to the prototype branch for the reduce-then-scan implementation.

Previously, the implementation was reliant on work-group sizes being >= 1024 so that the number of sub-groups is divisible by the sub-group size. This PR generalizes the work-group size requirement and allows for the case where the number of sub-groups is not divisible by the sub-group size. There is a small performance penalty when this is the case.

Additionally, support has been added for performing sub-group scans when we do not have a full sub-group size of input via the `num_remaining` parameter and avoids the need to pad with identity elements.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/two_pass_scan.h
+++ b/include/oneapi/dpl/experimental/kt/two_pass_scan.h
@@ -253,11 +253,11 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
                         }
                         // If we are past the input range, then the previous value of v is passed to the sub-group scan.
                         // It does not affect the result as our sub_group_scan will use a mask to only process in-range elements.
-                        if (sub_group_local_id < num_sub_groups_local)
+                        if (i * VL + sub_group_local_id < num_sub_groups_local)
                             v = sub_group_partials[sub_group_local_id];
                         std::tie(v, sub_group_carry) = sub_group_scan<VL, true>(
                             sub_group, std::move(v), binary_op, std::move(sub_group_carry), num_sub_groups_local);
-                        if (sub_group_local_id < num_sub_groups_local)
+                        if (i * VL + sub_group_local_id < num_sub_groups_local)
                             tmp_storage[start_idx + i * VL + sub_group_local_id] = v;
                     }
                 }
@@ -338,7 +338,7 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
                             sub_group_partials[i * VL + sub_group_local_id] =
                                 tmp_storage[csrc + i * VL + sub_group_local_id];
                         }
-                        if (sub_group_local_id < num_sub_groups_local)
+                        if (i * VL + sub_group_local_id < num_sub_groups_local)
                         {
                             sub_group_partials[i * VL + sub_group_local_id] =
                                 tmp_storage[csrc + i * VL + sub_group_local_id];
@@ -399,7 +399,7 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
                                 binary_op(adj_work_group_carry, sub_group_partials[carry_offset + sub_group_local_id]);
                             carry_offset += VL;
                         }
-                        if (sub_group_local_id < num_sub_groups_local)
+                        if (i * VL + sub_group_local_id < num_sub_groups_local)
                         {
                             sub_group_partials[carry_offset + sub_group_local_id] =
                                 binary_op(adj_work_group_carry, sub_group_partials[carry_offset + sub_group_local_id]);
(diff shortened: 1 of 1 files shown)


danhoeflinger:
We still need to come up with some strategy for "filling" elements because we need some constructed element in the `shift_group_right` call. Since we are not "using" the shifted value, we can possibly fill it with "garbage" or the last valid element in the subgroup.  I'm thinking about how to do this.

mmichel11:
> We still need to come up with some strategy for "filling" elements because we need some constructed element in the `shift_group_right` call. Since we are not "using" the shifted value, we can possibly fill it with "garbage" or the last valid element in the subgroup. I'm thinking about how to do this.

Yeah, this PR does not resolve this part of the issue which makes it still dependent on the identity. Currently, it relies on the previous value of whatever `v` was when it is unused in the scan which is initially set to the identity.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
```diff
@@ -267,13 +318,31 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
...
+                            sub_group_partials[i * VL + sub_group_local_id] =
+                                tmp_storage[csrc + i * VL + sub_group_local_id];
+                        }
+                    }
+                    else
                     {
-                        sub_group_partials[i * VL + sub_group_local_id] =
-                            tmp_storage[csrc + i * VL + sub_group_local_id];
+                        std::uint8_t i = 0;
+                        _ONEDPL_PRAGMA_UNROLL
```
danhoeflinger, line 335: I don't think we are going to get anything out of these unrolls anymore because we have removed the constexpr calculation of `iters`.  It probably doesn't hurt us, but it may be misleading.

danhoeflinger (commented):
I think this looks OK to merge.  If you want, you can remove the unrolls then merge.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
mmichel11: I have removed the unrolls here and in a few other places where our loops can't be unrolled.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
```diff
@@ -199,15 +224,37 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
...
+                        }
+                    }
+                    else
+                    {
+                        // In practice iters is usually 1 here when the number of sub-groups is less than the vector length.
+                        // An exception would be the unlikely case where the sub-group size does not divide the work-group size
+                        std::uint8_t i = 0;
+                        for (; i < iters - 1; i++)
+                        {
+                            v = sub_group_partials[sub_group_local_id];
```
danhoeflinger, line 245: I know this is merged, but I'm trying to merge conflicts with my branch and I notice that `i * VL` is missing from this index.  I think its required.  Am I missing something?

include/oneapi/dpl/experimental/kt/two_pass_scan.h
```diff
@@ -199,15 +224,37 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
...
+                            std::tie(v, sub_group_carry) = sub_group_scan<VL, true>(sub_group, std::move(v), binary_op,
+                                                                                    std::move(sub_group_carry));
+                            tmp_storage[start_idx + i * VL + sub_group_local_id] = v;
+                        }
+                        // If we are past the input range, then the previous value of v is passed to the sub-group scan.
+                        // It does not affect the result as our sub_group_scan will use a mask to only process in-range elements.
+                        if (i * VL + sub_group_local_id < num_sub_groups_local)
+                            v = sub_group_partials[sub_group_local_id];
```
danhoeflinger, line 253: Same comment here, is there a reason not to add `i * VL` here?

danhoeflinger (commented):
I can fix these in my upcoming commit, just please take a look and let me know if I'm missing something here.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
mmichel11: Good catch. It should be here.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
mmichel11: It should be here as well.

mmichel11: It should be here as well.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
danhoeflinger: OK I'll fix these in my next commit.

PR #1640, opened 2024-06-25
Addressing static analysis errors

timmiesmith:
Scanning oneDPL code with a static analysis tool identified the two versions of a complex comparator in the SIMD unseq backend.  This PR removes the user-defined copy constructors since they are explicit versions of the default copy constructor to address the errors from the static analysis tool.

In addition the scans also identified two functions used in the implementation of uninitialized memory algorithms that were tagged with the noexcept keyword.  In the case of the uninitialized memory functors new may be called, in which case an exception could be thrown.

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -104,7 +104,7 @@ __for_each_n_it_serial(_ForwardIterator __first, _Size __n, _Function __f)
 //------------------------------------------------------------------------
 template <class _ForwardIterator, class _Function>
 void
-__brick_walk1(_ForwardIterator __first, _ForwardIterator __last, _Function __f, /*vector=*/::std::false_type) noexcept
+__brick_walk1(_ForwardIterator __first, _ForwardIterator __last, _Function __f, /*vector=*/::std::false_type)
 {
     ::std::for_each(__first, __last, __f);
 }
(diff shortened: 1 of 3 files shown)


mmichel11 (approved):
LGTM

danhoeflinger (dismissed):
LGTM

adamfidel (approved):
Looks good, but I'm curious what were the errors from the static analysis tool? If the user-defined copy constructor is identical to the default copy constructor, what is the error? Or is the error that it the user-defined copy constructor is unnecessary?

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -104,7 +104,7 @@ __for_each_n_it_serial(_ForwardIterator __first, _Size __n, _Function __f)
 //------------------------------------------------------------------------
 template <class _ForwardIterator, class _Function>
 void
-__brick_walk1(_ForwardIterator __first, _ForwardIterator __last, _Function __f, /*vector=*/::std::false_type) noexcept
```
danhoeflinger, line 107: From the CI, it looks like there is a forward declaration you also need to change.

timmiesmith:
> Looks good, but I'm curious what were the errors from the static analysis tool? If the user-defined copy constructor is identical to the default copy constructor, what is the error? Or is the error that it the user-defined copy constructor is unnecessary?

Static analysis identified it as a violation of the [Rule of Three](https://en.cppreference.com/w/cpp/language/rule_of_three), basically seeing the user-defined copy constructor and expecting to see corresponding user-defined assignment operator and destructor.

timmiesmith:
CI issues have been resolved.

danhoeflinger (approved):
LGTM

adamfidel (approved):
LGTM

PR #1644, opened 2024-06-26
[Perf] Break loop in __early_exit_find_or after first found element

SergeyKopienko:
In this PR we implement break of for-loop inside `__early_exit_find_or` after first found element.
The approach from the PR https://github.com/oneapi-src/oneDPL/pull/1624 is applicable for the first/last element search.
This approach gives us a good performance boost.
Technically this `break` call implemented thought additional local variable with check inside of continue-condition in `for`-loop due performance results. 

Also in this PR, we analyze the result of `compare_exchange_strong()` call and break loop if the value of atomic has been changed correctly to avoid extra atomic value `load()` calls.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1051,10 +1051,7 @@ struct __early_exit_find_or
         // if our "line" is out of work group size, reduce the line to the number of the rest elements
         if (__wg_size - __leader < __shift)
             __shift = __wg_size - __leader;
-
-        bool __something_was_found = false;
-
-        for (_IterSize __i = 0; __i < __n_iter && !__something_was_found; ++__i)
+        for (_IterSize __i = 0; __i < __n_iter; ++__i)
         {
             //in case of find-semantic __shifted_idx must be the same type as the atomic for a correct comparison
             using _ShiftedIdxType = ::std::conditional_t<_OrTagType::value, decltype(__init_index + __i * __shift),
@@ -1080,7 +1077,7 @@ struct __early_exit_find_or
                     }
                 }
 
-                __something_was_found = true;
+                break;
             }
         }
     }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1099,7 +1099,9 @@ struct __early_exit_find_or
                 //    This means that after the first found entry there is no reason to process data anymore too.
                 // 3) __parallel_or_tag : when we search for any matching data entry, we process data from start to end (forward direction).
                 //    This means that after the first found entry there is no reason to process data anymore too.
-                __something_was_found = true; // break statement here shows poor perf then bool variable state check in the for-loop header in some cases
+                // But break statement here shows poor perf in some cases.
+                // So we use bool variable state check in the for-loop header.
+                __something_was_found = true;
             }
         }
     }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1075,7 +1075,8 @@ struct __early_exit_find_or
                 {
                     for (auto __old = __found_local.load(); __comp(__shifted_idx, __old); __old = __found_local.load())
                     {
-                        __found_local.compare_exchange_strong(__old, __shifted_idx);
+                        if (__found_local.compare_exchange_strong(__old, __shifted_idx))
+                            break;
                     }
                 }
 
@@ -1176,7 +1177,8 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
                             for (auto __old = __found.load(); __comp(__found_local.load(), __old);
                                  __old = __found.load())
                             {
-                                __found.compare_exchange_strong(__old, __found_local.load());
+                                if (__found.compare_exchange_strong(__old, __found_local.load()))
+                                    break;
                             }
                         }
                     }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1095,7 +1095,7 @@ struct __early_exit_find_or
                 //    This means that after first found entry there is no reason to process data anymore.
                 // 2) __parallel_find_backward_tag  : when we search for the last matching data entry, we process data from end to start (backward direction).
                 //    This means that after the first found entry there is no reason to process data anymore too.
-                // 3) __parallel_or_tag : when we find any data entry we processing data we processing data from start to end (forward direction).
+                // 3) __parallel_or_tag : when we search for any matching data entry, we process data from start to end (forward direction).
                 //    This means that after the first found entry there is no reason to process data anymore too.
                 break;
             }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1036,7 +1036,7 @@ struct __early_exit_find_or
     {
         using __par_backend_hetero::__parallel_or_tag;
 
-        // There is 3 possible tag types here:
+        // There are 3 possible tag types here:
         //  - __parallel_find_forward_tag : in case when we find the first value in the data;
         //  - __parallel_find_backward_tag : in case when we find the last value in the data;
         //  - __parallel_or_tag : in case when we find any value in the data.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1077,6 +1077,7 @@ struct __early_exit_find_or
                     {
                         __found_local.compare_exchange_strong(__old, __shifted_idx);
                     }
+                    break;
                 }
             }
         }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1093,7 +1093,7 @@ struct __early_exit_find_or
                 // This break is safe for all our cases:
                 // 1) __parallel_find_forward_tag : when we find the first data entry we processing data we process data from start to end (forward direction).
                 //    This means that after first found entry there is no reason to process data anymore.
-                // 2) __parallel_find_backward_tag  : when we find the last data entry we processing data we process data from end to start (backward direction).
+                // 2) __parallel_find_backward_tag  : when we search for the last matching data entry, we process data from end to start (backward direction).
                 //    This means that after the first found entry there is no reason to process data anymore too.
                 // 3) __parallel_or_tag : when we find any data entry we processing data we processing data from start to end (forward direction).
                 //    This means that after the first found entry there is no reason to process data anymore too.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1091,7 +1091,7 @@ struct __early_exit_find_or
 
                 // This break is mandatory from the performance point of view.
                 // This break is safe for all our cases:
-                // 1) __parallel_find_forward_tag : when we find the first data entry we processing data we process data from start to end (forward direction).
+                // 1) __parallel_find_forward_tag : when we search for the first matching data entry, we process data from start to end (forward direction).
                 //    This means that after first found entry there is no reason to process data anymore.
                 // 2) __parallel_find_backward_tag  : when we search for the last matching data entry, we process data from end to start (backward direction).
                 //    This means that after the first found entry there is no reason to process data anymore too.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1073,13 +1073,13 @@ struct __early_exit_find_or
             {
                 if constexpr (_OrTagType::value)
                 {
-                    // As far as we find any value of the data here, we can set the atomic value to 1
+                    // As far as we found any data entry, we can set the atomic value to 1.
                     // No additional actions required.
                     __found_local.store(1);
                 }
                 else
                 {
-                    // As far as we find the first/last value of the data here, we need to set the atomic value to the index of the found data.
+                    // As far as we found the first/last data entry here, we need to set the atomic value to the index of the found data entry.
                     // But only in this case when this value is less (if we find the first value)/greater(if we find the last value) than the current value of the atomic.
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1067,17 +1070,17 @@ struct __early_exit_find_or
...
                 {
                     for (auto __old = __found_local.load(); __comp(__shifted_idx, __old); __old = __found_local.load())
                     {
-                        __found_local.compare_exchange_strong(__old, __shifted_idx);
+                        if (__found_local.compare_exchange_strong(__old, __shifted_idx))
+                            break;
                     }
                 }
+
+                __something_was_found = true;
```
julianmi, line 1083: Can we simplify this by using a break here?

SergeyKopienko: Sure, but my performance checks show me that this variant works a little bit faster than the variant with break.

julianmi: That's interesting. Does this performance benefit apply for the other break statements as well?

SergeyKopienko: I think required additional performance checks for the other cases.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1067,17 +1070,17 @@ struct __early_exit_find_or
             if (__shifted_idx < __n && __pred(__shifted_idx, __rngs...))
             {
                 if constexpr (_OrTagType::value)
-                {
                     __found_local.store(1);
-                    break;
-                }
                 else
                 {
                     for (auto __old = __found_local.load(); __comp(__shifted_idx, __old); __old = __found_local.load())
```
danhoeflinger, line 1073: I wonder if this combination of two atomic operations (load then exchange) can actually be converted to a single `fetch_min` or `fetch_max` followed by non-atomic local check of the return vs the operand to check if early exit can be triggered.  This would also remove the possibility to "miss" which is what requires the loop in my understanding.

I believe `__comp` will always be `less` or `greater` in this branch, and no more exotic comparisons are required, which should enable such a change.

If this isn't quite right, then I'd consider the weak exchange comment here instead..

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1174,7 +1177,8 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
                             for (auto __old = __found.load(); __comp(__found_local.load(), __old);
```
danhoeflinger, line 1174: I believe all writes to `__found_local` happened before the barrier.  Can we just read from `__temp_local` directly now without involving atomics?

danhoeflinger, line 1174: Can we also reduce this to be a `fetch_min` / `fetch_max` operation depending on the tag?  It seems like both the find_first would be a `fetch_min`, and the find_last and find_or would be a `fetch_max`.  It needs more thought to confirm, but I think its OK.

danhoeflinger (commented):
Lets have a conversation offline if any of these comments don't make sense.  I think what I'm proposing should still be correct, and has the opportunity for performance improvement, but its worth discussing if you see any problems.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: But this additional perf improvement looks too small, so rolled back.
Let's use `break` statement.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1067,17 +1070,17 @@ struct __early_exit_find_or
             if (__shifted_idx < __n && __pred(__shifted_idx, __rngs...))
             {
                 if constexpr (_OrTagType::value)
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1067,17 +1070,17 @@ struct __early_exit_find_or
...
-                {
                     __found_local.store(1);
-                    break;
-                }
                 else
                 {
                     for (auto __old = __found_local.load(); __comp(__shifted_idx, __old); __old = __found_local.load())
                     {
-                        __found_local.compare_exchange_strong(__old, __shifted_idx);
+                        if (__found_local.compare_exchange_strong(__old, __shifted_idx))
```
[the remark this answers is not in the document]
SergeyKopienko: I already have checked this option today: the performance is the same for the `compare_exchange_weak` and `compare_exchange_strong` so I believe there is no reasons to make this change.

I will check additionally the improvement of the usage `fetch_min` and `fetch_max`.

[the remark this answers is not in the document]
SergeyKopienko: About the usage `fetch_min` and `fetch_max` - looks like it's give about additional 10% so I will apply it.
Good catch, @danhoeflinger, thanks!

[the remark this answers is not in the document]

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: About the usage `fetch_min` and `fetch_max` - looks like it's give about additional 10% so I will apply it.
Good catch, @danhoeflinger, thanks!

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
SergeyKopienko: https://github.com/oneapi-src/oneDPL/pull/1644#discussion_r1655117321

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Also fixed at https://github.com/oneapi-src/oneDPL/pull/1644/commits/42eb498d1a113aebf2e0ea1ed712c8cbc8b06c3c

SergeyKopienko: `__found_local` describes the state of find inside one group. So atomic required, I believe.

danhoeflinger: It is the workgroup-local "found" status, but all writes to this variable occur before the intra-workgroup barrier and we are now after the barrier. This should mean that **all** threads in the workgroup are past the barrier, and only reads to this memory remain.  I believe this means we can interact directly with the memory without any danger.  

My guess is that it wont change performance, but it has a chance for an improvement.

danhoeflinger: We are still doing a `__found.load()` and `__comp` operation.  I think we can avoid this completely, and replace it with `fetch_min` and `fetch_max`.  This should be applicable even in the case of `__or_tag_check`, because `fetch_max` should be a fine replacement for `__atomic_compare` as written.  

It *is* possible that this is slightly slower though, I'm not sure.  
As written, we have 1 atomic read guarantee, and an optional atomic write if the comparison returns true.  My proposal would replace that system with a single atomic read+write in the form of `fetch_max`. 

If we did this I think we might also be able to remove the `_Compare` type altogether.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1068,16 +1073,21 @@ struct __early_exit_find_or
...
-                    }
+                    // As far as we find the first/last value of the data here, we need to set the atomic value to the index of the found data.
+                    // But only in this case when this value is less (if we find the first value)/greater(if we find the last value) than the current value of the atomic.
+                    if constexpr (!_BackwardTagType::value)
+                        __found_local.fetch_min(__shifted_idx);
+                    else
+                        __found_local.fetch_max(__shifted_idx);
                 }
+
+                break;
```
julianmi, line 1090: Does it make sense to add a comment here about why the break is generally safe? I think especially the backward tag can be confusing if one is unaware that we process elements backwards in this case.

SergeyKopienko: Fixed: added additional comments.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: All of the atomic operations have this clause in their description (including the ones you reverted to).  What it is saying is that these 64 bit data type operations are only supported for device which support 64 bit atomics.  
I think we should use what is fastest here, and if we want to we can do some check of 
`__queue.get_device().has(sycl::aspect::atomic64)` when we have a 64 bit type like the one done here in single pass scan https://github.com/oneapi-src/oneDPL/blob/71de137931a0f0e23d29fb8dd6ac43352a9b46ed/include/oneapi/dpl/experimental/kt/single_pass_scan.h#L329

SergeyKopienko: Sure, we able to implement this approach.
But the amount of required changes in this case is too much.
Let's think about this in separate PR.

SergeyKopienko: And yes, of course, I already seen this code and condition check.

danhoeflinger: OK, that is fine to separate to another PR.

Just to be clear though, `fetch_min` and `fetch_max` impose no additional requirements as compared to the current PR.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1035,6 +1035,11 @@ struct __early_exit_find_or
                _LocalAtomic& __found_local, _BrickTag, _Ranges&&... __rngs) const
     {
         using __par_backend_hetero::__parallel_or_tag;
+
+        // There is 3 possible tag types here:
```
julianmi, line 1039:
```suggestion
        // There are 3 possible tag types here:
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1068,16 +1073,31 @@ struct __early_exit_find_or
...
                 {
+                    // As far as we found any data entry, we can set the atomic value to 1.
+                    // No additional actions required.
                     __found_local.store(1);
-                    break;
                 }
                 else
                 {
+                    // As far as we found the first/last data entry here, we need to set the atomic value to the index of the found data entry.
+                    // But only in this case when this value is less (if we find the first value)/greater(if we find the last value) than the current value of the atomic.
```
julianmi, line 1083:
```suggestion
                    // But only in this case when this value is less (if we find the first value)/greater (if we find the last value) than the current value of the atomic.
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1068,16 +1073,31 @@ struct __early_exit_find_or
...
-                        __found_local.compare_exchange_strong(__old, __shifted_idx);
+                        // If we replace the atomic value successfully, we should break the loop to avoid extra operations with atomic
+                        if (__found_local.compare_exchange_strong(__old, __shifted_idx))
+                            break;
                     }
                 }
+
+                // This break is mandatory from the performance point of view.
+                // This break is safe for all our cases:
+                // 1) __parallel_find_forward_tag : when we find the first data entry we processing data we process data from start to end (forward direction).
```
julianmi, line 1094:
```suggestion
                // 1) __parallel_find_forward_tag : when we search for the first matching data entry, we process data from start to end (forward direction).
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1068,16 +1073,31 @@ struct __early_exit_find_or
...
+                //    This means that after first found entry there is no reason to process data anymore.
+                // 2) __parallel_find_backward_tag  : when we find the last data entry we processing data we process data from end to start (backward direction).
```
julianmi, line 1096:
```suggestion
                // 2) __parallel_find_backward_tag  : when we search for the last matching data entry, we process data from end to start (backward direction).
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: You should be able to replace all instances of `found_local.load()` with `__temp_local[0]` after the barrier (or copy it into some local variable `found_local_state` for clarity.  Concurrent reads are safe, all writes will be resolved before the barrier.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1068,16 +1073,31 @@ struct __early_exit_find_or
...
+                //    This means that after the first found entry there is no reason to process data anymore too.
+                // 3) __parallel_or_tag : when we find any data entry we processing data we processing data from start to end (forward direction).
```
julianmi, line 1098:
```suggestion
                // 3) __parallel_or_tag : when we search for any matching data entry, we process data from start to end (forward direction).
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: I have checked this variant: there is no significant performance improvement during my experiments.
So for reduce the amount of changes in this PR I propose do not change it in this PR.

danhoeflinger: Fair enough.  Its good that the atomic implementation is good enough that reads alone are basically equivalent to just normal SLM reads.

SergeyKopienko:
@danhoeflinger, @julianmi how do you think, are we ready to merge this PR ?

danhoeflinger (approved):
LGTM (wait for green CI)

julianmi (approved):

julianmi (approved):
LGTM. However, I'd like to understand why this is faster than break statements. There might be potential performance benefits in other code parts where we use break statements in case this is a general issue.

SergeyKopienko:
> LGTM. However, I'd like to understand why this is faster than break statements. There might be potential performance benefits in other code parts where we use break statements in case this is a general issue.

Just now I found one place with break pattern in Kernel code and going to investigate this place separatelly.
It's inside `struct multiple_match_pred`

danhoeflinger (approved):
LGTM, It seems fine from a correctness perspective.  It should simplify when we make the switch to fetch_min / fetch_max.

danhoeflinger:
My guess about the break statements is that it is more difficult for thread divergence within a subgroup.  It does seem like the compiler should be able to generate code for a break; which is equivalently performant as a loop exit flag though.  
It would be interesting to boil down to a very simple example and examine the resulting generated code.

PR #1647, opened 2024-06-27
[PROTOTYPE] Two-pass scan - Support in-place exclusive scans and update tests

mmichel11:
This PR is targeted to the development branch for a new scan implementation.

I have done some evaluation of the scan implementation for the in-place case where the input and output buffers are the same. For inclusive scan, there are no issues with the current implementation. For exclusive scan, we have an issue in the multi-block case where we fetch the last element from the previous block's input to compute the current block's carry-in. In the in-place case, this input value would have been already overwritten with the exclusive scan value.

To fix this issue, the last input element in a block is written to temporary storage by the last sub-group's work item in the second kernel before the scan results are written and is then fetched by the next block. This operation does not add any additional overhead as the last sub-group would be idle either way waiting on the first sub-group to finish its work.

I have added a test-case for out-of-place and in-place inclusive / exclusive scan to the USM tests. The test build / run time is starting to grow, so I don't think we should add them to each case.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/two_pass_scan.h
+++ b/include/oneapi/dpl/experimental/kt/two_pass_scan.h
@@ -108,6 +108,7 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
     std::uint32_t work_group_size = q.get_device().get_info<sycl::info::device::max_work_group_size>();
     // TODO: develop simple heuristic to determine this value based on maximum compute units
     std::uint32_t num_work_groups = 128;
+    std::uint32_t num_work_items = work_group_size * num_work_groups;
     std::uint32_t num_sub_groups_local = work_group_size / VL;
     std::uint32_t num_sub_groups_global = num_sub_groups_local * num_work_groups;
     // Is set if the scanner sub-group that scans sub-group carries in a work-group divides evenly into
@@ -143,7 +144,7 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
     auto blockSize = (M < MAX_INPUTS_PER_BLOCK) ? M : MAX_INPUTS_PER_BLOCK;
     auto numBlocks = M / blockSize + (M % blockSize != 0);
 
-    auto globalRange = range<1>(num_work_groups * work_group_size);
+    auto globalRange = range<1>(num_work_items);
     auto localRange = range<1>(work_group_size);
     nd_range<1> range(globalRange, localRange);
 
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/experimental/kt/two_pass_scan.h
```diff
@@ -355,6 +359,19 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
...
                     }
                 }
+                // For the exclusive scan case:
+                // While the first sub-group is doing work, have the last item in the group store the last element
+                // in the block to temporary storage for use in the next block.
+                // This is required to support in-place exclusive scans as the input values will be overwritten.
+                if constexpr (!Inclusive)
+                {
+                    auto global_id = ndi.get_global_linear_id();
+                    if (global_id == num_work_groups * work_group_size - 1)
```
adamfidel, line 369: Could we create a variable for the expression `num_work_groups * work_group_size` (maybe called `num_work_items`) and use it both here and on line 146 (before creation of `globalRange`)?

test/kt/two_pass_scan.cpp
```diff
@@ -67,30 +67,58 @@ generate_scan_data(T* input, std::size_t size, std::uint32_t seed)
     }
 }
 
-template <typename T, sycl::usm::alloc _alloc_type, typename BinOp, typename KernelParam>
```
danhoeflinger, line 70: Can we just pass one Kernel Param and then use another layer of `get_new_kernel_params` within this function?
This seems more modular and extendable.

test/kt/two_pass_scan.cpp
```diff
@@ -67,30 +67,58 @@ generate_scan_data(T* input, std::size_t size, std::uint32_t seed)
...
+template <typename T, sycl::usm::alloc _alloc_type, typename BinOp, typename KernelParam1, typename KernelParam2>
 void
-test_usm(sycl::queue q, std::size_t size, BinOp bin_op, KernelParam param)
+test_usm(sycl::queue q, std::size_t size, BinOp bin_op, KernelParam1 param1, KernelParam2 param2)
```
danhoeflinger, line 72: We probably want to add exclusive scan for the other input types eventually, although its not a blocker for this PR.

danhoeflinger (commented):
I agree with the fix, and am happy to see exclusive scan tests.  couple minor comments, but otherwise looks good.

test/kt/two_pass_scan.cpp
mmichel11: I agree it may be good. We will probably want to cut down on some testing somewhere though or maybe added a short / long mode for testing. With integrated graphics, running this test is already taking 400+ seconds on my side.

test/kt/two_pass_scan.cpp
mmichel11: Yes, I have made this change.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
mmichel11: Yes, I have done this.

danhoeflinger (approved):
LGTM

mmichel11:
@adamfidel Was there any additional review you wanted to do here?

adamfidel (approved):
LGTM.

PR #1650, opened 2024-06-28
Using `fetch_min` / `fetch_max` in `__parallel_find_or` and in it's stuff

SergeyKopienko:
In this PR we using `fetch_min` / `fetch_max` in `__parallel_find_or` and it's stuff.
This approach give us additional performance boost:
- now we shouldn't try to setup atomic states inside `for`-based loops with additional checks.

This idea was proposed by @danhoeflinger.

Also in this PR we using `__dpl_ceiling_div` function where it's applicable in the code.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -331,7 +331,7 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         // Practically this is the better value that was found
         constexpr decltype(__wgroup_size) __iters_per_witem = 16;
         auto __size_per_wg = __iters_per_witem * __wgroup_size;
-        auto __n_groups = (__n - 1) / __size_per_wg + 1;
+        auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_wg);
         // Storage for the results of scan for each workgroup
         sycl::buffer<_Type> __wg_sums(__n_groups);
 
@@ -357,7 +357,7 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         // 2. Scan for the entire group of values scanned from each workgroup (runs on a single workgroup)
         if (__n_groups > 1)
         {
-            auto __iters_per_single_wg = (__n_groups - 1) / __wgroup_size + 1;
+            auto __iters_per_single_wg = oneapi::dpl::__internal::__dpl_ceiling_div(__n_groups, __wgroup_size);
             __submit_event = __exec.queue().submit([&](sycl::handler& __cgh) {
                 __cgh.depends_on(__submit_event);
                 auto __wg_sums_acc = __wg_sums.template get_access<access_mode::read_write>(__cgh);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -973,12 +973,12 @@ struct __parallel_find_forward_tag
         return __val;
     }
 
-    template <sycl::access::address_space _Space>
+    template <typename _TAtomic>
     static void
-    __save_state_to_atomic(__dpl_sycl::__atomic_ref<_AtomicType, _Space>& __atomic, _AtomicType __new_state)
+    __save_state_to_atomic(_TAtomic& __found, _AtomicType __new_state)
     {
         // As far as we make search from begin to the end of data, we should save the first (minimal) found state.
-        __atomic.fetch_min(__new_state);
+        __found.fetch_min(__new_state);
     }
 };
 
@@ -999,12 +999,12 @@ struct __parallel_find_backward_tag
         return _AtomicType{-1};
     }
 
-    template <sycl::access::address_space _Space>
+    template <typename _TAtomic>
     static void
-    __save_state_to_atomic(__dpl_sycl::__atomic_ref<_AtomicType, _Space>& __atomic, _AtomicType __new_state)
+    __save_state_to_atomic(_TAtomic& __found, _AtomicType __new_state)
     {
         // As far as we make search from end to the begin of data, we should save the last (maximal) found state.
-        __atomic.fetch_max(__new_state);
+        __found.fetch_max(__new_state);
     }
 };
 
@@ -1020,12 +1020,12 @@ struct __parallel_or_tag
         return 0;
     }
 
-    template <sycl::access::address_space _Space>
+    template <typename _TAtomic>
     static void
-    __save_state_to_atomic(__dpl_sycl::__atomic_ref<_AtomicType, _Space>& __atomic, _AtomicType /*__new_state*/)
+    __save_state_to_atomic(_TAtomic& __found, _AtomicType /*__new_state*/)
     {
         // Store that a match was found. Its position is not relevant for or semantics.
-        __atomic.store(1);
+        __found.store(1);
     }
 };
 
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1126,7 +1151,9 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
         oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<__find_or_kernel, _CustomName, _Brick,
                                                                                _BrickTag, _Ranges...>;
 
-    constexpr bool __or_tag_check = ::std::is_same_v<_BrickTag, __parallel_or_tag>;
+    assert("This device does not support 64-bit atomics" &&
+           (sizeof(_AtomicType) < 64 || __exec.queue().get_device().has(sycl::aspect::atomic64)));
```
julianmi, line 1155: `sizeof` uses bytes so this should be `sizeof(_AtomicType) < 8`. I think even better would be `<= 4` since only 32-bit data types are then allowed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -985,40 +999,70 @@ struct __parallel_find_backward_tag
...
     constexpr static _AtomicType __init_value(_DiffType)
     {
         return 0;
     }
+
+    template <sycl::access::address_space _Space>
+    static void
+    __save_state_to_atomic(__dpl_sycl::__atomic_ref<_AtomicType, _Space>& __atomic, _AtomicType /*__new_state*/)
+    {
+        // As far as we make search from end to the begin of data, we simple save the fact of finding.
```
julianmi, line 1041: The processing order isn't relevant here. We might want to simplify this to something like: "Store that a match was found. Its position is not relevant for or semantics."

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1185,39 +1212,25 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
                 });
         });
         //The end of the scope  -  a point of synchronization (on temporary sycl buffer destruction)
     }
 
-    if constexpr (__or_tag_check)
-        return __result;
-    else
-        return __result != __init_value ? __result : __rng_n;
+    return _BrickTag::__eval_res(__result, __rng_n);
```
julianmi, line 1233: This comes probably down to preference but I preferred the approach before. I think it was easier to read and didn't require duplicated routines in forward/backward tag.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1126,7 +1151,9 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
         oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<__find_or_kernel, _CustomName, _Brick,
                                                                                _BrickTag, _Ranges...>;
 
-    constexpr bool __or_tag_check = ::std::is_same_v<_BrickTag, __parallel_or_tag>;
+    assert("This device does not support 64-bit atomics" &&
```
danhoeflinger, line 1154: This is a bigger question than this PR is looking to address, but should we have a "general" version of this algorithm which supports hardware which doesn't support 64-bit atomics rather than merely an assert? Obviously this would come at a cost, and additional code.

For instance, I believe this is not supported in FPGA (if I remember correctly from work on histogram), which also may use 64-bit atomics based on the types passed in from the user).

Note that this is not a new problem added by this PR or any recent work, but just one which is been brought up by the addition of the assertion.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Good catch, @julianmi 
But this mean we have the same error here:
```C++
    assert("This device does not support 64-bit atomics" &&
           (sizeof(_Type) < 64 || __queue.get_device().has(sycl::aspect::atomic64)));
```
I mean https://github.com/oneapi-src/oneDPL/blob/e1fee6404e4ebda93ade0d3e913e1eb9e9b8595b/include/oneapi/dpl/experimental/kt/single_pass_scan.h#L328
It was introduced in https://github.com/oneapi-src/oneDPL/commit/e625bf013767577392c3b3e28fa1d8e46112f3fb

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Let's resolve this point a little bit later.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Probably we have one more option here: to replace this `assert` to `static_assert` and check `sizeof(_Type)`.
If problem really exist I think that there is no reasons to land this PR due it's about only 10% perf profit.

SergeyKopienko: @akukanov what do you think about this? I mean the case described by @danhoeflinger at https://github.com/oneapi-src/oneDPL/pull/1650#discussion_r1658876487 ?

SergeyKopienko: Probably this point has been already resolved early: in the structures `__parallel_find_forward_tag` and `__parallel_find_backward_tag` the `_AtomicType` has been defined as below:
```C++
// FPGA devices don't support 64-bit atomics
#if _ONEDPL_FPGA_DEVICE
    using _AtomicType = uint32_t;
#else
    using _AtomicType = oneapi::dpl::__internal::__difference_t<_RangeType>;
#endif
```

julianmi (approved):
LGTM. I agree with @danhoeflinger that we should extend this in a separate PR to support 64-bit types. Please wait for his approval before merging.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -985,40 +992,57 @@ struct __parallel_find_backward_tag
...
-      public:
-        template <typename _LocalAtomic, typename _GlobalAtomic>
-        bool
-        operator()(const _LocalAtomic& __found_local, const _GlobalAtomic& __found) const
-        {
-            return __found_local == 1 && __found == 0;
-        }
-    };
-
     using _AtomicType = int32_t;
```
danhoeflinger, line 1014: Should this use the same Atomic type as in `__parallel_find_[backward,forward]_tag`?

Follow ups:  
1) Does `backward` have a reason to be signed int rather than unsigned?
2) Do we check that `n` is less than max (index type)?  (perhaps we have other problems with `n` > 2 Billion)

danhoeflinger (approved):
Basically looks good to me, other than some questions about atomics and making consistent, which can be put in another PR if we don't want to address them here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: About "Should this use the same Atomic type as in __parallel_find_[backward,forward]_tag?"
 - it's simple not required because we save the fact that we found something, not real position. From this point of view it may be `bool` type too. But I assume that we haven't any additional perf proof from that.

danhoeflinger: Yes, ok that is right for the `or` case.  However, I think it cant be bool because atomics must be at least 32bits.  

I think my follow ups still make sense for `__parallel_find_[backward,forward]_tag` though.  
Should we align these with each other?

Issue #1631, opened 2024-06-21
including <oneapi/dpl/*> headers cause CUDA_ERROR_NOT_INITIALIZED after fork


jonasdelacour:
Including any <oneapi/dpl> header seems to construct a sycl::queue, which must not be done prior to fork() calls. If you do you get  CUDA_ERROR_NOT_INITIALIZED when the child process attempts to destroy this queue and underlying CUDA context.

Here's a minimum example to reproduce this error:

```
#include <oneapi/dpl/algorithm>
#include <unistd.h>

int main(){
    pid_t pid = fork();
    return 0;
}
```

Compiled with `icpx -fsycl`

produces the following stack trace from valgrind:

```
==11688== Process terminating with default action of signal 6 (SIGABRT)
==11688==    at 0x4FD79FC: __pthread_kill_implementation (pthread_kill.c:44)
==11688==    by 0x4FD79FC: __pthread_kill_internal (pthread_kill.c:78)
==11688==    by 0x4FD79FC: pthread_kill@@GLIBC_2.34 (pthread_kill.c:89)
==11688==    by 0x4F83475: raise (raise.c:26)
==11688==    by 0x4F697F2: abort (abort.c:79)
==11688==    by 0x4914B9D: ??? (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30)
==11688==    by 0x492020B: ??? (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30)
==11688==    by 0x4920276: std::terminate() (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30)
==11688==    by 0x4CB25BA: __clang_call_terminate (in /opt/intel/oneapi/compiler/2024.1/lib/libsycl.so.7.1.0)
==11688==    by 0x4CC0E0F: sycl::_V1::detail::queue_impl::~queue_impl() (in /opt/intel/oneapi/compiler/2024.1/lib/libsycl.so.7.1.0)
==11688==    by 0x4025EA: oneapi::dpl::execution::__dpl::device_policy<oneapi::dpl::execution::__dpl::DefaultKernelName>::~device_policy() (in /home/jonaslacour/playground/a.out)
==11688==    by 0x4F86494: __run_exit_handlers (exit.c:113)
==11688==    by 0x4F8660F: exit (exit.c:143)
==11688==    by 0x4F6AD96: (below main) (libc_start_call_main.h:74)
```

icpx version:
`
[...]

PR #1652, opened 2024-06-28
Resolve global policy issues in a backward-compatible manner, attempt 2

akukanov:
This patch is a follow-up to #1618 (with the commit history from there) with a somewhat modified approach. It is necessary to address #1631, one more issue caused by eager initialization of a SYCL queue in oneDPL headers. It still is supposed to address #1060.

The core idea is similar: to keep layout compatibility yet have lazy behavior, store SYCL queues inside a special `__queue_holder` class with the same size and alignment, and use the first `sizeof(uintptr_t)` bytes as a flag to detect if it holds a valid queue or - in case there are only zeros in these bytes - something else. To remind, 
> the rationale is that a SYCL queue is likely implemented as a shared_ptr (that's certainly the case for DPC++), which in turn typically holds several pointers to the actual object, a service block or a reference counter, etc. It is highly unlikely therefore that the first bytes in a properly constructed queue object will be equivalent to a null pointer. 

 Special device policy constructors are used only for predefined policies and, unlike #1618, always nullify the "flag" and additionally store a pointer to a factory function that generates a proper `sycl::queue` object as a copy of an internal "magic static" instance initialized on the first call to the factory. Each time when a device policy need to return its queue the "flag" is checked and, depending on its state, either a copy of the stored queue is returned or the factory is called.

It is worth noting that the workaround only applies to the predefined policies. Any explicitly created policies as well as copies of policies created by the implementation (if it chooses so) will contain a valid queue. Propagating factories to the copies of predefined policies is deliberately not done; it is impossible to always create queues on first use without breaking class layout, so having consistent behavior for all policies except predefined ones seems the best choice.

The implementation still introduces an undefined behavior, this time by possibly using an inactive union field (the "flag", in case a valid queue was constructed).

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -24,6 +24,9 @@
 
 #include <type_traits>
 
+#include <cstddef> // std::byte, std::nullptr_t
+#include <cassert>
+
 namespace oneapi
 {
 namespace dpl
@@ -32,6 +35,81 @@ namespace execution
 {
 inline namespace __dpl
 {
+
+namespace __internal {
+
+struct __lazy_ctor_tag {};
+
+struct alignas(sycl::queue) __lazy_queue
+{
+    static_assert(sizeof(sycl::queue) >= sizeof(void*));
+    static_assert(alignof(sycl::queue) >= alignof(void*));
+
+    std::byte __buf[sizeof(sycl::queue)];
+
+    template <typename... _Args>
+    __lazy_queue(_Args... __args)
+    {
+        new(__buf) sycl::queue(__args...);
+    }
+
+    __lazy_queue(__lazy_ctor_tag)
+    {
+        new(__buf) std::nullptr_t;
+    }
+
+    ~__lazy_queue()
+    {
+        if (__has_queue())
+            __get_queue_ref().~queue();
+    }
+
+    bool __has_queue() const
+    {
+        return nullptr!=*reinterpret_cast<void* const*>(__buf);
+    }
+
+    const sycl::queue& __get_queue_ref() const
+    {
+        assert(__has_queue());
+        return *reinterpret_cast<sycl::queue const*>(__buf);
+    }
+    sycl::queue& __get_queue_ref()
+    {
+        assert(__has_queue());
+        return *reinterpret_cast<sycl::queue*>(__buf);
+    }
+
+    sycl::queue __get_queue_with_fallback(sycl::queue(*__f)()) const
+    {
+        return __has_queue()? __get_queue_ref() : __f();
+    }
+};
+
+inline sycl::queue
+__get_default_queue()
+{
+    static sycl::queue __q(sycl::default_selector_v);
+    return __q;
+}
+
+#if _ONEDPL_FPGA_DEVICE
+using __fpga_default_selector = 
+#if _ONEDPL_FPGA_EMU
+    __dpl_sycl::__fpga_emulator_selector;
+#else
+    __dpl_sycl::__fpga_selector;
+#endif
+
+inline sycl::queue
+__get_fpga_default_queue()
+{
+    static sycl::queue __q(__fpga_default_selector());
+    return __q;
+}
+#endif // _ONEDPL_FPGA_DEVICE
+
+} // namespace __internal
 
 struct DefaultKernelName;
 
@@ -53,15 +131,22 @@ class device_policy
     }
     explicit device_policy(sycl::queue q_) : q(q_) {}
     explicit device_policy(sycl::device d_) : q(d_) {}
-    operator sycl::queue() const { return q; }
+    explicit device_policy(__internal::__lazy_ctor_tag t_) : q(t_) {}
+
+    operator sycl::queue() const { return queue(); }
     sycl::queue
     queue() const
     {
-        return q;
+        return __get_queue_with_fallback(__internal::__get_default_queue);
     }
 
+  protected:
+    sycl::queue __get_queue_with_fallback(sycl::queue(*__f)())
+    {
+        return q.__get_queue_with_fallback(__f);
+    }
   private:
-    sycl::queue q;
+    __internal::__lazy_queue q;
 };
 
 #if _ONEDPL_FPGA_DEVICE
@@ -74,14 +159,7 @@ class fpga_policy : public device_policy<KernelName>
   public:
     static constexpr unsigned int unroll_factor = factor;
 
-    fpga_policy()
-        : base(sycl::queue(
-#    if _ONEDPL_FPGA_EMU
-              __dpl_sycl::__fpga_emulator_selector()
-#    else
-              __dpl_sycl::__fpga_selector()
-#    endif // _ONEDPL_FPGA_EMU
-                  ))
+    fpga_policy() : base(sycl::queue(__internal::__fpga_default_selector()))
     {
     }
 
@@ -89,8 +167,15 @@ class fpga_policy : public device_policy<KernelName>
     fpga_policy(const fpga_policy<other_factor, OtherName>& other) : base(other.queue()){};
     explicit fpga_policy(sycl::queue q) : base(q) {}
     explicit fpga_policy(sycl::device d) : base(d) {}
-};
+    explicit fpga_policy(__internal::__lazy_ctor_tag t) : base(t) {}
 
+    operator sycl::queue() const { return queue(); }
+    sycl::queue
+    queue() const
+    {
+        return base::__get_queue_with_fallback(__internal::__get_fpga_default_queue);
+    }
+};
 #endif // _ONEDPL_FPGA_DEVICE
 
 // 2.8, Execution policy objects
@@ -101,9 +186,9 @@ class fpga_policy : public device_policy<KernelName>
 // Starting with c++17 we can simply define sycl as inline variable.
 #    if _ONEDPL___cplusplus >= 201703L
 
-inline device_policy<> dpcpp_default{};
+inline device_policy<> dpcpp_default{__internal::__lazy_ctor_tag{}};
 #        if _ONEDPL_FPGA_DEVICE
-inline fpga_policy<> dpcpp_fpga{};
+inline fpga_policy<> dpcpp_fpga{__internal::__lazy_ctor_tag{}};
 #        endif // _ONEDPL_FPGA_DEVICE
 
 #    endif // _ONEDPL___cplusplus >= 201703L
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -50,16 +50,19 @@ class alignas(sycl::queue) __queue_holder
     static_assert(sizeof(sycl::queue) >= sizeof(std::pair<uintptr_t, _Queue_factory>));
     static_assert(alignof(sycl::queue) >= alignof(std::uintptr_t));
 
-    union {
+    union
+    {
         sycl::queue __q;
         std::pair<uintptr_t, _Queue_factory> __flag_and_factory;
     };
 
-    bool __has_queue() const
+    bool
+    __has_queue() const
     {
         bool res = true;
 #if _ONEDPL_PREDEFINED_POLICIES
-        res = (__flag_and_factory.first != 0); // If the first size-of-pointer bytes are zeros, we consider there is no valid queue.
+        // If the first size-of-pointer bytes are zeros, we consider there is no valid queue.
+        res = (__flag_and_factory.first != 0);
         std::atomic_signal_fence(std::memory_order_acq_rel); // to prevent possible reordering due to type punning
 #endif
         return res;
@@ -70,9 +73,9 @@ class alignas(sycl::queue) __queue_holder
     __queue_holder(_Args... __args) : __q{std::forward<_Args>(__args)...} {}
 
 #if _ONEDPL_PREDEFINED_POLICIES
-     // The ctor for predefined policy instances does not create a queue but stores a queue factory.
-     // The first size-of-pointer bytes - the "flag" - are nullified to indicate that there is no valid queue.
-     // Then a pointer to a factory function is stored.
+    // The ctor for predefined policy instances does not create a queue but stores a queue factory.
+    // The first size-of-pointer bytes - the "flag" - are nullified to indicate that there is no valid queue.
+    // Then a pointer to a factory function is stored.
     __queue_holder(_Queue_factory __f) : __flag_and_factory{0, __f} {}
 #endif
 
@@ -88,21 +91,24 @@ class alignas(sycl::queue) __queue_holder
     __queue_holder(const __queue_holder& __h) : __q{__h.__get_queue()} {}
     __queue_holder(__queue_holder&& __h) : __q{__h.__get_queue()} {}
 
-    __queue_holder& operator=(const __queue_holder& __h)
+    __queue_holder&
+    operator=(const __queue_holder& __h)
     {
         if (this != &__h)
             __q = __h.__get_queue();
         return *this;
     }
 
-    __queue_holder& operator=(__queue_holder&& __h)
+    __queue_holder&
+    operator=(__queue_holder&& __h)
     {
         if (this != &__h)
             __q = __h.__get_queue();
         return *this;
     }
 
-    sycl::queue __get_queue() const
+    sycl::queue
+    __get_queue() const
     {
         if (__has_queue())
             return __q;
@@ -148,7 +154,7 @@ class device_policy
     }
 
 #if _ONEDPL_PREDEFINED_POLICIES
-    explicit device_policy(__internal::__global_instance_tag) : q(/*factory*/__get_default_queue) {}
+    explicit device_policy(__internal::__global_instance_tag) : q(/*factory*/ __get_default_queue) {}
 
   protected:
     device_policy(__internal::_Queue_factory __f) : q(__f) {}
@@ -180,6 +186,7 @@ class fpga_policy : public device_policy<KernelName>
         };
         return __q;
     }
+
   public:
     static constexpr unsigned int unroll_factor = factor;
 
@@ -190,7 +197,7 @@ class fpga_policy : public device_policy<KernelName>
     explicit fpga_policy(sycl::queue q) : base(q) {}
     explicit fpga_policy(sycl::device d) : base(d) {}
 #if _ONEDPL_PREDEFINED_POLICIES
-    explicit fpga_policy(__internal::__global_instance_tag) : base(/*factory*/__get_fpga_default_queue) {}
+    explicit fpga_policy(__internal::__global_instance_tag) : base(/*factory*/ __get_fpga_default_queue) {}
 #endif
 };
 #endif // _ONEDPL_FPGA_DEVICE
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -24,8 +24,7 @@
 
 #include <type_traits>
 
-#include <cstddef> // std::byte
-#include <cstring> // memset
+#include <cstdint> // std::uintptr_t
 #include <atomic>  // atomic_signal_fence
 #include <cassert>
 
@@ -47,15 +46,16 @@ struct __global_instance_tag {};
 
 class alignas(sycl::queue) __queue_holder
 {
-    static_assert(sizeof(sycl::queue) >= sizeof(void*));
-    static_assert(alignof(sycl::queue) >= alignof(void*));
+    static_assert(sizeof(sycl::queue) >= sizeof(std::uintptr_t));
+    static_assert(sizeof(sycl::queue) % sizeof(std::uintptr_t) == 0);
+    static_assert(alignof(sycl::queue) >= alignof(std::uintptr_t));
 
-    std::byte __buf[sizeof(sycl::queue)];
+    std::uintptr_t __buf[sizeof(sycl::queue)/sizeof(std::uintptr_t)];
 
     bool __has_queue() const
     {
-        bool res = (nullptr != reinterpret_cast<void* const&>(*this));
-        std::atomic_signal_fence(std::memory_order_acq_rel); // mitigate possible reordering due to type punning
+        bool res = (__buf[0] == 0); // If the first size-of-pointer bytes are zeros, we consider there is no valid queue
+        std::atomic_signal_fence(std::memory_order_acq_rel); // to prevent possible reordering due to type punning
         return res;
     }
 
@@ -74,7 +74,7 @@ class alignas(sycl::queue) __queue_holder
         else
         {
             // an "impossible" case of SYCL providing no devices, which we however must handle
-            std::memset(this, 0, sizeof(void*));
+            __buf[0] = 0; // nullify the first size-of-pointer bytes of the holder
             // Since the queue holder does not have a valid queue in this case,
             // the predefined device policies are de-facto unusable.
             // That seems OK though, because there is no device to run SYCL anyway.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -121,6 +121,29 @@ class alignas(sycl::queue) __queue_holder
     }
 };
 
+// Queue factory functions to use with the queue holder
+inline sycl::queue
+__get_default_queue()
+{
+    static sycl::queue __q(sycl::default_selector_v);
+    return __q;
+}
+
+#if _ONEDPL_FPGA_DEVICE
+inline sycl::queue
+__get_fpga_default_queue()
+{
+    static sycl::queue __q(
+#if _ONEDPL_FPGA_EMU
+        __dpl_sycl::__fpga_emulator_selector()
+#else
+        __dpl_sycl::__fpga_selector()
+#endif
+    );
+    return __q;
+}
+#endif // _ONEDPL_FPGA_DEVICE
+
 } // namespace __internal
 
 struct DefaultKernelName;
@@ -133,17 +156,10 @@ struct DefaultKernelName;
 template <typename KernelName = DefaultKernelName>
 class device_policy
 {
-    static sycl::queue
-    __get_default_queue()
-    {
-        static sycl::queue __q(sycl::default_selector_v);
-        return __q;
-    }
-
   public:
     using kernel_name = KernelName;
 
-    device_policy() : device_policy(__get_default_queue()) {}
+    device_policy() : device_policy(__internal::__get_default_queue()) {}
     template <typename OtherName>
     device_policy(const device_policy<OtherName>& other) : __qh(other.queue()) {}
     explicit device_policy(sycl::queue q) : __qh(q) {}
@@ -157,7 +173,8 @@ class device_policy
     }
 
 #if _ONEDPL_PREDEFINED_POLICIES
-    explicit device_policy(__internal::__global_instance_tag __t) : __qh(__t, /*factory*/__get_default_queue) {}
+    explicit device_policy(__internal::__global_instance_tag __t)
+        : __qh(__t, /*factory*/__internal::__get_default_queue) {}
 
   protected:
     device_policy(__internal::__global_instance_tag __t, __internal::__queue_factory __f) : __qh(__t, __f) {}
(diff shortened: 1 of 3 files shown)


test/general/fork.pass.cpp
```diff
@@ -0,0 +1,13 @@
+#include <oneapi/dpl/execution>
+#include "support/utils.h"
```
adamfidel, line 2: The inclusion of this header causes this new test to fail because the test utility headers define a static global policy:
https://github.com/oneapi-src/oneDPL/blob/e3a4ed39d5306dad1582a33eb7ac85035a26dbab/test/support/utils_sycl.h#L100

Once I removed this include (and changed the last line in `main` to `return 0`), this test now passes with the Codeplay NVIDIA backend.

adamfidel: After looking into it further, I believe the test header `utils_sycl.h` might be incorrect. It is creating a queue based on the macro `ONEDPL_USE_PREDEFINED_POLICIES`; however, I believe at that point it should be using `_ONEDPL_PREDEFINED_POLICIES` instead.

https://github.com/oneapi-src/oneDPL/blob/e3a4ed39d5306dad1582a33eb7ac85035a26dbab/test/support/utils_sycl.h#L99-L103

If I change the test header to use the latter macro, then your test passes without modification.

akukanov: Thanks Adam, I will take a look.
As for the macro, preferably tests should use public APIs (including macros) and avoid internal ones. Maybe this is the place to do it differently, but I'd need to make sure.

akukanov: @adamfidel You were right, the definition in `utils_sycl.h` did not properly handle the "default" behavior for `ONEDPL_USE_PREDEFINED_POLICIES` which is as if it was defined to 1 (and not 0, the preprocessor default value). I have updated that file.

test/support/utils_sycl.h
```diff
@@ -32,6 +32,7 @@
 #include "utils_invoke.h"
 #include "utils_test_base.h"
 
+#define TEST_USE_PREDEFINED_POLICIES (!defined(ONEDPL_USE_PREDEFINED_POLICIES) || ONEDPL_USE_PREDEFINED_POLICIES)
```
adamfidel, line 35: Although this works, it produces a warning:
```
test/support/utils_sycl.h:100:9: warning: macro expansion producing 'defined' has undefined behavior [-Wexpansion-to-defined]
  100 | #    if TEST_USE_PREDEFINED_POLICIES
```
The following does not give a warning and I believe is semantically identical:
```suggestion
#ifdef ONEDPL_USE_PREDEFINED_POLICIES
#  define TEST_USE_PREDEFINED_POLICIES ONEDPL_USE_PREDEFINED_POLICIES
#else
#  define TEST_USE_PREDEFINED_POLICIES 1
#endif
```

test/general/fork.pass.cpp
```diff
@@ -0,0 +1,13 @@
...
+
+#if __has_include(<unistd.h>)
+#include <unistd.h>
+#endif
+
+int main(){
+#ifdef _POSIX_VERSION // defined in <unistd.h>
+    [[maybe_unused]] pid_t pid = fork();
```
mmichel11, line 10: Would it be good to wait in the parent process and check the child process' exit status to make sure nothing went wrong?

akukanov: Why not; added.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,87 @@ namespace execution
...
+        if (this != &__h)
+            __q = __h.__get_queue();
+        return *this;
+    }
+
+    __queue_holder&
+    operator=(__queue_holder&& __h)
+    {
+        if (this != &__h)
+            __q = __h.__get_queue();
```
SergeyKopienko, line 106: Looks like simple assignment, move should work a little bit faster.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -23,6 +23,10 @@
 #include "sycl_defs.h"
 
 #include <type_traits>
+#include <utility> // std::pair
```
SergeyKopienko, line 26: Is these comments really required?

danhoeflinger (commented):
I think this looks good; a minor comment and a question.

test/general/fork.pass.cpp
```diff
@@ -0,0 +1,27 @@
+#include <oneapi/dpl/execution>
```
mmichel11, line 1: Do we need a header copyright comment for this test, or is it unnecessary?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: No. `__get_queue` returns an rvalue object, not a reference. Compilers should use return value optimization (RVO) techniques to not even create that object, but directly update the left side of the assignment instead. Adding `std::move`` will unfortunately prohibit RVO. as far as I understand. @rarutyun can comment more on this.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: No, these are not required, but I personally find it quite useful to explain why this or that standard header is included, especially when you need only a thing or two.

mmichel11 (commented):
Small question from me about the test. I've looked through the rest of the PR and it LGTM.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: Yes. And this mean that probably this operation should be implemented by some another way. For example - to avoid `syc::queue` creation inside `__get_queue` call.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,87 @@ namespace execution
...
+    // Then a pointer to a factory function is stored.
+    __queue_holder(__global_instance_tag, _Queue_factory __f) : __flag_and_factory(0, __f) {}
+#endif
+
+    ~__queue_holder()
+    {
+        if (__has_queue())
+            __q.~queue();
+        else
+            __flag_and_factory.second = nullptr;
```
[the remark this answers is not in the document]
akukanov: No, no other reason - just cleaning things up. Compilers can eliminate that assignment if this is the last thing happening in the object lifetime and the modified value is not read afterwards.
One can imagine that if there was a conservative garbage collector running and interpreting anything that looks like a pointer to be a potentially valid pointer, then nullification of pointers would probably make real difference (though I guess I would need to add something else to support such garbage collection).

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,87 @@ namespace execution
 inline namespace __dpl
 {
 
+namespace __internal
+{
+
+#if _ONEDPL_PREDEFINED_POLICIES
+struct __global_instance_tag {};
+#endif
+using _Queue_factory = sycl::queue (*)();
```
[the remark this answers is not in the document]
akukanov: That's a valid point, thanks. I prefer to make it clear that this is a type, so `_QueueFactory` is my preference.

test/general/fork.pass.cpp
akukanov: We do need some header, thanks for reminding.

danhoeflinger (approved):
LGTM, 
I looked at the clang formatting complaints.  I would adopt the ones suggesting a space after `/*factory*/`, but it is just preference.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: `__get_queue` can only return an object, because that's what the factory produces if it is used in case the holder has no valid queue.
I will think though if something reasonable can be done.

akukanov: Something like this can be done in order to move a queue in case `__h` already holds it (essentially, to move a `shared_ptr` instead of copying):
```
    __queue_holder(__queue_holder&& __h) {
        if (h.__has_queue())
            __q = std::move(__h.__g); // perhaps there should be placement new, for visibility if nothing else
        else
            __q = (__h.__flag_and_factory.second)(); // here as well
    }

    __queue_holder&
    operator=(__queue_holder&& __h)
    {
        if (this != &__h) {
            if (h.__has_queue())
                __q = std::move(__h.__g);
            else
                __q = (__h.__flag_and_factory.second)();
        }
        return *this;
    }
```
Opinions are welcome on whether it makes sense. I do not see how that can be simplified, given that the factory returns a temporary object. not a reference.

mmichel11 (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: Probably at least we may incapsulate this code into some private static function and pass `*this` and `__queue_holder&& __h` into it.

danhoeflinger: If I'm understanding correctly, I think something like what is proposed here is a benefit.  I don't like that we are "reaching in" to the `__h` object to be able to get a return type that we can optimize appropriately.  However, I think it comes with the territory of this sort of "split" union type.  At least the specialized code still lives within this `__queue_holder` class and it does not infect other places with complexity.

I agree with Sergey's suggestion that we can consolidate the repeated code for the two move routines.

akukanov: Upon the second thought, it is simpler than the above. For move operations, we should require that the moved policy has a proper queue, because moving the predefined device policies is a horrible idea.

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,92 @@ namespace execution
...
+#if _ONEDPL_PREDEFINED_POLICIES
+        // If the first size-of-pointer bytes are zeros, we consider there is no valid queue.
+        res = (__flag_and_factory.first != 0);
+        std::atomic_signal_fence(std::memory_order_acq_rel); // to prevent possible reordering due to type punning
+#endif
+        return res;
+    }
+
+  public:
+    template <typename... _Args>
```
rarutyun, line 72:
```suggestion
    template <typename... Args, std::enable_if_t<!(sizeof...(Args) == 1 && std::is_same_v<std::remove_cvref_t<std::tuple_element_t<0, std::tuple<Args...>>>, my<T>>), int> = 0>
```

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,92 @@ namespace execution
...
+using _QueueFactory = sycl::queue (*)();
+
+class alignas(sycl::queue) __queue_holder
+{
+    static_assert(sizeof(sycl::queue) >= sizeof(std::pair<uintptr_t, _QueueFactory>));
+    static_assert(alignof(sycl::queue) >= alignof(std::uintptr_t));
+
+    union
+    {
+        sycl::queue __q;
```
rarutyun, line 55: seems like we didn't break the ABI, if I am not mistaken. That's great!

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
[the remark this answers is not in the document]
rarutyun: Likely it's going to be no-op anyway with O3 (or even O2)

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,92 @@ namespace execution
...
+
+    ~__queue_holder()
+    {
+        if (__has_queue())
+            __q.~queue();
+        else
+            __flag_and_factory.second = nullptr;
+    }
+
+    // Copy and move operations have to be explicit
```
rarutyun, line 90:
```suggestion
    // Copy and move operations have to be user-defined
```
or something like that.

Sorry for being nerd but it might be confusing for somebody when you read the comment near constructor that is has to be explicit but then you don't use `explicit` keyword :)

Don't get me wrong, I totally understand what you mean. I am trying to avoid confusion in the future

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -43,54 +133,76 @@ struct DefaultKernelName;
...
+
+  protected:
+    device_policy(__internal::__global_instance_tag __t, __internal::_QueueFactory __f) : q(__t, __f) {}
+#endif
+
   private:
-    sycl::queue q;
+    static_assert(sizeof(__internal::__queue_holder) == sizeof(sycl::queue));
+    static_assert(alignof(__internal::__queue_holder) == alignof(sycl::queue));
+    __internal::__queue_holder q;
```
rarutyun, line 169: I understand that it's potentially more changes than "really necessary" but how about renaming `q`. for example to `__holder`.. The reason why I am asking about that is `queue` public method implementation. `q.__get_queue` looks weird to me.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -43,54 +133,76 @@ struct DefaultKernelName;
 template <typename KernelName = DefaultKernelName>
 class device_policy
 {
+    static inline sycl::queue
+    __get_default_queue()
+    {
+        static sycl::queue __q(sycl::default_selector_v);
```
rarutyun, line 139: I started to think about multithreading but seems like we already solved everything. Thanks to magic static!

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,92 @@ namespace execution
...
+        return *this;
+    }
+
+    __queue_holder&
+    operator=(__queue_holder&& __h)
+    {
+        if (this != &__h)
+        {
+            // predefined policies should never be moved, so the move-from one must have a queue
+            assert(h.__has_queue());
```
rarutyun, line 109: How about the same `assert` in move constructor?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,92 @@ namespace execution
...
+
+class alignas(sycl::queue) __queue_holder
+{
+    static_assert(sizeof(sycl::queue) >= sizeof(std::pair<uintptr_t, _QueueFactory>));
+    static_assert(alignof(sycl::queue) >= alignof(std::uintptr_t));
+
+    union
+    {
+        sycl::queue __q;
+        std::pair<uintptr_t, _QueueFactory> __flag_and_factory;
```
rarutyun, line 56: I don't see that you use it as a pointer (like it was in previous versions of implementation). Should it still be `uintptr_t`? Maybe `bool` is good enough? (FWIW, I would not care about padding or swap `first` and `second`)

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,92 @@ namespace execution
...
 
+namespace __internal
+{
+
+#if _ONEDPL_PREDEFINED_POLICIES
+struct __global_instance_tag {};
+#endif
+using _QueueFactory = sycl::queue (*)();
+
+class alignas(sycl::queue) __queue_holder
```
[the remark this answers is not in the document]
akukanov: It also make the code self-documented, being explicit that same alignment as for sycl::queue is the purpose.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: No, `bool` is not good enough. We want to check all bytes of the pointer not being zero, and not just a single byte.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,92 @@ namespace execution
...
+        // If the first size-of-pointer bytes are zeros, we consider there is no valid queue.
+        res = (__flag_and_factory.first != 0);
+        std::atomic_signal_fence(std::memory_order_acq_rel); // to prevent possible reordering due to type punning
+#endif
+        return res;
+    }
+
+  public:
+    template <typename... _Args>
+    __queue_holder(_Args... __args) : __q(std::forward<_Args>(__args)...) {}
```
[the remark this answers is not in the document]
akukanov: I still do not get the second part; give an example perhaps?

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: OK, I will change in a different way.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: No problem. Right, I did not want to change what does not have to be changed, but now you gave me an excuse :)

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,92 @@ namespace execution
 inline namespace __dpl
 {
 
+namespace __internal
+{
+
+#if _ONEDPL_PREDEFINED_POLICIES
+struct __global_instance_tag {};
+#endif
+using _QueueFactory = sycl::queue (*)();
```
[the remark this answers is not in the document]
akukanov: Well, if you **strongly** prefer... :)

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: I have done this exercise (declare `dpcpp_default` and 'dpcpp_fpga` as `const`) in https://github.com/oneapi-src/oneDPL/pull/1671 and waiting CI.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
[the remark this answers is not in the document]
akukanov: I hope this will work, and it is a better code anyway because again the purpose is clearly seen from the code itself:
```
    template<typename T, typename = std::enable_if_t<std::is_constructible_v<sycl::queue, T&&>>>
    __queue_holder(T&& __qarg) : __q(std::forward<T>(__qarg)) {}
```

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -33,6 +37,92 @@ namespace execution
...
+#if _ONEDPL_PREDEFINED_POLICIES
+        // If the first size-of-pointer bytes are zeros, we consider there is no valid queue.
+        res = (__flag_and_factory.first != 0);
+        std::atomic_signal_fence(std::memory_order_acq_rel); // to prevent possible reordering due to type punning
+#endif
+        return res;
+    }
+
+  public:
+    template <typename T, typename = std::enable_if_t<std::is_constructible_v<sycl::queue, T&&>>>
```
SergeyKopienko, line 72: As for me, it's a little bit strange that we change params pack to one param here but checking is constructor available.
From one point of view, we know how we going to use this code: only with one param and this check isn't really required. From another point of view, if we think we should know the usage of this code, params pack looks better: for example, because `sycl::queue` has a lot of formats of constructor:
```C++
  explicit queue(const property_list& propList = {});

  explicit queue(const async_handler& asyncHandler,
                 const property_list& propList = {});

  template <typename DeviceSelector>
  explicit queue(const DeviceSelector& deviceSelector,
                 const property_list& propList = {});

  template <typename DeviceSelector>
  explicit queue(const DeviceSelector& deviceSelector,
                 const async_handler& asyncHandler,
                 const property_list& propList = {});

  explicit queue(const device& syclDevice, const property_list& propList = {});

  explicit queue(const device& syclDevice, const async_handler& asyncHandler,
                 const property_list& propList = {});

  template <typename DeviceSelector>
  explicit queue(const context& syclContext,
                 const DeviceSelector& deviceSelector,
                 const property_list& propList = {});

  template <typename DeviceSelector>
  explicit queue(const context& syclContext,
                 const DeviceSelector& deviceSelector,
                 const async_handler& asyncHandler,
                 const property_list& propList = {});

  explicit queue(const context& syclContext, const device& syclDevice,
                 const property_list& propList = {});

  explicit queue(const context& syclContext, const device& syclDevice,
                 const async_handler& asyncHandler,
                 const property_list& propList = {});
```

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -43,54 +133,76 @@ struct DefaultKernelName;
...
 #if _ONEDPL_FPGA_DEVICE
 struct DefaultKernelNameFPGA;
+
 template <unsigned int factor = 1, typename KernelName = DefaultKernelNameFPGA>
 class fpga_policy : public device_policy<KernelName>
 {
     using base = device_policy<KernelName>;
 
+    static sycl::queue
+    __get_fpga_default_queue()
```
SergeyKopienko, line 181: Probably we may combine this function with `__get_fpga_default_queue()` and pass different arguments as params.
In this case this function `__get_fpga_default_queue()` will not required anymore.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: > From one point of view, we know how we going to use this code: only with one param and this check isn't really required. 

That's the view I take - it's an internal class which usage we fully control, and at this time we do not need more than one parameter. As of now, we do not need an abstract holder with an arbitrary number of parameters supported.

One specific thing with variadic pack ctor which I did not even realize is that it served as a default ctor for `__queue_holder` as well. Once I removed it, it became obvious that `device_policy() = default;` needs attention within this patch, and I have made it more solid.

SergeyKopienko: OK

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: No, we cannot; an important part is the static `queue` instance that the function "holds", which needs to be different for the `dpcpp_fpga` policy.

SergeyKopienko: Understood, agree. You are right.

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: > I cannot add it to the member init list, or can I? Changing the list to a placement new is possible but I do not see it worthful. Meanwhile, we can consider making the predefined policies `const` as Sergey suggested, which would make these assertions even less useful.

If you have some special static function, which receive something as param and return it, I hope you may use in in initialization list and it can contain `assert` check inside.

akukanov: OK, so it would be something like this:
```
    sycl::queue&&
    __move_queue() &&
    {
        // predefined policies should never be moved, so an rvalue holder must have a queue
        assert(h.__has_queue());
        return std::move(__q);
    }

    __queue_holder(__queue_holder&& __h) : __q(std::move(__h).__move_queue()) {}

    __queue_holder&
    operator=(__queue_holder&& __h)
    {
        if (this != &__h)
            __q = std::move(__h).__move_queue();
        return *this;
    }
```
Do we want this? :)

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
rarutyun: I got it, thanks

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -46,26 +159,36 @@ class device_policy
...
     template <typename OtherName>
-    device_policy(const device_policy<OtherName>& other) : q(other.queue())
-    {
-    }
-    explicit device_policy(sycl::queue q_) : q(q_) {}
-    explicit device_policy(sycl::device d_) : q(d_) {}
-    operator sycl::queue() const { return q; }
+    device_policy(const device_policy<OtherName>& other) : __qh(other.queue()) {}
+    explicit device_policy(sycl::queue q) : __qh(q) {}
+    explicit device_policy(sycl::device d) : __qh(d) {}
```
rarutyun, line 166:
```suggestion
    explicit device_policy(sycl::queue q) : __qh(std::move(q)) {}
    explicit device_policy(sycl::device d) : __qh(std::move(d)) {}
```

I understand that this is not a part of this PR but would you be so kind to make this improvement? :)

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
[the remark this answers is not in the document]
rarutyun: I think I still disagree that it's helpful. I believe it's redundant. But I don't want to hold a useful PR because of that.

rarutyun (commented):
My comments are minor, I want to double check the behavior of move semantics and then press approve

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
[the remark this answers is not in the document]
akukanov: Thanks.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
rarutyun: My slight preference is to have assertions because it's just safer. I don't know who is going to move from "alien" global objects but if we want to protect users from themselves let's do it in both assignment and constructor.

I know that we have a PR in progress, which adds `const` to global policies. If we decide to make them `const` I would prefer to remove asserts for all move semantics.

rarutyun (approved):
I go ahead and approve this pull request assuming:
- We resolve the question with asserts in move semantics one way or the other
- We know what to do with clang-format failures (FWIW, I briefly looked on those and I don't think all of them are awful)

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: I think we incline to add constness to global policies as in #1671, as well as fix the specification in that regard. I believe that makes the asserts useless as a typical attempt (i.e. without `const_cast` involved) to move from a global policy will result in a compilation error.

For this reason, I am going to remove the assertion. Nevertheless, I think it makes sense to document that a program which moves or destroys `dpcpp_default` etc. has undefined behavior.

akukanov:
> what to do with clang-format failures

My viewpoint: ignore the remaining failures.
- Splitting an empty class/function body into two extra lines, each only containing an open/closing brace and maybe a semicolon is a waste of space and should not be done.
- Indenting preprocessing directives (`#if` etc.), mixed with the regular C++ code that has its own indentation is ugly and should not be done.
- Adding a space between a function argument and the comment preceding and describing the argument slightly worsens readability, as it becomes somewhat less clear (visually) what the comment refers to. Modern IDEs use different coloring for comments and names, so in most cases the adjacent tokens remain visible.

rarutyun (approved):

SergeyKopienko (approved):
LGTM

PR #1656, opened 2024-07-02
Avoid extra sycl::buffer tests

SergeyKopienko:
Previously `sycl::buffer` tests rans from `test1buffer`, `test2buffers`, `test3buffers` and from `test4buffers` functions for each type of `alloc_type` template param: for `sycl::usm::alloc::shared` and for `sycl::usm::alloc::device`.

It's look like as extra overhead which we should avoid to reduce CI test time.

Now `sycl::buffer` tests runs only for `alloc_type` template param equal to `sycl::usm::alloc::shared` (by default).

### DIFF
--- a/test/support/utils_sycl.h
+++ b/test/support/utils_sycl.h
@@ -111,7 +111,15 @@ sycl::queue get_test_queue()
     return my_queue;
 }
 
-template <sycl::usm::alloc alloc_type, typename TestValueType, typename TestName>
+template <sycl::usm::alloc alloc_type>
+constexpr bool
+required_test_sycl_buffer()
+{
+    return alloc_type == sycl::usm::alloc::shared;
+}
+
+template <sycl::usm::alloc alloc_type, typename TestValueType, typename TestName,
+          bool TestSyclBuffer = required_test_sycl_buffer<alloc_type>()>
 void
 test1buffer(float ScaleStep = 1.0f, float ScaleMax = 1.0f)
 {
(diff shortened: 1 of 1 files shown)


test/support/utils_sycl.h
```diff
@@ -111,7 +111,15 @@ sycl::queue get_test_queue()
...
-template <sycl::usm::alloc alloc_type, typename TestValueType, typename TestName>
+template <sycl::usm::alloc alloc_type>
+constexpr bool
+required_test_sycl_buffer()
+{
+    return alloc_type == sycl::usm::alloc::shared;
+}
+
+template <sycl::usm::alloc alloc_type, typename TestValueType, typename TestName,
+          bool TestSyclBuffer = required_test_sycl_buffer<alloc_type>()>
```
danhoeflinger, line 122: Should we have the defaulted template parameter at both "levels"?
I think this overload is only called from within this file, so perhaps it should require explicit parameters.

Same for the 2, 3 buffer cases.

SergeyKopienko: Fixed

danhoeflinger (dismissed):
LGTM if CI comes back green.

danhoeflinger (approved):
Agree with reverting.  I missed that test3buffers is used at both levels.
LGTM with green CI.

SergeyKopienko:
> Agree with reverting. I missed that test3buffers is used at both levels. LGTM with green CI.

Not only `test3buffers` :
- for example, also we have the call of `test1buffer` like
```C++
test1buffer<sycl::usm::alloc::device, ValueType, test_is_heap<ValueType, PermItIndexTag>>();
```

danhoeflinger:
> > Agree with reverting. I missed that test3buffers is used at both levels. LGTM with green CI.
> 
> Not only `test3buffers` :
> 
> * for example, also we have the call of `test1buffer` like
> 
> ```c++
> test1buffer<sycl::usm::alloc::device, ValueType, test_is_heap<ValueType, PermItIndexTag>>();
> ```

Ah, OK.  Its confusing to position the overloads like this with both different template parameters and defaulted normal parameters.  I'd prefer these overloads to be named differently for clarity (but that's just my opinion, of course).  In any case, its clear that the default is needed in both places.

julianmi (approved):
LGTM

PR #1659, opened 2024-07-03
[CI] Update MacOS version, move to Clang++ from ICPC

dmitriy-sobolev:
MacOS 11 was removed from GitHub CI on June 28 according to https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners. It has been replaced with MacOS 13.

Intel® C++ Compiler Classic Release was discontinued in the oneAPI 2024.0 release. It's been replaced with Clang++, which tends to be a default c++ compiler on MacOS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -388,9 +388,10 @@ jobs:
           echo "::warning::CMake: $(cmake --version)"
           sysctl -a | grep machdep.cpu
           # workaround for CMake not being able to find OpenMP: see https://discourse.cmake.org/t/how-to-find-openmp-with-clang-on-macos/8860
-          export OpenMP_ROOT=$(brew --prefix)/opt/libomp 
+          # -DCMAKE_POLICY_DEFAULT_CMP0074=NEW below is forced to make sure CMake uses <PackageName>_ROOT variables.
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -373,31 +371,23 @@ jobs:
...
+          - os: macos-13
+            cxx_compiler: clang++
             std: 17
             build_type: release
             backend: omp
     steps:
       - uses: actions/checkout@v3
-      - name: Install Intel® oneAPI C++ Compiler Classic and Intel® oneAPI Threading Building Blocks
+      - name: Install OpenMP for Clang++
+        if: (matrix.backend == 'omp' && matrix.cxx_compiler == 'clang++')
```
timmiesmith, line 382: Is the if statement here necessary given that it exactly matches the matrix defined on the lines above?

timmiesmith (dismissed):
LGTM. If the if check in the Install OpenMP isn't needed I can reapprove when you remove it.

.github/workflows/ci.yml
dmitriy-sobolev: Done

timmiesmith (approved):
Looks good to me. Thank you.

PR #1662, opened 2024-07-03
[PROTOTYPE] Removing identity and default construction requirement from two_pass_scan

danhoeflinger:
This PR removes the identity requirement and the default constructor requirements from two_pass_scan to be more general so that it may cover more cases from mainline oneDPL.

It currently relies on a few strategies to do so:
1) Uses a lazy-constructing union in place of some value types. This allows us to avoid default construction, but allows variables to be scoped where we want them.  
2) delays creation of value type objects to interior scope, rather than reusing the same variable. 
3) Early exit for empty/ inactive workgroups, and properly avoiding empty / inactive subgroups for specific operations
 ** this seems to have the added benefit of performance improvement for small values of `n` **
4) Adapting the subgroup scan to make the `init_and_carry` parameter an optional input (as dictated by a template parameter), which lazily constructs it as an output if it does not exist, and uses and then overwrites if it is provided.  Then, first iterations of loops within a subgroup are pulled out to "skip" the input carry when it does not exist.

This PR also re-writes the "block-global" input carry calculated for `g > 0`  near the beginning of the second kernel.  The previous logic was convoluted and unnecessary, and this PR simplifies it and makes it easier to avoid the identity within the calculation.

This also adds a set of test cases which cover a type which has no default ctor, and no known identity to use. 

Performance:
It seems this PR does not generally harm performance beyond ~3% in the very large values of `n`, and improves performance for small `n` up to ~20%.  We can continue to improve those cases.  There is opportunity to make small adjustments which may regain any losses here, we will see.

Remaining work:
`std::inclusive_scan` optionally does not have an init value, whereas this code takes advantage of a guaranteed `init` value for both inclusive and exclusive scan.  Some more work is needed to specialize for the zeroth block of the cases without an incoming init.  This should be reasonably doable.  I will attempt to do it within this PR, but we can opt to do this as a follow up PR, if it takes too long.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/two_pass_scan.h
+++ b/include/oneapi/dpl/experimental/kt/two_pass_scan.h
@@ -185,10 +185,6 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
     size_t M = __in_rng.size();
     size_t num_remaining = M;
 
-    //    static_assert(oneapi::dpl::unseq_backend::__has_known_identity<BinaryOp, ValueType>::value,
-    //                  "The prototype currently supports only known identity operators + init type");
-    //    constexpr ValueType identity = oneapi::dpl::unseq_backend::__known_identity<BinaryOp, ValueType>;
-
     constexpr int J_max = 128;
     std::size_t MAX_INPUTS_PER_BLOCK = work_group_size * J_max * num_work_groups; // empirically determined for reduce_then_scan
     // items per PVC hardware thread
@@ -235,7 +231,7 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
                 oneapi::dpl::__par_backend_hetero::__lazy_ctor_storage<ValueType> sub_group_carry;
                 std::size_t group_start_idx = (b * blockSize) + (g * K * num_sub_groups_local);
                 if (M <= group_start_idx)
-                    return; // exit early for empty groups
+                    return; // exit early for empty groups (TODO: avoid launching these?)
 
                     
                 std::size_t elements_in_group = std::min(M - group_start_idx, std::size_t(num_sub_groups_local * K));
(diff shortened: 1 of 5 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -767,6 +767,22 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
     }
 };
 
+template <typename _Tp>
+union __lazy_ctor_storage
```
mmichel11, line 771: Would the location of this union be better suited as a general utility in `pstl/utils.h` as it is not SYCL specific?

include/oneapi/dpl/experimental/kt/two_pass_scan.h
```diff
@@ -56,45 +57,108 @@ sub_group_masked_scan(const SubGroup& sub_group, MaskOp mask_fn, InitBroadcastId
             value = binary_op(partial_carry_in, value);
         }
     }
-    value = binary_op(init, value);
-    // Adjust for an exclusive scan if requested
-    if constexpr (!Inclusive)
+    if constexpr (InitPresent)
```
mmichel11, line 60: Can this conditional be combined with the one below?

include/oneapi/dpl/experimental/kt/two_pass_scan.h
```diff
@@ -56,45 +57,108 @@ sub_group_masked_scan(const SubGroup& sub_group, MaskOp mask_fn, InitBroadcastId
...
+    _ONEDPL_PRAGMA_UNROLL
+    for (std::uint8_t shift = 1; shift <= VL / 2; shift <<= 1)
+    {
+        auto partial_carry_in = sycl::shift_group_right(sub_group, value, shift);
+        if (mask_fn(sub_group_local_id, shift))
+        {
+            value = binary_op(partial_carry_in, value);
+        }
+    }
+    if constexpr (InitPresent)
```
mmichel11, line 106: Similar comment to the exclusive scan case. We should be able to combine the two conditionals.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
```diff
@@ -319,45 +418,61 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
...
+                                auto reduction_idx = i * num_sub_groups_local * VL + num_sub_groups_local * sub_group_local_id + offset;
+                                value.__v = tmp_storage[reduction_idx];
+                                sub_group_scan<VL, true, true>(sub_group, value.__v, binary_op, carry_last);
+                            }
+
+                            // final partial iteration
+                            auto proposed_idx = (pre_carry_iters - 1) * num_sub_groups_local * VL + num_sub_groups_local * sub_group_local_id + offset;
+                            auto remaining_elements = elements_to_process - ((pre_carry_iters - 1) * VL);
+                            auto reduction_idx = (proposed_idx < subgroups_before_my_group) ? proposed_idx : subgroups_before_my_group - 1;
+                            value.__setup(tmp_storage[reduction_idx]);
```
mmichel11, line 473: Should we call `__setup` here? `value` has already been constructed here.

mmichel11 (commented):
First pass from my side.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
```diff
@@ -174,48 +228,81 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
...
-                    for (int j = 0; j < J_max; j++)
+                    // adjust for lane-id
+                    // compute sub-group local pfix on T0..63, K samples/T, send to accumulator kernel
+                    if (is_full_thread && is_full_block)
                     {
-                        v = unary_op(__in_rng[start_idx + j * VL]);
+                        auto v = unary_op(__in_rng[start_idx]);
                         // In principle we could use SYCL group scan. Stick to our own for now for full control of implementation.
-                        std::tie(v, sub_group_carry) = sub_group_scan<VL, Inclusive>(sub_group, std::move(v), binary_op, std::move(sub_group_carry));
+                        sub_group_scan<VL, Inclusive, false>(sub_group, v, binary_op, sub_group_carry);
```
adamfidel, line 251: For calls to `sub_group_scan`, can we annotate the last template parameter. For example:
```suggestion
                        sub_group_scan<VL, Inclusive, /*InitPresent=*/false>(sub_group, v, binary_op, sub_group_carry);
```

include/oneapi/dpl/experimental/kt/two_pass_scan.h
danhoeflinger: Yes, thanks

include/oneapi/dpl/experimental/kt/two_pass_scan.h
danhoeflinger: Yes, thanks

include/oneapi/dpl/experimental/kt/two_pass_scan.h
danhoeflinger: good catch.  fixed.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
```diff
@@ -174,48 +228,81 @@ two_pass_scan(sycl::queue q, _InRng&& __in_rng, _OutRng&& __out_rng,
...
-                bool is_full_thread = start_idx + J * VL <= M;
-                // adjust for lane-id
-                start_idx += sub_group_local_id;
-                // compute sub-group local pfix on T0..63, K samples/T, send to accumulator kernel
-                if (is_full_thread && is_full_block)
+                    
+                std::size_t elements_in_group = std::min(M - group_start_idx, std::size_t(num_sub_groups_local * K));
+                auto active_subgroups = oneapi::dpl::__internal::__dpl_ceiling_div(elements_in_group, K);
+                std::size_t subgroup_start_idx = group_start_idx + (sub_group_id * K);
+                bool is_full_thread = subgroup_start_idx + J * VL <= M;
```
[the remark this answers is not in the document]
danhoeflinger: I believe you are correct.  I made the change, and will check it with tests shortly.

include/oneapi/dpl/experimental/kt/two_pass_scan.h
danhoeflinger: good suggestion.  Added labels for unclear template params to `sub_group_scan*`

include/oneapi/dpl/experimental/kt/two_pass_scan.h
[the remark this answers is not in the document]
mmichel11: `sub_group_start_idx + J * VL` would be the first index processed by the next thread which is why we have the `<=`.  If `J=1`, `M=32`, `VL=32`, and `subgroup_start_idx = 0`, then we would have a full thread but `subgroup_start_idx + J * VL = 32` which is why we need `<=`.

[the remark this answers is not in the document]
danhoeflinger: ah, yeah.  We are comparing two loop end-conditions, not an index to a loop end condition.  I'll revert it.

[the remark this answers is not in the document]
danhoeflinger: This has been reverted.

mmichel11 (commented):
I have looked through the full PR logic again and it LGTM. Two minor comments on formatting.

In general in our entire implementation, I think there is room for better organization / structuring. However, I think this is best done once we are actually developing in oneDPL  main.

mmichel11 (approved):
LGTM. Feel free to merge this into the prototype branch now or merge #1680 into this branch first when it's review is complete.

PR #1666, opened 2024-07-04
Some improvements of struct reduce_over_group

SergeyKopienko:


SergeyKopienko:
@julianmi could you please check some perf improvement with these changes?

julianmi:
> @julianmi could you please check some perf improvement with these changes?

The performance on PVC seems to be unchanged for 4-byte data. Large data types shouldn't see a change in performance either since they are passed by reference like on mainline.

SergeyKopienko:
> > @julianmi could you please check some perf improvement with these changes?
> 
> The performance on PVC seems to be unchanged for 4-byte data. Large data types shouldn't see a change in performance either since they are passed by reference like on mainline.

Thanks a lot, @julianmi, for this update!.
Еhe idea, unfortunately, was not useful.

PR #1667, opened 2024-07-04
Limit work-group size to fix FPGA slowdowns in find_or (#1653)

julianmi:
#1653 saw significant slowdowns in FPGA emulation over mainline. These are due to high work-group sizes (65536 on my test system). This limits the work-group size to 2048, which has shown to provide a good balance between throughput and required synchronization in group reductions.


SergeyKopienko (approved):
LGTM

PR #1668, opened 2024-07-05
Remove local (in-group) atomic usage from __parallel_find_or (V1)

SergeyKopienko:
In this PR we remove local (in-group) atomic usage from `__parallel_find_or` implementation:
- `__found_local` now is local variable (for each item);
- barriers doesn't required anymore;
- we updates the state of `__found_local` within one group through `__dpl_sycl::__reduce_over_group` operation :
```C++
                    __found_local = __dpl_sycl::__reduce_over_group(__item_id.get_group(), __found_local,
                                                                    typename _BrickTag::_LocalResultsReduceOp{});
```
where `_LocalResultsReduceOp` is :
 - `__dpl_sycl::__minimum` (for the `__parallel_find_forward_tag`);
 - `__dpl_sycl::__maximum` (for the `__parallel_find_backward_tag`);
 - `__dpl_sycl::__bit_or` (for the `__parallel_or_tag`).

The new operation `__dpl_sycl::__bit_or` has been defined in the file [sycl_defs.h](https://github.com/oneapi-src/oneDPL/pull/1668/files#diff-521f7db06a55567c1a1dffa855dde585fb6bc5fbe8633d19b528356e3a501ea0) :
```C++
#if _ONEDPL_SYCL2020_FUNCTIONAL_OBJECTS_PRESENT
// ...
template <typename _T = void>
using __bit_or = sycl::bit_or<_T>;
#else  // _ONEDPL_SYCL2020_FUNCTIONAL_OBJECTS_PRESENT
// ...
template <typename _T = void>
using __bit_or = sycl::ONEAPI::bit_or<_T>;
#endif // _ONEDPL_SYCL2020_FUNCTIONAL_OBJECTS_PRESENT
```

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1035,7 +1035,7 @@ struct __parallel_or_tag
 {
     using _AtomicType = int32_t;
 
-    using _LocalResultsReduceOp = __dpl_sycl::__plus<_AtomicType>;
+    using _LocalResultsReduceOp = __dpl_sycl::__bit_or<_AtomicType>;
 
     // The template parameter is intended to unify __init_value in tags.
     template <typename _DiffType>
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -965,7 +965,7 @@ struct __parallel_find_forward_tag
     using _AtomicType = oneapi::dpl::__internal::__difference_t<_RangeType>;
 #endif
 
-    using FoundLocalReduceOp = __dpl_sycl::__minimum<_AtomicType>;
+    using _LocalResultsReduceOp = __dpl_sycl::__minimum<_AtomicType>;
 
     // The template parameter is intended to unify __init_value in tags.
     template <typename _DiffType>
@@ -1004,7 +1004,7 @@ struct __parallel_find_backward_tag
     using _AtomicType = oneapi::dpl::__internal::__difference_t<_RangeType>;
 #endif
 
-    using FoundLocalReduceOp = __dpl_sycl::__maximum<_AtomicType>;
+    using _LocalResultsReduceOp = __dpl_sycl::__maximum<_AtomicType>;
 
     template <typename _DiffType>
     constexpr static _AtomicType __init_value(_DiffType)
@@ -1035,7 +1035,7 @@ struct __parallel_or_tag
 {
     using _AtomicType = int32_t;
 
-    using FoundLocalReduceOp = __dpl_sycl::__plus<_AtomicType>;
+    using _LocalResultsReduceOp = __dpl_sycl::__plus<_AtomicType>;
 
     // The template parameter is intended to unify __init_value in tags.
     template <typename _DiffType>
@@ -1225,7 +1225,7 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
                     // or calculate the sum with __dpl_sycl::__plus (for the __parallel_or_tag)
                     // inside all our group items
                     __found_local = __dpl_sycl::__reduce_over_group(__item_id.get_group(), __found_local,
-                                                                    typename _BrickTag::FoundLocalReduceOp{});
+                                                                    typename _BrickTag::_LocalResultsReduceOp{});
 
                     // Set local found state value value to global atomic
                     if (__local_idx == 0 && __found_local != __init_value)
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1174,7 +1174,7 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     __wgroup_size = ::std::min(__wgroup_size, oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel));
 #endif
 
-#if _ONEDPL_FPGA_DEVICE
+#if _ONEDPL_FPGA_EMU
     // Limit the maximum work-group size to minimize the cost of work-group reduction.
     // Limiting this also helps to avoid huge work-group sizes on some devices (e.g., FPGU emulation).
     __wgroup_size = std::min(__wgroup_size, (std::size_t)2048);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1169,9 +1169,12 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     auto __kernel = __internal::__kernel_compiler<_FindOrKernel>::__compile(__exec);
     __wgroup_size = ::std::min(__wgroup_size, oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel));
 #endif
+
+#if _ONEDPL_FPGA_DEVICE
     // Limit the maximum work-group size to minimize the cost of work-group reduction.
     // Limiting this also helps to avoid huge work-group sizes on some devices (e.g., FPGU emulation).
     __wgroup_size = std::min(__wgroup_size, (std::size_t)2048);
+#endif
     auto __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
 
     auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -965,6 +965,8 @@ struct __parallel_find_forward_tag
     using _AtomicType = oneapi::dpl::__internal::__difference_t<_RangeType>;
 #endif
 
+    using FoundLocalReduceOp = __dpl_sycl::__minimum<_AtomicType>;
+
     // The template parameter is intended to unify __init_value in tags.
     template <typename _DiffType>
     constexpr static _AtomicType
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -139,6 +141,9 @@ using __maximum = sycl::ONEAPI::maximum<_T>;
 
 template <typename _T>
 using __minimum = sycl::ONEAPI::minimum<_T>;
+
+template <typename _T = void>
+using __bit_or = sycl::ONEAPI::bit_or<_T>;
```
SergeyKopienko, line 146: Described at https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/removed/sycl_ext_oneapi_group_algorithms.asciidoc

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -130,6 +130,8 @@ using __maximum = sycl::maximum<_T>;
 template <typename _T = void>
 using __minimum = sycl::minimum<_T>;
 
+template <typename _T = void>
+using __bit_or = sycl::bit_or<_T>;
```
SergeyKopienko, line 134: Described at https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.identities

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1139,11 +1168,17 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
 #if _ONEDPL_COMPILE_KERNEL
     auto __kernel = __internal::__kernel_compiler<_FindOrKernel>::__compile(__exec);
     __wgroup_size = ::std::min(__wgroup_size, oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel));
 #endif
+
+#if _ONEDPL_FPGA_DEVICE
+    // Limit the maximum work-group size to minimize the cost of work-group reduction.
+    // Limiting this also helps to avoid huge work-group sizes on some devices (e.g., FPGU emulation).
+    __wgroup_size = std::min(__wgroup_size, (std::size_t)2048);
+#endif
```
julianmi, line 1181: This limit was not intended for FPGA only. We don't want overly large work-groups on any device to not slow down the work-group reduction too much and prevent running in potential SLM limits. I would propose removing the if-statement. We might even limit this to 1024, which is picked on PVC anyways.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1182,37 +1215,35 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+                        _BrickTag::__save_state_to_atomic(__found, __found_local);
                     }
                 });
         });
         //The end of the scope  -  a point of synchronization (on temporary sycl buffer destruction)
     }
 
     if constexpr (__or_tag_check)
-        return __result;
+        return __result != __init_value;
```
julianmi, line 1246: Is this change really needed? Wouldn't `__result` be 0 anyways if no match is found?

SergeyKopienko: We should return the result of `bool` type in this case, so this expression looks like more correct then implicit conversion of some numeric type to `bool` type, I believe.

SergeyKopienko: Also this expression looks more clear from my point of view.

julianmi: I'm fine with your change. I just think there is no semantical difference. We could switch to `bool` in general. I don't expect much performance benefit from this though since this is only a single value in registers in `__early_exit_find_or` and we need an atomic type in the global update anyways.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1139,11 +1168,17 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
 #endif
+
+#if _ONEDPL_FPGA_DEVICE
+    // Limit the maximum work-group size to minimize the cost of work-group reduction.
+    // Limiting this also helps to avoid huge work-group sizes on some devices (e.g., FPGU emulation).
+    __wgroup_size = std::min(__wgroup_size, (std::size_t)2048);
+#endif
     auto __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
 
     auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size);
```
julianmi, line 1184: I've looked into oversubscribing the number of groups to hide latencies on the device, but this shows slowdowns on PVC. I expect this to be due to the additional congestion incurred for the global atomic updates. I think the `__n_groups` calculation is fine for now and we can remove the todo comment in the line below.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: In this PR I hadn't the goal to define `__wgroup_size` by any other way. It's changed here under this define (initially in your PR https://github.com/oneapi-src/oneDPL/pull/1667, thanks @julianmi) just for avoid long compilation time.
Let's do any changes about `__wgroup_size` in separate PRs.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: This `bool` returning value is not my change: it has place before.

SergeyKopienko: Resolved?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
julianmi: The issue with the test timeouts is not specific to FPGAs. This issue occurs on all device that support very large work-groups. This leads to more complex reductions which typically require more SLM. This slows the execution down and might even lead to timeouts as we've experienced. So, I don't think this is a compilation issue.

The proposed change in https://github.com/oneapi-src/oneDPL/pull/1667 limits the group size to a reasonable limit and should be safe on all architectures. Typical CPUs and GPUs group sizes are below this threshold.
 
 What other changes to the work-group size would you propose in another PR?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
SergeyKopienko: Test evidence for complier icx 2021.1.2 (whene group operations isn't present) ): https://godbolt.org/z/PcnYT4vqe

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: I propose to remove in the separate PR `_ONEDPL_FPGA_DEVICE` define here and change the maximal available value from `2048` to something else.

danhoeflinger: Yes, I agree that the change should be for all architectures as was done in #1667.  Generally, I think this is a good idea for most all of our patterns, as they are written with this sort of size in mind for the SYCL backend and it can lead to unexpected results for CPU and FPGA targets. Of course, there may be some reason for a specific algorithm to want larger workgroups when they are available.
  
I wonder if we would benefit from adding some cap like this within `__max_work_group_size` with some defaulted parameter which could be adjusted when specifically desired.  I think this could end up stopping these unexpected slowdowns / issues on non-GPU SYCL targets.

julianmi: > I wonder if we would benefit from adding some cap like this within `__max_work_group_size` with some defaulted parameter which could be adjusted when specifically desired. I think this could end up stopping these unexpected slowdowns / issues on non-GPU SYCL targets.

I agree with this. One drawback I see with adding this to `__max_work_group_size` is that it would require significant testing to not introduce any unwanted side effects. How about we introduce a helper such as `__limit_work_group_size` first for places where we explicitly limit this (e.g., `__parallel_find_or`, `__parallel_histogram_select_kernel`, `__parallel_transform_reduce`)? We could extend this over time and finally add something to `__max_work_group_size` or a similar generic helper.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
julianmi: My comment is not about the return type but the internal `_BrickTag::_AtomicType`. Anyways, there is no change required from my point of view.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Yes, something like that may be the better approach, and allow us to more deliberately add this to patterns while looking at them specifically, rather than just making a blanket change "silently".

SergeyKopienko:
@danhoeflinger, @julianmi is something that mandatory required to change in this PR?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1182,37 +1214,35 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
-                    // 2. Find any element that satisfies pred and set local atomic value to global atomic
+                    // 2. Find any element that satisfies pred
                     __pred(__item_id, __n_iter, __wgroup_size, __found_local, __brick_tag, __rngs...);
-                    __dpl_sycl::__group_barrier(__item_id);
 
-                    // Set local atomic value to global atomic
-                    if (__local_idx == 0)
+                    // 3. Reduce over group: find __dpl_sycl::__minimum (for the __parallel_find_forward_tag),
+                    // find __dpl_sycl::__maximum (for the __parallel_find_backward_tag)
+                    // or calculate the sum with __dpl_sycl::__plus (for the __parallel_or_tag)
```
julianmi, line 1225:
```suggestion
                    // or update state with __dpl_sycl::__bit_or (for the __parallel_or_tag)
```

julianmi (approved):
LGTM. Please wait for @danhoeflinger and green CI.

danhoeflinger (approved):
LGTM,  with the comment explaining the restriction of workgroup size, it is OK.  
I believe we can just make that restriction for all targets, but it should be examined, and we can do that in a separate PR.

julianmi (approved):

PR #1670, opened 2024-07-05
Fix mistakes in tests utils

SergeyKopienko:


danhoeflinger (approved):
yes LGTM.  Thanks for correcting these.

PR #1671, opened 2024-07-05
Declare predefined `dpcpp_default` and `dpcpp fpga` policies as `const`

SergeyKopienko:
In this PR Declare predefined `dpcpp_default` and `dpcpp fpga` policies as `const`.

I hope it's may help to implement move-constructor and move-assignment in the PR https://github.com/oneapi-src/oneDPL/pull/1652 properly.

PR with the changes for oneDPL specification: https://github.com/uxlfoundation/oneAPI-spec/pull/554

### DIFF
--- a/documentation/CHANGES.rst
+++ b/documentation/CHANGES.rst
@@ -369,8 +369,8 @@ Changes to Existing Features
 
 - The following API elements of oneDPL were changed or removed:
 
-  - the ``default_policy`` object was renamed to ``dpcpp_default`` and declared as ``const``;
-  - the ``fpga_policy`` object was renamed to ``dpcpp_fpga`` and declared as ``const``;
+  - the ``default_policy`` object was renamed to ``dpcpp_default``;
+  - the ``fpga_policy`` object was renamed to ``dpcpp_fpga``;
   - the ``fpga_device_policy`` class was renamed to ``fpga_policy``;
   - the ``_PSTL_FPGA_DEVICE`` macro was renamed to ``ONEDPL_FPGA_DEVICE``;
   - the ``_PSTL_FPGA_EMU`` macro was renamed to ``ONEDPL_FPGA_EMULATOR``;
--- a/documentation/CHANGES.rst
+++ b/documentation/CHANGES.rst
@@ -369,8 +369,8 @@ Changes to Existing Features
 
 - The following API elements of oneDPL were changed or removed:
 
-  - the ``default_policy`` object was renamed to ``dpcpp_default``;
-  - the ``fpga_policy`` object was renamed to ``dpcpp_fpga``;
+  - the ``default_policy`` object was renamed to ``dpcpp_default`` and declared as ``const``;
+  - the ``fpga_policy`` object was renamed to ``dpcpp_fpga`` and declared as ``const``;
   - the ``fpga_device_policy`` class was renamed to ``fpga_policy``;
   - the ``_PSTL_FPGA_DEVICE`` macro was renamed to ``ONEDPL_FPGA_DEVICE``;
   - the ``_PSTL_FPGA_EMU`` macro was renamed to ``ONEDPL_FPGA_EMULATOR``;
--- a/documentation/library_guide/macros.rst
+++ b/documentation/library_guide/macros.rst
@@ -94,8 +94,8 @@ Macro                              Description
                                    such as ``dpcpp_default`` and ``dpcpp_fpga``. When the macro is not defined (by default)
                                    or evaluates to non-zero, predefined policies objects can be used.
                                    When the macro is set to 0, predefined policies objects and make functions
-                                   without arguments, when ``make_device_policy()``,
-                                   ``make_fpga_policy()``, are not available.
+                                   without arguments - ``make_device_policy()`` and
+                                   ``make_fpga_policy()`` - are not available.
 ---------------------------------- ------------------------------
 ``ONEDPL_ALLOW_DEFERRED_WAITING``  This macro allows waiting for completion of certain algorithms executed with
                                    device policies to be deferred. (Disabled by default.)
--- a/documentation/library_guide/macros.rst
+++ b/documentation/library_guide/macros.rst
@@ -94,8 +94,7 @@ Macro                              Description
                                    such as ``dpcpp_default`` and ``dpcpp_fpga``. When the macro is not defined (by default)
                                    or evaluates to non-zero, predefined policies objects can be used.
                                    When the macro is set to 0, predefined policies objects and make functions
-                                   without arguments - ``make_device_policy()`` and
-                                   ``make_fpga_policy()`` - are not available.
+                                   without arguments (``make_device_policy()`` and ``make_fpga_policy()``) are not available.
 ---------------------------------- ------------------------------
 ``ONEDPL_ALLOW_DEFERRED_WAITING``  This macro allows waiting for completion of certain algorithms executed with
                                    device policies to be deferred. (Disabled by default.)
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -101,7 +101,7 @@ class fpga_policy : public device_policy<KernelName>
 // Starting with c++17 we can simply define sycl as inline variable.
 #    if _ONEDPL___cplusplus >= 201703L
 
-inline device_policy<> dpcpp_default{};
+inline const device_policy<> dpcpp_default{};
 #        if _ONEDPL_FPGA_DEVICE
 inline fpga_policy<> dpcpp_fpga{};
 #        endif // _ONEDPL_FPGA_DEVICE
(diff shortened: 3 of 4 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov, line 1: If we make this change, should it also be reflected in the documentation and in the specification? Opinions welcome.

SergeyKopienko: I added one commit into documentation: https://github.com/oneapi-src/oneDPL/pull/1671/commits/c1c778f9389d793f222a67083f9759a582380281
Looks like there is no more places there some changes about these `const` default policy declarations are required.
@dcbenito How do you think?

documentation/CHANGES.rst
```diff
@@ -369,8 +369,8 @@ Changes to Existing Features
 
 - The following API elements of oneDPL were changed or removed:
 
-  - the ``default_policy`` object was renamed to ``dpcpp_default``;
-  - the ``fpga_policy`` object was renamed to ``dpcpp_fpga``;
+  - the ``default_policy`` object was renamed to ``dpcpp_default`` and declared as ``const``;
+  - the ``fpga_policy`` object was renamed to ``dpcpp_fpga`` and declared as ``const``;
```
akukanov, line 373: No, old release notes are a wrong place for this new information.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
akukanov: There are two separate questions.

First of all, we need to think whether the change violates the specification (https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.3-rev-1/elements/onedpl/source/parallel_api#dpc-execution-policy), whether the specification should be changed or better kept as is, and if changed - whether it is a compatible change. This is where I would like others to give their opinions.

Second, the documentation here https://oneapi-src.github.io/oneDPL/parallel_api/execution_policies.html should be updated accordingly. And at last, a new release note will need to be added.

danhoeflinger: Its an interesting question.  I think as written by the spec, it is technically a violation to change `dpcpp_default` to be `const`.  However, as the `device_policy` is implemented **and specified**, it seems that it is not possible to change it via any publicly defined interface as it has no public member fields, and no non-const public member functions. In the implementation, the only member field, is `const`.  So, it is practically `const`, just not labelled as such. 
 
Therefore, in practice, I don't think this change can break any user code, even if it technically violates the specification. I think it is a compatible change, but more thought should be given to this.  I think the spec should be updated to include `const`.

documentation/CHANGES.rst
SergeyKopienko: Fixed (rolled back).

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
rarutyun: Technically it's a breaking change (believe it or not) because it affects `dpcpp_default` type (for example, in template instantiation). Please look at [this](https://godbolt.org/z/nbxbGcxj5) example.

I am not saying that this very use case is going to be broken. I am just saying that since it affects the type of the deduced parameters (as well as `decltype(dpcpp_default)` expression) it might break template/metaprogramming code as well as it might affect overload resolution (look [here](https://godbolt.org/z/Ere871a3c))

Whether it's going to break any of the user code in practice or not, I am not sure, but it's definitely possible

akukanov: OK, so if we make this change, some codes might stop working or change behavior. That does not sound good, but let's take a closer look.

First, what can users do with a non-const instance of `device_policy` which cannot be done with a const instance? As Dan noted, the specified methods of `device_policy` are non-modifying, so there is no difference. What remains, I think, are compiler-generated methods, i.e. copy/move constructors and assignments. That means, one can change a policy object by assigning it a new value or by moving from it to another object. These use cases are permitted - except that for the predefined policies it should better be forbidden, as such changes might make `dpcpp_default` etc. semantically invalid. I do not believe this will result in any practical functionality loss, as a copy of `dpcpp_default` can be made and then re-assigned or moved without problems.

Now let's look at Ruslan's examples:

- the first one simply ignores cv qualifiers when defining/using its `is_device_policy` trait; i.e. it will not work correctly with any const policy object, With `std::remove_cvref_t` added in proper places, the issue is gone.
- in the second example, a function takes a policy instance by non-const reference. Assuming that is intended (i.e. the policy argument is supposed to be modified by the function) and given the analysis above, we **do** want such functions to stop working with predefined policies.

danhoeflinger: Aha, yes.  These are both very good points.  

To me, this means that 
(1) it is a spec violation
(2) it is a breaking change
(3) we may want to change the spec to resolve these issues (and modify the code accordingly)

Also, I was wrong about the member field (queue) being `const` in our implementation.  The ways you can access it publicly are `const`, but the queue is not `const`. This means those generated move and assignment operators can definitely be problematic. It would be good to "break" the possibility of these changing the queue for default policies. 

While we have shown and discussed some examples of user code "breaking" with this change, I also think that in practice these are relatively unlikely to be in user code.  Using copy / move assignment operator to modify the default policy queue seems like a clear abuse of the intentions of the spec.  Relying upon a `device_policy<KernelName>` with an template `KernelName` for overload resolution or template instantiation is odd and would be surprising to see. So while this is a breaking change, I think it will not be a disruptive one, and we should not be taking away functionality that users are likely depending on.

akukanov: Trying to answer my own questions above:

> Does this change violate the specification?

If we only do it in the implementation without changing the specification - yes, it will be a violation of the current spec which does not preclude modification of the predefined policies. However, I see it rather as a bug in the specification. The intended use of predefined policies would not be affected by this change.

> Should the specification be changed or better kept as is?

I am sure it should be changed - better to fix the bug than to pretend that this is the intended behavior.

> If the spec is changed, is it a compatible change?

Technically incompatible, but if positioned as a bug fix that enforces the original design intent and only impacts (breaks) applications which either cannot properly handle **any** const policy or they modify `dpcpp_default` (i.e. exploit the bug), I think this change is OK to do and will not cause practical problems. That also means, I do not think that we need to bump the major version for either the specification or the implementation.

I also searched GitHub for `dpcpp_default` in C++ files and studied the results. I have not found any non-typical use of the policy and I believe all these codes will work with the const policy as well.

dcbenito:
We will need to update the docs if this change happens. 

I did a quick search and found that we cite device_policy in the following pages: Execution Policies, Macros, Pass Data to Algorithms. We cite fpga_policy in the following pages: Execution Policies and Macros.

As long as I have guidance, I will update the docs as needed.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
rarutyun: In principle, I don't object to make this change. I also think that even if we have some harm it's not going to be wide. Also we can put some examples of potentially broken code in release notes. I don't know procedurally whether we should bump library version or not. My inclination is we should but I am not a decision maker here (I need to better familiarize myself with oneAPI guidelines to have stronger opinion)

Another way to go is trough deprecation, however it's definitely more complicated path. Furthermore, we don't know whether it helps anybody in practice.

To conclude, I really like that idea and I think it brings more value than potential harm.

The only thing that we need to check it that PSTL offload still works with this change. I don't see how it can break immediately and since the way how we work with `dpcpp_default` is "implementation details of PSTL offload" and it's controlled by us, there is nothing we cannot fix. I am writing this part to mostly not forget

akukanov:
@dcbenito, @SergeyKopienko - I have added the documentation changes to the pull request, please take a look.

documentation/library_guide/macros.rst
```diff
@@ -94,8 +94,8 @@ Macro                              Description
                                    such as ``dpcpp_default`` and ``dpcpp_fpga``. When the macro is not defined (by default)
                                    or evaluates to non-zero, predefined policies objects can be used.
                                    When the macro is set to 0, predefined policies objects and make functions
-                                   without arguments, when ``make_device_policy()``,
-                                   ``make_fpga_policy()``, are not available.
+                                   without arguments - ``make_device_policy()`` and
```
dcbenito, line 97: Please change the dashes - to parentheses ():  "without arguments (``make_device_policy()`` and ``make_fpga_policy()``) are not available."

dcbenito (dismissed):
I have one change, but otherwise LGTM.

dcbenito (approved):
LGTM

danhoeflinger (approved):
LGTM, as discussed, lets wait for the spec change to go through before merging.

akukanov:
I have committed the changes to the specification: https://github.com/uxlfoundation/oneAPI-spec/pull/554

PR #1674, opened 2024-07-05
Some improvements to lazy policy init

SergeyKopienko:


SergeyKopienko:
Looks like this  approach will cause ODR rule violation.
Please see details at https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-unnamed .

PR #1683, opened 2024-07-09
Limit the maximum work-group size

julianmi:
Several algorithms query the maximum supported work-group size with `__max_work_group_size`. Some devices such as the FPGA emulator might return arbitrary numbers. This can lead to issues with huge resource usage such as SLM or increased runtimes. E.g., resulting test timeouts were observed in https://github.com/oneapi-src/oneDPL/pull/1653.

This PR proposes to limit the value returned by `__max_work_group_size` to the highest correct limit of our tested production hardware architectures. 8192 is observed for `opencl:cpu`, discrete GPUs have typically 1024, and integrated GPUs 512.

This change also allows to provide a custom limit if desired by the caller. This is useful if a specific algorithm shows higher throughputs due to less used hardware resources or reduced communication costs when using smaller work-groups.

I've tagged many people here for awareness.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1278,12 +1278,6 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     // TODO: find a way to generalize getting of reliable work-group size
     std::size_t __wgroup_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
 
-#if _ONEDPL_FPGA_EMU
-    // Limit the maximum work-group size to minimize the cost of work-group reduction.
-    // Limiting this also helps to avoid huge work-group sizes on some devices (e.g., FPGU emulation).
-    __wgroup_size = std::min(__wgroup_size, (std::size_t)2048);
-#endif
-
     const auto __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
     auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size);
     __n_groups = ::std::min(__n_groups, decltype(__n_groups)(__max_cu));
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1276,7 +1276,8 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     assert(__rng_n > 0);
 
     // TODO: find a way to generalize getting of reliable work-group size
-    std::size_t __wgroup_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
+    // Limit the work-group size to prevent large sizes on CPUs. Empirically found value.
+    std::size_t __wgroup_size = oneapi::dpl::__internal::__max_work_group_size(__exec, (std::size_t)4096);
 
     const auto __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
     auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -313,6 +313,9 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         // get the work group size adjusted to the local memory limit
         // TODO: find a way to generalize getting of reliable work-group sizes
         ::std::size_t __wgroup_size = oneapi::dpl::__internal::__slm_adjusted_work_group_size(__exec, sizeof(_Type));
+        // Limit the work-group size to prevent large sizes on CPUs. Empirically found value.
+        // This value matches the current practical limit for GPUs, but may need to be re-evaluated in the future.
+        __wgroup_size = std::min(__wgroup_size, (std::size_t)1024);
 
 #if _ONEDPL_COMPILE_KERNEL
         //Actually there is one kernel_bundle for the all kernels of the pattern.
@@ -1277,6 +1280,7 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
 
     // TODO: find a way to generalize getting of reliable work-group size
     // Limit the work-group size to prevent large sizes on CPUs. Empirically found value.
+    // This value exceeds the current practical limit for GPUs, but may need to be re-evaluated in the future.
     std::size_t __wgroup_size = oneapi::dpl::__internal::__max_work_group_size(__exec, (std::size_t)4096);
 
     const auto __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -56,10 +56,14 @@ __device_info(const _ExecutionPolicy& __policy)
 #endif
 
 template <typename _ExecutionPolicy>
-::std::size_t
-__max_work_group_size(const _ExecutionPolicy& __policy)
+std::size_t
+__max_work_group_size(const _ExecutionPolicy& __policy, std::size_t __wg_size_limit = 8192)
 {
-    return __policy.queue().get_device().template get_info<sycl::info::device::max_work_group_size>();
+    std::size_t __wg_size = __policy.queue().get_device().template get_info<sycl::info::device::max_work_group_size>();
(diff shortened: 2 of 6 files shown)


SergeyKopienko:
@julianmi , about "E.g., resulting test timeouts were observed in https://github.com/oneapi-src/oneDPL/pull/1653." - compilation time, correct?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -56,10 +56,14 @@ __device_info(const _ExecutionPolicy& __policy)
 #endif
 
 template <typename _ExecutionPolicy>
-::std::size_t
-__max_work_group_size(const _ExecutionPolicy& __policy)
-{
-    return __policy.queue().get_device().template get_info<sycl::info::device::max_work_group_size>();
+std::size_t
+__max_work_group_size(const _ExecutionPolicy& __policy, std::size_t __wg_size_limit = 8192)
```
SergeyKopienko, line 60: I am not sure that it's goo idea to specify max size on caller side.
As you described on this PR, we know the correct sizes for any devices.
Also we know that real incorrect info here has place only for FPGA_EMU device.
If it's correct, why we couldn't incorporate all this logic into this function?

julianmi:
> @julianmi , about "E.g., resulting test timeouts were observed in #1653." - compilation time, correct?

Compilation times were unchanged for me. I believe the timeouts were mostly test runtime (and potentially increased JIT time) since the work-groups were huge and the resulting reduction was very slow (a couple of seconds vs. 20 minutes).

SergeyKopienko:
> > @julianmi , about "E.g., resulting test timeouts were observed in #1653." - compilation time, correct?
> 
> Compilation times were unchanged for me. I believe the timeouts were mostly test runtime (and potentially increased JIT time) since the work-groups were huge and the resulting reduction was very slow (a couple of seconds vs. 20 minutes).

Yes, I am about JIT-compilation.
But it's doesn't explain why after splitting sycl iterator tests this problem has gone.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1174,11 +1174,6 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     __wgroup_size = ::std::min(__wgroup_size, oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel));
 #endif
 
-#if _ONEDPL_FPGA_EMU
-    // Limit the maximum work-group size to minimize the cost of work-group reduction.
-    // Limiting this also helps to avoid huge work-group sizes on some devices (e.g., FPGU emulation).
-    __wgroup_size = std::min(__wgroup_size, (std::size_t)2048);
```
danhoeflinger, line 1180: Does replacing this 2048 with 8192 reintroduce any detectable slowdown for CPU / FPGA?

I ask because I believe most codes are written with a max of 1024 mind (with the priority intended target being GPU).  

It would be interesting to see how different kernels in oneDPL perform on CPU / FPGA with 1024, 2048, 4096, 8192.  If in reality we are not getting any "benefit" out of more than 1024 / 2048, I think I would prefer that we use those numbers as the max.  

My suspicion is that sycl group algorithms, and our kernels in general perform worse with these higher workgroup sizes in practice.  

This could be made into an issue if we don't want to address it here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -56,10 +56,14 @@ __device_info(const _ExecutionPolicy& __policy)
...
+{
+    std::size_t __wg_size = __policy.queue().get_device().template get_info<sycl::info::device::max_work_group_size>();
+    // Limit the maximum work-group size supported by the device to optimize the throughput or minimize communication
+    // costs. This is limited to 8192 which is the highest current limit of the tested hardware (opencl:cpu devices) to
+    // prevent huge work-group sizes returned on some devices (e.g., FPGU emulation).
+    return std::min(__wg_size, __wg_size_limit);
```
danhoeflinger, line 66: Does it make sense to look at the `__kernel_work_group_size` function as well?  It uses a somewhat arbitrary division by 4 for CPU devices.  I think a maximum cap similar to here might be a better system, and it would be more consistent.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
julianmi: There is a huge range of CPU and FPGA device, but I've collected data on 5th gen Xeon as a reference. The `find_or` pattern and radix sort, scan, reduce-by-segment, and scan-by-segment algorithms use `__max_work_group_size` and/or `__kernel_work_group_size` without lowering the work-group size. Runs with 1024, 2048, 4096, 8192 didn't show a general pattern. Instead, the optimal work-group size is dependent on the specific algorithm. I think our assumption about communication-itensive kernels such as scan benefit from smaller work-group sizes holds though. However, radix sort seems to favor larger sizes. Thus, I've added specific limits based on the empirical data.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: Good point. I've removed this arbitrary value and used specific limits as discussed above.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Very interesting. Thanks for taking a look at this. 
I'm not sure how to feel about these magic values since like you mention, there is a large search space of device.  However, I don't really know a better way.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: Thanks again for looking at this.  

I wonder if we should continue to restrict this cap to CPU devices (for the caps which are replacing the CPU specific division by 4).  In practice, with current hardware and drivers, I believe 1024 is our GPU cap anyway, but we have no guarantees this does not extend further to 2048 and beyond, and we would not necessarily remember to update these values accordingly.  Presumably, if those sizes are available on the GPU, they will be advantageous to use.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
julianmi: I agree. I think one way to help with these tuning parameters would be to explicitly mark them and provide tooling to reevaluate them based on hardware changes. This would also help improve with some of the tuning work such as https://github.com/oneapi-src/oneDPL/pull/1054.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: I agree that this cap might change with future GPUs, but wouldn't we need to tune for their changed capabilities anyways? It looks to me like these issues might be better addressed by architecture-specific tuning parameters that are stored somewhere centrally and updated for new architectures.

danhoeflinger: Yes, a central location or system for these empirical values and tools to re-evaluate them within some range and options to have specialized values for different hardware would be ideal.  My guess is we don't want to try to solve that in this PR.  

Perhaps for now we can tag these with some specific keyword in the comments and create a github issue which describes how to find these empirical values in the code to later re-evaluate within something like #1054.

The comments you've added describing that these have been empirically determined for CPU are good.  If we don't want to limit with a branch specific to CPU devices, perhaps we can add to that comment saying something like:

// This value exceeds the current practical limit for GPUs, but may need to be re-evaluated in the future

julianmi: I agree that this would be another PR.

I've added comments about re-evaluating future GPUs as you suggested. I prefer this solution over a CPU-specific cap since the CPU and FPGA emulator are currently the only devices using larger work-group sizes making this cap generally applicable.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: Additional to the correctness issues you mentioned, the caller might know a better limit than the max. This is used throughout our code with `min` comparisons, which can be eliminated with this change.

julianmi:
@danhoeflinger @SergeyKopienko I've rebased this patch. I think the main open is how to handle the magic numbers limiting the WG sizes. This will require continued discussions and a design and implementation of a tuning architecture. I think this is out of scope for this patch and we should go ahead with these changes in the meantime. Do you have further comments or can this be merged?

danhoeflinger (approved):
LGTM, 

I agree that we are not going to fix the magic number question in this PR.  I think this is a step in the right direction, and should be good to merge.

SergeyKopienko (approved):
LGTM

PR #1689, opened 2024-07-10
Improvements of early exit for find-based algorithms (when we search for any matching data entry)

SergeyKopienko:
In this PR, we implemented the early exit from search inside one sub-group in all sub-group items for find-based algorithms
(when we search for any matching data entry).

It's give us some significant performance boost.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1138,7 +1138,7 @@ struct __early_exit_find_or
             }
 
             // Share found into state between items in our sub-group to early exit if something was found
-            // - this approach is applicable only for __parallel_or_tag (hen we search for any matching data entry)
+            // - this approach is applicable only for __parallel_or_tag (when we search for any matching data entry)
             // - for __parallel_find_forward_tag and __parallel_find_backward_tag we should process all data
             if constexpr (_OrTagType{})
             {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1090,7 +1090,8 @@ struct __early_exit_find_or
               typename... _Ranges>
     void
     operator()(const _NDItemId __item_id, const _IterSize __n_iter, const _WgSize __wg_size,
-               _LocalFoundState& __found_local, _BrickTag __brick_tag, _Ranges&&... __rngs) const
+               const _LocalFoundState __init_value, _LocalFoundState& __found_local, _BrickTag __brick_tag,
+               _Ranges&&... __rngs) const
     {
         // There are 3 possible tag types here:
         //  - __parallel_find_forward_tag : in case when we find the first value in the data;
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1138,6 +1137,16 @@ struct __early_exit_find_or
...
             }
+
+            // Share found into state between items in our sub-group to early exit if something was found
+            // - this approach is applicable only for __parallel_or_tag (hen we search for any matching data entry)
+            // - for __parallel_find_forward_tag and __parallel_find_backward_tag we should process all data
+            if constexpr (_OrTagType{})
+            {
+                __found_local = __dpl_sycl::__any_of_group(__item_id.get_sub_group(), __found_local);
+                if (__found_local != __init_value)
+                    __something_was_found = true;
```
danhoeflinger, line 1148: Can we just run `__any_of_group` directly on `__something_was_found` and avoid the branching?

julianmi: This would also remove the requirement for passing `__init_value` introduced here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1138,6 +1136,16 @@ struct __early_exit_find_or
                 // So we use bool variable state check in the for-loop header.
                 __something_was_found = true;
             }
+
+            // Share found into state between items in our sub-group to early exit if something was found
+            // - this approach is applicable only for __parallel_or_tag (hen we search for any matching data entry)
+            // - for __parallel_find_forward_tag and __parallel_find_backward_tag we should process all data
+            if constexpr (_OrTagType{})
```
danhoeflinger, line 1143: I believe we should be able to use this subgroup coordinated early exit for all cases, not just `_OrTagType`.  However, I think we are blocked currently by our access pattern.  Right now, each work-item iterates in the direction of the search, (front-to-back or back-to-front) so it is safe to early exit individual work-items. 

This synchronization makes each subgroup work in lock step, coordinating the results, and if it were loading contiguous blocks of the input data in the direction of the search, we could do this coordinated subgroup early exit safely as well for the other search types.  Right now, it is **not** loading elements in contiguous blocks of the subgroup size, but rather contiguous blocks of size 16, and striding each work-item by 16 for each iteration.   

For a subgroup of size 32, we would have a subgroup loading half its data from one contiguous section, and another half from a later contiguous section.  This means that we cannot use coordinated subgroup early-exit, because a different work item may find a "earlier" element on a different iteration.

The fix for this is to re-write the access pattern to load contiguous blocks based on subgroup size rather than hardcoded to 16.  Then we should be able to remove this constexpr branch and benefit from this optimization for all find algorithms.  This also may make loads faster because they should be contiguous blocks of data rather than gathering from two spots.

SergeyKopienko: I propose to fix this in the separate PR

danhoeflinger (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1138,6 +1136,16 @@ struct __early_exit_find_or
                 // So we use bool variable state check in the for-loop header.
                 __something_was_found = true;
             }
+
+            // Share found into state between items in our sub-group to early exit if something was found
+            // - this approach is applicable only for __parallel_or_tag (hen we search for any matching data entry)
```
julianmi, line 1141:
```suggestion
            // - this approach is applicable only for __parallel_or_tag (when we search for any matching data entry)
```

julianmi (approved):

danhoeflinger (approved):

PR #1690, opened 2024-07-10
[PROTOTYPE] Fix check for __no_init_value when __no_init_value is used with a non-void value type

mmichel11:
Our current check for if there is an init value assumes that `__no_init_value` will always provide a `__value_type` of `void`. As I've been integrating this code into the oneDPL SYCL backend, it seems that this is not always the case and the `__value_type` may be the actual input iterator's value type which leads to the wrong path being taken and compile time errors.

I have updated the check to determine if `InitType` is an instance of `__no_init_value`.


danhoeflinger (approved):
LGTM

PR #1692, opened 2024-07-11
Add stable_sort_by_key

dmitriy-sobolev:
Add the `stable_sort_by_key`, including the API, tests, and documentation.
This new API will provide flexibility while selecting algorithms for the best performance. Specifically, `sort_by_key` may divert execution into a non-stable sort implementation in the future.

Currently, it uses the same implementation as `sort_by_key` for device policies and new one for standard policies. The standard ones already use a non-stable algorithm, so new, stable, pattern has been introduced.

The test has been extended/reworked:
 - Stability is verified.
 - Data is generated randomly instead of being based on a simple pattern.
 - Data type coverages is extended from `int` to many arithmetic types and a non-arithmetic type.
 - Default comparator is used instead of  `std::less` to check the an overload without a comparator.
 - `std::greater` is replaced with a custom comparator (which does the same), to verify comparison-based implementations, like merge-sort.


test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,177 @@
...
+        if (i < (N - 1))
+        {
+            EXPECT_TRUE(h_skey[i] >= h_skey[i + 1], "wrong sort result with hetero policy, USM data");
+        }
+    }
+}
+
+template <typename StabilityTag>
+void
+test_with_buffers(sycl::queue& q, StabilityTag stability_tag)
```
SergeyKopienko, line 129: Early this function was written under `#if TEST_DPCPP_BACKEND_PRESENT`
Why this define is not used now?

dmitriy-sobolev: It is used to guard  `test_with_buffers` and `test_usm` as before. I additionally guarded `test_all_policies` by mistake, but is fixed now.

documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -87,6 +87,22 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
...
+  a comparioson operator. If no comparator is provided, then the elements are compared with ``operator<``.
+  The sequence's values are permutated according to the sorted sequence's keys. The prerequisite for correct
+  behavior is that the size for both keys sequence and values sequence shall be the same.  
+  For example::
+
+    keys:   [3,    5,   0,   4,   3,   0]
+    values: ['a', 'b', 'c', 'd', 'e', 'f']
+    output_keys:   [0,    0,   3,   3,   4,   5]
+    output_values: ['c', 'f', 'a', 'e', 'd', 'b']
+
```
akukanov, line 105: In order to  not duplicate the same information, would not it be better to just change `sort_by_key` to `stable_sort_by_key` in the current wording, and add a short section that describes `sort_by_key` as equivalent to the stable one and adds the caution note?

include/oneapi/dpl/pstl/glue_algorithm_impl.h
```diff
@@ -719,6 +719,29 @@ sort_by_key(_ExecutionPolicy&& __exec, _RandomAccessIterator1 __keys_first, _Ran
...
+
+template <typename _ExecutionPolicy, typename _RandomAccessIterator1, typename _RandomAccessIterator2>
+oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy>
+stable_sort_by_key(_ExecutionPolicy&& __exec, _RandomAccessIterator1 __keys_first, _RandomAccessIterator1 __keys_last,
+                   _RandomAccessIterator2 __values_first)
+{
+    oneapi::dpl::stable_sort_by_key(::std::forward<_ExecutionPolicy>(__exec), __keys_first, __keys_last, __values_first,
+                                    oneapi::dpl::__internal::__pstl_less());
+}
+
```
akukanov, line 744: Same here: why not to rename the current implementation and then define `sort_by_key` as a redirect?

documentation/library_guide/parallel_api/additional_algorithms.rst
dmitriy-sobolev: That's a good idea. It's done.

include/oneapi/dpl/pstl/glue_algorithm_impl.h
dmitriy-sobolev: `sort_by_key` now redirects to `stable_sort_by_key`.

documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -76,7 +76,7 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
     search sequence: [0, 2, 4, 7, 6]
     result sequence: [1, 4, 8, 10, 10]
 
-* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts the sequence's keys according to 
+* ``stable_sort_by_key``: performs a stable key-value sort. The algorithm sorts the sequence's keys according to 
   a comparioson operator. If no comparator is provided, then the elements are compared with ``operator<``.
   The sequence's values are permutated according to the sorted sequence's keys. The prerequisite for correct
   behavior is that the size for both keys sequence and values sequence shall be the same.
```
akukanov, line 82:
```suggestion
* ``stable_sort_by_key``: performs a stable key-value sort. The algorithm sorts a sequence of keys according to 
  a comparison operator. If no comparator is provided, then the elements are compared with ``operator<``.
  A sequence of values is simultaneously permuted according to the sorted order of keys. For correct behavior
  both sequences shall be of the same size.  

```

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,182 @@
...
+    // 2. Sorting
+    if constexpr (std::is_same_v<StabilityTag, StableSort>)
+        oneapi::dpl::stable_sort_by_key(policy, keys_begin, keys_begin + n, vals_begin);
+    else
+        oneapi::dpl::sort_by_key(policy, keys_begin, keys_begin + n, vals_begin);
+}
+
+template<typename _Keys, typename _Vals, typename _Size>
+void
+check_sort_by_key_results(const _Keys& keys_buf, const _Vals& vals_buf,  _Size n)
```
akukanov, line 52: There do not seem to be any checks for the stability of sorting; I think it is needed.

dmitriy-sobolev: Agree. I've added the check. It's revealed that the host-side implementation is not stable:

https://github.com/oneapi-src/oneDPL/blob/e81af7bc5ee7a40ef28040b9d6aa405bbdbf1223/include/oneapi/dpl/pstl/algorithm_impl.h#L2478-L2499

Tagging @danhoeflinger for awareness.

The documentation says that `sort_by_key` is stable. But host-side policies implement non-stable one. I see 2 ways to handle it:

1. Treat it as an implementation bug. Fix it.
2. Treat it as a documentation bug for host-side policies. Fix the documentation and introduce a stable host-side implementation.

I see the 1st approach as a better way to proceed, because some users may already have `sort_by_key` assuming stability. @akukanov, what is your opinion on that?

dmitriy-sobolev: I've partially followed the 2nd approach. `sort_by_key` does not guarantee the stability now, but there is a exception: device execution policies.

documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -76,17 +76,26 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
     search sequence: [0, 2, 4, 7, 6]
     result sequence: [1, 4, 8, 10, 10]
 
-* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts the sequence's keys according to 
-  a comparioson operator. If no comparator is provided, then the elements are compared with ``operator<``.
-  The sequence's values are permutated according to the sorted sequence's keys. The prerequisite for correct
-  behavior is that the size for both keys sequence and values sequence shall be the same.  
+* ``sort_by_key``: performs a key-value sort. The algorithm sorts a sequence of keys according to
+  a comparison operator. If no comparator is provided, then the elements are compared with ``operator<``.
```
MikeDvorskiy, line 80: Probably, should be `std::less` ? 
oneDPL impl uses its own predicate with "operator<", right... But, probably, should we change it as well?

include/oneapi/dpl/pstl/algorithm_fwd.h
```diff
@@ -778,6 +778,22 @@ void
 __pattern_sort_by_key(__parallel_tag<_IsVector>, _ExecutionPolicy&&, _RandomAccessIterator1, _RandomAccessIterator1,
                       _RandomAccessIterator2, _Compare);
 
+//------------------------------------------------------------------------
+// stable_sort_by_key
+//------------------------------------------------------------------------
+
+template <class _Tag, typename _ExecutionPolicy, typename _RandomAccessIterator1, typename _RandomAccessIterator2,
```
MikeDvorskiy, line 785: For consistency we write `typename `(instead of class), where it is appicable.

test/parallel_api/algorithm/alg.sorting/stable_sort_by_key.pass.cpp
```diff
@@ -0,0 +1,26 @@
...
+#include "support/test_config.h"
+#include "support/utils.h"
+
+#include "sort_by_key_common.h"
+
+int
+main()
+{
+    test_all_policies(StableSort{});
+    return TestUtils::done();
```
MikeDvorskiy, line 25: if we skip some tests in some cases we should return an error/skip code, for example
return TestUtils::done(_ENABLE_RANGES_TESTING);

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,250 @@
...
+        if constexpr (std::is_same_v<StabilityTag, StableSort>)
+            oneapi::dpl::stable_sort_by_key(policy, keys_begin, keys_begin + n, vals_begin, greater);
+        else
+            oneapi::dpl::sort_by_key(policy, keys_begin, keys_begin + n, vals_begin, greater);
+    }
+}
+
+template<typename _KeysIt, typename _ValsIt, typename _Size, typename StabilityTag, typename DirectionTag>
+void
+check_sort(const _KeysIt& keys_begin, const _ValsIt& vals_begin, _Size n, StabilityTag, DirectionTag)
```
MikeDvorskiy, line 102: A checker depends on a data generator, right?
Why whether do we use `std::is_sorted`? or/and std::sort with tuple type and check via EXPECT_EQ_N ?

test/parallel_api/algorithm/alg.sorting/stable_sort_by_key.pass.cpp
dmitriy-sobolev: It makes sense doing if there is a combination of macros when the test is skipped entirely. There is no such combination for this test.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,250 @@
...
+    call_sort(policy, keys_buf.begin(), vals_buf.begin(), n, stability_tag, direction_tag);
+    check_sort(keys_buf.begin(), vals_buf.begin(), n, stability_tag, direction_tag);
+}
+
+#if TEST_DPCPP_BACKEND_PRESENT
+template <sycl::usm::alloc alloc_type, typename StabilityTag, typename DirectionTag>
+void
+test_with_usm(sycl::queue& q, StabilityTag stability_tag, DirectionTag direction_tag)
+{
+    constexpr int n = (1 << 12) + 42;
```
MikeDvorskiy, line 158: Such magic formulas are difficult to understand....
I see that (1 << 12) :  12th degree of 2.... but why 42? 
Overall, such formulas should be explained in a comment.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,250 @@
...
+        if constexpr (std::is_same_v<StabilityTag, StableSort>)
+            oneapi::dpl::stable_sort_by_key(policy, keys_begin, keys_begin + n, vals_begin);
+        else
+            oneapi::dpl::sort_by_key(policy, keys_begin, keys_begin + n, vals_begin);
+    }
+    else
+    {
+        // Pass a custom comparator to check the corresponding implementation; merge-sort may be used
+        auto greater = [](const auto& lhs, const auto& rhs) { return lhs > rhs; };
+        if constexpr (std::is_same_v<StabilityTag, StableSort>)
```
MikeDvorskiy, line 93: Instead of a special Tag usage I would pass (into common test function) a functional object which represents stable or unstable sort.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,250 @@
...
+}
+
+template<typename _Policy, typename _KeyIt, typename _ValIt, typename _Size, typename StabilityTag, typename DirectionTag>
+void
+call_sort(_Policy&& policy, _KeyIt keys_begin, _ValIt vals_begin, _Size n, StabilityTag, DirectionTag)
+{
+    if constexpr (std::is_same_v<DirectionTag, AscendingSort>)
+    {
+        // Do not pass the comparator to check the API with a default comparator; radix sort may be used
+        if constexpr (std::is_same_v<StabilityTag, StableSort>)
```
MikeDvorskiy, line 84: Instead of embedded `if constexpr`  I would use a functional object (with necessary parameters) which represents stable or unstable sort.

documentation/library_guide/parallel_api/additional_algorithms.rst
dmitriy-sobolev: `std::sort` uses `operator<` until c++20, and `std::less` since c++20.
I suppose it does not result in backward compatibility issues since `std::less` is defined as:

> Function object for performing comparisons. The main template invokes operator< on type T.

Namely, `std::less` is a superset including `operator<`.

However, it is a change in the function signature, and it may break the ABI for users. Please, correct me if I am mistaken here.

dmitriy-sobolev: Ah, just ignore the note about ABI. There is no change in the signature.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -2498,6 +2498,54 @@ __pattern_sort_by_key(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _Ran
...
+    ::std::stable_sort(__beg, __end, __cmp_f);
+}
+
+template <typename _IsVector, typename _ExecutionPolicy, typename _RandomAccessIterator1,
+          typename _RandomAccessIterator2, typename _Compare>
+void
+__pattern_stable_sort_by_key(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomAccessIterator1 __keys_first,
+                             _RandomAccessIterator1 __keys_last, _RandomAccessIterator2 __values_first, _Compare __comp)
+{
+    static_assert(
```
MikeDvorskiy, line 2529: I would suggest omitting such type requirement checks with startic_assert, due to
1) a user has got an error directly in a code place where such requirement needs.
2) concept and requires are upcoming... 

I know that such checks were before this PR... But, it was for ensuring that we didn't miss a some special case after "move_constructible" had been removed from algo signature.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
danhoeflinger: I think its just a number that is non-zero so we aren't just testing powers of 2 or powers of 10.

42 is commonly used because of Hitchhikers guide:
[The Answer to the Ultimate Question of Life, the Universe, and Everything is 42](https://en.wikipedia.org/wiki/Phrases_from_The_Hitchhiker%27s_Guide_to_the_Galaxy)

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,250 @@
...
+
+template <typename StabilityTag>
+void
+test_all_policies(StabilityTag stability_tag)
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+    test_device_polcies(stability_tag);
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+#if !TEST_DPCPP_BACKEND_PRESENT
```
MikeDvorskiy, line 244: It looks as wrong test logic: if we have` dpcpp backend`, for example, we skip the tests with std policies....

dmitriy-sobolev: Now the standard policies are tested irrespective of device backend presence.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Thanks, done

include/oneapi/dpl/pstl/algorithm_impl.h
dmitriy-sobolev: Done

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Tag simplifies dispatching, like as it is done for `check_sort` function. I do not see an easy way to do the same with a functional object.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -2498,6 +2498,54 @@ __pattern_sort_by_key(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _Ran
...
+{
+    static_assert(__is_serial_tag_v<_Tag> || __is_parallel_forward_tag_v<_Tag>);
+
+    auto __beg = oneapi::dpl::make_zip_iterator(__keys_first, __values_first);
+    auto __end = __beg + (__keys_last - __keys_first);
+    auto __cmp_f = [__comp](const auto& __a, const auto& __b) {
+        return __comp(::std::get<0>(__a), ::std::get<0>(__b));
+    };
+
+    ::std::stable_sort(__beg, __end, __cmp_f);
```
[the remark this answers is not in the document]
dmitriy-sobolev: Removed

include/oneapi/dpl/pstl/algorithm_fwd.h
dmitriy-sobolev: Fixed

documentation/library_guide/parallel_api/additional_algorithms.rst
dmitriy-sobolev: There are plenty of algorithms, which should be changed accordingly. Would you mind if it is done separately?

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I've replaced this formula with a constant. You can find it as `small_size` variable, which accompanies `large_size`. They contain some "random" values, which are non-zero, non-power of 2 or 10 as described by Dan. Should it still be explained in a comment?

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I've made a more genetic checkers and generator, so they are now decoupled. `is_sorted` is not used now either.
Has it resolved your question? Frankly, I do not understand all of its details.

documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -76,17 +76,26 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
...
     result sequence: [1, 4, 8, 10, 10]
 
-* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts the sequence's keys according to 
-  a comparioson operator. If no comparator is provided, then the elements are compared with ``operator<``.
-  The sequence's values are permutated according to the sorted sequence's keys. The prerequisite for correct
-  behavior is that the size for both keys sequence and values sequence shall be the same.  
+* ``sort_by_key``: performs a key-value sort. The algorithm sorts a sequence of keys according to
+  a comparison operator. If no comparator is provided, then the elements are compared with ``operator<``.
+  A sequence of values is simultaneously permuted according to the sorted order of keys. For correct behavior
+  both sequences shall be of the same size.
```
MikeDvorskiy, line 82: As far as we are changing that sentence... Probably, it makes sense to change `shall be` => `muts be` ? We use `muts be` everywhere in the doc.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,286 @@
...
+    float velocity;
+    float coordinates[3];
+    using energy_type = float;
+    energy_type energy() const
+    {
+        return 0.5 * mass * velocity * velocity;
+    }
+    bool operator==(const Particle& other) const
+    {
+        return std::tie(coordinates[0], coordinates[1], coordinates[2]) ==
```
MikeDvorskiy, line 61: A bitwise comparison of float numbers might be not correct...

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,286 @@
...
+               std::tie(other.coordinates[0], other.coordinates[1], other.coordinates[2]);
+    }
+};
+
+constexpr std::size_t large_size = 81207;
```
MikeDvorskiy, line 66: Why do `small `and `large` have such constant? I think should be commented.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: It is guaranteed to be exact here, because the values are reordered and not transformed. Should I change the logic to apply some epsilon?

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Done

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I've additionally explained it.

documentation/library_guide/parallel_api/additional_algorithms.rst
dmitriy-sobolev: I like `must be` more. Done.

test/support/utils.h
```diff
@@ -930,6 +932,41 @@ get_new_kernel_params(KernelParams)
 }
 #endif //TEST_DPCPP_BACKEND_PRESENT
 
+template <typename T>
+typename ::std::enable_if_t<std::is_arithmetic_v<T>, void>
```
SergeyKopienko, line 936:
```suggestion
typename ::std::enable_if_t<std::is_arithmetic_v<T>>
```

according to https://en.cppreference.com/w/cpp/types/enable_if

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,289 @@
...
+    test_with_buffers<Particle::energy_type, Particle, 4>(q, large_size, GreaterComparator{}, stability_tag);
+    test_with_buffers<Particle::energy_type, Particle, 5>(q, small_size, DefaultComparator{}, stability_tag);
+}
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+template <typename KeyT, typename ValT, typename Size, typename Compare, typename StabilityTag>
+void
+test_std_polcies(Size n, Compare compare, StabilityTag stability_tag)
+{
+    test_with_std_policy<KeyT, ValT>(oneapi::dpl::execution::seq, n, compare, stability_tag);
```
SergeyKopienko, line 271: May we try to use `invoke_on_all_host_policies` instead of this function?
The `invoke_on_all_host_policies` implemented more more difficult then this function.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,289 @@
...
+        sycl::buffer<ValT> vals_device(vals.data(), n);
+        auto policy = TestUtils::make_device_policy<TestUtils::unique_kernel_name<class Buffer, KernelNameID>>(q);
+        call_sort(policy, oneapi::dpl::begin(keys_device), oneapi::dpl::begin(vals_device), n, compare, stability_tag);
+    }
+   // sort_by_key with device policy guarantees stability, hence StableSort{} is passed
+    check_sort(keys.begin(), vals.begin(), origin_keys.begin(), origin_vals.begin(), n, compare, StableSort{});
+}
+#endif // TEST_DPCPP_BACKEND_PRESENT
+
+#if TEST_DPCPP_BACKEND_PRESENT
```
SergeyKopienko, line 253:
```suggestion

```
Looks like these two lines simple not required.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1283,6 +1287,23 @@ __pattern_sort_by_key(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec
...
+//------------------------------------------------------------------------
+// stable_sort_by_key
+//------------------------------------------------------------------------
+
+template <typename _BackendTag, typename _ExecutionPolicy, typename _Iterator1, typename _Iterator2, typename _Compare>
+void
+__pattern_sort_by_key(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _Iterator1 __keys_first,
+                      _Iterator1 __keys_last, _Iterator2 __values_first, _Compare __comp)
+{
+    __pattern_stable_sort_by_key(__tag, ::std::forward<_ExecutionPolicy>(__exec), __keys_first, __keys_last,
```
MikeDvorskiy, line 1299: `::std::` -> `std::`

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
MikeDvorskiy: I see... in that case, I think, you should not.

test/support/utils.h
```diff
@@ -930,6 +932,41 @@ get_new_kernel_params(KernelParams)
...
+generate_arithmetic_data(T* input, std::size_t size, std::uint32_t seed)
+{
+    std::default_random_engine gen{seed};
+    std::size_t unique_threshold = 75 * size / 100;
```
MikeDvorskiy, line 940: I would add a comment why we use so "magic" constants (75 and 100)

test/support/utils.h
```diff
@@ -930,6 +932,41 @@ get_new_kernel_params(KernelParams)
...
+    {
+        std::uniform_int_distribution<int> dist(std::numeric_limits<T>::lowest(), std::numeric_limits<T>::max());
+        std::generate(input, input + unique_threshold, [&] { return T(dist(gen)); });
+    }
+    else if constexpr (std::is_integral_v<T>)
+    {
+        std::uniform_int_distribution<T> dist(std::numeric_limits<T>::lowest(), std::numeric_limits<T>::max());
+        std::generate(input, input + unique_threshold, [&] { return dist(gen); });
+    }
+    else
```
MikeDvorskiy, line 951: I would suggest joining these two cases (941-950) into one, like this:

```
template <typename T>
using GT = std::conditional_t<sizeof(T) < sizeof(short), int, T>;
std::uniform_int_distribution<GT> ....
```

test/support/utils.h
```diff
@@ -930,6 +932,41 @@ get_new_kernel_params(KernelParams)
...
+        std::uniform_real_distribution<T> dist_real(std::numeric_limits<T>::min(), log2(std::numeric_limits<T>::max()));
+        std::uniform_int_distribution<int> dist_binary(0, 1);
+        auto randomly_signed_real = [&dist_real, &dist_binary, &gen]()
+        {
+            auto v = exp2(dist_real(gen));
+            return dist_binary(gen) == 0 ? v : -v;
+        };
+        std::generate(input, input + unique_threshold, [&] { return randomly_signed_real(); });
+    }
+    for (uint32_t i = 0, j = unique_threshold; j < size; ++i, ++j)
```
MikeDvorskiy, line 965: It makes sense to add an assert before the loop, like:
`assert(unique_threshold >= size/2 && unique_threshold < size);`

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,289 @@
...
+{
+    if constexpr (std::is_same_v<Compare, DefaultComparator>)
+        oneapi::dpl::stable_sort_by_key(policy, keys_begin, keys_begin + n, vals_begin);
+    else
+        oneapi::dpl::stable_sort_by_key(policy, keys_begin, keys_begin + n, vals_begin, compare);
+}
+
+template<typename Policy, typename KeyIt, typename ValIt, typename Size, typename Compare>
+void
+call_sort(Policy&& policy, KeyIt keys_begin, ValIt vals_begin, Size n, Compare compare, UnstableSort)
```
MikeDvorskiy, line 124: Actually, I dont understand why we need that function... I would call the sort algo directly
`oneapi::dpl::sort_by_key(policy, keys_begin, keys_begin + n, vals_begin, compare);`

or if we would like to various stable/unstable yo can define a general caller like that:

auto call_sort = [](auto f, auto&&... args) { f(arg...);}

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,289 @@
...
+        oneapi::dpl::sort_by_key(policy, keys_begin, keys_begin + n, vals_begin);
+    else
+        oneapi::dpl::sort_by_key(policy, keys_begin, keys_begin + n, vals_begin, compare);
+}
+
+template<typename KeyIt, typename ValIt, typename Size, typename Compare>
+void
+call_reference_sort(KeyIt ref_keys_begin, ValIt ref_vals_begin, Size n, Compare)
+{
+    static_assert(std::is_same_v<Compare, DefaultComparator> || std::is_same_v<Compare, GreaterComparator>);
```
MikeDvorskiy, line 136: why only default  and great?
Why not general Compare?

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,289 @@
...
+    auto first = oneapi::dpl::make_zip_iterator(ref_keys_begin, ref_vals_begin);
+    if constexpr (std::is_same_v<Compare, DefaultComparator>)
```
MikeDvorskiy, line 138: I dont understand why we need 138-149 logic.
lets call sort for testing with Comparator type which is passed by a test case.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,289 @@
...
+
+struct Particle
+{
+    float mass;
+    float velocity;
+    float coordinates[3];
+    using energy_type = float;
+    energy_type energy() const
+    {
+        return 0.5 * mass * velocity * velocity;
```
MikeDvorskiy, line 57: I guess, a comment "kinetic energy" should be helpfull here.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,289 @@
...
+#include <vector> // std::vector
+#include <iterator> // std::distance
+#include <algorithm> // std::generate, std::remove_if, std::stable_sort, std::is_sorted, std::transform, std::shuffle
+#include <type_traits> // std::is_same
+#include <unordered_map> // std::unordered_map
+
+struct StableSort{};
+struct UnstableSort{};
+
+struct DefaultComparator{};
```
MikeDvorskiy, line 40: I think `DefaultComparator` confuses here... because "comparator" must be  callable with two arguments a return bool type.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,289 @@
...
+{
+    template<typename T>
+    bool operator()(const T& lhs, const T& rhs) const {
+        return lhs > rhs;
+    }
+};
+
+struct Particle
+{
+    float mass;
```
MikeDvorskiy, line 51: lines: 51 - 53: uninitialized fields

documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -76,19 +76,28 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
...
     result sequence: [1, 4, 8, 10, 10]
 
-* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts the sequence's keys according to 
-  a comparioson operator. If no comparator is provided, then the elements are compared with ``operator<``.
-  The sequence's values are permutated according to the sorted sequence's keys. The prerequisite for correct
-  behavior is that the size for both keys sequence and values sequence shall be the same.  
+* ``sort_by_key``: performs a key-value sort. The algorithm sorts a sequence of keys according to
+  a comparison operator. If no comparator is provided, then the elements are compared with ``operator<``.
+  A sequence of values is simultaneously permuted according to the sorted order of keys. For correct behavior,
+  both sequences must be of the same size.
```
MikeDvorskiy, line 82: The given iterator version of `sort_by_key `has 2 iterators for definition of a key sequence and one iterator for definition of a value sequence. 
So, I think the sentence "both sequences must be of the same size" is not quite correct.
For example, we can observe `thrust::sort_by_key ` documentation: 
https://nvidia.github.io/cccl/thrust/api/function_group__sorting_1ga667333ee2e067bb7da3fb1b8ab6d348c.html

test/support/utils.h
dmitriy-sobolev: I've added a comment

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Removed

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Making a test aligned with the core tests is a good idea, but I suggest doing it separately since it may complicate the PR.
Current PR adds new functionality and extends the test accordingly. I suppose it should be enough for now.

test/support/utils.h
dmitriy-sobolev: Done

test/support/utils.h
dmitriy-sobolev: Thanks, done.

test/support/utils.h
dmitriy-sobolev: Done

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Would you mind if I leave it as is?

It is not obvious to me how using a lambda can further simplify the code. There are 2 `call_sort` overloads, essentially 6-lines each. They are called from 3 functions. It is already pretty minimalistic and simple.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Note that a general comparator was not tested originally. This PR extends the testing as it is described, well, in the PR's description.

Do you think that adding a custom comparator is still necessary? It requires some kind of mapping between the comparators for the tested algorithm (`dpl::sort_by_key`) and the checker (`std::sort`).  Probably, it would be better to pass both comparators: one for `dpl::sort_by_key` and another for `std::sort` from `test` functions instead of a single comparator-tag.  

Would you prefer me implementing the idea above, or leaving it as is? I am inclined toward leaving it as is for now.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: `dpl::sort_by_key` is checked by `std::stable_sort`. Their comparators have different semantics.

For example: 
`std::stable_sort(..., [](const auto& lhs, const auto& rhs) { return std::get<0>(lhs) > std::get<0>(rhs)})` checks 
`dpl::sort_by_key(..., [](const T& lhs, const T& rhs) { return lhs > rhs; })`.

That is why the same comparator cannot be used.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Added

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I guess I should use something like `NoComparatorTag` instead. It is not a comparator, but a tag-like entity. I've renamed it.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Note that are initialized in `generate_data`.

documentation/library_guide/parallel_api/additional_algorithms.rst
dmitriy-sobolev: That's a good catch. Let me think about the wording. 
The documentation of the additional algorithms does not provide the signature of the functions (which is a huge gap on my opinion), what complicates giving a "crisp" description.

dmitriy-sobolev: I've corrected the documentation and improved the tests accordingly.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
MikeDvorskiy: Ok, i don't insist

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
MikeDvorskiy: > a general comparator was not tested originally

I means the test was written quite right originally...

As example, I keep in my mind the following simple test workflow:
```
test_sort(sort_algo, compare, auto&&... data)
{
      sort_algo(compare, data...);
      check_result(compare, data...)
}
//.....
//test case 1:
test_sort(sort_stable, std::less{}, values);
//test case 2:
test_sort(sort_unstable, my_comparator, values);
//test case 3:
test_sort(sort_stable_by_key, std::greater{}, values, keys);
//etc
```

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
MikeDvorskiy: I did not mean that the fields are non-intialized at all. I meant that having uninitialized class members  is a "rule of good manners" for C++ code. And it prevents UB.

documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -76,10 +76,11 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
...
 
-* ``sort_by_key``: performs a key-value sort. The algorithm sorts a sequence of keys according to
-  a comparison operator. If no comparator is provided, then the elements are compared with ``operator<``.
-  A sequence of values is simultaneously permuted according to the sorted order of keys. For correct behavior,
-  both sequences must be of the same size.
+* ``sort_by_key``: performs a key-value sort.
+  The algorithm sorts a sequence of keys based on a comparison function object.
+  If it is not provided, the elements are compared with ``operator<``.
+  A sequence of values is simultaneously permuted according to the sorted order of keys.
+  For correct behavior, a sequence of values must have at least as many elements as the sequence of keys.
```
MikeDvorskiy, line 83: I suggest following sentence:
`A sequence of values must have at least as many elements as the sequence of keys, otherwise the behavior is undefined.`

`the behavior is undefined` - "standard" wording on cppreference, f.e., more over we also use one in other parts of the documentation.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,297 @@
...
+{
+    if constexpr (std::is_same_v<Compare, NoComparatorTag>)
+        oneapi::dpl::sort_by_key(policy, keys_begin, keys_begin + n, vals_begin);
+    else
+        oneapi::dpl::sort_by_key(policy, keys_begin, keys_begin + n, vals_begin, compare);
+}
+
+template<typename KeyIt, typename ValIt, typename Size, typename Compare>
+void
+call_reference_sort(KeyIt ref_keys_begin, ValIt ref_vals_begin, Size n, Compare)
```
MikeDvorskiy, line 135: M.b. like that?

```
template<typename KeyIt, typename ValIt, typename Size, typename Compare = std::less<>>
void
call_reference_sort(KeyIt ref_keys_begin, ValIt ref_vals_begin, Size n, Compare comp = {})
{
    auto first = oneapi::dpl::make_zip_iterator(ref_keys_begin, ref_vals_begin);
    std::stable_sort(first, first + n, [](const auto& lhs, const auto& rhs) { return comp(std::get<0>(lhs), std::get<0>(rhs)); });
}

///....
call_reference_sort(....); //no comparator
call_reference_sort(...., GreaterComparatorTag{}); //with comparator
```

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
MikeDvorskiy: `comp(std::get<0>(lhs), std::get<0>(rhs))`  - the same comparator

documentation/library_guide/parallel_api/additional_algorithms.rst
SergeyKopienko: May be some `assert` required for this check or `exception` ?

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,297 @@
...
+#include <iterator> // std::distance
+#include <algorithm> // std::generate, std::remove_if, std::stable_sort, std::is_sorted, std::transform, std::shuffle
+#include <type_traits> // std::is_same
+#include <unordered_map> // std::unordered_map
+
+struct StableSortTag{};
+struct UnstableSortTag{};
+
+struct NoComparatorTag{};
+struct GreaterComparatorTag
```
MikeDvorskiy, line 41: -> `GreaterComparator` ?

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
MikeDvorskiy: It looks these constants are chosen  just for device policy... What about the host policies? Are the same values applied? Is it correct?

test/kt/esimd_radix_sort_out_of_place.cpp
```diff
@@ -44,7 +44,7 @@ test_all_view(sycl::queue q, std::size_t size, KernelParam param)
     std::cout << "\ttest_all_view(" << size << ") : " << TypeInfo().name<T>() << std::endl;
 #    endif
     std::vector<T> input(size);
-    generate_data(input.data(), size, 42);
+    TestUtils::generate_arithmetic_data(input.data(), size, 42);
```
MikeDvorskiy, line 47: I would suggest making one global (in the file) definition of this constant and use that name everywhere in file below.
`constexpr int __seed = 42;`

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,297 @@
...
+    if constexpr (std::is_same_v<Compare, NoComparatorTag>)
+    {
+        std::stable_sort(first, first + n, [](const auto& lhs, const auto& rhs) {
+            return std::get<0>(lhs) < std::get<0>(rhs);
+        });
+    }
+    if constexpr (std::is_same_v<Compare, GreaterComparatorTag>)
+    {
+        std::stable_sort(first, first + n, [](const auto& lhs, const auto& rhs) {
+            return std::get<0>(lhs) > std::get<0>(rhs);
```
MikeDvorskiy, line 148: it is not consistent (and might be non correct) with `call_sort` logic:
In case of not "NoComparatorTag" `call_sort` sorts data via `compare` passed by a caller,
but `call_reference_sort` doesn't work at all (`static_assert` is triggered).

test/parallel_api/algorithm/alg.sorting/stable_sort_by_key.pass.cpp
akukanov: I think Dmitry is correct, we return a skip code if the test had not run at all. If some tests were run, it's `done()`

documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -76,19 +76,29 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
     search sequence: [0, 2, 4, 7, 6]
     result sequence: [1, 4, 8, 10, 10]
 
-* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts the sequence's keys according to 
-  a comparioson operator. If no comparator is provided, then the elements are compared with ``operator<``.
-  The sequence's values are permutated according to the sorted sequence's keys. The prerequisite for correct
-  behavior is that the size for both keys sequence and values sequence shall be the same.  
+* ``sort_by_key``: performs a key-value sort.
+  The algorithm sorts a sequence of keys based on a comparison function object.
```
akukanov, line 80:
```suggestion
  The algorithm sorts a sequence of keys using a given comparison function object.
```

documentation/library_guide/parallel_api/additional_algorithms.rst
```diff
@@ -76,19 +76,29 @@ header.  All algorithms are implemented in the ``oneapi::dpl`` namespace.
...
 
-* ``sort_by_key``: performs a stable key-value sort. The algorithm sorts the sequence's keys according to 
-  a comparioson operator. If no comparator is provided, then the elements are compared with ``operator<``.
-  The sequence's values are permutated according to the sorted sequence's keys. The prerequisite for correct
-  behavior is that the size for both keys sequence and values sequence shall be the same.  
+* ``sort_by_key``: performs a key-value sort.
+  The algorithm sorts a sequence of keys based on a comparison function object.
+  If it is not provided, the elements are compared with ``operator<``.
+  A sequence of values is simultaneously permuted according to the sorted order of keys.
+  For correct behavior, a sequence of values must have at least as many elements as the sequence of keys.
```
akukanov, line 83:
```suggestion
  There must be at least as many values as the number of keys, otherwise the behavior is undefined.
```

documentation/library_guide/parallel_api/additional_algorithms.rst
MikeDvorskiy: We cannot check it, because for "values" we have only "begin"  iterator.

test/kt/esimd_radix_sort_out_of_place.cpp
dmitriy-sobolev: Let's avoid doing refactoring changes in tests which are not related to this PR. I can do it in a separate PR, though.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I've returned back `GreaterComparator`  together with `NoComparator` for consistency.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Let me clarify what happens in the code:

If `NoComparator` is passed, `call_sort` calls `dpl::sort_by_key` without comparator, and `call_reference_sort` calls `std::sort` with zipped pairs and std::less applied only to keys, which is an equivalent.

I suppose it is what is expected, unless I misunderstood you.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I like the suggested idea for refactoring, but it would require changing the test in many other places, e.g. changing majority of functions to accept a variadic number of arguments, getting rid of `GreaterComparator` and `NoComparator` objects.

In my opinion, the test is in better shape right now, that it was before the PR. I suppose it may be treated as an "evolutionary" change. For more radical ones, let's stick to the idea from https://github.com/oneapi-src/oneDPL/pull/1692#discussion_r1701557694, and do it separately.

documentation/library_guide/parallel_api/additional_algorithms.rst
dmitriy-sobolev: I've applied the suggestion, with slight changes proposed by Alexey below.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
SergeyKopienko: ok

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I've corrected it.

documentation/library_guide/parallel_api/additional_algorithms.rst
MikeDvorskiy: I don't  mind. Lets create an issue to don't forget.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: Now, there are dedicated functions for stable_sort/sort.

documentation/library_guide/parallel_api/additional_algorithms.rst
dmitriy-sobolev: Done: https://github.com/oneapi-src/oneDPL/issues/1779

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: > The last sentence tells: if a data generator is changed, the checker should be also changed (checker depends on data generation way). It is a downside of the test design.

By the way, the checker and generator do not depend on each other now. The current issue is that they are embedded - other variants cannot be provided unless the code structure is changed, but it is not necessary right now.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
MikeDvorskiy: Another thing which confuses: 
line 145: `if constexpr (std::is_same_v<Compare, GreaterComparator>)`
But on line 148, we dont call `GreaterComparator::operator()`....
It looks we dont call GreaterComparator::operator() everywhere. So, why we have structute GreaterComparator with operator() ?

dmitriy-sobolev: `GreaterComparator` is called by `call_sort`. `call_reference_sort` used a custom substitute with the same effect.
I've replaced it with the passed comparator using your previous suggestion.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I've added https://github.com/oneapi-src/oneDPL/issues/1780 for further refactoring.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I've reused that idea in https://github.com/oneapi-src/oneDPL/pull/1692#discussion_r1716714607, but still there is `if constexpr` branching.
I've created https://github.com/oneapi-src/oneDPL/issues/1780 for further refactoring.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
dmitriy-sobolev: I've created https://github.com/oneapi-src/oneDPL/issues/1780 for further refactoring.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -0,0 +1,293 @@
...
+void
+check_sort(const KeyIt& keys_begin, const ValIt& vals_begin,
+           const KeysOrigIt& keys_orig_begin, const ValsOrigIt& vals_orig_begin,
+           Size keys_n, Size vals_n, Compare compare, UnstableSortTag)
+{
+    using KeyT = typename std::iterator_traits<KeyIt>::value_type;
+    using ValT = typename std::iterator_traits<ValIt>::value_type;
+    std::vector<KeyT> keys_expected(keys_orig_begin, keys_orig_begin + keys_n);
+    std::vector<ValT> vals_expected(vals_orig_begin, vals_orig_begin + vals_n);
+    call_reference_sort(keys_expected.begin(), vals_expected.begin(), keys_n, compare);
```
MikeDvorskiy, line 179: It looks the same reference sorting (etalon) is called for stable and unstable cases...

dmitriy-sobolev: > Lets add //TODO about that.

Added

MikeDvorskiy (approved):
LGTM excepting some "potential refactoring things", non critical for now.

dmitriy-sobolev:
@dcbenito, could you check the changes in the documentation?

dcbenito (approved):
LGTM.

PR #1695, opened 2024-07-15
[CMake] Fix cmake workaround comment for clarity

danhoeflinger:
As discussed in :
https://github.com/oneapi-src/oneDPL/pull/1479#discussion_r1675897808

The comment for this cmake workaround was unclear, this PR should clarify things.

### DIFF
--- a/cmake/oneDPLWindowsIntelLLVMApply.cmake
+++ b/cmake/oneDPLWindowsIntelLLVMApply.cmake
@@ -36,6 +36,6 @@ endif()
 
 if (INTELLLVM_WIN_STD_IGNORE_FIX)
     include(Compiler/CMakeCommonCompilerMacros)
-    # Setting c++ standard default to 14, and enabling usage of CMAKE_CXX_STANDARD
+    # For IntelLLVM versions greater or equal to 2020, setting c++ standard default to 14, and enable usage of CMAKE_CXX_STANDARD
     __compiler_check_default_language_standard(CXX 2020 14)
 endif()


cmake/oneDPLWindowsIntelLLVMApply.cmake
```diff
@@ -36,6 +36,6 @@ endif()
 
 if (INTELLLVM_WIN_STD_IGNORE_FIX)
     include(Compiler/CMakeCommonCompilerMacros)
-    # Setting c++ standard default to 14, and enabling usage of CMAKE_CXX_STANDARD
+    # For IntelLLVM versions greater or equal to 2020, setting c++ standard default to 14, and enable usage of CMAKE_CXX_STANDARD
     __compiler_check_default_language_standard(CXX 2020 14)
```
akukanov, line 40: Should the default be set to C++17 now?

cmake/oneDPLWindowsIntelLLVMApply.cmake
akukanov, line 1: For my curiosity, where is `__compiler_check_default_language_standard` defined - somewhere in CMake files provided by the compiler itself?

danhoeflinger: This is an internal cmake function, CMake has specializations to setup individual compilers including intelLLVM.  This is a workaround which is "fixing" specifics of certain versions of cmake to be in a reasonably working state so those versions can be used on windows, so it reaches into the implementation details of cmake to accomplish this.  
Its not ideal, but it enabled usage of cmake on windows.  

At this point, there are working versions of cmake available which we could choose to "require" rather than attempting to fix. to stretch support back into versions which do not work well with intelLLVM.

cmake/oneDPLWindowsIntelLLVMApply.cmake
danhoeflinger: I suppose the correct thing would be to make it so that version 2020 defaults to 14, and 2021 defaults to 17.  I think this is when icpx switched to c++17 default,  I can check what current cmake intelLLVM code does here if we want to update it to be a more full "fix".    

However, this is only as a workaround for windows cmake in windows, so it is not really important.  Within oneDPL's cmake (after this applied workaround), we set the default to c++17 if left unspecified.  The important part of this call is the other effect, which is to enable the use of `CMAKE_CXX_STANDARD` to specify a standard library version.

danhoeflinger: I think actually this is still what CMake still calls for intelLLVM 

https://github.com/Kitware/CMake/blob/8b90d19630b0a80eed34df340f45eb1d61aff0ca/Modules/Compiler/IntelLLVM-CXX.cmake#L103

I'd be inclined to leave this as is, especially since this default value really goes unused in practice.

lslusarczyk: Thanks for clarifying comment. It clearly says what the call does. I think the most important comment would be also the reason why is it needed, which you explained here. 

Something like this: _"This is a workaround for Windows cmake. This function side effect is that it enables the use of CMAKE_CXX_STANDARD to specify a standard library version. Otherwise CMAKE_CXX_STANDARD is ignored."_

lslusarczyk (approved):

akukanov (approved):

PR #1697, opened 2024-07-15
Update to use internal version of enumerate in test

BenBrock:
This patch allows distributed ranges tests to be compiled with `gcc-14`.


lslusarczyk:
Solved conflicts and removed workaround in CI in https://github.com/oneapi-src/oneDPL/pull/1721 PR based on this work. Closing this.

PR #1698, opened 2024-07-16
Conditional use of libfmt

mateuszpn:
libfmt is used for detailed messaging in dr tests only if it is available.


test/distributed-ranges/shp/CMakeLists.txt
```diff
@@ -19,18 +19,19 @@ else ()
     message(STATUS "suitable for distributed ranges cmake version ${CMAKE_VERSION} found")
 endif()
 
+add_library(dr_shp INTERFACE)
+add_library(DR::shp ALIAS dr_shp)
```
lslusarczyk, line 23: can we implement it without this two variables dr_shp and DR::shp. This is something which is either empty or having `-DUSE_FMT`?

test/distributed-ranges/shp/CMakeLists.txt
```diff
@@ -76,7 +77,10 @@ add_executable(
 add_executable(shp-tests-3 shp-tests.cpp containers-3.cpp copy-3.cpp)
 
 foreach(test-exec IN ITEMS shp-tests shp-tests-3)
-  target_link_libraries(${test-exec} PRIVATE GTest::gtest_main fmt::fmt oneDPL)
+  target_link_libraries(${test-exec} PRIVATE GTest::gtest_main oneDPL)
+  if(fmt_FOUND)
+    target_link_libraries(${test-exec} PRIVATE fmt::fmt)
```
lslusarczyk, line 82: e.g. just write here `target_compile_options(${test-exec} PRIVATE -DUSE_FMT)` and we are done with one line and no extra variables.

test/distributed-ranges/shp/xhp-tests.hpp
```diff
@@ -18,12 +18,14 @@
...
 #include <oneapi/dpl/distributed-ranges>
 
+#ifdef USE_FMT
 #include <fmt/core.h>
 #include <fmt/format.h>
 #include <fmt/ranges.h>
 namespace drfmt {
-  using fmt::format;
+using fmt::format;
 }
```
lslusarczyk, line 27: add here
```
#else
namespace drfmt {
inline std::string format(...) { return "check failed"; }
}
```
... and get rid of most `ifdef`s added in the code.

lslusarczyk (changes requested):

test/distributed_ranges/include/common_tests.hpp
```diff
@@ -367,6 +362,7 @@ template <rng::range R1, rng::range R2> bool operator==(R1 &&r1, R2 &&r2) {
   return is_equal(std::forward<R1>(r1), std::forward<R2>(r2));
 }
 
+#ifdef USE_FMT
```
lslusarczyk, line 365: why is this needed? I guess these operators are not needed at all if code compiles well without USE_FMT flag.

test/distributed_ranges/include/common_tests.hpp
```diff
@@ -171,13 +166,13 @@ bool is_equal(std::forward_iterator auto it, rng::range auto &&r) {
...
   if (is_equal(ref, actual)) {
-    return drfmt::format("");
+    return "";
   }
   return drfmt::format("\n{}"
-                     "    ref:    {}\n"
-                     "    actual: {}\n  ",
-                     title == "" ? "" : "    " + title + "\n",
-                     rng::views::all(ref), rng::views::all(actual));
+                       "    ref:    {}\n"
```
lslusarczyk, line 172: only formatting change? and did it pass checks in changes shape?
did you run format checks?

test/distributed_ranges/include/common_tests.hpp
mateuszpn: Indeed. Removed

lslusarczyk: @mateuszpn , I see it is still not removed

test/distributed_ranges/shp/CMakeLists.txt
```diff
@@ -84,7 +81,11 @@ add_executable(
 add_executable(shp_tests_3 shp_tests.cpp containers_3.cpp copy_3.cpp)
 
 foreach(test-exec IN ITEMS shp_tests shp_tests_3)
-  target_link_libraries(${test-exec} PRIVATE GTest::gtest_main fmt::fmt oneDPL)
+  target_link_libraries(${test-exec} PRIVATE GTest::gtest_main oneDPL)
+  if(fmt_FOUND)
+    target_compile_options(${test-exec} INTERFACE -DUSE_FMT)
```
lslusarczyk, line 86: What does INTERFACE do? Why it is used here?

lslusarczyk (changes requested):
To check if this code works I introduced some error in a test and wanted to see whether I see pretty error message. Here is my change
```
diff --git a/test/distributed_ranges/common/for_each.cpp b/test/distributed_ranges/common/for_each.cpp
index 257151726..70bb55af5 100644
--- a/test/distributed_ranges/common/for_each.cpp
+++ b/test/distributed_ranges/common/for_each.cpp
@@ -32,6 +32,7 @@ void test_foreach_n(std::vector<T> v, int n, int initial_skip, auto func) {
 
   xhp::for_each_n(d_v.begin() + initial_skip, n, func);
   stdrng::for_each_n(v.begin() + initial_skip, n, func);
+  v[0] += 1;
 
   EXPECT_TRUE(equal(v, d_v));
 }
```
I've compiled in environment where fmt is installed. I veridfied it by looking that there is no warning about "no pretty messages" but a message thet
```
-- suitable for distributed ranges libfmt version 9.1.0 found
```

Then I ran foreach test, but I saw
```
/tmp/oneDPL/test/distributed_ranges/common/for_each.cpp:37: Failure
Value of: equal(v, d_v)
  Actual: false (check failed)
Expected: true
```
So it seems that despite of the fmt being present code is being still compiled without fmt.
Please fix it.

test/distributed_ranges/shp/CMakeLists.txt
```diff
@@ -29,15 +29,12 @@ endif()
 find_package(fmt QUIET)
 if(fmt_FOUND)
     if(fmt_VERSION VERSION_GREATER_EQUAL 8.1 AND fmt_VERSION VERSION_LESS 10)
-        message(STATUS "suitable for distributed ranges libfmt version ${fmt_VERSION} found")
+      message(STATUS "suitable for distributed ranges libfmt version ${fmt_VERSION} found")
```
lslusarczyk, line 32: here is 6 spaces and below 8, please revert this change

test/distributed_ranges/include/common_tests.hpp
```diff
@@ -217,7 +213,7 @@ std::string check_segments_message(auto &&r) {
                        "      segments: {}\n  ",
                        stdrng::views::all(r), stdrng::views::all(segments));
   }
-  return "";
+  return drfmt::format("");
```
lslusarczyk, line 216: without fmt all checks in tests fail

it is because lines like this returning non-empty (check failed) even if you pass "" to it

lslusarczyk (changes requested):
please fix failures of tests when they run without libfmt

lslusarczyk (approved):
works, thank you

PR #1700, opened 2024-07-16
[PROTOTYPE] Generalized two pass algorithm and copy_if

danhoeflinger:
Summary
---
This PR changes the two pass algorithm to be more generalized for use with other scan-like algorithms like copy_if.

This PR adds copy_if as an example.  partition and unique patterns should follow easily.  Set algorithms have not been evaluated, but should also be possible.

Structural changes
---
The "bridge" between blocks now uses a saved carry-out value from the previous block (where it used to depend on the output of the previous block).  The output of the previous block only applies to transform_scan patterns, and not scan_copy algorithms.  It also required different logic for inclusive and exclusive scans.  The change to use the carry unifies this logic between inclusive and exclusive scans. 

The following operations have been defined to encode a generalized scan-like pattern:
* `_GenReduceInput` : a function which accepts the input range and index to generate the data needed by the main output used in the reduction operation (to calculate the global carries)
* `_GenScanInput` :  a function which accepts the input range and index to generate the data needed by the final scan and write operations, for scan patterns
* `_ScanPred` : a unary function applied to the ouput of `_GenScanInput` to extract the component used in the scan, but not the part only required for the final write operation
* `_ReduceOp` : a binary function which is used in the reduction and scan operations
* `_FinalOp`: A function which accepts output range, index, and output of `_GenScanInput` applied to the input range.

For transform_scan, these are defined as follows:

* `_GenReduceInput` : apply the transfom to input at index and return
* `_GenScanInput` :  apply the transfom to input at index and return
* `_ScanPred` : no_op passthrough
* `_ReduceOp` : user supplied binary scan operation
* `_FinalOp`: Simple write value to output range at index

For copy_if these are defined as follows:
* `_GenReduceInput` : apply user supplied predicate to input at index and return `1` for true, `0` for false
* `_GenScanInput` :  a tuple with 3 elements:
                            1) apply user supplied predicate to input at index and return `1` for true, `0` for false (contribution to count)
                            2) apply user supplied predicate to input at index (condition for final copy)
                            3) original value (value for final copy)
* `_ScanPred` : `std::get<0>(input)`
* `_ReduceOp` : `std::plus`
* `_FinalOp`: if the second element of the tuple is true, write the third element to the output range using the index found in the first element of the tuple

Odds and ends
---
* Makes the two pass scan pattern asynchronous, depending on `__result_and_scratch_space` for async deallocation

* Changes the single workgroup copy_if to have a smaller threshold, and to use the `__result_and_scratch_space` struct to match return future type with the copy_if implementation.

* Limits range to only launch non-empty workgroups (removed early exit for empty workgroups)

TODO next:
---
Replace usages of  `__parallel_transform_scan_base` to route to `__parallel_transform_reduce_then_scan` instead.  Modify the calling code as necessary.  This should cover the rest of the scan-like patterns.  Prioritize partition, then unique, then finally, set algorithms as time allows.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -774,6 +774,26 @@ __group_scan_fits_in_slm(const sycl::queue& __queue, ::std::size_t __n, ::std::s
     return (__n <= __single_group_upper_limit && __max_slm_size >= __req_slm_size);
 }
 
+template <typename _UnaryOp>
+struct __gen_transform_input
+{
+    template <typename InRng>
+    auto operator()(InRng&& __in_rng, std::size_t __idx) const
+    {
+        return __unary_op(__in_rng[__idx]);
+    }
+    _UnaryOp __unary_op;
+};
+
+struct __simple_write_to_idx
+{
+    template<typename _OutRng, typename ValueType>
+    void operator()(_OutRng&& __out, std::size_t __idx, const ValueType& __v) const
+    {
+        __out[__idx] = __v;
+    }
+};
+
 template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _UnaryOperation, typename _InitType,
           typename _BinaryOperation, typename _Inclusive>
 auto
@@ -811,9 +831,12 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
     //_Assigner __assign_op;
     //_NoAssign __no_assign_op;
     //_NoOpFunctor __get_data_op;
+    oneapi::dpl::__par_backend_hetero::__gen_transform_input<_UnaryOperation> __gen_transform{__unary_op};
     return __future(__parallel_transform_reduce_then_scan(__backend_tag, ::std::forward<_ExecutionPolicy>(__exec),
-                                                          ::std::forward<_Range1>(__in_rng), ::std::forward<_Range2>(__out_rng),
-                                                          __binary_op, __unary_op, __init, _Inclusive{})
+                                                          ::std::forward<_Range1>(__in_rng),
+                                                          ::std::forward<_Range2>(__out_rng),
+                                                          __gen_transform, __binary_op, __gen_transform, __binary_op,
+                                                          __simple_write_to_idx{}, __init, _Inclusive{})
                     .event());
 }
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -779,7 +779,7 @@ __group_scan_fits_in_slm(const sycl::queue& __queue, ::std::size_t __n, ::std::s
 template <typename _ValueType, typename _UnaryOp>
 struct __gen_transform_input
 {
-    using __out_value_type = typename std::invoke_result<_UnaryOp, _ValueType>::type;
+    using __out_value_type = std::decay_t<typename std::invoke_result<_UnaryOp, _ValueType>::type>;
     template <typename InRng>
     auto
     operator()(InRng&& __in_rng, std::size_t __idx) const
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -805,7 +805,7 @@ struct __gen_count_pred
     using __out_value_type = _SizeType;
     template <typename _InRng>
     _SizeType
-    operator()(_InRng&& __in_rng, _SizeType __idx)
+    operator()(_InRng&& __in_rng, _SizeType __idx) const
     {
         return __pred(__in_rng[__idx]) ? _SizeType{1} : _SizeType{0};
     }
@@ -817,7 +817,7 @@ struct __gen_expand_count_pred
 {
     template <typename _InRng>
     auto
-    operator()(_InRng&& __in_rng, _SizeType __idx)
+    operator()(_InRng&& __in_rng, _SizeType __idx) const
     {
         auto ele = __in_rng[__idx];
         bool mask = __pred(ele);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -860,32 +860,62 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
                           _InitType __init, _BinaryOperation __binary_op, _Inclusive)
 {
     using _Type = typename _InitType::__value_type;
-
-    // Next power of 2 greater than or equal to __n
-    auto __n_uniform = __n;
-    if ((__n_uniform & (__n_uniform - 1)) != 0)
-        __n_uniform = oneapi::dpl::__internal::__dpl_bit_floor(__n) << 1;
-
-    // TODO: can we reimplement this with support fort non-identities as well? We can then use in reduce-then-scan
-    // for the last block if it is sufficiently small
-    constexpr bool __can_use_group_scan = unseq_backend::__has_known_identity<_BinaryOperation, _Type>::value;
-    if constexpr (__can_use_group_scan)
+    // Reduce-then-scan is dependent on sycl::shift_group_right which requires the underlying type to be trivially
+    // copyable. If this is not met, then we must fallback to the legacy implementation. The single work-group implementation
+    // requires a fundamental type which must also be trivially copyable.
+    if constexpr (std::is_trivially_copyable_v<_Type>)
     {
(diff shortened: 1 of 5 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -774,6 +776,77 @@ __group_scan_fits_in_slm(const sycl::queue& __queue, ::std::size_t __n, ::std::s
     return (__n <= __single_group_upper_limit && __max_slm_size >= __req_slm_size);
 }
 
+template <typename _ValueType, typename _UnaryOp>
+struct __gen_transform_input
+{
+    using __out_value_type = std::decay_t<decltype(::std::declval<_UnaryOp>()(::std::declval<_ValueType>()))>;
```
adamfidel, line 782: Could this be rewritten with `invoke_result`? Something like:
```suggestion
    using __out_value_type = typename std::invoke_result<_UnaryOp, _ValueType>::type;
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -604,109 +612,96 @@ struct __parallel_reduce_then_scan_scan_submitter<__sub_group_size, __max_inputs
...
-          typename _UnaryOperation, typename _InitType, typename _Inclusive>
+// General scan-like algorithm helpers
+// _GenReduceInput - a function which accepts the input range and index to generate the data needed by the main output
+//                   used in the reduction operation (to calculate the global carries)
+// _GenScanInput - a function which accepts the input range and index to generate the data needed by the final scan
+//                 and write operations, for scan patterns
+// _ScanPred - a unary function applied to the ouput of `_GenScanInput` to extract the component used in the scan, but
+//             not the part only required for the final write operation
+// _ReduceOp - a binary function which is used in the reduction and scan operations
+// _FinalOp - a function which accepts output range, index, and output of `_GenScanInput` applied to the input range
```
adamfidel, line 696: I don't mean to bikeshed too much, but perhaps we could consider a different name for the "final op"? The name doesn't really convey what its intended use is. 

From what I can tell, this operator will typically be used for writing the output. Maybe we can consider "write op" or "output op" or maybe even "final write op" if we want to keep the word final.

adamfidel (commented):
Just some initial comments.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: I think its important to make these names good to make the algorithm understandable and usable for other patterns, so please help suggest names.  I agree `_WriteOp` is better, I'll make the change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: yep thats better.  applied.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -774,6 +776,77 @@ __group_scan_fits_in_slm(const sycl::queue& __queue, ::std::size_t __n, ::std::s
...
+};
+
+template <typename _SizeType, typename _Predicate>
+struct __gen_expand_count_pred
+{
+    template <typename _InRng>
+    auto
+    operator()(_InRng&& __in_rng, _SizeType __idx)
+    {
+        auto ele = __in_rng[__idx];
```
danhoeflinger, line 822: Should we forward the `__in_rng` as we use its bracket operator?
I don't think we normally do that, but I suppose we could, right?

mmichel11 (commented):
First few minor things I see.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -604,109 +610,95 @@ struct __parallel_reduce_then_scan_scan_submitter<__sub_group_size, __max_inputs
...
                                                    /*__init_present=*/false,
                                                    /*__capture_output=*/true, __max_inputs_per_item>(
-                        __sub_group, __unary_op, __binary_op, __sub_group_carry, __in_rng, __out_rng, __start_idx, __n,
-                        __inputs_per_item, __subgroup_start_idx, __sub_group_id, __active_subgroups);
+                        __sub_group, __gen_scan_input, __scan_pred, __reduce_op, __write_op, __sub_group_carry,
+                        __in_rng, __out_rng, __start_idx, __n, __inputs_per_item, __subgroup_start_idx, __sub_group_id,
+                        __active_subgroups);
+                }
+                //if at the last element in the sequence, then we need to write out the last carry out
+                if (__sub_group_local_id == 0 && (__active_groups == __g + 1) &&
```
adamfidel, line 650: Not asking to change, but just pointing out that these conditions with `+ 1` tripped me up since I'm so used to seeing conditions like this formatted as `g == __active_groups - 1` instead. Maybe we can consider switching the condition around to be more consistent with the rest of the codebase, but it's not a big deal.

danhoeflinger: Thats fair.  I did this with the thought that `__active_[sub]groups` could be 0, and both are unsigned types.  I think now in the current state of the codebase they cannot be 0, but I was trying to avoid underflow issues.

Its also reasonable to make these signed types.  They wont be overflowing their respective types if we make them signed.

danhoeflinger: I improved the comment to make this more clear (hopefully)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -148,56 +148,56 @@ __sub_group_scan_partial(const _SubGroup& __sub_group, _ValueType& __value, _Bin
...
-          typename _LazyValueType, typename _InRng, typename _OutRng>
+          std::uint32_t __max_inputs_per_item, typename _SubGroup, typename _GenInput, typename _ScanPred,
+          typename _BinaryOp, typename _WriteOp, typename _LazyValueType, typename _InRng, typename _OutRng>
 void
-__scan_through_elements_helper(const _SubGroup& __sub_group, _UnaryOp __unary_op, _BinaryOp __binary_op,
-                               _LazyValueType& __sub_group_carry, _InRng __in_rng, _OutRng __out_rng,
-                               std::size_t __start_idx, std::size_t __n, std::uint32_t __iters_per_item,
-                               std::size_t __subgroup_start_idx, std::uint32_t __sub_group_id,
-                               std::uint32_t __active_subgroups)
+__scan_through_elements_helper(const _SubGroup& __sub_group, _GenInput __gen_input, _ScanPred __scan_pred,
```
mmichel11, line 154: When I see pred or predicate I usually think of a function that returns true or false which confused me a bit when I saw this at first. Naming this something like `__scan_op` also sounds confusing though. How about something like `__scan_input_transform`?

mmichel11: Although I do see why it is named `__scan_pred` since the unary predicate would be applied here for some scan-based algorithms, so it is difficult to think of a good name for this.

danhoeflinger: Yes, I agree.  Your suggestion looks good.  

Another option is `_ScanUnaryOp` __scan_unary_op`, but I think I like your option better.

danhoeflinger: applied.

danhoeflinger: applied.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I think we can, but I am not sure what all it gets us here.

danhoeflinger: Yeah, its probably not any real benefit, and not worth the cost to readability.  Resolving this.

mmichel11 (approved):
LGTM

adamfidel (approved):
LGTM.

PR #1702, opened 2024-07-17
Use only C++20 features

BenBrock:
There are still some compilation issues with chained views.

### DIFF
--- a/include/oneapi/dpl/internal/distributed_ranges_impl/detail/enumerate.hpp
+++ b/include/oneapi/dpl/internal/distributed_ranges_impl/detail/enumerate.hpp
@@ -16,6 +16,7 @@
 #pragma once
 
 #include <oneapi/dpl/internal/distributed_ranges_impl/detail/ranges_shim.hpp>
+#include <oneapi/dpl/internal/distributed_ranges_impl/shp/zip_view.hpp>
 
 namespace oneapi::dpl::experimental::dr
 {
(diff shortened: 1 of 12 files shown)


include/oneapi/dpl/internal/distributed_ranges_impl/detail/enumerate.hpp
```diff
@@ -57,11 +58,11 @@ class enumerate_adapter_closure
         using W = std::conditional_t<std::weakly_incrementable<S>, S, std::size_t>;
         if constexpr (rng::sized_range<R>)
         {
-            return rng::views::zip(rng::views::iota(W{0}, W{rng::size(r)}), std::forward<R>(r));
+            return oneapi::dpl::experimental::dr::shp::views::zip(rng::views::iota(W{0}, W{rng::size(r)}), std::forward<R>(r));
```
lslusarczyk, line 61: This code is already in "oneapi::dpl::experimental::dr" namespace, use just "shp::"... Please use shorter namespaces also in other places where full namespace is not needed.

include/oneapi/dpl/internal/distributed_ranges_impl/detail/enumerate.hpp
```diff
@@ -16,6 +16,7 @@
 #pragma once
 
 #include <oneapi/dpl/internal/distributed_ranges_impl/detail/ranges_shim.hpp>
+#include <oneapi/dpl/internal/distributed_ranges_impl/shp/zip_view.hpp>
```
lslusarczyk, line 19: This is common code for mhp and shp. We can't use shp specific code here.

lslusarczyk (changes requested):

PR #1707, opened 2024-07-17
[PROTOTYPE] Fallback to legacy scan implementation if underlying init type is not trivially copyable

mmichel11:
We require the init type to be trivially copyable due to requirements imposed by `sycl::shift_group_[left|right]`. If this is not the case, then we must fallback to the legacy implementation.

By the point this check is added, `std::tuple` types will have already been repacked to `oneapi::dpl::__internal::tuple` allowing them to be trivially copyable.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -879,7 +879,7 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
         }
     }
 
-    // Reduce-then-scan is dependent on sycl::sub_group_shift which requires the underlying type to be trivially
+    // Reduce-then-scan is dependent on sycl::shift_group_right which requires the underlying type to be trivially
     // copyable. If this is not met, then we must fallback to the legacy implementation.
     if constexpr (std::is_trivially_copyable_v<_Type>)
     {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -879,13 +879,45 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
         }
     }
 
-    oneapi::dpl::__par_backend_hetero::__gen_transform_input<_UnaryOperation>
-        __gen_transform{__unary_op};
-    return __future(__parallel_transform_reduce_then_scan(
-                        __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), ::std::forward<_Range1>(__in_rng),
-                        ::std::forward<_Range2>(__out_rng), __gen_transform, __binary_op, __gen_transform,
-                        oneapi::dpl::__internal::__no_op{}, __simple_write_to_idx{}, __init, _Inclusive{})
-                        .event());
+    // Reduce-then-scan is dependent on sycl::sub_group_shift which requires the underlying type to be trivially
+    // copyable. If this is not met, then we must fallback to the legacy implementation.
+    if constexpr (std::is_trivially_copyable_v<_Type>)
+    {
+        oneapi::dpl::__par_backend_hetero::__gen_transform_input<_UnaryOperation> __gen_transform{__unary_op};
+        return __future(__parallel_transform_reduce_then_scan(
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -879,13 +879,45 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
...
+        oneapi::dpl::__par_backend_hetero::__gen_transform_input<_UnaryOperation> __gen_transform{__unary_op};
+        return __future(__parallel_transform_reduce_then_scan(
+                            __backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__in_rng),
+                            std::forward<_Range2>(__out_rng), __gen_transform, __binary_op, __gen_transform,
+                            oneapi::dpl::__internal::__no_op{}, __simple_write_to_idx{}, __init, _Inclusive{})
+                            .event());
+    }
+    else
+    {
+        // Either we can't use group scan or this input is too big for one workgroup
```
adamfidel, line 895: I don't think this comment is correct at this point, since we are now falling back if we can't use group scan, or use `shift_group_right`, or the input is larger than a single work-group. Maybe we can just remove the comment.

mmichel11: I agree we do not need this comment and have removed it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -879,13 +879,45 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
...
+        using _Assigner = unseq_backend::__scan_assigner;
+        using _NoAssign = unseq_backend::__scan_no_assign;
+        using _UnaryFunctor = unseq_backend::walk_n<_ExecutionPolicy, _UnaryOperation>;
+        using _NoOpFunctor = unseq_backend::walk_n<_ExecutionPolicy, oneapi::dpl::__internal::__no_op>;
+
+        _Assigner __assign_op;
+        _NoAssign __no_assign_op;
+        _NoOpFunctor __get_data_op;
+
+        return __future(
```
danhoeflinger, line 904: Are there any types which are non-trivially-copyable for which `__can_use_group_scan` is `true_type`?

If so, then this would be a compile time issue because of two different run-time future types which can't both be run-time return values.

I think we are in the clear, but it may be good to represent that better via the code somehow.

danhoeflinger: Perhaps single-workgroup scan should be within the if side of the `if` `else` rather than above.  Either that or some `static_assert` to make it clear.

danhoeflinger: To be clear, this is because I switched single-wg scan to be use `__result_and_scratch_space` rather than a buffer of size 1 for the return, so that it matches two-pass-scan.

adamfidel: > Are there any types which are non-trivially-copyable for which `__can_use_group_scan` is `true_type`?

`__can_use_group_scan` is only true if the operator/type combination has a known identity. From what I can tell, the SYCL spec only defines known identities for types that are trivially copyable (e.g., integral types, floating point types, bool, `sycl::half). See [Table 120 of the SYCL spec](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.identities).

Based on that, it seems that there cannot be a case where `__can_use_group_scan` is true but the type is non-trivially-copyable.

mmichel11: I agree. The spec requires them to be fundamental types. I also agree it is clearer to have them in the same branch of the `if` `else`, so I will make this change.

danhoeflinger: I agree.  What do you guys think about making this "official" by only considering group scan for trivially_copyable types? 

I think this would be better for future maintenance personally, but I could be convinced otherwise.

mmichel11: This makes the most sense to me. Do you also agree? @adamfidel

danhoeflinger (approved):
LGTM

PR #1708, opened 2024-07-17
[PROTOTYPE] partition, unique families and ranges API

danhoeflinger:
Adds partition, unique families as well as the ranges API to use two pass scan.

There is some refactoring to the generalized helpers, to simplify sharing between families of scan-like algorithms using this.  
Otherwise, this just adds other entry points to use two pass scan.

This also standardizes / adds some zero element short circuits in the ranges API

Note:  for the ranges API, I try to hook in to the path to re-use what we have in the main APIs.  Specifically, this enables ranges API copy_if to be able to use single workgroup copy_if as well, where it was not previously exposed to that.

Tentative performance results for these families of algorithms are similar to what we have seen with copy_if.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -886,10 +886,10 @@ __pattern_mismatch(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterat
 //------------------------------------------------------------------------
 
 template <typename _BackendTag, typename _ExecutionPolicy, typename _Iterator1, typename _IteratorOrTuple,
-          typename _GenMask, typename _WriteOp>
+          typename _GenMask, typename _WriteOp, typename _IsUniquePattern>
 ::std::pair<_IteratorOrTuple, typename ::std::iterator_traits<_Iterator1>::difference_type>
 __pattern_scan_copy(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __last,
-                    _IteratorOrTuple __output_first, _GenMask __gen_mask, _WriteOp __write_op)
+                    _IteratorOrTuple __output_first, _GenMask __gen_mask, _WriteOp __write_op, _IsUniquePattern __is_unique_pattern)
 {
     using _It1DifferenceType = typename ::std::iterator_traits<_Iterator1>::difference_type;
 
@@ -898,15 +898,26 @@ __pattern_scan_copy(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Itera
 
     _It1DifferenceType __n = __last - __first;
 
+    if constexpr (_IsUniquePattern::value)
+    {
+        if (__n == 1)
+        {
+            oneapi::dpl::__internal::__pattern_walk2_brick(
+                __hetero_tag<_BackendTag>{}, std::forward<_ExecutionPolicy>(__exec), __first, __last, __output_first,
+                oneapi::dpl::__internal::__brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
+            return std::make_pair(__output_first + 1, 1);
+        }
+    }
+
     auto __keep1 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator1>();
     auto __buf1 = __keep1(__first, __last);
     auto __keep2 =
         oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _IteratorOrTuple>();
     auto __buf2 = __keep2(__output_first, __output_first + __n);
 
-    auto __res =
-        __par_backend_hetero::__parallel_scan_copy(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
-                                                   __buf1.all_view(), __buf2.all_view(), __n, __gen_mask, __write_op);
+    auto __res = __par_backend_hetero::__parallel_scan_copy(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
+                                                            __buf1.all_view(), __buf2.all_view(), __n, __gen_mask,
+                                                            __write_op, __is_unique_pattern);
 
     ::std::size_t __num_copied = __res.get();
     return ::std::make_pair(__output_first + __n, __num_copied);
@@ -958,7 +969,7 @@ __pattern_partition_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __e
             __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::write>(__result1),
             __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::write>(__result2)),
         oneapi::dpl::__par_backend_hetero::__gen_mask<_UnaryPredicate>{__pred},
-        oneapi::dpl::__par_backend_hetero::__write_to_idx_if_else{});
+        oneapi::dpl::__par_backend_hetero::__write_to_idx_if_else{}, /*_IsUniquePattern=*/std::false_type{});
 
     return ::std::make_pair(__result1 + __result.second, __result2 + (__last - __first - __result.second));
 }
@@ -978,7 +989,7 @@ __pattern_unique_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec
     auto __result =
         __pattern_scan_copy(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first, __last, __result_first,
                             oneapi::dpl::__par_backend_hetero::__gen_unique_mask<_BinaryPredicate>{__pred},
-                            oneapi::dpl::__par_backend_hetero::__write_to_idx_if{});
+                            oneapi::dpl::__par_backend_hetero::__write_to_idx_if<1>{}, /*_IsUniquePattern=*/std::true_type{});
 
     return __result_first + __result.second;
 }
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -939,9 +939,9 @@ __pattern_partition_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __e
     auto __buf2 = __keep2(__zipped_res, __zipped_res + __n);
 
     auto __result = oneapi::dpl::__par_backend_hetero::__parallel_partition_copy(
-        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(), __buf2.all_view(), __pred);
+        _BackendTag{}, std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(), __buf2.all_view(), __pred);
 
-    return ::std::make_pair(__result1 + __result.get(), __result2 + (__last - __first - __result.get()));
+    return std::make_pair(__result1 + __result.get(), __result2 + (__last - __first - __result.get()));
 }
 
 //------------------------------------------------------------------------
@@ -974,7 +974,7 @@ __pattern_unique_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec
     auto __buf2 = __keep2(__result_first, __result_first + __n);
 
     auto __result = oneapi::dpl::__par_backend_hetero::__parallel_unique_copy(
-        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(), __buf2.all_view(), __pred);
+        _BackendTag{}, std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(), __buf2.all_view(), __pred);
 
     return __result_first + __result.get();
 }
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -885,44 +885,6 @@ __pattern_mismatch(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterat
 // copy_if
 //------------------------------------------------------------------------
 
-template <typename _BackendTag, typename _ExecutionPolicy, typename _Iterator1, typename _IteratorOrTuple,
-          typename _GenMask, typename _WriteOp, typename _IsUniquePattern>
-::std::pair<_IteratorOrTuple, typename ::std::iterator_traits<_Iterator1>::difference_type>
-__pattern_scan_copy(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator1 __first, _Iterator1 __last,
-                    _IteratorOrTuple __output_first, _GenMask __gen_mask, _WriteOp __write_op, _IsUniquePattern __is_unique_pattern)
-{
-    using _It1DifferenceType = typename ::std::iterator_traits<_Iterator1>::difference_type;
-
-    if (__first == __last)
-        return ::std::make_pair(__output_first, _It1DifferenceType{0});
-
-    _It1DifferenceType __n = __last - __first;
-
-    if constexpr (_IsUniquePattern::value)
-    {
-        if (__n == 1)
-        {
-            oneapi::dpl::__internal::__pattern_walk2_brick(
-                __hetero_tag<_BackendTag>{}, std::forward<_ExecutionPolicy>(__exec), __first, __last, __output_first,
-                oneapi::dpl::__internal::__brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>{});
-            return std::make_pair(__output_first + 1, 1);
-        }
-    }
-
-    auto __keep1 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator1>();
-    auto __buf1 = __keep1(__first, __last);
-    auto __keep2 =
-        oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _IteratorOrTuple>();
-    auto __buf2 = __keep2(__output_first, __output_first + __n);
-
-    auto __res = __par_backend_hetero::__parallel_scan_copy(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
-                                                            __buf1.all_view(), __buf2.all_view(), __n, __gen_mask,
-                                                            __write_op, __is_unique_pattern);
-
-    ::std::size_t __num_copied = __res.get();
-    return ::std::make_pair(__output_first + __n, __num_copied);
-}
-
 template <typename _BackendTag, typename _ExecutionPolicy, typename _Iterator1, typename _Iterator2,
           typename _Predicate>
 _Iterator2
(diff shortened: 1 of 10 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -958,51 +999,29 @@ struct __invoke_single_group_copy_if
...
+        /*_Inclusive=*/std::true_type{});
 }
 
-template <typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Size, typename _Pred>
+template <typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Size, typename _Pred,
+          typename _Assign = oneapi::dpl::__internal::__pstl_assign>
 auto
 __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _ExecutionPolicy&& __exec,
-                   _InRng&& __in_rng, _OutRng&& __out_rng, _Size __n, _Pred __pred)
+                   _InRng&& __in_rng, _OutRng&& __out_rng, _Size __n, _Pred __pred, _Assign __assign = _Assign{})
```
danhoeflinger, line 1024: Noting this for myself:
  I did not extend single-wg scan to support `__assign`.  I don't think this will be a problem, but we need to do it in this PR.

danhoeflinger: This is done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -840,16 +867,32 @@ struct __get_zeroth_element
...
+};
+
+template <typename Assign = oneapi::dpl::__internal::__pstl_assign>
+struct __write_to_idx_if_else
+{
+    template <typename _OutRng, typename _SizeType, typename ValueType>
+    void
+    operator()(_OutRng&& __out, _SizeType __idx, const ValueType& __v) const
+    {
+        if (std::get<1>(__v))
```
mmichel11, line 890: We will likely need a fix similar to https://github.com/oneapi-src/oneDPL/pull/1714 here.

danhoeflinger: yes, agreed.  I'll rebase and make the appropriate changes.

danhoeflinger: should be done.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -904,9 +904,9 @@ __pattern_scan_copy(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Itera
         oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _IteratorOrTuple>();
     auto __buf2 = __keep2(__output_first, __output_first + __n);
 
-    auto __res = __par_backend_hetero::__parallel_scan_copy(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
-                                                            __buf1.all_view(), __buf2.all_view(), __n, __create_mask_op,
-                                                            __copy_by_mask_op);
+    auto __res =
+        __par_backend_hetero::__parallel_scan_copy(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
```
mmichel11, line 908: Since we're touching this code, we can make the `::std::forward` `std::forward`

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -334,50 +334,20 @@ __pattern_count(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range&& _
...
-                                                                  __get_data_op},
-            // global scan
-            __copy_by_mask_op)
-            .get();
+    auto __n = __rng1.size();
+    if (__n == 0)
+        return 0;
 
-    return __res;
+    auto __res = __par_backend_hetero::__parallel_scan_copy(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
```
mmichel11, line 347: Same comment here as above.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -815,19 +817,46 @@ struct __simple_write_to_idx
...
+};
+
+template <typename _BinaryPredicate>
+struct __gen_unique_mask
+{
+    template <typename _InRng>
+    bool
+    operator()(_InRng&& __in_rng, std::size_t __idx) const
+    {
+        if (__idx == 0)
```
mmichel11, line 838: Are there any tricks we could use to remove this conditional entirely? It feels like this is a lot of overhead for special handling of the first element.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -941,9 +989,11 @@ struct __invoke_single_group_copy_if
     // Specialization for devices that have a max work-group size of at least 1024
     static constexpr ::std::uint16_t __targeted_wg_size = 1024;
 
-    template <::std::uint16_t _Size, typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Pred>
+    template <::std::uint16_t _Size, typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Pred,
+              typename _Assign = oneapi::dpl::__internal::__pstl_assign>
     auto
-    operator()(_ExecutionPolicy&& __exec, ::std::size_t __n, _InRng&& __in_rng, _OutRng&& __out_rng, _Pred&& __pred)
+    operator()(_ExecutionPolicy&& __exec, ::std::size_t __n, _InRng&& __in_rng, _OutRng&& __out_rng, _Pred&& __pred,
```
mmichel11, line 995: Is there a reason why `Pred` is a forwarding reference and `_Assign` is not?

mmichel11 (commented):
I've had a pass over this PR and have a few comments.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -815,19 +817,46 @@ struct __simple_write_to_idx
...
 
 template <typename _Predicate>
-struct __gen_count_pred
+struct __gen_mask
+{
+    template <typename _InRng>
+    bool
+    operator()(_InRng&& __in_rng, std::size_t __idx) const
+    {
+        return __pred(__in_rng[__idx]);
```
adamfidel, line 826: Do we need to forward `__in_rng` here?

adamfidel: After second thought, I don't think it would buy us anything to forward the range here just to call `operator[]` on it, so disregard this comment.

danhoeflinger: Yeah, I questioned this myself earlier in the previous PR.  I think we *can* do it, but I don't think we have to and I don't think we get much from it.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: fixed, thanks

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Nope, fixed this and some other instances of the same.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: This is a good question...here's my best idea:
Its a little ugly, but it will be faster than the "cleaner" options.

Add a template bool to the `parallel_scan_copy` (and turn it on for `pattern_unique_copy`) which special cases the first element in the second kernel, and adds an offset of 1 to the scan operation. This would allow us to just have a single operation for the 0th index work item to make the copy, and then scan without the branch starting at element 1 instead of 0 for both kernels.  

The cleaner option would be a separate memcpy which handles the first element (and offsetting the start of the list), but if we handled that at a high level, it would be a separate memcpy or kernel which I think ends up slower.

What do you think?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -845,71 +1005,47 @@ struct __invoke_single_group_copy_if
...
 
-template <typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Size, typename _CreateMaskOp,
-          typename _CopyByMaskOp>
+template <typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Size, typename _GenMask,
+          typename _WriteOp>
 auto
 __parallel_scan_copy(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _ExecutionPolicy&& __exec,
-                     _InRng&& __in_rng, _OutRng&& __out_rng, _Size __n, _CreateMaskOp __create_mask_op,
-                     _CopyByMaskOp __copy_by_mask_op)
+                     _InRng&& __in_rng, _OutRng&& __out_rng, _Size __n, _GenMask __generate_mask, _WriteOp __write_op)
```
mmichel11, line 1030: Since we've had to add a fallback to the legacy scan implementation, I think we will need to have a `__parallel_scan_copy` and a `__parallel_reduce_then_scan_copy` instead of rewriting `__parallel_scan_copy`. Or we could add some parameter to switch between reduce-then-scan and the legacy implementation.

danhoeflinger: Yes, I'll resolve this after rebase.  Good point.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: This has been applied via #1743 (merged).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -747,14 +774,19 @@ __parallel_transform_reduce_then_scan(oneapi::dpl::__internal::__device_backend_
     const std::size_t __n = __in_rng.size();
     const std::size_t __max_inputs_per_block = __work_group_size * __max_inputs_per_item * __num_work_groups;
     std::size_t __num_remaining = __n;
+    if constexpr (__is_unique_pattern_v)
+    {
+        // skip scan of zeroth element in unique patterns
+        __num_remaining -= 1;
```
mmichel11, line 780: Should we `assert(__num_remaining > 0)` before this to make sure we are properly handling the `__n == 0` case everywhere?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -740,9 +724,10 @@ __parallel_transform_scan_single_group(oneapi::dpl::__internal::__device_backend
         using _DynamicGroupScanKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
             __par_backend_hetero::__scan_single_wg_dynamic_kernel<_BinaryOperation, _CustomName>>;
 
-        auto __event = __parallel_transform_scan_dynamic_single_group_submitter<_Inclusive::value, _DynamicGroupScanKernel>()(
-            ::std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng), std::forward<_OutRng>(__out_rng),
-            __n, __init, __binary_op, __unary_op, __max_wg_size);
+        auto __event =
+            __parallel_transform_scan_dynamic_single_group_submitter<_Inclusive::value, _DynamicGroupScanKernel>()(
+                ::std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
```
mmichel11, line 729:
```suggestion
                std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -847,7 +858,7 @@ struct __get_zeroth_element
         return std::get<0>(std::forward<_Tp>(__a));
     }
 };
-
+template <int32_t __offset = 0, typename Assign = oneapi::dpl::__internal::__pstl_assign>
```
mmichel11, line 861: Is there a reason we use a signed type here?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -951,19 +924,24 @@ __pattern_partition_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __e
...
 
-    return ::std::make_pair(__result1 + __result.second, __result2 + (__last - __first - __result.second));
+    auto __keep2 =
+        oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, decltype(__zipped_res)>();
+    auto __buf2 = __keep2(__zipped_res, __zipped_res + __n);
+
+    auto __result = oneapi::dpl::__par_backend_hetero::__parallel_partition_copy(
+        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(), __buf2.all_view(), __pred);
+
+    return ::std::make_pair(__result1 + __result.get(), __result2 + (__last - __first - __result.get()));
```
mmichel11, line 944:
```suggestion
    return std::make_pair(__result1 + __result.get(), __result2 + (__last - __first - __result.get()));
```

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -977,16 +955,28 @@ __pattern_unique_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec
...
+        return __result_first + 1;
+    }
+
+    auto __keep1 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator1>();
+    auto __buf1 = __keep1(__first, __last);
+    auto __keep2 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _Iterator2>();
+    auto __buf2 = __keep2(__result_first, __result_first + __n);
+
+    auto __result = oneapi::dpl::__par_backend_hetero::__parallel_unique_copy(
+        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(), __buf2.all_view(), __pred);
```
mmichel11, line 977:
```suggestion
        _BackendTag{}, std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(), __buf2.all_view(), __pred);
```

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -433,17 +387,27 @@ template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typ
...
-                                         ::std::forward<_Range1>(__rng), ::std::forward<_Range2>(__result),
-                                         __create_mask_op, __copy_by_mask_op);
+    auto __n = __rng.size();
+    if (__n == 0)
+        return 0;
+    if (__n == 1)
+    {
+        using CopyBrick = oneapi::dpl::__internal::__brick_copy<__hetero_tag<_BackendTag>, _ExecutionPolicy>;
+        oneapi::dpl::__par_backend_hetero::__parallel_for(
+            _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
```
mmichel11, line 399:
```suggestion
            _BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
```

mmichel11 (commented):
First pass over the recent changes. I marked a few of the places where we touched the code and left `::std`. There are a few more but I didn't want to pollute the review comments with them.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -327,14 +304,16 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         auto __size_per_wg = __iters_per_witem * __wgroup_size;
         auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_wg);
         // Storage for the results of scan for each workgroup
-        sycl::buffer<_Type> __wg_sums(__n_groups);
+
+        using _TempStorage = __result_and_scratch_storage<std::decay_t<_ExecutionPolicy>, _Type>;
+        _TempStorage __result_and_scratch{__exec, __n_groups + 1};
```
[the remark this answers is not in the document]
danhoeflinger: The legacy algorithm here used to use the last element of the temp buffer as the return of the future.  This was convenient because it also happens to be the "final carry-out" or our desired return value. 

This PR changes the functionality when `__result_and_scratch_storage` is used to have an additional return value in addition to the  `__n_groups` tempoary wg sums, and for the scan patterns that use the return, we explicitly copy the last value of the temporary array into the explicit return value.  

`__result_and_scratch_storage` when handled by the future only has the option to include the value of the "return" in the future, unless we extend it to do something else.

With all that, I think the update `__ngroups + 1` is the appropriate value.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: It provides it with a little more utility in the future to do other things.  Its very unlikely we would use a negative value here though.  The only options right now are `0` or `1`, but I think we don't lose anything to allow it to be signed. I think it will be combined at compile time with the `- 1` which precedes it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: Yes, this is a good idea to add after this constexpr block, because the kernel itself is not written to withstand an "empty" scan.

danhoeflinger: done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -702,6 +728,16 @@ struct __parallel_reduce_then_scan_scan_submitter<
     _InitType __init;
 };
 
+// reduce_then_scan requires subgroup size of 32, and performs well only on devices with fast coordinated subgroup
+// operations.  We do not want to run this can on CPU targets, as they are not performant with this algorithm.
+template <typename _ExecutionPolicy>
+bool
+__is_best_alg_reduce_then_scan(_ExecutionPolicy&& __exec)
```
mmichel11, line 735: Not a big deal but since we don't forward the forwarding reference, should we just take it as `const _ExecutionPolicy& __exec`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: No objections from my side. I guess we should just use `std::int32_t` to align with our convention.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: sure, done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed.

mmichel11 (approved):
LGTM

PR #1709, opened 2024-07-18
Split SYCL code by kernel to improve JIT compilation in CI

julianmi:


danhoeflinger (commented):
I think this is reasonable to consider, but I'm not necessarily sold here.  
This per-commit CI is not our only check on correctness (we have much more exhaustive testing).  However, it is the first line of defense against issues coming in.  
This is lazy-JIT setting is not the default compiler setting. I generally prefer the defaults to best simulate the common case for the user.  I have no real reason to believe that we may arrive at different generated code or correctness issues due to using this flag, but there is some chance there for differences.

We are only saving 2.5 minutes it seems.  Is this usually the bottleneck job or on the bottleneck runner?

The question in my head is whether this 2.5 minutes is worth some (admittedly small) risk of missed issues.

SergeyKopienko:
...and probably I know what kind of issues we can miss : it's Kernel's names when unnamed lambda is switched off: we will compile and check only names of used in Kernels. Am i correct?

danhoeflinger:
> ...and probably I know what kind of issues we can miss: it's Kernel's names when unnamed lambda is switched off: we will compile and check only names of used in Kernels. Am I correct?

Umm, that is a possibility I think.  Kernels compiled but not encountered at runtime would not be checked for same mangled name errors, that sounds plausible to me.

julianmi:
> I think this is reasonable to consider, but I'm not necessarily sold here. This per-commit CI is not our only check on correctness (we have much more exhaustive testing). However, it is the first line of defense against issues coming in. This is lazy-JIT setting is not the default compiler setting. I generally prefer the defaults to best simulate the common case for the user. I have no real reason to believe that we may arrive at different generated code or correctness issues due to using this flag, but there is some chance there for differences.
> 
> We are only saving 2.5 minutes it seems. Is this usually the bottleneck job or on the bottleneck runner?
> 
> The question in my head is whether this 2.5 minutes is worth some (admittedly small) risk of missed issues.

I agree that the speedup is limited. The longest-running CI test seems to be TBB backend on Windows so this might have no real benefit in overall CI time. However, this could become more beneficial when we increase the per-commit CI test coverage as discussed e.g. in https://github.com/oneapi-src/oneDPL/pull/1493 and in the recent splitting of the SYCL iterator tests.

SergeyKopienko:
> > ...and probably I know what kind of issues we can miss: it's Kernel's names when unnamed lambda is switched off: we will compile and check only names of used in Kernels. Am I correct?
> 
> Umm, that is a possibility I think. Kernels compiled but not encountered at runtime would not be checked for same mangled name errors, that sounds plausible to me.

Is far as I understand, when we using `fsycl-device-code-split=per_kernel` we will compile only running Kernels. It's the reason why  I have used this option in tests for `std::complex` - because we shouldn't try to compile Kernels with used `double` type when it doesn't supported on device : intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2023-2/just-in-time-compilation.html - `per_kernel: generates code to do JIT compilation of a kernel only when it is called`

danhoeflinger:
> Is far as I understand, when we using `fsycl-device-code-split=per_kernel` we will compile only called Kernels. It's the reason why I have used this option in tests for `std::complex` - because we shouldn't try to compile Kernels with used `double` type when it doesn't supported on device : intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2023-2/just-in-time-compilation.html - `per_kernel: generates code to do JIT compilation of a kernel only when it is called`

We are in agreement. I used imprecise language in my response .  What I meant to say is:
1) With the default compiler option, all kernels which are not excluded by compile time branches so the could, in theory, be encountered at runtime, and are therefore eagerly compiled and would show this issue of same mangled name.
2) With the proposed compiler option, only kernels which are encountered at runtime are compiled and would possibly show the issue of same mangled name.
3) Therefore, the set of kernels which we could "miss" by using this option are kernels who's entry point are valid runtime possibilities at the start of the execution, but are never actually encountered before exiting the test.

julianmi:
I'll close this for now since the benefit is limited and we don't have a clear picture of the potential reduced test coverage.

PR #1710, opened 2024-07-18
Upgrade CI versions

julianmi:
This PR proposes to upgrade the runners' versions to the latest OS to keep the CI running on current environments. This was noticed by @danhoeflinger in https://github.com/oneapi-src/oneDPL/pull/1479#discussion_r1670866048.

- Ubuntu 20.04 -> `ubuntu-latest` (currently 22.04)
- Windows Server 2019 -> `windows-latest` (currently 2022)
- macOS 13 -> `macos-latest` (currently 14)

It further upgrades the following GitHub actions:

- `actions/checkout`: warns about using deprecated Node.js 16
- `actions/setup-python`: warns about using deprecated Node.js 16
- `actions/upload-artifact`: warns about deprecation

I've also increased the build concurrency to 4 on Linux and Windows to match the core counts on the latest runners. This reduces the test times for these configurations.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -194,6 +194,10 @@ jobs:
         if: (matrix.device_type == 'FPGA_EMU')
         run: |
           sudo apt-get install intel-oneapi-compiler-fpga -y
+      - name: Install OpenMP dependencies
+        if: (matrix.os == 'ubuntu-latest' && matrix.cxx_compiler == 'clang++' && matrix.backend == 'omp')
+        run: |
+          sudo apt-get install libomp5 libomp-dev -y
       - name: Run testing
         shell: bash
         run: |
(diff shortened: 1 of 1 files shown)


timmiesmith:
There is a known issue with the use of checkout@v4 and Ubuntu 22.04.  See https://github.com/actions/checkout/issues/1667.  I'm not aware of a workaround at this point so I don't think it's possible to make all of the proposed changes in the set of move to the latest Ubuntu, Node20, and checkout@v4.

danhoeflinger:
Do we know why the Linux omp test is failing? 
It seems the nightly CI does not have issues with ubuntu22.04 and omp together.  Is there something we are missing in this per-commit ci?

julianmi:
> There is a known issue with the use of checkout@v4 and Ubuntu 22.04. See [actions/checkout#1667](https://github.com/actions/checkout/issues/1667). I'm not aware of a workaround at this point so I don't think it's possible to make all of the proposed changes in the set of move to the latest Ubuntu, Node20, and checkout@v4.

Thank you for making me aware of this. This issue seems to be for outdated self-hosted runners only, right? We should use the GitHub-managed runners and I don't see any error messages with the proposed switch to `checkout@v4` and Ubuntu 22.04.

julianmi:
> Do we know why the Linux omp test is failing? It seems the nightly CI does not have issues with ubuntu22.04 and omp together. Is there something we are missing in this per-commit ci?

We need to apparently install `libomp5` and `libomp-dev` packages on Ubuntu 22.04. All tests are now green, and we don't have any deprecation messages anymore.

danhoeflinger (approved):
LGTM, It may be good to get confirmation from someone like @kboyarinov or @Alexandr-Konovalov  that there is no need for a specific ubuntu version for the pstl-offload job.

Alexandr-Konovalov:
It should be no requirement for the specific version of Ubuntu from PSTL offload side.

Alexandr-Konovalov (approved):

PR #1712, opened 2024-07-18
[PROTOTYPE] Add unused temporary storage to single work-group scan to fix use-after free error

adamfidel:
We were throwing away the result and scratch container inside of the future in `__parallel_transform_scan` because we needed to return a future without storage. This was causing a use-after free error because the scratch storage was freed before it was used in the kernel.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -426,10 +426,10 @@ struct __parallel_transform_scan_dynamic_single_group_submitter<_Inclusive,
                                                                 __internal::__optional_kernel_name<_ScanKernelName...>>
 {
     template <typename _Policy, typename _InRng, typename _OutRng, typename _InitType, typename _BinaryOperation,
-              typename _UnaryOp>
+              typename _UnaryOp, typename _TempStorage>
     auto
     operator()(const _Policy& __policy, _InRng&& __in_rng, _OutRng&& __out_rng, ::std::size_t __n, _InitType __init,
-               _BinaryOperation __bin_op, _UnaryOp __unary_op, ::std::uint16_t __wg_size)
+               _BinaryOperation __bin_op, _UnaryOp __unary_op, ::std::uint16_t __wg_size, _TempStorage __temp_storage)
     {
         using _ValueType = typename _InitType::__value_type;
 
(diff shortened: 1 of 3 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -681,6 +681,10 @@ __parallel_transform_scan_single_group(oneapi::dpl::__internal::__device_backend
     // Specialization for devices that have a max work-group size of 1024
     constexpr ::std::uint16_t __targeted_wg_size = 1024;
 
+    using _ValueType = typename _InitType::__value_type;
+    using _TempStorage = __result_and_scratch_storage<std::decay_t<_ExecutionPolicy>, _ValueType>;
+    _TempStorage __result_and_scratch{__exec, 0};
```
danhoeflinger, line 686: Can we name this `__dummy_result_and_scratch` or something? 
Also can we add a comment explaining the purpose of this dummy scratch space?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -699,17 +703,17 @@ __parallel_transform_scan_single_group(oneapi::dpl::__internal::__device_backend
                         ::std::integral_constant<::std::uint16_t, __num_elems_per_item>, _BinaryOperation,
                         /* _IsFullGroup= */ std::true_type, _Inclusive, _CustomName>>>()(
                     ::std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
-                    std::forward<_OutRng>(__out_rng), __n, __init, __binary_op, __unary_op);
+                    std::forward<_OutRng>(__out_rng), __n, __init, __binary_op, __unary_op, __result_and_scratch);
```
danhoeflinger, line 706: Do we need to pass this scratch space in only to return it?  
Perhaps it is more clear if we instead return an event from the submitter, and construct the future here with the dummy variable.  It would make it more clear that it goes unused, and minimize the changes required.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: I renamed it and added a comment explaining why it's here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: Good point. I changed the submitters to return an event and constructed the future in the outer context instead.

mmichel11 (approved):
LGTM

danhoeflinger:
LGTM (from phone)

PR #1714, opened 2024-07-19
[PROTOTYPE] Resolve conversion issues between internal tuple and std::tuple in zip_iterator.pass

mmichel11:
`zip_iterator.pass` runs into some compilation issues currently due to conversion issues between `std::tuple` and our internal tuple in cases in which they should be convertible but a compilation failure is still encountered. 

To fix this issue, we apply the same trick applied in the legacy `copy_if` implementation:  https://github.com/oneapi-src/oneDPL/blob/e3ebb58c1cc97261909c3bb3b47e4b3747be8dde/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L618

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -856,8 +856,13 @@ struct __write_to_idx_if
     void
     operator()(_OutRng&& __out, _SizeType __idx, const ValueType& __v) const
     {
+        // Use of an explicit cast to our internal tuple type is required to resolve conversion issues between our
+        // internal tuple and std::tuple. If the underlying type is not a tuple, then the type will just be passed through.
+        using _ConvertedTupleType =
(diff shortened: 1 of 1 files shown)


danhoeflinger (approved):
LGTM.  We will need to incorporate this into the changes for partition / unique as well.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -856,8 +856,13 @@ struct __write_to_idx_if
     void
     operator()(_OutRng&& __out, _SizeType __idx, const ValueType& __v) const
     {
+        // Use of an explicit cast to our internal tuple type is required to resolve conversion issues between our
+        // internal tuple and std::tuple. If the underlying type is not a tuple, then the type will just be passed through.
```
danhoeflinger, line 860: I think we may need a similar fix on line 808.

mmichel11: We don't have any tests catching it here, but I agree the issue may arise so I have added a similar fix.

danhoeflinger:
LGTM after similar fix.

PR #1715, opened 2024-07-19
Remove slice

BenBrock:
Remove `index.hpp`, which is only needed for matrix code currently not included.  Remove `slice`, which was a non-standard range adaptor and the only place `index` was used in the oneDPL codebase.


BenBrock:
Please ignore; meant to make this PR against `distributed-ranges`, and looks like I can't delete it, just close it.

PR #1720, opened 2024-07-22
Remove incorrect noexcept specifications from algorithms that may throw

timmiesmith:
Additional static analysis of oneDPL tests has identified other algorithms specified as 'noexcept' that may throw.  This PR extends similar work done in #1640.

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -112,7 +112,7 @@ __brick_walk1(_ForwardIterator __first, _ForwardIterator __last, _Function __f, 
 template <class _RandomAccessIterator, class _Function>
 void
 __brick_walk1(_RandomAccessIterator __first, _RandomAccessIterator __last, _Function __f,
-              /*vector=*/::std::true_type) noexcept
+              /*vector=*/::std::true_type)
 {
     __unseq_backend::__simd_walk_1(__first, __last - __first, __f);
 }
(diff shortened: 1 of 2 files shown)


mmichel11 (approved):
LGTM after CI passes

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -112,7 +112,7 @@ __brick_walk1(_ForwardIterator __first, _ForwardIterator __last, _Function __f,
 template <class _RandomAccessIterator, class _Function>
 void
 __brick_walk1(_RandomAccessIterator __first, _RandomAccessIterator __last, _Function __f,
-              /*vector=*/::std::true_type) noexcept
+              /*vector=*/::std::true_type)
 {
```
danhoeflinger, line 116: So is it the provided function that may throw, or just simply the bracket dereference operator? (within `__simd_walk_1`)

SergeyKopienko:
I think this problem i smore complicated.
Looks like this PR fixed only one problem case in code and probably we still have a lot of another cases.

include/oneapi/dpl/pstl/algorithm_impl.h
timmiesmith: It's the body of the functor '__f'' that may throw.  It's several layers up in the stack from this point.

timmiesmith:
> I think this problem i smore complicated. Looks like this PR fixed only one problem case in code and probably we still have a lot of another cases.

I'm addressing the issues as they're identified by the static analysis.  I'd prefer to continue this approach instead of taking a broader approach that may remove the qualifier where we may actually be able to use it.

SergeyKopienko (approved):
LGTM

PR #1723, opened 2024-07-23
Tune the amount of groups in `__parallel_find_or` pattern

SergeyKopienko:
In this PR we tune the amount of groups in `__parallel_find_or` pattern.
This approach give us some performance boost on bigger data sizes.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1182,7 +1182,7 @@ struct __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backe
             // If our work capacity is not enough to process all data in one iteration, will tune the number of work-groups
             if (__iters_per_work_item > 1)
             {
-                // Empirically found formula for typical devices.
+                // Empirically found formula for GPU devices.
                 const auto __rng_x = __rng_n / __base_rng_n;
                 const auto __required_iters_per_work_item = std::max(std::sqrt(__rng_x), 1.);
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1182,8 +1182,8 @@ struct __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backe
             {
                 // Empirically found formula for GPU devices.
                 // TODO : need to re-evaluate this formula.
-                const auto __rng_x = __rng_n / 4096;
-                const float __desired_iters_per_work_item = std::max(std::sqrt(__rng_x), 1.);
+                const float __rng_x = (float)__rng_n / 4096.f;
+                const float __desired_iters_per_work_item = std::max(std::sqrt(__rng_x), 1.f);
 
                 if (__iters_per_work_item < __desired_iters_per_work_item)
                 {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1182,12 +1182,14 @@ struct __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backe
             {
                 // Empirically found formula for GPU devices.
                 const auto __rng_x = __rng_n / 4096;
-                const auto __desired_iters_per_work_item = std::max(std::sqrt(__rng_x), 1.);
+                const float __desired_iters_per_work_item = std::max(std::sqrt(__rng_x), 1.);
 
                 if (__iters_per_work_item < __desired_iters_per_work_item)
                 {
-
-                    auto __k = oneapi::dpl::__internal::__dpl_bit_ceil(
+                    // We should halve the number of work-groups until the number of iterations per work-item
+                    // is greater than or equal to the desired number of iterations per work-item.
+                    // For that we evaluate the max power of 2 not smaller than the given value.
+                    std::size_t __k = oneapi::dpl::__internal::__dpl_bit_ceil(
                         (std::size_t)std::ceil(__desired_iters_per_work_item / __iters_per_work_item));
                     __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size *
                                                                                          __iters_per_work_item * __k);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1158,8 +1158,6 @@ struct __parallel_find_or_nd_range_tuner
             __n_groups = 1;
         }
 
-        _PRINT_INFO_IN_DEBUG_MODE(__exec, __wgroup_size, __max_cu);
-
         return {__n_groups, __wgroup_size};
     }
 };
@@ -1205,8 +1203,6 @@ struct __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backe
                     __iters_per_work_item =
                         oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __n_groups * __wgroup_size);
                 }
-
-                _PRINT_INFO_IN_DEBUG_MODE(__exec, __wgroup_size);
 
                 __nd_range_params = {__n_groups, __wgroup_size};
             }
@@ -1362,6 +1358,8 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
         __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backend_tag>{}(__exec, __rng_n);
     const auto __n_groups = std::get<0>(__nd_range_params);
     const auto __wgroup_size = std::get<1>(__nd_range_params);
+
+    _PRINT_INFO_IN_DEBUG_MODE(__exec, __wgroup_size);
 
     using _AtomicType = typename _BrickTag::_AtomicType;
     const _AtomicType __init_value = _BrickTag::__init_value(__rng_n);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1174,11 +1174,8 @@ struct __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backe
     std::tuple<std::size_t, std::size_t>
     operator()(const _ExecutionPolicy& __exec, const std::size_t __rng_n) const
     {
-        // Define common tuner type
-        using __parallel_find_or_nd_range_tuner_common = __parallel_find_or_nd_range_tuner<int>;
-
         // Call common tuning function to get the work-group size
-        auto [__n_groups, __wgroup_size] = __parallel_find_or_nd_range_tuner_common{}(__exec, __rng_n);
+        auto [__n_groups, __wgroup_size] = __parallel_find_or_nd_range_tuner<int>{}(__exec, __rng_n);
 
         if (__n_groups > 1)
         {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1339,12 +1339,12 @@ __parallel_find_or_impl_multiple_wgs(oneapi::dpl::__internal::__device_backend_t
 }
 
 // Base pattern for __parallel_or and __parallel_find. The execution depends on tag type _BrickTag.
-template <typename _ExecutionPolicy, typename _Brick, typename _BrickTag, typename _GroupsTuner, typename... _Ranges>
+template <typename _ExecutionPolicy, typename _Brick, typename _BrickTag, typename... _Ranges>
 ::std::conditional_t<
     ::std::is_same_v<_BrickTag, __parallel_or_tag>, bool,
     oneapi::dpl::__internal::__difference_t<typename oneapi::dpl::__ranges::__get_first_range_type<_Ranges...>::type>>
 __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Brick __f,
-                   _BrickTag __brick_tag, const _GroupsTuner& __n_groups_tuner, _Ranges&&... __rngs)
+                   _BrickTag __brick_tag, _Ranges&&... __rngs)
 {
     using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
     using _FindOrKernelOneWG =
@@ -1358,7 +1358,8 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     assert(__rng_n > 0);
 
     // Evaluate the amount of work-groups and work-group size
-    const auto __nd_range_params = __n_groups_tuner(__exec, __rng_n);
+    const auto __nd_range_params =
+        __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backend_tag>{}(__exec, __rng_n);
     const auto __n_groups = std::get<0>(__nd_range_params);
     const auto __wgroup_size = std::get<1>(__nd_range_params);
 
@@ -1417,9 +1418,7 @@ __parallel_or(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _Exec
     return oneapi::dpl::__par_backend_hetero::__parallel_find_or(
         __backend_tag,
         __par_backend_hetero::make_wrapped_policy<__or_policy_wrapper>(::std::forward<_ExecutionPolicy>(__exec)), __f,
-        __parallel_or_tag{},
-        __par_backend_hetero::__parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backend_tag>{},
-        __buf.all_view(), __s_buf.all_view());
+        __parallel_or_tag{}, __buf.all_view(), __s_buf.all_view());
 }
 
 // Special overload for single sequence cases.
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,27 +1144,112 @@ struct __early_exit_find_or
...
+
+            const auto __it_bound = std::find_if(__lower_bounds_of_sizes.crbegin(), __lower_bounds_of_sizes.crend(),
+                                                 [__rng_n](std::size_t __i) { return __i <= __rng_n; });
+            if (__it_bound != __lower_bounds_of_sizes.crend())
+            {
+                const auto __offset = std::distance(__lower_bounds_of_sizes.cbegin(), __it_bound.base()) - 1;
+                const auto __it_size = __required_iters_per_work_items.cbegin() + __offset;
+
+                const std::size_t __required_iters_per_work_item = *__it_size;
+                if (__required_iters_per_work_item > 0)
```
adamfidel, line 1209: From what I can tell, `__required_iters_per_work_item` must be one of the values in `__required_iters_per_work_items`, which are all greater than 0. What is the reason for guarding this section with `if (__required_iters_per_work_item > 0)`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,27 +1144,112 @@ struct __early_exit_find_or
...
+                {
+                    auto __iters_per_work_item =
```
adamfidel, line 1211: A comment explaining what this block is doing might be helpful. Perhaps something like "We halve the number of work-groups until the number of iterations per work-item is greater than or equal to the desired number of iterations per work-item.

adamfidel (commented):
A few questions:
- It is stated that the performance is better for larger input sizes. Does this have any affect on smaller input sizes?
- For which devices do we see a performance benefit?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,27 +1144,113 @@ struct __early_exit_find_or
...
+        if (__rng_n <= __wgroup_size)
+            return 1;
+
+        // Size: [ 16'777'216, ..,  67'108'864 ) -> minimum number of iterations per work-item is 164
+        // Size: [  4'194'304, ..,  16'777'216 ) -> minimum number of iterations per work-item is  32
+        // Size: [  1'048'576, ...,  4'194'304 ) -> minimum number of iterations per work-item is  16
+        // Size: [    262'144, ...,  1'048'576 ) -> minimum number of iterations per work-item is   8
+        if (__n_groups > 1)
+        {
+            constexpr std::array<std::size_t, 6> __lower_bounds_of_sizes         = { 262'144, 1'048'576, 4'194'304, 16'777'216, 67'108'864 };
```
julianmi, line 1196: I think I follow this approach but we might want to simplify this. The current values could be derived from the input range since we double `__required_iters_per_work_items ` for multiples of four of `__lower_bounds_of_sizes`. Calculating them could remove the need for `find_if`.

julianmi (commented):
I think a simpler tuning approach you might want to try is to fix `__iters_per_work_item` to powers of two based on the input sizes. This might generate more optimized code and would remove the need for the complex approach of calculating `__n_groups`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,27 +1144,114 @@ struct __early_exit_find_or
 // parallel_find_or - sync pattern
 //------------------------------------------------------------------------
 
+template <typename Tag>
+struct __parallel_find_or_n_groups_tuner
+{
+    // Calculate the number of work groups.
+    //
+    // @param _ExecutionPolicy&& __exec - policy
```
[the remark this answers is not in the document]
SergeyKopienko: Removed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,27 +1144,114 @@ struct __early_exit_find_or
...
+    operator()(_ExecutionPolicy&& /*__exec*/, std::size_t __n_groups, const std::size_t __wgroup_size,
+               const std::size_t __rng_n) const
+    {
+        // No tuning for FPGA_EMU
+#if _ONEDPL_FPGA_EMU
+        return __n_groups;
+#endif
+
+        // If all source data fits into one work-group, then we need only one work-group
+        if (__rng_n <= __wgroup_size)
```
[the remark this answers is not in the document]
SergeyKopienko: It's some short exit + processing of one corner case.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,27 +1144,114 @@ struct __early_exit_find_or
...
+            return 1;
+
+        // Size: [ 67'108'864, .., 268'435'456 ) -> minimum number of iterations per work-item is 128
+        // Size: [ 16'777'216, ..,  67'108'864 ) -> minimum number of iterations per work-item is 164
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,27 +1144,113 @@ struct __early_exit_find_or
...
+#endif
+
+        // If all source data fits into one work-group, then we need only one work-group
+        if (__rng_n <= __wgroup_size)
+            return 1;
+
+        // Size: [ 16'777'216, ..,  67'108'864 ) -> minimum number of iterations per work-item is 164
+        // Size: [  4'194'304, ..,  16'777'216 ) -> minimum number of iterations per work-item is  32
+        // Size: [  1'048'576, ...,  4'194'304 ) -> minimum number of iterations per work-item is  16
+        // Size: [    262'144, ...,  1'048'576 ) -> minimum number of iterations per work-item is   8
```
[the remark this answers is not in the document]
SergeyKopienko: I have tried to made it, but no performance profit has been detected.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: I think my approach is more customizable in the future: the current values is only some best values found by me on concrete device. In the future we may correct them.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
SergeyKopienko: Removed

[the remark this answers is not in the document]
SergeyKopienko: Removed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Do we understand why this gives us a performance boost?  
Perhaps it is some combination of a minimum level of occupancy mixed with sufficiently large number of iterations to get maximal benefit from early exit opportunities...

I'm asking because I'm wondering if there is a more targeted way to reach this same result, as this includes many empirically tuned numbers. Reducing the number of these magic values, or even better, relying upon some query-able device characteristics would be more maintainable certainly.

SergeyKopienko: "I'm asking because I'm wondering if there is a more targeted way to reach this same result, as this includes many empirically tuned numbers. Reducing the number of these magic values, or even better, relying upon some query-able device characteristics would be more maintainable certainly."
 - unfortunately I haven't any ideas about this at current time.

SergeyKopienko: We can simple drop this PR, it's give us not more then 10% profit.

danhoeflinger: I'm not necessarily suggesting to drop it, I just wonder if there is a more direct way of going about it. 

The method before the PR was to just use the number of compute units as the number of groups.  This makes sense as it allows us to fill the device, but also have large wgs to maximize early exit.

Our benchmarks have a good variety of data configurations, including cases with no data to find at all.  However, I'm wondering this is tuning too much to the benchmark data. I think our basic understanding is that there is some penalty from underutilizing the hardware (by progressive factors of 2), which we are weighing against the expected savings we have from potential to early exit by increasing the iters per workitem by progressive factors of 2.

I would expect that we would want to target either: 
(a) a specific number of work groups
-or-
(b) a specific number of iters per work item

Beyond that I fear we might be somehow training to our benchmarking data, unless we have a compelling reasoning behind it.

julianmi: I agree with this. I think the speedup on PVC shows that there is a better launch configuration. I think we should first understand why this improves the runtime and then come up with some calculation potential based of the number of compute units. Also, we should benchmark this on different hardware architectures to not introduce slowdowns.

I made a quick prototype to use iters per work item with powers of two, but this was worse. I suspect that the tuned n values don't match anymore because we end up with a different number of groups than in the current approach. So, collecting data for the iterations or tuning the number of groups might be better suited here.

SergeyKopienko:
@julianmi, @danhoeflinger implementation has been updated.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Please see my comment at https://github.com/oneapi-src/oneDPL/pull/1723#issuecomment-2250116248

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,27 +1144,84 @@ struct __early_exit_find_or
...
+    template <typename _ExecutionPolicy>
+    std::size_t
+    operator()(_ExecutionPolicy&& /*__exec*/, std::size_t __n_groups, const std::size_t __wgroup_size,
+               const std::size_t __rng_n) const
+    {
+        assert(__rng_n > 0);
+
+        if (__n_groups > 1)
+        {
+            constexpr std::size_t __base_rng_n = 4096;
```
danhoeflinger, line 1175: I think this is going in a better direction.  

Does it make sense to represent `4096` here as some combination of `max_compute_units`, `[max_]subgroup_size`, or `max_workgroup_size`?

Do you have an understanding of why `sqrt` is appropriate here?

SergeyKopienko:
@danhoeflinger, @julianmi, @adamfidel Could you please take a look again?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: We use `sqrt` here to have required (found) best values in some check data size points.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -298,7 +298,7 @@ class __kernel_compiler
 template <typename _Policy>
 inline void
 // Passing policy by value should be enough for debugging
-__print_device_debug_info(_Policy __policy, size_t __wg_size = 0, size_t __max_cu = 0)
+__print_device_debug_info(const _Policy& __policy, size_t __wg_size = 0, size_t __max_cu = 0)
```
SergeyKopienko, line 301: We did copy on execution policy here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -309,7 +309,8 @@ __print_device_debug_info(_Policy __policy, size_t __wg_size = 0, size_t __max_c
 }
 #else
 template <typename _Policy>
-inline void __print_device_debug_info(_Policy, size_t = 0, size_t = 0)
+inline void
+__print_device_debug_info(const _Policy& __policy, size_t = 0, size_t = 0)
```
SergeyKopienko, line 313: We had copying of execution policy here in case when `_ONEDPL_DEBUG_SYCL` is undefined too .

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1134,6 +1134,85 @@ struct __early_exit_find_or
...
+    std::tuple<std::size_t, std::size_t>
+    operator()(const _ExecutionPolicy& __exec, const std::size_t __rng_n) const
+    {
+        // Define common tuner type
+        using __parallel_find_or_nd_range_tuner_common = __parallel_find_or_nd_range_tuner<int>;
+
+        // Call common tuning function to get the work-group size
+        auto __nd_range_params = __parallel_find_or_nd_range_tuner_common{}(__exec, __rng_n);
+
+        auto __n_groups = std::get<0>(__nd_range_params);
```
akukanov, line 1183: Here I would use structured binding as well.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1134,6 +1134,85 @@ struct __early_exit_find_or
...
+    static constexpr std::size_t __base_rng_n = 4096;
+
+    // Tune the amount of work-groups and work-group size
+    template <typename _ExecutionPolicy>
+    std::tuple<std::size_t, std::size_t>
+    operator()(const _ExecutionPolicy& __exec, const std::size_t __rng_n) const
+    {
+        // Define common tuner type
+        using __parallel_find_or_nd_range_tuner_common = __parallel_find_or_nd_range_tuner<int>;
+
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1134,6 +1134,85 @@ struct __early_exit_find_or
...
+        return {__n_groups, __wgroup_size};
+    }
+};
+
+// No tuning for FPGA_EMU because we are not going to tune here the performance for FPGA emulation.
+#if !_ONEDPL_FPGA_EMU
+template <>
+struct __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backend_tag>
+{
+    static constexpr std::size_t __base_rng_n = 4096;
```
akukanov, line 1170: Is not it rather `__base_wg_size`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1134,6 +1134,85 @@ struct __early_exit_find_or
...
+        {
+            auto __wgroup_size = std::get<1>(__nd_range_params);
+
+            auto __iters_per_work_item =
+                oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __n_groups * __wgroup_size);
+            
+            // If our work capacity is not enough to process all data in one iteration, will tune the number of work-groups
+            if (__iters_per_work_item > 1)
+            {
+                // Empirically found formula for typical devices.
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1134,6 +1134,72 @@ struct __early_exit_find_or
...
+            {
+                // Empirically found formula for GPU devices.
+                const auto __rng_x = __rng_n / 4096;
+                const auto __desired_iters_per_work_item = std::max(std::sqrt(__rng_x), 1.);
+
+                if (__iters_per_work_item < __desired_iters_per_work_item)
+                {
+
+                    auto __k = oneapi::dpl::__internal::__dpl_bit_ceil(
+                        (std::size_t)std::ceil(__desired_iters_per_work_item / __iters_per_work_item));
```
danhoeflinger, line 1191: Does this `ceil` actually do anything?
It seems to me that the types should be `std::size_t` here, but maybe I'm missing something.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1134,6 +1134,72 @@ struct __early_exit_find_or
...
+        {
+            auto __iters_per_work_item =
+                oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __n_groups * __wgroup_size);
+
+            // If our work capacity is not enough to process all data in one iteration, will tune the number of work-groups
+            if (__iters_per_work_item > 1)
+            {
+                // Empirically found formula for GPU devices.
+                const auto __rng_x = __rng_n / 4096;
+                const auto __desired_iters_per_work_item = std::max(std::sqrt(__rng_x), 1.);
```
danhoeflinger, line 1185: We may want to add a TODO to revisit this question in the future when we are considering some more automated way of finding these empirical values / formulas.  

With the infrastructure we have now, I think this is a reasonable way of doing things, but I'm a little worried that we are just fitting to our benchmarking data and target hardware, and with a different set of data we might end up with a different formula here.  This is specifically interesting to these "early exit" routines because there can be such large variations in performance based on when the "early exit" is allowed by the data.

I don't think its any worse in this respect than before this PR so I don't object.  

With the addition of a TODO, I think I'm OK with the PR as it stands.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1134,6 +1134,75 @@ struct __early_exit_find_or
...
+
+                if (__iters_per_work_item < __desired_iters_per_work_item)
+                {
+                    // We should halve the number of work-groups until the number of iterations per work-item
+                    // is greater than or equal to the desired number of iterations per work-item.
+                    // For that we evaluate the max power of 2 not smaller than the given value.
+                    const std::size_t __k = oneapi::dpl::__internal::__dpl_bit_ceil(
+                        (std::size_t)std::ceil(__desired_iters_per_work_item / __iters_per_work_item));
+                    __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size *
+                                                                                         __iters_per_work_item * __k);
```
akukanov, line 1196:
```suggestion
                    // Multiply work per item by a power of 2 to reach the desired number of iterations.
                    // Bit ceiling rounds the ratio up to the next power of 2.
                    const std::size_t __k = oneapi::dpl::__internal::__dpl_bit_ceil(
                        (std::size_t)std::ceil(__desired_iters_per_work_item / __iters_per_work_item));
                    // Proportionally reduce the number of work groups.
                    __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(
                        __rng_n, __wgroup_size * __iters_per_work_item * __k);
```

akukanov (approved):

PR #1724, opened 2024-07-23
Implement TODO: move the loop into peer algorithms

dmitriy-sobolev:
This essentially implements this TODO:
> // TODO: move the whole loop to be a part of peer algorithms

This will allow adding one more peer algorithm which iterates over radix bits rather than radix states; or it will be easier to experiment at least.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
@@ -367,7 +367,7 @@ struct __peer_prefix_helper<__radix_bits, _OffsetT, __peer_prefix_algo::atomic_f
 
     template <typename _OffsetHistogramAcc>
     _OffsetT
-    __peer_contribution(std::uint32_t __bucket, _OffsetHistogramAcc& __offset_histogram)
+    __peer_contribution(std::uint32_t __bucket, _OffsetHistogramAcc& __histogram)
     {
         _OffsetT __offset = 0;
         _ONEDPL_PRAGMA_UNROLL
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -357,40 +357,50 @@ struct __peer_prefix_helper<_OffsetT, __peer_prefix_algo::atomic_fetch_or>
...
-        sycl::group_barrier(__sgroup);
-        ::std::uint32_t __peer_mask_bits = __atomic_peer_mask.load();
-        ::std::uint32_t __sg_total_offset = sycl::popcount(__peer_mask_bits);
-
-        // get the local offset index from the bits set in the peer mask with index less than the work item ID
-        __peer_mask_bits &= __item_mask;
-        __new_offset_idx |= __is_current_bucket * (__offset_prefix + sycl::popcount(__peer_mask_bits));
-        return __sg_total_offset;
+        _OffsetT __offset = 0;
+        _ONEDPL_PRAGMA_UNROLL
```
julianmi, line 373: We use 8 radix bits and 256 resulting states, right? This seems like a lot to me unrolling. Did you experiment without this directive? Or with using a smaller value?

We've seen increased compile times when using unrolling directives in the past. Did you check if this has an effect here as well?

dmitriy-sobolev: We currently sort 4 bits per pass, but it may be more in the future. I am going to remove the pragma, to limit the PR only to such refactoring.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -400,30 +410,38 @@ struct __peer_prefix_helper<_OffsetT, __peer_prefix_algo::scan_then_broadcast>
...
-struct __peer_prefix_helper<_OffsetT, __peer_prefix_algo::subgroup_ballot>
+template <std::uint32_t __radix_bits, typename _OffsetT>
+struct __peer_prefix_helper<__radix_bits, _OffsetT, __peer_prefix_algo::subgroup_ballot>
 {
     using _TempStorageT = __empty_peer_temp_storage;
 
     sycl::sub_group __sgroup;
     ::std::uint32_t __self_lidx;
     sycl::ext::oneapi::sub_group_mask __item_sg_mask;
+    static constexpr std::uint32_t __radix_states = 1 << __radix_bits;
```
adamfidel, line 444: Instead of calculating `__radix_states` in each helper, maybe it makes sense to instead pass in `__radix_states` as a template parameter. It doesn't appear that `__radix_bits` is used for anything else besides calculating `__radix_states`.

dmitriy-sobolev: Done

adamfidel (approved):
LGTM. Thanks for making these changes.

PR #1729, opened 2024-07-25
Fixes in __detail namespace

mateuszpn:
- Implementation functions moved to __detail namespace.
- `sp/containers/details.hpp` removed - not used


include/oneapi/dpl/internal/distributed_ranges_impl/detail/iterator_adaptor.hpp
```diff
@@ -23,17 +23,12 @@
...
-{
-
 template <typename R>
 concept has_segments_method = requires(R r)
 {
     {r.segments()};
 };
 
-} // namespace
-
```
akukanov, line 36: This change adds `concept has_segments_method` into `namespace oneapi::dpl::experimental::dr`, which is sort of equivalent to making it a public API. Is it intentional? If not, I would use `namespace __detail` instead.

Just in case, in my comment https://github.com/oneapi-src/oneDPL/pull/1479#discussion_r1686830006 about unnamed namespaces, I did not imply it is always bad - I pointed out to a particular strange case of an unnamed namespace used directly within the default global namespace.

include/oneapi/dpl/internal/distributed_ranges_impl/detail/ranges.hpp
```diff
@@ -53,9 +53,9 @@ template <typename Iter>
 concept is_remote_iterator_shadow_impl_ =
     std::forward_iterator<Iter> && has_rank_method<Iter> && !disable_rank<std::remove_cv_t<Iter>>;
 
-} // namespace
+} // namespace __detail
 
-namespace
+namespace __detail
 {
```
akukanov, line 59: The same namespace is closed and immediately reopened.

include/oneapi/dpl/internal/distributed_ranges_impl/detail/ranges.hpp
```diff
@@ -163,11 +163,11 @@ struct segments_fn_
...
 };
 
-} // namespace
+} // namespace __detail
 
-inline constexpr auto segments = segments_fn_{};
+inline constexpr auto segments = __detail::segments_fn_{};
 
-namespace
+namespace __local
```
akukanov, line 170: Why not also `__detail`?

mateuszpn: `__local::local_fn_` and `__detail::local_fn_` are different, not sure however at the moment why we need both

include/oneapi/dpl/internal/distributed_ranges_impl/detail/iterator_adaptor.hpp
mateuszpn: Should be in `__detail`.
`__detail::has_segments_method` concept is duplicated in `iterator_adaptor.hpp` -> to be removed

include/oneapi/dpl/internal/distributed_ranges_impl/detail/ranges.hpp
akukanov: I see, `__detail::local_fn_` uses `ranges::local` if available, otherwise behaves (almost) as `std::identity`.
I think the best is to make this behavior clear from the name - for example, name it as `local_or_identity`.

Note also that `std::identity` uses forwarding to preserve the value category, see https://eel.is/c++draft/func.identity. The "identity" `operator()` misses it.

include/oneapi/dpl/internal/distributed_ranges_impl/detail/ranges.hpp
```diff
@@ -359,17 +346,19 @@ struct local_fn_
...
 };
 
-inline constexpr auto local = local_fn_{};
-
 } // namespace __detail
 
+inline constexpr auto segments = __detail::segments_fn_{};
+
+inline constexpr auto local_or_identity = __detail::local_or_identity_fn_{};
+
```
akukanov, line 361: `local_or_identity` is seemingly a helper for internal use; should it really be added into the same namespace with the public APIs? I would keep it in `__detail`.

lslusarczyk: The same for
```
inline constexpr auto local = __detail::local_fn_{};
inline constexpr auto segments = __detail::segments_fn_{};
```
?
Do you think we should also put them in detail? I think not. Regarding _segments_ I it is a public interface. Local is public interface too. 

And regarding local_or_identity I think it looks better if it is in the same namespace as local and segments. But if you have strong feeling it should be in __detail I can move (leaving segments and local outside __detail). Please decide.

include/oneapi/dpl/internal/distributed_ranges_impl/detail/ranges.hpp
akukanov, line 1: There are two seemingly very similar internal concepts defined in the file: `is_localizable` and `has_local`. Can these be united?

include/oneapi/dpl/internal/distributed_ranges_impl/detail/ranges.hpp
akukanov: Maybe `local_or_identity` coud even be moved into another file, if there is a suitable place. That would make the distinction even clearer.

include/oneapi/dpl/internal/distributed_ranges_impl/detail/ranges.hpp
lslusarczyk: is_localizable is for range-of-ranges that is for sliding_view which is not included in the part we add to oneDPL, I'm removing it.

include/oneapi/dpl/internal/distributed_ranges_impl/detail/ranges.hpp
lslusarczyk: moved to __detail

PR #1733, opened 2024-07-26
Replace usage of std::complex in by_segment tests

mmichel11:
Previously, we used `std::complex` in the tests for `inclusive_scan_by_segment`, `exclusive_scan_by_segment`, and `reduce_by_segment` algorithms. The purpose of this was to test with some non-fundamental type. However, `std::complex` is technically not officially supported in SYCL kernels and failures in these tests over time have reflected the state of `std::complex` support as opposed to the actual correctness of the by_segment algorithms. We already have explicit tests for current `std::complex` in our testing suite, so I believe these should be fully separated from our parallel API tests. We also do not use `std::complex` throughout our other parallel API tests.

I've replaced these tests with tests over a custom 2D point type. Essentially, a similar option is performed as before but no longer depends on `std::complex`.

### DIFF
--- a/test/support/utils.h
+++ b/test/support/utils.h
@@ -822,14 +822,14 @@ template <typename _Tp>
 struct MaxFunctor<MatrixPoint<_Tp>>
 {
     auto
-    euclidean_distance(const MatrixPoint<_Tp>& __x) const
+    sum(const MatrixPoint<_Tp>& __x) const
     {
-        return std::sqrt(__x.m * __x.m + __x.n * __x.n);
+        return __x.m + __x.n;
     }
     MatrixPoint<_Tp>
     operator()(const MatrixPoint<_Tp>& __x, const MatrixPoint<_Tp>& __y) const
     {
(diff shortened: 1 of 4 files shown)


test/support/utils.h
```diff
@@ -777,6 +777,37 @@ struct UserBinaryPredicate
     }
 };
 
+template <typename T>
+struct MatrixPoint
```
adamfidel, line 781: Instead of defining a new class here, can we use something like `sycl::vec<T, 2>` or even `std::pair<T>`? I believe it'll effectively have the same level of test coverage but with less code.

mmichel11: I don't think `sycl::vec` will work here since it will also go through host policy testing. Also, since `std::pair` doesn't have an `operator+` we would need to remove the testing that relies on the default binary operator.

adamfidel: Ah, okay. Then I think it's fine to leave it as is in that case.

julianmi (approved):
LGTM

danhoeflinger (approved):
LGTM, it makes sense not to rely on complex here.

PR #1734, opened 2024-07-26
Resolve compilation errors with sycl::queue default device selection for older icpx compilers

mmichel11:
Currently, we create our default  `sycl::queue` through providing `sycl::default_selector_v` to its constructor. Despite being apart of SYCL2020, this is problematic for some older compilers we still support and test with where this feature has not been implemented (e.g. icpx 2022.3):
```
error: no type named 'default_selector_v' in namespace 'sycl'; did you mean 'default_selector'?
    static sycl::queue __q(sycl::default_selector_v);
```
The default constructor of `sycl::queue` guarantees that the default device will be selected, so relying on this resolves our issue.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
@@ -123,7 +123,7 @@ class alignas(sycl::queue) __queue_holder
 inline sycl::queue
 __get_default_queue()
 {
-    static sycl::queue __q(sycl::default_selector_v);
+    static sycl::queue __q;
     return __q;
 }
 


include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
```diff
@@ -123,7 +123,7 @@ class alignas(sycl::queue) __queue_holder
 inline sycl::queue
 __get_default_queue()
 {
-    static sycl::queue __q(sycl::default_selector_v);
+    static sycl::queue __q;
```
SergeyKopienko, line 126: 1) Which warranties we have that the work will the same as before?
2) May be better to check compiler version?

dmitriy-sobolev (approved):
LGTM.

I confirm the the queue() and queue(sycl::default_selector_v) are equivalents.

dmitriy-sobolev:
I see that Sergey have just left comments. It might be safer to address them than to merge it as is.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
danhoeflinger: According to the sycl 2020 spec, a default constructed queue ... "Constructs a SYCL queue instance using the device constructed from the default_selector_v.".

danhoeflinger (approved):
LGTM, according to the spec, these should be equivalent.  

We can't guarantee the implementation makes these exactly equivalent I guess but I think specializing to the compiler version is redundant.

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
SergeyKopienko: ok

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h
mmichel11: And just to add on: for compilers that do not support `default_selector_v`, similar guarantees are made in older SYCL specifications (SYCL 1.2.1):
```
Constructs a SYCL queue instance using the device returned by an instance of
default_selector
```

PR #1735, opened 2024-07-29
merge-sort: optimize leaf stage

dmitriy-sobolev:


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -66,36 +212,37 @@ struct __parallel_sort_submitter<_IdType, __internal::__optional_kernel_name<_Le
...
 
         // 2. Merge sorting
         oneapi::dpl::__par_backend_hetero::__buffer<_ExecutionPolicy, _Tp> __temp_buf(__exec, __n);
         auto __temp = __temp_buf.get_buffer();
         bool __data_in_temp = false;
         _IdType __n_sorted = __leaf;
+        const bool __is_cpu = __exec.queue().get_device().is_cpu();
         const std::uint32_t __chunk = __is_cpu ? 32 : 4;
-        __steps = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __chunk);
+        const std::uint32_t __steps = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __chunk);
```
julianmi, line 245: Is `uint32_t` enough since `__n` is `size_t`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -57,6 +59,150 @@ struct __leaf_sort_kernel
...
+
+    void
+    sort(const sycl::nd_item<1>& __item) const
+    {
+        sycl::sub_group __sg = __item.get_sub_group();
+        sycl::group __wg = __item.get_group();
+        const std::uint32_t __wg_id = __wg.get_group_linear_id();
+        const std::uint32_t __sg_id = __sg.get_group_linear_id();
+        const std::uint32_t __sg_size = __sg.get_local_linear_range();
+        const std::uint32_t __sg_inner_id = __sg.get_local_linear_id();
```
julianmi, line 148:
```suggestion
        const std::uint32_t __sg_local_id = __sg.get_local_linear_id();
```
Should we change this to follow SYCL's language?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -159,6 +306,51 @@ struct __parallel_sort_submitter<_IdType, __internal::__optional_kernel_name<_Le
...
+
+    using _Tp = oneapi::dpl::__internal::__value_t<_Range>;
+
+    std::variant<_LeafXL, _LeafL, _LeafM, _LeafS>
+    select(const sycl::queue& __q, _Range& __rng, _Compare __comp) const
+    {
+        const auto __max_wg_size = __q.get_device().template get_info<sycl::info::device::max_work_group_size>();
+        auto __max_slm_items = __q.get_device().template get_info<sycl::info::device::local_mem_size>() / sizeof(_Tp);
+        // Get the work group size adjusted to the local memory limit.
+        // Pessimistically reduce it by 0.8 to take into account memory used by compiled kernel.
```
julianmi, line 331: Is .8 enough since we use .5 in similar places?

dmitriy-sobolev: I stuck to something larger than 0.5, because now group algorithms are used. 0.8 was based on "gut feeling".
However, I've replaced it with 0.5, because this is what was agreed upon (I do not know the reason, though). it can be reconsidered later, if needed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: It may be not enough, thanks for noticing. I've corrected it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -57,6 +59,150 @@ struct __leaf_sort_kernel
...
+        const std::uint32_t __wg_id = __wg.get_group_linear_id();
+        const std::uint32_t __sg_id = __sg.get_group_linear_id();
+        const std::uint32_t __sg_size = __sg.get_local_linear_range();
+        const std::uint32_t __sg_inner_id = __sg.get_local_linear_id();
+        const std::uint32_t __sg_process_size = __sg_size * __data_per_workitem;
+        const std::size_t __wg_start = __wg_id * __process_size;
+        const std::uint32_t __sg_start = __sg_id * __sg_process_size;
+        const std::size_t __wg_end = __wg_start + std::min<std::size_t>(__process_size, __n - __wg_start);
+        const std::uint32_t __adjusted_wg_size = __wg_end - __wg_start;
+        // 1. Load
```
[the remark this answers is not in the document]
dmitriy-sobolev: That's a good suggestion. However, it implies that it also should be applied to other parts of the library: it is orthogonal to this PR, which I want to limit to the merge-sort. It can be done in a separate PR.

Could you point any other part of library which may benefit from it? I am going to investigate it, however, I suppose, you may already have something in mind.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -159,6 +306,51 @@ struct __parallel_sort_submitter<_IdType, __internal::__optional_kernel_name<_Le
...
+    select(const sycl::queue& __q, _Range& __rng, _Compare __comp) const
+    {
+        const auto __max_wg_size = __q.get_device().template get_info<sycl::info::device::max_work_group_size>();
+        auto __max_slm_items = __q.get_device().template get_info<sycl::info::device::local_mem_size>() / sizeof(_Tp);
+        // Get the work group size adjusted to the local memory limit.
+        // Pessimistically reduce it by 0.8 to take into account memory used by compiled kernel.
+        // TODO: find a way to generalize getting of reliable work-group size.
+        __max_slm_items /= 0.8;
+
+        if (__max_slm_items > _LeafXL::storage_size() && __max_wg_size >= 1024 && __rng.size() >= (1 << 20))
```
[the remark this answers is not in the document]
dmitriy-sobolev: I've changed the methodology to avoid magical numbers where possible.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -159,6 +306,51 @@ struct __parallel_sort_submitter<_IdType, __internal::__optional_kernel_name<_Le
     }
 };
 
+template <typename _Range, typename _Compare>
+struct __leaf_sorter_selector
+{
+    // 1024 is greater than or equal the maximum work-group size for the majority of devices
```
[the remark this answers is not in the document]
dmitriy-sobolev: Good suggestion. I've followed it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -159,6 +310,72 @@ struct __parallel_sort_submitter<_IdType, __internal::__optional_kernel_name<_Le
...
+        const std::uint8_t __oversubscription = __d.is_gpu() ? 8 : 1;
+        const auto __max_cu = __d.template get_info<sycl::info::device::max_compute_units>();
+        const auto __saturation_point = __max_cu * __max_sg_size * __oversubscription;
+        const auto __desired_data_per_workitem = __n / __saturation_point;
+
+        // Pessimistically double the memory requirement to take into account memory used by compiled kernel.
+        // TODO: investigate if the adjustment can be less conservative
+        const std::size_t __max_slm_items =
+            __d.template get_info<sycl::info::device::local_mem_size>() / (sizeof(_Tp) * 2);
+        if (__max_slm_items >= _Leaf8::storage_size(__max_wg_size) && __desired_data_per_workitem >= 8)
```
SergeyKopienko, line 355: May be better get this `8` number in condition `__desired_data_per_workitem >= 8` from `_Leaf8` struct - I am about `static constexpr std::uint16_t __data_per_workitem = _DataPerWorkitem;`

SergeyKopienko: At this place and below

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -159,6 +310,72 @@ struct __parallel_sort_submitter<_IdType, __internal::__optional_kernel_name<_Le
...
+        {
+            return _Leaf8(__rng, __comp, __max_wg_size);
+        }
+        else if (__max_slm_items >= _Leaf4::storage_size(__max_wg_size) && __desired_data_per_workitem >= 4)
```
SergeyKopienko, line 359: `else` not required.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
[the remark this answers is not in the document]
julianmi: I agree that this is out of scope. Let's discuss offline.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
julianmi: I think .5 was very pessimistically since we've seen issues with debug builds. .8 might be fine for this kernel though.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -159,6 +310,72 @@ struct __parallel_sort_submitter<_IdType, __internal::__optional_kernel_name<_Le
...
+    // TODO: reconsider the value if another algorithm is used,
+    //       or an internal cap is set (e.g. sorting 2 sequences of 4/8 items each)
+    using _Leaf8 = __leaf_sorter<8, _Range, _Compare>;
+    using _Leaf4 = __leaf_sorter<4, _Range, _Compare>;
+    // 2 is the smallest reasonable value for merge-path group sorter since it loads 2 values at least
+    using _Leaf2 = __leaf_sorter<2, _Range, _Compare>;
+
+    using _Tp = oneapi::dpl::__internal::__value_t<_Range>;
+
+    std::variant<_Leaf8, _Leaf4, _Leaf2>
```
SergeyKopienko, line 326: My assumption that `variant` / `visit` pattern probably not required here due you return only one object from this function. Also, in the call of `std::visit` you have only one object and index.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -16,12 +16,15 @@
 #ifndef _ONEDPL_PARALLEL_BACKEND_SYCL_MERGE_SORT_H
 #define _ONEDPL_PARALLEL_BACKEND_SYCL_MERGE_SORT_H
 
-#include <cmath>     // std::log2
-#include <limits>    // std::numeric_limits
-#include <cassert>   // assert
-#include <utility>   // std::swap
-#include <cstdint>   // std::uint32_t, ...
-#include <algorithm> // std::min
+#include <bit>         // std::bit_floor
```
mmichel11, line 19: Should we switch this include for `../../utils.h` since that is where `__dpl_bit_floor` is defined?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -16,12 +16,15 @@
...
+#include <cmath>       // std::log2
+#include <limits>      // std::numeric_limits
+#include <cassert>     // assert
+#include <utility>     // std::swap
+#include <cstdint>     // std::uint32_t, ...
+#include <variant>     // std::variant, std::visit
```
SergeyKopienko, line 25: @dmitriy-sobolev please take a look to https://github.com/oneapi-src/oneDPL/pull/1755 - I have implemented required logic without variant - visitor usage.

dmitriy-sobolev:
I've found that the changes result on incorrect sort on Nvidia GPUs. I am debugging it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: `../../utils.h` is included now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: Thanks for the idea of the strategy. I see it as introduction of 2 layers, where each one serves its one purpose: dispatching basing on an index or a leaf type. This is one more proof of "We can solve any problem by introducing an extra level of indirection" theorem :)
 
I've adopted it [0daae6e](https://github.com/oneapi-src/oneDPL/pull/1735/commits/0daae6e5e218f458c47d11c89600900673ea70de) with some modifications.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: Would you mind if I leave it as is? Using just 8 looks simpler.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: Would you mind if a leave it as is? A single block of if-else just looks more visually appealing to me. 
I guess having "else" here is a matter of code look.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: `variant`/`visit` has been removed in [0daae6e](https://github.com/oneapi-src/oneDPL/pull/1735/commits/0daae6e5e218f458c47d11c89600900673ea70de).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
SergeyKopienko: Is these changes from the PR https://github.com/oneapi-src/oneDPL/pull/1735 ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -34,19 +35,20 @@ namespace dpl
...
+        for (_Size i = __start; i < __end; ++i)
         {
-            for (_Size1 j = __start + 1; j < __start + __end - i; ++j)
+            for (_Size j = __start + 1; j < __start + __end - i; ++j)
             {
                 // forwarding references allow binding of internal tuple of references with rvalue
-                auto&& __first_item = __acc[j - 1];
-                auto&& __second_item = __acc[j];
+                // TODO: avoid bank conflicts, or reconsider the algorithm
+                auto&& __first_item = __storage[j - 1];
```
SergeyKopienko, line 50: As we may see at https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_interface_for_local_accessors, `operator[]` return reference:
```C++
  /* Available only when: (Dimensions > 0) */
  reference operator[](id<Dimensions> index) const;
```

So why you are using `&&` here ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -57,6 +59,154 @@ struct __leaf_sort_kernel
...
+        bool __data_in_temp = false;
+        std::uint32_t __next_sorted = __sorted * 2;
+        // ctz precisely calculates log2 of an integral value which is a power of 2, while
+        // std::log2 may be prone to rounding errors on some architectures
+        std::int16_t __iters = sycl::ctz(__sorted_final) - sycl::ctz(__sorted);
+        for (std::int16_t __i = 0; __i < __iters; ++__i)
+        {
+            const std::uint32_t __id_local = __id % __next_sorted;
+            // Borders of the ranges to be merged
+            const std::uint32_t __start1 = std::min<std::uint32_t>((__id / __next_sorted) * __next_sorted, __end);
```
SergeyKopienko, line 83: May you explain, why should we use here `<std::uint32_t>` ?

SergeyKopienko:
@dmitriy-sobolev , in this PR you have two calls of `depends_on` : mandatory required to capture their params by value (now it captured by reference so it's will work incorrectly).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
SergeyKopienko: What's about
```C++
const std::uint32_t __start1 = std::min(__id - __id % __next_sorted, __end);
```
?

dmitriy-sobolev: > May you explain, why should we use here <std::uint32_t> ?

It is used to help `std::min` with selection of a common type for the arguments. Perhaps, it is a sign that the function must explicitly take `__end` parameter as `std::uint32_t` to highlight its requirements.

> What's about const std::uint32_t __start1 = std::min(__id - __id % __next_sorted, __end);

That's a good idea, especially because `__id % __next_sorted` part has already been calculated.

dmitriy-sobolev: Done

dmitriy-sobolev:
> @dmitriy-sobolev , in this PR you have two calls of `depends_on` : mandatory required to capture their params by value (now it captured by reference so it's will work incorrectly).

Thanks. I've fixed it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: Indeed, forwarding reference is not needed with the current changes. I've removed it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
SergeyKopienko: I don't see the change for
https://github.com/oneapi-src/oneDPL/pull/1735/files/d29e35c10604b2f0e55336ab59bba3e67fbcb738#r1704399376

dmitriy-sobolev: It is in [1cd360d](https://github.com/oneapi-src/oneDPL/pull/1735/commits/1cd360dce432a016372508852871ef44cc99d0bf).

SergeyKopienko:
@dmitriy-sobolev please take a look to the code like this:
```C++
                            const auto& __rng1 = oneapi::dpl::__ranges::drop_view_simple(__dst, __offset);
                            const auto& __rng2 = oneapi::dpl::__ranges::drop_view_simple(__dst, __offset + __n1);
```
1) `&` isn't required here;
2) Really we create the new instance of `drop_view_simple` in this code, so I propose to rewrite it:
```C++
                            const oneapi::dpl::__ranges::drop_view_simple __rng1(__dst, __offset);
                            const oneapi::dpl::__ranges::drop_view_simple __rng2(__dst, __offset + __n1);
```

dmitriy-sobolev:
> @dmitriy-sobolev please take a look to the code like this:
> 
> ```c++
>                             const auto& __rng1 = oneapi::dpl::__ranges::drop_view_simple(__dst, __offset);
>                             const auto& __rng2 = oneapi::dpl::__ranges::drop_view_simple(__dst, __offset + __n1);
> ```
> 
> 1. `&` isn't required here;
> 2. Really we create the new instance of `drop_view_simple` in this code, so I propose to rewrite it:
> 
> ```c++
>                             const oneapi::dpl::__ranges::drop_view_simple __rng1(__dst, __offset);
>                             const oneapi::dpl::__ranges::drop_view_simple __rng2(__dst, __offset + __n1);
> ```
> 
> There are four places...

Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -16,16 +16,17 @@
 #ifndef _ONEDPL_PARALLEL_BACKEND_SYCL_MERGE_SORT_H
 #define _ONEDPL_PARALLEL_BACKEND_SYCL_MERGE_SORT_H
 
-#include <cmath>     // std::log2
-#include <limits>    // std::numeric_limits
-#include <cassert>   // assert
-#include <utility>   // std::swap
-#include <cstdint>   // std::uint32_t, ...
-#include <algorithm> // std::min
+#include <cmath>       // std::log2
```
mmichel11, line 19: I think we can remove this include now that we use `sycl::ctz`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -168,39 +319,112 @@ class __sort_global_kernel;
...
+
+    const std::size_t __max_wg_size = __device.template get_info<sycl::info::device::max_work_group_size>();
+    std::size_t __wg_size = oneapi::dpl::__internal::__dpl_bit_floor(__max_wg_size);
+
+    const bool __is_cpu = __device.is_cpu();
+    std::uint32_t __max_sg_size{};
+    if (__is_cpu)
+    {
+        const auto __sg_sizes = __device.template get_info<sycl::info::device::sub_group_sizes>();
+        __max_sg_size = __sg_sizes.empty() ? 1 : *std::max_element(__sg_sizes.begin(), __sg_sizes.end());
```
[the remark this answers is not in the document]
dmitriy-sobolev: This is for a case if the call returns an empty vector. The specification says nothing whether it can be empty or not. 
However, implicitly, it seems it should not be empty, since the minimum number of sub-groups in a work-group is 1:

> info::device::max_num_sub_groups: Returns the maximum number of sub-groups in a work-group for any kernel executed on the device. The minimum value is 1.

I will get rid of this check.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -168,39 +319,112 @@ class __sort_global_kernel;
...
+    }
+    // Assume CPUs handle one sub-group (SIMD) per CU;
+    // Assume GPUs handle multiple sub-groups per CU,
+    // while the maximum work-group size takes hardware multithreading (occupancy) into account
+    const std::size_t __max_hw_wg_size = __is_cpu ? __max_sg_size : __max_wg_size;
+    const auto __max_cu = __device.template get_info<sycl::info::device::max_compute_units>();
```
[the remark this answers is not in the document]
dmitriy-sobolev: The `.template` instruction is for `get_info` which is a template member. 

I tried to get rid of it and got a error:

```
oneDPL/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h:363:48: error: use 'template' keyword to treat 'get_info' as a dependent template name
  363 |     const std::size_t __max_wg_size = __device.get_info<sycl::info::device::max_work_group_size>();
      |                                                ^
      |                                                template

```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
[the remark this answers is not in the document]
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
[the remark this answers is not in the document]
dmitriy-sobolev: I've faced the same issue with ICPX 2022.2, 2023.2 and 2024.2.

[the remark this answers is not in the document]
dmitriy-sobolev: I've no clear understanding why .template is necessary in this exact context.
I have not managed to reproduce it using a simpler example (https://godbolt.org/z/c75s1aPYd), however it seems, `.template` construction does not harm.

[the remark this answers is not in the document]
mmichel11: Interesting, it is no big deal, so let's keep it as is.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -168,39 +318,112 @@ class __sort_global_kernel;
...
+    const auto __saturation_point = __max_cu * __max_hw_wg_size;
+    const auto __desired_data_per_workitem = __n / __saturation_point;
+
+    // Pessimistically double the memory requirement to take into account memory used by compiled kernel.
+    // TODO: investigate if the adjustment can be less conservative
+    const std::size_t __max_slm_items =
+        __device.template get_info<sycl::info::device::local_mem_size>() / (sizeof(_Tp) * 2);
+    auto __caller = __parallel_sort_submitter_caller<_IndexT>();
+    if (__max_slm_items >= _Leaf8::storage_size(__wg_size) && __desired_data_per_workitem >= 8)
```
SergeyKopienko, line 386: @dmitriy-sobolev Do we want to use `__static_monotonic_dispatcher` here?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -168,39 +318,112 @@ class __sort_global_kernel;
...
+    // 2 is the smallest reasonable value for merge-path group sorter since it loads 2 values at least
+    using _Leaf2 = __leaf_sorter<2, std::decay_t<_Range>, _Compare>;
+
+    using _Tp = oneapi::dpl::__internal::__value_t<_Range>;
+
+    const std::size_t __n = __rng.size();
+    auto __device = __exec.queue().get_device();
+
+    const std::size_t __max_wg_size = __device.template get_info<sycl::info::device::max_work_group_size>();
+    std::size_t __wg_size = oneapi::dpl::__internal::__dpl_bit_floor(__max_wg_size);
```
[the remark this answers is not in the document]
dmitriy-sobolev: > My understanding is that we need the work-group size to be a power-of-2 which is why we do this bit floor. 

Correct

> is there any guarantee that this floored __wg_size is supported by the device?

I've never thought about such a case, assuming one can use any arbitrary work-groups size in the range [1, max-work-group-size]. I was not able to find any explicit proof in the spec, but here is an example of doing almost the same thing a I'm doing:

https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_kernel_introspection

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -168,39 +318,112 @@ class __sort_global_kernel;
...
+    if (__is_cpu)
+    {
+        const auto __sg_sizes = __device.template get_info<sycl::info::device::sub_group_sizes>();
+        __max_sg_size = *std::max_element(__sg_sizes.begin(), __sg_sizes.end());
+    }
+    // Assume CPUs handle one sub-group (SIMD) per CU;
+    // Assume GPUs handle multiple sub-groups per CU,
+    // while the maximum work-group size takes hardware multithreading (occupancy) into account
+    const std::size_t __max_hw_wg_size = __is_cpu ? __max_sg_size : __max_wg_size;
+    const auto __max_cu = __device.template get_info<sycl::info::device::max_compute_units>();
```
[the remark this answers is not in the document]
dmitriy-sobolev: I decided to stick to that value because it provides significantly better performance comparing to the previous leaf implementation, but retains generalization.

Let me share calculations regarding Intel® Data Center GPU Max (1100 model, for example) saturation:

1.  Real saturation (how it maps HW in my view): 
`VE Count * Max Sub-Group Size * Max Occupancy = 448 * 32 * 8 = 114'688`
2. Current calculation: 
`VE Count * Max Work-Group Size = 448 * 1024 = 458'752`
3. If CU count was reported as XC count: 
`XC Count * Max Work-Group Size = 57'344`

Note that there is ticket for Intel GPU driver to rise max work-group size to 2048 to achieve full utilization of XC with a single group. So 1st or 3rd option is what is needed for Intel GPUs. 3rd is better, because it is aligned with what we get for other vendors. So, I suppose saturation is already correct for Nvidia GPUs. 

Currently, the difference between desired and real saturation is 4x for Intel GPU. Let me check the performance effect on dividing it by 4. I did some experiments previously, performance was not affected much, but let me do a more targeted one. 

I will refrain from any vendor dispatch as possible, though. If the performance impact is small, I will leave it as is.

Anyway, I will add a note in our code, and try to submit an issue for reporting XC instead of VE (probably, it was discussed before and there is a reason, but let me try).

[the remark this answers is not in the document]
mmichel11: Thanks, I agree we should avoid vendor dispatch as much as possible and only do it if the performance benefit is large. After this experiment on the performance impact, I am ready to approve.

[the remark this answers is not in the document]
dmitriy-sobolev: Speed up for different saturation points on Intel® Data Center GPU Max 1550 (1T, 512 VE) against **512 * 1024** saturation point:

  | n | 64 * 1024 | 128 *   1024 | 256 * 1024
-- | -- | -- | -- | --
merge sort (float) | 256 | 0.98 | 1.02 | 1.02
merge sort (float) | 512 | 1.00 | 1.04 | 1.00
merge sort (float) | 1024 | 1.00 | 1.00 | 1.00
merge sort (float) | 2048 | 1.00 | 1.00 | 1.03
merge sort (float) | 4096 | 1.00 | 1.02 | 1.00
merge sort (float) | 8192 | 1.02 | 1.00 | 1.02
merge sort (float) | 16384 | 1.00 | 0.98 | 1.00
merge sort (float) | 32768 | 1.02 | 1.00 | 1.00
merge sort (float) | 65536 | 1.03 | 1.01 | 1.03
merge sort (float) | 131072 | 0.99 | 0.99 | 0.99
merge sort (float) | 262144 | 0.98 | 0.99 | 1.00
merge sort (float) | 524288 | 1.04 | 1.11 | 1.00
merge sort (float) | 1048576 | 1.29 | 1.29 | 1.18
merge sort (float) | 2097152 | 1.15 | 1.14 | 1.15
merge sort (float) | 4194304 | 1.00 | 1.00 | 1.00
merge sort (float) | 8388608 | 1.00 | 1.00 | 1.00
merge sort (float) | 16777216 | 1.00 | 1.00 | 1.00
merge sort (float) | 33554432 | 1.01 | 1.01 | 1.01
merge sort (float) | 67108864 | 1.01 | 1.01 | 1.01
merge sort (float) | 134217728 | 1.01 | 1.01 | 1.01
merge sort (float) | 268435456 | 1.01 | 1.01 | 1.01

I've measured performance adjusting the current saturation point. It appears 4x adjustment to match theoretical HW saturation makes a lot for 500K - 4M sizes.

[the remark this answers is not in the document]
dmitriy-sobolev: I think this benefit worth thinking about. Let me check, what can be done...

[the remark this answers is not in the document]
dmitriy-sobolev: I haven't managed to find a solution, which is not controversial, and I've added comments.
Let's proceed with this PR, and discuss this issue more separately.

mmichel11 (approved):
LGTM, although you may want to check with other reviewers before merging

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -57,6 +56,153 @@ struct __leaf_sort_kernel
...
+          __process_size(__data_per_workitem * __workgroup_size), __sub_group_sorter(), __group_sorter()
+    {
+        assert((__process_size & (__process_size - 1)) == 0 && "Process size must be a power of 2");
+    }
+
+    void
+    sort(const sycl::nd_item<1>& __item, const _StorageAcc& __storage_acc) const
+    {
+        sycl::sub_group __sg = __item.get_sub_group();
+        sycl::group __wg = __item.get_group();
```
danhoeflinger, line 142: small nit: 
I think `__wg` is unused.

dmitriy-sobolev: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -168,39 +317,115 @@ class __sort_global_kernel;
...
-        using _LeafSortKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-            __sort_leaf_kernel<_CustomName, _wi_index_type>>;
-        using _GlobalSortKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-            __sort_global_kernel<_CustomName, _wi_index_type>>;
-        using _CopyBackKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-            __sort_copy_back_kernel<_CustomName, _wi_index_type>>;
-        return __parallel_sort_submitter<_wi_index_type, _LeafSortKernel, _GlobalSortKernel, _CopyBackKernel>()(
-            oneapi::dpl::__internal::__device_backend_tag{}, std::forward<_ExecutionPolicy>(__exec),
-            std::forward<_Range>(__rng), __comp);
+        return __submit_selecting_leaf<std::size_t>(std::forward<_ExecutionPolicy>(__exec), std::forward<_Range>(__rng),
```
SergeyKopienko, line 427: May be better to use `uint32_t` type?

dmitriy-sobolev: Did you mean `std::uint64_t` since `std::size_t` may be even "smaller" than `std::uint32_t` above?

SergeyKopienko: Yes, exactly, i believe then `std::uint64_t` for cases when `std::uint32_t` is not enough.

dmitriy-sobolev: Done

SergeyKopienko (dismissed):
LGTM

akukanov:
Is it critical for performance that data-per-workitem is a template parameter of `__leaf_sorter` and `__group_merge_path_sorter`, instead of a runtime parameter? I am concerned about the proliferation of kernel instantiations, and the only reason why leaf sorters have different C++ type is that parameter.

dmitriy-sobolev:
> Is it critical for performance that data-per-workitem is a template parameter of `__leaf_sorter` and `__group_merge_path_sorter`, instead of a runtime parameter? I am concerned about the proliferation of kernel instantiations, and the only reason why leaf sorters have different C++ type is that parameter.

No it is not. As far as I remember it contributes to ~3% of better performance. Probably, it is better to get rid of it for now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -57,6 +56,152 @@ struct __leaf_sort_kernel
     }
 };
 
+struct __group_merge_path_sorter
+{
+    template <typename _StorageAcc, typename _Compare>
+    bool
+    sort(const sycl::nd_item<1>& __item, const _StorageAcc& __storage_acc, _Compare __comp, std::uint32_t __start,
```
SergeyKopienko, line 63: Let's declare not changed params as const?

dmitriy-sobolev: Done

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
dmitriy-sobolev: We discussed it offline and came to an agreement that it was optional.

mmichel11 (approved):
LGTM

PR #1736, opened 2024-07-29
Single WG implementation of `__parallel_find_or`

SergeyKopienko:
In this PR we prepare single work-group implementation of `__parallel_find_or` :
- they doesn't used `atomic`-based synchronization;
- they doesn't used `syc::buffer` for return result when USM-memory is available on device (we using `__result_and_scratch_storage` prepared by @julianmi earlier);
- kernel's compilation has been removed from `__parallel_find_or` and their staff.

This approach gives us a big performance boost for small data sizes.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1332,13 +1332,6 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size);
     __n_groups = ::std::min(__n_groups, decltype(__n_groups)(__max_cu));
 
-    // Pass all small data into single WG implementation
-    constexpr std::size_t __max_iters_per_work_item = 32;
-    if (__rng_n <= __wgroup_size * __max_iters_per_work_item)
-    {
-        __n_groups = 1;
-    }
-
     _PRINT_INFO_IN_DEBUG_MODE(__exec, __wgroup_size, __max_cu);
 
     using _AtomicType = typename _BrickTag::_AtomicType;
@@ -1346,6 +1339,13 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     const auto __pred = oneapi::dpl::__par_backend_hetero::__early_exit_find_or<_ExecutionPolicy, _Brick>{__f};
 
     constexpr bool __or_tag_check = std::is_same_v<_BrickTag, __parallel_or_tag>;
+
+    // Pass all small data into single WG implementation
+    constexpr std::size_t __max_iters_per_work_item = 32;
+    if (__rng_n <= __wgroup_size * __max_iters_per_work_item)
+    {
+        __n_groups = 1;
+    }
 
     _AtomicType __result;
     if (__n_groups == 1)
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1203,7 +1203,7 @@ __parallel_find_or_impl_one_wg(oneapi::dpl::__internal::__device_backend_tag, _E
                     __found_local = __dpl_sycl::__reduce_over_group(__item_id.get_group(), __found_local,
                                                                     typename _BrickTag::_LocalResultsReduceOp{});
 
-                // Set local found state value value to global atomic to have correct result
+                // Set local found state value value to global state to have correct result
                 // - do not check __found_local != __init_value due __result_storage isn't initialized with __init_value on host side.
                 if (__local_idx == 0)
                 {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1181,10 +1181,7 @@ __parallel_find_or_impl_one_wg(oneapi::dpl::__internal::__device_backend_tag, _E
                 // Set local found state value value to global state to have correct result
                 if (__local_idx == 0)
                 {
-                    __FoundStateType& __found =
-                        *__result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__result_acc);
-
-                    __found = __found_local;
+                    __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__result_acc)[0] = __found_local;
                 }
             });
     });
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1216,10 +1216,8 @@ __parallel_find_or_impl_one_wg(oneapi::dpl::__internal::__device_backend_tag, _E
             });
     });
 
-    __future __f_obj(__event_id, __result_storage);
-    auto __result = __f_obj.get();
-
-    return __result;
+    // Wait and return result
+    return __future(__event_id, __result_storage).get();
 }
 
 // Base pattern for __parallel_or and __parallel_find. The execution depends on tag type _BrickTag.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1190,10 +1190,7 @@ __parallel_find_or_impl_one_wg(oneapi::dpl::__internal::__device_backend_tag, _E
     });
 
     // Wait and return result
-    if (__result_storage.is_USM())
-        __event.wait_and_throw();
-
-    return __result_storage.__get_value();
+    return __future(__event, __result_storage).get();
 }
 
 // Base pattern for __parallel_or and __parallel_find. The execution depends on tag type _BrickTag.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -196,6 +196,9 @@ template <typename... _Name>
 class __scan_group_kernel;
 
 template <typename... _Name>
+class __find_or_kernel_one_wg;
+
+template <typename... _Name>
 class __find_or_kernel;
 
 template <typename... _Name>
@@ -1137,49 +1140,112 @@ struct __early_exit_find_or
 //------------------------------------------------------------------------
 
 // Base pattern for __parallel_or and __parallel_find. The execution depends on tag type _BrickTag.
-template <typename _ExecutionPolicy, typename _Brick, typename _BrickTag, typename... _Ranges>
-::std::conditional_t<
-    ::std::is_same_v<_BrickTag, __parallel_or_tag>, bool,
-    oneapi::dpl::__internal::__difference_t<typename oneapi::dpl::__ranges::__get_first_range_type<_Ranges...>::type>>
-__parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Brick __f,
-                   _BrickTag __brick_tag, _Ranges&&... __rngs)
+template <typename KernelName, bool __or_tag_check, typename _ExecutionPolicy, typename _BrickTag,
+#if _ONEDPL_COMPILE_KERNEL && _ONEDPL_KERNEL_BUNDLE_PRESENT
+          typename _Kernel,
+#endif
+          typename _AtomicType, typename _Predicate,
+          typename... _Ranges>
+_AtomicType
+__parallel_find_or_impl_one_wg(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec,
+                               _BrickTag __brick_tag,
+#if _ONEDPL_COMPILE_KERNEL && _ONEDPL_KERNEL_BUNDLE_PRESENT
+                               _Kernel& __kernel,
+#endif
+                               const std::size_t __rng_n, const std::size_t __wgroup_size,
+                               const _AtomicType __init_value, _Predicate __pred,
+                               _Ranges&&... __rngs)
 {
-    using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
-    using _AtomicType = typename _BrickTag::_AtomicType;
-    using _FindOrKernel =
-        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<__find_or_kernel, _CustomName, _Brick,
-                                                                               _BrickTag, _Ranges...>;
-    constexpr bool __or_tag_check = ::std::is_same_v<_BrickTag, __parallel_or_tag>;
+    const std::size_t __n_groups = 1;
 
+    // We shouldn't have any restrictions for _AtomicType type here
+    // because we have a single work-group and we don't need to use atomics for inter-work-group communication.
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Starts main work...
+
+    using __result_and_scratch_storage_t = __result_and_scratch_storage<_ExecutionPolicy, _AtomicType>;
+    __result_and_scratch_storage_t __result_storage(__exec, 0);
+
+    // main parallel_for
+    auto __event_id = __exec.queue().submit([&](sycl::handler& __cgh) {
+        oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
+        auto __result_acc = __result_storage.__get_result_acc(__cgh);
+
+#if _ONEDPL_COMPILE_KERNEL && _ONEDPL_KERNEL_BUNDLE_PRESENT
+        __cgh.use_kernel_bundle(__kernel.get_kernel_bundle());
+#endif
+        __cgh.parallel_for<KernelName>(
+#if _ONEDPL_COMPILE_KERNEL && !_ONEDPL_KERNEL_BUNDLE_PRESENT
+            __kernel,
+#endif
+            sycl::nd_range</*dim=*/1>(sycl::range</*dim=*/1>(__n_groups * __wgroup_size),
+                                        sycl::range</*dim=*/1>(__wgroup_size)),
+            [=](sycl::nd_item</*dim=*/1> __item_id) {
+                auto __local_idx = __item_id.get_local_id(0);
+
+                // 1. Set initial value to local found state
+                _AtomicType __found_local = __init_value;
+
+                // 2. Find any element that satisfies pred
+                //  - after this call __found_local may still have initial value:
+                //    1) if no element satisfies pred;
+                //    2) early exit from sub-group occurred: in this case the state of __found_local will updated in the next group operation (3)
+                __pred(__item_id, __rng_n, __n_groups * __wgroup_size, __found_local, __brick_tag, __rngs...);
+
+                // 3. Reduce over group: find __dpl_sycl::__minimum (for the __parallel_find_forward_tag),
+                // find __dpl_sycl::__maximum (for the __parallel_find_backward_tag)
+                // or update state with __dpl_sycl::__any_of_group (for the __parallel_or_tag)
+                // inside all our group items
+                if constexpr (__or_tag_check)
+                    __found_local = __dpl_sycl::__any_of_group(__item_id.get_group(), __found_local);
+                else
+                    __found_local = __dpl_sycl::__reduce_over_group(__item_id.get_group(), __found_local,
+                                                                    typename _BrickTag::_LocalResultsReduceOp{});
+
+                // Set local found state value value to global atomic to have correct result
+                // - do not check __found_local != __init_value due __result_storage isn't initialized with __init_value on host side.
+                if (__local_idx == 0)
+                {
+                    _AtomicType& __found =
+                        *__result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__result_acc);
+
+                    // Update global (for all groups) non-atomic state with the found index
+                    __found = __found_local;
+                }
+            });
+    });
+
+    __future __f_obj(__event_id, __result_storage);
+    auto __result = __f_obj.get();
+
+    return __result;
+}
+
+// Base pattern for __parallel_or and __parallel_find. The execution depends on tag type _BrickTag.
+template <typename KernelName, bool __or_tag_check, typename _ExecutionPolicy, typename _BrickTag,
+#if _ONEDPL_COMPILE_KERNEL && _ONEDPL_KERNEL_BUNDLE_PRESENT
+          typename _Kernel,
+#endif
+          typename _AtomicType, typename _Predicate,
+          typename... _Ranges>
+_AtomicType
+__parallel_find_or_impl(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec,
+                        _BrickTag __brick_tag,
+#if _ONEDPL_COMPILE_KERNEL && _ONEDPL_KERNEL_BUNDLE_PRESENT
+                        _Kernel& __kernel,
+#endif
+                        const std::size_t __rng_n, const std::size_t __n_groups, const std::size_t __wgroup_size,
+                        const _AtomicType __init_value, _Predicate __pred,
+                        _Ranges&&... __rngs)
+{
     assert("This device does not support 64-bit atomics" &&
            (sizeof(_AtomicType) < 8 || __exec.queue().get_device().has(sycl::aspect::atomic64)));
 
-    auto __rng_n = oneapi::dpl::__ranges::__get_first_range_size(__rngs...);
-    assert(__rng_n > 0);
+    ///////////////////////////////////////////////////////////////////////////
+    // Starts main work...
 
-    // TODO: find a way to generalize getting of reliable work-group size
-    std::size_t __wgroup_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
-#if _ONEDPL_COMPILE_KERNEL
-    auto __kernel = __internal::__kernel_compiler<_FindOrKernel>::__compile(__exec);
-    __wgroup_size = ::std::min(__wgroup_size, oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel));
-#endif
-
-#if _ONEDPL_FPGA_EMU
-    // Limit the maximum work-group size to minimize the cost of work-group reduction.
-    // Limiting this also helps to avoid huge work-group sizes on some devices (e.g., FPGU emulation).
-    __wgroup_size = std::min(__wgroup_size, (std::size_t)2048);
-#endif
-    auto __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
-
-    auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size);
-    __n_groups = ::std::min(__n_groups, decltype(__n_groups)(__max_cu));
-
-    _PRINT_INFO_IN_DEBUG_MODE(__exec, __wgroup_size, __max_cu);
-
-    const _AtomicType __init_value = _BrickTag::__init_value(__rng_n);
     auto __result = __init_value;
-
-    auto __pred = oneapi::dpl::__par_backend_hetero::__early_exit_find_or<_ExecutionPolicy, _Brick>{__f};
 
     // scope is to copy data back to __result after destruction of temporary sycl:buffer
     {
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
+                               const std::size_t __rng_n, const std::size_t __wgroup_size,
+                               const _AtomicType __init_value, _Predicate __pred, _Ranges&&... __rngs)
 {
-    using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
-    using _AtomicType = typename _BrickTag::_AtomicType;
-    using _FindOrKernel =
-        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<__find_or_kernel, _CustomName, _Brick,
-                                                                               _BrickTag, _Ranges...>;
-    constexpr bool __or_tag_check = ::std::is_same_v<_BrickTag, __parallel_or_tag>;
+    const std::size_t __n_groups = 1;
```
julianmi, line 1153: Why don't we remove this throughout the single WG implementation? The compiler will probably optimize this out, but it would shorten the code.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
 
-    assert("This device does not support 64-bit atomics" &&
-           (sizeof(_AtomicType) < 8 || __exec.queue().get_device().has(sycl::aspect::atomic64)));
+    // We shouldn't have any restrictions for _AtomicType type here
```
julianmi, line 1155: Should we move this comment to the main routine to make clear how the single WG implementation works? This would probably follow the reader.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
+    // because we have a single work-group and we don't need to use atomics for inter-work-group communication.
 
-    auto __rng_n = oneapi::dpl::__ranges::__get_first_range_size(__rngs...);
-    assert(__rng_n > 0);
+    ///////////////////////////////////////////////////////////////////////////
+    // Starts main work...
```
julianmi, line 1159: This looks misplaced and you have a `main parallel_for` comment below. I would remove this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
+                // or update state with __dpl_sycl::__any_of_group (for the __parallel_or_tag)
+                // inside all our group items
+                if constexpr (__or_tag_check)
+                    __found_local = __dpl_sycl::__any_of_group(__item_id.get_group(), __found_local);
+                else
+                    __found_local = __dpl_sycl::__reduce_over_group(__item_id.get_group(), __found_local,
+                                                                    typename _BrickTag::_LocalResultsReduceOp{});
+
+                // Set local found state value value to global state to have correct result
+                // - do not check __found_local != __init_value due __result_storage isn't initialized with __init_value on host side.
```
julianmi, line 1205: These look outdated. I would remove them.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
+                //  - after this call __found_local may still have initial value:
+                //    1) if no element satisfies pred;
+                //    2) early exit from sub-group occurred: in this case the state of __found_local will updated in the next group operation (3)
+                __pred(__item_id, __rng_n, __iters_per_work_item, __n_groups * __wgroup_size, __found_local,
+                       __brick_tag, __rngs...);
+
+                // 3. Reduce over group: find __dpl_sycl::__minimum (for the __parallel_find_forward_tag),
+                // find __dpl_sycl::__maximum (for the __parallel_find_backward_tag)
+                // or update state with __dpl_sycl::__any_of_group (for the __parallel_or_tag)
+                // inside all our group items
```
julianmi, line 1197: This comment looks outdated. I would remove it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
+                                                                    typename _BrickTag::_LocalResultsReduceOp{});
+
+                // Set local found state value value to global state to have correct result
+                // - do not check __found_local != __init_value due __result_storage isn't initialized with __init_value on host side.
+                if (__local_idx == 0)
+                {
+                    _AtomicType& __found =
+                        *__result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__result_acc);
+
+                    // Update global (for all groups) non-atomic state with the found index
```
julianmi, line 1211: This comment looks outdated. I would remove it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
+                    __found = __found_local;
+                }
+            });
+    });
 
-    auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size);
-    __n_groups = ::std::min(__n_groups, decltype(__n_groups)(__max_cu));
+    // Wait and return result
+    return __future(__event_id, __result_storage).get();
```
julianmi, line 1218: Why do we create a future here? Can't we just wait on the event and call `__get_value` on `__result_storage`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
 
-#if _ONEDPL_FPGA_EMU
-    // Limit the maximum work-group size to minimize the cost of work-group reduction.
-    // Limiting this also helps to avoid huge work-group sizes on some devices (e.g., FPGU emulation).
-    __wgroup_size = std::min(__wgroup_size, (std::size_t)2048);
+    // Calculate the number of elements to be processed by each work-item.
+    const auto __iters_per_work_item = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __n_groups * __wgroup_size);
+
+    // main parallel_for
+    auto __event_id = __exec.queue().submit([&](sycl::handler& __cgh) {
```
julianmi, line 1168: `__event_id` suggests to me that this is the event's index instead of the event. I would simply name it `__event`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
+#endif
+                                     const std::size_t __rng_n, const std::size_t __n_groups,
+                                     const std::size_t __wgroup_size, const _AtomicType __init_value, _Predicate __pred,
+                                     _Ranges&&... __rngs)
+{
+    assert("This device does not support 64-bit atomics" &&
+           (sizeof(_AtomicType) < 8 || __exec.queue().get_device().has(sycl::aspect::atomic64)));
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Starts main work...
```
julianmi, line 1241: Same comment about removing this since I don't think it adds value to the reader being placed right in the beginning.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1235,13 +1300,105 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+
+    // TODO: find a way to generalize getting of reliable work-group size
+    std::size_t __wgroup_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
+#if _ONEDPL_COMPILE_KERNEL
+    auto __kernels = __internal::__kernel_compiler<_FindOrKernelOneWG, _FindOrKernel>::__compile(__exec);
+    auto __kernel_one_wg = __kernels[0];
+    auto __kernel_many_wg = __kernels[1];
+    __wgroup_size = std::min({__wgroup_size, oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel_one_wg),
+                              oneapi::dpl::__internal::__kernel_work_group_size(__exec, __kernel_many_wg)});
+#endif
```
julianmi, line 1339: Extending my comment about compiling the kernels: I would be interested to see if we even need it for the multi WG implementation since we don't use shared resources.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1235,13 +1300,105 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Calculates the amount of work-groups taking into account the required number of elements per work-item
+    const auto __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
+    auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size);
+    __n_groups = ::std::min(__n_groups, decltype(__n_groups)(__max_cu));
+
+    // Pass all small data into single WG implementation
+    constexpr std::size_t __max_iters_per_work_item = 32;
+    if (__rng_n <= __wgroup_size * __max_iters_per_work_item)
```
julianmi, line 1355: Should we move this done to line 1371 for brevity?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1235,13 +1300,105 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+    using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
+    using _FindOrKernelOneWG =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<__find_or_kernel_one_wg, _CustomName,
+                                                                               _Brick, _BrickTag, _Ranges...>;
+    using _FindOrKernel =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<__find_or_kernel, _CustomName, _Brick,
+                                                                               _BrickTag, _Ranges...>;
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Calculate source data size
```
julianmi, line 1323: I would remove this and similar comments since they weren't there before and these are typical patterns in our code.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1235,13 +1300,105 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+
+    auto __rng_n = oneapi::dpl::__ranges::__get_first_range_size(__rngs...);
+    assert(__rng_n > 0);
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Calculate work-group size
```
julianmi, line 1329: Same about removing this comment.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1235,13 +1300,105 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+#endif
+
+#if _ONEDPL_FPGA_EMU
+    // Limit the maximum work-group size to minimize the cost of work-group reduction.
+    // Limiting this also helps to avoid huge work-group sizes on some devices (e.g., FPGU emulation).
+    __wgroup_size = std::min(__wgroup_size, (std::size_t)2048);
+#endif
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Calculates the amount of work-groups taking into account the required number of elements per work-item
```
julianmi, line 1348: Same about removing this comment.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1235,13 +1300,105 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+    constexpr std::size_t __max_iters_per_work_item = 32;
+    if (__rng_n <= __wgroup_size * __max_iters_per_work_item)
+    {
+        __n_groups = 1;
+    }
+
+    _PRINT_INFO_IN_DEBUG_MODE(__exec, __wgroup_size, __max_cu);
+
+    ///////////////////////////////////////////////////////////////////////////
+    // Eval initial value and create predicate
```
julianmi, line 1363: Same about removing this comment.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1235,13 +1300,105 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+    _AtomicType __result;
+    if (__n_groups == 1)
+    {
+        // Single WG implementation
+        __result = __parallel_find_or_impl_one_wg<_FindOrKernelOneWG, __or_tag_check>(
+            oneapi::dpl::__internal::__device_backend_tag{}, std::forward<_ExecutionPolicy>(__exec), __brick_tag,
+#if _ONEDPL_COMPILE_KERNEL && _ONEDPL_KERNEL_BUNDLE_PRESENT
+            __kernel_one_wg,
+#endif
+            __rng_n, __wgroup_size, // We shouldn't pass __n_groups to this call due it's single WG implementation
```
julianmi, line 1379: I would remove this comment. I think it's very clear at this point. Instead, I would add some comments about the single and multi WG implementations above and below.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Removed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Why?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1081,15 +1084,11 @@ struct __early_exit_find_or
 
     template <typename _NDItemId, typename _SrcDataSize, typename _IterationDataSize, typename _LocalFoundState,
               typename _BrickTag, typename... _Ranges>
-    void
+    inline void
```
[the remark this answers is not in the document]
SergeyKopienko: Perf looks equal, this change has been rolled back.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
julianmi: I misread the comment in line 1197. I would still suggest removing this line here and in the multi WG kernel since we explain that these are group operations already in line 1194 and 1196.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Is it really significant?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Who will call `sycl::queue::wait()` method in this case?

julianmi: The caller of `__get_value` needs to ensure that the result is ready. See the comment in https://github.com/oneapi-src/oneDPL/blob/825ba2f02fe8110e93afa1a49c7292d646851690/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h#L628-L629.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,114 @@ struct __early_exit_find_or
...
 
 // Base pattern for __parallel_or and __parallel_find. The execution depends on tag type _BrickTag.
-template <typename _ExecutionPolicy, typename _Brick, typename _BrickTag, typename... _Ranges>
-::std::conditional_t<
-    ::std::is_same_v<_BrickTag, __parallel_or_tag>, bool,
-    oneapi::dpl::__internal::__difference_t<typename oneapi::dpl::__ranges::__get_first_range_type<_Ranges...>::type>>
-__parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Brick __f,
-                   _BrickTag __brick_tag, _Ranges&&... __rngs)
+template <typename KernelName, bool __or_tag_check, typename _ExecutionPolicy, typename _BrickTag,
+#if _ONEDPL_COMPILE_KERNEL && _ONEDPL_KERNEL_BUNDLE_PRESENT
```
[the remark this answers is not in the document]
SergeyKopienko: Kernel's compilation for `__parallel_find_or` and their staff has been removed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: How our __future::get() is implemented:
```C++
    auto
    get()
    {
        if constexpr (sizeof...(_Args) > 0)
        {
            auto& __val = std::get<0>(*this);
            return __wait_and_get_value(__val);
        }
        else
            wait();
    }
```
 
How __wait_and_get_value is implemented:
```C++
    constexpr auto
    __wait_and_get_value(__result_and_scratch_storage<_ExecutionPolicy, _T>& __storage)
    {
        if (__storage.is_USM())
            wait();
        return __storage.__get_value();
    }
```
 
So for me this mean that we really should call wait() only for USM memory.

@julianmi, am I correct?

julianmi: Yes, we only need explicit synchronization when using USM. I think we should remove the future since it is not used and change this line to `if (__result_storage.is_USM()) wait(); __result_storage.__get_value();`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
julianmi: No but it would shorten the code. Your decision.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,76 @@ struct __early_exit_find_or
...
+                    __found_local = __dpl_sycl::__reduce_over_group(__item_id.get_group(), __found_local,
+                                                                    typename _BrickTag::_LocalResultsReduceOp{});
+
+                // Set local found state value value to global state to have correct result
+                if (__local_idx == 0)
+                {
+                    _AtomicType& __found =
+                        *__result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__result_acc);
+
+                    __found = __found_local;
```
danhoeflinger, line 1187: I don't know if this works if `__result_and_scratch_storage_t` is in buffer storage mode.  This would return an accessor, which doesn't overload the `operator*()`, I believe.

I think the better way is to not dereference, but rather assign to an `auto` typed variable, then assign `__found_local` to the zeroth element of the accessor / usm ptr.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,49 +1136,76 @@ struct __early_exit_find_or
...
 
 // Base pattern for __parallel_or and __parallel_find. The execution depends on tag type _BrickTag.
-template <typename _ExecutionPolicy, typename _Brick, typename _BrickTag, typename... _Ranges>
-::std::conditional_t<
-    ::std::is_same_v<_BrickTag, __parallel_or_tag>, bool,
-    oneapi::dpl::__internal::__difference_t<typename oneapi::dpl::__ranges::__get_first_range_type<_Ranges...>::type>>
-__parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Brick __f,
-                   _BrickTag __brick_tag, _Ranges&&... __rngs)
+template <typename KernelName, bool __or_tag_check, typename _ExecutionPolicy, typename _BrickTag, typename _AtomicType,
+          typename _Predicate, typename... _Ranges>
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed: renamed to `__FoundStateType`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: As I may see below, the function `__get_usm_or_buffer_accessor_ptr` always return pointer:
```C++
    template <typename _Acc>
    static auto
    __get_usm_or_buffer_accessor_ptr(const _Acc& __acc, ::std::size_t __scratch_n = 0)
    {
        return __acc.__get_pointer();
#else
        return &__acc[__scratch_n];
#endif
    }
```

So we simple apply `operator*` to this pointer and save result into reference:
```C++
                    __FoundStateType& __found =
                        *__result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__result_acc);
```

Looks like this code hasn't problems here...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Let's save as is.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
julianmi: I would suggest adding a `__result_and_scratch_storage.__wait_and_get_value` with this logic and use it here and in `__future.__wait_and_get_value` to not duplicate code.

SergeyKopienko: Should this new implementation wait under `ONEDPL_ALLOW_DEFERRED_WAITING` macro check like in the `__future` struct?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
julianmi: I agree with Dan. Assigning the value to the zeroth element seems the safer choice here and is used in our existing code already. If you want to stick with this code, please test it with the buffer mode by setting `_ONEDPL_SYCL_UNIFIED_USM_BUFFER_PRESENT` to false or use an older compiler. Another path you would need to try is setting `__use_USM_host` and `__supports_USM_device` to false in `__result_and_scratch_storage`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: > I would suggest adding a `__result_and_scratch_storage.__wait_and_get_value` with this logic and use it here and in `__future.__wait_and_get_value` to not duplicate code.

done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: `sycl::accessor` based on `sycl::buffer` return in this case `reference` type.
After that inside the method
```C++
    auto
    __get_pointer() const // should be cached within a kernel
    {
        return __usm ? __ptr + __offset : &__acc[__offset];
    }
```
we return address from this reference.
So I think it's look ok too.

SergeyKopienko: About the other our cases, for example `__work_group_reduce_kernel` - there is double referencing: we pass pointer into `__work_group_reduce_kernel` by const reference, names it as `__res_acc` and then, as you wrote, assign value to in through `[0]'. For me this practice isn't very good.

SergeyKopienko: Fixed: I changed this place for compatibility with other code, but not because it's incorrect.

SergeyKopienko:
@julianmi , @danhoeflinger , @adamfidel have somebody additional comments for this PR? Looks like all comments has been fixed now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Ah, I guess I misunderstood how `__get_usm_or_buffer_accessor_ptr()` worked, but your logic is correct I think @SergeyKopienko . 

The naming of the function gives it away I suppose, a pointer to a buffer accessor.   Its just not a normal thing to do, but it does allow the return to be treated equivalently to the usm pointer, I guess.

I think its more clear as you have it written now, but I'm OK with either option.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1235,13 +1253,84 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+    const auto __pred = oneapi::dpl::__par_backend_hetero::__early_exit_find_or<_ExecutionPolicy, _Brick>{__f};
+
+    constexpr bool __or_tag_check = std::is_same_v<_BrickTag, __parallel_or_tag>;
+
+    // Pass all small data into single WG implementation
+    constexpr std::size_t __max_iters_per_work_item = 32;
+    if (__rng_n <= __wgroup_size * __max_iters_per_work_item)
+    {
+        __n_groups = 1;
+    }
```
julianmi, line 1304: Updated comment https://github.com/oneapi-src/oneDPL/pull/1736#discussion_r1697050711. Should we remove this and integrate the if-statement into line 1307 for brevity?

danhoeflinger (dismissed):
LGTM, please come to some agreement with @julianmi on his comment before merging, I don't have a strong preference there.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: I prefer to save this as is because we should have two separate steps
1) tune amount of groups
2) run one of two implementations.
I prefer to have separate steps and do not mix them in one condition.

julianmi: We discussed to move this to where we define the launch configuration based on Sergey's preference.

julianmi (approved):
LGTM. Please wait for green CI.

PR #1737, opened 2024-07-30
[RNG] Fixed failures in ranlux_24_48_base and ranlux_24_48 tests of subtract_with_carry and discard_block engines

egrabovskaya:
Since `operator>>` takes the engine with template parameters for `subtract_with_carry_engine` and `discard_block_engine` classes we have multiple competing definitions in the same namespace when 2 or more of these engine types are instantiated.
Therefore, it's necessary to move the definition outside of the class, then it's only defined once, and add its declaration before the class definition.

### DIFF
--- a/include/oneapi/dpl/internal/random_impl/discard_block_engine.h
+++ b/include/oneapi/dpl/internal/random_impl/discard_block_engine.h
@@ -35,6 +35,14 @@ template <class _Engine, ::std::size_t _P, ::std::size_t _R>
 class discard_block_engine;
 
 template <class CharT, class Traits, class __Engine, std::size_t __P, std::size_t __R>
+::std::basic_ostream<CharT, Traits>&
+operator<<(::std::basic_ostream<CharT, Traits>&, const discard_block_engine<__Engine, __P, __R>&);
+
+template <class __Engine, std::size_t __P, std::size_t __R>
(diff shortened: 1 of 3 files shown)


include/oneapi/dpl/internal/random_impl/discard_block_engine.h
```diff
@@ -166,22 +173,7 @@ class discard_block_engine
...
-        std::size_t __n_;
-        if (__is >> __engine_ >> __n_)
-        {
-            __e.engine_ = __engine_;
-            __e.n_ = __n_;
-        }
-
-        return __is;
-    }
+    operator>>(::std::basic_istream<CharT, Traits>&, discard_block_engine<__Engine, __P, __R>&);
```
egrabovskaya, line 176: Should I also move definitions of output operators and !=, == outside of the discard_block_engine and subtract_with_carry_engine classes? And should I do that for linear_congruential_engine as well?

aelizaro: yes, please, align operators>> and << definitions. (== and != should be ok to leave as is)

aelizaro (approved):
LGTM, thank you!

timmiesmith:
Thank you for the quick work on this issue!

PR #1742, opened 2024-07-30
Extend RFC announcement to include discussions

timmiesmith:
The PR updates the action that currently announces PR labeled as RFC to the oneDPL Slack channel to include discussions labeled as RFC as well.


timmiesmith:
I'm using https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#discussion to identify the keywords needed.

danhoeflinger:
It looks like the action checks `github.event.pull_request.labels.*.name`, for RFC, but is this the same variable we should be checking for discussions as well?
I assume we need a second check or to modify this one to cover `github.event.discussion.labels.*.name`  (guessing the formulation of this).

timmiesmith:
goofed the rebase. I'm closing this PR and will open another.

PR #1746, opened 2024-07-31
distributed_device_policy renamed to sycl_device_collection

mateuszpn:


include/oneapi/dpl/internal/distributed_ranges_impl/sp/init.hpp
```diff
@@ -82,7 +82,7 @@ nprocs()
     return __detail::ngpus();
 }
 
-inline distributed_device_policy par_unseq;
+inline sycl_device_collection par_unseq;
```
akukanov, line 85: I still do not like `par_unseq` as the policy name, finding it misleading.

I can see why you may want it to make the use of DR algorithms similar to the standard ones. But this is only syntactic similarity, while semantically the differences are rather broad. Also, I am afraid it will get confused with `oneapi::dpl::execution::par_unseq` which closely follows the standard semantics.

My comment https://github.com/oneapi-src/oneDPL/pull/1479#discussion_r1686750865 was admittedly not quite clear, sorry if it mislead you. But I actually meant the `par_unseq` name of the object, not the type.

How about `struct device_set_policy` and `inline device_set_policy preset_devices`?

@lslusarczyk @BenBrock @rscohn2

akukanov: Moved this discussion to the main PR https://github.com/oneapi-src/oneDPL/pull/1479#discussion_r1700286815

PR #1747, opened 2024-07-31
Fix an error in the oneDPL `__future` implementation

SergeyKopienko:
In this PR we fix an error in the oneDPL `__future` implementation.

Let's take look to current implementation of `__future::wait()` function :
```C++
    wait()
    {
#if !ONEDPL_ALLOW_DEFERRED_WAITING
        __my_event.wait_and_throw();
#endif
    }
```
This mean that real call of `wait_and_throw()` will depends on the state of `ONEDPL_ALLOW_DEFERRED_WAITING` macro (what is this and for that - described at https://oneapi-src.github.io/oneDPL/macros.html).

Now let's take a look to implementation of `__future::__wait_and_get_value()` function :
```C++
    template <typename _T>
    constexpr auto
    __wait_and_get_value(_T& __val)
    {
        wait();
        return __val;
    }
```

So at first - we waits and then return value. But this mean that real call of `wait_and_throw()` will depends on the state of `ONEDPL_ALLOW_DEFERRED_WAITING` macro too and we may return value without real `wait_and_throw()` call.

### DIFF
--- a/documentation/library_guide/macros.rst
+++ b/documentation/library_guide/macros.rst
@@ -103,8 +103,8 @@ Macro                              Description
                                    a device policy might return before the computation completes on the device.
 
                                    .. Warning:: Before accessing data produced or modified by the call, waiting
-                                   for completion of all tasks in the corresponding SYCL queue is required;
-                                   otherwise, the program behavior is undefined.
+                                      for completion of all tasks in the corresponding SYCL queue is required;
+                                      otherwise, the program behavior is undefined.
 ---------------------------------- ------------------------------
 ``ONEDPL_FPGA_DEVICE``             Use this macro to build your code containing |onedpl_short| parallel
                                    algorithms for FPGA devices. (Disabled by default.)
--- a/documentation/library_guide/macros.rst
+++ b/documentation/library_guide/macros.rst
@@ -98,6 +98,13 @@ Macro                              Description
 ---------------------------------- ------------------------------
 ``ONEDPL_ALLOW_DEFERRED_WAITING``  This macro allows waiting for completion of certain algorithms executed with
                                    device policies to be deferred. (Disabled by default.)
+
+                                   When the macro evaluates to non-zero, a call to a oneDPL algorithm with
+                                   a device policy might return before the computation completes on the device.
+
+                                   .. Warning:: Before accessing data produced or modified by the call, waiting
+                                   for completion of all tasks in the corresponding SYCL queue is required;
+                                   otherwise, the program behavior is undefined.
 ---------------------------------- ------------------------------
 ``ONEDPL_FPGA_DEVICE``             Use this macro to build your code containing |onedpl_short| parallel
                                    algorithms for FPGA devices. (Disabled by default.)
--- a/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
+++ b/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
@@ -48,8 +48,7 @@ __pattern_walk1_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
     return __future_obj;
 }
 
-template <typename _WaitMode = __par_backend_hetero::__async_mode,
-          __par_backend_hetero::access_mode __acc_mode1 = __par_backend_hetero::access_mode::read,
+template <__par_backend_hetero::access_mode __acc_mode1 = __par_backend_hetero::access_mode::read,
           __par_backend_hetero::access_mode __acc_mode2 = __par_backend_hetero::access_mode::write,
           typename _BackendTag, typename _ExecutionPolicy, typename _ForwardIterator1, typename _ForwardIterator2,
           typename _Function>
@@ -69,9 +68,6 @@ __pattern_walk2_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
     auto __future = oneapi::dpl::__par_backend_hetero::__parallel_for(
         _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
-
-    // Call no wait, wait or deferrable wait depending on _WaitMode
-    __future.wait(_WaitMode{});
 
     return __future.__make_future(__first2 + __n);
 }
--- a/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
+++ b/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
@@ -71,7 +71,7 @@ __pattern_walk2_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
     if constexpr (_IsSync::value)
-        __future.wait();
+        __future.deferrable_wait();
 
     return __future.__make_future(__first2 + __n);
 }
--- a/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
+++ b/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
@@ -71,7 +71,7 @@ __pattern_walk2_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
     if constexpr (_IsSync::value)
-        __future.deferrable_wait();
+        __future.__deferrable_wait();
 
     return __future.__make_future(__first2 + __n);
 }
--- a/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
+++ b/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
@@ -71,7 +71,7 @@ __pattern_walk2_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
     // Call no wait, wait or deferrable wait depending on _WaitMode
-    oneapi::dpl::__par_backend_hetero::__wait_future_result<_WaitMode>{}(__future);
+    __future.wait(_WaitMode{});
 
     return __future.__make_future(__first2 + __n);
 }
--- a/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
+++ b/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
@@ -71,7 +71,7 @@ __pattern_walk2_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
     // Call no wait, wait or deferrable wait depending on _WaitMode
-    oneapi::dpl::__par_backend_hetero::__wait_future_result<_WaitMode>{}(__future_obj);
+    oneapi::dpl::__par_backend_hetero::__wait_future_result<_WaitMode>{}(__future);
 
     return __future.__make_future(__first2 + __n);
 }
--- a/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
+++ b/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
@@ -48,7 +48,7 @@ __pattern_walk1_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
     return __future_obj;
 }
 
-template <typename _IsSync = ::std::false_type,
+template <typename _WaitMode = __par_backend_hetero::__async_mode,
           __par_backend_hetero::access_mode __acc_mode1 = __par_backend_hetero::access_mode::read,
           __par_backend_hetero::access_mode __acc_mode2 = __par_backend_hetero::access_mode::write,
           typename _BackendTag, typename _ExecutionPolicy, typename _ForwardIterator1, typename _ForwardIterator2,
@@ -70,8 +70,8 @@ __pattern_walk2_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
         _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
-    if constexpr (_IsSync::value)
-        __future.__deferrable_wait();
+    // Call no wait, wait or deferrable wait depending on _WaitMode
+    oneapi::dpl::__par_backend_hetero::__wait_future_result<_WaitMode>{}(__future_obj);
 
     return __future.__make_future(__first2 + __n);
 }
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -75,59 +75,10 @@ __pattern_walk1_n(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _F
 // walk2
 //------------------------------------------------------------------------
 
-// Tag __async_mode describe a pattern call mode which should be executed asynchronously
-struct __async_mode
-{
-};
-// Tag __sync_mode describe a pattern call mode which should be executed synchronously
-struct __sync_mode
-{
-};
-// Tag __deferrable_mode describe a pattern call mode which should be executed
-// synchronously/asynchronously : it's depends on ONEDPL_ALLOW_DEFERRED_WAITING macro state
-struct __deferrable_mode
-{
-};
-
-template <typename _Mode>
-struct __wait_future_result;
-
-template <>
-struct __wait_future_result<__async_mode>
-{
-    template <typename _Future>
-    void
-    operator()(_Future& /*__future*/) const
-    {
-    }
-};
-
-template <>
-struct __wait_future_result<__sync_mode>
-{
-    template <typename _Future>
-    void
-    operator()(_Future& __future) const
-    {
-        __future.wait();
-    }
-};
-
-template <>
-struct __wait_future_result<__deferrable_mode>
-{
-    template <typename _Future>
-    void
-    operator()(_Future& __future) const
-    {
-        __future.__deferrable_wait();
-    }
-};
-
 // TODO: A tag _WaitMode is used for provide a patterns call pipeline, where the last one should be synchronous
 // Probably it should be re-designed by a pipeline approach, when a pattern returns some sync objects
 // and ones are combined into a "pipeline" (probably like Range pipeline)
-template <typename _WaitMode = __deferrable_mode,
+template <typename _WaitMode = __par_backend_hetero::__deferrable_mode,
           __par_backend_hetero::access_mode __acc_mode1 = __par_backend_hetero::access_mode::read,
           __par_backend_hetero::access_mode __acc_mode2 = __par_backend_hetero::access_mode::write,
           typename _BackendTag, typename _ExecutionPolicy, typename _ForwardIterator1, typename _ForwardIterator2,
@@ -151,7 +102,7 @@ __pattern_walk2(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _ForwardIt
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
     // Call no wait, wait or deferrable wait depending on _WaitMode
-    __wait_future_result<_WaitMode>{}(__future_obj);
+    oneapi::dpl::__par_backend_hetero::__wait_future_result<_WaitMode>{}(__future_obj);
 
     return __first2 + __n;
 }
@@ -175,7 +126,8 @@ _ForwardIterator2
 __pattern_swap(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _ForwardIterator1 __first1,
                _ForwardIterator1 __last1, _ForwardIterator2 __first2, _Function __f)
 {
-    return __pattern_walk2</*_WaitMode*/ __deferrable_mode, __par_backend_hetero::access_mode::read_write,
+    return __pattern_walk2</*_WaitMode*/ __par_backend_hetero::__deferrable_mode,
+                           __par_backend_hetero::access_mode::read_write,
                            __par_backend_hetero::access_mode::read_write>(
         __tag, ::std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2, __f);
 }
@@ -298,8 +250,8 @@ __pattern_walk2_transform_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&&
 {
     // Require `read_write` access mode for output sequence to force a copy in for host iterators to capture incoming
     // values of the output sequence for elements where the predicate is false.
-    return __pattern_walk2</*_WaitMode*/ __deferrable_mode, __par_backend_hetero::access_mode::read,
-                           __par_backend_hetero::access_mode::read_write>(
+    return __pattern_walk2</*_WaitMode*/ __par_backend_hetero::__deferrable_mode,
+                           __par_backend_hetero::access_mode::read, __par_backend_hetero::access_mode::read_write>(
         __tag,
         __par_backend_hetero::make_wrapped_policy<__walk2_transform_if_wrapper>(
             ::std::forward<_ExecutionPolicy>(__exec)),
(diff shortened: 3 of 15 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -688,12 +688,20 @@ class __future : private std::tuple<_Args...>
...
+
     void
     wait()
     {
-#if !ONEDPL_ALLOW_DEFERRED_WAITING
         __my_event.wait_and_throw();
+    }
+
+    void
+    deferrable_wait()
```
danhoeflinger, line 701: Do we want users to be encouraged to call a `deferrable_wait()` function?  I would think that is somewhat implementation details or perhaps to convoluted for public consumption.
  
Obviously we need this to be `public`, but should we uglify it maybe to discourage usage outside the library (or would we want some internal version of `__future` even which we don't return to the users?

test/general/header_inclusion_order_async_0.pass.cpp
```diff
@@ -34,7 +34,7 @@ main()
     std::vector<T, allocator> data(n, 1, alloc);
 
     auto f = oneapi::dpl::experimental::reduce_async(TestUtils::make_device_policy(q), data.begin(), data.end());
-    f.wait();
+    f.deferrable_wait();
```
danhoeflinger, line 37: As mentioned above, I'm not sure we want users calling this function, but rather they would be calling `wait()`

test/general/header_inclusion_order_async_1.pass.cpp
```diff
@@ -34,7 +34,7 @@ main()
     std::vector<T, allocator> data(n, 1, alloc);
 
     auto f = oneapi::dpl::experimental::reduce_async(TestUtils::make_device_policy(q), data.begin(), data.end());
-    f.wait();
+    f.deferrable_wait();
```
danhoeflinger, line 37: Same as above, users probably should call `wait()`

danhoeflinger (commented):
I'm wondering if this macro is still used by anyone, and if we still need / want to support it.  It seems like the more appropriate venue for something like this is to use the async API.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: Our `__future` implemented in the namespace 
```C++
namespace oneapi
{
namespace dpl
{
namespace __internal
{
```
This mean (for me) that user will never works with our `__future` implementation directly.

test/general/header_inclusion_order_async_0.pass.cpp
SergeyKopienko: @akukanov how do you think, which call we should use in this case?

test/general/header_inclusion_order_async_1.pass.cpp
SergeyKopienko: @akukanov how do you think, which call we should use in this case?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
akukanov: Actually they might, if an object of this type is returned by our public API functions. It might be documented as an unspecified type, yet it should be considered as a part of public API. Is that the case with `__future`?

test/general/header_inclusion_order_async_0.pass.cpp
SergeyKopienko: Fixed: restored `wait()` call.

test/general/header_inclusion_order_async_1.pass.cpp
SergeyKopienko: Fixed: restored `wait()` call.

test/general/header_inclusion_order_async_0.pass.cpp
danhoeflinger: I recommend at least uglifying it, but we could consider having an `__internal_future` type with this functionality, and another which we return to users which can be created from an `__internal_future` which does not have this functionality.

julianmi:
> I'm wondering if this macro is still used by anyone, and if we still need / want to support it. It seems like the more appropriate venue for something like this is to use the async API.

I agree with this.

If we want to keep it, I think we should improve the documentation for `ONEDPL_ALLOW_DEFERRED_WAITING` in https://github.com/oneapi-src/oneDPL/blob/main/documentation/library_guide/macros.rst. It is currently very unspecific especially regarding the need for synchronization and potential data issues:
"This macro allows waiting for completion of certain algorithms executed with device policies to be deferred. (Disabled by default.)"

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: Yes, its an unspecified future-like type in our documentation, which lists `wait()` and `get()`.  It of course is not in the spec because it is part of the experimental async API.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -739,7 +739,7 @@ __pattern_reduce_by_segment(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&&
         oneapi::dpl::__ranges::take_view_simple(experimental::ranges::views::all_read(__idx),
                                                 __intermediate_result_end),
         ::std::forward<_Range2>(__values), experimental::ranges::views::all_write(__tmp_out_values))
-        .wait();
+        .__deferrable_wait();
 
     // Round 2: final reduction to get result for each segment of equal adjacent keys
```
akukanov, line 744: I am concerned that deferring the wait might not be appropriate here, in particular if the input for "Round 2" depends on completion of the kernel(s) above it.

test/general/header_inclusion_order_async_0.pass.cpp
SergeyKopienko: Uglified.

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h
```diff
@@ -172,7 +172,7 @@ __pattern_transform_scan_base(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&
         oneapi::dpl::__par_backend_hetero::__parallel_transform_scan(_BackendTag{}, __policy, __buf1.all_view(),
                                                                      __buf2.all_view(), __n, __unary_op, __init,
                                                                      __binary_op, _Inclusive{})
-            .wait();
+            .__deferrable_wait();
 
         // Move data from temporary buffer into results
         oneapi::dpl::__internal::__pattern_walk2_brick(
```
akukanov, line 178: Here I have the same kind of concern as well, as another computation follows the pattern call.

akukanov:
> I'm wondering if this macro is still used by anyone, and if we still need / want to support it. It seems like the more appropriate venue for something like this is to use the async API.

@danhoeflinger, you are right of course. However, the async API is still in preview, and the macro is public so formally needs deprecation.
- 
> If we want to keep it, I think we should improve the documentation for ONEDPL_ALLOW_DEFERRED_WAITING in https://github.com/oneapi-src/oneDPL/blob/main/documentation/library_guide/macros.rst. It is currently very unspecific especially regarding the need for synchronization and potential data issues.

Yes, something might be added along the lines that callers are responsible for waiting for completion on the SYCL queue associated with the execution policy before accessing any data modified by deferred algorithms.

danhoeflinger:
> @danhoeflinger, you are right of course. However, the async API is still experimental, and the macro is public so formally needs deprecation.


Yes, you are right, it would require deprecation.  Should we consider adding deprecation for the next release?  In my opinion the experimental async API is already a much better supported and better documented version of what this macro provides.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h

include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h

akukanov:
I think we should first have a plan for async API productization, then talk about deprecation.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -54,7 +54,7 @@ __pattern_walk1(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _ForwardIt
 
     oneapi::dpl::__par_backend_hetero::__parallel_for(
         _BackendTag{}, __exec, unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf.all_view())
-        .wait();
+        .__deferrable_wait();
```
danhoeflinger, line 57: I cant currently find any problems with this one currently in the repo, but it is to some extent a "building block", which other algorithms may use internally as a piece of a multi-kernel operation which may have internal dependencies.  

This seems like a possible source of errors future the way it is currently implemented.
It may be worth it to do something like what is done in `__patern_walk2` to provide control of `_IsSync` to the caller, then use `ONEDPL_ALLOW_DEFERRED_WAITING` to control that template variable in the cases which make sense.  Alternatively, we could allow algorithms to explicitly set this pattern to synchronous, and otherwise, by default depend on `__deferrable_wait()`.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -102,7 +102,7 @@ __pattern_walk2(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _ForwardIt
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
     if constexpr (_IsSync())
```
danhoeflinger, line 104: `__pattern_partial_sort_copy` uses `__pattern_walk2` in a way which requires synchronousity.  It needs the copy to finish before sort is started on the copied data.    
Also, this function has a template parameter which specifies `_IsSync`, and then we ignore it by using `__deferrable_wait()`.  Should we instead selectively use the macro to to control the `_IsSync` template parameter (only when it is safe to do so)?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -158,7 +158,7 @@ __pattern_walk3(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _ForwardIt
     oneapi::dpl::__par_backend_hetero::__parallel_for(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
                                                       unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n,
                                                       __buf1.all_view(), __buf2.all_view(), __buf3.all_view())
-        .wait();
+        .__deferrable_wait();
```
danhoeflinger, line 161: similar comment to `__pattern_walk1`.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1185,7 +1185,7 @@ __pattern_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _Ite
 
         __par_backend_hetero::__parallel_merge(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
                                                __buf1.all_view(), __buf2.all_view(), __buf3.all_view(), __comp)
-            .wait();
+            .__deferrable_wait();
```
danhoeflinger, line 1188: `__pattern_inplace_merge` requires this pattern to by synchronous, so the merge is finished before the copy.

danhoeflinger (commented):
After investigating a handful of these individual patterns, it makes me question if this is the right strategy.

It should be said that we are not adding errors where things were working before.  This is already broken when the macro is enabled.  However, if we are looking at improving this, we should try to fix these sorts of errors and "traps" for future bugs.  

What we actually want from this macro is optional synchronicity at the API level, but we do not have a consistent standard or expectation for synchronicity at the pattern level, which can be problematic.

@reble This PR is probably something to consider in our look at aligning with the asynchronous API.  Even if it is effectively a more crude precursor to that API.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
akukanov: @SergeyKopienko please pay attention to this.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
akukanov: @SergeyKopienko please take into account.

danhoeflinger:
> @danhoeflinger @julianmi While this patch might still leave some problems unresolved, I think at this time it's an improvement - first, due to limiting the number of places where waiting can be deferred and making these visible and searchable in the code, and second, due to improving the documentation for the macro. Therefore I would still propose to commit it, maybe with adding some TODOs and/or an issue to revisit later.

OK, do we want to try to resolve issues with the individual usages like https://github.com/oneapi-src/oneDPL/pull/1747#discussion_r1710009584 in this PR, or leave those as TODO?  
If we want to try to resolve them, I can keep looking in detail through the rest of the patterns.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: @danhoeflinger, @akukanov Just for note - this error hasn't been introduced in this PR.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: @danhoeflinger please explain which chain of call exactly you mean.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: @akukanov, @danhoeflinger Fixed.

documentation/library_guide/macros.rst
```diff
@@ -98,6 +98,13 @@ Macro                              Description
 ---------------------------------- ------------------------------
 ``ONEDPL_ALLOW_DEFERRED_WAITING``  This macro allows waiting for completion of certain algorithms executed with
                                    device policies to be deferred. (Disabled by default.)
+
+                                   When the macro evaluates to non-zero, a call to a oneDPL algorithm with
+                                   a device policy might return before the computation completes on the device.
+
+                                   .. Warning:: Before accessing data produced or modified by the call, waiting
```
julianmi, line 105: There seems to be a line break in the rendered version between "waiting" and "for".

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -672,27 +672,35 @@ template <typename _Event, typename... _Args>
...
+        auto __val = __storage.__wait_and_get_value(__my_event);
+        __set_waited_for_result();
+        return __val;
     }
 
     template <typename _T>
     constexpr auto
     __wait_and_get_value(const _T& __val)
     {
         wait();
```
julianmi, line 700: Do we need to wait here again even if the user has called `__future.wait()`? Could we instead check for `__waited_for_result` and skip the wait if it is true.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -672,27 +672,35 @@ template <typename _Event, typename... _Args>
...
+        __set_waited_for_result();
+        return __val;
     }
 
     template <typename _ExecutionPolicy, typename _T>
     constexpr auto
     __wait_and_get_value(const __result_and_scratch_storage<_ExecutionPolicy, _T>& __storage)
     {
-        return __storage.__wait_and_get_value(__my_event);
+        auto __val = __storage.__wait_and_get_value(__my_event);
```
julianmi, line 691: Similar comment here about unneeded waits in case we've waited before.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -101,8 +151,8 @@ __pattern_walk2(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _ForwardIt
         _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
-    if constexpr (_IsSync())
-        __future_obj.wait();
+    // Call optional wait: no wait, wait or deferrable wait.
```
julianmi, line 154:
```suggestion
    // Call no wait, wait or deferrable wait depending on _WaitMode
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: My assumption that `__waited_for_result` state is debugging tool only, may be we remove it at all later or place under some `debug` mode macro check. So I believe that we shouldn't use it for choice how to get result and so on. With this tool I am trying to find errors in tests now on `CPU` and on `GPU`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: https://github.com/oneapi-src/oneDPL/pull/1747#discussion_r1715079486

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: Fixed.

documentation/library_guide/macros.rst
SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: UPD: all tests passed on `CPU` and `GPU` devices and no `assert` has been occurred,
I mean `assert` from
```C++
    template <typename _T>
    constexpr auto
    __wait_and_get_value(const _T& __val)
    {
        wait();

        assert(__get_waited_for_result() && "Result was not waited");

        return __val;
    }
```
So we may have hope that the current state of the code is correct.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1510,7 +1559,7 @@ __pattern_partial_sort_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&&
 
         auto __buf_first = __buf.get();
 
-        auto __buf_last = __pattern_walk2</*_IsSync=*/::std::false_type>(
+        auto __buf_last = __pattern_walk2<__sync_mode>(
```
danhoeflinger, line 1562: I believe this can stay `__async_mode` based on the comment below.  
The temporary buffer will order the kernels appropriately, and no explicit wait is required. (Also, since `__buf_last` does not depend on the result of the kernel, it is a fixed increment of `__buf_first`)

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -516,7 +516,7 @@ __pattern_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _Ran
         __par_backend_hetero::__parallel_merge(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
                                                ::std::forward<_Range1>(__rng1), ::std::forward<_Range2>(__rng2),
                                                ::std::forward<_Range3>(__rng3), __comp)
-            .wait();
+            .__deferrable_wait();
```
danhoeflinger, line 519: I originally thought  `__pattern_inplace_merge` could have an issue, but I believe it is saved by the implicit dependency between kernels from the temporary buffer, so I think nothing needs to be done.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: as result, all `waited_for_result` stuff has been removed to reduce the amount of changes.

SergeyKopienko: @julianmi may we resolve this conversation?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: as result, all `waited_for_result` stuff has been removed to reduce the amount of changes.

SergeyKopienko: @julianmi may we resolve this conversation?

include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
```diff
@@ -70,8 +70,8 @@ __pattern_walk2_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
         _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
-    if constexpr (_IsSync::value)
-        __future.wait();
+    // Call no wait, wait or deferrable wait depending on _WaitMode
+    oneapi::dpl::__par_backend_hetero::__wait_future_result<_WaitMode>{}(__future_obj);
```
danhoeflinger, line 74: I believe this is the wrong variable name `__future` vs `__future_obj`

SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -738,6 +744,55 @@ class __future : private std::tuple<_Args...>
...
+
+template <>
+struct __wait_future_result<__async_mode>
+{
+    template <typename _Future>
+    void
+    operator()(_Future& /*__future*/) const
+    {
+    }
+};
```
akukanov, line 772: Do we really need to make this dispatch generic, for an "arbitrary" `_Future` type? Or would it make more sense to have that dispatch being really within the actual future type used in the implementation?

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: We discussed and think its best to leave this for now prior to release, and perhaps make more changes after release to switch some patterns to a system like `__pattern_walk2`.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: We discussed and think its best to leave this for now prior to release, and perhaps make more changes after release to switch some patterns to a system like __pattern_walk2.

danhoeflinger (approved):
Wait for a Green CI, and please resolve the feedback from @akukanov , but overall LGTM.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
akukanov: Something like this, for example (not tested):
```
// within class template __future
template<typename _ModeTag>
void wait(_ModeTag)
{
    if constexpr (std::is_same_v<_ModeTag,__sync_mode>)
        wait();
    else if constexpr (std::is_same_v<_ModeTag,__deferrable_mode>)
        __deferrable_wait();
}
```

include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
```diff
@@ -48,7 +48,7 @@ __pattern_walk1_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
     return __future_obj;
 }
 
-template <typename _IsSync = ::std::false_type,
+template <typename _WaitMode = __par_backend_hetero::__async_mode,
```
akukanov, line 51: Do we really need this tag for `__pattern_walk2_async`? I mean, it is supposed to be asynchronous, is not it? And of course currently it is not used with other than the default template argument.

SergeyKopienko: The source idea was to have the same format for `__pattern_walk2_async` and for `__pattern_walk2`

akukanov: I know; and I know it is not in your patch. But my question is, is it really needed? And we are changing the pattern slightly anyway, so I think it's appropriate to revisit this.

SergeyKopienko: From current use cases state it's really not required.

SergeyKopienko: Fixed: removed.

akukanov (approved):

danhoeflinger (approved):
It would be best to wait for green CI, LGTM.

julianmi (dismissed):

julianmi (approved):

PR #1749, opened 2024-07-31
[PROTOTYPE] Add fallback to legacy scan implementation for CPU devices and devices that lack size 32 sub-groups

mmichel11:
This PR adds a fallback from reduce-then-scan to the legacy scan implementation in two separate cases:

- As a fallback if the device does not support size 32 sub-groups. We explicitly require this size and must fallback if it's not supported.
- For CPU devices. The sub-group level operations do not translate well to CPU devices and the legacy implementation is significantly faster on CPUs.


mmichel11:
Moving back to draft. Some issues occurred when merging with the target branch.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1052,6 +1064,23 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag __backend_tag,
             oneapi::dpl::unseq_backend::__no_init_value<_Size>{},
             /*_Inclusive=*/std::true_type{});
     }
+    else
+    {
+        using _ReduceOp = ::std::plus<_Size>;
```
danhoeflinger, line 1069: `::std`->`std`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -769,6 +770,15 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
     }
 };
 
+template <typename _ExecutionPolicy, typename _SizeType>
+bool
+__supports_sub_group_size(const _ExecutionPolicy& __exec, _SizeType __target_size)
+{
+    const auto __subgroup_sizes = __exec.queue().get_device().template get_info<sycl::info::device::sub_group_sizes>();
+    return std::find(__subgroup_sizes.begin(), __subgroup_sizes.end(), static_cast<std::size_t>(32)) !=
```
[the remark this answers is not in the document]
mmichel11: Thanks, sloppy mistake on my side

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -769,6 +770,15 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
     }
 };
 
+template <typename _ExecutionPolicy, typename _SizeType>
+bool
+__supports_sub_group_size(const _ExecutionPolicy& __exec, _SizeType __target_size)
```
adamfidel, line 775: Since the only usage of `__target_size` in this function casts it to `std::size_t`, would it make sense to take the parameter by `std::size_t` directly instead of using a template?
```suggestion
__supports_sub_group_size(const _ExecutionPolicy& __exec, std::size_t __target_size)
```

danhoeflinger (commented):
one minor nitpick, otherwise, I think LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: Good point. I've added this and removed the `static_cast` below.

danhoeflinger (approved):
LGTM

PR #1751, opened 2024-08-01
Fix double dereference in the function `__work_group_reduce_kernel`

SergeyKopienko:
In this PR I propose the way how to avoid double dereference in Kernel code inside  function `__work_group_reduce_kernel` .
Double dereference has place here in Kernel's code in cases when we pass into this function simple `pointer` : we pass it by reference.

This problem also demonstrated in example, prepared by @julianmi : https://godbolt.org/z/rP7WbW94o
```C++
template <typename _Res>
void __kernel1(const _Res __res_acc){
	//  mov rax, qword ptr [rbp - 8]
	// mov dword ptr [rax], 0
    *__res_acc = 0;
}
 
template <typename _Res>
void __kernel2(const _Res& __res_acc){
    // mov rax, qword ptr [rbp - 8]
    // mov rax, qword ptr [rax]
    // mov dword ptr [rax], 1                  <<< EXTRA INSTRUCTION due we pass pointer by reference
    *__res_acc = 1;
}
 
template <typename _Res>
void __kernel3(const _Res __res_acc){
    // mov rax, qword ptr [rbp - 8]
    // mov dword ptr [rax], 2
    __res_acc[0] = 2;
}
 
template <typename _Res>
void __kernel4(const _Res& __res_acc){
    // mov rax, qword ptr [rbp - 8]
    // mov rax, qword ptr [rax]
    // mov dword ptr [rax], 3                     <<< EXTRA INSTRUCTION  due we pass pointer by reference
    __res_acc[0] = 3;
}
```

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -483,6 +483,8 @@ struct __usm_or_buffer_accessor
     size_t __offset = 0;
 
   public:
+    using value_type = _T;
+
     // Buffer accessor
     __usm_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf)
         : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{}))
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -500,7 +502,7 @@ struct __usm_or_buffer_accessor
     {
     }
 
-    auto
+    _T*
```
SergeyKopienko, line 505: This is not mandatory change, but we know return type here.

danhoeflinger (approved):
LGTM, 
As far as I can tell, it only saves a single instruction per work-item, but it does make things more clear with naming / types, and I think it is correct.

SergeyKopienko:
Yes, exactly: we save only one asm instruction in Kernel code per work-item, but I believe it make sense.

julianmi (approved):
LGTM

PR #1756, opened 2024-08-02
limited CI

lslusarczyk:
temprorarily disabled all tests except DR ones to speed up applying DR review comments

change to be reverted

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,78 +99,78 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - os: ubuntu-latest
-            cxx_compiler: icpx
-            std: 17
-            build_type: release
-            backend: dpcpp
-            device_type: CPU
-          - os: ubuntu-latest
-            cxx_compiler: icpx
-            std: 17
-            build_type: release
-            backend: dpcpp
-            device_type: FPGA_EMU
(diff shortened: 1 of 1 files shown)


akukanov:
@danhoeflinger @dmitriy-sobolev @timmiesmith - is there a "better" way to do what @lslusarczyk wants for CI in the [distributed-ranges](https://github.com/oneapi-src/oneDPL/tree/distributed-ranges) branch?

danhoeflinger:
> @danhoeflinger @dmitriy-sobolev @timmiesmith - is there a "better" way to do what @lslusarczyk wants for CI in the [distributed-ranges](https://github.com/oneapi-src/oneDPL/tree/distributed-ranges) branch?

Umm, this is kind of the "quick and dirty" way of doing it, which I don't really object to.  It makes it reasonably obvious that it should be removed.  

I think there are ways to exclude runs from the matrix based upon branch name matching.  That would allow us to do the same with some fewer lines of code I think.  Is it necessary in this scenario where the intention is to revert this?  
I'm not sure.

Perhaps there is some better way still that I'm missing...

lslusarczyk:
Thanks for looking into this. If we find a better way I will revert this change and disable in a new way. For now I'm merging it.

It has an advantage that it is so dirty that nobody will forget about reverting this change.

PR #1757, opened 2024-08-02
Rework the use of ranges-v3 and remove owning_view

akukanov:
In this patch I propose to support ranges-v3 in a different manner. The common DR code in oneDPL would support redefinition of the `stdrng` namespace alias, as well as the `_ONEDPL_DR_STD_RANGES_NAMESPACE` macro, but adjusting these definitions for ranges-v3 would be done in the "standalone" DR repository, like below:

```
// File: dr/detail/ranges_v3_shim.hpp
#pragma once

// In order to use ranges-v3 instead of C++ 20 ranges,
// #define _ONEDPL_DR_STD_RANGES_SHIM <dr/detail/ranges_v3_shim.hpp>
// before including oneDPL DR headers

#include <range/v3/all.hpp>

namespace stdrng = ::ranges;

#define _ONEDPL_DR_STD_RANGES_NAMESPACE ranges

#include "owning_view.hpp"

namespace _ONEDPL_DR_STD_RANGES_NAMESPACE
{
    using dr::__detail::owning_view;
}
```
The macro `_ONEDPL_DR_STD_RANGES_SHIM` should name the "replacement" shim header, in this case, the one that redirects to ranges-v3. It will be included by the shim in oneDPL.

Consequently, the custom implementation of `owning_view` which is only needed for ranges-v3 would also remain in the DR repository.


lslusarczyk:
Both DR tests timed out on "Zip/0.Drop" on this change.

akukanov:
I will take a look at the test. Yet, I would like to have some pros & cons discussion as well, as I might easily miss some reasons for `owning_view` to really be the proper way.

lslusarczyk:
> I will take a look at the test. Yet, I would like to have some pros & cons discussion as well, as I might easily miss some reasons for `owning_view` to really be the proper way to return the segments.

Sure. I have to look and @BenBrock too. It would be great to simplify the code and remove not needed parts. I had no enough time to rereview owning-view so I only managed to spot this failing test.

BenBrock:
The issue here is that we have some functions that create views of segments.  They take in a range of segments and return a modified view of those segments (e.g. [`take_segments` on line 33 of `segments_tools.hpp`](https://github.com/oneapi-src/oneDPL/pull/1479/files#diff-64c0eb389b436a415e819a57224d70c32ab7b7830c2766aa8ff4e1036bf5a9d9R33)).

These functions are implemented, naturally, using standard range adaptors like `enumerate`, `take`, and `transform`.  Those range adaptors must store a view, which they create using `views::all` (the stored view being of type `views::all_t`).  If you pass in an lvalue to an owning range (like a vector of `device_span`s), it will store a `ref_view`, assuming that the owning range will outlive the lifetime of the newly created `ref_view`.

I assume this is what's causing problems in the PR---we assume that `segments(my_zip_view)` returns a view and use an lvalue reference to it to create more views.  This can be fixed by either ensuring `segments()` always returns a view (my preference), or by modifying the code to never use an lvalue reference to the return of `segments()` (somewhat trickier, and the user could potentially always violate this).

My preference would probably be to simplify using `std::views::owning_view`.  I don't think we have significant uses of copying segment views.  If I remember correctly I primarily added a copy constructor because `mhp` needed it.

akukanov:
Ben's response gave me the clue for where to look. Apparently, the implementation of `take_ & drop_segments` does not properly handle rvalue arguments. Simply adding `std::forward` fixed the tests.

The semantical questions remain though - what to expect from `ranges::segments(dr)`? Does it return a view, a range of views, or a range of ranges (assuming that each view is a range, but not vice versa)? Is the returned type copyable, movable but not copyable, or it depends on the range?

PR #1759, opened 2024-08-05
Fix logic error in `__early_exit_find_or::operator()`

SergeyKopienko:
In this PR we fix logic error in `__early_exit_find_or::operator()` :
- not enough simple update the state of `__something_was_found` inside the  current sub-group in case of forward/backward search.

I think the error has been introduced in the PR https://github.com/oneapi-src/oneDPL/pull/1689


SergeyKopienko:
As I understood there is no error actually. Closed.

PR #1762, opened 2024-08-05
Add reduce then scan algorithm for transform scan family

danhoeflinger:
This PR Adds a generic reduce_then_scan algorithm for scan-like algorithms, and uses this algorithm for transform scan family of scan-like algorithms where it is beneficial.  

This PR is targeted after a pair of other PRs #1769 + #1770, which lay some plumbing for these changes.  Those two PRs were separated from this one to provide more focus to this change.

It is important to understand that subsequent PRs will add other families of scan-like algorithms (copy_if, partition, unique, ...), which may justify some aspects of the implementation of the reduce_then_scan kernels that could seem overcomplicated for the transform scan family alone.  

---

This PR is targeted to #1770, to allow for a clean diff, and is a part of the following sequence of PRs meant to be merged in order:

#1769 [MERGED] ~~Relocate __lazy_ctor_storage to utils header~~
#1770 [MERGED] ~~Use __result_and_scratch_storage within scan kernels~~
#1762 Add reduce_then_scan algorithm for transform scan family (This PR)

### DIFF
--- a/include/oneapi/dpl/experimental/kt/single_pass_scan.h
+++ b/include/oneapi/dpl/experimental/kt/single_pass_scan.h
@@ -332,7 +332,7 @@ __single_pass_scan(sycl::queue __queue, _InRange&& __in_rng, _OutRange&& __out_r
     auto __n_uniform = ::oneapi::dpl::__internal::__dpl_bit_ceil(__n);
 
     // Perform a single-work group scan if the input is small
-    if (oneapi::dpl::__par_backend_hetero::__group_scan_fits_in_slm<_Type>(__queue, __n, __n_uniform, 16384))
+    if (oneapi::dpl::__par_backend_hetero::__group_scan_fits_in_slm<_Type>(__queue, __n, __n_uniform, /*limit=*/16384))
     {
         return oneapi::dpl::__par_backend_hetero::__parallel_transform_scan_single_group(
             oneapi::dpl::__internal::__device_backend_tag{},
--- a/include/oneapi/dpl/experimental/kt/single_pass_scan.h
+++ b/include/oneapi/dpl/experimental/kt/single_pass_scan.h
@@ -332,7 +332,7 @@ __single_pass_scan(sycl::queue __queue, _InRange&& __in_rng, _OutRange&& __out_r
     auto __n_uniform = ::oneapi::dpl::__internal::__dpl_bit_ceil(__n);
 
     // Perform a single-work group scan if the input is small
-    if (oneapi::dpl::__par_backend_hetero::__group_scan_fits_in_slm<_Type>(__queue, __n, __n_uniform))
+    if (oneapi::dpl::__par_backend_hetero::__group_scan_fits_in_slm<_Type>(__queue, __n, __n_uniform, 16384))
     {
         return oneapi::dpl::__par_backend_hetero::__parallel_transform_scan_single_group(
             oneapi::dpl::__internal::__device_backend_tag{},
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -809,8 +809,6 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
     // work-group implementation requires a fundamental type which must also be trivially copyable.
     if constexpr (std::is_trivially_copyable_v<_Type>)
     {
-        // Next power of 2 greater than or equal to __n
-        auto __n_uniform = oneapi::dpl::__internal::__dpl_bit_ceil(__n);
         bool __use_reduce_then_scan = oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec);
 
         // TODO: Consider re-implementing single group scan to support types without known identities. This could also
@@ -818,6 +816,9 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
         constexpr bool __can_use_group_scan = unseq_backend::__has_known_identity<_BinaryOperation, _Type>::value;
         if constexpr (__can_use_group_scan)
         {
+            // Next power of 2 greater than or equal to __n
+            auto __n_uniform = oneapi::dpl::__internal::__dpl_bit_ceil(__n);
+
             // Empirically found values for reduce-then-scan and multi pass scan implementation for single wg cutoff
             std::size_t __single_group_upper_limit = __use_reduce_then_scan ? 2048 : 16384;
             if (__group_scan_fits_in_slm<_Type>(__exec.queue(), __n, __n_uniform, __single_group_upper_limit))
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -840,8 +840,8 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
                 _WriteOp{}, __init, _Inclusive{});
         }
     }
-    //else use legacy scan implementation
 
+    //else use legacy scan implementation
     using _Assigner = unseq_backend::__scan_assigner;
     using _NoAssign = unseq_backend::__scan_no_assign;
     using _UnaryFunctor = unseq_backend::walk_n<_ExecutionPolicy, _UnaryOperation>;
@@ -852,18 +852,18 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
     _NoOpFunctor __get_data_op;
 
     return __parallel_transform_scan_base(
-        __backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__in_rng),
-        std::forward<_Range2>(__out_rng), __binary_op, __init,
-        // local scan
-        unseq_backend::__scan<_Inclusive, _ExecutionPolicy, _BinaryOperation, _UnaryFunctor, _Assigner, _Assigner,
-                                _NoOpFunctor, _InitType>{__binary_op, _UnaryFunctor{__unary_op}, __assign_op,
-                                                        __assign_op, __get_data_op},
-        // scan between groups
-        unseq_backend::__scan</*inclusive=*/::std::true_type, _ExecutionPolicy, _BinaryOperation, _NoOpFunctor,
-                                _NoAssign, _Assigner, _NoOpFunctor, unseq_backend::__no_init_value<_Type>>{
-            __binary_op, _NoOpFunctor{}, __no_assign_op, __assign_op, __get_data_op},
-        // global scan
-        unseq_backend::__global_scan_functor<_Inclusive, _BinaryOperation, _InitType>{__binary_op, __init});
+            __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), ::std::forward<_Range1>(__in_rng),
+            ::std::forward<_Range2>(__out_rng), __binary_op, __init,
+            // local scan
+            unseq_backend::__scan<_Inclusive, _ExecutionPolicy, _BinaryOperation, _UnaryFunctor, _Assigner, _Assigner,
+                                  _NoOpFunctor, _InitType>{__binary_op, _UnaryFunctor{__unary_op}, __assign_op,
+                                                           __assign_op, __get_data_op},
+            // scan between groups
+            unseq_backend::__scan</*inclusive=*/::std::true_type, _ExecutionPolicy, _BinaryOperation, _NoOpFunctor,
+                                  _NoAssign, _Assigner, _NoOpFunctor, unseq_backend::__no_init_value<_Type>>{
+                __binary_op, _NoOpFunctor{}, __no_assign_op, __assign_op, __get_data_op},
+            // global scan
+            unseq_backend::__global_scan_functor<_Inclusive, _BinaryOperation, _InitType>{__binary_op, __init});
 }
 
 template <typename _SizeType>
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -785,7 +785,7 @@ struct __simple_write_to_id
 {
     template <typename _OutRng, typename ValueType>
     void
-    operator()(const _OutRng& __out_rng, std::size_t __id, const ValueType& __v) const
+    operator()(_OutRng& __out_rng, std::size_t __id, const ValueType& __v) const
     {
         // Use of an explicit cast to our internal tuple type is required to resolve conversion issues between our
         // internal tuple and std::tuple. If the underlying type is not a tuple, then the type will just be passed through.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -771,7 +771,7 @@ struct __gen_transform_input
 {
     template <typename InRng>
     auto
-    operator()(InRng&& __in_rng, std::size_t __idx) const
+    operator()(const InRng& __in_rng, std::size_t __idx) const
     {
         using _ValueType = oneapi::dpl::__internal::__value_t<InRng>;
         using _OutValueType = oneapi::dpl::__internal::__decay_with_tuple_specialization_t<
@@ -785,7 +785,7 @@ struct __simple_write_to_idx
 {
     template <typename _OutRng, typename ValueType>
     void
-    operator()(_OutRng&& __out_rng, std::size_t __idx, const ValueType& __v) const
+    operator()(const _OutRng& __out_rng, std::size_t __idx, const ValueType& __v) const
     {
         // Use of an explicit cast to our internal tuple type is required to resolve conversion issues between our
         // internal tuple and std::tuple. If the underlying type is not a tuple, then the type will just be passed through.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -868,25 +868,6 @@ struct __write_to_idx_if
     Assign __assign;
 };
 
-template <typename Assign = oneapi::dpl::__internal::__pstl_assign>
-struct __write_to_idx_if_else
-{
-    template <typename _OutRng, typename _SizeType, typename ValueType>
-    void
-    operator()(_OutRng&& __out_rng, _SizeType __idx, const ValueType& __v) const
-    {
-        using _ConvertedTupleType =
-            typename oneapi::dpl::__internal::__get_tuple_type<std::decay_t<decltype(std::get<2>(__v))>,
-                                                               std::decay_t<decltype(__out_rng[__idx])>>::__type;
-        if (std::get<1>(__v))
-            __assign(static_cast<_ConvertedTupleType>(std::get<2>(__v)), std::get<0>(__out_rng[std::get<0>(__v) - 1]));
-        else
-            __assign(static_cast<_ConvertedTupleType>(std::get<2>(__v)),
-                     std::get<1>(__out_rng[__idx - std::get<0>(__v)]));
-    }
-    Assign __assign;
-};
-
 template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _UnaryOperation, typename _InitType,
           typename _BinaryOperation, typename _Inclusive>
 auto
@@ -1064,33 +1045,6 @@ __parallel_scan_copy(oneapi::dpl::__internal::__device_backend_tag __backend_tag
                                                               __get_data_op},
         // global scan
         __copy_by_mask_op);
-}
-
-template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _UnaryPredicate>
-auto
-__parallel_partition_copy(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _ExecutionPolicy&& __exec,
-                          _Range1&& __rng, _Range2&& __result, _UnaryPredicate __pred)
-{
-    auto __n = __rng.size();
-    if (oneapi::dpl::__par_backend_hetero::__prefer_reduce_then_scan(__exec))
-    {
-        using _GenMask = oneapi::dpl::__par_backend_hetero::__gen_mask<_UnaryPredicate>;
-        using _WriteOp =
-            oneapi::dpl::__par_backend_hetero::__write_to_idx_if_else<oneapi::dpl::__internal::__pstl_assign>;
-
-        return __parallel_reduce_then_scan_copy(__backend_tag, std::forward<_ExecutionPolicy>(__exec),
-                                                std::forward<_Range1>(__rng), std::forward<_Range2>(__result), __n,
-                                                _GenMask{__pred}, _WriteOp{});
-    }
-    else
-    {
-        using _ReduceOp = std::plus<decltype(__n)>;
-        using _CreateOp = unseq_backend::__create_mask<_UnaryPredicate, decltype(__n)>;
-        using _CopyOp = unseq_backend::__partition_by_mask<_ReduceOp, /*inclusive*/ std::true_type>;
-
-        return __parallel_scan_copy(__backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__rng),
-                                    std::forward<_Range2>(__result), __n, _CreateOp{__pred}, _CopyOp{_ReduceOp{}});
-    }
 }
 
 template <typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Size, typename _Pred,
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -555,10 +555,10 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
                                                         __internal::__optional_kernel_name<_ScanKernelName...>>
 {
     template <typename _Policy, typename _InRng, typename _OutRng, typename _InitType, typename _BinaryOperation,
-              typename _UnaryOp, typename _Assign>
+              typename _UnaryOp>
     auto
     operator()(_Policy&& __policy, _InRng&& __in_rng, _OutRng&& __out_rng, ::std::size_t __n, _InitType __init,
-               _BinaryOperation __bin_op, _UnaryOp __unary_op, _Assign __assign)
+               _BinaryOperation __bin_op, _UnaryOp __unary_op)
     {
         using _ValueType = ::std::uint16_t;
 
@@ -622,8 +622,7 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
                     for (::std::uint16_t __idx = __item_id; __idx < __n; __idx += _WGSize)
                     {
                         if (__lacc[__idx])
-                            __assign(static_cast<__tuple_type>(__in_rng[__idx]),
-                                     __out_rng[__lacc[__idx + __elems_per_wg]]);
+                            __out_rng[__lacc[__idx + __elems_per_wg]] = static_cast<__tuple_type>(__in_rng[__idx]);
                     }
 
                     const ::std::uint16_t __residual = __n % _WGSize;
@@ -632,8 +631,7 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
                     {
                         auto __idx = __residual_start + __item_id;
                         if (__lacc[__idx])
-                            __assign(static_cast<__tuple_type>(__in_rng[__idx]),
-                                     __out_rng[__lacc[__idx + __elems_per_wg]]);
+                            __out_rng[__lacc[__idx + __elems_per_wg]] = static_cast<__tuple_type>(__in_rng[__idx]);
                     }
 
                     if (__item_id == 0)
@@ -798,76 +796,6 @@ struct __simple_write_to_idx
     }
 };
 
-template <typename _Predicate>
-struct __gen_mask
-{
-    template <typename _InRng>
-    bool
-    operator()(_InRng&& __in_rng, std::size_t __idx) const
-    {
-        return __pred(__in_rng[__idx]);
-    }
-    _Predicate __pred;
-};
-
-template <typename _GenMask>
-struct __gen_count_mask
-{
-    template <typename _InRng, typename _SizeType>
-    _SizeType
-    operator()(_InRng&& __in_rng, _SizeType __idx) const
-    {
-        return __gen_mask(std::forward<_InRng>(__in_rng), __idx) ? _SizeType{1} : _SizeType{0};
-    }
-    _GenMask __gen_mask;
-};
-
-template <typename _GenMask>
-struct __gen_expand_count_mask
-{
-    template <typename _InRng, typename _SizeType>
-    auto
-    operator()(_InRng&& __in_rng, _SizeType __idx) const
-    {
-        // Explicitly creating this element type is necessary to avoid modifying the input data when _InRng is a
-        //  zip_iterator which will return a tuple of references when dereferenced. With this explicit type, we copy
-        //  the values of zipped the input types rather than their references.
-        using _ElementType =
-            oneapi::dpl::__internal::__decay_with_tuple_specialization_t<oneapi::dpl::__internal::__value_t<_InRng>>;
-        _ElementType ele = __in_rng[__idx];
-        bool mask = __gen_mask(__in_rng, __idx);
-        return std::tuple(mask ? _SizeType{1} : _SizeType{0}, mask, ele);
-    }
-    _GenMask __gen_mask;
-};
-
-struct __get_zeroth_element
-{
-    template <typename _Tp>
-    auto&
-    operator()(_Tp&& __a) const
-    {
-        return std::get<0>(std::forward<_Tp>(__a));
-    }
-};
-template <std::int32_t __offset = 0, typename Assign = oneapi::dpl::__internal::__pstl_assign>
-struct __write_to_idx_if
-{
-    template <typename _OutRng, typename _SizeType, typename ValueType>
-    void
-    operator()(_OutRng&& __out_rng, _SizeType __idx, const ValueType& __v) const
-    {
-        // Use of an explicit cast to our internal tuple type is required to resolve conversion issues between our
-        // internal tuple and std::tuple. If the underlying type is not a tuple, then the type will just be passed through.
-        using _ConvertedTupleType =
-            typename oneapi::dpl::__internal::__get_tuple_type<std::decay_t<decltype(std::get<2>(__v))>,
-                                                               std::decay_t<decltype(__out_rng[__idx])>>::__type;
-        if (std::get<1>(__v))
-            __assign(static_cast<_ConvertedTupleType>(std::get<2>(__v)), __out_rng[std::get<0>(__v) - 1 + __offset]);
-    }
-    Assign __assign;
-};
-
 template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _UnaryOperation, typename _InitType,
           typename _BinaryOperation, typename _Inclusive>
 auto
@@ -944,16 +872,13 @@ struct __invoke_single_group_copy_if
     // Specialization for devices that have a max work-group size of at least 1024
     static constexpr ::std::uint16_t __targeted_wg_size = 1024;
 
-    template <std::uint16_t _Size, typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Pred,
-              typename _Assign = oneapi::dpl::__internal::__pstl_assign>
+    template <::std::uint16_t _Size, typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Pred>
     auto
-    operator()(_ExecutionPolicy&& __exec, std::size_t __n, _InRng&& __in_rng, _OutRng&& __out_rng, _Pred&& __pred,
-               _Assign&& __assign)
+    operator()(_ExecutionPolicy&& __exec, ::std::size_t __n, _InRng&& __in_rng, _OutRng&& __out_rng, _Pred&& __pred)
     {
         constexpr ::std::uint16_t __wg_size = ::std::min(_Size, __targeted_wg_size);
         constexpr ::std::uint16_t __num_elems_per_item = ::oneapi::dpl::__internal::__dpl_ceiling_div(_Size, __wg_size);
         const bool __is_full_group = __n == __wg_size;
-
         using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
         using _InitType = unseq_backend::__no_init_value<::std::uint16_t>;
         using _ReduceOp = ::std::plus<::std::uint16_t>;
@@ -967,8 +892,7 @@ struct __invoke_single_group_copy_if
             return __par_backend_hetero::__parallel_copy_if_static_single_group_submitter<
                 _SizeType, __num_elems_per_item, __wg_size, true, _FullKernelName>()(
                 std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
-                std::forward<_OutRng>(__out_rng), __n, _InitType{}, _ReduceOp{}, std::forward<_Pred>(__pred),
-                std::forward<_Assign>(__assign));
+                std::forward<_OutRng>(__out_rng), __n, _InitType{}, _ReduceOp{}, std::forward<_Pred>(__pred));
         }
         else
         {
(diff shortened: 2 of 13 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,832 @@
...
+    using _ScanKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
+        __reduce_then_scan_scan_kernel<_CustomName>>;
+    using _ValueType = typename _InitType::__value_type;
+
+    constexpr std::size_t __sub_group_size = 32;
+    // Empirically determined maximum. May be less for non-full blocks.
+    constexpr std::uint8_t __max_inputs_per_item = 128;
+    constexpr bool __inclusive = _Inclusive::value;
+
+    // TODO: This min call is temporary until PR #1683 is merged.
```
mmichel11, line 735: I think we can remove the min call now since PR 1683 was just merged.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -526,8 +508,8 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
                         }
                     }
 
-                    __scan_work_group<_ValueType, _Inclusive>(__group, __lacc_ptr, __lacc_ptr + __n,
-                                                              __lacc_ptr, __bin_op, __init);
+                    __scan_work_group<_ValueType, _Inclusive>(__group, __lacc_ptr, __lacc_ptr + __n, __lacc_ptr,
+                                                              __bin_op, __init);
```
danhoeflinger, line 512: formatting only change, should be reverted.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -631,7 +614,7 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
 
                     __scan_work_group<_ValueType, /* _Inclusive */ false>(
                         __group, __lacc_ptr, __lacc_ptr + __elems_per_wg, __lacc_ptr + __elems_per_wg, __bin_op,
-                         __init);
+                        __init);
```
danhoeflinger, line 617: formatting only change, should be reverted

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -755,7 +756,7 @@ template <typename _Type>
 bool
 __group_scan_fits_in_slm(const sycl::queue& __queue, ::std::size_t __n, ::std::size_t __n_uniform)
 {
-    constexpr int __single_group_upper_limit = 16384;
+    constexpr int __single_group_upper_limit = 2048;
```
dmitriy-sobolev, line 759: Is that based on some empirical finding? If so, it may sense to leave a comment.

dmitriy-sobolev (commented):
Thank your for the PR.

I am familiarizing myself with the structure of the PR, while leaving minor comments like these on the way.
Probably, more important comments will follow later if any.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I've refactored this a little to add a comment, and have a separate threshold for reduce-then-scan and legacy scan.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -773,24 +804,44 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
...
+    // Reduce-then-scan is dependent on sycl::shift_group_right which requires the underlying type to be trivially
+    // copyable. If this is not met, then we must fallback to the legacy implementation. The single work-group implementation
+    // requires a fundamental type which must also be trivially copyable.
+    if constexpr (std::is_trivially_copyable_v<_Type>)
     {
-        if (__group_scan_fits_in_slm<_Type>(__exec.queue(), __n, __n_uniform))
+        // Next power of 2 greater than or equal to __n
+        auto __n_uniform = __n;
+        if ((__n_uniform & (__n_uniform - 1)) != 0)
+            __n_uniform = oneapi::dpl::__internal::__dpl_bit_floor(__n) << 1;
```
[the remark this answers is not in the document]
danhoeflinger: Good catch, Using `dpl_bit_ceil` now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -801,15 +852,15 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
     _NoOpFunctor __get_data_op;
 
     return __parallel_transform_scan_base(
-        __backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__in_rng),
-        std::forward<_Range2>(__out_rng), __binary_op, __init,
+        __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), ::std::forward<_Range1>(__in_rng),
```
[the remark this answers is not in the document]
danhoeflinger: This looks like the effect of a bad rebase.  Removed all these changes.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+};
+
+// reduce_then_scan requires subgroup size of 32, and performs well only on devices with fast coordinated subgroup
+// operations.  We do not want to run this can on CPU targets, as they are not performant with this algorithm.
+template <typename _ExecutionPolicy>
+bool
+__prefer_reduce_then_scan(const _ExecutionPolicy& __exec)
+{
+    const bool __dev_has_sg32 = __par_backend_hetero::__supports_sub_group_size(__exec, 32);
+    return (!__exec.queue().get_device().is_cpu() && __dev_has_sg32);
```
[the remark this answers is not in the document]
danhoeflinger: Sure, I have no objection to this.  Changed. If this somehow stops icpx from eliminating dead code JIT kernels on with CPU device type, we will revert it, but I expect it shouldn't change that. Changed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+            oneapi::dpl::__ranges::__require_access(__cgh, __in_rng, __out_rng);
+            auto __temp_acc = __scratch_container.__get_scratch_acc(__cgh);
+            auto __res_acc = __scratch_container.__get_result_acc(__cgh);
+
+            __cgh.parallel_for<_KernelName...>(__nd_range, [=,
+                                                            *this](sycl::nd_item<1> __ndi) [[sycl::reqd_sub_group_size(
+                                                               __sub_group_size)]] {
+                auto __tmp_ptr = _TmpStorageAcc::__get_usm_or_buffer_accessor_ptr(__temp_acc);
+                auto __res_ptr =
+                    _TmpStorageAcc::__get_usm_or_buffer_accessor_ptr(__res_acc, __num_sub_groups_global + 2);
```
julianmi, line 454: Don't you have a single result only? If so, I think we should get the zeroth element of `__res_acc`, i.e., the default.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                auto __lid = __ndi.get_local_id(0);
+                auto __g = __ndi.get_group(0);
```
julianmi, line 456: Similar comments here about the naming: `__local_id` and `__group_id`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                auto __sub_group = __ndi.get_sub_group();
+                auto __sub_group_id = __sub_group.get_group_linear_id();
+                auto __sub_group_local_id = __sub_group.get_local_linear_id();
+
+                auto __group_start_idx =
+                    (__block_num * __max_block_size) + (__g * __inputs_per_sub_group * __num_sub_groups_local);
```
julianmi, line 462: Similar comment about adding a helper for `__inputs_per_sub_group * __num_sub_groups_local` and reusing it on the next line.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                    std::min(__n - __group_start_idx, std::size_t(__num_sub_groups_local * __inputs_per_sub_group));
+                std::uint32_t __active_subgroups =
+                    oneapi::dpl::__internal::__dpl_ceiling_div(__elements_in_group, __inputs_per_sub_group);
+                oneapi::dpl::__internal::__lazy_ctor_storage<_InitValueType> __carry_last;
+
+                // propagate carry in from previous block
+                oneapi::dpl::__internal::__lazy_ctor_storage<_InitValueType> __sub_group_carry;
+
+                // on the first sub-group in a work-group (assuming S subgroups in a work-group):
+                // 1. load S sub-group local carry pfix sums (T0..TS-1) to slm
```
julianmi, line 474:
```suggestion
                // 1. load S sub-group local carry prefix sums (T0..TS-1) to SLM
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                // 2. load 32, 64, 96, etc. TS-1 work-group carry-outs (32 for WG num<32, 64 for WG num<64, etc.),
+                //    and then compute the prefix sum to generate global carry out
+                //    for each WG, i.e., prefix sum on TS-1 carries over all WG.
+                // 3. on each WG select the adjacent neighboring WG carry in
+                // 4. on each WG add the global carry-in to S sub-group local pfix sums to
```
julianmi, line 479:
```suggestion
                // 4. on each WG add the global carry-in to S sub-group local prefix sums to
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                //    get a T-local global carry in
+                // 5. recompute T-local pfix values, add the T-local global carries,
```
julianmi, line 481:
```suggestion
                // 5. recompute T-local prefix values, add the T-local global carries,
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                //    and then write back the final values to memory
+                if (__sub_group_id == 0)
+                {
+                    // step 1) load to Xe slm the WG-local S prefix sums
```
julianmi, line 485:
```suggestion
                    // step 1) load to Xe SLM the WG-local S prefix sums
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                    for (; __i < __iters - 1; __i++)
+                    {
+                        __sub_group_partials[__i * __sub_group_size + __sub_group_local_id] =
+                            __tmp_ptr[__subgroups_before_my_group + __i * __sub_group_size + __sub_group_local_id];
+                    }
+                    if (__i * __sub_group_size + __sub_group_local_id < __active_subgroups)
+                    {
+                        __sub_group_partials[__i * __sub_group_size + __sub_group_local_id] =
+                            __tmp_ptr[__subgroups_before_my_group + __i * __sub_group_size + __sub_group_local_id];
+                    }
```
julianmi, line 502: We might want to add a helper for ` __sub_group_size + __sub_group_local_id` and add `__sub_group_size` in each iteration to reuse it throughout this code. Same for `__subgroups_before_my_group + __i * __sub_group_size + __sub_group_local_id`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                            // multiple iterations
+                            // first 1 full
+                            auto __value = __tmp_ptr[__num_sub_groups_local * __sub_group_local_id + __offset];
+                            __sub_group_scan<__sub_group_size, /*__is_inclusive=*/true, /*__init_present=*/false>(
+                                __sub_group, __value, __reduce_op, __carry_last);
+
+                            // then some number of full iterations
+                            for (std::uint32_t __i = 1; __i < __pre_carry_iters - 1; __i++)
+                            {
+                                auto __reduction_idx = __i * __num_sub_groups_local * __sub_group_size +
```
julianmi, line 539: We might want to create helpers for `__num_sub_groups_local * __sub_group_size` and `__num_sub_groups_local * __sub_group_local_id + __offset` and reuse them in this loop and the partial iteration below.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                        //            and apply to local sub-group prefix carries
+                        auto __carry_offset = 0;
+
+                        std::uint8_t __iters =
+                            oneapi::dpl::__internal::__dpl_ceiling_div(__active_subgroups, __sub_group_size);
+
+                        std::uint8_t __i = 0;
+                        for (; __i < __iters - 1; ++__i)
+                        {
+                            __sub_group_partials[__carry_offset + __sub_group_local_id] = __reduce_op(
```
julianmi, line 570: We might want to create a helper for `__carry_offset + __sub_group_local_id` and reuse it in this loop and below.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                else // first group first block, no subgroup carry
+                {
+                    __scan_through_elements_helper<__sub_group_size, __is_inclusive,
+                                                   /*__init_present=*/false,
+                                                   /*__capture_output=*/true, __max_inputs_per_item>(
+                        __sub_group, __gen_scan_input, __scan_input_transform, __reduce_op, __write_op,
+                        __sub_group_carry, __in_rng, __out_rng, __start_idx, __n, __inputs_per_item,
+                        __subgroup_start_idx, __sub_group_id, __active_subgroups);
+                }
+                //If within the last active group and subgroup of the block, use the 0th work item of the subgroup
```
julianmi, line 663:
```suggestion
                // If within the last active group and sub-group of the block, use the 0th work-item of the sub-group
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                if (__sub_group_local_id == 0 && (__active_groups == __g + 1) &&
+                    (__active_subgroups == __sub_group_id + 1))
+                {
+                    if (__block_num + 1 == __num_blocks)
+                    {
+                        __res_ptr[0] = __sub_group_carry.__v;
+                    }
+                    else
+                    {
+                        //capture the last carry out for the next block
```
julianmi, line 674:
```suggestion
                        // capture the last carry out for the next block
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -773,24 +803,45 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
...
+
+            return __parallel_transform_reduce_then_scan(
+                __backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__in_rng),
+                std::forward<_Range2>(__out_rng), __gen_transform, __binary_op, __gen_transform, _ScanInputTransform{},
+                _WriteOp{}, __init, _Inclusive{});
         }
     }
 
-    // Either we can't use group scan or this input is too big for one workgroup
+    //else use legacy scan implementation
```
julianmi, line 844: We might want to call it by the algorithm used instead of "legacy".

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
julianmi: My bad. This code looks correct to me in case we have a fallback buffer.

danhoeflinger (commented):
Thanks very much for the review, and for helping clean up the read-ability.  I think it is in a better place now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                                   __inputs_per_sub_group, __inputs_per_item, __b);
+        __num_remaining -= std::min(__num_remaining, __block_size);
+        // We only need to resize these parameters prior to the last block as it is the only non-full case.
+        if (__b + 2 == __num_blocks)
+        {
+            __inputs_per_sub_group =
+                __num_remaining >= __max_inputs_per_block
+                    ? __max_inputs_per_block / __num_sub_groups_global
+                    : std::max(__sub_group_size,
+                               oneapi::dpl::__internal::__dpl_bit_ceil(__num_remaining) / __num_sub_groups_global);
```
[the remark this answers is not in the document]
danhoeflinger: done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,835 @@
...
+                                    __init};
+    sycl::event __event;
+    // Data is processed in 2-kernel blocks to allow contiguous input segment to persist in LLC between the first and second kernel for accelerators
+    // with sufficiently large L2 / L3 caches.
+    for (std::size_t __b = 0; __b < __num_blocks; ++__b)
+    {
+        auto __elements_in_block = oneapi::dpl::__internal::__dpl_ceiling_div(
+            std::min(__num_remaining, __max_inputs_per_block), __inputs_per_item);
+        auto __ele_in_block_round_up_workgroup =
+            oneapi::dpl::__internal::__dpl_ceiling_div(__elements_in_block, __work_group_size) * __work_group_size;
```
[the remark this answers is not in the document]
danhoeflinger: I renamed things generally in the direction of `inputs` where it makes sense.  Thanks.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -765,6 +765,36 @@ __group_scan_fits_in_slm(const sycl::queue& __queue, ::std::size_t __n, ::std::s
...
+struct __gen_transform_input
+{
+    template <typename InRng>
+    auto
+    operator()(const InRng& __in_rng, std::size_t __idx) const
+    {
+        using _ValueType = oneapi::dpl::__internal::__value_t<InRng>;
+        using _OutValueType = oneapi::dpl::__internal::__decay_with_tuple_specialization_t<
+            typename std::invoke_result<_UnaryOp, _ValueType>::type>;
+        return _OutValueType{__unary_op(__in_rng[__idx])};
```
akukanov, line 778: As far as I can tell, all the current implementations of scan kernels use nothing like `__decay_with_tuple_specialization_t` but instead assign the result of the unary operation to either an `auto` variable or to a value of `typename _InitType::__value_type`. So I wonder what's different in this implementation? If nothing, then I would suggest to add the expected output type as the template parameter of `__gen_transform_input`, which would be set by the caller to the init value type as noted above.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -332,7 +332,7 @@ __single_pass_scan(sycl::queue __queue, _InRange&& __in_rng, _OutRange&& __out_r
     auto __n_uniform = ::oneapi::dpl::__internal::__dpl_bit_ceil(__n);
 
     // Perform a single-work group scan if the input is small
-    if (oneapi::dpl::__par_backend_hetero::__group_scan_fits_in_slm<_Type>(__queue, __n, __n_uniform))
+    if (oneapi::dpl::__par_backend_hetero::__group_scan_fits_in_slm<_Type>(__queue, __n, __n_uniform, 16384))
```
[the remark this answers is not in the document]
danhoeflinger: applied

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -773,24 +803,45 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
...
-    if constexpr (__can_use_group_scan)
+    // Reduce-then-scan is dependent on sycl::shift_group_right which requires the underlying type to be trivially
+    // copyable. If this is not met, then we must fallback to the multi pass scan implementation. The single
+    // work-group implementation requires a fundamental type which must also be trivially copyable.
+    if constexpr (std::is_trivially_copyable_v<_Type>)
     {
-        if (__group_scan_fits_in_slm<_Type>(__exec.queue(), __n, __n_uniform))
+        // Next power of 2 greater than or equal to __n
+        auto __n_uniform = oneapi::dpl::__internal::__dpl_bit_ceil(__n);
+        bool __pref_reduce_then_scan = oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec);
```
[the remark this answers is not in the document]
danhoeflinger: sure, done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,840 @@
...
+    }
+
+    const std::size_t __max_block_size;
+    const std::size_t __num_sub_groups_local;
+    const std::size_t __num_sub_groups_global;
+    const std::size_t __num_work_items;
+    const std::size_t __num_blocks;
+    const std::size_t __n;
+
+    const _GenReduceInput __gen_reduce_input;
```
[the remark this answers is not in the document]
danhoeflinger: Yes, this is unused in the second kernel.  Removed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,840 @@
...
+            //   1 element for the accumulated block-local carry-in from previous groups in the block
+            __dpl_sycl::__local_accessor<_InitValueType> __sub_group_partials(__num_sub_groups_local + 1, __cgh);
+            __cgh.depends_on(__prior_event);
+            oneapi::dpl::__ranges::__require_access(__cgh, __in_rng, __out_rng);
+            auto __temp_acc = __scratch_container.__get_scratch_acc(__cgh);
+            auto __res_acc = __scratch_container.__get_result_acc(__cgh);
+
+            __cgh.parallel_for<_KernelName...>(__nd_range, [=,
+                                                            *this](sycl::nd_item<1> __ndi) [[sycl::reqd_sub_group_size(
+                                                               __sub_group_size)]] {
```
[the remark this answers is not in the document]
danhoeflinger: formatting fixed, much to clang-format's dismay. :)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -813,6 +814,14 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
     }
 };
 
+template <typename _ExecutionPolicy>
+bool
+__supports_sub_group_size(const _ExecutionPolicy& __exec, std::size_t __target_size)
```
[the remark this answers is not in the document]
danhoeflinger: moved.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: We initially ran into issues with an `auto` variable for `__in_rng[__idx]` when processing `zip_iterator` input where the dereference type is a tuple of references. It led to the caller modifying the input range when modifying the return of this function.

However, what we have here with `__decay_with_tuple_specialization_t` is unnecessary. What is important is ensuring that `__in_rng[__idx]` gets converted to the value type of the input range which can be achieved with the following:
```
        using _ValueType = oneapi::dpl::__internal::__value_t<_InRng>;
        return __unary_op(_ValueType{__in_rng[__idx]});
```
I do not think an additional template parameter is needed in this case, but we can discuss more if needed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -773,24 +804,45 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
...
-    constexpr bool __can_use_group_scan = unseq_backend::__has_known_identity<_BinaryOperation, _Type>::value;
-    if constexpr (__can_use_group_scan)
+    // Reduce-then-scan is dependent on sycl::shift_group_right which requires the underlying type to be trivially
+    // copyable. If this is not met, then we must fallback to the multi pass scan implementation. The single
+    // work-group implementation requires a fundamental type which must also be trivially copyable.
+    if constexpr (std::is_trivially_copyable_v<_Type>)
     {
-        if (__group_scan_fits_in_slm<_Type>(__exec.queue(), __n, __n_uniform))
+        // Next power of 2 greater than or equal to __n
+        auto __n_uniform = oneapi::dpl::__internal::__dpl_bit_ceil(__n);
```
akukanov, line 813: Move the computation of `__n_uniform` into the scope of its use.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -765,6 +765,37 @@ __group_scan_fits_in_slm(const sycl::queue& __queue, ::std::size_t __n, ::std::s
...
+    template <typename _OutRng, typename ValueType>
+    void
+    operator()(const _OutRng& __out_rng, std::size_t __idx, const ValueType& __v) const
+    {
+        // Use of an explicit cast to our internal tuple type is required to resolve conversion issues between our
+        // internal tuple and std::tuple. If the underlying type is not a tuple, then the type will just be passed through.
+        using _ConvertedTupleType =
+            typename oneapi::dpl::__internal::__get_tuple_type<std::decay_t<decltype(__v)>,
+                                                               std::decay_t<decltype(__out_rng[__idx])>>::__type;
+        __out_rng[__idx] = static_cast<_ConvertedTupleType>(__v);
```
akukanov, line 795: Does the conversion to internal tuple type also happen in the existing code (perhaps implicit), or is it only required for the new code? If the latter - is it due to the extra abstraction of `__simple_write_to_idx`? My concern is that it might introduce extra copying with some pressure on registers etc.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,836 @@
...
+{
+namespace dpl
+{
+namespace __par_backend_hetero
+{
+
+template <std::uint8_t __sub_group_size, bool __init_present, typename _MaskOp, typename _InitBroadcastId,
+          typename _SubGroup, typename _BinaryOp, typename _ValueType, typename _LazyValueType>
+void
+__exclusive_sub_group_masked_scan(const _SubGroup& __sub_group, _MaskOp __mask_fn, _InitBroadcastId __init_broadcast_id,
```
akukanov, line 41: Why subgroup is defined with a template parameter in this and all other helper functions, instead of being just `sycl::sub_group`? It's not even a template class in SYCL.

adamfidel: If I recall, this might have something to do with the `sub_group` class existing in a separate extensions namespace in older versions of libsycl. A similar pattern occurs in radix sort where we have to query for the subgroup type: https://github.com/oneapi-src/oneDPL/blob/82aa13f3060e144e7bc2312df47702b8a0ea285e/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h#L400

Higher up in that file, we only use `sycl::sub_group` directly if we know that we are using a version of libsycl of at least 5.7.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
adamfidel: Good point. It has been moved.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: This same conversion is present in single wg scan [here](https://github.com/oneapi-src/oneDPL/blob/82aa13f3060e144e7bc2312df47702b8a0ea285e/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h#L566), and multi-pass scan [here](https://github.com/oneapi-src/oneDPL/blob/82aa13f3060e144e7bc2312df47702b8a0ea285e/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L706), [here](https://github.com/oneapi-src/oneDPL/blob/82aa13f3060e144e7bc2312df47702b8a0ea285e/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L656), and [here](https://github.com/oneapi-src/oneDPL/blob/82aa13f3060e144e7bc2312df47702b8a0ea285e/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L600)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
akukanov: Can we handle it by defining a type alias in `namespace __dpl_sycl` that depends on the version of libsycl - similarly as we do it for many APIs in `hetero/dpcpp/sylc_defs.h`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
akukanov, line 1: I do not like the proliferation of `auto`-typed variables across the code base, In this file alone, it's more than on 60 occasions. I firmly believe that it's rather bad for code readability and understanding/

I think `auto` should be used in rather specific cases, such as:
- The actual type of the variable is unambiguously and easily follows from the code, while `auto` reduces verbosity. For example, `std::vector<int> v; ... auto v_iter = v.begin();` would be fine.
- The actual type is not known or hard to express as it depends on other types. A classic example is the parameter type of a generic lambda, and the type of variables that depend on these parameters. Even in such cases, it might be useful to give a local name to the type via `using type_name = decltype(expression)`.
- The actual type does not matter in the context. This should be a rare case I believe; for example, if you store a function return value to a temporary variable which you then just pass to another function, i.e. the variable serves no other purpose than readability/code formatting.

At the same time, `auto` should not be used only to save programmer's time on checking what a function returns etc. In most cases, being explicit about the type results in a better code.

Using `auto` sparingly allows it to attract attention to the places that deserve it, while using it everywhere will make it "commodity" hiding these potentially important places in the noise.

Please take another look at how `auto` is used in the patch, and replace it with the actual time unless there are solid reasons not to.

danhoeflinger: Sure, we can do that. I agree that it is overused.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
adamfidel: Yes, that is a good suggestion. I made this change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,843 @@
...
+            __dpl_sycl::__local_accessor<_InitValueType> __sub_group_partials(__num_sub_groups_local, __cgh);
+            __cgh.depends_on(__prior_event);
+            oneapi::dpl::__ranges::__require_access(__cgh, __in_rng);
+            auto __temp_acc = __scratch_container.__get_scratch_acc(__cgh);
+            __cgh.parallel_for<_KernelName...>(
+                    __nd_range, [=, *this](sycl::nd_item<1> __ndi) [[sycl::reqd_sub_group_size(__sub_group_size)]] {
+                _InitValueType* __temp_ptr = _TmpStorageAcc::__get_usm_or_buffer_accessor_ptr(__temp_acc);
+                std::size_t __group_id = __ndi.get_group(0);
+                __dpl_sycl::__sub_group __sub_group = __ndi.get_sub_group();
+                std::uint32_t __sub_group_id = __sub_group.get_group_linear_id();
```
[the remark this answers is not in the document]
danhoeflinger: For a sub_group, the return type is `uint32_t` rather than `size_t` from the sycl spec, better to use a smaller type when we can.
We may be able to shrink a few other types, but this one is clear.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -765,6 +765,37 @@ __group_scan_fits_in_slm(const sycl::queue& __queue, ::std::size_t __n, ::std::s
...
+        // We explicitly convert __in_rng[__idx] to the value type of _InRng to properly handle the case where we
+        // process zip_iterator input where the reference type is a tuple of a references. This prevents the caller
+        // from modifying the input range when altering the return of this functor.
+        using _ValueType = oneapi::dpl::__internal::__value_t<_InRng>;
+        return __unary_op(_ValueType{__in_rng[__idx]});
+    }
+    _UnaryOp __unary_op;
+};
+
+struct __simple_write_to_idx
```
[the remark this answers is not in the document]
danhoeflinger: sure, done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,843 @@
...
+    assert(__inputs_remaining > 0);
+    const std::uint32_t __max_inputs_per_subgroup = __max_inputs_per_block / __num_sub_groups_global;
+    std::uint32_t __evenly_divided_remaining_inputs =
+        std::max(std::size_t{__sub_group_size},
+                 oneapi::dpl::__internal::__dpl_bit_ceil(__inputs_remaining) / __num_sub_groups_global);
+    std::uint32_t __inputs_per_sub_group =
+        __inputs_remaining >= __max_inputs_per_block ? __max_inputs_per_subgroup : __evenly_divided_remaining_inputs;
+    std::uint32_t __inputs_per_item = __inputs_per_sub_group / __sub_group_size;
+    const std::size_t __block_size =
+        (__inputs_remaining < __max_inputs_per_block) ? __inputs_remaining : __max_inputs_per_block;
```
julianmi, line 770: Could this be `min(__inputs_remaining, __max_inputs_per_block)`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,843 @@
...
+    // Empirically determined maximum. May be less for non-full blocks.
+    constexpr std::uint8_t __max_inputs_per_item = 128;
+    constexpr bool __inclusive = _Inclusive::value;
+
+    const std::size_t __work_group_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
+
+    // TODO: Investigate potentially basing this on some scale of the number of compute units. 128 work-groups has been
+    // found to be reasonable number for most devices.
+    const std::size_t __num_work_groups = 128;
+    const std::size_t __num_work_items = __num_work_groups * __work_group_size;
```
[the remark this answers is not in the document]
danhoeflinger: I think this is reasonable, but if we do think, I think I want to explicitly add `8192` here above as the supplied `__max_work_group_size`.  Without this, we are subject to an issue if we modify the default at a later time and it breaks our assumptions.  
In this specific case I see it as unlikely we will ever have a workgroup size which exceeds the limits of `uint32_t`, (or 1/128th of that), but it is bad form to leave it like this with the return type being `size_t`.  I'll make this change then follow it through to make type adjustments accordingly.
Thanks.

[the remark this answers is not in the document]
danhoeflinger: I made these changes, and shrunk a bunch of types (and some other minor changes)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -765,6 +765,37 @@ __group_scan_fits_in_slm(const sycl::queue& __queue, ::std::size_t __n, ::std::s
...
+        return __unary_op(_ValueType{__in_rng[__id]});
+    }
+    _UnaryOp __unary_op;
+};
+
+struct __simple_write_to_id
+{
+    template <typename _OutRng, typename ValueType>
+    void
+    operator()(const _OutRng& __out_rng, std::size_t __id, const ValueType& __v) const
```
MikeDvorskiy, line 788: The constant reference `const _OutRng& __out_rng` looks suspicious, because range is output.

danhoeflinger: Good point, I removed the `const` here this and fixed a couple other range types in the kernel.  I will go through the remaining PRs to check for similar issues.  (I think output ranges for the other helpers probably have the same issue)

julianmi (approved):
LGTM. Please wait for another approval since there are partial reviews of several people.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,840 @@
...
+                //    and then compute the prefix sum to generate global carry out
+                //    for each WG, i.e., prefix sum on TS-1 carries over all WG.
+                // 3. on each WG select the adjacent neighboring WG carry in
+                // 4. on each WG add the global carry-in to S sub-group local prefix sums to
+                //    get a T-local global carry in
+                // 5. recompute T-local prefix values, add the T-local global carries,
+                //    and then write back the final values to memory
+                if (__sub_group_id == 0)
+                {
+                    // step 1) load to Xe SLM the WG-local S prefix sums
```
adamfidel, line 488: Should we remove the reference to `Xe` here since this algorithm is designed to work on non-Xe-based devices as well?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,840 @@
...
+
+                // Get inter-work group and adjusted for intra-work group prefix
+                bool __sub_group_carry_initialized = true;
+                if (__block_num == 0)
+                {
+                    if (__sub_group_id > 0)
+                    {
+                        _InitValueType __value = __sub_group_id - 1 < __active_subgroups
+                                                     ? __sub_group_partials[__sub_group_id - 1]
+                                                     : __sub_group_partials[__active_subgroups - 1];
```
danhoeflinger, line 601: This should be a min operation.
```suggestion
                        _InitValueType __value = __sub_group_partials[std::min( __sub_group_id - 1, __active_subgroups - 1)];
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,840 @@
...
+                        }
+                    }
+                }
+                else
+                {
+                    if (__sub_group_id > 0)
+                    {
+                        _InitValueType __value = __sub_group_id - 1 < __active_subgroups
+                                                     ? __sub_group_partials[__sub_group_id - 1]
+                                                     : __sub_group_partials[__active_subgroups - 1];
```
danhoeflinger, line 632: Same comment: min operation
```suggestion
                        _InitValueType __value = __sub_group_partials[std::min( __sub_group_id - 1, __active_subgroups - 1)];
```

danhoeflinger (commented):
a few minor things I found.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -94,6 +95,15 @@ __max_compute_units(const _ExecutionPolicy& __policy)
     return __policy.queue().get_device().template get_info<sycl::info::device::max_compute_units>();
 }
 
+template <typename _ExecutionPolicy>
+bool
+__supports_sub_group_size(const _ExecutionPolicy& __exec, std::size_t __target_size)
+{
+    const std::vector<size_t> __subgroup_sizes =
```
mmichel11, line 102:
```suggestion
    const std::vector<std::size_t> __subgroup_sizes =
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -0,0 +1,837 @@
...
+        __reduce_then_scan_scan_kernel<_CustomName>>;
+    using _ValueType = typename _InitType::__value_type;
+
+    constexpr std::uint8_t __sub_group_size = 32;
+    constexpr std::uint8_t __block_size_scale = std::max(std::size_t{1}, sizeof(double) / sizeof(_ValueType));
+    // Empirically determined maximum. May be less for non-full blocks.
+    constexpr std::uint16_t __max_inputs_per_item = 64 * __block_size_scale;
+    constexpr bool __inclusive = _Inclusive::value;
+
+    const std::uint32_t __work_group_size = oneapi::dpl::__internal::__max_work_group_size(__exec, 8192);
```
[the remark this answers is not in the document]
danhoeflinger: done.

adamfidel (approved):
This PR is in a good state and is okay to be merged.

mmichel11 (approved):
LGTM. We have checked tests and performance with the latest version.

julianmi (approved):

include/oneapi/dpl/pstl/hetero/numeric_ranges_impl_hetero.h
```diff
@@ -91,35 +91,15 @@ oneapi::dpl::__internal::__difference_t<_Range2>
 __pattern_transform_scan_base(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2,
                               _UnaryOperation __unary_op, _InitType __init, _BinaryOperation __binary_op, _Inclusive)
 {
-    if (__rng1.empty())
+    oneapi::dpl::__internal::__difference_t<_Range1> __n = __rng1.size();
```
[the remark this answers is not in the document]
adamfidel: Thanks for pointing this out. We created PR #1811 to address this issue.

PR #1766, opened 2024-08-06
Updates from Adam's comments

BenBrock:
This update is probably fine, but needs to be tested on PVC.  Currently having difficulty compiling due to lack of frontend compiler bug in Clang/lack of GCC-14, which includes [a workaround](https://github.com/gcc-mirror/gcc/commit/be34a8b538c0f04b11a428bd1a9340eb19dec13f).


lslusarczyk:
I've enabled tests which were disabled due to the bug.

PR #1768, opened 2024-08-06
Cxx20standard

lslusarczyk:
CXX-20 lacks zip_view (used one from oneDPL), enumerate (implemented self simple version).
Anyway still there are some problems and I think I can no longer hold everything else to work on this further.

Any help on this is appreciated.

```
/tmp/oneDPL/include/oneapi/dpl/internal/distributed_ranges_impl/sp/algorithms/../../detail/segments_tools.hpp:59:37:
 error: invalid operands to binary expression ('decltype(zip_view<iota_view<unsigned long, unsigned long>, vector<remote_vector<int, device_allocator<int, 0>>, allocator<remote_vector<int, device_allocator<int, 0>>>> &>(
::std::forward<std::ranges::iota_view<unsigned long, unsigned long>>(args), ::
std::forward<std::vector<oneapi::dpl::experimental::dr::sp::remote_vector<int, oneapi::dpl::experimental::dr::sp::device_allocator<int>>> &>(args)))'
 (aka 'zip_view<std::ranges::iota_view<unsigned long, unsigned long>, std::vector<oneapi::dpl::experimental::dr::sp::remote_vector<int, oneapi::dpl::experimental::dr::sp::device_allocator<int, 0>>, std::allocator<oneapi::dpl::experimental::dr::sp::remote_vector<int, oneapi::dpl::experimental::dr::sp::device_allocator<int, 0>>>> &>')
 and '_Partial<_Take, decay_t<unsigned long>>' 
(aka '_Partial<std::ranges::views::_Take, unsigned long>'))
   59 |     return make_enumerate(segments) | stdrng::views::take(last_seg + 1) |
```


lslusarczyk:
@akukanov , @MikeDvorskiy , here is my work in progress regarding making DR working C++-20 standard.
Fixing one problem reveals a new ones. I don't know when it will end, but if you think continuing this work is worth our effort, then we may continue and you're welcome to contribute.

lslusarczyk:
... maybe alternative is to officially say users to use ranges-v3 with oneDPL::DR if they need lower than 23 standard. This should work out of the box.

include/oneapi/dpl/internal/distributed_ranges_impl/sp/algorithms/sort.hpp
```diff
@@ -112,7 +112,7 @@ sort(R&& r, Compare comp = Compare())
 
     T* medians = sycl::malloc_device<T>(n_segments * n_splitters, sp::devices()[0], sp::context());
 
-    for (auto&& [segment_id_, segment] : stdrng::views::enumerate(segments))
+    for (auto&& [segment_id_, segment] : dr::__detail::make_enumerate(segments))
```
MikeDvorskiy, line 115: It turns out that we are using a functional approach for the sake of a functional approach...
I would suggest a little bit re-writing this loop is following way:

```
for (auto segment_id = 0, n =  stdrng::size(segment); segment_id < n; ++segment_id)
{
         auto segment = segments[segment_id ];
.......
```

lslusarczyk: @MikeDvorskiy , thanks for looking into this. Looks acceptable, although it is more lines of code. 

Are other places with enumarate also easily changeable? Do you think it is more clean than having `make_enumerate' or does it fixes some problems which `make_enumerate` has (if it has)?

Enumerate is quite common programming pattern making code shorter and more readable.

I don't object of changing all places where there is enumerate to some code like this if it is needed to add C++20 support. Please feel free to clone my change to your branch and continue adding any changes you think may be good for C++20 support.

akukanov:
Talked to Mikhail today. It sounds like our internal zip_view has some assumptions that complicate its use as a replacement for the standard one. Let's keep this patch as a reference - as I still think there should be our own C++20 zip_view for both "regular" and distributed range algorithms, - but I am fine to postpone resolution of this question for a later update.

PR #1765, opened 2024-08-05
Make Unique family of APIs use reduce_then_scan

danhoeflinger:
This PR changes the unique family of scan-like APIs to use reduce_then_scan when it is beneficial.

This PR allows us to remove `__pattern_scan_copy` functions because they are no longer used.  
We have moved the algorithm decisions to go through at the level of `__parallel_[copy_if/partition/unique]_copy`, so we no longer need the "scan_copy" at the pattern level.

Moves all algorithm selection decisions to `__parallel_unique_copy`, and unifies range API to also use this function. This allows us to unify the algorithmic selection, and provide performance improvements to the ranges API.

Unique requires some `constexpr` special casing in the kernel to allow us to avoid an extra branch for each element in the `_GenMask` for unique to avoid underflow when `index == 0`.  We special case the kernels to skip and always copy the 0th element for unique family APIs, and start the scan at element `1`.  This allows us to handle this copy of the 0th element without any additional kernel launches.  The `n==1` case is handled specially with a simple copy call.

---

This PR is targeted to #1764, to allow for a clean diff, and is a part of the following sequence of PRs meant to be merged in order:

#1769 [MERGED] ~~Relocate __lazy_ctor_storage to utils header~~
#1770 [MERGED] ~~Use __result_and_scratch_storage within scan kernels~~
#1762 Add reduce_then_scan algorithm for transform scan family
#1763 Make Copy_if family of APIs use reduce_then_scan algorithm
#1764 Make Partition family of APIs use reduce_then_scan algorithm 
#1765 Make Unique family of APIs use reduce_then_scan algorithm (This PR)

This work is a collaboration between @mmichel11 @adamfidel and @danhoeflinger, and based upon an original prototype by Ted Painter.

PR #1763, opened 2024-08-05
Make Copy_if family of APIs use reduce_then_scan algorithm

danhoeflinger:
This PR adds copy_if family to use reduce_then_scan algorithm where it is beneficial.

Moves all algorithm selection decisions to `__parallel_copy_if`, and unifies range API to also use this function.  Adding support for an assignment operator for single work group copy_if.  This allows us to unify the algorithmic selection, and provide performance improvements to the ranges API.

Adds `__parallel_reduce_then_scan_copy` function which will be used by partition and unique in the future to take advantage of shared functionality.

---

This PR is targeted to #1762, to allow for a clean diff, and is a part of the following sequence of PRs meant to be merged in order:

#1769 [MERGED] ~~Relocate __lazy_ctor_storage to utils header~~
#1770 [MERGED] ~~Use __result_and_scratch_storage within scan kernels~~
#1762 Add reduce_then_scan algorithm for transform scan family
#1763 Make Copy_if family of APIs use reduce_then_scan algorithm (This PR)
#1764 Make Partition family of APIs use reduce_then_scan algorithm
#1765 Make Unique family of APIs use reduce_then_scan algorithm

This work is a collaboration between @mmichel11 @adamfidel and @danhoeflinger, and based upon an original prototype by Ted Painter.

PR #1769, opened 2024-08-06
Relocate __lazy_ctor_storage to utils header

adamfidel:
One of the changes to the upcoming scan rework (#1762) is to move the helper union `__lazy_ctor_storage` out of the reduce header into the main utils header, as it will eventually be used in scan as well. 

This PR makes that movement, as well as adds a few helper methods to `__lazy_ctor_storage`.

---

This PR is a part of the following sequence of PRs meant to be merged in order:

#1769 Relocate __lazy_ctor_storage to utils header (This PR)
#1770 Use __result_and_scratch_storage within scan kernels
#1762 Add reduce_then_scan algorithm for transform scan family

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -47,14 +47,6 @@ class __reduce_mid_work_group_kernel;
 template <typename... _Name>
 class __reduce_kernel;
 
-// Storage helper since _Tp may not have a default constructor.
-template <typename _Tp>
-union __lazy_ctor_storage
-{
-    _Tp __v;
-    __lazy_ctor_storage() {}
-};
-
 // Adjust number of sequential operations per work-item based on the vector size. Single elements are kept to
 // improve performance of small arrays or remainder loops.
 template <std::uint8_t _VecSize, typename _Size>
@@ -76,7 +68,7 @@ __work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Si
 {
     auto __local_idx = __item_id.get_local_id(0);
     const _Size __group_size = __item_id.get_local_range().size();
-    __lazy_ctor_storage<_Tp> __result;
+    oneapi::dpl::__internal::__lazy_ctor_storage<_Tp> __result;
     // 1. Initialization (transform part). Fill local memory
     __transform_pattern(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0, __is_full,
                         /*__n_groups*/ (_Size)1, __result, __acc...);
@@ -89,7 +81,7 @@ __work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Si
         __reduce_pattern.apply_init(__init, __result.__v);
         __res_ptr[0] = __result.__v;
     }
-    __result.__v.~_Tp();
+    __result.__destroy();
 }
 
 // Device kernel that transforms and reduces __n elements to the number of work groups preliminary results.
@@ -104,7 +96,7 @@ __device_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Size _
     auto __local_idx = __item_id.get_local_id(0);
     auto __group_idx = __item_id.get_group(0);
     const _Size __group_size = __item_id.get_local_range().size();
-    __lazy_ctor_storage<_Tp> __result;
+    oneapi::dpl::__internal::__lazy_ctor_storage<_Tp> __result;
     // 1. Initialization (transform part). Fill local memory
     __transform_pattern(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0, __is_full, __n_groups,
                         __result, __acc...);
@@ -114,7 +106,7 @@ __device_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Size _
     __result.__v = __reduce_pattern(__item_id, __n_items, __result.__v, __local_mem);
     if (__local_idx == 0)
         __temp_acc[__group_idx] = __result.__v;
-    __result.__v.~_Tp();
+    __result.__destroy();
 }
 
 //------------------------------------------------------------------------
@@ -394,7 +386,7 @@ struct __parallel_transform_reduce_impl
                         // 1. Initialization (transform part). Fill local memory
                         _Size __n_items;
                         const bool __is_full = __n == __size_per_work_group * __n_groups;
-                        __lazy_ctor_storage<_Tp> __result;
+                        oneapi::dpl::__internal::__lazy_ctor_storage<_Tp> __result;
                         if (__is_first)
                         {
                             __transform_pattern1(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0,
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -47,14 +47,6 @@ class __reduce_mid_work_group_kernel;
 template <typename... _Name>
 class __reduce_kernel;
 
-// Storage helper since _Tp may not have a default constructor.
```
julianmi, line 50: Should we keep a comment similar to this one to clarify what this helper does? E.g., "Storage helper to support all types including ones without default constructor.".

include/oneapi/dpl/pstl/utils.h
```diff
@@ -765,6 +765,24 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
 static constexpr bool __is_iterator_type_v = __is_iterator_type<_T>::value;
 
+template <typename _Tp>
+union __lazy_ctor_storage
+{
+    using __value_type = _Tp;
+    _Tp __v;
+    __lazy_ctor_storage() {}
+    void
+    __setup(const _Tp& init)
```
mmichel11, line 775: Since we have this function now, I think we can replace the placement news in the `transform_reduce` struct in `unseq_backend_sycl.h` to call this function for consistency.

danhoeflinger: Oh yes, this was part of the original changes in the transform_scan branch,  I didn't see those were missing here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
adamfidel: Yes, I added this comment back.

include/oneapi/dpl/pstl/utils.h
adamfidel: Good catch. I made those changes.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -213,7 +213,7 @@ struct transform_reduce
     void
     vectorized_reduction_first(const _Size __start_idx, _Res& __res, const _Acc&... __acc) const
     {
-        new (&__res.__v) _Tp(__unary_op(__start_idx, __acc...));
```
danhoeflinger, line 216: Do we need the `_Tp()` here?  I think that should be redundant.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -213,7 +213,7 @@ struct transform_reduce
...
+        __res.__setup(_Tp(__unary_op(__start_idx, __acc...)));
```
danhoeflinger, line 216: Do we need the `_Tp()` here?  I think that should be redundant.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -765,6 +765,25 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
 
+// Storage helper since _Tp may not have a default constructor.
+template <typename _Tp>
+union __lazy_ctor_storage
+{
+    using __value_type = _Tp;
+    _Tp __v;
+    __lazy_ctor_storage() {}
+    void
+    __setup(const _Tp& init)
```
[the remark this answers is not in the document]
danhoeflinger: yeah, good point, we should probably take a forwarding reference and forward it along into the ctor.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
adamfidel: You are correct; they are redundant and have been removed.

include/oneapi/dpl/pstl/utils.h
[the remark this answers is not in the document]
adamfidel: I wasn't sure if a union can have a templated member method but it appears that they can, so I changed the formal parameter to a forwarding reference.

SergeyKopienko:
One comment about `__result.__destroy();`
In common case we have this call as latest in some functions.
The problem here is that in case of exception throwing before this code we never execute this line and theoretically may have memory leak.

Probably we may have some helper on stack which will execute this code on it's destructor.

danhoeflinger:
> One comment about `__result.__destroy();` In common case we have this call as latest in some functions. The problem here is that in case of exception throwing before this code we never execute this line and theoretically may have memory leak.
> 
> Probably we may have some helper on stack which will execute this code on it's destructor.

Its a good point, but I'm not sure how to handle it really. Also, I think the issue is more about not destroying a constructed type than a memory leak.  For some types with dynamically allocated memory, not destroying could theoretically lead to a memory leak, but not for simple types.  For simple types, the memory would be deallocated as the `__lazy_ctor_storage` object leaves scope.

The problem is we do not know if it has been setup / constructed at the time of the exception. The point of the structure is to allow us to separate the scope from the construction / destruction without extra memory to record this at runtime. We rely upon the structure of the algorithm to provide the knowledge of if it is constructed.

julianmi (approved):
LGTM

adamfidel:
> > One comment about `__result.__destroy();` In common case we have this call as latest in some functions. The problem here is that in case of exception throwing before this code we never execute this line and theoretically may have memory leak.
> > Probably we may have some helper on stack which will execute this code on it's destructor.
> 
> Its a good point, but I'm not sure how to handle it really. Also, I think the issue is more about not destroying a constructed type than a memory leak. For some types with dynamically allocated memory, not destroying could theoretically lead to a memory leak, but not for simple types. For simple types, the memory would be deallocated as the `__lazy_ctor_storage` object leaves scope.
> 
> The problem is we do not know if it has been setup / constructed at the time of the exception. The point of the structure is to allow us to separate the scope from the construction / destruction without extra memory to record this at runtime. We rely upon the structure of the algorithm to provide the knowledge of if it is constructed.

I agree with @danhoeflinger here. Additionally, the level of exception safety isn't really changing with this PR as we are ultimately not introducing anything new that wasn't in the code before -- just reorganizing a bit.
 
@SergeyKopienko are you okay with these changes as they are and tackling the case you brought up in a separate issue / PR?

SergeyKopienko:
> > > One comment about `__result.__destroy();` In common case we have this call as latest in some functions. The problem here is that in case of exception throwing before this code we never execute this line and theoretically may have memory leak.
> > > Probably we may have some helper on stack which will execute this code on it's destructor.
> > 
> > 
> > Its a good point, but I'm not sure how to handle it really. Also, I think the issue is more about not destroying a constructed type than a memory leak. For some types with dynamically allocated memory, not destroying could theoretically lead to a memory leak, but not for simple types. For simple types, the memory would be deallocated as the `__lazy_ctor_storage` object leaves scope.
> > The problem is we do not know if it has been setup / constructed at the time of the exception. The point of the structure is to allow us to separate the scope from the construction / destruction without extra memory to record this at runtime. We rely upon the structure of the algorithm to provide the knowledge of if it is constructed.
> 
> I agree with @danhoeflinger here. Additionally, the level of exception safety isn't really changing with this PR as we are ultimately not introducing anything new that wasn't in the code before -- just reorganizing a bit.
> 
> @SergeyKopienko are you okay with these changes as they are and tackling the case you brought up in a separate issue / PR?

I understood and completely agree. Thanks for these explanations.

SergeyKopienko (approved):
LGTM

PR #1770, opened 2024-08-06
Use __result_and_scratch_storage within scan kernels

danhoeflinger:
This PR adjust existing scan kernels to use `__result_and_scratch_storage` rather than a temporary buffer for both scratch space and for the return within the resulting future.

This is important because: 
1) It should be faster (as evidenced from previous work with reduce)
2) It serves as plumbing in preparation for the upcoming `reduce_then_scan` kernels which utilize this structure for temporary and return storage.  Since we have runtime branches selecting which algorithm to use (existing scan, single wg scan, single wg copy_if, and soon reduce_then_scan), these all need to have the same return future type.  This means they should all use `__result_and_scratch_storage` for their return (and therefore also any temporary storage).

Single WG scan does not have a return or temporary storage, so it simply creates a dummy `__result_and_scratch_storage` to make a consistent type for the return, and no changes are needed in the kernel.

For copy_if, unique and partition scan kernels, we need to write to the output result storage.  Previous to to PR, we simply used the last element of the temporary storage buffer in the returned future.

This PR also avoids double dereferences, by passing pointers around and using them as pointers, rather than as const lvalue references to "accessors".  (as mentioned in #1751)  Until this PR, these actually were accessors, but now they are pointers.

---

This PR is targeted to #1769, to allow for a clean diff, and is a part of the following sequence of PRs meant to be merged in order:

#1769 [MERGED] Relocate __lazy_ctor_storage to utils header  
#1770 Use __result_and_scratch_storage within scan kernels (This PR)
#1762 Add reduce_then_scan algorithm for transform scan family

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -835,23 +835,21 @@ struct __invoke_single_group_copy_if
         if (__is_full_group)
             return __par_backend_hetero::__parallel_copy_if_static_single_group_submitter<
                 _SizeType, __num_elems_per_item, __wg_size, true,
-                   oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-                    __scan_copy_single_wg_kernel<::std::integral_constant<::std::uint16_t, __wg_size>,
-                                                ::std::integral_constant<::std::uint16_t, __num_elems_per_item>,
-                                                /* _IsFullGroup= */ std::true_type, _CustomName>>
-                >()(
-                __exec, ::std::forward<_InRng>(__in_rng), ::std::forward<_OutRng>(__out_rng), __n, _InitType{},
-                _ReduceOp{}, ::std::forward<_Pred>(__pred));
+                oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
+                    __scan_copy_single_wg_kernel<std::integral_constant<std::uint16_t, __wg_size>,
+                                                 std::integral_constant<std::uint16_t, __num_elems_per_item>,
+                                                 /* _IsFullGroup= */ std::true_type, _CustomName>>>()(
+                std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
+                std::forward<_OutRng>(__out_rng), __n, _InitType{}, _ReduceOp{}, std::forward<_Pred>(__pred));
         else
             return __par_backend_hetero::__parallel_copy_if_static_single_group_submitter<
                 _SizeType, __num_elems_per_item, __wg_size, false,
-                   oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-                    __scan_copy_single_wg_kernel<::std::integral_constant<::std::uint16_t, __wg_size>,
-                                                ::std::integral_constant<::std::uint16_t, __num_elems_per_item>,
-                                                /* _IsFullGroup= */ std::false_type, _CustomName>>
-                >()(
-                __exec, ::std::forward<_InRng>(__in_rng), ::std::forward<_OutRng>(__out_rng), __n, _InitType{},
-                _ReduceOp{}, ::std::forward<_Pred>(__pred));
+                oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
+                    __scan_copy_single_wg_kernel<std::integral_constant<std::uint16_t, __wg_size>,
+                                                 std::integral_constant<std::uint16_t, __num_elems_per_item>,
+                                                 /* _IsFullGroup= */ std::false_type, _CustomName>>>()(
+                std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
+                std::forward<_OutRng>(__out_rng), __n, _InitType{}, _ReduceOp{}, std::forward<_Pred>(__pred));
     }
 };
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -833,23 +833,30 @@ struct __invoke_single_group_copy_if
         using _InitType = unseq_backend::__no_init_value<::std::uint16_t>;
         using _ReduceOp = ::std::plus<::std::uint16_t>;
         if (__is_full_group)
+        {
+            using _FullKernel =
+                __scan_copy_single_wg_kernel<std::integral_constant<std::uint16_t, __wg_size>,
+                                             std::integral_constant<std::uint16_t, __num_elems_per_item>,
+                                             /* _IsFullGroup= */ std::true_type, _CustomName>;
+            using _FullKernelName = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<_FullKernel>;
             return __par_backend_hetero::__parallel_copy_if_static_single_group_submitter<
-                _SizeType, __num_elems_per_item, __wg_size, true,
-                oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-                    __scan_copy_single_wg_kernel<std::integral_constant<std::uint16_t, __wg_size>,
-                                                 std::integral_constant<std::uint16_t, __num_elems_per_item>,
-                                                 /* _IsFullGroup= */ std::true_type, _CustomName>>>()(
+                _SizeType, __num_elems_per_item, __wg_size, true, _FullKernelName>()(
                 std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
                 std::forward<_OutRng>(__out_rng), __n, _InitType{}, _ReduceOp{}, std::forward<_Pred>(__pred));
+        }
         else
+        {
+            using _NonFullKernel =
+                __scan_copy_single_wg_kernel<std::integral_constant<std::uint16_t, __wg_size>,
+                                             std::integral_constant<std::uint16_t, __num_elems_per_item>,
+                                             /* _IsFullGroup= */ std::false_type, _CustomName>;
+            using _NonFullKernelName =
+                oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<_NonFullKernel>;
             return __par_backend_hetero::__parallel_copy_if_static_single_group_submitter<
-                _SizeType, __num_elems_per_item, __wg_size, false,
-                oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-                    __scan_copy_single_wg_kernel<std::integral_constant<std::uint16_t, __wg_size>,
-                                                 std::integral_constant<std::uint16_t, __num_elems_per_item>,
-                                                 /* _IsFullGroup= */ std::false_type, _CustomName>>>()(
+                _SizeType, __num_elems_per_item, __wg_size, false, _NonFullKernelName>()(
                 std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
                 std::forward<_OutRng>(__out_rng), __n, _InitType{}, _ReduceOp{}, std::forward<_Pred>(__pred));
+        }
     }
 };
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -310,8 +310,8 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_wg);
         // Storage for the results of scan for each workgroup
 
-        using _TempStorage = __result_and_scratch_storage<std::decay_t<_ExecutionPolicy>, _Type>;
-        _TempStorage __result_and_scratch{__exec, __n_groups + 1};
+        using __result_and_scratch_storage_t = __result_and_scratch_storage<_ExecutionPolicy, _Type>;
+        __result_and_scratch_storage_t __result_and_scratch{__exec, __n_groups + 1};
 
         _PRINT_INFO_IN_DEBUG_MODE(__exec, __wgroup_size, __max_cu);
 
@@ -328,7 +328,7 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
                 __kernel_1,
 #endif
                 sycl::nd_range<1>(__n_groups * __wgroup_size, __wgroup_size), [=](sycl::nd_item<1> __item) {
-                    auto __temp_ptr = _TempStorage::__get_usm_or_buffer_accessor_ptr(__temp_acc);
+                    auto __temp_ptr = __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__temp_acc);
                     __local_scan(__item, __n, __local_acc, __rng1, __rng2, __temp_ptr, __size_per_wg, __wgroup_size,
                                  __iters_per_witem, __init);
                 });
@@ -350,7 +350,7 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
 #endif
                     // TODO: try to balance work between several workgroups instead of one
                     sycl::nd_range<1>(__wgroup_size, __wgroup_size), [=](sycl::nd_item<1> __item) {
-                        auto __temp_ptr = _TempStorage::__get_usm_or_buffer_accessor_ptr(__temp_acc);
+                        auto __temp_ptr = __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__temp_acc);
                         __group_scan(__item, __n_groups, __local_acc, __temp_ptr, __temp_ptr,
                                      /*dummy*/ __temp_ptr, __n_groups, __wgroup_size, __iters_per_single_wg);
                     });
@@ -364,8 +364,9 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
             auto __temp_acc = __result_and_scratch.__get_scratch_acc(__cgh);
             auto __res_acc = __result_and_scratch.__get_result_acc(__cgh);
             __cgh.parallel_for<_PropagateScanName...>(sycl::range<1>(__n_groups * __size_per_wg), [=](auto __item) {
-                auto __temp_ptr = _TempStorage::__get_usm_or_buffer_accessor_ptr(__temp_acc);
-                auto __res_ptr = _TempStorage::__get_usm_or_buffer_accessor_ptr(__res_acc, __n_groups + 1);
+                auto __temp_ptr = __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__temp_acc);
+                auto __res_ptr =
+                    __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__res_acc, __n_groups + 1);
                 __global_scan(__item, __rng2, __rng1, __temp_ptr, __res_ptr, __n, __size_per_wg);
             });
         });
@@ -566,8 +567,8 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
             typename ::std::decay_t<decltype(__in_rng[0])>, typename ::std::decay_t<decltype(__out_rng[0])>>::__type;
 
         constexpr ::std::uint32_t __elems_per_wg = _ElemsPerItem * _WGSize;
-
-        __result_and_scratch_storage<std::decay_t<_Policy>, _Size> __result{__policy, 0};
+        using __result_and_scratch_storage_t = __result_and_scratch_storage<_Policy, _Size>;
+        __result_and_scratch_storage_t __result{__policy, 0};
 
         auto __event = __policy.queue().submit([&](sycl::handler& __hdl) {
             oneapi::dpl::__ranges::__require_access(__hdl, __in_rng, __out_rng);
@@ -580,8 +581,7 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
 
             __hdl.parallel_for<_ScanKernelName...>(
                 sycl::nd_range<1>(_WGSize, _WGSize), [=](sycl::nd_item<1> __self_item) {
-                    auto __res_ptr =
-                        __result_and_scratch_storage<_Policy, _Size>::__get_usm_or_buffer_accessor_ptr(__res_acc);
+                    auto __res_ptr = __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__res_acc);
                     const auto& __group = __self_item.get_group();
                     const auto& __subgroup = __self_item.get_sub_group();
                     // This kernel is only launched for sizes less than 2^16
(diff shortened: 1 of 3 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -380,15 +361,16 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         auto __final_event = __exec.queue().submit([&](sycl::handler& __cgh) {
             __cgh.depends_on(__submit_event);
```
SergeyKopienko, line 362: From my point of view, capture the source event by reference is not quite correct.
Here and in other places.
Better to capture by value.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -654,11 +636,11 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
                     if (__item_id == 0)
                     {
                         // Add predicate of last element to account for the scan's exclusivity
-                        __res_acc[0] = __lacc[__elems_per_wg + __n - 1] + __lacc[__n - 1];
+                        __res_ptr[0] = __lacc[__elems_per_wg + __n - 1] + __lacc[__n - 1];
```
SergeyKopienko, line 639: It's correct. But may be better to write `*__res_ptr` as far as it is simple pointer?
Here and in other places...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -706,6 +696,7 @@ __parallel_transform_scan_single_group(oneapi::dpl::__internal::__device_backend
                         /* _IsFullGroup= */ ::std::false_type, _Inclusive, _CustomName>>>()(
                     ::std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
                     std::forward<_OutRng>(__out_rng), __n, __init, __binary_op, __unary_op);
+            return __future(__event, __dummy_result_and_scratch);
```
SergeyKopienko, line 699:
```suggestion
            return __future(__event, _TempStorage{__exec, 0});
```

SergeyKopienko: In this case `_TempStorage __dummy_result_and_scratch{__exec, 0};` will not required anymore.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -735,9 +726,11 @@ __parallel_transform_scan_single_group(oneapi::dpl::__internal::__device_backend
...
             __par_backend_hetero::__scan_single_wg_dynamic_kernel<_BinaryOperation, _CustomName>>;
 
-        return __parallel_transform_scan_dynamic_single_group_submitter<_Inclusive::value, _DynamicGroupScanKernel>()(
-            ::std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng), std::forward<_OutRng>(__out_rng),
-            __n, __init, __binary_op, __unary_op, __max_wg_size);
+        auto __event =
+            __parallel_transform_scan_dynamic_single_group_submitter<_Inclusive::value, _DynamicGroupScanKernel>()(
+                std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
+                std::forward<_OutRng>(__out_rng), __n, __init, __binary_op, __unary_op, __max_wg_size);
+        return __future(__event, __dummy_result_and_scratch);
```
SergeyKopienko, line 733: The same

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -547,7 +547,8 @@ struct __result_and_scratch_storage
     }
 
   public:
-    __result_and_scratch_storage(_ExecutionPolicy& __exec, ::std::size_t __scratch_n)
+    template <typename _Policy>
+    __result_and_scratch_storage(_Policy&& __exec, ::std::size_t __scratch_n)
```
SergeyKopienko, line 551: What if we write `__scratch_n = 0` and remove `0` as second param in all instantiations?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Do you mean the capture for the lambda being submitted here to capture `__submit_event`?
This is not new code, and is also how most submit calls in the library are done.  

If my understanding is correct, I think this is outside the scope of this PR if we want to make any change to that.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -617,6 +624,11 @@ struct __copy_by_mask
                 // is performed(i.e. __typle_type is the same type as its operand).
                 __assigner(static_cast<__tuple_type>(get<0>(__in_acc[__item_idx])), __out_acc[__out_idx]);
         }
+        if (__item_idx == 0)
+        {
+            //copy final result to output
+            __ret_ptr[0] = __wg_sums_ptr[(__n - 1) / __size_per_wg];
```
SergeyKopienko, line 630: `*__ret_ptr'

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -656,10 +668,15 @@ struct __partition_by_mask
...
                 if (__not_first_wg)
-                    __out_idx -= __wg_sums_acc[__wg_sums_idx - 1];
+                    __out_idx -= __wg_sums_ptr[__wg_sums_idx - 1];
                 get<1>(__out_acc[__out_idx]) = static_cast<__tuple_type>(get<0>(__in_acc[__item_idx]));
             }
         }
+        if (__item_idx == 0)
+        {
+            //copy final result to output
+            __ret_ptr[0] = __wg_sums_ptr[(__n - 1) / __size_per_wg];
```
SergeyKopienko, line 678: The same

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: ok.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I disagree with this suggestion, because while we do save a line of code, we lose the ability to name it `__dummy_result_and_scratch`. 

I think this is descriptive, and helpful to show that this is unused to the future reader.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: see above

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: I think I also prefer to keep this as is, to help code using `__result_and_scratch_storage` make it more clear that they are not using the "scratch space" portion of this structure, but I don't have strong opinions on it.  

I'm interested in @julianmi  opinion here also.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: done.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
danhoeflinger: done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -596,10 +576,12 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
             // predicate on each element of the input range. The second half stores the index of the output
             // range to copy elements of the input range.
             auto __lacc = __dpl_sycl::__local_accessor<_ValueType>(sycl::range<1>{__elems_per_wg * 2}, __hdl);
-            auto __res_acc = __res.template get_access<access_mode::write>(__hdl);
+            auto __res_acc = __result.__get_result_acc(__hdl);
 
             __hdl.parallel_for<_ScanKernelName...>(
                 sycl::nd_range<1>(_WGSize, _WGSize), [=](sycl::nd_item<1> __self_item) {
+                    auto __res_ptr =
+                        __result_and_scratch_storage<_Policy, _Size>::__get_usm_or_buffer_accessor_ptr(__res_acc);
```
SergeyKopienko, line 584: At line 570 you wrote
`__result_and_scratch_storage<std::decay_t<_Policy>, _Size> __result{__policy, 0};`

but at this line `std::decay_t` is absent.

Which variant is correct?

SergeyKopienko:
I believe we should pass `__scratch_container` param by reference at
https://github.com/oneapi-src/oneDPL/blob/aec12a4c55ac5de94753f8e3259b50cd6befb0c3/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h#L202

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
julianmi: I tend to agree that the current version (without the default value) is clearer.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -332,14 +309,16 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         auto __size_per_wg = __iters_per_witem * __wgroup_size;
         auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_wg);
         // Storage for the results of scan for each workgroup
-        sycl::buffer<_Type> __wg_sums(__n_groups);
+
+        using _TempStorage = __result_and_scratch_storage<std::decay_t<_ExecutionPolicy>, _Type>;
```
julianmi, line 313: Why is `std::decay_t` required?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -677,6 +659,13 @@ __parallel_transform_scan_single_group(oneapi::dpl::__internal::__device_backend
     // Specialization for devices that have a max work-group size of 1024
     constexpr ::std::uint16_t __targeted_wg_size = 1024;
 
+    using _ValueType = typename _InitType::__value_type;
+
+    // Although we do not actually need result storage in this case, we need to construct
```
julianmi, line 664: I have a prototype branch `dev/julianmi/scratch_storage` that supports cases where we don't need a result but want to keep the same return type. Would this be applicable here?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -547,7 +547,8 @@ struct __result_and_scratch_storage
     }
 
   public:
-    __result_and_scratch_storage(_ExecutionPolicy& __exec, ::std::size_t __scratch_n)
+    template <typename _Policy>
```
julianmi, line 550: Why does this policy need to be different to `_ExecutionPolicy`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: ok, let's save as is

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Yes, I believe this is what we would like here. I've not looked in full detail at the branch, but on first glance it looks like what we would want.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I think the decay is all unnecessary.  Working to confirm this and discussing with the others who contributed here who may have a better understanding of why this was introduced.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: I think this is more about being able to accept rvalue references to policies.  However, I'm not sure it is specifically necessary in any of the subsequent PRs.  

I think either we need to forward it into the `__exec` constructor, or accept a `const _ExecutionPolicy&`, which will enable all variety of argument.  

Also, I think we should rename the parameter to be distinct from the member variable.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
julianmi: Ok, I'll productize it then and open a PR.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I don't think it is.  We added it at one point to rectify differences in types between future returns, but I don't think it is necessary, especially with the proposed changes to the `__result_and_scratch_storage` ctor to use a const lvalue ref policy.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: I see that it's fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: I propose to use a `const _ExecutionPolicy&` here instead.  Any copies required will happen once on entry to this constructor.  This should allow incoming rvalue policies, (or any other policy reference flavor).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Thanks!  We will review it and rebase upon it once its merged

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I've made the changes to remove the decays

danhoeflinger:
> I believe we should pass `__scratch_container` param by reference at
> 
> https://github.com/oneapi-src/oneDPL/blob/aec12a4c55ac5de94753f8e3259b50cd6befb0c3/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h#L202
> 
> 
> Due it's deep copy is heavy (smart-pointers inside and so on).
> And may be the same in `__parallel_transform_reduce_work_group_kernel_submitter::operator` - pass `__scratch_container` by reference or moreover as `r-value` due it's never used after this call on caller side.

It is a deep copy in the location you mention, although I don't think its that expensive since it is only smart pointers, not a copy of the data they point to.  In that case, I think we do use it after the fact (to send it to the `__future`), so an rvalue isn't appropriate until the `__future` itself.
I'm not sure its in the scope of this PR to change that to a reference, but I'm happy to add it if we have consensus on it (@julianmi, etc.).  

The only deep copies we have in this PR are into `__future`, and we could possible `std::move` into the future, but changes would be needed in `__future` to accommodate that. 
@julianmi, what do you think of that change?

SergeyKopienko:
> > I believe we should pass `__scratch_container` param by reference at
> > https://github.com/oneapi-src/oneDPL/blob/aec12a4c55ac5de94753f8e3259b50cd6befb0c3/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h#L202
> > 
> > Due it's deep copy is heavy (smart-pointers inside and so on).
> > And may be the same in `__parallel_transform_reduce_work_group_kernel_submitter::operator` - pass `__scratch_container` by reference or moreover as `r-value` due it's never used after this call on caller side.
> 
> It is a deep copy in the location you mention, although I don't think its that expensive since it is only smart pointers, not a copy of the data they point to. In that case, I think we do use it after the fact (to send it to the `__future`), so an rvalue isn't appropriate until the `__future` itself. I'm not sure its in the scope of this PR to change that to a reference, but I'm happy to add it if we have consensus on it (@julianmi, etc.).
> 
> The only deep copies we have in this PR are into `__future`, and we could possible `std::move` into the future, but changes would be needed in `__future` to accommodate that. @julianmi, what do you think of that change? (edit: I'm not sure if `__future` changes are required... investigating)

Exactly, under deep copy I mean copy of smart-pointers here with all atomic counter operations in their copy implementations. But we are able to avoid these at all.

julianmi:
> > I believe we should pass `__scratch_container` param by reference at
> > https://github.com/oneapi-src/oneDPL/blob/aec12a4c55ac5de94753f8e3259b50cd6befb0c3/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h#L202
> > 
> > Due it's deep copy is heavy (smart-pointers inside and so on).
> > And may be the same in `__parallel_transform_reduce_work_group_kernel_submitter::operator` - pass `__scratch_container` by reference or moreover as `r-value` due it's never used after this call on caller side.
> 
> It is a deep copy in the location you mention, although I don't think its that expensive since it is only smart pointers, not a copy of the data they point to. In that case, I think we do use it after the fact (to send it to the `__future`), so an rvalue isn't appropriate until the `__future` itself. I'm not sure its in the scope of this PR to change that to a reference, but I'm happy to add it if we have consensus on it (@julianmi, etc.).
> 
> The only deep copies we have in this PR are into `__future`, and we could possible `std::move` into the future, but changes would be needed in `__future` to accommodate that. @julianmi, what do you think of that change? (edit: I'm not sure if `__future` changes are required... investigating)

Deep copy: I agree that it is probably not that expensive but still unneeded. I will change it in a separate PR. Let's focus on getting the reduce-then-scan implementation in. 

`std::move`: I think this might be useful. This is probably also out of scope for this PR. I will investigate and prepare a patch.

SergeyKopienko (dismissed):
LGTM

danhoeflinger:
> Deep copy: I agree that it is probably not that expensive but still unneeded. I will change it in a separate PR. Let's focus on getting the reduce-then-scan implementation in.
> 
> `std::move`: I think this might be useful. This is probably also out of scope for this PR. I will investigate and prepare a patch.

Thanks for taking a look at these.  
Here is a godbolt starting to look at the future changes which could be helpful.  https://godbolt.org/z/WvqK7G37z
I think we can `std::move` right now, and it wont cause errors, but we also wont save a copy with the current implementation.  I think we can save a copy with something like what is in the godbolt, but it requires explicit template parameter specification of the `__future` type.  This is not necessarily a bad thing...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -844,23 +835,21 @@ struct __invoke_single_group_copy_if
...
-                                                /* _IsFullGroup= */ std::false_type, _CustomName>>
-                >()(
-                __exec, ::std::forward<_InRng>(__in_rng), ::std::forward<_OutRng>(__out_rng), __n, _InitType{},
-                _ReduceOp{}, ::std::forward<_Pred>(__pred));
+                oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
+                    __scan_copy_single_wg_kernel<std::integral_constant<std::uint16_t, __wg_size>,
+                                                 std::integral_constant<std::uint16_t, __num_elems_per_item>,
+                                                 /* _IsFullGroup= */ std::false_type, _CustomName>>>()(
+                std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
+                std::forward<_OutRng>(__out_rng), __n, _InitType{}, _ReduceOp{}, std::forward<_Pred>(__pred));
```
danhoeflinger, line 852: I had to fix a missing `std::forward` of the execution policy here, otherwise in situations where we had a rvalue policy type, we do in fact encounter compiler errors of mismatched return type between branches.  

I believe this was the original source of the `decay_t` calls earlier, but they are unnecessary with this change. 

There may be a situation in the future, where such a decay of policy type is required for `__result_and_scratch_storage`  if different runtime paths may result in different handling of forwarding policies.

julianmi (dismissed):
LGTM

danhoeflinger:
Waiting to merge this until #1769 is merged, to avoid rebase complexity, unless that PR gets stuck for a long time.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: One more moment.
What about extract `__kernel_name_provider` here into some extra type?

```C++
        template <bool _IsFullGroup>
        using __kernel_name_provider_t =
            oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__scan_copy_single_wg_kernel<
                std::integral_constant<std::uint16_t, __wg_size>,
                std::integral_constant<std::uint16_t, __num_elems_per_item>, _IsFullGroup, _CustomName>>;

        if (__is_full_group)
            return __par_backend_hetero::__parallel_copy_if_static_single_group_submitter<
                _SizeType, __num_elems_per_item, __wg_size, true,
                __kernel_name_provider_t</*_IsFullGroup*/ std::true_type>>()(
                std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
                std::forward<_OutRng>(__out_rng), __n, _InitType{}, _ReduceOp{}, std::forward<_Pred>(__pred));
        else
            return __par_backend_hetero::__parallel_copy_if_static_single_group_submitter<
                _SizeType, __num_elems_per_item, __wg_size, false,
                __kernel_name_provider_t</*_IsFullGroup*/ std::false_type>>()(
                std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
                std::forward<_OutRng>(__out_rng), __n, _InitType{}, _ReduceOp{}, std::forward<_Pred>(__pred));
```

danhoeflinger: It doesn't get much that much better without extra external template structures, but I've pulled out a couple layers to make it a bit easier to read.  What do you think?

SergeyKopienko: I like your changes here. OK.

julianmi (approved):

SergeyKopienko (approved):
LGTM

PR #1771, opened 2024-08-06
Workaround Windows correctness issue in reduce_by_segment.pass

mmichel11:
This PR is put up in anticipation of the reduce-then-scan implementation being committed. It can currently be produced on `dev/shared/reduce_than_scan_impl`.

When using the new reduce-then-scan implementation, a correctness issue has been identified on Windows with `reduce_by_segment.pass.cpp`. The issue can only be produced with the current ordering of test cases (and I assume ordering in which kernels are compiled). After running the two flag predicate tests, the third `MatrixPoint` test return incorrect results. We run into this issue even if the flag predicate tests are not run and their kernels only compiled. Reordering the tests to run the `MatrixPoint` test prior to the flag predicate tests, or completely removing the flag predicate tests results in a pass.

We believe this issue related to the order of kernel compilation is a compiler bug and are following up with the compiler team. For now, I have added a broken macro to reorder the tests on Windows.

### DIFF
--- a/test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
+++ b/test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
@@ -346,13 +346,22 @@ run_test()
 int
 main()
 {
+    // On Windows, we observe incorrect results with this test with a specific compilation order of the
+    // kernels. This is being filed to the compiler team. In the meantime, we can rearrange this test
+    // to resolve the issue on our side.
+#if _PSTL_RED_BY_SEG_WINDOWS_COMPILE_ORDER_BROKEN
+    run_test<MatrixPoint<float>, UserBinaryPredicate<MatrixPoint<float>>, MaxFunctor<MatrixPoint<float>>>();
+#endif
+
 #if TEST_DPCPP_BACKEND_PRESENT
     // test with flag pred
     test_flag_pred<sycl::usm::alloc::device, class KernelName1, std::uint64_t>();
     test_flag_pred<sycl::usm::alloc::device, class KernelName2, MatrixPoint<float>>();
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
+#if !_PSTL_RED_BY_SEG_WINDOWS_COMPILE_ORDER_BROKEN
     run_test<MatrixPoint<float>, UserBinaryPredicate<MatrixPoint<float>>, MaxFunctor<MatrixPoint<float>>>();
+#endif
 
     run_test<int, ::std::equal_to<int>, ::std::plus<int>>();
     run_test<float, ::std::equal_to<float>, ::std::plus<float>>();
--- a/test/support/test_config.h
+++ b/test/support/test_config.h
@@ -224,6 +224,7 @@
 // A specific kernel compilation order causes incorrect results on Windows with the DPCPP backend. For now, we reorder
 // the test while the issue is being reported to the compiler team. Once it is resolved, this macro can be removed
 // or limited to older compiler versions.
-#define _PSTL_RED_BY_SEG_WINDOWS_COMPILE_ORDER_BROKEN (_MSC_VER && TEST_DPCPP_BACKEND_PRESENT)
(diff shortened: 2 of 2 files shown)


test/support/test_config.h
```diff
@@ -221,4 +221,9 @@
 #    define _PSTL_ICPX_FPGA_TEST_USM_VECTOR_ITERATOR_BROKEN 0
 #endif
 
+// A specific kernel compilation order causes incorrect results on Windows with the DPCPP backend. For now, we reorder
+// the test while the issue is being reported to the compiler team. Once it is resolved, this macro can be removed
+// or limited to older compiler versions.
+#define _PSTL_RED_BY_SEG_WINDOWS_COMPILE_ORDER_BROKEN (_MSC_VER && TEST_DPCPP_BACKEND_PRESENT)
```
danhoeflinger, line 227: Does the compiler version matter here or does it effect all versions of the compiler?

SergeyKopienko: In case if it effect all versions of the compiler, probably better to use compiler version too as newest version of existing compilers. The goal - to recheck it again on newest versions later.

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
```diff
@@ -346,13 +346,22 @@ run_test()
 int
 main()
 {
+    // On Windows, we observe incorrect results with this test with a specific compilation order of the
```
julianmi, line 349: I wonder whether we even need this logic. Was there a specific reason for ordering these tests in the current way? If not, couldn't we simply move the `MatrixPoint` test in general?

danhoeflinger: Sure, we could do this, but it is not very satisfying to move tests merely because it suppresses an error, it feels like sweeping dirt under the rug.  The advantage of this is that it records that there is an error, and the situation where it exists.  Obviously, we are also planning to pursue it with the compiler team as well.

I don't think there is anything stopping us from just changing the order of the tests across the board, but that would lose the information and ability to reproduce the issue via these tests in the future.

julianmi: That makese sense, thanks.

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
```diff
@@ -346,13 +346,22 @@ run_test()
...
+    // kernels. This is being filed to the compiler team. In the meantime, we can rearrange this test
+    // to resolve the issue on our side.
+#if _PSTL_RED_BY_SEG_WINDOWS_COMPILE_ORDER_BROKEN
```
SergeyKopienko, line 352: Is it OK here that we wrote this statement without negation?

mmichel11: Yes, we want to still keep the test enabled but if the macro is set, then we want to place it above flag predicate tests which works around the issue. This is why we have `#if _PSTL_RED_BY_SEG_WINDOWS_COMPILE_ORDER_BROKEN` here and `#if !_PSTL_RED_BY_SEG_WINDOWS_COMPILE_ORDER_BROKEN` below.

test/support/test_config.h
mmichel11: Added the compiler version check here. The crash I mentioned with icpx 2024.0 is likely due to outdated drivers on my local Windows setup.

I have used a fresh system with up-to-date drivers and can reproduce with icpx 2024.0 as well.

mmichel11:
I've added this to the milestone even though it's a test change, since it will be important for testing of the milestone.

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
SergeyKopienko: ok

SergeyKopienko (approved):
LGTM

danhoeflinger (approved):
LGTM

julianmi (approved):

PR #1772, opened 2024-08-08
Update scan kernel template documentation to note single-work-group is asynchronous

adamfidel:
After revisiting the code for the scan kernel template, we discovered that the algorithm is only synchronous in the non-single-work-group case. The documentation should be amended slightly to note of this discrepancy.

### DIFF
--- a/documentation/library_guide/kernel_templates/single_pass_scan.rst
+++ b/documentation/library_guide/kernel_templates/single_pass_scan.rst
@@ -78,7 +78,7 @@ Parameters
 
   Current limitations:
 
-  - Except for the case described in :ref:`scan-single-wg`, the function will internally block until the issued kernels have completed execution.
+  - Except for the case described in the :ref:`single-work-group note <scan-single-wg>`, the function will internally block until the issued kernels have completed execution.
     Although intended in the future to be an asynchronous call, the algorithm is currently synchronous.
   - The SYCL device associated with the provided queue must support 64-bit atomic operations if the element type is 64-bits.
   - There must be a known identity value for the provided combination of the element type and the binary operation. That is, ``sycl::has_known_identity_v`` must evaluate to true. Such operators are listed in the `SYCL 2020 specification <https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.identities>`_.
--- a/documentation/library_guide/kernel_templates/single_pass_scan.rst
+++ b/documentation/library_guide/kernel_templates/single_pass_scan.rst
@@ -78,7 +78,7 @@ Parameters
 
   Current limitations:
 
-  - The function will internally block until the issued kernels have completed execution.
+  - Except for the case described in :ref:`scan-single-wg`, the function will internally block until the issued kernels have completed execution.
     Although intended in the future to be an asynchronous call, the algorithm is currently synchronous.
   - The SYCL device associated with the provided queue must support 64-bit atomic operations if the element type is 64-bits.
   - There must be a known identity value for the provided combination of the element type and the binary operation. That is, ``sycl::has_known_identity_v`` must evaluate to true. Such operators are listed in the `SYCL 2020 specification <https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#table.identities>`_.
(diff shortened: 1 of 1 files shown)


documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -78,7 +78,7 @@ Parameters
 
   Current limitations:
 
-  - The function will internally block until the issued kernels have completed execution.
+  - Except for the case described in the :ref:`single-work-group note <scan-single-wg>`, the function will internally block until the issued kernels have completed execution.
```
dcbenito, line 81: Have you tested to see if this cross link works? I ask because it didn't work for me. The issue might be the underscore on line 168 _scan-single-wg, where on line 81 it is referenced as scan-single-wg. 

The sentence LGTM otherwise.

dcbenito (changes requested):

documentation/library_guide/kernel_templates/single_pass_scan.rst
adamfidel: Yes, in the generated HTML, it does create a link. When clicked on, it scrolls down to the correct note.
image

According to the [reStructuredText documentation](https://www.sphinx-doc.org/en/master/usage/referencing.html#ref-role) on cross-referencing, they state the following:

> Reference labels must start with an underscore. When referencing a label, the underscore must be omitted.

So I believe the underscore is necessary in the label but not in the reference to the label.

dcbenito: Thank you for the additional information! I approve.

dcbenito (dismissed):

mmichel11 (dismissed):
LGTM

documentation/library_guide/kernel_templates/single_pass_scan.rst
```diff
@@ -78,7 +78,7 @@ Parameters
...
     Although intended in the future to be an asynchronous call, the algorithm is currently synchronous.
```
danhoeflinger, line 82: Do we need to update the language in this sentence as well?
It seems to be in conflict with the previous sentence.

danhoeflinger: We can say something like 
"The API should be treated as asynchronous, and returns a sycl::event, which should be waited upon to guarantee completion, but in the majority of cases, it is currently synchronous and will return an event which has already been waited upon."

danhoeflinger: I suppose to me what I'm interested in is how the user is meant to use the function.  I think because there is a chance that the function may be asynchronous currently (if it fits in single wg mode), and the intention is that it will eventually be asynchronous for all cases, we should start with that expectation for users, as I wouldn't look at this function as safe to treat as blocking with that in mind.  

It doesn't hurt us to basically recommend users call `wait()` on the returned event, even if that call may currently be unnecessary for some users who are sure their data will be larger than will fit in a single wg.

danhoeflinger: To state it another way, I think the way we should portray this is:
This is intended to be used as an asynchronous API, but currently due to some technical limitations, some modes currently will not return until the algorithm fully completes.

Rather than:
This is a synchronous call, except for this special case.

adamfidel: I believe I understand where you are coming from. Based on that, maybe we can not even mention the single-work-group case and replace both bullet points with a single sentence:

"The function is intended to asynchronous, but in some cases, the function will not return until the algorithm fully completes.

danhoeflinger: I think that makes sense to me.  Does it make sense that to explicitly state that waiting on the returned `sycl::event` can be used in either case to ensure the algorithm has completed?   (I don't know the answer to this)

adamfidel: I don't think we should recommend that users call `wait()` explicitly, as they may wish to use the returned `sycl::event` in other ways (such as passing it to a handler's `depends_on()` method).

Maybe it makes sense to change the wording of the "Return type" section, which currently states:

> A `sycl::event` object representing the status of the algorithm execution.

Perhaps we can add something to it like: "The event may or may not be completed by the time it is returned."

danhoeflinger: Yeah, I think the current language is actually fine "A `sycl::event object` representing the status of the algorithm execution."
and your last proposed "The function is intended to asynchronous, but in some cases, the function will not return until the algorithm fully completes." is also good.  I don't think we need any more than that.

danhoeflinger (approved):
LGTM

PR #1774, opened 2024-08-12
Move `sycl::event` and `__result_and_scratch_storage` into `__future`

SergeyKopienko:
In this PR we move the instances of `sycl::event` and `__result_and_scratch_storage` into `__future` constructor to avoid extra operations with counters inside `std::shared_ptr` objects which are the members of `__result_and_scratch_storage` (and probably in `sycl::event` implementation too).

Also we avoid `copy` operation for `_ExecutionPolicy` instance inside `__result_and_scratch_storage` too.

Please take a look to https://stackoverflow.com/questions/41871115/why-would-i-stdmove-an-stdshared-ptr for details.

Additionally, in this PR we fix issue https://github.com/oneapi-src/oneDPL/issues/1776 : we remove ability to create `__future` instance by copy and copy-assignment:
```C++
    using FutureType = __future<_Event, _Args...>;
    // ....
    __future(const FutureType&) = delete;
    __future(FutureType&&) = default;

    FutureType&
    operator=(const FutureType&) = delete;
    FutureType&
    operator=(FutureType&&) = default;
```

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -695,10 +695,13 @@ struct __deferrable_mode
 template <typename _Event, typename... _Args>
 class __future
 {
-    _Event __my_event;
+  public:
 
-    using _Data = std::tuple<_Args...>;
-    _Data __data;
+      using _DataTuple = std::tuple<_Args...>;
+
+  protected:
+    _Event __my_event;
+    _DataTuple __my_data;
 
     template <typename _T>
     constexpr auto
@@ -727,7 +730,9 @@ class __future
 
     using FutureType = __future<_Event, _Args...>;
 
-    __future(_Event __e, _Args... __args) : __my_event(__e), __data(std::forward<_Args>(__args)...) {}
+    __future(_Event __e) : __my_event(__e) {}
+    __future(_Event __e, const _DataTuple& __data) : __my_event(__e), __my_data(__data) {}
+    __future(_Event __e, _DataTuple&& __data) : __my_event(__e), __my_data(std::forward<_DataTuple>(__data)) {}
     __future(const FutureType&) = delete;
     __future(FutureType&&) = default;
 
@@ -770,7 +775,7 @@ class __future
     {
         if constexpr (sizeof...(_Args) > 0)
         {
-            auto& __val = std::get<0>(__data);
+            auto& __val = std::get<0>(__my_data);
             return __wait_and_get_value(__val);
         }
         else
@@ -784,7 +789,7 @@ class __future
     __make_future(_T __t) const
     {
         auto new_val = std::tuple<_T>(__t);
-        auto new_tuple = std::tuple_cat(new_val, __data);
+        auto new_tuple = std::tuple_cat(new_val, __my_data);
         return __future<_Event, _T, _Args...>(__my_event, new_tuple);
     }
 };
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -693,9 +693,12 @@ struct __deferrable_mode
 //A contract for future class: <sycl::event or other event, a value, sycl::buffers..., or __usm_host_or_buffer_storage>
 //Impl details: inheritance (private) instead of aggregation for enabling the empty base optimization.
 template <typename _Event, typename... _Args>
-class __future : private std::tuple<_Args...>
+class __future
 {
     _Event __my_event;
+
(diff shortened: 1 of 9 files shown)


danhoeflinger:
For this to stop the copy from happening, changes are required for `__future` as well I believe.  I think @julianmi may be working on this already, we have discussed it a little.  

Here is a godbolt that shows that even with this move, it still ends up with a copy when creating the tuple, unless we forward it in.  The downside to such a change is it requires explicit specification of the `__future` type, rather than allowing it to use deduction.  Julian has been looking at a way to at least allow the user to skip specifying the event type, but I don't think we can avoid requiring users to define the future value types.  In the end, this may not be a bad thing, as it requires users of `__future` to be more deliberate with their types.

https://godbolt.org/z/vEWz6svMT

julianmi:
> For this to stop the copy from happening, changes are required for `__future` as well I believe. I think @julianmi may be working on this already, we have discussed it a little.
> 
> Here is a godbolt that shows that even with this move, it still ends up with a copy when creating the tuple, unless we forward it in. The downside to such a change is it requires explicit specification of the `__future` type, rather than allowing it to use deduction. Julian has been looking at a way to at least allow the user to skip specifying the event type, but I don't think we can avoid requiring users to define the future value types. In the end, this may not be a bad thing, as it requires users of `__future` to be more deliberate with their types.
> 
> https://godbolt.org/z/vEWz6svMT

That's right. I've also included @reble in the discussion since he designed the original `__future` class. Another alternative would be to remove the tuple and specialize for our current use cases (event and/or one result container). However, this would be less generic. I would propose to close this PR and have an offline design discussion.

SergeyKopienko:
> > For this to stop the copy from happening, changes are required for `__future` as well I believe. I think @julianmi may be working on this already, we have discussed it a little.
> > Here is a godbolt that shows that even with this move, it still ends up with a copy when creating the tuple, unless we forward it in. The downside to such a change is it requires explicit specification of the `__future` type, rather than allowing it to use deduction. Julian has been looking at a way to at least allow the user to skip specifying the event type, but I don't think we can avoid requiring users to define the future value types. In the end, this may not be a bad thing, as it requires users of `__future` to be more deliberate with their types.
> > https://godbolt.org/z/vEWz6svMT
> 
> That's right. I've also included @reble in the discussion since he designed the original `__future` class. Another alternative would be to remove the tuple and specialize for our current use cases (event and/or one result container). However, this would be less generic. I would propose to close this PR and have an offline design discussion.

I like this variant!

SergeyKopienko:
@danhoeflinger, @reble I wrote new issue about our `__future` implementation: https://github.com/oneapi-src/oneDPL/issues/1776

SergeyKopienko:
@danhoeflinger , please take a look to this PR again.
Probably this `__future` implementations is ok for us?
Also I fixed issue https://github.com/oneapi-src/oneDPL/issues/1776 here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -669,9 +669,15 @@ struct __result_and_scratch_storage
...
-class __future : private std::tuple<_Args...>
+class __future
 {
+  public:
+
+      using _DataTuple = std::tuple<_Args...>;
+
+  protected:
     _Event __my_event;
+    _DataTuple __my_data;
```
danhoeflinger, line 680: I know that this is part of the experimental async API, but is this an ABI breakage?
Do we care if it is for experimental API?

What advantage do we get from moving it to a member variable from inheritance?

SergeyKopienko: In principal, the advantage is that inheritance should have place only when it's really required. I don't think that it's this place.
But I think you are correct about ABI compatibility, will fix. Thanks.

SergeyKopienko: But looks like we in any case should break ABI at this place if we want to disable `copy`-construction and `copy`-assignment for the `__future` class:
```C++
    __future(const FutureType&) = delete;
    // ...
    FutureType&
    operator=(const FutureType&) = delete;
```
Is it correct?

SergeyKopienko: @akukanov, @rarutyun what do you think about this?
Are we able to break `ABI` in the `__future` class?

SergeyKopienko: Restored original inheritance from `std::tuple`

SergeyKopienko:
We made decision together with @akukanov and @danhoeflinger to postpone this PR.
So I clear milestone state and change it's state to Draft.

SergeyKopienko:
@reble please take a look to this draft PR and take anything if it's useful for you.

PR #1775, opened 2024-08-12
Fix double reference in `__device_reduce_kernel` param

SergeyKopienko:
In this PR we fix double reference in `__device_reduce_kernel` param `__temp_acc`:
 - we have only one call of `__device_reduce_kernel` and we pass into `__temp_acc` param simple pointer in all cases:
```C++
                    auto __temp_ptr =
                        __result_and_scratch_storage<_ExecutionPolicy2, _Tp>::__get_usm_or_buffer_accessor_ptr(
                            __temp_acc);
                    // ...
                    __device_reduce_kernel<_Tp>(..., __temp_ptr, ...);
```
where `__get_usm_or_buffer_accessor_ptr` is :
```C++
    template <typename _Acc>
    static auto
    __get_usm_or_buffer_accessor_ptr(const _Acc& __acc, ::std::size_t __scratch_n = 0)
    {
        return __acc.__get_pointer();
#else
        return &__acc[__scratch_n];
#endif
    }
```


danhoeflinger (approved):
LGTM

julianmi (approved):
LGTM

PR #1782, opened 2024-08-14
Adjust block size for reduce-then-scan based on input type

adamfidel:
Targeting to be included in #1762. This PR adjusts the block size to be bigger or smaller based on the input type.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -738,7 +738,8 @@ __parallel_transform_reduce_then_scan(oneapi::dpl::__internal::__device_backend_
 
     constexpr std::uint8_t __sub_group_size = 32;
     // Empirically determined maximum. May be less for non-full blocks.
-    constexpr std::uint8_t __max_inputs_per_item = 128;
+    constexpr std::uint8_t __block_size_factor = sizeof(_ValueType) > sizeof(std::uint32_t) ? sizeof(_ValueType) / sizeof(std::uint32_t) : sizeof(std::uint32_t) / sizeof(_ValueType);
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -738,7 +738,8 @@ __parallel_transform_reduce_then_scan(oneapi::dpl::__internal::__device_backend_
 
     constexpr std::uint8_t __sub_group_size = 32;
     // Empirically determined maximum. May be less for non-full blocks.
-    constexpr std::uint8_t __max_inputs_per_item = 128;
+    constexpr std::uint8_t __block_size_factor = sizeof(_ValueType) > sizeof(std::uint32_t) ? sizeof(_ValueType) / sizeof(std::uint32_t) : sizeof(std::uint32_t) / sizeof(_ValueType);
+    constexpr std::uint16_t __max_inputs_per_item = sizeof(_ValueType) > sizeof(std::uint32_t) ? 128 / __block_size_factor : 128 * __block_size_factor;
```
danhoeflinger, line 742: Can we adjust this so our baseline is for `sizeof(double)` types: `64` elements, then scale it by however much smaller the `ValueType` is?

``` 
std::uint8_t __scale = std::min(1, sizeof(double) / sizeof(_ValueType));
std::uing16_t __max_inputs_per_item = 64 * __scale; 
```
We could also limit it to 4x scale if we don't think think 1 byte types will benefit from 512.

mmichel11:
Can we also adjust the `__max_inputs_per_item` template parameters to be `std::uint16_t`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
adamfidel: Thanks, this is a cleaner way to write this expression. Although I believe we want to use `std::max` here instead of `std::min`.

danhoeflinger (approved):
LGTM

PR #471, opened 2022-01-10
Fix warnings of unused entities (library only)

dmitriy-sobolev:
Let's have different patches for the library and for the tests for easier review and prioritization.

PR #1784, opened 2024-08-19
Cleaning up a few places in the code

akukanov:
A couple of changes lurking in my working copy, plus the changes for unused variables from #471 that still apply, and dropping colons before `std` in the code around.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -23,6 +23,7 @@
 #include "../../iterator_impl.h"
 #include "sycl_iterator.h"
 #include "sycl_defs.h"
+#include "execution_sycl_defs.h"
 
 namespace oneapi
 {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -21,7 +21,6 @@
 
 #include "../../utils_ranges.h"
 #include "../../iterator_impl.h"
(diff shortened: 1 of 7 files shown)


SergeyKopienko:
@akukanov I think technically this patch is OK.
But is it a good practice to mix some significant changes like remove unused param from `__parallel_transform_scan_base` together with some unsignificant changes like remove `::` in one PR?

akukanov:
> @akukanov I think technically this patch is OK. But is it a good practice to mix some significant changes like remove unused param from `__parallel_transform_scan_base` together with some unsignificant changes like remove `::` in one PR?

We have decided in the past that we will gradually remove `::` as we update the code base, Essentially, I removed it from the functions & classes touched by the more significant changes - but yes, it's somewhat wider than only the modified expressions. Since the whole patch is about cleaning up the code without any functionality add or loss, I tend to think it's OK.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -21,7 +21,6 @@
 
 #include "../../utils_ranges.h"
 #include "../../iterator_impl.h"
-#include "../../glue_numeric_defs.h"
```
dmitriy-sobolev, line 24: I suggest including `execution_sycl_defs.h` if we remove `glue_numeric_defs.h`. The later transitively includes  `execution_sycl_defs.h`, which defines `__hetero_tag` and  `device_policy` used in the current file.

akukanov: Done.

dmitriy-sobolev (approved):

SergeyKopienko (commented):
LGTM

PR #1636, opened 2024-06-24
[RNG] Adding operators ==, !=, << and >> for engines and distributions

egrabovskaya:
There are the following requirements for operators <<, >> in C++ standard

- **for distributions**:

image

- **for engines**:

image
image

**Note:** llvm and gcc comply with the same requirements for distributions as for generators

Operator support was added for Bernoulli distribution and linear_congruential engine

PR #1785, opened 2024-08-20
[RNG] Added experimental support of Philox counter-based PRNG

ElenaTyuleneva:
C++ draft proposal was used as a design for public API - [p2075r6.pdf](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2075r6.pdf), the latest correction by the time of PR creation is [issue4134](https://cplusplus.github.io/LWG/issue4134)

Things to make before the merge:
- [x] Add testing for parameter n = 2 <- added statistics test [philox_uniform_real_distribution_test.pass.cpp](https://github.com/oneapi-src/oneDPL/blob/e73422b7cad34c9f0a0b613c490ed04d2efa5db8/test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp) to check that we are generating a statistically valid sequence.
- [x] Add uglification
- [x] Mark the feature as experimental since C++ std proposal has not yet got the final revision
- [x] Add statistics test for scalar-input Philox for W!=32 and W!=64
- [x] Apply clang-format


include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,91 @@
...
+    result_type res_hi, res_lo;
+    /* multiplication fits standard types */
+    if(W <= 32) {
+        uint_fast64_t mult_result = (uint_fast64_t)a * (uint_fast64_t)b;
+        res_hi = mult_result >> W;
+        res_lo =  mult_result & detail::fffmask<result_type, W>;
+    }
+    /* pen-pencil multiplication by 32-bit chunks */
+    else if(W == 64) {
+        res_lo = a * b;
```
aelizaro, line 64: I assume this approach should also work for W < 64, you do it as for 64 and then apply & mask

or as an else here we can check if compiler supports __int128 (e.g. if CLANG or GCC) and have a multiply as 2 128-bit numbers

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,91 @@
...
+// Implement w-bit mulhilo with an 2w-wide integer - returns
+// the w hi and w low bits of the 2w-bit product of a and b.
+template <typename UIntType, unsigned W>
+static std::pair<UIntType, UIntType> mulhilo(UIntType a, UIntType b)
+{
+    using result_type = UIntType;
+
+    result_type res_hi, res_lo;
+    /* multiplication fits standard types */
+    if(W <= 32) {
```
aelizaro, line 57: this if can be constexpr

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,366 @@
...
+    explicit philox_engine(scalar_type value) { seed(value); }
+    void seed(scalar_type value = default_seed) { seed_internal({ value & in_mask }); }
+
+    /* Set the state to arbitrary position */
+    void set_counter(const ::std::array<scalar_type, word_count>& counter) {
+        auto start = counter.begin();
+        auto end = counter.end();
+        for (size_t i = 0; i < word_count; i++) {
+            // all counters are set in everse order
+            state_.X[i] = (start == end) ? 0 : (*--end) & in_mask;
```
aelizaro, line 187: I assume we don't have to check start == end here as counter is guaranteed to have word_count size.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,366 @@
...
+        }
+    }
+
+    /* Generating functions */
+    result_type operator()() {
+        result_type ret = this->operator()<internal::type_traits_t<result_type>::num_elems>();
```
aelizaro, line 193: optional: you can have private function name e.g. generate() instead of operator(), so this line would be a bit simpler

test/xpu_api/random/conformance_tests/common_for_conformance_tests.hpp
```diff
@@ -62,8 +64,36 @@ typename Engine::scalar_type test(sycl::queue& queue) {
             return 0;
         }
     }
-
+    std::cout << "\n\t\tres sycl: " << dpstd_samples[REF_SAMPLE_ID] << std::endl;
     return dpstd_samples[REF_SAMPLE_ID];
+
+#else
```
aelizaro, line 70: please, remove #if1 later

test/xpu_api/random/conformance_tests/philox_test.pass.cpp
```diff
@@ -0,0 +1,62 @@
+// -*- C++ -*-
+//===-- minstd_rand_rand0_test.cpp -----------------------------------------===//
```
aelizaro, line 2: wrong file name in the comment (maybe we should remove it if it is not against the common rules)

documentation/library_guide/api_for_sycl_kernels/random.rst
```diff
@@ -37,6 +38,8 @@ Type                                                                  Descriptio
 ``ranlux48_base``                                                     ``oneapi::dpl::subtract_with_carry_engine<std::uint64_t, 48, 5, 12>``
 ``ranlux24``                                                          ``oneapi::dpl::discard_block_engine<ranlux24_base, 223, 23>``
 ``ranlux48``                                                          ``oneapi::dpl::discard_block_engine<ranlux48_base, 389, 11>``
+``philox4x32``                                                        ``oneapi::dpl::philox_engine<uint_fast32_t, 32, 4, 10, 0xD2511F53, 0x9E3779B9, 0xCD9E8D57, 0xBB67AE85>``
```
aelizaro, line 41: to check before the merge (I am not sure yet), if we align the order of consts with P2075R6 or wait till the corrections are published (where the multipliers should be swapped)

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,366 @@
...
+    }
+    os << engine.state_.idx;
+    
+    return os;
+}
+
+template<typename UIntType, ::std::size_t w, ::std::size_t n, ::std::size_t r, UIntType... consts>
+const sycl::stream&
+operator<<(const sycl::stream& os, const philox_engine<UIntType, w, n, r, consts...>& engine) {
+    for(auto x_elm: engine.state_.X) {
```
aelizaro, line 320: should we set:

internal::save_stream_flags<CharT, Traits> __flags(os);
  os.setf(::std::ios_base::dec | ::std::ios_base::left);
  CharT sp = os.widen(' ');
  os.fill(sp);
  
  here as well? and ' ' -> sp in lines 321, 324, 327

ElenaTyuleneva: I'm not sure. I did `operator<<` for `sycl::stream` similar to Kate's changes here #1636. All API for `sycl::stream` don't have the format set-up at the beginning(probably, it's not applicable for `sycl::stream`?)

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,91 @@
+// -*- C++ -*-
+//===-- philox-engine.h ------------------------------------------===//
```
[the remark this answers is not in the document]
ElenaTyuleneva: Thanks, fixed.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
ElenaTyuleneva: > I assume this approach should also work for W < 64, you do it as for 64 and then apply & mask

Yep, I've corrected the condition

> or as an else here we can check if compiler supports __int128 (e.g. if CLANG or GCC) and have a multiply as 2 128-bit numbers
Decided with @paveldyakov to implement this later, as an improvement of the current version. The condition of `__int128` type enabling should be defined very carefully to prevent compilation errors on user's size.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
ElenaTyuleneva: Done.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,366 @@
...
+
+public:
+    static constexpr ::std::array<scalar_type, array_size> multipliers =
+        detail::get_even_array_from_tuple<scalar_type>(::std::make_tuple(consts...),
+                                                    ::std::make_index_sequence<array_size>{});
+    static constexpr ::std::array<scalar_type, array_size> round_consts =
+        detail::get_odd_array_from_tuple<scalar_type>(::std::make_tuple(consts...),
+                                                    ::std::make_index_sequence<array_size>{});
+    static constexpr scalar_type min() { return 0; }
+    static constexpr scalar_type max() { return ::std::numeric_limits<scalar_type>::max(); }
```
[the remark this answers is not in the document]
ElenaTyuleneva: Changed, thanks!

test/xpu_api/random/conformance_tests/philox_test.pass.cpp
ElenaTyuleneva: I've renamed the header.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,373 @@
...
+
+        state_.idx = word_count;
+    }
+    
+    /* Increment counter by 1 */
+    void increase_counter_internal() {
+        state_.X[0] = (state_.X[0] + 1) & in_mask;
+        for (::std::size_t i = 1; i < word_count; ++i) {
+            if (state_.X[i - 1]) {
+                [[likely]] return;
```
paveldyakov, line 104: Could you please describe what [[likely]] means?

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,373 @@
...
+        }
+        
+        // There are already generated numbers in the buffer
+        tmp = state_.Y[curr_idx];
+        state_.idx = ++curr_idx;
+
+        return tmp;
+    }
+
+public:
```
paveldyakov, line 171: Question: why private methods described first?

ElenaTyuleneva: Hmm, I didn't pay a lot of attention to the order of public/private members inside the class. If it's important I can reorder them.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: A standard attribute to indicate to the compiler that a statement has a high probability of being executed. Here is the full description - https://en.cppreference.com/w/cpp/language/attributes/likely
As I see, it's c++20 feature, so possibly it's better to be removed for oneDPL integration.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,373 @@
...
+    /* engine characteristics */
+    static constexpr ::std::size_t word_size = w;
+    static constexpr ::std::size_t word_count = n;
+    static constexpr ::std::size_t round_count = r;
+
+private:
+    static_assert(n == 2 || n == 4, "n must be 2 or 4");
+    static_assert(sizeof...(consts) == n, "the amount of consts must be equal to n");
+    static_assert(r > 0, "r must be more than 0");
+    static_assert(w > 0 && w <= ::std::numeric_limits<scalar_type>::digits, "w must 0 < w < ::std::numeric_limits<UIntType>::digits");
```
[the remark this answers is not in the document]
ElenaTyuleneva: Ok, changed a bit.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,373 @@
...
+
+#include "random_common.h"
+#include "counter_based_engines_stuff.h"
+
+namespace oneapi
+{
+namespace dpl
+{
+
+template<typename UIntType, ::std::size_t w, ::std::size_t n, ::std::size_t r,  internal::element_type_t<UIntType> ...consts>
```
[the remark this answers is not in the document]
ElenaTyuleneva: I wanted to do that on the last stage of the review, since it makes the code less readable. I've specified that in my ["before-merge" toDo](https://github.com/oneapi-src/oneDPL/pull/1785#issue-2475206702).

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,91 @@
...
+    /* Other types are proceeds school multiplication - not supported yet */
+    else {
+        ; 
+    }
+    
+    return { res_hi, res_lo };
+}
+} // namespace detail
+
+#endif // _ONEDPL_CTR_ENGINES_STUFF_H
```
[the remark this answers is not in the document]
ElenaTyuleneva: Added.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,91 @@
...
+
+        // 64-bit product + two 32-bit values
+        result_type middle = p10 + (p00 >> 32) + (p01 & detail::fffmask<result_type, 32>);
+
+        // 64-bit product + two 32-bit values
+        res_hi = p11 + (middle >> 32) + (p01 >> 32);
+    }
+    /* Other types are proceeds school multiplication - not supported yet */
+    else {
+        ;
```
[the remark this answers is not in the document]
ElenaTyuleneva: This branch is covered by `static_assert` at the beginning of the function, That code can be removed at all.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,91 @@
...
+#ifndef _ONEDPL_CTR_ENGINES_STUFF_H
+#define _ONEDPL_CTR_ENGINES_STUFF_H
+
+#include <concepts>
+#include <iterator>
+#include <utility>
+#include <tuple>
+#include <limits>
+
+namespace detail {
```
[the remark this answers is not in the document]
ElenaTyuleneva: Yep, it should. Fixed.

test/xpu_api/random/conformance_tests/common_for_conformance_tests.hpp
ElenaTyuleneva: I don't need that anymore, I think. Removed.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: A good suggestion, I've added `generate_internal`. Also, I renamed  `generate()` to `philox_kernel()` to reflect the function sense better.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Hm, sounds reasonable, removed.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,98 @@
...
+#ifndef _ONEDPL_CTR_ENGINES_STUFF_H
+#define _ONEDPL_CTR_ENGINES_STUFF_H
+
+#include <concepts>
+#include <iterator>
+#include <utility>
+#include <tuple>
+#include <limits>
+
+namespace oneapi
```
paveldyakov, line 32: Suggestion: to move it under oneapi::dpl::detail::random:: namespace. The functionality is not needed in the other parts

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,406 @@
...
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+//
+// Abstract:
+//
+// Public header file provides implementation for Philox Engine
+
+#ifndef _ONEDPL_PHILOX_ENGINE_H
+#define _ONEDPL_PHILOX_ENGINE_H
```
paveldyakov, line 21: If I am not mistaken - the support is experimental. Please check how experimental features are handled in oneDPL.

include/oneapi/dpl/internal/random_impl/philox_engine.h
paveldyakov: Just a question - please decide what is better here

include/oneapi/dpl/internal/random_impl/philox_engine.h
paveldyakov: Yes, it makes sense to remove it for now

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,373 @@
...
+    static constexpr ::std::size_t word_size = w;
+    static constexpr ::std::size_t word_count = n;
+    static constexpr ::std::size_t round_count = r;
+
+private:
+    static_assert(n == 2 || n == 4, "n must be 2 or 4");
+    static_assert(sizeof...(consts) == n, "the amount of consts must be equal to n");
+    static_assert(r > 0, "r must be more than 0");
+    static_assert(w > 0 && w <= ::std::numeric_limits<scalar_type>::digits, "w must 0 < w < ::std::numeric_limits<UIntType>::digits");
+    static_assert(::std::numeric_limits<scalar_type>::digits <= 64, "UIntType size must be less than 64 bits");
```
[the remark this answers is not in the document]
paveldyakov: Suggestion: size of the scalar UIntType (in case of sycl::vec<T, N> the size of T) must be less than 64 bits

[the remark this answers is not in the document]
ElenaTyuleneva: The suggestion was applied

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
ElenaTyuleneva: Discussed offline to make it similar to `include/oneapi/dpl/internal/random_impl/random_common.h`
image
done in https://github.com/oneapi-src/oneDPL/pull/1785/commits/50c586c904501a28a5bcd6d440bbb260ba34cea8

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: I took a look at oneMKL and llvm implementations - in oneMKL all public methods are specified first in all engine and distribution classes. In llvm the order of public and private members is a bit mixed, but it also tends to be "public at the top". So I also moved all public methods first.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Discussed offline with @aelizaro for sycl::stream not applicable

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: I've marked Philox class as experimental by moving it to `experimental` namespace. I'll need a review from oneDPL developers to double-check it was done correctly.

paveldyakov (approved):
Thank you for applying the comments, Elena! 
Looks good to me

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,105 @@
...
+//
+// Copyright (C) Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// This file incorporates work covered by the following copyright and permission
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
```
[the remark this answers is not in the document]
ElenaTyuleneva: @akukanov ,  so the right copyright to be used for the new files is:

```
// -*- C++ -*-
//===-- <filename> ------------------------------------------===//
//
// Copyright (C) Intel Corporation
//
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
```
, right?

documentation/library_guide/api_for_sycl_kernels/random.rst
```diff
@@ -37,6 +38,8 @@ Type                                                                  Descriptio
 ``ranlux48_base``                                                     ``oneapi::dpl::subtract_with_carry_engine<std::uint64_t, 48, 5, 12>``
 ``ranlux24``                                                          ``oneapi::dpl::discard_block_engine<ranlux24_base, 223, 23>``
 ``ranlux48``                                                          ``oneapi::dpl::discard_block_engine<ranlux48_base, 389, 11>``
+``philox4x32``                                                        ``oneapi::dpl::experimental::philox_engine<uint_fast32_t, 32, 4, 10, 0xD2511F53, 0x9E3779B9, 0xCD9E8D57, 0xBB67AE85>``
```
aelizaro, line 41: please, add std:: to uint_fast to align with other lines

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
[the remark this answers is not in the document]
akukanov: That's right.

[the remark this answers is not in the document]
ElenaTyuleneva: Thanks for the comment, copyrights were changed in all new files added by this PR.

include/oneapi/dpl/random
```diff
@@ -51,6 +52,12 @@ using ranlux24_base = subtract_with_carry_engine<uint_fast32_t, 24, 10, 24>;
 using ranlux48_base = subtract_with_carry_engine<uint_fast64_t, 48, 5, 12>;
 using ranlux24 = discard_block_engine<ranlux24_base, 223, 23>;
 using ranlux48 = discard_block_engine<ranlux48_base, 389, 11>;
+namespace experimental
+{
+using philox4x32 = philox_engine<uint_fast32_t, 32, 4, 10, 0xD2511F53, 0x9E3779B9, 0xCD9E8D57, 0xBB67AE85>;
```
aelizaro, line 57: LWG issue for the philox is submitted and it has a different order of constants: https://cplusplus.github.io/LWG/issue4134

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+          __UIntType... __consts>
+::std::basic_istream<_CharT, _Traits>&
+operator>>(::std::basic_istream<_CharT, _Traits>&, philox_engine<__UIntType, __w, __n, __r, __consts...>&);
+
+template <typename _UIntType, ::std::size_t _w, ::std::size_t _n, ::std::size_t _r,
+          internal::element_type_t<_UIntType>... _consts>
+class philox_engine
+{
+    /* The size of the consts arrays */
+    static constexpr ::std::size_t array_size = _n / 2;
```
aelizaro, line 50: array_size should also be uglified

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+    static_assert(sizeof...(_consts) == _n, "the amount of _consts must be equal to _n");
+    static_assert(_r > 0, "_r must be more than 0");
+    static_assert(_w > 0 && _w <= ::std::numeric_limits<scalar_type>::digits,
+                  "_w must satisfy 0 < _w < ::std::numeric_limits<_UIntType>::digits");
+    static_assert(::std::numeric_limits<scalar_type>::digits <= 64,
+                  "size of the scalar _UIntType (in case of sycl::vec<T, N> the size of T) must be less than 64 bits");
+    static_assert(::std::is_unsigned_v<scalar_type>, "_UIntType must be unsigned type or vector of unsigned types");
+
+    static constexpr ::std::array<scalar_type, array_size> multipliers =
+        internal::experimental::get_even_array_from_tuple<scalar_type>(::std::make_tuple(_consts...),
```
aelizaro, line 72: you can try to use std::array instead of make_tuple as consts... have the same type

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+    static_assert(_n == 2 || _n == 4, "_n must be 2 or 4");
+    static_assert(sizeof...(_consts) == _n, "the amount of _consts must be equal to _n");
+    static_assert(_r > 0, "_r must be more than 0");
+    static_assert(_w > 0 && _w <= ::std::numeric_limits<scalar_type>::digits,
+                  "_w must satisfy 0 < _w < ::std::numeric_limits<_UIntType>::digits");
+    static_assert(::std::numeric_limits<scalar_type>::digits <= 64,
+                  "size of the scalar _UIntType (in case of sycl::vec<T, N> the size of T) must be less than 64 bits");
+    static_assert(::std::is_unsigned_v<scalar_type>, "_UIntType must be unsigned type or vector of unsigned types");
+
+    static constexpr ::std::array<scalar_type, array_size> multipliers =
```
aelizaro, line 71:
```suggestion
    static constexpr std::array<scalar_type, array_size> multipliers =
```

In the new code oneDPL stops using :: before std namespace

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+{
+namespace dpl
+{
+namespace internal
+{
+namespace experimental
+{
+
+/* word_mask<_U, __W> - an unsigned integral type with the low __W bits set */
+template <typename _U, unsigned __W,
```
aelizaro, line 33: __W should be std::size_t instead of unsigned

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+    discard(unsigned long long __z)
+    {
+        discard_internal(__z);
+    }
+
+    /* Equality operators */
+    friend bool
+    operator==(const philox_engine& __x, const philox_engine& __y)
+    {
+        if (!::std::equal(__x.state_.X.begin(), __x.state_.X.end(), __y.state_.X.begin()) ||
```
aelizaro, line 136: this can be simplified to:
return 
std::equal(__x.state_.X.begin(), __x.state_.X.end(), __y.state_.X.begin()) &&
std::equal(__x.state_.K.begin(), __x.state_.K.end(), __y.state_.K.begin()) &&
std::equal(__x.state_.Y.begin(), __x.state_.Y.end(), __y.state_.Y.begin()) &&
__x.state_.idx != __y.state_.idx;

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+    }
+    static constexpr scalar_type default_seed = 20111115u;
+
+    /* Constructors and seeding functions */
+    philox_engine() : philox_engine(default_seed) {}
+    explicit philox_engine(scalar_type __seed) { seed(__seed); }
+    void
+    seed(scalar_type __seed = default_seed)
+    {
+        seed_internal({__seed & in_mask});
```
aelizaro, line 95: having std::initializer_list for seed seems to be overkill, as we only have scalar seed

documentation/library_guide/api_for_sycl_kernels/random.rst
aelizaro: need to apply: https://cplusplus.github.io/LWG/issue4134

rarutyun (commented):
The comments were made together with Alina. I suggest you guys to talk, so Alina could provide extra feedback.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+        {
+            state_.K[__i] = (__start == __end) ? 0 : (*__start++) & in_mask;
+        }
+        // results are set to zero
+        for (::std::size_t __i = 0; __i < word_count; __i++)
+        {
+            state_.Y[__i] = 0;
+        }
+
+        state_.idx = word_count;
```
aelizaro, line 200: we are not aligned with paper here (https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2075r6.pdf): according to the paper we need to set idx as word_count -1 and when we do generate, first increment counter, then check idx == word_count:
image
Ideally we need to align but it will cause logic shift in generate and discard

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+                __R0 = __hi0 ^ __K0 ^ __L0;
+                __L0 = __lo0;
+                __K0 = (__K0 + round_consts[0]) & in_mask;
+            }
+            state_.Y[0] = __R0;
+            state_.Y[1] = __L0;
+        }
+        else if constexpr (word_count == 4)
+        {
+            scalar_type __R0 = (state_.X[0]) & in_mask;
```
aelizaro, line 372: when the constants will be changed this part can be represented in a more generic way through a loop for n/2:
image

test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp
```diff
@@ -0,0 +1,597 @@
...
+#include <iostream>
+
+#if TEST_UNNAMED_LAMBDAS
+#include <vector>
+#include <random>
+#include <oneapi/dpl/random>
+#include "statistics_common.h"
+
+// Tested engines
+using philox2x32 = oneapi::dpl::experimental::philox_engine<std::uint_fast32_t, 32, 2, 10, 0xd256d193, 0x0>;
```
aelizaro, line 24: we can also add tests with w not equal to 32 and 64 here

test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp
```diff
@@ -0,0 +1,597 @@
...
+
+    // memory allocation
+    std::vector<oneapi::dpl::internal::element_type_t<RealType>> samples(nsamples);
+
+    constexpr int num_elems = oneapi::dpl::internal::type_traits_t<RealType>::num_elems == 0
+                                  ? 1
+                                  : oneapi::dpl::internal::type_traits_t<RealType>::num_elems;
+    // generation
+    {
+        sycl::buffer<oneapi::dpl::internal::element_type_t<RealType>, 1> buffer(samples.data(), nsamples);
```
aelizaro, line 53:
```suggestion
        sycl::buffer<oneapi::dpl::internal::element_type_t<RealType>> buffer(samples.data(), nsamples);
```

test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp
```diff
@@ -0,0 +1,597 @@
...
+                Engine engine;
+                engine.discard(offset);
+                oneapi::dpl::uniform_real_distribution<RealType> distr(left, right);
+
+                sycl::vec<oneapi::dpl::internal::element_type_t<RealType>, num_elems> res = distr(engine, part);
+                for (int i = 0; i < n_elems; ++i)
+                    acc[offset + i] = res[i];
+            });
+        });
+        queue.wait_and_throw();
```
aelizaro, line 117: inconsistent wait() and wait_and_throw() - if we want to use throw it is better to have try - catch block

test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp
```diff
@@ -0,0 +1,597 @@
...
+        err += tests_set_portion<sycl::vec<double, 16>, sycl::vec<std::uint_fast64_t, 2>, philox2x64>(queue, 160, 17);
+        err += tests_set_portion<sycl::vec<double, 16>, sycl::vec<std::uint_fast64_t, 1>, philox2x64>(queue, 160, 17);
+#    endif // TEST_LONG_RUN
+        EXPECT_TRUE(!err, "Test FAILED");
+    }
+
+#endif // TEST_UNNAMED_LAMBDAS
+
+    return TestUtils::done(TEST_UNNAMED_LAMBDAS);
+}
```
aelizaro, line 597: Please, add a separate test (can be host-side only) with thorough discard check:
generate - then discard and discard after generate + overflow check, you can use base from: https://github.com/rarutyun/iso_cpp_papers/pull/5/files

test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp
```diff
@@ -0,0 +1,597 @@
...
+    }
+    else
+    {
+        std::cout << "\tPassed" << std::endl;
+    }
+
+    return err;
+}
+
+template <class RealType, class UIntType, typename Engine>
```
aelizaro, line 86: please, align class and typename

documentation/library_guide/api_for_sycl_kernels/random.rst
ElenaTyuleneva: Applied, thanks for pointing this out.

include/oneapi/dpl/random
ElenaTyuleneva: Done

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Agreed, done.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Thanks for the proposal, done! Only a minor remark - the last statement should be `_x.state.idx == _y.state.idx;`

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Yep, I've made things simpler here.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+    {
+        result_type __ret = generate_internal<internal::type_traits_t<result_type>::num_elems>(__random_nums);
+        return __ret;
+    }
+
+    /* Shift the counter only forward relative to its current position */
+    void
+    discard(unsigned long long __z)
+    {
+        discard_internal(__z);
```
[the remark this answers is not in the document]
ElenaTyuleneva: I tried to minimize the size of the public section of the class to make it more readable. That's why all internal logic was moved to internal methods.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Agree, I've replaced tuple with array.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: All occurrences were refactored

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
ElenaTyuleneva: Thanks for noticing that, changed.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+          typename = ::std::enable_if_t<::std::is_unsigned_v<_U> && (__W <= ::std::numeric_limits<_U>::digits)>>
+constexpr _U word_mask = __W ? (_U(~(_U(0))) >> (::std::numeric_limits<_U>::digits - __W)) : 0;
+
+/* For unpacking variadic of constants into two arrays */
+template <typename _UIntType, typename _Tuple, ::std::size_t... __Is>
+constexpr auto
+get_even_array_from_tuple(_Tuple __t, ::std::index_sequence<__Is...>)
+{
+    return ::std::array<_UIntType, ::std::index_sequence<__Is...>::size()>{::std::get<__Is * 2>(__t)...};
+}
```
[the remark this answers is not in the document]
ElenaTyuleneva: Thanks for the suggestion, I've replaced tuple with array here

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+    {
+        return ::std::numeric_limits<scalar_type>::max() & in_mask;
+    }
+    static constexpr scalar_type default_seed = 20111115u;
+
+    /* Constructors and seeding functions */
+    philox_engine() : philox_engine(default_seed) {}
+    explicit philox_engine(scalar_type __seed) { seed(__seed); }
+    void
+    seed(scalar_type __seed = default_seed)
```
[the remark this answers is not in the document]
ElenaTyuleneva: Applied

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+template <typename _UIntType, typename _Tuple, ::std::size_t... __Is>
+constexpr auto
+get_odd_array_from_tuple(_Tuple __t, std::index_sequence<__Is...>)
+{
+    return ::std::array<_UIntType, ::std::index_sequence<__Is...>::size()>{::std::get<__Is * 2 + 1>(__t)...};
+}
+
+/* Implement __W-bit mulhilo - returns the __W hi and __W low 
+   bits of the 2__W-bit product of __a and __b */
+template <typename _UIntType, unsigned __W>
```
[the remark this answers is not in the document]
ElenaTyuleneva: > please, align amount of "_" in templates


Initially I followed the logic `__` for values and `_` for types, so the amount of underscores may vary inside one template. But no problem, I've aligned amount of "_" in all templates.(_NOTE:_ I had to add two underscores for I/O methods, otherwise there was a problem of template parameters shadowing)

> _W should be std::size_t

I moved `mulhilo` into Philox class, so this template param is not necessary anymore.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+{
+namespace internal
+{
+namespace experimental
+{
+
+/* word_mask<_U, __W> - an unsigned integral type with the low __W bits set */
+template <typename _U, unsigned __W,
+          typename = ::std::enable_if_t<::std::is_unsigned_v<_U> && (__W <= ::std::numeric_limits<_U>::digits)>>
+constexpr _U word_mask = __W ? (_U(~(_U(0))) >> (::std::numeric_limits<_U>::digits - __W)) : 0;
```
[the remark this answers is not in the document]
ElenaTyuleneva: Thanks, done!

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+get_odd_array_from_tuple(_Tuple __t, std::index_sequence<__Is...>)
+{
+    return ::std::array<_UIntType, ::std::index_sequence<__Is...>::size()>{::std::get<__Is * 2 + 1>(__t)...};
+}
+
+/* Implement __W-bit mulhilo - returns the __W hi and __W low 
+   bits of the 2__W-bit product of __a and __b */
+template <typename _UIntType, unsigned __W>
+static ::std::pair<_UIntType, _UIntType>
+mulhilo(_UIntType __a, _UIntType __b)
```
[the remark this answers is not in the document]
ElenaTyuleneva: yep, I've decided that is a good idea, moved.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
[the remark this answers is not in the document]
ElenaTyuleneva: Corrected

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+{
+    static_assert(__W <= 64, "__W must be 0 < __W <= 64");
+
+    using result_type = _UIntType;
```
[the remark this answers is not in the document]
ElenaTyuleneva: No needed anymore, the function now is member of Philox.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+    result_type __res_hi, __res_lo;
+
+    /* multiplication fits standard types */
+    if constexpr (__W <= 32)
+    {
+        uint_fast64_t __mult_result = (uint_fast64_t)__a * (uint_fast64_t)__b;
```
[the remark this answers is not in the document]
ElenaTyuleneva: Added and checked also all  other places(NOTE: I've left `uint_fast64_t` and `uint_fast32_t`  without `std::` only in `include/oneapi/dpl/random` for the consistency with other code in this file, should I also add there std::?)

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+        __res_hi = __mult_result >> __W;
+        __res_lo = __mult_result & internal::experimental::word_mask<result_type, __W>;
+    }
+    /* pen-pencil multiplication by 32-bit chunks */
+    else if constexpr (__W > 32)
+    {
+        __res_lo = __a * __b;
```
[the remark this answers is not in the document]
ElenaTyuleneva: Applied mask to both `__res_lo ` and `__res_hi`.

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+        result_type __p11 = __x1 * __y1;
+        result_type __p01 = __x0 * __y1;
+        result_type __p10 = __x1 * __y0;
+        result_type __p00 = __x0 * __y0;
+
+        // 64-bit product + two 32-bit values
+        result_type __middle = __p10 + (__p00 >> 32) + (__p01 & internal::experimental::word_mask<result_type, 32>);
+
+        // 64-bit product + two 32-bit values
+        __res_hi = __p11 + (__middle >> 32) + (__p01 >> 32);
```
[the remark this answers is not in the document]
ElenaTyuleneva: Applied mask to both `__res_lo ` and `__res_hi`.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+namespace dpl
+{
+namespace experimental
+{
+
+template <typename _UIntType, ::std::size_t _w, ::std::size_t _n, ::std::size_t _r,
+          internal::element_type_t<_UIntType>... _consts>
+class philox_engine;
+
+template <class _CharT, class _Traits, typename __UIntType, ::std::size_t __w, ::std::size_t __n, ::std::size_t __r,
```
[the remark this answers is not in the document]
ElenaTyuleneva: `class` was replaced with `typename` in templates.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+                                                                      ::std::make_index_sequence<array_size>{});
+    static constexpr scalar_type
+    min()
+    {
+        return 0;
+    }
+    static constexpr scalar_type
+    max()
+    {
+        return ::std::numeric_limits<scalar_type>::max() & in_mask;
```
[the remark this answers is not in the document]
ElenaTyuleneva: Good catch! Fixed.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+
+        return __loc_result;
+    }
+
+    /* generate_internal() specified for a scalar output */
+    template <unsigned int _N = 0>
+    ::std::enable_if_t<(_N == 0), result_type>
+    generate_internal()
+    {
+        result_type __loc_result;
```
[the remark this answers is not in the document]
ElenaTyuleneva: Actually, we can. I removed usage of the extra variable here. The method now is like the following:

```
    /* generate_internal() specified for a scalar output */
    template <unsigned int _N = 0>
    std::enable_if_t<(_N == 0), result_type>
    generate_internal()
    {
        scalar_type __curr_idx = state_.idx;

        // check if buffer is empty
        if (__curr_idx == word_count)
        { 
            philox_kernel();
            increase_counter_internal();
            __curr_idx = 0;
        }

        state_.idx = __curr_idx + 1;

        return state_.Y[__curr_idx];;
    }
```

test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp
ElenaTyuleneva: Applied.

test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp
ElenaTyuleneva: The test was fixed

test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp
ElenaTyuleneva: I've extended `philox_test.pass` with this corner-cases testing - https://github.com/oneapi-src/oneDPL/blob/ccc0f1fcb2db03af15f0f87a0d1be2dee45c69c1/test/xpu_api/random/conformance_tests/philox_test.pass.cpp 

@aelizaro , could you please take a look and let me know if that looks fine?

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+
+    /* generate_internal() specified for sycl_vec output */
+    template <unsigned int _N = 0>
+    ::std::enable_if_t<(_N > 0), result_type>
+    generate_internal()
+    {
+        result_type __loc_result;
+        scalar_type __curr_idx = state_.idx;
+
+        for (int __elm_count = 0; __elm_count < _N; __elm_count++)
```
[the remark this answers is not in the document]
ElenaTyuleneva: In one method the loop range is compile-time information(template param `_N`), but in another it's a variable `__random_nums`. If we combine them, we may miss potential compiler optimization for the 1st case.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+
+    /* Set the state to arbitrary position */
+    void
+    set_counter(const ::std::array<scalar_type, word_count>& __counter)
+    {
+        auto __end = __counter.end();
+        for (::std::size_t __i = 0; __i < word_count; __i++)
+        {
+            // all counters are set in reverse order
+            state_.X[__i] = (*--__end) & in_mask;
```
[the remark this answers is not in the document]
ElenaTyuleneva: Thanks for the proposal! The index of `X` array here is also not obvious, but at least we get rid of prefix decrement logic. Changed.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+          internal::element_type_t<_UIntType>... _consts>
+class philox_engine
+{
+    /* The size of the consts arrays */
+    static constexpr std::size_t __array_size = _n / 2;
+
+    /* Methods for unpacking variadic of constants into two arrays */
+    template <typename _Array, std::size_t... _Is>
+    static constexpr auto
+    get_even_elm_array(_Array __inp, std::index_sequence<_Is...>)
```
aelizaro, line 63: minor: I would suggest get_even_elm_array -> get_even_elem_array or get_even_element_array

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+    {
+        return 0;
+    }
+
+    static constexpr scalar_type
+    max()
+    {
+        return (std::numeric_limits<scalar_type>::max() & in_mask) - 1;
+    }
+    static constexpr scalar_type default_seed = 20111115u;
```
aelizaro, line 110:
```suggestion

    static constexpr scalar_type default_seed = 20111115u;
```
extra '\n"

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+    /* Constructors */
+    philox_engine() : philox_engine(default_seed) {}
+
+    explicit philox_engine(scalar_type __seed) { seed(__seed); }
+    
+    /* Seeding function */
+    void
+    seed(scalar_type __seed = default_seed)
+    {
+        seed_internal({__seed & in_mask});
```
aelizaro, line 121:
```suggestion
        seed_internal(__seed & in_mask);
```

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+            state_.X[word_count - __i - 1] = __counter[__i] & in_mask;
+        }
+    }
+
+    /* Generating functions */
+    result_type
+    operator()()
+    {
+        result_type __ret = generate_internal<internal::type_traits_t<result_type>::num_elems>();
+        return __ret;
```
aelizaro, line 140: can this be done in one line?
just `return generate_internal<internal::type_traits_t<result_type>::num_elems>();`?

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+    }
+    /* operator () overload for result portion generation */
+    result_type
+    operator()(unsigned int __random_nums)
+    {
+        result_type __ret = generate_internal<internal::type_traits_t<result_type>::num_elems>(__random_nums);
```
aelizaro, line 146: same as above, can it be done in one line?

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+    /* Equality operators */
+    friend bool
+    operator==(const philox_engine& __x, const philox_engine& __y)
+    {
+        return (std::equal(__x.state_.X.begin(), __x.state_.X.end(), __y.state_.X.begin()) &&
+                std::equal(__x.state_.K.begin(), __x.state_.K.end(), __y.state_.K.begin()) &&
+                std::equal(__x.state_.Y.begin(), __x.state_.Y.end(), __y.state_.Y.begin()) &&
+                __x.state_.idx == __y.state_.idx);
+    }
+    friend bool
```
aelizaro, line 166:
```suggestion

    friend bool
```

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+    struct state
+    {
+        std::array<scalar_type, word_count> X;     // counters
+        std::array<scalar_type, word_count / 2> K; // keys
+        std::array<scalar_type, word_count> Y;     // results
+        scalar_type idx;                           // index
+    } state_;
+
+    /* __word_mask<_W> - scalar_type with the low _W bits set */
+    template <std::size_t _W>
```
aelizaro, line 198: do we need template _W here or can we just use word_size from the class?

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+    static constexpr scalar_type __word_mask = _W ? (~scalar_type(0) >> (std::numeric_limits<scalar_type>::digits - _W)) : 0;
+
+    /* Processing mask */
+    static constexpr auto in_mask = __word_mask<word_size>;
```
aelizaro, line 202: this can be merged with the previous one if _w is not used

test/xpu_api/random/statistics_tests/philox_uniform_real_distribution_test.pass.cpp
ElenaTyuleneva: Thanks, test coverage was extended here!

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: The logic of idx iteration is changed as well as the generation methods and discard method.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+    {
+        // set to zero counters and results
+        for (std::size_t __i = 0; __i < word_count; __i++)
+        {
+            state_.X[__i] = 0;
+            state_.Y[__i] = 0;
+        }
+        // 0th key element is set as seed, others are 0
+        state_.K[0] = __seed & in_mask;
+        for (std::size_t __i = 1; __i < (word_count / 2); __i++)
```
aelizaro, line 215: minor style: inconsistent `i++` and `++i` across the file

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,544 @@
...
+        }
+
+        state_.idx = word_count;
+    }
+
+    /* Increment counter by 1 */
+    void
+    increase_counter_internal()
+    {
+        state_.X[0] = (state_.X[0] + 1) & in_mask;
```
aelizaro, line 227: minor: can be reordered just to not duplicate the same line

```
for (std::size_t __i = 0; __i < word_count; ++__i)
 {
          state_.X[__i] = (state_.X[__i] + 1) & in_mask;
          if (state_.X[__i])
          {
              return;
          }
}
 ```

File: include/oneapi/dpl/internal/random_impl/philox_engine.h [hunk 62]
```diff
@@ -0,0 +1,527 @@
...
+                __V1 = __lo0;
+                __K0 = (__K0 + round_consts[0]) & in_mask;
+            }
+            state_.Y[0] = __V0;
+            state_.Y[1] = __V1;
+        }
+        else if constexpr (word_count == 4)
+        {
+            // permute X to V
+            scalar_type __V2 = (state_.X[0]) & in_mask; // ToDo to check if we need mask
```
aelizaro, line 376: ToDo left here, I assume we don't need to do & in_mask as we do it in every init/increment_counter/set_counter

test/xpu_api/random/interface_tests/engines_methods.pass.cpp
```diff
@@ -38,6 +40,17 @@ constexpr auto RANLUX24_BASE_MIN = 0;
...
+constexpr auto PHILOX4x32_W   = 32;
+constexpr auto PHILOX4x64_W   = 64;
+constexpr auto PHILOX4x32_N   = 4;
+constexpr auto PHILOX4x64_N   = 4;
+constexpr auto PHILOX4x32_R   = 10;
+constexpr auto PHILOX4x64_R   = 10;
+constexpr auto PHILOX4x32_MIN = 0;
+constexpr auto PHILOX4x32_MAX = (std::numeric_limits<ex::philox4x32::result_type>::max() & 0xFFFFFFFF) - 1;
+constexpr auto PHILOX4x64_MIN = 0;
+constexpr auto PHILOX4x64_MAX = (std::numeric_limits<ex::philox4x64::result_type>::max() & 0xFFFFFFFFFFFFFFFF) - 1;
```
aelizaro, line 52: do we still need -1 here?

aelizaro: I think it is better to do numeric_limits<uint64_fast_t/uint32_fast_t>::max() - 1 here for predefined engines, not to rely on philox api which we are testing in the same file

test/xpu_api/random/unit_tests/philox_unit_test.pass.cpp
```diff
@@ -0,0 +1,323 @@
...
+}
+
+/*
+ * All functions below have the following return values:
+ *           0 in case of success
+ *           1 in case of failure
+ */
+
+template <typename Engine>
+int api_test() {
```
aelizaro, line 96: this one should be covered by engine_methods, may be removed

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: I'd like to leave `__word_mask ` configurable by a different `_W`, since it's used in `mulhilo` to form a mask different from `in_mask`

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: It's used, please see the comment above.

test/xpu_api/random/interface_tests/engines_methods.pass.cpp
ElenaTyuleneva: > do we still need -1 here?

I think so. Since method `max()` returns  `(std::numeric_limits<scalar_type>::max() & in_mask) - 1;` and we would like to check its work outside.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Renamed to `get_even_element_array` and `get_odd_element_array`

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Appled

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Thanks for the proposal, applied.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,527 @@
+// -*- C++ -*-
+//===-- philox-engine.h ---------------------------------------------------===//
```
aelizaro, line 2:
```suggestion
//===-- philox_engine.h ---------------------------------------------------===//
```

test/xpu_api/random/unit_tests/philox_unit_test.pass.cpp
ElenaTyuleneva: Removed

test/xpu_api/random/interface_tests/engines_methods.pass.cpp
ElenaTyuleneva: > I think it is better to do numeric_limits<uint64_fast_t/uint32_fast_t>::max() - 1 here for predefined engines, not to rely on philox api which we are testing in the same file

Yes, that would be better! changed

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: >  I assume we don't need to do & in_mask as we do it in every init/increment_counter/set_counter

Agree. Also unnecessary comment was removed.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Thanks for the suggestion, applied!

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Aligned to prefix incrementation.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Applied

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Yes, it can, changed.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Yep, done

include/oneapi/dpl/internal/random_impl/counter_based_engines_stuff.h
```diff
@@ -0,0 +1,99 @@
...
+    static_assert(__W <= 64, "__W must be 0 < __W <= 64");
+
+    using result_type = _UIntType;
+    result_type __res_hi, __res_lo;
+
+    /* multiplication fits standard types */
+    if constexpr (__W <= 32)
+    {
+        uint_fast64_t __mult_result = (uint_fast64_t)__a * (uint_fast64_t)__b;
+        __res_hi = __mult_result >> __W;
```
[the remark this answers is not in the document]
ElenaTyuleneva: My logic is - here we have the hi an lo parts of the multiplication stored in first and second `word_size` bits of the product  `__mult_result `. So to get the hi part we need to shift the amount of lo bits, which is equal to `__W`(in the updated code to `word_size`).

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,533 @@
...
+
+            //          select high chunk             shift for addition with the next counter chunk
+            __ctr_inc = (__ctr_inc & (~in_mask)) >> (std::numeric_limits<unsigned long long>::digits - word_size);
+        }
+    }
+
+    /* generate_internal() specified for sycl_vec output 
+       and overload for result portion generation */
+    template <unsigned int _N = 0>
+    std::enable_if_t<(_N > 0), result_type>
```
aelizaro, line 266: this enable_if can be removed as it is another overload

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,533 @@
...
+                state_.idx = 0;
+            }
+            __loc_result[__elm_count] = state_.Y[state_.idx];
+        }
+
+        return __loc_result;
+    }
+
+    /* generate_internal() specified for sycl_vec output */
+    template <unsigned int _N = 0>
```
aelizaro, line 291: let's remove default template parameter

test/xpu_api/random/unit_tests/philox_unit_test.pass.cpp
```diff
@@ -0,0 +1,318 @@
...
+
+    std::cout << "void skip_test() [Engine = philox4x32]";
+    err += skip_test<ex::philox4x32>();
+    std::cout << "void skip_test() [Engine = philox4x64]";
+    err += skip_test<ex::philox4x64>();
+
+    std::cout << "void counter_overflow_test() [Engine = philox4x32]";
+    err += counter_overflow_test<ex::philox4x32>();
+    std::cout << "void counter_overflow_test() [Engine = philox4x64]";
+    err += counter_overflow_test<ex::philox4x64>();
```
aelizaro, line 70: let's add w != 32/64 for overflow tests

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,533 @@
...
+{
+    /* The size of the consts arrays */
+    static constexpr std::size_t __array_size = _n / 2;
+
+    /* Methods for unpacking variadic of constants into two arrays */
+    template <typename _Array, std::size_t... _Is>
+    static constexpr auto
+    get_even_element_array(_Array __inp, std::index_sequence<_Is...>)
+    {
+        return std::array<scalar_type, std::index_sequence<_Is...>::size()>{std::get<_Is * 2>(__inp)...};
```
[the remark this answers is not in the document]
ElenaTyuleneva: Done.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,483 @@
...
+    /* Types */
+    using result_type = _UIntType;
+    using scalar_type = internal::element_type_t<result_type>;
+
+    /* Engine characteristics */
+    static constexpr ::std::size_t word_size = _w;
+    static constexpr ::std::size_t word_count = _n;
+    static constexpr ::std::size_t round_count = _r;
+
+    static_assert(_n == 2 || _n == 4, "_n must be 2 or 4");
```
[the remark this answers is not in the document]
ElenaTyuleneva: All param names were aligned with names in spec

test/xpu_api/random/unit_tests/philox_unit_test.pass.cpp
ElenaTyuleneva: Added

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,533 @@
...
+          internal::element_type_t<_UIntType>... _consts>
+class philox_engine
+{
+    /* The size of the consts arrays */
+    static constexpr std::size_t __array_size = _n / 2;
+
+    /* Methods for unpacking variadic of constants into two arrays */
+    template <typename _Array, std::size_t... _Is>
+    static constexpr auto
+    get_even_element_array(_Array __inp, std::index_sequence<_Is...>)
```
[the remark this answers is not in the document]
ElenaTyuleneva: Applied.

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -0,0 +1,533 @@
...
+    static constexpr scalar_type
+    min()
+    {
+        return 0;
+    }
+
+    static constexpr scalar_type
+    max()
+    {
+        return (std::numeric_limits<scalar_type>::max() & in_mask) - 1;
```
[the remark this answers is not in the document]
aelizaro: discussed offline, it seems like an issue in paper. it should be 2^w - 1 which is basically just:
return in_mask;

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: Decided offline to leave it consistent with other engines.

include/oneapi/dpl/internal/random_impl/philox_engine.h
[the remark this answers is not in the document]
ElenaTyuleneva: Fixed + added a comment to make the operation clearer.

ElenaTyuleneva:
@aelizaro , the PR is ready for the final review, I've pushed all planned commits here.

aelizaro (approved):
Thank you for the huge work @ElenaTyuleneva!

The changes looks good to me.
Reviewed offline with @rarutyun. All the major design comments are resolved; we agreed to merge the feature as experimental. The algorithmic part of the implementation wasn't checked too thoroughly, but it relies on good test coverage, which is done in this PR.

PR #1787, opened 2024-08-21
[test] Skip std::unique UserType test for dpcpp backend

danhoeflinger:
This PR skips the custom `UserType` test of `std::equal` for dpcpp backends.

This is done because `UserType` is not [device copyable](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable) (and shouldn't be explicitly specialized to be device copyable).  It has a custom copy assignment operator which is **not the same as a bitwise copy of the object**.  

Instead of changing this custom type to be device copyable, we instead skip this test to avoid conflicts with LLVM upstream.


MikeDvorskiy (approved):
LGTM

PR #1789, opened 2024-08-21
[test] Update broken macro for comparison

danhoeflinger:
Version 2025.0.0 of icpx will still has the issue when combined with MSVC STD implementation comparing std::optional, std::pair etc.
Updating the version broken macro to reflect this.


mmichel11 (approved):
LGTM

SergeyKopienko (approved):
LGTM

PR #1790, opened 2024-08-21
Resolve ESIMD sort KT compilation failures with new compiler

mmichel11:
Our ESIMD radix sort has two issues with newer compilers:

1. We unnecessarily use the `template` keyword prior to calling a function without a template argument list. This causes compilation failures with newer compilers. This has been fixed by removing `template` in this instance.
2. The `sycl::ext::intel::esimd::sw_barrier` enumerator has been removed and as far as I am aware there is no other ESIMD equivalent. We have a note that these `sw_barrier` fences are for performance reasons. There is one fence in the inter-wg synchronization that I believe is necessary for correctness (I see a hang otherwise). However, it should use a global memory fence and not a software barrier as we need ordering of reads which are from global memory. I have removed the other sw_barrier fences and do not observe any measurable performance impact.

I can share performance data for 2 if requested.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/internal/esimd_defs.h
+++ b/include/oneapi/dpl/experimental/kt/internal/esimd_defs.h
@@ -21,8 +21,6 @@
 // https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_esimd/sycl_ext_intel_esimd.md#static-allocation-of-slm-using-slm_init-function
 #define _ONEDPL_ESIMD_INLINE inline __attribute__((always_inline))
 
-#define _ONEDPL_SORT_KT_EXTRA_BARRIER_NEEDED (__INTEL_LLVM_COMPILER > 0 && __INTEL_LLVM_COMPILER <= 20250000)
-
 namespace oneapi::dpl::experimental::kt::gpu::esimd::__impl
 {
 
--- a/include/oneapi/dpl/experimental/kt/internal/esimd_defs.h
+++ b/include/oneapi/dpl/experimental/kt/internal/esimd_defs.h
@@ -21,8 +21,7 @@
 // https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_esimd/sycl_ext_intel_esimd.md#static-allocation-of-slm-using-slm_init-function
 #define _ONEDPL_ESIMD_INLINE inline __attribute__((always_inline))
 
-#define _ONEDPL_SORT_KT_EXTRA_BARRIER_NEEDED \
-    (__INTEL_LLVM_COMPILER > 0 && __INTEL_LLVM_COMPILER <= 20250000)
+#define _ONEDPL_SORT_KT_EXTRA_BARRIER_NEEDED (__INTEL_LLVM_COMPILER > 0 && __INTEL_LLVM_COMPILER <= 20250000)
 
 namespace oneapi::dpl::experimental::kt::gpu::esimd::__impl
 {
--- a/include/oneapi/dpl/experimental/kt/internal/esimd_defs.h
+++ b/include/oneapi/dpl/experimental/kt/internal/esimd_defs.h
@@ -21,6 +21,9 @@
 // https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_esimd/sycl_ext_intel_esimd.md#static-allocation-of-slm-using-slm_init-function
 #define _ONEDPL_ESIMD_INLINE inline __attribute__((always_inline))
 
+#define _ONEDPL_SORT_KT_EXTRA_BARRIER_NEEDED \
+    (__INTEL_LLVM_COMPILER > 0 && __INTEL_LLVM_COMPILER <= 20250000)
+
 namespace oneapi::dpl::experimental::kt::gpu::esimd::__impl
 {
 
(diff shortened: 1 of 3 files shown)


danhoeflinger (dismissed):
LGTM, 
I think its worthwhile to get @dmitriy-sobolev  approval before merging, unless there is urgent reason to go forward before next week.

adamfidel:
I had to use the [Wayback Machine](https://web.archive.org/web/20230929210820/https://intel.github.io/llvm-docs/doxygen/group__sycl__esimd__memory.html) to see what `sw_barrier` was meant to do since it was scrubbed from the documentation after being removed.

Its intended effect was 

| Name| Description |
| ---|---|
| `sw_barrier` | Creates a software (compiler) barrier, which does not generate any instruction and only prevents instruction scheduler from reordering instructions across this barrier at compile time. |

It doesn't appear that that the fences that are being removed are guarding against any critical reordering across the fence from what I can tell, so I think it's fine there. It also looks like we're not touching any fences that were recently added in #1676.

It seems okay at a glance but I would still perhaps wait on input from @dmitriy-sobolev as suggested by @danhoeflinger.

mmichel11:
I will wait on @dmitriy-sobolev's input next week before merging.

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
```diff
@@ -644,10 +639,7 @@ struct __radix_sort_onesweep_kernel
                         _GlobOffsetT, __bin_width, __dpl_esimd::__ens::lsc_data_size::default_size,
                         __dpl_esimd::__ens::cache_hint::uncached, __dpl_esimd::__ens::cache_hint::cached>(
                         __p_prev_group_hist + __local_tid * __bin_width);
-                    // Software barrier is used in order to motivate the compiler
-                    // to generate memory operations in an order which results in better performance
-                    // TODO: check if it is still necessary.
-                    __dpl_esimd::__ns::fence<__dpl_esimd::__ns::fence_mask::sw_barrier>();
+                    __dpl_esimd::__ns::fence<__dpl_esimd::__ns::memory_kind::global>();
```
dmitriy-sobolev, line 642: Let me think about it since I do not have clear understanding of why there is a hang after removal of `sw_barrier` and why the introduced barrier helps... One confusing fact is that the fence is applied on group-scope by default (i.e. it guarantees global memory ops visibility within a group), however it is a cross-group synchronization section.

I was able to reproduce the hang while removing the fence. However, it does not reproduce when I replace it with:
`__dpl_esimd::__ns::fence<__dpl_esimd::__ns::memory_kind::local>();`

Note that this fence above:
https://github.com/oneapi-src/oneDPL/pull/1790/files#diff-b0f1146f4a5f7425f04ba4973e688c2c834dc1c4f7563ecb95a3e937bab81c1bR601-R603
guarantees that the global histogram and histograms local for each group are observable by other groups. This should be enough to avoid hanging. In the worst case, a group "N" can scan all the previous histograms.

dmitriy-sobolev: I've gathered the kernels compiled in asm format for Intel® Data Center GPU Max:
 - [sw_fence.txt](https://github.com/user-attachments/files/16748126/sw_fence.txt)
 - [local_fence.txt](https://github.com/user-attachments/files/16748127/local_fence.txt)
 - [no_fence.txt](https://github.com/user-attachments/files/16748128/no_fence.txt)

The major difference between `sw_fence` and `no_fence` is the following:

```asm
_L_k43_10_preHeader:
BB_9:
// B018: Preds:{B017, B018},  Succs:{B018}
(W)     jmpi                                 _L_k43_10_preHeader                                     // $487
```
It looks like an plain endless loop.

I suggest changing the barrier to the local one, guarding it with a macro to turn it off after the next minor version of icpx, and adding a comment. I will create a bug ticket in the meantime.

mmichel11: Thanks for taking a deeper look and clarifying on the barriers. I agree that it should not be needed.

I have switched to the local barrier, added a compiler check macro, and added a comment.

include/oneapi/dpl/experimental/kt/internal/esimd_defs.h
```diff
@@ -21,6 +21,8 @@
 // https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_esimd/sycl_ext_intel_esimd.md#static-allocation-of-slm-using-slm_init-function
 #define _ONEDPL_ESIMD_INLINE inline __attribute__((always_inline))
 
+#define _ONEDPL_SORT_KT_EXTRA_BARRIER_NEEDED (__INTEL_LLVM_COMPILER > 0 && __INTEL_LLVM_COMPILER <= 20250000)
```
dmitriy-sobolev, line 24: Let's also apply the fix with [oneAPI DPC++ compiler](https://github.com/intel/llvm/tree/sycl?tab=readme-ov-file#oneapi-dpc-compiler): I was able to reproduce the hang with this compiler.

I would recommend relying on `_ONEDPL_LIBSYCL_VERSION` macro instead of `__INTEL_LLVM_COMPILER`. The drawback is that is not directly connected to the compiler version, which has the bug, but they are shipped side-by-side. I think it would be safe to assume `8` as a major version and `1` is a minor. The latest combination is 8.0, but 8.1 should provide more safety in case if a minor update happens soon.

dmitriy-sobolev: Probably, 9.0 would be even better.

dmitriy-sobolev: One more thing: there is only one place where the fix is applicable. You may consider moving the macro definition closer the the usage or even add the condition "in-place".

mmichel11: I've switched the macro to rely on `_ONEDPL_LIBSYCL_VERSION < 90000`, so it will expire with the next major update to 9. The comment above the macro explains the issue, so I have in-lined the check instead of making a separate macro.

dmitriy-sobolev (approved):
Thank you very much for the fix! LGTM.

danhoeflinger (approved):

PR #1791, opened 2024-08-22
Add code change workaround for 64-bit reduce_by_segment bug

mmichel11:
There is an IGC bug that affects `reduce_by_segment` with 64-bit types on GPU Series Max devices which has previously required us to provide the `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` macro workaround. This workaround invokes the legacy implementation which is around ~3x slower but produces correct results.

The IGC bug still exists, but I have a found a workaround with negligible performance impact within our `reduce_by_segment` implementation. This enables users to invoke the faster `reduce_by_segment` implementation without correctness issues.

By first initializing the private memory arrays to the known identity element prior to loading real data into some of the array indices, the register filling bug is avoided. I have verified with oneDPL tests (which previously caught this issue) and with external tests.

I have also removed the macro workaround and additional test.

### DIFF
--- a/include/oneapi/dpl/internal/reduce_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/reduce_by_segment_impl.h
@@ -352,6 +352,9 @@ __sycl_reduce_by_segment(__internal::__hetero_tag<_BackendTag>, _ExecutionPolicy
 #endif
             sycl::nd_range<1>{__n_groups * __wgroup_size, __wgroup_size}, [=](sycl::nd_item<1> __item) {
                 auto __identity = unseq_backend::__known_identity<_BinaryOperator, __val_type>;
+                // TODO: Remove this initialization to the identity when possible. We load real data to __loc_partials
+                // in the first loop below but this initialization to the identity works around an IGC register
+                // filling bug.
                 std::array<__val_type, __vals_per_item> __loc_partials = {__identity};
 
                 auto __group = __item.get_group();
--- a/include/oneapi/dpl/internal/reduce_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/reduce_by_segment_impl.h
@@ -351,7 +351,8 @@ __sycl_reduce_by_segment(__internal::__hetero_tag<_BackendTag>, _ExecutionPolicy
             __seg_reduce_wg_kernel,
 #endif
             sycl::nd_range<1>{__n_groups * __wgroup_size, __wgroup_size}, [=](sycl::nd_item<1> __item) {
-                ::std::array<__val_type, __vals_per_item> __loc_partials;
+                auto __identity = unseq_backend::__known_identity<_BinaryOperator, __val_type>;
+                std::array<__val_type, __vals_per_item> __loc_partials = {__identity};
 
                 auto __group = __item.get_group();
                 ::std::size_t __group_id = __item.get_group(0);
@@ -368,7 +369,6 @@ __sycl_reduce_by_segment(__internal::__hetero_tag<_BackendTag>, _ExecutionPolicy
 
                 ::std::size_t __max_end = 0;
                 ::std::size_t __item_segments = 0;
-                auto __identity = unseq_backend::__known_identity<_BinaryOperator, __val_type>;
 
                 __val_type __accumulator = __identity;
                 for (::std::size_t __i = __start; __i < __end; ++__i)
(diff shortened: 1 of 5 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -34,21 +34,11 @@ namespace unseq_backend
 //This optimization depends on Intel(R) oneAPI DPC++ Compiler implementation such as support of binary operators from std namespace.
 //We need to use defined(SYCL_IMPLEMENTATION_INTEL) macro as a guard.
 
-template <typename _Tp>
-inline constexpr bool __can_use_known_identity =
-#    if ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION
```
mmichel11, line 39: Is it okay if we directly remove this macro check, or should it be deprecated first with a `#warning` or something similar? From the user's perspective, they should just see `reduce_by_segment` speedup with 64-bit types.

include/oneapi/dpl/internal/reduce_by_segment_impl.h
```diff
@@ -351,7 +351,11 @@ __sycl_reduce_by_segment(__internal::__hetero_tag<_BackendTag>, _ExecutionPolicy
             __seg_reduce_wg_kernel,
 #endif
             sycl::nd_range<1>{__n_groups * __wgroup_size, __wgroup_size}, [=](sycl::nd_item<1> __item) {
-                ::std::array<__val_type, __vals_per_item> __loc_partials;
+                auto __identity = unseq_backend::__known_identity<_BinaryOperator, __val_type>;
+                // TODO: Remove this initialization to the identity when possible. We load real data to __loc_partials
+                // in the first loop below but this initialization to the identity works around an IGC register
+                // filling bug.
+                std::array<__val_type, __vals_per_item> __loc_partials = {__identity};
```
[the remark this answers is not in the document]
mmichel11: Thanks, it looks like the rest of the elements may be initialized to 0: https://en.cppreference.com/w/c/language/array_initialization.

The way the fix was implemented still worked since it does not matter what is loaded into the array as long as it's something. However, I switched to your suggestion to be consistent.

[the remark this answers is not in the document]
mmichel11: Filling the array after its definition seems to reintroduce the bug. I will see if I can find a better solution. I suppose what we originally had adds a default constructability requirement we do not want.

include/oneapi/dpl/internal/reduce_by_segment_impl.h
```diff
@@ -351,7 +351,12 @@ __sycl_reduce_by_segment(__internal::__hetero_tag<_BackendTag>, _ExecutionPolicy
             __seg_reduce_wg_kernel,
 #endif
             sycl::nd_range<1>{__n_groups * __wgroup_size, __wgroup_size}, [=](sycl::nd_item<1> __item) {
-                ::std::array<__val_type, __vals_per_item> __loc_partials;
+                auto __identity = unseq_backend::__known_identity<_BinaryOperator, __val_type>;
```
SergeyKopienko, line 354: Let's use `__val_type` instead of `auto` - it's will more readable, I think.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
timmiesmith: It is ok to directly remove it.

include/oneapi/dpl/internal/reduce_by_segment_impl.h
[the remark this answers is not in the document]
mmichel11: I have reverted the change I made here. My last statement is wrong, the default constructor for each array element is already called when we declare the array, so we are not adding any additional requirements.

I have explored some different ways to try to workaround the issue, but this seems to be the only thing that works. I've confirmed that register filling bug is avoided as our tests pass along with internal reproducers where the issue was reported.

include/oneapi/dpl/internal/reduce_by_segment_impl.h
mmichel11: Done

mmichel11:
Thank you for the reviews so far in this PR. I am closing it in favor of https://github.com/oneapi-src/oneDPL/pull/1915. The entire `reduce_by_segment` SYCL-based implementation is replaced with a reduce-then-scan call and is unaffected by this device compilation issue while also bringing significant performance improvement.

If for any reason we still need this fix, I can re-open in the future.

PR #1793, opened 2024-08-23
Add spec compliant histogram API while supporting legacy extension API

danhoeflinger:
This PR adjust the API overloads for histogram to create an API which complies to the current PR for [the specification](https://github.com/uxlfoundation/oneAPI-spec/pull/546), while continuing to support the existing API as an extension.

It removes an unnecessary and unused default parameter for `ValueType` which was incorrectly added in the original PR.  The new spec compliant API implements what was the intended signature originally motivated from [this discussion](https://github.com/oneapi-src/oneDPL/pull/1273#discussion_r1395697209).  

This change in overload / signature should intercept the normal usages of histogram which have matching types for the bound limits with the value type of the input iterators, and use the spec compliant signature.  

For those using histogram that provide bound limits which do not match the type, the legacy overload will still support this usage.  

We should not stop supporting any existing working cases with this change, but we will add some additional supported invocations where the type of the boundaries do not match each other.  If they are both implicitly convertible to the value type of the input iterator, they will be converted, and it should work (possibly with warnings).

Example in godbolt to play with: [Godbolt example](https://godbolt.org/z/4ePTobz4c)

### DIFF
--- a/include/oneapi/dpl/pstl/histogram_impl.h
+++ b/include/oneapi/dpl/pstl/histogram_impl.h
@@ -46,9 +46,32 @@ __pattern_histogram(_Tag, _ExecutionPolicy&& exec, _RandomAccessIterator1 __firs
 
 } // namespace __internal
 
+template <typename _ExecutionPolicy, typename _RandomAccessIterator1, typename _Size, typename _RandomAccessIterator2>
+oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _RandomAccessIterator2>
+histogram(_ExecutionPolicy&& exec, _RandomAccessIterator1 first, _RandomAccessIterator1 last, _Size num_bins,
+          typename std::iterator_traits<_RandomAccessIterator1>::value_type first_bin_min_val,
+          typename std::iterator_traits<_RandomAccessIterator1>::value_type last_bin_max_val,
+          _RandomAccessIterator2 histogram_first)
+{
+    using _BoundaryType = typename std::iterator_traits<_RandomAccessIterator1>::value_type;
+    const auto __dispatch_tag = oneapi::dpl::__internal::__select_backend(exec, first, histogram_first);
+
+    oneapi::dpl::__internal::__pattern_histogram(
+        __dispatch_tag, ::std::forward<_ExecutionPolicy>(exec), first, last, num_bins,
+        oneapi::dpl::__internal::__evenly_divided_binhash<_BoundaryType>(first_bin_min_val, last_bin_max_val, num_bins),
+        histogram_first);
+    return histogram_first + num_bins;
+}
+
+// This overload is provided to support an extension to the oneDPL specification to support the original implementation
+// of the histogram API, where the boundary type _ValueType could differ from the value type of the input iterator,
+// and required `operator<` and `operator<=` to be defined between _ValueType and
+// std::iterator_traits<_RandomAccessIterator1>::value_type rather than enforcing they were the same type
 template <typename _ExecutionPolicy, typename _RandomAccessIterator1, typename _Size, typename _RandomAccessIterator2,
           typename _ValueType>
-oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _RandomAccessIterator2>
+std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<::std::decay_t<_ExecutionPolicy>> &&
+                     !std::is_same_v<_ValueType, typename std::iterator_traits<_RandomAccessIterator1>::value_type>,
+                 _RandomAccessIterator2>
 histogram(_ExecutionPolicy&& exec, _RandomAccessIterator1 first, _RandomAccessIterator1 last, _Size num_bins,
           _ValueType first_bin_min_val, _ValueType last_bin_max_val, _RandomAccessIterator2 histogram_first)
 {
--- a/include/oneapi/dpl/pstl/histogram_extension_defs.h
+++ b/include/oneapi/dpl/pstl/histogram_extension_defs.h
@@ -22,10 +22,18 @@ namespace oneapi
 {
 namespace dpl
 {
+template <typename _ExecutionPolicy, typename _RandomAccessIterator1, typename _Size, typename _RandomAccessIterator2>
+oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _RandomAccessIterator2>
+histogram(_ExecutionPolicy&& exec, _RandomAccessIterator1 first, _RandomAccessIterator1 last, _Size num_bins,
+          typename std::iterator_traits<_RandomAccessIterator1>::value_type first_bin_min_val,
+          typename std::iterator_traits<_RandomAccessIterator1>::value_type last_bin_max_val,
+          _RandomAccessIterator2 histogram_first);
 
 template <typename _ExecutionPolicy, typename _RandomAccessIterator1, typename _Size, typename _RandomAccessIterator2,
-          typename _ValueType = typename ::std::iterator_traits<_RandomAccessIterator1>::value_type>
-oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _RandomAccessIterator2>
+          typename _ValueType>
+std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<::std::decay_t<_ExecutionPolicy>> &&
+                     !std::is_same_v<_ValueType, typename std::iterator_traits<_RandomAccessIterator1>::value_type>,
+                 _RandomAccessIterator2>
 histogram(_ExecutionPolicy&& exec, _RandomAccessIterator1 first, _RandomAccessIterator1 last, _Size num_bins,
           _ValueType first_bin_min_val, _ValueType last_bin_max_val, _RandomAccessIterator2 histogram_first);
 


timmiesmith (dismissed):
LGTM.

include/oneapi/dpl/pstl/histogram_impl.h
```diff
@@ -47,8 +47,32 @@ __pattern_histogram(_Tag, _ExecutionPolicy&& exec, _RandomAccessIterator1 __firs
 } // namespace __internal
 
 template <typename _ExecutionPolicy, typename _RandomAccessIterator1, typename _Size, typename _RandomAccessIterator2,
-          typename _ValueType>
+          typename> //final unused template param support extension API in combination with overload below
```
akukanov, line 50: I do not see why this extra template parameter is needed. As far as I know, it is fine to have function template overloads with different numbers of template parameters.

include/oneapi/dpl/pstl/histogram_impl.h
```diff
@@ -47,8 +47,32 @@ __pattern_histogram(_Tag, _ExecutionPolicy&& exec, _RandomAccessIterator1 __firs
...
+// of the histogram API, where the boundary type _ValueType could differ from the value type of the input iterator,
+// and required `<`, `<=`, `+`, `-`, and `/` to be defined between _ValueType and
+// std::iterator_traits<_RandomAccessIterator1>::value_type rather than enforcing they were the same type or convertible
+template <typename _ExecutionPolicy, typename _RandomAccessIterator1, typename _Size, typename _RandomAccessIterator2,
+          typename _ValueType>
+std::enable_if_t<oneapi::dpl::execution::is_execution_policy_v<std::decay_t<_ExecutionPolicy>> &&
+                     !std::is_same_v<_ValueType, typename std::iterator_traits<_RandomAccessIterator1>::value_type>,
+                 _RandomAccessIterator2>
 histogram(_ExecutionPolicy&& exec, _RandomAccessIterator1 first, _RandomAccessIterator1 last, _Size num_bins,
           _ValueType first_bin_min_val, _ValueType last_bin_max_val, _RandomAccessIterator2 histogram_first)
```
akukanov, line 77: I wonder if keeping the old variant of the function available sort of defeats the purpose of adding the new one, as Dan wrote in a comment to the specification PR:

> I think it is better for force users to provide bounds which are convertible to the value_type of the input sequence, and do the conversion once when accepting the arguments rather than each time in the comparison operator. 

If this overload is kept under the condition of type not being the same as the iterator value type, it will be selected even for the cases where we would want to have the conversion done only once. If the condition is changed to that of type not being convertible, will the code even work for such types? Or should the condition be for types that are not **implicitly** convertible to the desired one, while still explicitly convertible?

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: In combination with the other overload, this unused defaulted template parameters covers the case where a user specified all template parameters explicitly, and the specified type matches the input iterator value type.

`histogram<Policy, int*, std::size_t, uint32_t*, int>(policy, first, last, num_bins, bin_min, bin_max, histogram_first);`

The other overload would exclude it because of the matching type.

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: I need to experiment here a little to see if there is a better check.  I think if we check `!std::is_convertible` instead, I think it should still work.

danhoeflinger:
After discussion with @rarutyun and @MikeDvorskiy, I think we may have better options here (specification / implementation).  I will provide links for the alternatives tonight or tomorrow.  I sent and email with some details which you can look at in the mean time.

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: Technically, I believe this works with `!std::is_convertible_v` in place of `!istd::is_same_v` in that all the cases which build and run before still build and run in the new version as well.  

However, the unused and defaulted template parameter feels a little *more* hacky with this version. With the `!std::is_same_v` version, we know that the only time the spec compliant signature is picked, the types are the same, so doing this conversion is a no-op.  Alternately, with `!std::is_convertible_v`, if someone explicitly lists out their parameter list, including a `_ValueType` which is convertible but not the same as the input iterator value type to match the old signature, we would still pick the new spec compliant overload and do the conversion.  

I have an improved version in a sandbox godbolt which adds a (sigh) 4th overload, but manages to remove the hacky unused and defaulted template parameter from the spec compliant signature.

This version breaks up the extension overloads into two signatures:  
1) type not convertible to input iterator value type
  This is basically what we have here, but checking `!std::is_convertible_v`

2) Explicit template list 
  This makes the 4th template argument non-deducable, but has 5th and 6th template arguments to represent the types of the two boundary limit parameters, and

I'll get an commit in to update this PR with what it would look like tomorrow, its better, but I think I still prefer 1a or 1b.

include/oneapi/dpl/pstl/histogram_impl.h
akukanov: Since C++20, the standard has a clause in [algorithms.requirements] similar to this (taken from the current draft):

> The well-formedness and behavior of a call to an algorithm with an explicitly-specified template argument list is unspecified, except where explicitly stated otherwise.
[Note 3: Consequently, an implementation can declare an algorithm with different template parameters than those presented. — end note]

I believe this was added due to range algorithms, which are usually implemented in a way that they cannot be called with explicit template arguments. The restriction however, as I read it, is not limited to range algorithms.

Since we are adding parallel range algorithms both to the library and the spec as well, we will need a similar clause, and maybe it makes sense to apply to all algorithms.

danhoeflinger: Very interesting, I wasn't aware of that.

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: I've updated the code to implement the improved logic for supporting the extension API cases, and added tests to check these overloads specifically.

include/oneapi/dpl/pstl/histogram_impl.h
```diff
@@ -77,6 +79,54 @@ histogram(_ExecutionPolicy&& exec, _RandomAccessIterator1 first, _RandomAccessIt
...
+{
+    const auto __dispatch_tag = oneapi::dpl::__internal::__select_backend(exec, first, histogram_first);
+
+    oneapi::dpl::__internal::__pattern_histogram(
+        __dispatch_tag, ::std::forward<_ExecutionPolicy>(exec), first, last, num_bins,
+        oneapi::dpl::__internal::__evenly_divided_binhash<_ValueType>(_ValueType{first_bin_min_val},
+                                                                      _ValueType{last_bin_max_val}, num_bins),
+        histogram_first);
+    return histogram_first + num_bins;
+}
```
akukanov, line 128: In this implementation, why not instead call the main specified overload, just converting the parameters to `_ValueType`?

include/oneapi/dpl/pstl/histogram_impl.h
```diff
@@ -77,6 +79,54 @@ histogram(_ExecutionPolicy&& exec, _RandomAccessIterator1 first, _RandomAccessIt
...
+        __dispatch_tag, ::std::forward<_ExecutionPolicy>(exec), first, last, num_bins,
+        oneapi::dpl::__internal::__evenly_divided_binhash<_ValueType>(first_bin_min_val, last_bin_max_val, num_bins),
+        histogram_first);
+    return histogram_first + num_bins;
+}
+
+// This overload is provided to support an extension to the oneDPL specification to support the original implementation
+// of the histogram API, where if users explicitly-specify all template arguments, their arguments for bin boundary min
+// and max are convertible to the specified _ValueType, and the _ValueType is convertible to the value type of the
+// input iterator.
```
akukanov, line 109: It makes sense to mention that `_ValueType` is not deducible and so this overload can only be chosen if it's explicitly specified.

akukanov (commented):
Should we claim the old APIs deprecated, so that their use causes warnings and encourages the change to the supported overload?

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: We could I think.  I was trying to follow the convention to not call external oneDPL APIs from within implementation, but this may be a case where an exception would be useful.

akukanov:
> Does that have any implication on the oneDPL version numbering, etc.?

At most, I think, we'd need to get an approval to deprecate. Given that the API remains and only some corner case uses are potentially impacted, I am not sure what the process should be. @timmiesmith, what do you think on this?

akukanov:
> Does that have any implication on the oneDPL version numbering, etc.?

The version is not impacted, as the API is still there. But it makes sense to check what needs to be done to claim something deprecated. Tagging @timmiesmith.

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: After some testing, from my perspective this is not a good idea.

The two options are subtly different from each other.  As written currently in the PR, the functionality for this overload with explicit template arguments specified is exactly the same as it was before.  The user specifies that the bins be calculated from their chosen type, and we do so, using the operations (`+` `-` `\` `<` `<=`).

If we change to instead wrap and call the spec compliant signature, we do a conversion to the value type of the input iterator at the start, then internally operate in that type.  
With that change, we are doing `IteratorValueType{ExplicitValueType{argument}}` where the current implementation just does `ExplicitValueType{argument}`. If we are supporting these as an extension only, I think it makes sense to support them exactly as they were.  Both options work when matched with the correct corresponding call to the serial implementation in the tests, but I prefer it as it is written in the PR currently.

timmiesmith:
> > Does that have any implication on the oneDPL version numbering, etc.?
> 
> The version is not impacted, as the API is still there. But it makes sense to check what needs to be done to claim something deprecated. Tagging @timmiesmith.

Deprecations have to be announced 1 year before the API is removed, and it's encouraged to do it at the major oneAPI release. We need to add the warning macros to the APIs so the user does see the warnings at comepile time now, and then next year we can remove the APIs. https://github.com/oneapi-src/oneDPL/commit/5c4840f0f12b8c98664f6b9e76554789184a5871 is an example of the notice required.

danhoeflinger:
> Deprecations have to be announced 1 year before the API is removed, and it's encouraged to do it at the major oneAPI release. We need to add the warning macros to the APIs so the user does see the warnings at comepile time now, and then next year we can remove the APIs. [5c4840f](https://github.com/oneapi-src/oneDPL/commit/5c4840f0f12b8c98664f6b9e76554789184a5871) is an example of the notice required.

OK, I can add deprecation warnings here to the extension APIs.  Thanks @timmiesmith.

danhoeflinger:
Deprecations have been added.  Its added a clang-format suggestion which I very much disagree with.

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: Fixed.

include/oneapi/dpl/pstl/histogram_impl.h
akukanov: OK, I don't mind enough to argue :)

danhoeflinger:
We determined that we want to go with #1800  instead.  Converting to draft and removing from milestone.

danhoeflinger:
closing this in favor of #1800.

PR #1795, opened 2024-08-26
[RNG] Conversion fixes in uniform real distribution

ElenaTyuleneva:
When trying to build the test `test/xpu_api/random/device_tests/engine_device_test.pass.cpp` with enabled macro `TEST_LONG_RUN` , the following error appears:

`...
/oneDPL/include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h:277:29: error: use of overloaded operator '-' is ambiguous (with operand types 'std::enable_if_t<!std::is_same_v<vec_data_t<DataT>, vec_data_t<float>> && !std::is_same_v<detail::ConvertToOpenCLType_t<vec_data_t<DataT>>, detail::ConvertToOpenCLType_t<vec_data_t<float>>>, vec<float, 1>>' (aka 'sycl::vec<float, 1>') and 'scalar_type' (aka 'unsigned long'))
  277 |                 ((__res_tmp - __engine.min()) / (1 + static_cast<scalar_type>(__engine.max() - __engine.min()))) *
      |                   ~~~~~~~~~ ^ ~~~~~~~~~~~~~~`

### DIFF
--- a/include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
+++ b/include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
@@ -274,7 +274,7 @@ class uniform_real_distribution
             auto __engine_output = __engine();
             auto __res_tmp = __engine_output.template convert<scalar_type, sycl::rounding_mode::rte>();
             __res_tmp =
-                ((__res_tmp - __engine.min()) / (1 + static_cast<scalar_type>(__engine.max() - __engine.min()))) *
+                ((__res_tmp - static_cast<scalar_type>(__engine.min())) / (1 + static_cast<scalar_type>(__engine.max() - __engine.min()))) *
                     (__params.b() - __params.a()) +
                 __params.a();
 
@@ -288,7 +288,7 @@ class uniform_real_distribution
             auto __engine_output = __engine(__tail_size);
             auto __res_tmp = __engine_output.template convert<scalar_type, sycl::rounding_mode::rte>();
             __res_tmp =
-                ((__res_tmp - __engine.min()) / (1 + static_cast<scalar_type>(__engine.max() - __engine.min()))) *
+                ((__res_tmp - static_cast<scalar_type>(__engine.min())) / (1 + static_cast<scalar_type>(__engine.max() - __engine.min()))) *
                     (__params.b() - __params.a()) +
                 __params.a();
             for (int __j = 0; __j < __tail_size; __j++)
--- a/include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
+++ b/include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
@@ -212,11 +212,13 @@ class uniform_real_distribution
         auto __engine_output = __engine();
         result_type __res{};
 
-        for (int __i = 0; __i < _Ndistr; ++__i){
-            __res[__i] = ((__engine_output[__i] - __engine.min()) / static_cast<scalar_type>(1 + (__engine.max() - __engine.min()))) *
-                        (__params.b() - __params.a()) +
-                    __params.a();
-        }
+        for (int __i = 0; __i < _Ndistr; ++__i)
+        {
+            __res[__i] = ((__engine_output[__i] - __engine.min()) /
+                          static_cast<scalar_type>(1 + (__engine.max() - __engine.min()))) *
+                             (__params.b() - __params.a()) +
+                         __params.a();
+         }
 
         return __res;
     }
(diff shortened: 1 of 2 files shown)


aelizaro (dismissed):
Looks good to me!

ElenaTyuleneva:
> There are more uses of __engine.min() etc. without type conversion in this file.

Thanks for pointing this out! I reproduced the issue on `test/xpu_api/random/device_tests/engine_device_test.pass.cpp` test and it showed only 2 problematic places, which were fixed. Build of the another tests revealed more places, I'll also fix them. 

> Actually, so many type conversions needed suggest to check why the types do not match, and if possible eliminate the reason for that. Is the public API of engines such that this mismatch is inevitable?

Discussed with @paveldyakov, that we can eliminate the explicit casts by reordering the operations. Currently the logic is like the following:
1. get the output of an `__engine`
2. **Convert it to `scalar_type` of the distribution**
3. Subtract the `__engine.min()` from the converted value(type mismatch here in case scalar_type of the distribution!=scalar_type ) and do other calculations according to the distribution algorithm.

If we reorder that in the following way we can potentially remove the need in the explicit casts.:
1. get the output of an `__engine`
2. **Subtract the `__engine.min()` from the engine's output (both numbers are of the same type)**
3. **Convert temporary result to `scalar_type` of the distribution**
4. Do other calculations according to the distribution algorithm.


@aelizaro , do you have any objections against such fix?

andreyfe1:
I've made changes for `uniform_real_distribution` only. If these are ok, I'll change other files

include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
```diff
@@ -210,14 +210,14 @@ class uniform_real_distribution
     generate(_Engine& __engine, const param_type& __params)
     {
         auto __engine_output = __engine();
-        result_type __res;
+        result_type __res{};
 
-        for (int __i = 0; __i < _Ndistr; ++__i)
-            __res[__i] = static_cast<scalar_type>(__engine_output[__i]);
+        for (int __i = 0; __i < _Ndistr; ++__i){
+            __res[__i] = ((__engine_output[__i] - __engine.min()) / static_cast<scalar_type>(1 + (__engine.max() - __engine.min()))) *
```
paveldyakov, line 216: I expect that here can be overflow "1 + (__engine.max() - __engine.min())"
In case min() returns 0 and max() returns INT_MAX

Question: will the performance be affected? You changed the calculation from sycl::vec based to loop

include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
```diff
@@ -226,9 +226,8 @@ class uniform_real_distribution
     generate(_Engine& __engine, const param_type& __params)
     {
         auto __engine_output = __engine();
-        auto __res = static_cast<scalar_type>(__engine_output);
-        __res = ((__res - static_cast<scalar_type>(__engine.min())) /
-                 (1 + static_cast<scalar_type>(__engine.max() - __engine.min()))) *
+        scalar_type __res = ((__engine_output - __engine.min()) /
+                 static_cast<scalar_type>(1 + (__engine.max() - __engine.min()))) *
```
paveldyakov, line 230: Here and at the next places the question about overflow for "1 + (__engine.max() - __engine.min())"

include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
paveldyakov: Note: the case for _Ndistr == _Nengine is considered as the most performant one.

andreyfe1: I don't see any performance issues since we don't call any SYCL math functions. AFAIK, sycl::vec is just syntactic sugar that doesn't improve performance of the code if we don't call special math functions like sycl::sin, sycl::sqrt, ...

paveldyakov (approved):
My question have been addressed

include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
```diff
@@ -210,14 +210,16 @@ class uniform_real_distribution
     generate(_Engine& __engine, const param_type& __params)
     {
         auto __engine_output = __engine();
-        result_type __res;
+        result_type __res{};
```
aelizaro, line 213: why do we need {} here?

include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
```diff
@@ -226,11 +228,10 @@ class uniform_real_distribution
     generate(_Engine& __engine, const param_type& __params)
     {
         auto __engine_output = __engine();
-        auto __res = static_cast<scalar_type>(__engine_output);
-        __res = ((__res - static_cast<scalar_type>(__engine.min())) /
-                 (1 + static_cast<scalar_type>(__engine.max() - __engine.min()))) *
-                    (__params.b() - __params.a()) +
-                __params.a();
+        scalar_type __res =
```
aelizaro, line 231: if we change 1 line on a loop, will we call and compute static_cast<scalar_type>(1 + (__engine.max() - __engine.min()))) each time?

include/oneapi/dpl/internal/random_impl/uniform_real_distribution.h
andreyfe1: To not keep uninitialized values in the code

aelizaro (approved):

paveldyakov (approved):

paveldyakov (approved):
LGTM

aelizaro (approved):

PR #1796, opened 2024-08-27
[test] Remove single_pass_scan target duplicate

dmitriy-sobolev:
The PR removes a `single_pass_scan` target, which is a duplicate of `single_pass_scan_dpwi8_wgs256_uint32`.
The main motivation is to have all the targets under `build-scan-kt-tests`/`run-scan-kt-tests` umbrellas for easier test automation.

### DIFF
--- a/test/kt/CMakeLists.txt
+++ b/test/kt/CMakeLists.txt
@@ -131,7 +131,6 @@ if (ONEDPL_TEST_ENABLE_KT_ESIMD)
 endif()
 
 function (_generate_gpu_scan_test _data_per_work_item _work_group_size _type)
-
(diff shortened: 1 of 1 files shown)


dmitriy-sobolev:
@adamfidel, let me know if you prefer adding `single_pass_scan` into `build-scan-kt-tests`/`run-scan-kt-tests`. I am slightly inclined towards simply removing `single_pass_scan`, but I may be persuaded to follow the other way.

adamfidel:
The `single_pass_scan` target was mainly used as a convenience while developing and debugging the scan kernel template. As it is redundant and not strictly necessary, I am not opposed to removing it.

timmiesmith (approved):
LGTM. Thank you.

PR #1797, opened 2024-08-27
Iterators C++20 disable `operator*` when it invalid

danhoeflinger:
oneDPL's iterators like `zip_iterator`, `transform_iterator`, `permutation_iterator` can all use `sycl_iterator` as a source iterator, but `sycl_iterator` does not allow dereference.  
This PR makes the `operator*` require `std::indirectly_readable` for our iterators when c++20 concepts exist.  This fixes a bug in windows MSVC STL which was previously unable to correctly determine if a `transform_iterator<sycl_iterator>` is dereferenceable.

In practice, we never use the dereference because it is converted to the appropriate accessor for use within a kernel by our input data processing.  However, we do not control if STL implementations might try to use dereference, so we must make sure they can correctly determine if it is available.

### DIFF
--- a/include/oneapi/dpl/pstl/iterator_impl.h
+++ b/include/oneapi/dpl/pstl/iterator_impl.h
@@ -106,11 +106,12 @@ class zip_forward_iterator
     zip_forward_iterator() : __my_it_() {}
     explicit zip_forward_iterator(_Types... __args) : __my_it_(__tuple_t<_Types...>{__args...}) {}
 
-    reference operator*() const
+    reference
+    operator*() const
 #if _ONEDPL_CPP20_CONCEPTS_PRESENT
         // On windows, this is required so that concepts in MSVC STL do not detect the iterator as dereferenceable when
         // a source iterator is a sycl_iterator, which is a supported type.
-        requires (std::indirectly_readable<_Types> && ...)
+        requires(std::indirectly_readable<_Types> &&...)
 #endif
     {
         return __make_references<reference>()(__my_it_, ::std::make_index_sequence<__num_types>());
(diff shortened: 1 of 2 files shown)


danhoeflinger:
I would like this to get in for 2022.7.0, to fix some build errors on windows c++20, but we should be careful in modifying the iterators.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -105,7 +106,13 @@ class zip_forward_iterator
     zip_forward_iterator() : __my_it_() {}
     explicit zip_forward_iterator(_Types... __args) : __my_it_(__tuple_t<_Types...>{__args...}) {}
 
-    reference operator*() const
+    reference
+    operator*() const
+#if _ONEDPL_CPP20_CONCEPTS_PRESENT
```
mmichel11, line 111: What about instead of a adding a preprocessor guard and adding the `requires` expression inside, we introduce a macro function that checks `_ONEDPL_CPP20_CONCEPTS_PRESENT` internally and conditionally adds the constraint? I believe oneTBB does something like this.

Something like: `_ONEDPL_REQUIRES(std::indirectory_readable<_Types> &&...)` where the macro is defined as:
```
#   define _ONEDPL_REQUIRES(req) \
    requires req
#else
#   define _ONEDPL_REQUIRES(req)
#endif
```

danhoeflinger: Yeah, this is a good idea.  Thanks, I'll make this change.

timmiesmith (approved):
This looks good to me.

mmichel11 (approved):
LGTM

PR #1800, opened 2024-08-28
Reorder histogram template parameters

danhoeflinger:
Reorder histogram template parameters and remove unused default template argument.

One option for specification of histogram (version 1a from email).

### DIFF
--- a/include/oneapi/dpl/pstl/histogram_extension_defs.h
+++ b/include/oneapi/dpl/pstl/histogram_extension_defs.h
@@ -23,8 +23,8 @@ namespace oneapi
 namespace dpl
 {
 
-template <typename _ExecutionPolicy, typename _RandomAccessIterator1, typename _Size, typename _RandomAccessIterator2,
-          typename _ValueType = typename ::std::iterator_traits<_RandomAccessIterator1>::value_type>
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/histogram_extension_defs.h
akukanov, line 1: The change is fine if we consider the use of explicit template arguments as invalid/not allowed. If template arguments can be set explicitly, then it is a breaking change.

mmichel11 (approved):
LGTM

adamfidel:
Based on the discussion today, LGTM.

adamfidel (approved):

include/oneapi/dpl/pstl/histogram_extension_defs.h
danhoeflinger: After the discussion offline today, it is my understanding that we are going to view this as not a breaking change, but rather a bug fix, considering there is a very small likelihood that any user code explicitly specifies template parameters in code out in the wild.
If such an issue arises, our suggestion will be to remove the explicit template arguments, and if they are used to force a conversion to instead add an explicit conversion to the calling code.

akukanov: Agreed.

PR #1803, opened 2024-08-29
Removing oneapi::dpl::is_execution_policy_v alias from Ranges MVP1

danhoeflinger:
Patch offered to remove the `oneapi::dpl::is_execution_policy_v` alias from Ranges MVP1 for now, in anticipation of a separate patch which implements the soon to be codified specification changes.

Take it if it is useful to you.  I tried to format reasonably in context without messing up your existing formatting choices.

### DIFF
--- a/include/oneapi/dpl/pstl/execution_defs.h
+++ b/include/oneapi/dpl/pstl/execution_defs.h
@@ -82,8 +82,6 @@ inline constexpr bool is_execution_policy_v = oneapi::dpl::execution::is_executi
 } // namespace v1
 } // namespace execution
 
-using oneapi::dpl::execution::is_execution_policy_v;
-
 namespace __internal
 {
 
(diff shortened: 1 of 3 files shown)


MikeDvorskiy:
> Patch offered to remove the `oneapi::dpl::is_execution_policy_v` alias from Ranges MVP1 for now, in anticipation of a separate patch which implements the soon to be codified specification changes.
> 
> Take it if it is useful to you. I tried to format reasonably in context without messing up your existing formatting choices.

I appreciate you, Dan, but:
1) To tell the truth I don't like such big changes, which affects the all signatures and formatting and makes the signatures longer.  
2) I discussed that trait with Ruslan yesterday. He told me that the spec tells "the both traits (oneapi::dpl::is_execution_policy and oneapi::dpl::is_execution_policy_v) should be available as in oneapi::dpl and in oneapi::dpl::exeution as well".
So, we came to agreement just making visibility of both traits it is enough change for now.

danhoeflinger:
No problem at all.  What you agreed upon with Ruslan was my original suggestion, just add both.

PR #1804, opened 2024-08-29
Removing oneapi::dpl::is_execution_policy_v usage from Ranges MVP2

danhoeflinger:
Patch offered to remove the `oneapi::dpl::is_execution_policy_v` alias from Ranges MVP2 for now, in anticipation of a separate patch which implements the soon to be codified specification changes.  This PR requires the previous PR #1803  to be applied and rebased upon

Take it if it is useful to you.  I tried to format reasonably in context without messing up your existing formatting choices.


MikeDvorskiy:
Dan, please see https://github.com/oneapi-src/oneDPL/pull/1803#issuecomment-2320744089

PR #1809, opened 2024-08-30
[Tests] Consolidate "Latest" macros; update broken macros

danhoeflinger:
This PR consolidates the "Latest" macros into one place, to indicate where we currently are experiencing an issue from another library, compiler, etc.  so we want to skip certain broken tests, but not skip them indefinitely and forget about them.

The idea will be to have a task to revisit this section at least once every oneDPL release and audit the usages of these "Latest" macros to see if the issue is still present.  Any issues which have been fixed should be frozen to the last broken version.  Then once we have done this audit, we should update the macro in this location to the most recent version of the external entity.

This will help mark these "Latest" macros in the code explicitly, and also have a list of such instances to re-check prior to each release to help avoid re-interrogating the same errors in the CI.

This PR also updates the __GLIBCXX__ latest macro, for which the complex issues persist.  I will keep these as separate commits, rather than squashing upon merge.

This PR also updates the _PSTL_ICPX_FPGA_TEST_USM_VECTOR_ITERATOR_BROKEN to the latest icpx version, as it is still broken.  I plan to keep this also as a separate commit.

### DIFF
--- a/test/support/test_config.h
+++ b/test/support/test_config.h
@@ -24,11 +24,11 @@
 #endif
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// *** When updating we must audit each usage to ensure that the issue is still exists in the latest version ***
+//
 // This section contains macros representing the "Latest" version of compilers, STL implementations, etc. for use in
 // broken macros to represent the latest version of something which still has an ongoing issue. The intention is to
 // update this section regularly to reflect the latest version. 
-//
-// *** When updating we must audit each usage to ensure that the issue is still exists in the latest version ***
 //
 // When such an issue is fixed, we must replace the usage of these "Latest" macros with the appropriate version number
 // before updating to the newest version in this section.
--- a/test/support/test_config.h
+++ b/test/support/test_config.h
@@ -23,6 +23,26 @@
 #   include <ciso646>
 #endif
 
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// This section contains macros representing the "Latest" version of compilers, STL implementations, etc. for use in
+// broken macros to represent the latest version of something which still has an ongoing issue. The intention is to
+// update this section regularly to reflect the latest version. 
+//
+// *** When updating we must audit each usage to ensure that the issue is still exists in the latest version ***
+//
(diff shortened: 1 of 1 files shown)


danhoeflinger:
I'm cancelling CI actions for now, so that more urgent PRs get access to the CI.  I'll trigger a new CI once the queue has calmed down.  (Probably next week)

test/support/test_config.h
```diff
@@ -23,6 +23,26 @@
 #   include <ciso646>
 #endif
 
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// This section contains macros representing the "Latest" version of compilers, STL implementations, etc. for use in
+// broken macros to represent the latest version of something which still has an ongoing issue. The intention is to
+// update this section regularly to reflect the latest version. 
+//
+// *** When updating we must audit each usage to ensure that the issue is still exists in the latest version ***
```
adamfidel, line 31: I think this line should be moved either to the start of the comment block or the end of the comment block. As it is currently, it might be easy to miss.

adamfidel (commented):
I think this PR looks good and should make things clearer whenever a new platform enters the test matrix. I just had a small note about one of the comments.

mmichel11:
LGTM also when Adam's suggestion is added.

test/support/test_config.h
danhoeflinger: Moved it to the top.  Thanks.

mmichel11 (approved):
LGTM

test/support/test_config.h
```diff
@@ -23,6 +23,26 @@
 #   include <ciso646>
 #endif
 
+////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
+// *** When updating we must audit each usage to ensure that the issue is still exists in the latest version ***
```
timmiesmith, line 27:
```suggestion
// *** When updating we must audit each usage to ensure that the issue still exists in the latest version ***

```

adamfidel (dismissed):
LGTM

timmiesmith (commented):
one small nit, otherwise it looks good to me.

timmiesmith (approved):
LGTM. Thank you.

SergeyKopienko (commented):
LGTM

PR #1811, opened 2024-08-30
Fix return type of transform scan and copy if patterns

adamfidel:
As discovered in https://github.com/oneapi-src/oneDPL/pull/1762#discussion_r1730044386, there is a mismatch in the return type for the transform scan pattern. A similar issue exists with `copy_if`. This PR uses the difference type of the output range as the size type for `n` to match the return type of the patterns.


danhoeflinger (approved):
LGTM

mmichel11 (approved):
LGTM

julianmi (approved):
LGTM

PR #1812, opened 2024-08-30
Bug fix for fwd definition of __compare in sycl_traits.h

danhoeflinger:
As noted by a comment from @dmitriy-sobolev https://github.com/oneapi-src/oneDPL/pull/1621#discussion_r1739290933
after the PR was merged, there was a problem with the forward declaration of `__compare` in that is was misaligned with the definition of the struct. 

The PR resolves it.

@rarutyun  if we want to make any other changes to `__compare` here for ranges, we can.


dmitriy-sobolev (approved):
LGTM.

By the way, I have added std::invoke to __compare in MVP2 PR. I suppose it would be better to leave this PR as it is to prevent merge conflicts.

PR #1814, opened 2024-09-03
[RNG] Fixed warnings in the <random> part

andreyfe1:
Fixed warnings in the `<random>` part. Warnings are shown if  `-Wall -Wextra` are switching on

### DIFF
--- a/test/xpu_api/random/interface_tests/engines_methods.pass.cpp
+++ b/test/xpu_api/random/interface_tests/engines_methods.pass.cpp
@@ -99,8 +99,6 @@ test_vec
 public:
     bool run(sycl::queue& queue)
     {
-        using result_type = typename Engine::scalar_type;
-
         int sum = 0;
 
         // Memory allocation
@@ -173,8 +171,6 @@ test_vec<oneapi::dpl::ranlux24_vec<N>>
 public:
     bool run(sycl::queue& queue)
     {
-        using result_type = typename oneapi::dpl::ranlux24_vec<N>::scalar_type;
-
         int sum = 0;
 
         // Memory allocation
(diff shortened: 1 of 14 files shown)


test/xpu_api/random/interface_tests/engines_methods.pass.cpp
```diff
@@ -99,8 +99,6 @@ test_vec
 public:
     bool run(sycl::queue& queue)
     {
-        using result_type = typename Engine::scalar_type;
```
paveldyakov, line 102: I think this was done by the reason
The main aim here - to test that result_type is available in the engine

paveldyakov (commented):
Just one question, other changes look great

test/xpu_api/random/interface_tests/engines_methods.pass.cpp
andreyfe1: I see. So, we can keep it and make additional changes to avoid the warning about unused variable

paveldyakov: Sounds good, Andrey

test/xpu_api/random/statistics_tests/bernoulli_distribution_test.pass.cpp
```diff
@@ -50,7 +50,7 @@ test(sycl::queue& queue, double p, int nsamples)
 {
 
     // memory allocation
-    bool samples[nsamples];
+    bool* samples = new bool[nsamples];
```
aelizaro, line 53: maybe std::vector is better here?

andreyfe1: Unfortunately, no. We can not get pointer to data using `vector<bool>`. If we speak about `vector<int>` it needs a data conversion from `sycl::vec<bool>` to `sycl::vec<int>` that complicates the code of tests

aelizaro (approved):
LGTM, thank you!

PR #1815, opened 2024-09-03
Use a proper swapping method in ranges::sort for host backends

dmitriy-sobolev:
Add separate patterns for `std::ranges::sort` due to `std::indirectly_swappable` requirement, which implies the use of `std::ranges::iter_swap`, which can be customized externally. 

The fix is not applicable to the device backend since it does not keep the information about the original range; `std::ranges::iter_swap` is applied to a local accessor rather than the original range iterators.

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -2421,37 +2421,6 @@ __pattern_sort(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomAcce
     });
 }
 
-// Add separate patterns for std::ranges::sort due to std::indirectly_swappable requirement,
-// which implies the use of std::ranges::iter_swap, which can be customized externally
-#if _ONEDPL_CPP20_RANGES_PRESENT
-template <class _Tag, class _ExecutionPolicy, class _RandomAccessIterator, class _Compare>
-void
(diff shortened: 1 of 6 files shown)


include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -2452,6 +2452,37 @@ __pattern_stable_sort(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _Ran
     });
 }
 
+// Add separate patterns for std::ranges::sort due to std::indirectly_swappable requirement,
+// which implies the use of std::ranges::iter_swap, which can be customized externally
+#if _ONEDPL_CPP20_RANGES_PRESENT
+template <class _Tag, class _ExecutionPolicy, class _RandomAccessIterator, class _Compare>
+void
+__pattern_stable_sort_ranges(_Tag, _ExecutionPolicy&&, _RandomAccessIterator __first, _RandomAccessIterator __last,
```
MikeDvorskiy, line 2460: I would suggest not introducing another pattern for sort, just add LeafSort type to existed __pattern_stable_sort instead.

dmitriy-sobolev: Yep, I will try it

dmitriy-sobolev: I've implemented it in #1843. If you find it useful, then lets continue work there and close this PR.

dmitriy-sobolev:
I'm closing the PR since the same issue has already be addressed as a part of #1843

PR #1819, opened 2024-09-04
Do not compile `philox_test.cpp` test code if unnamed lambdas are not used

mmichel11:
The recently added `philox_test.cpp` test does not do a preprocessor check if we are testing with unnamed lambda support, so we are seeing some compilation failures in CI.

I added this check. The rest of the Philox tests seem to be compiling fine.

### DIFF
--- a/test/xpu_api/random/conformance_tests/philox_test.pass.cpp
+++ b/test/xpu_api/random/conformance_tests/philox_test.pass.cpp
@@ -24,6 +24,7 @@ namespace ex = oneapi::dpl::experimental;
 int
 main()
 {
+#if TEST_UNNAMED_LAMBDAS
     sycl::queue queue = TestUtils::get_test_queue();
 
     // Reference values
@@ -54,6 +55,7 @@ main()
     err += test<ex::philox4x64_vec<16>, 10000, 16>(queue) != philox4_64_ref;
 
     EXPECT_TRUE(!err, "Test FAILED");
+#endif
 
     return TestUtils::done(TEST_UNNAMED_LAMBDAS);
 }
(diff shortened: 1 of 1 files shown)


danhoeflinger (commented):
I don't know enough about why we need to disable the RNG tests with unnamed lambdas off, but it seems this is the norm for the RNG tests.  
Do we also need this for `philox_unit_test.pass.cpp` as well? 
It seems that it calls `return TestUtils::done(TEST_UNNAMED_LAMBDAS);` but it does not actually skip the test with a preprocessor macro.
It seems like we need to add the guards, or change the `done()` call.

aelizaro:
> I don't know enough about why we need to disable the RNG tests with unnamed lambdas off, but it seems this is the norm for the RNG tests. Do we also need this for `philox_unit_test.pass.cpp` as well? It seems that it calls `return TestUtils::done(TEST_UNNAMED_LAMBDAS);` but it does not actually skip the test with a preprocessor macro. It seems like we need to add the guards, or change the `done()` call.

philox_unit_test doesn't contain any offload to devices, so it can be done with `return TestUtils::done(TEST_UNNAMED_LAMBDAS);` -> `done()` if it doesn't affect CI time significantly (I don't know in general when random numbers tests are run and when they are skipped)

ElenaTyuleneva:
Hi @mmichel11 , thanks for catching that. The issue was caused by my changes, I've fixed this thing and `philox_unit_test.pass` in a separate PR(#1820 ).

mmichel11:
@ElenaTyuleneva Thanks! Closing this PR.

PR #1820, opened 2024-09-05
[RNG] Added compilation guards for `philox_test.pass` and `philox_unit_test.pass`

ElenaTyuleneva:
These guards are necessary to filter out some testing, they were forgotten to be added in the initial PR.
Also, 2 unnecessary headers were noticed in the test, they were removed.


paveldyakov (approved):
Looks good from my point of view. @aelizaro, please review

ElenaTyuleneva:
@timmiesmith , we would like to merge these small changes to fix CI issue when `TEST_UNNAMED_LAMBDAS` is set up to 0. Can we do that or oneDPL is already passed code freeze date?

aelizaro (approved):

PR #1822, opened 2024-09-05
Additional comments for `#endif`

SergeyKopienko:
In this PR we add additional comments for `#endif` macro instruction.


danhoeflinger (commented):
Most of these are within a few lines of their matching `#if`.  
It seems unnecessary, especially for the ones with a single line.

SergeyKopienko:
Not required.

PR #1823, opened 2024-09-05
Update fsycl-pstl-offload content.

dcbenito:
Removed fsycl-pstl-offload from Introduction. 
Added fsycl-pstl-offload info as a NOTE to Execution Policies.

### DIFF
--- a/documentation/library_guide/introduction.rst
+++ b/documentation/library_guide/introduction.rst
@@ -94,10 +94,6 @@ When called with |dpcpp_short| execution policies, |onedpl_short| algorithms app
 Known Limitations
 *****************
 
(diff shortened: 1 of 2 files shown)


akukanov:
@dcbenito, that limitation note, though related to PSTL offload, does not provide a sufficiently good general description I think.

I am working on one more documentation update on top of #1825 that is now in review, and plan to add some PSTL offload information as a part of it. I think it may/will go into two places: the introduction: the "Develop and build your code" part of the introduction, and maybe "Use the C++ Standard Execution Policies" part for some additional details. The note also needs to remain I think, perhaps rephrased to avoid confusion.

akukanov:
> Hi, @akukanov should we kill this PR? Or do you want to use this as a starting point?

@dcbenito feel free to close.

PR #1827, opened 2024-09-09
Fix `std::complex` broken tests

SergeyKopienko:
In this PR we fix broken tests in current `icpx` compiler.

Also we fix text which have used `INFINITY` macro in Kernel code when `double` type is unavailable:
 - now we using instead it the new const:
```C++
namespace TestUtils
{
    // This constant was declared to avoid the substitution and evaluation of the macro INFINITY
    // inside Kernel code. Since the implementation of INFINITY macro is not regulated, there are
    // cases where temporary expressions of type double are used to calculate the value of INFINITY.
    // This lead to an error on devices where the double type is not supported.
    static constexpr float infinity_val = INFINITY;
} /* namespace TestUtils */
```


test/support/test_config.h
```diff
@@ -180,6 +180,7 @@
 #define _PSTL_ICC_TEST_COMPLEX_POW_SCALAR_COMPLEX_PASS_BROKEN_TEST_EDGES  _PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX
 #define _PSTL_ICC_TEST_COMPLEX_NORM_MINUS_INF_NAN_BROKEN_TEST_EDGES       _PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX
 #define _PSTL_ICC_TEST_COMPLEX_POLAR_BROKEN_TEST_EDGES                    _PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX
+#define _PSTL_ICC_TEST_COMPLEX_EXP_BROKEN_TEST_EDGES                     (__INTEL_LLVM_COMPILER > 0 && 20220819 < __INTEL_LLVM_COMPILER && __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER)
```
danhoeflinger, line 183: Should this be `20220819 < __GLIBCXX__` ?

SergeyKopienko: No: in the previous compiler version without these new errors and in the some current compiler version with the errors the state of `__GLIBCXX__ ` is the same:
```C++
_GLIBCXX__ = 20220819
```

test/support/test_config.h
```diff
@@ -170,19 +170,21 @@
...
-#define _PSTL_ICC_TEST_COMPLEX_POW_COMPLEX_SCALAR_PASS_BROKEN_TEST_EDGES  _PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX
-#define _PSTL_ICC_TEST_COMPLEX_POW_SCALAR_COMPLEX_PASS_BROKEN_TEST_EDGES  _PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX
-#define _PSTL_ICC_TEST_COMPLEX_NORM_MINUS_INF_NAN_BROKEN_TEST_EDGES       _PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX
-#define _PSTL_ICC_TEST_COMPLEX_POLAR_BROKEN_TEST_EDGES                    _PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX
-#define _PSTL_TEST_COMPLEX_ACOS_BROKEN_IN_KERNEL_GLIB_CXX                (_PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX && __SYCL_DEVICE_ONLY__)
-#define _PSTL_TEST_COMPLEX_EXP_BROKEN_IN_KERNEL_GLIB_CXX                 (_PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX && __SYCL_DEVICE_ONLY__)
-#define _PSTL_TEST_COMPLEX_TANH_BROKEN_IN_KERNEL_GLIB_CXX                (_PSTL_TEST_COMPLEX_OP_BROKEN_GLIBCXX && __SYCL_DEVICE_ONLY__)
+#define _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER (0 < __INTEL_LLVM_COMPILER && __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER)
+
+#define _PSTL_ICC_TEST_COMPLEX_ASIN_MINUS_INF_NAN_BROKEN_SIGNBIT          _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
```
SergeyKopienko, line 175: At current state they are really depends on `__INTEL_LLVM_COMPILER` but not from `__GLIBCXX__`
So these macros has been renamed and modified.

test/support/test_config.h
```diff
@@ -170,19 +170,21 @@
...
+#define _PSTL_ICC_TEST_COMPLEX_POW_COMPLEX_SCALAR_PASS_BROKEN_TEST_EDGES  _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_POW_SCALAR_COMPLEX_PASS_BROKEN_TEST_EDGES  _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_NORM_MINUS_INF_NAN_BROKEN_TEST_EDGES       _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_POLAR_BROKEN_TEST_EDGES                    _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_EXP_BROKEN_TEST_EDGES                     (20220819 < __INTEL_LLVM_COMPILER && __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER)
+#define _PSTL_ICC_TEST_COMPLEX_COSH_BROKEN_TEST_EDGES                    (20220819 < __INTEL_LLVM_COMPILER && __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER)
+#define _PSTL_TEST_COMPLEX_ACOS_BROKEN_IN_KERNEL                         (_PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER && __SYCL_DEVICE_ONLY__)
+#define _PSTL_TEST_COMPLEX_EXP_BROKEN                                     _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
```
SergeyKopienko, line 186: For the current compiler state this test fails on host too, so `__SYCL_DEVICE_ONLY__` part has been removed.

test/xpu_api/numerics/complex.number/complex.transcendentals/acos.pass.cpp
```diff
@@ -142,7 +142,7 @@ void test_edges()
 
 ONEDPL_TEST_NUM_MAIN
 {
-    test<float>();
+    IF_DOUBLE_SUPPORT(test<float>())
```
SergeyKopienko, line 145: The problem lies in the usage of the macro `INFINITY`.
It seems that this macro ex[ands into an expression, during the evaluation of which a `double` type appears.

test/xpu_api/numerics/complex.number/complex.transcendentals/log.pass.cpp
```diff
@@ -124,7 +124,7 @@ void test_edges()
 
 ONEDPL_TEST_NUM_MAIN
 {
-    test<float>();
+    IF_DOUBLE_SUPPORT(test<float>())
```
SergeyKopienko, line 127: The problem lies in the usage of the macro `INFINITY`.
It seems that this macro ex[ands into an expression, during the evaluation of which a `double` type appears.

test/xpu_api/numerics/complex.number/complex.transcendentals/log10.pass.cpp
```diff
@@ -62,7 +62,7 @@ void test_edges()
 
 ONEDPL_TEST_NUM_MAIN
 {
-    test<float>();
+    IF_DOUBLE_SUPPORT(test<float>())
```
SergeyKopienko, line 65: The problem lies in the usage of the macro `INFINITY`.
It seems that this macro ex[ands into an expression, during the evaluation of which a `double` type appears.

test/support/test_config.h
```diff
@@ -37,7 +37,7 @@
 // According to https://gcc.gnu.org/develop.html#timeline use last known _GLIBCXX_ to check the version of libstdc++
 #define _PSTL_TEST_LATEST_GLIBCXX 20240801
```
SergeyKopienko, line 38: @danhoeflinger This macro is no longer used by anyone and I suppose I should remove it.

test/support/test_config.h
```diff
@@ -34,10 +34,7 @@
 // When such an issue is fixed, we must replace the usage of these "Latest" macros with the appropriate version number
 // before updating to the newest version in this section.
 
-// According to https://gcc.gnu.org/develop.html#timeline use last known _GLIBCXX_ to check the version of libstdc++
-#define _PSTL_TEST_LATEST_GLIBCXX 20240801
-
-#define _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER 20250000
+#define _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER 20250100
```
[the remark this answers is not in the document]
SergeyKopienko: The state has been restored

test/xpu_api/numerics/complex.number/complex.transcendentals/acosh.pass.cpp
```diff
@@ -149,7 +149,7 @@ void test_edges()
 
 ONEDPL_TEST_NUM_MAIN
 {
-    test<float>();
+    IF_DOUBLE_SUPPORT(test<float>())
```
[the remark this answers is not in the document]
SergeyKopienko: It's only macros: https://en.cppreference.com/w/c/numeric/math/INFINITY
With /*implementation defined*/
So this mean that my case with temporary `double` values inside is possible.

test/support/test_complex.h
```diff
@@ -117,6 +117,12 @@ run_test()
 
 namespace TestUtils
 {
+    // This constant was declared to avoid the substitution and evaluation of the macro INFINITY
+    // inside Kernel code. Since the implementation of INFINITY macro is not regulated, there are
+    // cases where temporary expressions of type double are used to calculate the value of INFINITY.
+    // This lead to an error on devices where the double type is not supported.
+    static constexpr float infinity_val = INFINITY;
```
SergeyKopienko, line 124: This static `constexpr float infinity_val` we should use in Kernel code due it's evaluated to `float` value in compile-time.

SergeyKopienko:
The state of `__INTEL_LLVM_COMPILER` from [__INTEL_LLVM_COMPILER ](https://godbolt.org/z/GWMqxnc68) :
| icx version  | `__INTEL_LLVM_COMPILER` state |
|-------------|-------------------------------------|
| 2021.1.2 |   |
| 2021.2.0 |   | 
| 2021.3.0 |   |
| 2021.4.0 |   |
| 2022.0.0 |   |
| 2022.1.0 |   |
| 2022.2.0 |   |
| 2022.2.1 |   |
| 2023.0.0 |   |
| 2023.1.0 |   |
| 2023.2.1 |   |
| 2024.0.0 |   |
| 2024.1.0 |   |
| 2024.2.0 | 202110 |

test/support/test_config.h
```diff
@@ -170,19 +167,21 @@
...
+
+#define _PSTL_ICC_TEST_COMPLEX_ASIN_MINUS_INF_NAN_BROKEN_SIGNBIT          _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_COSH_MINUS_INF_MINUS_ZERO_BROKEN_SIGNBIT   _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_COSH_MINUS_ZERO_MINUS_ZERO_BROKEN_SIGNBIT  _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_POW_COMPLEX_COMPLEX_PASS_BROKEN_TEST_EDGES _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_POW_COMPLEX_SCALAR_PASS_BROKEN_TEST_EDGES  _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_POW_SCALAR_COMPLEX_PASS_BROKEN_TEST_EDGES  _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_NORM_MINUS_INF_NAN_BROKEN_TEST_EDGES       _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_POLAR_BROKEN_TEST_EDGES                    _PSTL_TEST_COMPLEX_OP_BROKEN_IN_INTEL_LLVM_COMPILER
+#define _PSTL_ICC_TEST_COMPLEX_EXP_BROKEN_TEST_EDGES                     (20240201 < __INTEL_LLVM_COMPILER && __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER)
```
SergeyKopienko, line 180: I have used here the value `20240201` reasons:
1) it's the last public version of - please see [oneAPI 2024.2, Compiler Release 2024.2.1](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-dpc-c-compiler-release-notes.html)
2) how it's enumerated usually - please see my [table](https://github.com/oneapi-src/oneDPL/pull/1827#issuecomment-2343022524)

So I believe the value `20240201` is correct.

test/support/test_config.h
```diff
@@ -266,9 +265,9 @@
     (_MSC_VER && TEST_DPCPP_BACKEND_PRESENT && __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER)
 
 // Intel(R) oneAPI DPC++/C++ compiler produces 'Unexpected kernel lambda size issue' error
-#define _PSTL_LAMBDA_PTR_TO_MEMBER_WINDOWS_BROKEN (_MSC_VER && TEST_DPCPP_BACKEND_PRESENT && __INTEL_LLVM_COMPILER <= 20250100)
+#define _PSTL_LAMBDA_PTR_TO_MEMBER_WINDOWS_BROKEN (_MSC_VER && TEST_DPCPP_BACKEND_PRESENT && __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER)
 
 // Intel(R) oneAPI DPC++/C++ compiler produces 'Unexpected kernel lambda size issue' error
-#define _PSTL_LAMBDA_PTR_TO_MEMBER_WINDOWS_BROKEN (_MSC_VER && TEST_DPCPP_BACKEND_PRESENT && __INTEL_LLVM_COMPILER <= 20250100)
+#define _PSTL_LAMBDA_PTR_TO_MEMBER_WINDOWS_BROKEN (_MSC_VER && TEST_DPCPP_BACKEND_PRESENT && __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER)
```
danhoeflinger, line 271: These seem to be unrelated to this PR, but I still support making the change.  I'd prefer then end up as a separate commit in the history, but I'm not very concerned about this either way.

test/support/test_complex.h
danhoeflinger: This is clever, I like it.

test/xpu_api/numerics/complex.number/complex.transcendentals/acosh.pass.cpp
```diff
@@ -29,7 +29,7 @@ template <class T>
 void
 test()
 {
-    test(dpl::complex<T>(INFINITY, 1), dpl::complex<T>(INFINITY, 0));
+    test(dpl::complex<T>(TestUtils::infinity_val, 1), dpl::complex<T>(TestUtils::infinity_val, 0));
```
danhoeflinger, line 32: So, this is a nit pick, and probably not worth caring about, because I think all standard library implementations we see have a float infinity, but technically the standard doesn't require it
https://en.cppreference.com/w/cpp/types/numeric_limits/has_infinity

Just leaving this comment to note that, but I don't think anything needs to be done.  If we wanted this to be portable to any possible STL implementation, we should probably check `has_infinity` for the types at compile time before continuing with any INFINITY tests.

test/xpu_api/numerics/complex.number/complex.transcendentals/acos.pass.cpp
```diff
@@ -54,12 +54,12 @@ void test_edges()
         }
         else if (std::isfinite(testcases[i].real()) && std::isinf(testcases[i].imag()))
         {
-#if !_PSTL_TEST_COMPLEX_ACOS_BROKEN_IN_KERNEL_GLIB_CXX      // testcases[45]
+#if !_PSTL_TEST_COMPLEX_ACOS_BROKEN_IN_KERNEL               // testcases[45]
             assert(is_about(r.real(), pi/2));
-#endif // _PSTL_TEST_COMPLEX_ACOS_BROKEN_IN_KERNEL_GLIB_CXX
-#if !_PSTL_TEST_COMPLEX_ACOS_BROKEN_IN_KERNEL_GLIB_CXX      // testcases[147]
+#endif
+#if !_PSTL_TEST_COMPLEX_ACOS_BROKEN_IN_KERNEL               // testcases[147]
```
danhoeflinger, line 60: Can we remove these two lines, and just protect this with one guard?
Or do we separate them for some specific reason

SergeyKopienko: Sure, done.

danhoeflinger (approved):
LGTM

PR #1833, opened 2024-09-10
Always use size 16 sub-groups in single work-group radix sort if supported

mmichel11:
Previously, an IGC workaround has caused kernels compiled with `[[sycl::reqd_sub_group_size(32)]]` to be silently compiled with size 16 sub-groups with `-O0` on certain devices. To comply with the SYCL spec, this case will now throw a synchronous exception at JIT time.

Single work-group radix sort uses a sub-group sizes of 32 for the smallest inputs and 16 for larger inputs to minimize register pressure. I have experimented with the current mainline version, a version only using sub-group size of 16, and a version removing the requirement altogether. The first two versions performed similarly with no measurable difference, and explicitly requiring a sub-group size of 16 offered up to a ~15% benefit over the version with no requirement at all.

By using only sub-group sizes of 16 in single work-group radix sort, we are able to avoid the IGC issue while not impacting performance.


danhoeflinger:
There is a comment on line 819 of `parallel_backend_sycl_radix_sort.h` which needs to be updated with this PR.

danhoeflinger (approved):
LGTM, but I'd like @MikeDvorskiy  to also be able to look at this.

mmichel11:
> There is a comment on line 819 of `parallel_backend_sycl_radix_sort.h` which needs to be updated with this PR.

I made a few tweaks to the comment to match the new behavior.

PR #1834, opened 2024-09-10
removed unused _ValueType type

lslusarczyk:
Fixed oneDPL which has some unused type and that fails DR compilation.


julianmi (approved):
LGTM

PR #1835, opened 2024-09-10
More documentation improvements

akukanov:
This patch proposes
- A new paragraph in "Introduction" that explains interactions with `-fscyl-pstl-offload`;
- A reworked "known issue" note associated with PSTL offload;
- Using links to the "current" versions of other oneAPI documents instead of some specific versions;
- *[WIP, not yet in the draft]* Clarification of dependencies between policy types and passing data to algorithms
- A number of modifications related to execution policies, in particular:
  - The terms "standard aligned" aka "host" execution policies for `par_unseq`, `par` etc.;
  - Consistent use of "SIMD execution" when referring to unsequenced policies
  - Extended and improved desccription
 
as well as formatting and other improvements.


dcbenito (changes requested):
Small changes, but otherwise LGTM.

documentation/library_guide/introduction.rst
```diff
@@ -72,10 +72,22 @@ Follow the steps below to build your code with |onedpl_short|:
 Below is an example of a command line used to compile code that contains |onedpl_short| parallel algorithms
 on Linux* (depending on the code, parameters within [] could be unnecessary):
 
-.. code:: cpp
+.. code::
 
   icpx [-fsycl] [-fiopenmp] program.cpp [-ltbb] -o program
 
+You may also use `the -fsycl-pstl-offload option
```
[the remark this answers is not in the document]
akukanov: Done.

documentation/library_guide/kernel_templates/esimd_main.rst
```diff
@@ -2,7 +2,7 @@ ESIMD-Based Kernel Templates
 ############################
 
 The ESIMD kernel templates are based on `Explicit SIMD SYCL extension
-<https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2024-0/explicit-simd-sycl-extension.html>`_
+<https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/current/explicit-simd-sycl-extension.html>`_
```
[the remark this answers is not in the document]
akukanov: Done.

documentation/library_guide/parallel_api/execution_policies.rst
```diff
@@ -175,7 +178,7 @@ Use it to create customized policy objects or pass directly when invoking an alg
    To find out how to choose a more precise value, refer to the `unroll Pragma <https://www.intel.com/content/www/us/en/docs/oneapi-fpga-add-on/developer-guide/current/unroll-pragma.html>`_
    and `Loop Analysis <https://www.intel.com/content/www/us/en/docs/oneapi-fpga-add-on/developer-guide/current/loop-analysis.html>`_ content in
    the `Intel® oneAPI FPGA Handbook
-   <https://www.intel.com/content/www/us/en/docs/oneapi-fpga-add-on/developer-guide/2024-0/intel-oneapi-fpga-handbook.html>`_.
+   <https://www.intel.com/content/www/us/en/docs/oneapi-fpga-add-on/developer-guide/current/intel-oneapi-fpga-handbook.html>`_.
```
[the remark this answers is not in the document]
akukanov: The change is removed.

documentation/library_guide/common_cross_document_links.txt
```diff
@@ -1,26 +1,32 @@
...
+
+.. |vector_pragma| replace:: pragma vector documentation in the Intel® oneAPI DPC++/C++ Compiler Developer Guide and Reference
+.. _vector_pragma: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2024-2/vector.html
+
+.. |pstl_offload_option| replace:: ``-fsycl-pstl-offload`` option
+.. _pstl_offload_option: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2024-2/fsycl-pstl-offload.html
+
+.. |esimd_sycl_extension| replace:: Explicit SIMD SYCL extension
+.. _esimd_sycl_extension: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2024-2/explicit-simd-sycl-extension.html
+
```
akukanov, line 21: These two substitutions are new, everything else is the same, while the diff highlights end-of-line modifications made by Git.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -1,21 +1,36 @@
...
-* Unified shared memory (USM) pointers
+* ``oneapi::dpl::begin`` and ``oneapi::dpl::end`` functions for a SYCL buffer
+* `Unified shared memory (USM)<https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:usm>`_ pointers
 * ``std::vector`` with or without a USM allocator
 
+For an algorithm to access data, it is important that the used execution policy matches the data storage type.
+The following table shows which execution policies can be used with various data storage types.
+
+================================== =============== =============
+Data Storage                       Device policies Host policies
```
dcbenito, line 24: Please capitalize policies in both header cells.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -1,21 +1,36 @@
...
+================================== =============== =============
+SYCL buffer                        Yes             No
+Device-allocated USM               Yes             No
+Shared and host-allocated USM      Yes             Yes
+std::vector with a USM allocator   Yes             Yes
```
dcbenito, line 29: Does std::vector need monospaced font? (``std::vector``)

dcbenito (approved):
I have two minor changes, otherwise LGTM.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: Done.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: Also done.

dcbenito (approved):

documentation/library_guide/introduction.rst
```diff
@@ -70,10 +70,21 @@ Follow the steps below to build your code with |onedpl_short|:
 Below is an example of a command line used to compile code that contains |onedpl_short| parallel algorithms
 on Linux* (depending on the code, parameters within [] could be unnecessary):
 
-.. code:: cpp
+.. code::
 
   icpx [-fsycl] [-fiopenmp] program.cpp [-ltbb] -o program
 
+You may also use the |pstl_offload_option|_ of |dpcpp_cpp| powered by |onedpl_short|
```
timmiesmith, line 77: I like the wording here with "powered by".

documentation/library_guide/introduction.rst
```diff
@@ -154,10 +164,3 @@ Known Limitations
   the dereferenced value type of the provided iterators should satisfy the ``DefaultConstructible`` requirements.
 * For ``remove``, ``remove_if``, ``unique`` the dereferenced value type of the provided
   iterators should be ``MoveConstructible``.
-* The algorithms that process uninitialized storage: ``uninitialized_copy``, ``uninitialized_copy_n``, ``uninitialized_fill``, ``uninitialized_fill_n``, ``uninitialized_fill_n``, ``uninitialized_move``,
-  ``uninitialized_move_n``, ``uninitialized_default_construct``, ``uninitialized_default_construct_n``, ``uninitialized_value_construct``, ``uninitialized_value_construct_n``
-  should be called with a device policy when using device data and should be called with a host policy when using host data. Otherwise, the result is undefined.
-* The algorithms that destroy data: ``destroy`` and ``destroy_n`` should be called with a host policy when using host data that was initialized on the host, and should be called with a device policy when using device data that was initialized on the device. Otherwise, the result is undefined.
```
danhoeflinger, line 160: Are these notes no longer relevant?

documentation/library_guide/parallel_api/execution_policies.rst
```diff
@@ -1,65 +1,70 @@
...
---------------------------------- ------------------------------
-``dpcpp_default``                 Massive parallel execution on devices using |dpcpp_short|.
---------------------------------- ------------------------------
-``dpcpp_fpga``                    Massive parallel execution on FPGA devices.
-================================= ==============================
+According to `the oneAPI specification
+<https://uxlfoundation.github.io/oneAPI-spec/spec/elements/oneDPL/source/index.html>`_,
+|onedpl_long| (|onedpl_short|) provides execution policies semantically aligned with the C++ standard,
+also referred to as *standard aligned* or *host execution policies*, and the *device execution policies*
+to run data parallel computations on heterogeneous systems.
```
danhoeflinger, line 8: It may be more clear if we make this into two sentences.  Change the comma to a period then start the next sentence as "|onedpl_short| also provides the *device execution policies* ..."

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -1,20 +1,36 @@
 Pass Data to Algorithms
 #######################
 
-When using the C++ standard execution policies, oneDPL supports data being passed to its algorithms as specified
-in the ISO/IEC 14882:2017 standard (commonly called C++17). According to the standard, the calling code
-must prevent data races when using algorithms with parallel execution policies.
+When using the C++ standard aligned (or *host*) execution policies, oneDPL supports data being passed to its
```
timmiesmith, line 4:
```suggestion
When using the C++ standard aligned (or *host*) execution policies, |onedpl_short| supports data being passed to its
```

timmiesmith: This isn't needed to merge, but if there are other changes prior to final approval this would make the use consistent.

documentation/library_guide/introduction.rst
akukanov: I attempted to provide the same information in the additions to "Pass data to algorithms", where I think it is a better fit.

documentation/library_guide/parallel_api/execution_policies.rst
akukanov: OK, I will think of it - but perhaps in another patch, unless bigger changes are requested for this one.

danhoeflinger: not required, the rest LGTM.  About to approve.

danhoeflinger (approved):
LGTM

documentation/library_guide/introduction.rst
```diff
@@ -70,10 +70,21 @@ Follow the steps below to build your code with |onedpl_short|:
...
 
+You may also use the |pstl_offload_option|_ of |dpcpp_cpp| powered by |onedpl_short|
+to build the standard C++ code for execution on a SYCL device:
+
+.. code::
+
+  icpx -fsycl -fsycl-pstl-offload=gpu program.cpp -o program
+
+This option redirects C++ parallel algorithms invoked with the ``std::execution::par_unseq`` policy
+to |onedpl_short| algorithms. It does not change the behavior of the |onedpl_short| execution policies and algorithms
```
rarutyun, line 85: While I understand (after several reading) that onedpl_short is related to both execution policies and algorithms, perhaps adding it explicitly for the second time helps make this sentence clearer Maybe adding "themselves" also could help. I mean something like:
```suggestion
to |onedpl_short| algorithms. It does not change the behavior of the |onedpl_short| execution policies and |onedpl_short| algorithms themselves
```

documentation/library_guide/introduction.rst
danhoeflinger: Makes sense, I see that addition now.

documentation/library_guide/parallel_api/execution_policies.rst
```diff
@@ -1,65 +1,70 @@
...
+``par_unseq``          The standard aligned policy with the combined effect of ``unseq`` and ``par``.
+---------------------- -----------------------------------------------------
+``device_policy``      The class template to create device policies for data parallel execution.
+---------------------- -----------------------------------------------------
+``dpcpp_default``      The device policy for data parallel execution on the default SYCL device.
+---------------------- -----------------------------------------------------
+``fpga_policy``        The class template to create policies for execution on FPGA devices.
+---------------------- -----------------------------------------------------
+``dpcpp_fpga``         The device policy for data parallel execution on a SYCL FPGA device.
+====================== =====================================================
```
rarutyun, line 32: I agree that it's bad having the fpga policies for so many years and not list them in the documentation, on the other hand, we were discussing that we probably want to revise the names of our global policies. Having them documented would require us to go through deprecation way, because it will be a breaking change. How do you feel about leaving the status quo (before changes in the PR)?

documentation/library_guide/parallel_api/execution_policies.rst
```diff
@@ -1,65 +1,70 @@
...
+<https://uxlfoundation.github.io/oneAPI-spec/spec/elements/oneDPL/source/index.html>`_,
+|onedpl_long| (|onedpl_short|) provides execution policies semantically aligned with the C++ standard,
+also referred to as *standard aligned* or *host execution policies*, and the *device execution policies*
+to run data parallel computations on heterogeneous systems.
+
+The execution policies are defined in the ``oneapi::dpl::execution`` namespace and provided
+in the ``oneapi/dpl/execution`` header. The policies have the following meaning:
+
+====================== =====================================================
+Policy Value or Type   Description
```
MikeDvorskiy, line 14: Probably, does it makes sense to "split" (to have two tables) for the host policies
(policy values only) and for device policy types and policy values.
Otherwise, the column "Policy Value or Type" might confuse a user.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -1,20 +1,36 @@
...
+
+For an algorithm to access data, it is important that the used execution policy matches the data storage type.
+The following table shows which execution policies can be used with various data storage types.
+
+====================================== =============== =============
+Data Storage                           Device Policies Host Policies
+====================================== =============== =============
+SYCL buffer                            Yes             No
+Device-allocated USM                   Yes             No
+Shared and host-allocated USM          Yes             Yes
```
rarutyun, line 29: I am not sure if it makes sense but how about something like 
```suggestion
Shared and host-allocated USM (including USM allocator)          Yes             Yes
```
The reason why I am asking that is because we don't mention `usm_allocator` separately from `std::vector`. To me having `usm_allocator` without `std::vector` is a valid use-case

documentation/library_guide/parallel_api/execution_policies.rst
akukanov: Thanks.
The description indicates what is type and what is a value. I can also add "type" to the types :)

documentation/library_guide/parallel_api/execution_policies.rst
```diff
@@ -1,65 +1,70 @@
...
-   #. ``#include <oneapi/dpl/memory>``
+   - ``#include <oneapi/dpl/algorithm>``
+   - ``#include <oneapi/dpl/numeric>``
+   - ``#include <oneapi/dpl/memory>``
 
-#. Pass a |onedpl_short| execution policy object, defined in the ``oneapi::dpl::execution``
-   namespace, to a parallel algorithm.
-#. Use the C++ standard execution policies:
+#. Pass a |onedpl_short| execution policy object as the first argument to a parallel algorithm
+   to specify the desired execution behavior.
```
MikeDvorskiy, line 57: "desired execution behavior" - I remembered that a host execution policy doesn't  guarantee execution behavior, it may have the desired behavior, it is just a hint. 
For a device policy provides the desired (device execution) execution behavior.
Probably this difference should be explained in the doc.

akukanov: Thanks, I will think of it.

documentation/library_guide/introduction.rst
akukanov: "... that are directly used in the code" below should convey similar meaning. Anyway, I will think how to make it clearer.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -1,20 +1,36 @@
...
 
 When using a device execution policy, you can use one of the following ways to pass data to an algorithm:
 
-* ``oneapi::dpl::begin`` and ``oneapi::dpl::end`` functions
-* Unified shared memory (USM) pointers
-* ``std::vector`` with or without a USM allocator
+* ``oneapi::dpl::begin`` and ``oneapi::dpl::end`` functions for a `SYCL buffer
+  <https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:buffers>`_;
+* `Unified shared memory (USM) <https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:usm>`_ pointers;
+* ``std::vector`` with or without a USM allocator.
```
MikeDvorskiy, line 19: `std::vector with or without a USM allocator.`
It is not quite correct and  not consistent with the first and the second bullets  above.
Here we are telling  about  Iterator based API, and the first and the second bullet tells about first and last iterators (or pointers).
So, the third bullet should tell the same as the first I think: 
`std::begin and std::end functions for a `std::vector with or without a USM allocator.``
(Or even `oneapi::dpl::begin` `oneapi::dpl::end` for USM std::vector or we already have such customization)

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -1,20 +1,36 @@
...
+For an algorithm to access data, it is important that the used execution policy matches the data storage type.
+The following table shows which execution policies can be used with various data storage types.
+
+====================================== =============== =============
+Data Storage                           Device Policies Host Policies
+====================================== =============== =============
+SYCL buffer                            Yes             No
+Device-allocated USM                   Yes             No
+Shared and host-allocated USM          Yes             Yes
+``std::vector`` with a USM allocator   Yes             Yes
```
MikeDvorskiy, line 30: I think should be like that:
```std::vector`` USM allocator with a shared or host types`

File: documentation/library_guide/parallel_api/pass_data_algorithms.rst (hunk 14 above)
Comment (MikeDvorskiy) [inline documentation/library_guide/parallel_api/pass_data_algorithms.rst:29]: => `Shared or host-allocated USM` ?

File: documentation/library_guide/parallel_api/pass_data_algorithms.rst [hunk 18]
```diff
@@ -1,20 +1,36 @@
...
+``std::vector`` with a host allocator  See below       Yes
+Other host-allocated data              No              Yes
```
MikeDvorskiy, line 32: `Other host-allocated data No`
It is not quite right. We have implicit data coping to a device in case if the `first, last` are not pointer types.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
```diff
@@ -1,20 +1,36 @@
...
+The following table shows which execution policies can be used with various data storage types.
+
+====================================== =============== =============
+Data Storage                           Device Policies Host Policies
+====================================== =============== =============
+SYCL buffer                            Yes             No
+Device-allocated USM                   Yes             No
+Shared and host-allocated USM          Yes             Yes
+``std::vector`` with a USM allocator   Yes             Yes
+``std::vector`` with a host allocator  See below       Yes
```
MikeDvorskiy, line 31: ```std::vector`` with a host allocator See below`
1) "See below" - there is no link where I should "see below".
2) ````std::vector`` with a host allocator` confuses a little bit. Does it mean any host allocator? i that case it is "general C++ host  code" case... Probably, we can dont remind about that (?)

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: We need to be careful what we say here, we had a discussion about this when we last changed this page and chose this wording specifically because we don't want to tell people to use iterators to `std::vector` with a usm allocator.  This is explained in the final paragraph of this page.  
Some implementations of STL do not include the allocator information in the iterator, so it may trigger an additional copy to use the iterators rather than `std::vector::data()` and `std::vector::size()`.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: There is no such thing as a USM allocator for USM device memory (intentional choice in SYCL specification).  It still could be useful to point it out for clarity, but I believe this is technically true as written.  (Again, we want to encourage people to use the direct pointers to the underlying data if they used this type, rather than the iterators.)

documentation/library_guide/parallel_api/pass_data_algorithms.rst
danhoeflinger: So, even if it is technically possible that we support some "other host-allocated data" which follow some rules with device policies, this may be the best way to have it in the documentation.  I think we don't want to advertise any stronger support for types which we don't test in some capacity. 
If people figure out how to get it to work anyway, that's OK, but its relying upon undefined behavior.

documentation/library_guide/parallel_api/execution_policies.rst
rarutyun: Got it, thanks

akukanov:
The new feedback will be addressed in another patch.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: Correct, the only provided guarantee is for `std::vector`. If requested, we can consider extending it, for example, to any contiguous iterators that are not a pointer, But for now, the table matches what previously was and still is stated in this file.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: Dan is correct, see https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_c_allocator_interface

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: Are there problems with supporting "any host allocator"? For as long as the iterator is not a pointer, data copying will be used, right? The "Use std::vector" section in this very document (see below :)) does not specify details of host allocators either.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: It is a valid use case, however since we mention USM in general and not a specific way of allocation, I think it implicitly covers the allocator as well. The next rows actually focus on the vector, not on the allocator.

akukanov: Let's try full sentences:

> Device policies can be used with data in both shared and host-allocated USM.
> Device policies can be used with data in either shared or host-allocated USM.

I like the first one more. I do not think the context assumes simultaneous use. Moreover, I guess simultaneous use is possible when processing two or more data sequences.

documentation/library_guide/introduction.rst
akukanov: Resolved by swapping the words: "It does not change the behavior of the |onedpl_short| algorithms and
execution policies that are directly used in the code."

documentation/library_guide/parallel_api/execution_policies.rst
akukanov: Tried a few variations, and ended up with changing the column heading to "Policy Name / Type".

documentation/library_guide/parallel_api/execution_policies.rst
akukanov: I rephrased it as "to indicate the desired behavior" to sound a bit less formal. Anyway, the sentence describes a user action, and the behavior is desired, not guaranteed. Also I added  "possible" to execution descriptions for `unseq` and `par`.

documentation/library_guide/parallel_api/execution_policies.rst
akukanov: Changed "and ..." to "as well as device execution policies", but otherwise left it as is. I do not want to split, as the "According to the spec" part applies to both policy kinds.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: Added the link.

documentation/library_guide/parallel_api/pass_data_algorithms.rst
akukanov: I reworked this page quite significantly for the ranges doc patch, please take a look.

PR #1837, opened 2024-09-10
[test] Use macro workaround to avoid calling vectorized MSVC algorithms from within a SYCL kernel

mmichel11:
With newer MSVC versions, we see compilation errors such as:
```
error: SYCL kernel cannot call an undefined function without SYCL_EXTERNAL attribute
  334 |         return __std_mismatch_4(_First1, _First2, _Count);
      |                ^
```
which occurs when vectorized MSVC algorithms are attempted to be called from within a SYCL kernel. This is showing up with array comparisons within kernels.

To resolve this issue, we can set `_USE_STD_VECTOR_ALGORITHMS=0` which has already been applied to similar tests.

### DIFF
--- a/test/xpu_api/containers/sequences/array/array.comparison/greater_g.pass.cpp
+++ b/test/xpu_api/containers/sequences/array/array.comparison/greater_g.pass.cpp
@@ -13,6 +13,11 @@
 //
 //===----------------------------------------------------------------------===//
 
+// In Windows, as a temporary workaround, disable vector algorithm calls to avoid calls within sycl kernels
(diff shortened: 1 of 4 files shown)


danhoeflinger:
To put this in slightly more context: 

These are hand-rolled precompiled vectorized functions called from the MSVC STL implementation.  This functions are not callable from SYCL kernels, so we must define this macro to `0` to avoid using these functions *prior* to including any standard library headers.  Its unclear if a real fix is possible or necessary beyond this workaround.

danhoeflinger (approved):
LGTM

PR #1839, opened 2024-09-11
[onedpl][ranges] + a fix for "nanoranges" drop_view

MikeDvorskiy:
[onedpl][ranges] + a fix for "nanoranges" drop_view


dmitriy-sobolev (commented):
LGTM. It corresponds to the signature shown in https://en.cppreference.com/w/cpp/ranges/drop_view/end. 

Two additional points:

- I will approve it when @timmiesmith allows merging.
- I wonder if we want to sync this file with the original in the future, and how we keep track of such changes if so. Do you have any strategy?

rarutyun (approved):

dmitriy-sobolev (approved):

PR #1841, opened 2024-09-12
Fix __lifetime_keeper_base destructor

SergeyKopienko:
This PR based on issue https://github.com/oneapi-src/oneDPL/issues/1469

In this PR we rewrite the destructor of `struct __lifetime_keeper_base` :
 - instead of empty body we rewrote it as `= default;`

Why it's better: https://isocpp.org/blog/2015/04/quick-q-how-is-default-different-from-for-default-constructor-and-destructo


MikeDvorskiy (approved):
LGTM

PR #1843, opened 2024-09-13
Get rid of duplication in sort patterns

dmitriy-sobolev:
Parameterize sort patterns with a leaf sorting algorithm to get rid of duplication for unstable/stable and classic/ranged sort.
It also fixed an issue already addressed in #1815 in a less verbose way.


MikeDvorskiy (approved):
LGTM

PR #1845, opened 2024-09-16
Fix warnings in tests

SergeyKopienko:


test/xpu_api/numerics/c.math/cmath.pass.cpp
```diff
@@ -240,20 +240,23 @@ void test_isinf()
...
+            // GLIBC < 2.23 defines 'isinf(double)' with a return type of 'int' in
+            // all C++ dialects. The test should tolerate this when libc++ can't work
+            // around it.
+            // See: https://sourceware.org/bugzilla/show_bug.cgi?id=19439
+            static_assert((std::is_same_v<DoubleRetType, bool> || std::is_same_v<DoubleRetType, int>));
 #endif
-    };
-    IF_DOUBLE_SUPPORT_L(fnc)
+        };
+        fnc();
```
kboyarinov, line 258: Do we really need to define the lambda and immediately execute it? It seems like we can perform all checks inside the `if constexpr` body.

Similar simplification can be applied in other places.

SergeyKopienko: it's required, because otherwise we have compile error (in case when we place old macro argument simple under `constexpr if`

kboyarinov: I don't understand why there should be any compilation issues. What makes the lambda necessary?

I have tried manually with abs1.pass and it works.

SergeyKopienko: Fixed

SergeyKopienko: Fixed

Issue #1229, opened 2023-10-18
Add a test with multiple TUs to verify ODR compliance


dmitriy-sobolev:
Currently, oneDPL does not have such tests. It might be useful to verify ODR compliance.

dmitriy-sobolev:
It was implemented in #1854.

PR #1849, opened 2024-09-17
Fix warning and compilation errors

Xewar313:
When oneDPL is compiled without spirv, code in parallel_backend_sycl_radix_sort_one_wg.h causes unused variable error, so removing this variable should help.

Presence of function definitions in header files sometimes causes multiple definitions error, changing them to inline should solve the issue

### DIFF
--- a/include/oneapi/dpl/pstl/execution_impl.h
+++ b/include/oneapi/dpl/pstl/execution_impl.h
@@ -106,25 +106,25 @@ __select_backend(oneapi::dpl::execution::parallel_unsequenced_policy, _IteratorT
 namespace __ranges
 {
 
-::oneapi::dpl::__internal::__serial_tag<std::false_type>
+inline ::oneapi::dpl::__internal::__serial_tag<std::false_type>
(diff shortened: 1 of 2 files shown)


lslusarczyk (approved):
LGTM, @Xewar313 thank you for fixing bug that blocks distribute-ranges compilation

dmitriy-sobolev (approved):
Hi @Xewar313. Thank you for this important fix. The changes look good enough to me.

Meanwhile, I've prioritized #1229 to detect ODR violations earlier, and added #1850 for detecting warnings like the fixed one.

dmitriy-sobolev:
I've added #1851 which reverts unnecessary refactoring changes. Besides that, it would have been great to have separate PRs for each problem solved, but it is not critical, especially after merging.

dmitriy-sobolev:
@Xewar313, do not hesitate to create a PR with the description of your contribution in https://github.com/oneapi-src/oneDPL/blob/main/CREDITS.txt if you wish it to be mentioned there. It certainly deserves it.

PR #1853, opened 2024-09-17
Fix comparison of expected and actual values in tests

SergeyKopienko:
Issue: https://github.com/oneapi-src/oneDPL/issues/1382

In this PR we fixing value comparisons in oneDPL tests: the goal is to compare `floating point` values through `std::numeric_limits<T>::epsilon()` precision usage.


kboyarinov (approved):
LGTM

kboyarinov (approved):

PR #1859, opened 2024-09-18
Disable copy-constructor and copy-assignment operations for the `__future` class

SergeyKopienko:
Issue: https://github.com/oneapi-src/oneDPL/issues/1776

What we change in this PR:
- remove copy constructor and copy assignment from the `__future` : to be closet for `std::future` implementation and to avoid multiple resource (`sycl::event') owners;
- change format of `__future` constructors: in all use cases we able to move `sycl::event` into the constructor instead of it's copy;
- change format of `__future` constructors: we pass `std::tuple` into constructor by const reference or as `r-value`;
- we change the semantic of `__make_future` functions: we declare it as `static` and rewrite it to avoid extra data copy and use data move if possible.

### DIFF
--- a/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
+++ b/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
@@ -69,7 +69,8 @@ __pattern_walk2_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
         _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
         unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n, __buf1.all_view(), __buf2.all_view());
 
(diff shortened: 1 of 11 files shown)


danhoeflinger (commented):
I would resist making direct changes until we have had the chance to discuss this holistically in the context of the asynchronous API discussion.  
It is a public (experimental) API, so we may want to wait and then change it only a single time, in a deliberate way (perhaps in the context of productization).

SergeyKopienko:
> I would resist making direct changes until we have had the chance to discuss this holistically in the context of the asynchronous API discussion. It is a public (experimental) API, so we may want to wait and then change it only a single time, in a deliberate way (perhaps in the context of productization).

Unfortunately, not only (It is a public (experimental) API) : we widely using `__future` inside our code I would like to made it a little bit more optimal.

PR #1860, opened 2024-09-18
[RNG] exponential distribution operator>> fix

egrabovskaya:
I came across my copy-paste mistake in exponential_distribution code. Now it's fixed


aelizaro (approved):
LGTM, thank you for catching this!

PR #1861, opened 2024-09-18
Add Xewar313 to CREDITS.txt

Xewar313:
As @dmitriy-sobolev mentioned in https://github.com/oneapi-src/oneDPL/pull/1849 , I would be very glad if I could be added to CREDITS.txt

### DIFF
--- a/CREDITS.txt
+++ b/CREDITS.txt
@@ -18,3 +18,7 @@ N: Qually Jiang
 W: https://github.com/Quallyjiang
 E: qually.jiang@intel.com
 D: Implemented a prototype of ESIMD-based radix sort, which was adapted into oneDPL radix sort kernel templates.
+
+N: Mikołaj Komar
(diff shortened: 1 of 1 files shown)


CREDITS.txt
```diff
@@ -18,3 +18,7 @@ N: Qually Jiang
 W: https://github.com/Quallyjiang
 E: qually.jiang@intel.com
 D: Implemented a prototype of ESIMD-based radix sort, which was adapted into oneDPL radix sort kernel templates.
+
+N: Mikołaj Komar
+W: https://github.com/Xewar313
+D: Fixed an issue with compilation that occurred when including the library multiple times in a single executable.
```
dmitriy-sobolev, line 24: What do you think about making it more concise like below?

```suggestion
D: Fixed One Definition Rule violation.
```

timmiesmith (approved):
LGTM

PR #1863, opened 2024-09-19
[RNG] A note about distributions that can't be used on GPUs not supporting FP64 precision

egrabovskaya:


documentation/library_guide/api_for_sycl_kernels/random.rst
```diff
@@ -107,6 +107,9 @@ Distribution                   Description
 ``cauchy_distribution``        Produces real values according to the Cauchy distribution
 ============================== ============================================================================
 
+.. note::
+  ``bernoulli_distribution``, ``geometric_distribution``, and ``uniform_int_distribution`` cannot be used on GPU devices that don't support FP64 precision because they use this data type inside the implementation.
```
aelizaro, line 111: I propose to rephrase it in a more positive way:
```suggestion
  ``bernoulli_distribution``, ``geometric_distribution``, and ``uniform_int_distribution`` can only be used on devices with FP64 support as they rely on double precision in their implementation.
```

Maybe it is worth adding that in order to check if device supports FP64 `sycl::aspect::fp64` may be used

documentation/library_guide/api_for_sycl_kernels/random.rst
```diff
@@ -107,6 +107,9 @@ Distribution                   Description
 ``cauchy_distribution``        Produces real values according to the Cauchy distribution
 ============================== ============================================================================
 
+.. note::
+  ``bernoulli_distribution``, ``geometric_distribution``, and ``uniform_int_distribution`` can only be used on devices with FP64 support as they rely on double precision in their implementation.
```
akukanov: Should it say "... as their implementation requires double precision"? As far as I can say, this is certainly true for the first two distributions.

By the way, can the implementation of `uniform_int_distribution` be modified to use FP32 in case FP64 is not available?

aelizaro (approved):

documentation/library_guide/api_for_sycl_kernels/random.rst
akukanov: Thanks, no objections.

PR #1864, opened 2024-09-20
Remove extra `__internal` namespace specification before brick calls 

SergeyKopienko:
Issue: https://github.com/oneapi-src/oneDPL/issues/398

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -69,7 +69,7 @@ __pattern_any_of(_Tag, _ExecutionPolicy&&, _ForwardIterator __first, _ForwardIte
 {
     static_assert(__is_serial_tag_v<_Tag> || __is_parallel_forward_tag_v<_Tag>);
 
-    return __internal::__brick_any_of(__first, __last, __pred, typename _Tag::__is_vector{});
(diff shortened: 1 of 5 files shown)


dmitriy-sobolev (commented):
The PR looks good to me. I am ready to approve it once clang-format suggestions have bee applied.

Also, I added Mikhail as a reviewer to check if these changes affect the synchronization with the upstream somehow. I hope they do not.

dmitriy-sobolev (dismissed):
LGTM

MikeDvorskiy:
I have a concern here. We are changing  the full unqualified calls to unqualified ones. it can be dangerous with ADL perspective.

SergeyKopienko:
> I have a concern here. We are changing the full unqualified calls to unqualified ones. it can be dangerous with ADL perspective.

@MikeDvorskiy could you please explain?
In all changes cases we are already inside `__internal` and why we should specify `__internal` directly in this case?

dmitriy-sobolev:
Here is my understanding and the reason to dismiss the approval. 

For unqualified calls, ADL precedes the regular recursive namespace lookup. This is especially dangerous due to a chance to have the bricks with the same names in a standard library such as libstdc++. The danger is that the bricks from that library can be selected instead.

PR #1867, opened 2024-09-20
Adding note about potential issue with ranges::iter_swap customizations.

timmiesmith:


documentation/library_guide/parallel_api/parallel_range_algorithms.rst
```diff
@@ -106,6 +106,13 @@ Usage Example for Parallel Range Algorithms
         oneapi::dpl::ranges::copy(oneapi::dpl::execution::dpcpp_default, view_in, std::span(vec_out));
     }
 
+Potential Issues
+----------------
+The ``sort`` and ``stable_sort`` algorithms provided use ``std::swap`` internally instead of ``std::ranges::iter_swap``
+for swapping elements. Note that the ``std::ranges::sort`` algorithm in several C++ standard library
+implementations also does not make use of ``std::ranges::iter_swap``. The potential issue is that customizations
+targeting ``std::ranges::iter_swap`` will not be respected.
+
```
dmitriy-sobolev, line 115: I suggest slightly rephrasing the note and changing the title. 

```suggestion
Implementation Notes
---------------------------
 - The sort and ``stable_sort`` algorithms in the ``oneapi::dpl::ranges`` namespace use ``std::swap`` internally instead of ``std::ranges::iter_swap`` for swapping elements, as the C++ standard suggests but does not explicitly require.
 As a result, customizations targeting ``std::ranges::iter_swap`` will not be respected.

```

We can also add this note together with other related changes.

Signatures of some algorithms do not conform C++ standard and it is expected to be mentioned later (as noted in https://github.com/oneapi-src/oneDPL/pull/1596#discussion_r1758917326). There is a similarity with this issue, even though the addressed discrepancy is more subtle. They are likely to fit the same section, which will be more complete.

documentation/library_guide/parallel_api/parallel_range_algorithms.rst
```diff
@@ -106,6 +106,12 @@ Usage Example for Parallel Range Algorithms
         oneapi::dpl::ranges::copy(oneapi::dpl::execution::dpcpp_default, view_in, std::span(vec_out));
     }
 
+Implementation Notes
+--------------------
+The ``sort`` and ``stable_sort`` algorithms provided use ``std::swap`` internally instead of 
+``std::ranges::iter_swap`` for swapping elements, as the C++ standard suggests but does not explicitly require.
+As a result, customizations targeting ``std::ranges::iter_swap`` will not be respected.
```
akukanov, line 113: I am not sure that "the C++ standard suggests but does not require" is accurate. The standard requirements for `sort` include `sortable` which eventually **requires** `std::ranges::iter_swap(iter1, iter2)` to be a valid expression (see `indirectly_swappable`. But `sortable` also assumes `indirectly_movable_storable` which, as far as I can tell, guarantees enough for `std::swap(*iter1, *iter2)` to work. And the only kind of "suggestion" is that `std::ranges::iter_swap` checks ADL `iter_swap` (i.e. a customization) before falling back to other methods.

To avoid misinterpretation, and also make the wording more concise. here is my suggestion:
```suggestion
The ``sort`` and ``stable_sort`` algorithms use ``std::swap`` and not ``std::ranges::iter_swap`` for swapping elements.
As a result, customizations targeting ``std::ranges::iter_swap`` will not be respected.
```

akukanov (approved):

PR #1870, opened 2024-09-20
Improve SYCL backend `__parallel_for` performance for large  input sizes

mmichel11:
## Summary
This PR improves `__parallel_for` performance for large input sizes by switching to an nd-range kernel to process multiple inputs per work item which enables us to use the full hardware bandwidth.
## Details
On some target architectures, we are currently not hitting roofline memory bandwidth performance in our `__parallel_for` pattern. The cause is that our SYCL basic kernel implementation only processes a single element per item. This is insufficient to fully utilize memory bandwidth on some target architectures. Processing multiple inputs per work item enables us to perform enough loads / stores to saturate the hardware bandwidth. Explicitly using a coalesced pattern through either a sub-group or work-group stride ensures that a good access pattern is achieved.

A nd-range kernel has been added for large input sizes that uses a heuristic based upon the smallest sized type in the set of provided ranges to determine the number of iterations to process per input item. This drastically improves performance on target architectures for large inputs across nearly all for-based algorithms.

A second kernel has been added as opposed to merging both paths within a single kernel to prevent extra runtime dispatch within the kernel which hurt performance for small inputs. There is a smaller runtime overhead for selecting the best path from the host and compiling two kernels. For small-to-medium inputs, the SYCL basic kernel performs the best.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -252,7 +252,6 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
             const std::size_t __iters_per_compute_unit = oneapi::dpl::__internal::__dpl_ceiling_div(__count, __max_cu * __work_group_size);
             // For small data sizes, distribute the work evenly among compute units.
             const std::uint8_t __iters_per_work_item = std::min(__iters_per_compute_unit, static_cast<std::size_t>(__max_iters_per_work_item));
-            const bool __can_unroll_loop = __max_iters_per_work_item == __iters_per_work_item;
             const std::size_t __num_groups =
                          oneapi::dpl::__internal::__dpl_ceiling_div(__count, (__work_group_size * __iters_per_work_item));
             const std::size_t __num_items = __num_groups * __work_group_size;
@@ -263,16 +262,7 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
                     // TODO: Investigate using a vectorized approach similar to reduce.
                     // Initial investigation showed benefits for in-place for-based algorithms (e.g. std::for_each) but
                     // performance regressions for out-of-place (e.g. std::copy).
-                    if (__is_full && __can_unroll_loop)
-                    {
-                        _ONEDPL_PRAGMA_UNROLL
-                        for (std::uint8_t __i = 0; __i < __max_iters_per_work_item; ++__i)
-                        {
-                            __brick(__idx, __rngs...);
-                            __idx += __stride;
-                        }
-                    }
-                    else if (__is_full)
+                    if (__is_full)
                     {
                         for (std::uint8_t __i = 0; __i < __iters_per_work_item; ++__i)
                         {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -252,16 +252,16 @@ struct __parallel_for_small_submitter<__internal::__optional_kernel_name<_Name..
     }
 };
 
-template <typename _KernelName, typename... _Ranges>
+template <typename _KernelName, typename... _RangeTypes>
 struct __parallel_for_large_submitter;
 
-template <typename... _Name, typename... _Ranges>
-struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name...>, _Ranges...>
+template <typename... _Name, typename... _RangeTypes>
+struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name...>, _RangeTypes...>
 {
     // Flatten the range as std::tuple value types in the range are likely coming from separate ranges in a zip
     // iterator.
     using _FlattenedRangesTuple = typename oneapi::dpl::__internal::__flatten_std_or_internal_tuple<
-        std::tuple<oneapi::dpl::__internal::__value_t<_Ranges>...>>::type;
+        std::tuple<oneapi::dpl::__internal::__value_t<_RangeTypes>...>>::type;
     static constexpr std::size_t __min_type_size =
         oneapi::dpl::__internal::__min_tuple_type_size_v<_FlattenedRangesTuple>;
     // __iters_per_work_item is set to 1, 2, 4, 8, or 16 depending on the smallest type in the
@@ -320,7 +320,7 @@ struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name..
         return __work_group_size * __iters_per_work_item * __max_cu;
     }
 
-    template <typename _ExecutionPolicy, typename _Fp, typename _Index>
+    template <typename _ExecutionPolicy, typename _Fp, typename _Index, typename... _Ranges>
     auto
     operator()(_ExecutionPolicy&& __exec, _Fp __brick, _Index __count, _Ranges&&... __rngs) const
     {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -214,6 +214,12 @@ class __scan_single_wg_dynamic_kernel;
 template <typename... Name>
 class __scan_copy_single_wg_kernel;
 
+template <typename... Name>
+class __parallel_for_small_kernel;
+
+template <typename... Name>
+class __parallel_for_large_kernel;
+
 //------------------------------------------------------------------------
 // parallel_for - async pattern
 //------------------------------------------------------------------------
@@ -222,10 +228,35 @@ class __scan_copy_single_wg_kernel;
 // as the parameter pack that can be empty (for unnamed kernels) or contain exactly one
 // type (for explicitly specified name by the user)
 template <typename _KernelName>
-struct __parallel_for_submitter;
+struct __parallel_for_small_submitter;
 
 template <typename... _Name>
-struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
+struct __parallel_for_small_submitter<__internal::__optional_kernel_name<_Name...>>
+{
+    template <typename _ExecutionPolicy, typename _Fp, typename _Index, typename... _Ranges>
+    auto
+    operator()(_ExecutionPolicy&& __exec, _Fp __brick, _Index __count, _Ranges&&... __rngs) const
+    {
+        assert(oneapi::dpl::__ranges::__get_first_range_size(__rngs...) > 0);
+        _PRINT_INFO_IN_DEBUG_MODE(__exec);
+        auto __event = __exec.queue().submit([&__rngs..., &__brick, __count](sycl::handler& __cgh) {
+            //get an access to data under SYCL buffer:
+            oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
+
+            __cgh.parallel_for<_Name...>(sycl::range</*dim=*/1>(__count), [=](sycl::item</*dim=*/1> __item_id) {
+                auto __idx = __item_id.get_linear_id();
+                __brick(__idx, __rngs...);
+            });
+        });
+        return __future(__event);
+    }
+};
+
+template <typename _KernelName>
+struct __parallel_for_large_submitter;
+
+template <typename... _Name>
+struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name...>>
 {
     template <typename _ExecutionPolicy, typename _Fp, typename _Index, typename... _Ranges>
     auto
@@ -244,14 +275,13 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
             using _ValueType =
                 oneapi::dpl::__internal::__value_t<std::decay_t<std::tuple_element_t<0, std::tuple<_Ranges...>>>>;
 
-            // Process up to 16 bytes per work-item. This results in 512 bytes loaded input range per size 32 sub-group which
-            // has yielded best performance on target architectures. For larger data types, load a single element.
+            // Process up to 16 bytes per work-item per input range. This value has been the empirically determined minimum
+            // number of bytes for a single input range to saturate HW bandwidth on target architecures.
             constexpr std::uint8_t __bytes_per_work_item = 16;
-            constexpr std::uint8_t __max_iters_per_work_item = oneapi::dpl::__internal::__dpl_ceiling_div(__bytes_per_work_item, sizeof(_ValueType));
-            const std::uint32_t __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
-            const std::size_t __iters_per_compute_unit = oneapi::dpl::__internal::__dpl_ceiling_div(__count, __max_cu * __work_group_size);
-            // For small data sizes, distribute the work evenly among compute units.
-            const std::uint8_t __iters_per_work_item = std::min(__iters_per_compute_unit, static_cast<std::size_t>(__max_iters_per_work_item));
+            // TODO: Better handle this heuristic for the case where the input is a zip iterator
+            constexpr std::uint8_t __iters_per_work_item =
+                oneapi::dpl::__internal::__dpl_ceiling_div(__bytes_per_work_item, sizeof(_ValueType));
+
             const std::size_t __num_groups =
                          oneapi::dpl::__internal::__dpl_ceiling_div(__count, (__work_group_size * __iters_per_work_item));
             const std::size_t __num_items = __num_groups * __work_group_size;
@@ -264,6 +294,7 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
                     // performance regressions for out-of-place (e.g. std::copy).
                     if (__is_full)
                     {
+                        _ONEDPL_PRAGMA_UNROLL
                         for (std::uint8_t __i = 0; __i < __iters_per_work_item; ++__i)
                         {
                             __brick(__idx, __rngs...);
@@ -295,10 +326,23 @@ __parallel_for(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&&
                _Ranges&&... __rngs)
 {
     using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
-    using _ForKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<_CustomName>;
+    using _ForKernelSmall =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__parallel_for_small_kernel<_CustomName>>;
+    using _ForKernelLarge =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__parallel_for_large_kernel<_CustomName>>;
 
-    return __parallel_for_submitter<_ForKernel>()(::std::forward<_ExecutionPolicy>(__exec), __brick, __count,
-                                                  ::std::forward<_Ranges>(__rngs)...);
+    // Compile two kernels: one for small-to-medium inputs and a second for large. This avoids runtime checks within a single
+    // kernel that worsen performance for small cases.
+    if (__count <= 262144)
+    {
+        return __parallel_for_small_submitter<_ForKernelSmall>()(std::forward<_ExecutionPolicy>(__exec), __brick, __count,
+                                                            std::forward<_Ranges>(__rngs)...);
+    }
+    else
+    {
+        return __parallel_for_large_submitter<_ForKernelLarge>()(std::forward<_ExecutionPolicy>(__exec), __brick, __count,
+                                                                 std::forward<_Ranges>(__rngs)...);
+    }
 }
 
 //------------------------------------------------------------------------
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -238,22 +238,21 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
             std::size_t __work_group_size = oneapi::dpl::__internal::__max_work_group_size(__exec);
 
-            // For target architectures, 512 bytes is the maximum amount of data that can be performed in a single load / store
-            // transaction. Assuming a sub-group size of 32, 512 / 32 = 16 which is the number of bytes we wish to load / store
-            // per work-item. For architectures that do not support load / stores of 512 bytes (e.g. 128 bytes), several smaller
-            // but coalesced transactions will be made and performance should still be maximized.
+            // Processing 512 bytes per sub-group has shown the best performance on target architectures.
             // Grab the value type of the first range to estimate the optimal iters per work item.
             using _ValueType =
                 oneapi::dpl::__internal::__value_t<std::decay_t<std::tuple_element_t<0, std::tuple<_Ranges...>>>>;
-            constexpr std::uint16_t __max_bytes_per_transaction = 512;
+
+            constexpr std::uint16_t __max_bytes_per_sub_group = 512;
             constexpr std::uint16_t __predicted_sub_group_size = 32;
-            constexpr std::uint16_t __bytes_per_work_item = __max_bytes_per_transaction / __predicted_sub_group_size;
+            constexpr std::uint16_t __bytes_per_work_item = __max_bytes_per_sub_group / __predicted_sub_group_size;
             // If the _ValueType > 128 bytes (unlikely), then perform a single iteration per work item.
             constexpr std::uint16_t __iters_per_work_item =
                 std::max(std::size_t{1}, __bytes_per_work_item / sizeof(_ValueType));
-            std::size_t __num_items =
-                std::max(static_cast<_Index>(__work_group_size),
-                         oneapi::dpl::__internal::__dpl_ceiling_div(__count, __iters_per_work_item));
+            std::size_t __num_groups =
+                std::max(__work_group_size,
+                         oneapi::dpl::__internal::__dpl_ceiling_div(__count, (__work_group_size * __iters_per_work_item)));
+            std::size_t __num_items = __num_groups * __work_group_size;
             // TODO: optimize for small data sizes that do not saturate the device with this scheme
             __cgh.parallel_for<_Name...>(
                 sycl::nd_range(sycl::range<1>(__num_items), sycl::range<1>(__work_group_size)),
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -302,14 +302,14 @@ struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name..
             const std::size_t __num_items = __num_groups * __work_group_size;
             __cgh.parallel_for<_Name...>(
                 sycl::nd_range(sycl::range<1>(__num_items), sycl::range<1>(__work_group_size)),
-                [=](sycl::nd_item</*dim=*/1> __ndi) {
+                [=](sycl::nd_item</*dim=*/1> __item) {
                     // TODO: Investigate adding a vectorized path similar to reduce.
                     // Initial investigation showed benefits for in-place for-based algorithms (e.g. std::for_each) but
                     // performance regressions for out-of-place (e.g. std::copy) where the compiler was unable to
                     // vectorize our code. Vectorization may also improve performance of for-algorithms over small data
                     // types.
                     auto [__idx, __group_start_idx, __stride, __is_full] =
-                        __stride_recommender(__ndi, __count, __iters_per_work_item, __work_group_size);
+                        __stride_recommender(__item, __count, __iters_per_work_item, __work_group_size);
                     if (__is_full)
                     {
                         _ONEDPL_PRAGMA_UNROLL
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -252,6 +252,7 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
             const std::size_t __iters_per_compute_unit = oneapi::dpl::__internal::__dpl_ceiling_div(__count, __max_cu * __work_group_size);
             // For small data sizes, distribute the work evenly among compute units.
             const std::uint8_t __iters_per_work_item = std::min(__iters_per_compute_unit, static_cast<std::size_t>(__max_iters_per_work_item));
+            const bool __can_unroll_loop = __max_iters_per_work_item == __iters_per_work_item;
             const std::size_t __num_groups =
                          oneapi::dpl::__internal::__dpl_ceiling_div(__count, (__work_group_size * __iters_per_work_item));
             const std::size_t __num_items = __num_groups * __work_group_size;
@@ -262,7 +263,16 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
                     // TODO: Investigate using a vectorized approach similar to reduce.
                     // Initial investigation showed benefits for in-place for-based algorithms (e.g. std::for_each) but
                     // performance regressions for out-of-place (e.g. std::copy).
-                    if (__is_full)
+                    if (__is_full && __can_unroll_loop)
+                    {
+                        _ONEDPL_PRAGMA_UNROLL
+                        for (std::uint8_t __i = 0; __i < __max_iters_per_work_item; ++__i)
+                        {
+                            __brick(__idx, __rngs...);
+                            __idx += __stride;
+                        }
+                    }
+                    else if (__is_full)
                     {
                         for (std::uint8_t __i = 0; __i < __iters_per_work_item; ++__i)
                         {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -308,7 +308,7 @@ struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name..
                     // performance regressions for out-of-place (e.g. std::copy) where the compiler was unable to
                     // vectorize our code. Vectorization may also improve performance of for-algorithms over small data
                     // types.
-                    auto [__idx, __group_start_idx, __stride, __is_full] =
+                    auto [__idx, __stride, __is_full] =
                         __stride_recommender(__item, __count, __iters_per_work_item, __work_group_size);
                     if (__is_full)
                     {
@@ -319,21 +319,15 @@ struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name..
                             __idx += __stride;
                         }
                     }
-                    else
+                    // If we are not full, then take this branch only if there is work to process.
+                    else if (__idx < __count)
                     {
-                        // Recompute iters per item and manually unroll last loop iteration to remove most branching.
-                        if (__group_start_idx >= __count)
-                            return;
                         const std::uint8_t __adjusted_iters_per_work_item =
-                            oneapi::dpl::__internal::__dpl_ceiling_div(__count - __group_start_idx, __stride);
-                        for (std::uint8_t __i = 0; __i < __adjusted_iters_per_work_item - 1; ++__i)
+                            oneapi::dpl::__internal::__dpl_ceiling_div(__count - __idx, __stride);
+                        for (std::uint8_t __i = 0; __i < __adjusted_iters_per_work_item; ++__i)
                         {
                             __brick(__idx, __rngs...);
                             __idx += __stride;
-                        }
-                        if (__idx < __count)
-                        {
-                            __brick(__idx, __rngs...);
                         }
                     }
                 });
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -241,20 +241,20 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
             std::size_t __work_group_size = oneapi::dpl::__internal::__max_work_group_size(__exec, 512);
             __work_group_size = std::min(__work_group_size, static_cast<std::size_t>(__count));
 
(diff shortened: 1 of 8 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +252,96 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+            __cgh.parallel_for<_Name...>(
+                sycl::nd_range(sycl::range<1>(__num_items), sycl::range<1>(__work_group_size)),
+                [=](sycl::nd_item</*dim=*/1> __ndi) {
+                    // TODO: Investigate adding a vectorized path similar to reduce.
+                    // Initial investigation showed benefits for in-place for-based algorithms (e.g. std::for_each) but
+                    // performance regressions for out-of-place (e.g. std::copy) where the compiler was unable to
+                    // vectorize our code. Vectorization may also improve performance of for-algorithms over small data
+                    // types.
+                    auto [__idx, __group_start_idx, __stride, __is_full] =
+                        __stride_recommender(__ndi, __count, __iters_per_work_item, __work_group_size);
```
SergeyKopienko, line 312: May be better to specify `__stride_recommender` as template param type with ability to change from caller side if required?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +252,96 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+                        for (std::uint8_t __i = 0; __i < __adjusted_iters_per_work_item - 1; ++__i)
+                        {
+                            __brick(__idx, __rngs...);
+                            __idx += __stride;
+                        }
+                        if (__idx < __count)
+                        {
+                            __brick(__idx, __rngs...);
+                        }
+                    }
```
SergeyKopienko, line 338:
```suggestion
					// Recompute iters per item and manually unroll last loop iteration to remove most branching.
					if (__group_start_idx < __count)
					{
                        const std::uint8_t __adjusted_iters_per_work_item =
                            oneapi::dpl::__internal::__dpl_ceiling_div(__count - __group_start_idx, __stride);
                        for (std::uint8_t __i = 0; __i < __adjusted_iters_per_work_item - 1; ++__i)
                        {
                            __brick(__idx, __rngs...);
                            __idx += __stride;
                        }
                        if (__idx < __count)
                        {
                            __brick(__idx, __rngs...);
                        }
                    }
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -254,17 +350,32 @@ __parallel_for(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&&
...
+    using _ForKernelSmall =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__parallel_for_small_kernel<_CustomName>>;
+    using _ForKernelLarge =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__parallel_for_large_kernel<_CustomName>>;
+
+    using __small_submitter = __parallel_for_small_submitter<_ForKernelSmall>;
+    using __large_submitter = __parallel_for_large_submitter<_ForKernelLarge, _Ranges...>;
+    // Compile two kernels: one for small-to-medium inputs and a second for large. This avoids runtime checks within a single
+    // kernel that worsen performance for small cases.
+    if (__count < __large_submitter::__estimate_best_start_size(__exec))
```
SergeyKopienko, line 362: I am not sure that `__large_submitter` able to calculate to best size in all cases and it doesn't depends on Kernel code and it's logic at all. So may be we should have ability to customize this condition check to?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -823,6 +823,40 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
...
 };
 
+// Utility to recommend a stride for the best-performing memory access pattern from empirical testing on different
+// devices. This utility can only be called from the device.
+//
+// SPIR-V compilation targets show best performance with a stride of the sub-group size.
+// Other compilation targets perform best with a work-group size stride.
+template <typename NdItem>
+std::tuple<std::size_t, std::size_t, std::size_t, bool>
+__stride_recommender(const NdItem& __ndi, std::size_t __count, std::size_t __iters_per_work_item,
```
SergeyKopienko, line 833: I propose to rename `__ndi` to `__nd_item` or `__item`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +252,96 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+                            __brick(__idx, __rngs...);
+                            __idx += __stride;
+                        }
+                    }
+                    else
+                    {
+                        // Recompute iters per item and manually unroll last loop iteration to remove most branching.
+                        if (__group_start_idx >= __count)
+                            return;
+                        const std::uint8_t __adjusted_iters_per_work_item =
```
SergeyKopienko, line 327: Why you are using exactly the `std::uint8_t` type here?
May be better to use `decltype(__count - __group_start_idx)` or `auto` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +252,96 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+                            oneapi::dpl::__internal::__dpl_ceiling_div(__count - __group_start_idx, __stride);
+                        for (std::uint8_t __i = 0; __i < __adjusted_iters_per_work_item - 1; ++__i)
```
SergeyKopienko, line 329: The same.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -254,17 +350,32 @@ __parallel_for(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&&
...
+        return __large_submitter()(std::forward<_ExecutionPolicy>(__exec), __brick, __count,
+                                   std::forward<_Ranges>(__rngs)...);
+    }
 }
 
 //------------------------------------------------------------------------
 // parallel_transform_scan - async pattern
 //------------------------------------------------------------------------
 
-// Please see the comment for __parallel_for_submitter for optional kernel name explanation
```
danhoeflinger, line 267: There are a few more instances of this same comment which needs renaming (or instead adding the explanation in place):
`esimd_radix_sort_submitters.h`
`parallel_backend_sycl_fpga.h`
`parallel_backend_sycl_merge_sort.h`
`parallel_backend_sycl_merge.h`
`parallel_backend_sycl_reduce.h`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I think that the intention is that this will always be for non-full subgroups, which means it is `<=__iters_per_work_item` (max 16) after the division.  It is a narrowing cast, but it should be a safe one if the logic otherwise is correct.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Its an interesting point.  I'm trying to think about how different workloads / algorithms would impact this number...

I think the decision is largely dependent on overhead vs memory bandwidth optimization.  Its possible that more computation would make this less important because we are less reliant on memory bandwidth.  However, depending on user provided call-ables in the library makes this very difficult to make good decisions, unless there are some APIs which we know are always very high computationally that use parallel_for internally (I don't know of any).  

Another aspect to consider, the larger the size of the minimum type size in the input ranges, the fewer iterations would be run by the large submitter.  At the limit, I imagine there is no advantage of the large submitter when we only would run a single iteration.  This is knowable at compile time, perhaps we should detect this case and always choose the small submitter when a single iteration would be used.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Yes, @danhoeflinger is correct. `__adjusted_iters_per_work_item <= __iters_per_work_item` when going through the non-full path. `std::uint8_t` is used for `__iters_per_work_item` which is why it is used here as well.

SergeyKopienko: OK

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +252,96 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+                            __idx += __stride;
+                        }
+                    }
+                    else
+                    {
+                        // Recompute iters per item and manually unroll last loop iteration to remove most branching.
+                        if (__group_start_idx >= __count)
+                            return;
+                        const std::uint8_t __adjusted_iters_per_work_item =
+                            oneapi::dpl::__internal::__dpl_ceiling_div(__count - __group_start_idx, __stride);
```
danhoeflinger, line 328: Is this better than just using 
`__adjusted_iters_per_work_item = oneapi::dpl::__internal::__dpl_ceiling_div(__count - __idx, __stride);`
directly and not having the extra separated last iteration?

It seems like the compiler may end up with the same thing (and same amount of work-item divergence). I'm probably missing something though. 

I think we may be able to even remove `__group_start_idx` and the early exit above, merely relying upon the loop bounds to short circuit work-items beyond the end of the list.

Of course if we get a benefit from this, lets keep it but I'd like to understand it better.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -783,6 +783,26 @@ union __lazy_ctor_storage
...
+
+template <typename _T, typename... _Ts>
+class __min_tuple_type<std::tuple<_T, _Ts...>>
+{
+    using __min_type_ts = typename __min_tuple_type<std::tuple<_Ts...>>::type;
+
+  public:
+    using type = std::conditional_t<(sizeof(_T) < sizeof(__min_type_ts)), _T, __min_type_ts>;
+};
+
```
danhoeflinger, line 805: Do you think we will ever want the `type` specifics from a utility like this, or are we always only interested in the `sizeof(type)`?

If its likely to be the latter, we could just make this directly have a public `value` equal to the size of the minimum type rather than a public `type`.

This would simplify the code which uses it as well.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Thanks, I have updated these.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: I have changed the name to `__item` here and in the submitter code as well.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -823,6 +823,40 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
     }
 };
 
+// Utility to recommend a stride for the best-performing memory access pattern from empirical testing on different
+// devices. This utility can only be called from the device.
+//
+// SPIR-V compilation targets show best performance with a stride of the sub-group size.
+// Other compilation targets perform best with a work-group size stride.
+template <typename _NdItem>
```
dmitriy-sobolev, line 831: Minor thing. I assume there is no need to pass `nd_item` as a template, the function since only `nd_item<1>` is expected.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +252,90 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+    // iterator.
+    using _FlattenedRangesTuple = typename oneapi::dpl::__internal::__flatten_std_or_internal_tuple<
+        std::tuple<oneapi::dpl::__internal::__value_t<_Ranges>...>>::type;
+    using _MinValueType = typename oneapi::dpl::__internal::__min_tuple_type<_FlattenedRangesTuple>::type;
+    // __iters_per_work_item is set to 1, 2, 4, 8, or 16 depending on the smallest type in the
+    // flattened ranges. This allows us to launch enough work per item to saturate the device's memory
+    // bandwidth. This heuristic errs on the side of launching more work per item than what is needed to
+    // achieve full bandwidth utilization. 16 bytes per range per work item has been found as a good
+    // value across the different for-based algorithms.
+    static constexpr std::uint8_t __bytes_per_work_item = 16;
```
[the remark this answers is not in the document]
mmichel11: Since the reduce discussions, I believe there have been some improvements to coalesce style accesses in the compiler. For some cases with parallel for, there is benefit for adding a vertically vectorized path (sequential processing within a work item) but getting the compiler to generate the correct instructions is tricky to do for all cases.

For 32-bit and 64-bit types, I have seen similar performance with `std::copy` and a handcrafted version which generates vectorized loads / stores. Interestingly, there seems to be small benefits for `std::for_each` for these types. For smaller types such as 16-bit, vectorization brings significant improvement as the compiler is able to combine memory transactions. The bulk of further performance improvements would come with these small types I believe.

The reason that I have not added it now is that it has been non-trivial to implement in a generic fashion with our `__parallel_for` pattern to accommodate the different bricks. I have seen spotty performance thus far when trying to do so. When doing just a vector load such as in reduce, the compiler seems to easily generate the correct instructions. For load / store such as in `std::copy` it has been difficult to enable vectorization with our bricks. If the compiler doesn't generate the correct vector instructions, the performance significantly degrades. I think it is possible to implement a version that does this, but requires additional work that should come in a subsequent round of improvements that build on top of these. For the most common cases, this PR closes most of the gap and puts us near the memory bandwidth roof.

include/oneapi/dpl/pstl/utils.h
mmichel11: We are probably not going to ever use this to just get the type, so I have added this change to instead give a `value` which is the size of the smallest type. Also added a variable template.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: A similar change has been committed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Answered in the other comment regarding `uint8_t`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: One thing to note is that the performance difference at this "estimated point" is very small between the two versions so computing an optimal point to switch for each algorithm does not bring us any significant benefit and would likely be overtuned for a specific case / architecture, so I am in favor of leaving it as it currently is. The improvements are really observed once we scale to large sizes.

With regards to @danhoeflinger's comment, I agree that a user providing a heavily compute intensive operator might minimize the observed benefit if we become compute bound. I do not think we have encountered such a case yet, but I do not think there is a high risk of performance loss in this scenario although there may be small differences around this estimated point.

Good point on the case where we load one element per item. I will look into adding this as a compile time decision.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
dmitriy-sobolev: Thanks for so detailed reply. I agree: no need to do it in this PR.
 If you think that would be useful and feasible, I suggest leaving a TODO comment.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: In my opinion, leaving it as is currently is the best although maybe I should rename the function since stride is a loaded term.

We want to enforce a good access pattern. Work-group strides to enable coalescing is a good general choice for devices, and sub-group strides are used in the oneAPI GPU optimization guide and show slightly better performance for SPIR-V compiled targets. I do not see a need to change this as we should always use the best performing stride. I could see future improvements modifying `__stride_recommender` itself, but I do not see a need to accept a templated functor at this point. Do you see a potential use case for this?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: Good point, added!

dmitriy-sobolev: Could you also make the function inline for ODR compliance?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: This makes sense to me.  Thanks for the reply.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: Yes, done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
mmichel11: There is a TODO comment in the kernel submitter about adding a vectorized path.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I have added a compile time decision to only compile the basic kernel if the iters per work item is 1 for the large kernel. I benchmarked with a 128-bit type and confirmed performance is the same.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -352,16 +352,25 @@ __parallel_for(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&&
 
     using __small_submitter = __parallel_for_small_submitter<_ForKernelSmall>;
     using __large_submitter = __parallel_for_large_submitter<_ForKernelLarge, _Ranges...>;
-    // Compile two kernels: one for small-to-medium inputs and a second for large. This avoids runtime checks within a single
-    // kernel that worsen performance for small cases.
-    if (__count < __large_submitter::__estimate_best_start_size(__exec))
+    // Compile two kernels: one for small-to-medium inputs and a second for large. This avoids runtime checks within a
+    // single kernel that worsen performance for small cases. If the number of iterations of the large submitter is 1,
+    // then only compile the basic kernel as the two versions are effectively the same.
+    if constexpr (__large_submitter::__iters_per_work_item > 1)
```
danhoeflinger, line 358: I think if you rearrange the large submitter to be the `if` rather than the `else`, you can enable the if statement only when the constexpr is true, and then you can consolidate the two copies of the small submitter invocation into a single one (just using lack of return in place of `else`).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -254,17 +345,41 @@ __parallel_for(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&&
...
+    using __small_submitter = __parallel_for_small_submitter<_ForKernelSmall>;
+    using __large_submitter = __parallel_for_large_submitter<_ForKernelLarge, _Ranges...>;
+    // Compile two kernels: one for small-to-medium inputs and a second for large. This avoids runtime checks within a
+    // single kernel that worsen performance for small cases. If the number of iterations of the large submitter is 1,
+    // then only compile the basic kernel as the two versions are effectively the same.
+    if constexpr (__large_submitter::__iters_per_work_item > 1)
+    {
+        if (__count < __large_submitter::__estimate_best_start_size(__exec))
+        {
+            return __small_submitter()(std::forward<_ExecutionPolicy>(__exec), __brick, __count,
```
SergeyKopienko, line 362: As proposed @akukanov in some another PR, lets use `{}` for constructor param (as far as I remember).
So this mean here, that probably better to rewrite this like: `__small_submitter{}(`

SergeyKopienko:
@mmichel11 I have take a look to the history of this branch, probably make sense to rebase your branch from the current main state or merge main branch into your PR: we have a lot of new commits in the main branch now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -823,6 +823,39 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
     }
 };
 
+// Utility to recommend a stride for the best-performing memory access pattern from empirical testing on different
+// devices. This utility can only be called from the device.
+//
+// SPIR-V compilation targets show best performance with a stride of the sub-group size.
+// Other compilation targets perform best with a work-group size stride.
+inline std::tuple<std::size_t, std::size_t, bool>
+__stride_recommender(const sycl::nd_item<1>& __item, std::size_t __count, std::size_t __iters_per_work_item,
```
SergeyKopienko, line 832: How about to move it into `__parallel_for_large_submitter` as member function?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Good point, done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: Agreed, I have moved it here. If it's useful elsewhere in the future, then it can be made into a general utility.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +252,124 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+    {
+        const std::size_t __work_group_size =
+            oneapi::dpl::__internal::__max_work_group_size(__exec, __max_work_group_size);
+        const std::uint32_t __max_cu = oneapi::dpl::__internal::__max_compute_units(__exec);
+        return __work_group_size * __iters_per_work_item * __max_cu;
+    }
+
+    template <typename _ExecutionPolicy, typename _Fp, typename _Index>
+    auto
+    operator()(_ExecutionPolicy&& __exec, _Fp __brick, _Index __count, _Ranges&&... __rngs) const
```
adamfidel, line 325: I believe since `_Ranges` is a template parameter of the struct and not the function, `_Ranges&&` here means we are only accepting r-value references instead of forwarding references.

I'm not sure we want to exclude the ability to pass l-value reference ranges, so we might want to consider adding another template to this function for the actual ranges so we can use forwarding references.

mmichel11: You are correct. I added this as a template parameter to the struct to estimate the best start size. I have renamed this to `_RangeTypes` and have added `_Ranges` as a template parameter to the function call operator, so they are forwarding references again.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: OK, let's save as is.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: This change has been added.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -784,6 +784,29 @@ union __lazy_ctor_storage
...
+class __min_tuple_type_size<std::tuple<_T, _Ts...>>
+{
+    static constexpr std::size_t __min_type_value_ts = __min_tuple_type_size<std::tuple<_Ts...>>::value;
+
+  public:
+    static constexpr std::size_t value = std::min(sizeof(_T), __min_type_value_ts);
+};
+
+template <typename _Tuple>
+inline constexpr std::size_t __min_tuple_type_size_v = __min_tuple_type_size<_Tuple>::value;
```
dmitriy-sobolev, line 808: This can be simplified:

```
template<typename Tuple>
struct __min_tuple_type_size;

template<typename... Ts>
struct __min_tuple_type_size<std::tuple<Ts...>> {
    static constexpr std::size_t value = std::min({sizeof(Ts)...});
};
```

`_v` alias is not-necessary as it is used only once.

include/oneapi/dpl/pstl/tuple_impl.h
```diff
@@ -793,6 +793,25 @@ struct __decay_with_tuple_specialization<::std::tuple<_Args...>>
 template <typename... _Args>
 using __decay_with_tuple_specialization_t = typename __decay_with_tuple_specialization<_Args...>::type;
 
+// Flatten nested std::tuple or oneapi::dpl::__internal::tuple types into a single std::tuple.
+template <typename _T>
+struct __flatten_std_or_internal_tuple
```
dmitriy-sobolev, line 798: Optional suggestion: `__flatten_std_or_internal_tuple` -> `__flatten_tuple`.

dmitriy-sobolev, line 798: I'd recommend moving `__flatten_std_or_internal_tuple` into `utils.h`. It is a niche utility not related to the core part of the class.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +252,124 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+    static constexpr std::uint16_t __max_work_group_size = 512;
+
+    // SPIR-V compilation targets show best performance with a stride of the sub-group size.
+    // Other compilation targets perform best with a work-group size stride. This utility can only be called from the
+    // device.
+    static inline std::tuple<std::size_t, std::size_t, bool>
+    __stride_recommender(const sycl::nd_item<1>& __item, std::size_t __count, std::size_t __iters_per_work_item,
+                         std::size_t __work_group_size)
+    {
+        if constexpr (oneapi::dpl::__internal::__is_spirv_target_v)
```
dmitriy-sobolev, line 285: Could you include `utils.h` where `__is_spirv_target_v` is defined?

dmitriy-sobolev (commented):
There are some things not directly related to the main part of the PR, which looks good to me.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +252,124 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+            const std::uint32_t __sub_group_id = __sub_group.get_group_linear_id();
+            const std::uint32_t __sub_group_local_id = __sub_group.get_local_linear_id();
+            const std::size_t __work_group_id = __item.get_group().get_group_linear_id();
+
+            const std::size_t __sub_group_start_idx =
+                __iters_per_work_item * (__work_group_id * __work_group_size + __sub_group_size * __sub_group_id);
+            const bool __is_full_sub_group =
+                __sub_group_start_idx + __iters_per_work_item * __sub_group_size <= __count;
+            const std::size_t __work_item_idx = __sub_group_start_idx + __sub_group_local_id;
+            return std::make_tuple(__work_item_idx, __sub_group_size, __is_full_sub_group);
```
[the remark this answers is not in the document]
mmichel11: Done

include/oneapi/dpl/pstl/utils.h
mmichel11: Thanks, I have replaced this `__min_tuple_type_size` with `__min_nested_type_size` which avoids having to flatten the tuple first and applied these ideas there.

include/oneapi/dpl/pstl/tuple_impl.h
mmichel11: This utility has been removed

mmichel11: This was originally done since `tuple_impl.h` includes `utils.h`, so we would have to forward declare our internal tuple otherwise to avoid a circular dependency.

The new utility supports on arbitrary type and doesn't require any specializations for our internal tuple, so it can be easily placed in `utils.h`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Done

mmichel11:
Thanks for the reviews everyone. I have addressed all current comments.

As discussed offline, we will use the current state of the PR as a starting point to introduce vectorized load / store paths where it is performant by rewriting our bricks for parallel for. Likely, I will have a second PR into this branch with these changes once they are complete.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -254,17 +376,33 @@ __parallel_for(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&&
...
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__parallel_for_small_kernel<_CustomName>>;
+    using _ForKernelLarge =
+        oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__parallel_for_large_kernel<_CustomName>>;
+
+    using __small_submitter = __parallel_for_small_submitter<_ForKernelSmall>;
+    using __large_submitter = __parallel_for_large_submitter<_ForKernelLarge, _Ranges...>;
+    // Compile two kernels: one for small-to-medium inputs and a second for large. This avoids runtime checks within a
+    // single kernel that worsen performance for small cases. If the number of iterations of the large submitter is 1,
+    // then only compile the basic kernel as the two versions are effectively the same.
+    if constexpr (__large_submitter::__iters_per_work_item > 1)
```
SergeyKopienko, line 389: Can we combine this `if constexpr` together with the next `if`?
Will we have some real profit from these two conditions?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +254,120 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+        return __work_group_size * __iters_per_work_item * __max_cu;
+    }
+
+    template <typename _ExecutionPolicy, typename _Fp, typename _Index, typename... _Ranges>
+    auto
+    operator()(_ExecutionPolicy&& __exec, _Fp __brick, _Index __count, _Ranges&&... __rngs) const
+    {
+        assert(oneapi::dpl::__ranges::__get_first_range_size(__rngs...) > 0);
+        _PRINT_INFO_IN_DEBUG_MODE(__exec);
+        auto __event = __exec.queue().submit([&__rngs..., &__brick, &__exec, __count](sycl::handler& __cgh) {
```
SergeyKopienko, line 327: I think we haven't any other cases where we capture `policy` into `submit` call.
May be better to eval
```C++
            const std::size_t __work_group_size =
                oneapi::dpl::__internal::__max_work_group_size(__exec, __max_work_group_size);
```
outside of `submit` and capture `__work_group_size` by value?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +254,120 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+            return std::make_tuple(__work_item_idx, __sub_group_size, __is_full_sub_group);
+        }
+        else
+        {
+            const std::size_t __work_group_start_idx =
+                __item.get_group().get_group_linear_id() * __work_group_size * __iters_per_work_item;
+            const std::size_t __work_item_idx = __work_group_start_idx + __item.get_local_linear_id();
+            const bool __is_full_work_group =
+                __work_group_start_idx + __iters_per_work_item * __work_group_size <= __count;
+            return std::make_tuple(__work_item_idx, __work_group_size, __is_full_work_group);
```
SergeyKopienko, line 305: The same.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +254,120 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+        auto __event = __exec.queue().submit([&__rngs..., &__brick, &__exec, __count](sycl::handler& __cgh) {
+            //get an access to data under SYCL buffer:
+            oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
+            const std::size_t __work_group_size =
+                oneapi::dpl::__internal::__max_work_group_size(__exec, __max_work_group_size);
+            const std::size_t __num_groups =
+                oneapi::dpl::__internal::__dpl_ceiling_div(__count, (__work_group_size * __iters_per_work_item));
+            const std::size_t __num_items = __num_groups * __work_group_size;
+            __cgh.parallel_for<_Name...>(
+                sycl::nd_range(sycl::range<1>(__num_items), sycl::range<1>(__work_group_size)),
```
SergeyKopienko, line 336:
```suggestion
            __cgh.parallel_for<_Name...>(
                sycl::nd_range(sycl::range<1>(__num_groups * __work_group_size), sycl::range<1>(__work_group_size)),
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +254,120 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+    // Other compilation targets perform best with a work-group size stride. This utility can only be called from the
+    // device.
+    static inline std::tuple<std::size_t, std::size_t, bool>
+    __stride_recommender(const sycl::nd_item<1>& __item, std::size_t __count, std::size_t __iters_per_work_item,
+                         std::size_t __work_group_size)
+    {
+        if constexpr (oneapi::dpl::__internal::__is_spirv_target_v)
+        {
+            const __dpl_sycl::__sub_group __sub_group = __item.get_sub_group();
+            const std::uint32_t __sub_group_size = __sub_group.get_local_linear_range();
```
SergeyKopienko, line 286: All these functions returns `std::size_t`. Could you please explain why you are using `std::uint32_t` instead?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -246,6 +254,120 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
...
+                            __brick(__idx, __rngs...);
+                            __idx += __stride;
+                        }
+                    }
+                    // If we are not full, then take this branch only if there is work to process.
+                    else if (__idx < __count)
+                    {
+                        const std::uint8_t __adjusted_iters_per_work_item =
+                            oneapi::dpl::__internal::__dpl_ceiling_div(__count - __idx, __stride);
+                        for (std::uint8_t __i = 0; __i < __adjusted_iters_per_work_item; ++__i)
```
SergeyKopienko, line 359: Should we use `_ONEDPL_PRAGMA_UNROLL` for this `for-loop` too?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: The signature for the sub-group member function is `uint32_t get_local_linear_range() const` and related functions also return `uint32_t`. Were you thinking about the `group` class maybe?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Not in this case (there is a similar case in the strided loop utility in the new PR). Because the loop end variable is computed at run-time, the loop cannot be unrolled.

This path is called for the last sub-group / work-group, so the performance impact is negligible.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Combining the two would make it a runtime if conditional. Even if `__iters_per_work_item` is known at compile-time and the compiler can optimize it out, there may still be a chance for the kernel to be unnecessarily compiled. I think it is best to keep the `if constexpr`, so we can be sure to avoid compiling the large submitter if possible.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Sure, applied to new PR (this one will be closed soon).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Agreed, it is in the new PR.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
mmichel11: Just `std::tuple` is now used in the new PR.

PR #1871, opened 2024-09-23
Fixing include guard names

SergeyKopienko:
Issue: https://github.com/oneapi-src/oneDPL/issues/391

In this PR we fixing include guard names.

### DIFF
--- a/include/oneapi/dpl/pstl/omp/parallel_merge.h
+++ b/include/oneapi/dpl/pstl/omp/parallel_merge.h
@@ -13,8 +13,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-#ifndef _ONEDPL_INTERNAL_OMP_PARALLEL_MERGE_H
-#define _ONEDPL_INTERNAL_OMP_PARALLEL_MERGE_H
+#ifndef _ONEDPL_OMP_PARALLEL_MERGE_H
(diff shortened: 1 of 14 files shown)


include/oneapi/dpl/pstl/omp/parallel_merge.h
```diff
@@ -13,8 +13,8 @@
 //
 //===----------------------------------------------------------------------===//
 
-#ifndef _ONEDPL_INTERNAL_OMP_PARALLEL_MERGE_H
-#define _ONEDPL_INTERNAL_OMP_PARALLEL_MERGE_H
+#ifndef _ONEDPL_PARALLEL_MERGE_H
```
MikeDvorskiy, line 16: I think it is potentially dangerous. "OMP" substring is omitted.
What else if "tomorrow" we will add another host backend (another folder) with similar filenames and guards?

SergeyKopienko: I propose to think about this situation tomorrow, if we really will have this case.
And now we may just follow to our existing guard names rule.

SergeyKopienko: UPD: `_OMP` restored.

PR #1873, opened 2024-09-23
[Test] Fix for ranges testing in multi translational units

danhoeflinger:
Bugfix for multi translational units test.
Ranges header can only be included in combination with certain STL versions, we need to guard the ranges header include.


dmitriy-sobolev (approved):
Thank you for the fix!

mmichel11 (approved):
LGTM, thanks!

PR #1874, opened 2024-09-24
Avoid extra loops on sorted data in the `__subgroup_bubble_sorter::sort`

SergeyKopienko:
In this PR we avoid extra loops on sorted data in the `__subgroup_bubble_sorter::sort`,
which calls in sort algorithms on the device for every work item.

This is beneficial when the data is already sorted, which according to https://en.wikipedia.org/wiki/Sorting_algorithm is a common real-world case.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
@@ -40,16 +40,20 @@ struct __subgroup_bubble_sorter
     void
     sort(const _StorageAcc& __storage_acc, _Compare __comp, std::uint32_t __start, std::uint32_t __end) const
     {
-        for (std::uint32_t i = __start; i < __end; ++i)
+        bool __changed = true;
+
+        for (std::uint32_t i = __start; i < __end && __changed; ++i)
         {
+            __changed = false;
+
             for (std::uint32_t j = __start + 1; j < __start + __end - i; ++j)
             {
                 auto& __first_item = __storage_acc[j - 1];
                 auto& __second_item = __storage_acc[j];
                 if (__comp(__second_item, __first_item))
                 {
-                    using std::swap;
-                    swap(__first_item, __second_item);
+                    std::swap(__first_item, __second_item);
+                    __changed = true;
                 }
             }
         }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
@@ -54,7 +54,10 @@ struct __subgroup_bubble_sorter
                 auto& __first_item = __storage_acc[__start];
                 auto& __second_item = __storage_acc[__start + 1];
                 if (__comp(__second_item, __first_item))
-                    std::swap(__first_item, __second_item);
+                {
+                    using std::swap;
+                    swap(__first_item, __second_item);
+                }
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -40,18 +40,39 @@ struct __subgroup_bubble_sorter
...
+
+        _IndexType __n = __end - __start;
+
+        switch (__n)
         {
-            for (std::uint32_t j = __start + 1; j < __start + __end - i; ++j)
+        case 0:
+        case 1:
+            break;
+        case 2:
```
SergeyKopienko, line 52: I propose to ignore clang format at this place.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -40,18 +40,39 @@ struct __subgroup_bubble_sorter
     void
     sort(const _StorageAcc& __storage_acc, _Compare __comp, std::uint32_t __start, std::uint32_t __end) const
     {
-        for (std::uint32_t i = __start; i < __end; ++i)
+        using _IndexType = std::make_signed_t<decltype(__end)>;
+
+        _IndexType __n = __end - __start;
+
+        switch (__n)
```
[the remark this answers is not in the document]
mmichel11: For large number of elements `switch` should perform better although I do not know in device code. However, with just 4 cases there's probably no difference.

In my opinion, unless there is a measurable performance difference between switch and if-else, then we should use if-else.

mmichel11, line 47: Could you add a small comment justifying these extra checks?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -40,18 +40,39 @@ struct __subgroup_bubble_sorter
     void
     sort(const _StorageAcc& __storage_acc, _Compare __comp, std::uint32_t __start, std::uint32_t __end) const
     {
-        for (std::uint32_t i = __start; i < __end; ++i)
+        using _IndexType = std::make_signed_t<decltype(__end)>;
+
+        _IndexType __n = __end - __start;
```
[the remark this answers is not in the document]
SergeyKopienko: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
SergeyKopienko: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -40,18 +40,41 @@ struct __subgroup_bubble_sorter
...
+                auto& __second_item = __storage_acc[__start + 1];
                 if (__comp(__second_item, __first_item))
                 {
                     using std::swap;
                     swap(__first_item, __second_item);
                 }
             }
+            break;
+        default:            // The case when __end > __start + 2    three or more source data items require full bubble sort
+            do
```
SergeyKopienko, line 62: Implementation from https://en.wikipedia.org/wiki/Bubble_sort

PR #1876, opened 2024-09-24
Use __dpl_sycl accessor in single pass scan

danhoeflinger:
Single pass scan neglects to use `__dpl_sycl::__local_accessor` which causes problems when used with older versions of the SYCL library.  This is causing some errors in the multiple translation unit tests when used with older compilers.

This PR fixes that issue.


dmitriy-sobolev (approved):

timmiesmith (approved):
LGTM. Thank you.

adamfidel (approved):
LGTM. Thanks!

PR #1773, opened 2024-08-09
Support scratch space without result space

julianmi:
This patch extends `__result_and_scratch_storage` to support cases where no result space is required. This removes the overhead of allocating result storage even when it is unneeded such as in the dummy container introduced in #1770. The signature is kept intact to support the different scan kernels.

There is also some code cleanup and the usage in reduce is aligned with the recently merged scan. The container is also passed by reference to the reduce kernels to prevent copies.

PR #1879, opened 2024-09-30
Enabling reduce_then_scan for "Set" family of scan APIs

danhoeflinger:
Enabling reduce then scan algorithm to be used for set family APIs:
`set_intersection`
`set_union`
`set_difference`
`set_symmetric_difference`

Adds high-level helpers required to convert set family to use reduce_then_scan on GPU with size 32 subgroups.

Also makes set consistent with the other scan algorithms to select algorithm in the `__parallel_*` level of function call.

This provides significant performance improvements, especially at small sizes of `n`.  At larger sizes of `n`, this is an improvement over the existing algorithm, but not by a lot.  

There is still significant room for improvement here, this algorithm is very inefficient.  

Three future things to try (from easiest to hardest) to improve the set family within the reduce then scan alg:
1) We could explore our alternative `__pstl_lower_bound`, `__shars_lower_bound` to see if it provides benefit.
2) We could propagate a "hint" minimum search index between blocks (this provides a lower bound and better "blocks" the second set, in addition to the blocking we do inherently on the first set) 
3) We could propagate a similar "hint minimum search index subgroup from one iteration to the next.   (This prevents redundant computation for consecutive iterations for a subgroup.  Since the sets are ordered, subsequent searches will never go backward as long as the element we are searching's index in the first set is greater than the previous iteration's search).

Right now, we are wasting lots of memory accesses and comparisons on sections of the second set which should be able to ruled out from knowledge we have access to at the time we are making the search (in the "Reduce" predicate).  We also lose the "blocking" cache advantage of reduce_then_scan, because the second set always searches from the start of the list.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1726,8 +1726,10 @@ __pattern_hetero_set_op(__hetero_tag<_BackendTag> __backend_tag, _ExecutionPolic
     auto __keep3 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _OutputIterator>();
     auto __buf3 = __keep3(__result, __result + __n1);
 
-    auto __result_size = __par_backend_hetero::__parallel_set_op(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(),
-                                         __buf2.all_view(), __buf3.all_view(), __comp, __is_op_difference).get();
+    auto __result_size = __par_backend_hetero::__parallel_set_op(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
+                                                                 __buf1.all_view(), __buf2.all_view(),
+                                                                 __buf3.all_view(), __comp, __is_op_difference)
+                             .get();
 
     return __result + __result_size;
 }
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1708,7 +1708,7 @@ __pattern_rotate_copy(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Bid
 template <typename _BackendTag, typename _ExecutionPolicy, typename _ForwardIterator1, typename _ForwardIterator2,
           typename _OutputIterator, typename _Compare, typename _IsOpDifference>
 _OutputIterator
-__pattern_hetero_set_op(__hetero_tag<_BackendTag> __backend_tag, _ExecutionPolicy&& __exec, _ForwardIterator1 __first1,
+__pattern_hetero_set_op(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _ForwardIterator1 __first1,
                         _ForwardIterator1 __last1, _ForwardIterator2 __first2, _ForwardIterator2 __last2,
                         _OutputIterator __result, _Compare __comp, _IsOpDifference __is_op_difference)
 {
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1726,7 +1726,7 @@ __pattern_hetero_set_op(__hetero_tag<_BackendTag> __backend_tag, _ExecutionPolic
     auto __keep3 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::write, _OutputIterator>();
     auto __buf3 = __keep3(__result, __result + __n1);
 
-    auto result_size = __parallel_set_op(__backend_tag, std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(),
+    auto __result_size = __par_backend_hetero::__parallel_set_op(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec), __buf1.all_view(),
                                          __buf2.all_view(), __buf3.all_view(), __comp, __is_op_difference).get();
 
     return __result + __result_size;
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1708,34 +1708,13 @@ __pattern_rotate_copy(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Bid
 template <typename _BackendTag, typename _ExecutionPolicy, typename _ForwardIterator1, typename _ForwardIterator2,
           typename _OutputIterator, typename _Compare, typename _IsOpDifference>
 _OutputIterator
-__pattern_hetero_set_op(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _ForwardIterator1 __first1,
+__pattern_hetero_set_op(__hetero_tag<_BackendTag> __backend_tag, _ExecutionPolicy&& __exec, _ForwardIterator1 __first1,
                         _ForwardIterator1 __last1, _ForwardIterator2 __first2, _ForwardIterator2 __last2,
-                        _OutputIterator __result, _Compare __comp, _IsOpDifference)
+                        _OutputIterator __result, _Compare __comp, _IsOpDifference __is_op_difference)
 {
     typedef typename ::std::iterator_traits<_ForwardIterator1>::difference_type _Size1;
-    typedef typename ::std::iterator_traits<_ForwardIterator2>::difference_type _Size2;
 
     const _Size1 __n1 = __last1 - __first1;
-    const _Size2 __n2 = __last2 - __first2;
-
-    //Algo is based on the recommended approach of set_intersection algo for GPU: binary search + scan (copying by mask).
-    using _ReduceOp = ::std::plus<_Size1>;
-    using _Assigner = unseq_backend::__scan_assigner;
-    using _NoAssign = unseq_backend::__scan_no_assign;
-    using _MaskAssigner = unseq_backend::__mask_assigner<2>;
-    using _InitType = unseq_backend::__no_init_value<_Size1>;
-    using _DataAcc = unseq_backend::walk_n<_ExecutionPolicy, oneapi::dpl::__internal::__no_op>;
-
-    _ReduceOp __reduce_op;
-    _Assigner __assign_op;
-    _DataAcc __get_data_op;
-    unseq_backend::__copy_by_mask<_ReduceOp, oneapi::dpl::__internal::__pstl_assign, /*inclusive*/ ::std::true_type, 2>
-        __copy_by_mask_op;
-    unseq_backend::__brick_set_op<_ExecutionPolicy, _Compare, _Size1, _Size2, _IsOpDifference> __create_mask_op{
-        __comp, __n1, __n2};
-
-    // temporary buffer to store boolean mask
-    oneapi::dpl::__par_backend_hetero::__buffer<_ExecutionPolicy, int32_t> __mask_buf(__exec, __n1);
 
     auto __keep1 =
         oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _ForwardIterator1>();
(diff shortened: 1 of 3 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -837,18 +901,19 @@ struct __gen_count_mask
...
 
-template <typename _GenMask>
+template <typename _GenMask, typename _RangeTransform = oneapi::dpl::__internal::__no_op>
 struct __gen_expand_count_mask
 {
     template <typename _InRng, typename _SizeType>
     auto
     operator()(_InRng&& __in_rng, _SizeType __id) const
     {
+        auto __transformed_input = _RangeTransform{}(__in_rng);
```
adamfidel, line 911: Should we forward `__in_rng` here?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -825,6 +825,70 @@ struct __gen_unique_mask
...
+        }
+        else
+        {
+            auto __val_b = __set_b[__res];
+
+            //Difference operation logic: if number of duplication in __set_a on left side from __id > total number of
+            //duplication in __set_b than a mask is 1
+
+            //Intersection operation logic: if number of duplication in __set_a on left side from __id <= total number of
+            //duplication in __set_b than a mask is 1
```
adamfidel, line 861:
```suggestion
            //duplication in __set_b then a mask is 1
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -825,6 +825,70 @@ struct __gen_unique_mask
...
+
+            const std::size_t __count_a_left =
+                __id - oneapi::dpl::__internal::__pstl_left_bound(__set_a, std::size_t{0}, __id, __val_a, __comp) + 1;
+
+            const std::size_t __count_b =
+                oneapi::dpl::__internal::__pstl_right_bound(__set_b, __res, __nb, __val_b, __comp) - __res + __res -
```
adamfidel, line 867: I don't understand the ` - __res + __res` portion of this calculation. Is it necessary? If so, can we add a comment explaining it?

adamfidel (commented):
First pass of comments. I think I generally understand the algorithm and the logic seems good to me.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: This is me propagating an benign error from the multi-pass scan implementation which inspired this one.  I think they can just be removed in both places, good catch.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: No, because we use / forward `__in_rng` again later in the mask generation.

adamfidel: Ah, I didn't see the subsequent usage.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -798,14 +798,14 @@ struct __simple_write_to_id
...
+template <typename _Predicate, typename _RangeTransform = oneapi::dpl::__internal::__no_op>
 struct __gen_mask
 {
     template <typename _InRng>
     bool
-    operator()(const _InRng& __in_rng, std::size_t __id) const
+    operator()(_InRng&& __in_rng, std::size_t __id) const
     {
-        return __pred(__in_rng[__id]);
+        return __pred((_RangeTransform{}(std::forward<_InRng>(__in_rng)))[__id]);
```
mmichel11, line 808: For consistency with our functors, how about we add a `_RangeTransform __rng_transform` member after `__pred`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11, line 911: Similar comment to above with regards to adding `_RangeTransform` as a member.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -825,6 +825,70 @@ struct __gen_unique_mask
...
+        auto __set_mask = std::get<2>(__in_rng.tuple()); // mask sequence
+
+        std::size_t __nb = __set_b.size();
+
+        auto __val_a = __set_a[__id];
+
+        auto __res = oneapi::dpl::__internal::__pstl_lower_bound(__set_b, std::size_t{0}, __nb, __val_a, __comp);
+
+        bool bres =
+            _IsOpDifference::value; //initialization in true in case of difference operation; false - intersection.
```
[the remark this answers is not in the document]
danhoeflinger: i reverted the `constexpr` part of this suggestion since it is not used as a `constexpr` other than in this initialization.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1198,6 +1263,113 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag __backend_tag,
...
+    using _WriteOp = oneapi::dpl::__par_backend_hetero::__write_to_id_if<0, oneapi::dpl::__internal::__pstl_assign>;
+    using _Size = oneapi::dpl::__internal::__difference_t<_Range3>;
+
+    using _GenReduceInput = oneapi::dpl::__par_backend_hetero::__gen_count_mask<_GenMaskReduce>;
+    using _ReduceOp = std::plus<_Size>;
+    using _GenScanInput = oneapi::dpl::__par_backend_hetero::__gen_expand_count_mask<
+        _GenMaskScan, oneapi::dpl::__par_backend_hetero::__extract_range_from_zip<0>>;
+    using _ScanInputTransform = oneapi::dpl::__par_backend_hetero::__get_zeroth_element;
+
+    oneapi::dpl::__par_backend_hetero::__buffer<_ExecutionPolicy, int32_t> __mask_buf(__exec, __rng1.size());
```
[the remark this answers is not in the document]
danhoeflinger: I have not tried this.  I believe the only place we do this currently is `__result_and_scratch_storage`.  I could try to see the impact we would stand to gain, but we would need to create some infrastructure around it similar to that which could support situations which don't support USM.  
I propose tagging this as an issue for the future, and make the scope wider for a general utility structure, there are many such temp storage buffers in oneDPL I believe which could benefit from such treatment.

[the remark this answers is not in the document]
danhoeflinger: done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1198,6 +1263,113 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag __backend_tag,
...
+
+    return __parallel_transform_reduce_then_scan(
+        __backend_tag, std::forward<_ExecutionPolicy>(__exec),
+        oneapi::dpl::__ranges::make_zip_view(
+            std::forward<_Range1>(__rng1), std::forward<_Range2>(__rng2),
+            oneapi::dpl::__ranges::all_view<int32_t, __par_backend_hetero::access_mode::read_write>(
```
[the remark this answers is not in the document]
danhoeflinger: done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed, thanks.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed, thanks.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
danhoeflinger: Actually, now that I'm looking to make the issue I'm remembering #1773 where `__result_and_scratch_storage` was already adapted to be able to provide temporary-only storage, so not much is to be done here. 
I still think it perhaps deserves to be a separate issue to convert this and other temporary storage buffers to use this struct, or perhaps a layer on top of it.
We had agreed to not spend much effort polishing this PR with the intention to improve the set algorithms further.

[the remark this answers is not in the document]
danhoeflinger: I've created this issue.  https://github.com/oneapi-src/oneDPL/issues/1906
It has some nice potential for a performance improvement across a number of APIs with not too much development effort.  It will require a good amount of correctness and benchmarking to check it though.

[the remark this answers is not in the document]
mmichel11: Thanks!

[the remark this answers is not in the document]
danhoeflinger: Thank you for the idea.

danhoeflinger:
I realize I never expanded the set of reviewers from a small group, though I know we agreed not to attempt to spend too much effort on perfecting this approach (for performance) before merging.  I've now added a few more for visibility.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1690,32 +1690,11 @@ template <typename _BackendTag, typename _ExecutionPolicy, typename _ForwardIter
 _OutputIterator
 __pattern_hetero_set_op(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _ForwardIterator1 __first1,
                         _ForwardIterator1 __last1, _ForwardIterator2 __first2, _ForwardIterator2 __last2,
-                        _OutputIterator __result, _Compare __comp, _IsOpDifference)
+                        _OutputIterator __result, _Compare __comp, _IsOpDifference __is_op_difference)
 {
-    typedef typename std::iterator_traits<_ForwardIterator1>::difference_type _Size1;
-    typedef typename std::iterator_traits<_ForwardIterator2>::difference_type _Size2;
+    typedef typename ::std::iterator_traits<_ForwardIterator1>::difference_type _Size1;
```
mmichel11, line 1695: ::std -> std

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -825,6 +826,70 @@ struct __gen_unique_mask
...
+            else
+                bres = __count_a_left <= __count_b; /*intersection*/
+        }
+        __set_mask[__id] = bres;
+        return bres;
+    }
+    _Compare __comp;
+};
+
+template <std::size_t I>
```
mmichel11, line 882: I guess we should name this `_I`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1197,6 +1264,112 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag __backend_tag,
...
+                    _IsOpDifference __is_op_difference)
+{
+    using _Size1 = oneapi::dpl::__internal::__difference_t<_Range1>;
+    using _Size2 = oneapi::dpl::__internal::__difference_t<_Range2>;
+
+    _Size1 __n1 = __rng1.size();
+    _Size2 __n2 = __rng2.size();
+
+    //Algo is based on the recommended approach of set_intersection algo for GPU: binary search + scan (copying by mask).
+    using _ReduceOp = ::std::plus<_Size1>;
```
mmichel11, line 1316: ::std -> std

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1197,6 +1264,112 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag __backend_tag,
...
+    using _Assigner = unseq_backend::__scan_assigner;
+    using _NoAssign = unseq_backend::__scan_no_assign;
+    using _MaskAssigner = unseq_backend::__mask_assigner<2>;
+    using _InitType = unseq_backend::__no_init_value<_Size1>;
+    using _DataAcc = unseq_backend::walk_n<_ExecutionPolicy, oneapi::dpl::__internal::__no_op>;
+
+    _ReduceOp __reduce_op;
+    _Assigner __assign_op;
+    _DataAcc __get_data_op;
+    unseq_backend::__copy_by_mask<_ReduceOp, oneapi::dpl::__internal::__pstl_assign, /*inclusive*/ ::std::true_type, 2>
```
mmichel11, line 1326: ::std -> std

mmichel11 (commented):
A few small styling nitpicks. I am ready to approve after.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: renamed to something meaningful

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed

mmichel11 (approved):
LGTM

PR #1880, opened 2024-10-04
Align with SYCL 2020: remove sycl::mode_tag_t

dmitriy-sobolev:
`sycl::mode_tag_t` is not a part of SYCL 2020. It was a part of the provisional specification, but it is absent in the newer revisions. 
There is a more radical approach: remove seemingly unused `access_mode` from `sycl_iterator` class entirely, but I am not sure about design / ABI implications.

Note, that the host_task tags (such as `read_write_host_task`) were not added, because they are not currently supported, and there is no sense to do so. The oneDPL specification suggest that the `begin` and `end` functions should be  provided during code submission to a device, which is not applicable within a host task context:

> The begin and end functions can take SYCL 2020 deduction tags and sycl::no_init as arguments to explicitly mention which access mode should be applied to the buffer accessor when submitting a SYCL kernel to a device.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
@@ -219,13 +219,11 @@ begin(sycl::buffer<T, /*dim=*/1, Allocator> buf, ModeTagT)
 }
 
 template <typename T, typename Allocator, typename ModeTagT>
-__internal::sycl_iterator<
-    __internal::__access_mode_resolver<ModeTagT, __dpl_sycl::__no_init>::__value, T, Allocator>
+__internal::sycl_iterator<__internal::__access_mode_resolver<ModeTagT, __dpl_sycl::__no_init>::__value, T, Allocator>
 begin(sycl::buffer<T, /*dim=*/1, Allocator> buf, ModeTagT, __dpl_sycl::__no_init)
 {
-    return __internal::sycl_iterator<
-        __internal::__access_mode_resolver<ModeTagT, __dpl_sycl::__no_init>::__value, T,
-        Allocator>{buf, 0};
+    return __internal::sycl_iterator<__internal::__access_mode_resolver<ModeTagT, __dpl_sycl::__no_init>::__value, T,
+                                     Allocator>{buf, 0};
 }
 
 template <typename T, typename Allocator>
@@ -245,13 +243,11 @@ end(sycl::buffer<T, /*dim=*/1, Allocator> buf, ModeTagT)
 }
 
 template <typename T, typename Allocator, typename ModeTagT>
-__internal::sycl_iterator<
-    __internal::__access_mode_resolver<ModeTagT, __dpl_sycl::__no_init>::__value, T, Allocator>
+__internal::sycl_iterator<__internal::__access_mode_resolver<ModeTagT, __dpl_sycl::__no_init>::__value, T, Allocator>
 end(sycl::buffer<T, /*dim=*/1, Allocator> buf, ModeTagT, __dpl_sycl::__no_init)
 {
-    return __internal::sycl_iterator<
-        __internal::__access_mode_resolver<ModeTagT, __dpl_sycl::__no_init>::__value, T,
-        Allocator>{buf, __dpl_sycl::__get_buffer_size(buf)};
+    return __internal::sycl_iterator<__internal::__access_mode_resolver<ModeTagT, __dpl_sycl::__no_init>::__value, T,
+                                     Allocator>{buf, __dpl_sycl::__get_buffer_size(buf)};
 }
 
 template <typename T, typename Allocator>
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
@@ -132,42 +132,32 @@ struct sycl_iterator
 };
 
 // map access_mode tag to access_mode value
-template <typename _ModTagT, typename _NoInitT = void>
+// TODO: consider removing the logic for discard_read_write and discard_write which are deprecated in SYCL 2020
+template <typename _ModeTagT, typename _NoInitT = void>
 struct __access_mode_resolver
 {
 };
 
-template <>
-struct __access_mode_resolver<std::decay_t<decltype(sycl::read_only)>, void>
+template <typename _NoInitT>
+struct __access_mode_resolver<std::decay_t<decltype(sycl::read_only)>, _NoInitT>
 {
     static constexpr access_mode __mode = access_mode::read;
 };
 
-template <>
-struct __access_mode_resolver<std::decay_t<decltype(sycl::write_only)>, void>
+template <typename _NoInitT>
+struct __access_mode_resolver<std::decay_t<decltype(sycl::write_only)>, _NoInitT>
 {
-    static constexpr access_mode __value = access_mode::write;
+    static constexpr access_mode __value = std::is_same_v<_NoInitT, void> ? access_mode::write :
+                                                                            access_mode::discard_write;
 };
 
-template <>
-struct __access_mode_resolver<std::decay_t<decltype(sycl::read_write)>, void>
+template <typename _NoInitT>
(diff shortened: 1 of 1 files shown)


akukanov:
Looks reasonable to me, but why not combine two distinct helpers ("converter" and "resolver") into a single one having two template parameters?

> There is a more radical approach: remove seemingly unused access_mode from sycl_iterator class entirely, but I am not sure about design / ABI implications.

If we are sure it is not only unused **but also unnecessary**, the first step would be to just set the parameter always to `read_write`. Since the type is unspecified, there should be nobody relying on the specific template arguments. Then we would need to change the specification to deprecate the extra parameters of `begin()/end()`, including `no_init`, declaring that they have no effect. Then at some point we could change the implementation.

However I am not sure about the tag being unnecessary. Support for C++20 concepts such as `indirectly_writable`, if we decide to try implementing it, might need this information.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -213,18 +234,20 @@ __internal::sycl_iterator<access_mode::discard_read_write, T, Allocator>
...
+    return __internal::sycl_iterator<__internal::__access_mode_resolver<ModeTagT>::__value, T, Allocator>{
+        buf, __dpl_sycl::__get_buffer_size(buf)};
 }
 
-template <typename T, typename Allocator, access_mode Mode>
-__internal::sycl_iterator<__internal::_ModeConverter<Mode>::__value, T, Allocator>
-    end(sycl::buffer<T, /*dim=*/1, Allocator> buf, sycl::mode_tag_t<Mode>, __dpl_sycl::__no_init)
+template <typename T, typename Allocator, typename ModeTagT>
+__internal::sycl_iterator<__internal::__access_mode_resolver<ModeTagT, __dpl_sycl::__no_init>::__value, T, Allocator>
+end(sycl::buffer<T, /*dim=*/1, Allocator> buf, ModeTagT, __dpl_sycl::__no_init)
```
danhoeflinger, line 247: Am I correct in understanding that this should not break any existing code which would pass in `sycl::mode_tag_t<Mode>` to specify an access mode (without host_task stuff)?

Is it just that the sycl spec leaves the types of deduction tags unspecified and we should not use the internals in our code either?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -131,25 +131,44 @@ struct sycl_iterator
...
 template <>
-struct _ModeConverter<access_mode::read_write>
+struct __access_mode_resolver<std::decay_t<decltype(sycl::read_only)>, void>
 {
-    static constexpr access_mode __value = access_mode::discard_read_write;
+    static constexpr access_mode __mode = access_mode::read;
+};
+
+template <>
+struct __access_mode_resolver<std::decay_t<decltype(sycl::write_only)>, void>
```
danhoeflinger, line 147: Should this `std::decay_t` be here?

If `decltype(sycl::write_only)` is `const`, won't this no longer match with someone passing in `sycl::write_only{}` to the public api?  

I might be wrong, but either I think we should either not use decay here, or probably even better, we should use decay also below when using `__access_mode_resolver`, or in some implementation layer automatically on the incoming type.

dmitriy-sobolev: I spent quite a lot of time to get to `std::decay_t`, originally having decay-less specializations. Here is an example to play with, where the types do not match: https://godbolt.org/z/58Edjs4fr. You can uncomment lines 40-41 and check the error message. This puzzled me, and you may get surprised too. 

> If decltype(sycl::write_only) is const, won't this no longer match with someone passing in sycl::write_only{} to the public api?

I guess this is the exact problem. These tags are `constexpr`, which assumes `const`. 

> we should either not use decay here, or probably even better, we should use decay also below

I will do so. I am not sure if decay is needed when we call `__access_mode_resolver`, as the specialization without `const` is more permissive, but it will not hurt.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
dmitriy-sobolev: > Am I correct in understanding that this should not break any existing code which would pass in sycl::mode_tag_t<Mode> to specify an access mode (without host_task stuff)?

You are right, it allows passing `sycl::mode_tag_t` as any other implementation-defined deduction tag type.

> Is it just that the sycl spec leaves the types of deduction tags unspecified and we should not use the internals in our code either?

This is also correct. The types of the [SYCL 2020 deduction tags](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_deduction_tags) are unspecified, and we do not want to rely on the internals of the specific implementations.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
danhoeflinger: You are right, I am surprised, I need to dig into this better because there is something I must not understand.  I think it is about the types in the template function instantiation.

danhoeflinger: Ah, ok I understand now, I believe it is simply that it loses its `const` qualifier when it instantiates the template function.  
Passing a `const` qualified argument by value into a template function does not force that function to treat that variable as a `const` type.  It is the responsibility of the writer of the template function to figure out if that is a const param or not.  

I think your implementation is fine, but it may prevent any future confusion to decay on the way in to `__access_mode_resolver`, or alternatively, mark `ModeTagT` const and name the parameter `tag`, then take `decltype(tag)` on the way in the `__access_mode_resolver`.   

Really any of these options works for me.  Thanks for indulging me in my confusion.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -131,25 +131,44 @@ struct sycl_iterator
...
-struct _ModeConverter<access_mode::write>
+struct __access_mode_resolver<std::decay_t<decltype(sycl::write_only)>, __dpl_sycl::__no_init>
 {
     static constexpr access_mode __value = access_mode::discard_write;
 };
 
+template <>
+struct __access_mode_resolver<std::decay_t<decltype(sycl::read_write)>, __dpl_sycl::__no_init>
+{
+    static constexpr access_mode __value = access_mode::discard_read_write;
```
adamfidel, line 169: I suppose it's a matter of preference, but would it make sense to combine the init/no_init overloads? Something like:
```c++
template <typename _NoInitT>
struct __access_mode_resolver<std::decay_t<decltype(sycl::read_write)>, _NoInitT>
{
    static constexpr access_mode __mode = std::is_same_v<_NoInitT, void> ? access_mode::read_write : access_mode::discard_read_write;
};
```

danhoeflinger (approved):
LGTM, happy to reapprove for decay changes if you want to make them.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
dmitriy-sobolev: Thanks, I've applied it.

dmitriy-sobolev:
> LGTM, happy to reapprove for decay changes if you want to make them.

Thank you. I would prefer leaving it as is with `decay` provided only in the specializations. It looks simpler to me compared to the alternatives.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
danhoeflinger: Doesn't this kill the compile-time part of this?  
I thought the conditional operator `?` was not usable at compile time in a `constexpr`.

danhoeflinger: Never mind, I think I have a case of Friday afternoon brain.  

I checked it out on godbolt and it seems to be fine in this context, as well as being used like this plenty of other places in the codebases.  I need to remind myself the details of `?` and `constexpr` because I know there is some nuance, but I'll withdraw my objection.

danhoeflinger (approved):

PR #1881, opened 2024-10-04
Make SYCL backend detection more portable

dmitriy-sobolev:
SYCL specification allows SYCL to be implemented purely as a library: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_library_only_implementation. It means that inclusion of `sycl/sycl.hpp` is a portable way to get `SYCL_LANGUAGE_VERSION`, in contrast to expecting it to be defined by the compiler itself. Additionally, AdaptiveCPP defines this macro in `sycl/sycl.hpp`.


danhoeflinger (commented):
I have a memory of us running into issues with some compilers providing `sycl.hpp`, but not actually providing a working SYCL implementation when we were adding this to CMake detection of the availability of SYCL.  Wouldn't this have the potential for some problems in that case?

My memory is that this was with clang++, but I may be mistaken.  Let me see if I can pull up that discussion.

danhoeflinger:
> I have a memory of us running into issues with some compilers providing `sycl.hpp`, but not actually providing a working SYCL implementation when we were adding this to CMake detection of the availability of SYCL. Wouldn't this have the potential for some problems in that case?
> 
> My memory is that this was with clang++, but I may be mistaken. Let me see if I can pull up that discussion.

It turns out its the opposite issue in reality https://github.com/oneapi-src/oneDPL/pull/959#issue-1720293495
`fsycl` option was available but headers not available.

akukanov:
Let's do some analysis first.

`__ONEDPL_SYCL_AVAILABLE` expectedly says whether it is possible to use SYCL code. It is then used to
a) check SYCL availability when it is requested via `ONEDPL_USE_DPCPP_BACKEND`
b) enable SYCL backend implicitly when not requested

There are two SYCL availability tests we use:
1) check if `<sycl/sycl.hpp>` is accessible
2) check if `SYCL_LANGUAGE_VERSION` is defined.

According to the SYCL standard, eventually both checks need to pass.
(1) is a mandatory condition prior for any use of SYCL, while (2) might depend on the implementation, as Dmitriy points out.
For some compilers, including DPC++, (2) is independently enabled by a compiler option.
For other compilers or library only implementations, (2) might never be true until the header is included. However even after the header is included, (2) is not always true I think, and so it still needs to be checked.

So the simplest generic SYCL availability test would be:
1)   check if `<sycl/sycl.hpp>` is accessible
1.5) include `<sycl/sycl.hpp>`
2)   check if `SYCL_LANGUAGE_VERSION` is defined

However, we do not want to include the SYCL header until it's needed. Also, if the header code depends on compiler support,
including it without prior checking (2) might in theory cause errors. So an intermediate state "SYCL is possibly available" - which essentially means we observe the header but not the language macro - seems useful.

Also note that the use cases (a) and (b) would treat "possibly available" differently:
- if the use of SYCL is explicitly requested, i.e. `ONEDPL_USE_DPCPP_BACKEND` is set to a non-zero value,
  we may assume that the environment is ready for that, and including the header seems reasonable.
- if there was no direct request for SYCL, we need to be cautious and should not include the header
  unless we are sure it will not cause errors.

The general procedure could be:
1) check if `<sycl/sycl.hpp>` is accessible;
   - if no, SYCL is **definitely not available**.
2) check if `SYCL_LANGUAGE_VERSION` is defined;
   - if yes, SYCL is **definitely available**.
   - if no, but we know it is required (as for Intel Compiler), SYCL is **definitely not available**.
   - otherwise, SYCL is **possibly available**
3) Case (a): if the use of SYCL is requested but SYCL is definitely not available, issue an error.
4) If the use of SYCL is requested and SYCL is [possibly] available (i.e. not definitely unavailable), enable the SYCL backend.
6) Case (b): If the use of SYCL is not requested but we are sure including the header is safe, enable the SYCL backend.
   - Currently, we are only sure when SYCL is definitely available, but we can expand it later for "known good" environments.

With this procedure, the SYCL backend might be enabled without requiring `SYCL_LANGUAGE_VERSION` a priori,
and therefore the backend should check `SYCL_LANGUAGE_VERSION` after including the header:

7) include `<sycl/sycl.hpp>`
8) if `SYCL_LANGUAGE_VERSION` is still not defined:
   - complete the case (a): if SYCL was requested, issue an error.
   - disable the SYCL backend and skip the rest of it.
   - alternatively, if disabling the SYCL backend from inside does not seem feasible, just issue an error.

dmitriy-sobolev:
Thank you for such a detail analysis, Alexey. I've implemented this suggestion.

CMakeLists.txt
```diff
@@ -301,6 +301,7 @@ if (ONEDPL_BACKEND MATCHES "^(tbb|dpcpp|dpcpp_only)$")
         target_compile_definitions(oneDPL INTERFACE
             $<$<OR:$<BOOL:${ONEDPL_USE_DEVICE_FPGA_HW}>,$<BOOL:${ONEDPL_USE_DEVICE_FPGA_EMU}>>:ONEDPL_FPGA_DEVICE>
             $<$<BOOL:${ONEDPL_USE_DEVICE_FPGA_EMU}>:ONEDPL_FPGA_EMULATOR>
+            ONEDPL_USE_DPCPP_BACKEND=1
```
akukanov, line 304: Is this necessary?
Or rather - with this change, will our tests still cover the case of "ONEDPL_USE_DPCPP_BACKEND is not defined"?

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -54,14 +54,34 @@
...
+#endif
+#if defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
+#    define _ONEDPL_SYCL_LANGUAGE_VERSION_PRESENT 1
+#endif
+#if _ONEDPL_SYCL_HEADER_PRESENT && _ONEDPL_SYCL_LANGUAGE_VERSION_PRESENT
 #    define _ONEDPL_SYCL_AVAILABLE 1
 #endif
-#if ONEDPL_USE_DPCPP_BACKEND && !_ONEDPL_SYCL_AVAILABLE
+// Intel(R) oneAPI DPC++/C++ Compiler provides SYCL_LANGUAGE_VERSION without sycl.hpp inclusion
+#if _ONEDPL_SYCL_HEADER_PRESENT && !_ONEDPL_SYCL_LANGUAGE_VERSION_PRESENT && !defined(__INTEL_LLVM_COMPILER)
```
akukanov, line 68: Will this cover the open-source intel/llvm implementation as well?

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -54,14 +54,34 @@
...
+#    define _ONEDPL_SYCL_POSSIBLY_AVAILABLE 1
+#endif
```
akukanov, line 70: The code is correct, however that's how I would do it:
```
#if _ONEDPL_SYCL_HEADER_PRESENT
#  if _ONEDPL_SYCL_LANGUAGE_VERSION_PRESENT
#    define _ONEDPL_SYCL_AVAILABLE 1
// Intel(R) oneAPI DPC++/C++ Compiler pre-defines SYCL_LANGUAGE_VERSION with -fsycl option
#  elif !defined(__INTEL_LLVM_COMPILER)
// Other compilers might define the macro in the SYCL header
#    define _ONEDPL_SYCL_POSSIBLY_AVAILABLE 1
#  endif
#endif
```

test/support/test_config.h
```diff
@@ -88,13 +88,28 @@
...
+// which enables DPCPP backend when SYCL is definitely available
+#if (__has_include(<sycl/sycl.hpp>) || __has_include(<CL/sycl.hpp>))
+#    define TEST_SYCL_HEADER_PRESENT 1
+#endif
+#if defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
+#    define TEST_SYCL_LANGUAGE_VERSION_PRESENT 1
+#endif
+#if TEST_SYCL_HEADER_PRESENT && TEST_SYCL_LANGUAGE_VERSION_PRESENT
+#    define TEST_SYCL_AVAILABLE 1
+#endif
```
akukanov, line 106:
```suggestion
#define TEST_SYCL_HEADER_PRESENT  (__has_include(<sycl/sycl.hpp>) || __has_include(<CL/sycl.hpp>))
#define TEST_SYCL_LANGUAGE_VERSION_PRESENT  (CL_SYCL_LANGUAGE_VERSION || SYCL_LANGUAGE_VERSION)
#define TEST_SYCL_AVAILABLE  (TEST_SYCL_HEADER_PRESENT && TEST_SYCL_LANGUAGE_VERSION_PRESENT)
```
I think such definitions should be OK because these macros are not supposed to be used outside of other preprocessor directives.

akukanov (commented):
I have some questions and suggestions.

Also, there is no check in the backend that the language version macro is properly defined after the header is included.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: It will. This implementation does not define `__INTEL_LLVM_COMPILER`

akukanov: So, the open source implementation enables SYCL in the same way as Intel(R) oneAPI DPC++/C++ Compiler (as far as I know), yet the logic in our header would work differently for them? Sounds strange to me.

CMakeLists.txt
dmitriy-sobolev: If `-ONEDPL_BACKEND=<dpcpp|dpcpp_only>` option is passed to cmake, it explicitly says that dpcpp backend should be used, what makes it logical to define `ONEDPL_USE_DPCPP_BACKEND` to a non-zero value. 

The case with the macro being undefined should remain covered as well. Let me think how to achieve it.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: Let me think how to distinguish the open-sourced version from genetic clang++ to align the logic. Note that https://github.com/intel/llvm appears to have the same set of predefined macros as clang++, so we should rely on some indirect traits.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: Applied

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: `#if __has_cpp_attribute(intel::kernel_args_restrict)` or any other intel-specific attribute looks like an option to distinguish the compiler. Although, it does not look conventional.

test/support/test_config.h
dmitriy-sobolev: Applied

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: I've added a check with `#if __has_cpp_attribute(intel::kernel_args_restrict)`, but I am leaning towards removing it, as it looks non-conventinal and possibly fragile.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -333,4 +340,47 @@
...
+
+#if _ONEDPL_BACKEND_SYCL
+// Include sycl specific options
+// FPGA doesn't support sub-groups
+#    if !(_ONEDPL_FPGA_DEVICE)
+#        define _USE_SUB_GROUPS 1
+#        define _USE_GROUP_ALGOS 1
+#    endif
+
+#    define _USE_RADIX_SORT (_USE_SUB_GROUPS && _USE_GROUP_ALGOS)
```
akukanov, line 374: Please add the `_ONEDPL` prefix.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -333,4 +340,47 @@
...
+
+// Compilation of a kernel is requiried to obtain valid work_group_size
+// when target devices are CPU or FPGA emulator. Since CPU and GPU devices
+// cannot be distinguished during compilation, the macro is enabled by default.
+#    if !defined(_ONEDPL_COMPILE_KERNEL)
+#        define _ONEDPL_COMPILE_KERNEL 1
+#    endif
```
akukanov, line 381: It would be good to mention in the comment when/where the macro can be turned off.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -333,4 +340,47 @@
...
+#    define _ONEDPL_FPGA_EMU ONEDPL_FPGA_EMULATOR
+#endif
+
+#if defined(ONEDPL_USE_PREDEFINED_POLICIES)
+#    undef _ONEDPL_PREDEFINED_POLICIES
+#    define _ONEDPL_PREDEFINED_POLICIES ONEDPL_USE_PREDEFINED_POLICIES
+#elif !defined(_ONEDPL_PREDEFINED_POLICIES)
+#    define _ONEDPL_PREDEFINED_POLICIES 1
+#endif
+
```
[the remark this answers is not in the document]
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -54,15 +38,71 @@
...
+// Proceed to inclusion only if it is safe or the backend was explicitly requested
+#if _ONEDPL_SYCL_POSSIBLY_AVAILABLE && (ONEDPL_USE_DPCPP_BACKEND || _ONEDPL_SAFE_TO_INCLUDE_SYCL)
+#    if __has_include(<sycl/sycl.hpp>)
+#        include <sycl/sycl.hpp>
+#    else
+#        include <CL/sycl.hpp>
+#    endif
+#    if defined(CL_SYCL_LANGUAGE_VERSION) || defined(SYCL_LANGUAGE_VERSION)
+#        define _ONEDPL_SYCL_AVAILABLE 1
+#    endif
```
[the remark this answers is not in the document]
dmitriy-sobolev: >  I would move the lines 80-82 into the backend header

There is no a single entry point into a heterogenous backend and its utilities, unfortunately. I've checked the usage of `_ONEDPL_HETERO_BACKEND` and `_ONEDPL_BACKEND_SYCL` macros in the headers and which public headers they depend on:

_ONEDPL_HETERO_BACKEND |  
-- | --
pstl\algorithm_impl.h | execution, algorithm
pstl\glue_algorithm_impl.h | execution, algorithm
pstl\glue_algorithm_ranges_impl.h | execution, algorithm, ranges
pstl\glue_memory_impl.h | execution, memory
pstl\glue_numeric_impl.h | execution, numeric, algorithm
pstl\glue_numeric_ranges_impl.h | execution, ranges
pstl\histogram_impl.h | execution, algorithm
  |  
_ONEDPL_BACKEND_SYCL |  
async | async
dynamic_selection | dynamic_selection
iterator | iterator
internal\binary_search_impl.h | execution, algorithm
internal\exclusive_scan_by_segment_impl.h | execution, algorithm
internal\function.h | execution, algorithm
internal\inclusive_scan_by_segment_impl.h | execution, algorithm
internal\iterator_impl.h | execution, iterator, algorithm
internal\reduce_by_segment_impl.h | execution, algorithm
internal\scan_by_segment_impl.h | execution, algorithm
pstl\execution_defs.h | execution, algorithm
pstl\glue_execution_defs.h | execution, algorithm
pstl\glue_numeric_defs.h | execution, numeric, algorithm
pstl\glue_numeric_impl.h | execution, numeric, algorithm
pstl\parallel_backend.h | execution
pstl\ranges_defs.h | execution, algorithm, ranges
pstl\utils.h | execution, algorithm, numeric, memory, ranges, kernel_templates
pstl\hetero\algorithm_impl_hetero.h | execution, algorithm, numeric, memory, ranges
pstl\hetero\algorithm_ranges_impl_hetero.h | execution, algorithm, ranges
pstl\hetero\histogram_impl_hetero.h | execution, algorithm
pstl\hetero\numeric_impl_hetero.h | execution, numeric, algorithm
pstl\hetero\numeric_ranges_impl_hetero.h | execution, ranges



Given that, I suggest moving `sycl.hpp` inclusion and the check of SYCL (probably, with the rest backend configuration) into a separate file, e.g. `hetero_backend_config.h` and include it in `execution`, `algorithm`, `numeric`, `memory`, `ranges`, `kernel_templates`, `async`, `dynamic_selection` after `onedpl_config.h`.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -54,15 +38,71 @@
...
+// -- Check availability of heterogeneous backends --
+
+// Detect both Intel(R) oneAPI DPC++/C++ Compiler and oneAPI DPC++ compiler
+// Rely on an extension attribute, which is present in both compilers
+// A predefined macro cannot be used since oneAPI DPC++/C++ Compiler provides the same set of macros as Clang
+#if __has_cpp_attribute(intel::kernel_args_restrict)
+#    define _ONEDPL_DPCPP_COMPILER 1
+#else
+#    define _ONEDPL_DPCPP_COMPILER 0
+#endif
```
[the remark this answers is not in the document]
dmitriy-sobolev: The comment is not relevant anymore. I've simplified the logic to implicitly enable DPCPP backend only when both sycl.hpp and `SYCL_LANGUAGE_VERSION` with no header inclusion.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: This question is not relevant right now: all the compilers are treated equally.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: I think that it deserves a separate investigation, because I do not know the exact root cause why `_ONEDPL_COMPILE_KERNEL` can be turned off, which is necessary to answer that question.

What I know is that we compile a kernel to reliably understand the maximum work-group size according to the resource allocation per work-item, and for some reason (it looks like due to empirical findings) it can be avoidable for GPUs.

Would you mind if a create an issue to address that problem?

dmitriy-sobolev: I think that it deserves a separate investigation, because I have no answer and it is not related directly to the PR.

What I know is:
 - We compile a kernel to reliably understand the maximum work-group size. This kernel-specific size may be smaller than the maximum supported work-group size by a device, e.g. due to using a large amount of registers or SLM.
 - For GPUs, it seems, we have found some empirical rules to avoiding out-of-resources issues. I do not know if they are applicable, e.g. to CPU, FPGA, or other GPU kernels. Even though, I do not expect that everything will work out-of-the-box for GPU if the macro is set to 0.
 - Explicit kernel compilation introduces overhead, impactful on small problem sizes. I do not know which sizes exactly, though.
 - Some implementations may not provide kernels (e.g. AdaptiveCPP), so it should be turned off there.

Would you mind if I create a separate issue to improve the comment?

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -333,4 +313,46 @@
...
+
+#if _ONEDPL_BACKEND_SYCL
+// Include sycl specific options
+// FPGA doesn't support sub-groups
+#    if !(_ONEDPL_FPGA_DEVICE)
+#        define _ONEDPL_USE_SUB_GROUPS 1
+#        define _ONEDPL_USE_GROUP_ALGOS 1
+#    endif
+
+#    define _USE_RADIX_SORT (_ONEDPL_USE_SUB_GROUPS && _ONEDPL_USE_GROUP_ALGOS)
```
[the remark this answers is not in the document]
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/onedpl_config.h
akukanov: I have traced it back to the very first snapshot of oneDPL source files, and even there the code was like today. I can guess that it was added as an internal knob to manually disable kernel compilation.

Yes, let's revisit it later, and not for improving the comment but for improving the definition or removing it altogether.

akukanov (dismissed):
It looks good to me.

danhoeflinger (approved):
LGTM, these are all good changes.  If you want to take the optional nitpicks, I will approve after

danhoeflinger (approved):
LGTM, the clang format suggestion is bad and should be ignored.

PR #1884, opened 2024-10-07
Fix includes in scan_2 test

dmitriy-sobolev:
scan algorithms are defined in `numeric` header instead of `algorithm`.


SergeyKopienko (approved):
LGTM

PR #1885, opened 2024-10-07
[oneDPL][ranges] + fix for std::ranges::sort using (windows); 

MikeDvorskiy:
1. Microsoft standard libraries older than _MSC_VER == 1939 have std::ranges::<algo> definition as function instead of a global function object. It leads to a error in oneDPL implementation. The PR fixes this issue via a workaround. https://godbolt.org/z/sTfcsj897
2.  Added _ONEDPL_STD_RANGES_WINDOWS_NON_GLOBAL_FUN_OBJ  macro

### DIFF
--- a/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
+++ b/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
@@ -423,7 +423,13 @@ struct __stable_sort_fn
     {
         const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
         return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(__dispatch_tag,
-            std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj, std::ranges::stable_sort);
+            std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
+#if _ONEDPL_STD_RANGES_FUN_OBJ_WINDOWS_BROKEN
+            [](auto&&... __args) { std::ranges::stable_sort(std::forward<decltype(__args)>(__args)...); }
+#else
+            std::ranges::stable_sort
+#endif
+            );
     }
 }; //__stable_sort_fn
 }  //__internal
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -326,4 +326,10 @@
 #    define _ONEDPL_STD_BIT_FLOOR_BROKEN 0
 #endif
 
+#if defined(_MSC_VER) && (_MSC_VER < 1939)
+#    define _ONEDPL_STD_RANGES_WINDOWS_NON_GLOBAL_FUN_OBJ 1
+#else
+#    define _ONEDPL_STD_RANGES_WINDOWS_NON_GLOBAL_FUN_OBJ 0
+#endif
```
dmitriy-sobolev, line 333: Could you comment the reason to add this macro? Its purpose is not obvious for a one who does not see the PR's description.

Also, this macro looks like a broken macro, because it is a workaround for an issue in the MSVC library. That is why I suggest renaming: `_ONEDPL_STD_RANGES_WINDOWS_NON_GLOBAL_FUN_OBJ` -> `_ONEDPL_STD_RANGES_FUN_OBJ_WINDOWS_BROKEN`.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -423,7 +423,13 @@ struct __stable_sort_fn
     {
         const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
         return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(__dispatch_tag,
-            std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj, std::ranges::stable_sort);
+            std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
+#if _ONEDPL_STD_RANGES_WINDOWS_NON_GLOBAL_FUN_OBJ
+            [](auto&&... args) { std::ranges::stable_sort(std::forward<decltype(args)>(args)...); }
```
dmitriy-sobolev, line 428: Do not forget to uglify `args`

include/oneapi/dpl/pstl/onedpl_config.h
MikeDvorskiy: dont mind

dmitriy-sobolev (approved):
LGTM. 
The clang-format suggestions look optional to me.

PR #1894, opened 2024-10-09
Fix inclusion order of numeric header

### LINKED ISSUE #1886: Error when using reduce
Hello, today I have noticed that project that have been compiling fine before, stopped working when I updated oneDPL to latest version. The error message I keep getting is as follows:
> In file included from /home/local/current/distributed-ranges/build/_deps/dpl-src/include/oneapi/dpl/execution:39:
> In file included from /home/local/current/distributed-ranges/build/_deps/dpl-src/include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h:23:
> /home/local/current/distributed-ranges/build/_deps/dpl-src/include/oneapi/dpl/internal/../pstl/glue_numeric_impl.h:43:12: error: call to function 'transform_reduce' that is neither visible in the template definition nor found by argument-dependent lookup
>    43 |     return transform_reduce(::std::forward<_ExecutionPolicy>(__exec), __first, __last, __init, __binary_op,
>       |            ^
> /home/local/current/distributed-ranges/include/./dr/mp/algorithms/reduce.hpp:38:19: note: in instantiation of function template specialization 'oneapi::dpl::reduce<oneapi::dpl::execution::device_policy<>, dr::__detail::direct_iterator<double *>, double, (lambda at /home/local/current/distributed-ranges/benchmarks/gbench/mp/shallow_water.cpp:891:14)>' requested here
>    38 |       return std::reduce(dpl_policy(),
>       |                   ^
> /home/local/current/distributed-ranges/include/./dr/mp/algorithms/reduce.hpp:138:20: note: in instantiation of function template specialization 'dr::mp::__detail::reduce<double, dr::mp::distributed_mdarray<double, 2> &, (lambda at /home/local/current/distributed-ranges/benchmarks/gbench/mp/shallow_water.cpp:891:14) &>' requested here
>   138 |   return __detail::reduce(0, false, std::forward<DR>(dr), init, binary_op);
>       |                    ^
> /home/local/current/distributed-ranges/build/_deps/dpl-src/include/oneapi/dpl/pstl/glue_numeric_defs.h:60:1: note: 'transform_reduce' should be declared prior to the call site or in an associated namespace of one of its arguments
>    60 | transform_reduce(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _Tp __init,
>       | ^

Failing code can be seen for example here: https://github.com/oneapi-src/distributed-ranges/actions/runs/11216478807/job/31175864199
If you have any idea what could cause that, I will be happy to provide any information that you need to solve it!

dmitriy-sobolev:
The PR's aim is to fix #1886.

#1886 reproduces when oneDPL headers are included in the following order:
```
#include <oneapi/dpl/numeric>
```

The reason appears to be that  `glue_numeric_impl.h` is included before `glue_numeric_defs.h` in `inclusive_scan_by_segment_impl.h`, what means that `reduce` uses `transform_reduce`, which is defined later.

Unfortunately, I was not able to reproduce the issue with a simple reproducer (outside of the distributed-ranges context), to come up with a new test.

### DIFF
--- a/include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
@@ -20,7 +20,10 @@
 #define _ONEDPL_INCLUSIVE_SCAN_BY_SEGMENT_IMPL_H
 
 #include "by_segment_extension_defs.h"
-#include "../numeric" // inclusive_scan, transform_inclusive_scan
+
+#include "../pstl/glue_numeric_defs.h"
+#include "../pstl/glue_numeric_impl.h"
+
 #include "../pstl/parallel_backend.h"
 #include "function.h"
 #include "../pstl/utils.h"
--- a/include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
@@ -20,7 +20,7 @@
 #define _ONEDPL_INCLUSIVE_SCAN_BY_SEGMENT_IMPL_H
 
 #include "by_segment_extension_defs.h"
-#include "../pstl/glue_numeric_impl.h"
+#include "../numeric" // inclusive_scan, transform_inclusive_scan
 #include "../pstl/parallel_backend.h"
 #include "function.h"
 #include "../pstl/utils.h"
--- a/include/oneapi/dpl/pstl/glue_numeric_impl.h
+++ b/include/oneapi/dpl/pstl/glue_numeric_impl.h
@@ -33,6 +33,35 @@ namespace oneapi
 namespace dpl
 {
 
+// [reduce]
+
+template <class _ExecutionPolicy, class _ForwardIterator, class _Tp, class _BinaryOperation>
+oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _Tp>
+reduce(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _Tp __init,
+       _BinaryOperation __binary_op)
+{
+    return transform_reduce(::std::forward<_ExecutionPolicy>(__exec), __first, __last, __init, __binary_op,
+                            oneapi::dpl::__internal::__no_op());
+}
+
+template <class _ExecutionPolicy, class _ForwardIterator, class _Tp>
+oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _Tp>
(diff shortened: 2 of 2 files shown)


SergeyKopienko:
Probably we have another way to fix issues like this: we should include `include/oneapi/dpl/pstl/glue_numeric_defs.h` before `include/oneapi/dpl/pstl/glue_numeric_impl.h` or from `include/oneapi/dpl/pstl/glue_numeric_impl.h`.
I think it's normal when some functions calls another functions implemented below. And not always we are able to fix issue like https://github.com/oneapi-src/oneDPL/issues/1886 by changing functions order in one file.

dmitriy-sobolev:
> Probably we have another way to fix issues like this: we should include `include/oneapi/dpl/pstl/glue_numeric_defs.h` before `include/oneapi/dpl/pstl/glue_numeric_impl.h` or from `include/oneapi/dpl/pstl/glue_numeric_impl.h`. I think it's normal when some functions calls another functions implemented below. And not always we are able to fix issue like #1886 by changing functions order in one file.

I agree that swapping the functions is not the best idea, which can be treated as a short-term patching and code convolution. I've included `numeric` instead of both impl and defs headers for simplicity and reliability (`numeric` public header has all the logic to properly include the needed algorithms).

include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
```diff
@@ -20,7 +20,7 @@
 #define _ONEDPL_INCLUSIVE_SCAN_BY_SEGMENT_IMPL_H
 
 #include "by_segment_extension_defs.h"
-#include "../pstl/glue_numeric_impl.h"
+#include "../numeric" // inclusive_scan, transform_inclusive_scan
```
SergeyKopienko, line 23: Looks like usually we include it as `#include <numeric>` as I seen.

dmitriy-sobolev: We need oneDPL header here rather than the standard one. `../numeric` guarantees that.

SergeyKopienko: ok

SergeyKopienko (dismissed):
LGTM

include/oneapi/dpl/internal/inclusive_scan_by_segment_impl.h
MikeDvorskiy, line 23: I think it is better way just to include `glue_numeric_defs.h` for making the necessary declarations visible.

dmitriy-sobolev: Done

PR #1896, opened 2024-10-11
Include all necessary standard headers in utils_ranges.h

dmitriy-sobolev:
The PR includes all the necessary standard headers while fixing the issue:
```bash
utils_ranges.h:81:23: error: 'uint8_t' in namespace 'std' does not name a type; did you mean 'wint_t'?
```


MikeDvorskiy (approved):
LGTM

PR #1897, opened 2024-10-11
Split zip_iterator.pass into 8 smaller tests to prevent out-of-memory crashes

mmichel11:
On some client discrete GPUs with debug builds, our `zip_iterator` test is crashing with out-of-memory exceptions. This is due to the large number of kernels that are compiled within the test. These kernels are stored in device memory after compilation which gets fully exhausted with the large kernel sizes produced in debug builds.

I have split `zip_iterator.pass` into a set of smaller tests based on type and confirmed that the tests now pass on these failing devices. 

There are a few additional changes made:

- `test_min_element` and `test_find_if` were implemented but never actually called. I have added calls to these tests and created `_PSTL_TEST_MIN_ELEMENT` and `_PSTL_TEST_FIND_IF` macros.
- `_PSTL_TEST_LEXICOGRAPHICAL_COMPARE` was spelled incorrectly and corrected.
- `::std -> std`
- Some unnecessary includes were removed.


test/parallel_api/iterator/zip_iterator.h
```diff
@@ -0,0 +1,50 @@
...
+//
+// This file incorporates work covered by the following copyright and permission
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#ifndef _ZIP_ITERATOR_H
```
timmiesmith, line 16: I suggest renaming this to zip_iterator_funcs.h to better reflect what is here.

timmiesmith (changes requested):
@mmichel11 are there any code changes to the individual cpp files to improve the tests or is this a strict splitting of the code with the required duplication of header includes only?

mmichel11:
@timmiesmith The changes in the description are new. Some removal of unnecessary internal includes as well as adding test calls to `test_min_element` and `test_find_if`. The classes for these two tests were implemented but were never actually called.

test/parallel_api/iterator/zip_iterator.h
mmichel11: Done

test/parallel_api/iterator/zip_iterator_merge.pass.cpp
```diff
@@ -0,0 +1,137 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+#include "support/test_config.h"
+#include "support/utils.h"
+
+#include _PSTL_TEST_HEADER(execution)
+#include _PSTL_TEST_HEADER(algorithm)
+
+#if !defined(_PSTL_TEST_MERGE)
```
SergeyKopienko, line 22: Why we need this code in this file and in some another new files too?

test/parallel_api/iterator/zip_iterator_equal.pass.cpp
```diff
@@ -0,0 +1,179 @@
...
+
+#if !defined(_PSTL_TEST_EQUAL) && \
+    !defined(_PSTL_TEST_FOR_EACH_STRUCTURED_BINDING)
+#define _PSTL_TEST_EQUAL
+#define _PSTL_TEST_EQUAL_STRUCTURED_BINDING
+#endif
+
+using namespace TestUtils;
+
+#if TEST_DPCPP_BACKEND_PRESENT
```
SergeyKopienko, line 30: I propose write all includes in the begin of the file: it's better for reading.

mmichel11: I have moved the includes to the beginning of each file.

test/parallel_api/iterator/zip_iterator_merge.pass.cpp
mmichel11: It is a good question. I have just split up what was already in `zip_iterator.pass.cpp` into each new file. I agree if these are not needed, then we should just remove them.

Do you know who may have a good idea of why these were initially added?

danhoeflinger: My understanding is that these are used for debugging purposes to isolate specific aspects of a large test.   Since that is effectively what this PR is doing, i don't think they are necessary anymore.

mmichel11: Thanks for the clarification. I agree these guards are no longer necessary and have removed them.

test/parallel_api/iterator/zip_iterator_sort.pass.cpp
```diff
@@ -80,12 +76,10 @@ void
 test_usm_and_buffer()
 {
     using ValueType = std::int32_t;
-// sorting with zip iterator does not meet limits of RAM usage on FPGA.
-// TODO: try to investigate and reduce RAM consumption
-#if defined(_PSTL_TEST_STABLE_SORT) && !ONEDPL_FPGA_DEVICE
```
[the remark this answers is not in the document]
mmichel11: Thanks for pointing this out. The tests pass locally on my side with FPGA_EMU in both release and debug builds. I will do a limited CI run to confirm and then remove this comment.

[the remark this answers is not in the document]
mmichel11: I see no memory issues with `zip_iterator_sort` with FPGA emulator, so both the comment and guard are now removed.

danhoeflinger (approved):
LGTM

test/parallel_api/iterator/zip_iterator_find.pass.cpp
```diff
@@ -0,0 +1,115 @@
...
+        auto tuple_res4 = std::find(make_new_policy<new_kernel_name<Policy, 3>>(exec), tuple_first1, tuple_last1,
+                                    std::make_tuple(T1{-1}, T1{-1}));
+#if _PSTL_SYCL_TEST_USM
+        exec.queue().wait_and_throw();
+#endif
+        EXPECT_TRUE(tuple_res4 == tuple_last1, "wrong effect from find (tuple)");
+    }
+};
+
+template <sycl::usm::alloc alloc_type>
```
SergeyKopienko, line 90: This can be moved under `TEST_DPCPP_BACKEND_PRESENT` too.

SergeyKopienko (approved):
LGTM

timmiesmith (approved):
LGTM. Thank you.

PR #1898, opened 2024-10-14
[onedpl][ranges] copy_if implementation for sized output

MikeDvorskiy:


include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1233,6 +1265,28 @@ __brick_calc_mask_1(_ForwardIterator __first, _ForwardIterator __last, bool* __r
     return ::std::make_pair(__count_true, __size - __count_true);
 }
 
+template <class _DifferenceType, class _ForwardIterator, class _Bound, class _UnaryPredicate>
+std::pair<_DifferenceType, _ForwardIterator>
+__brick_calc_mask_1(_ForwardIterator __first, _ForwardIterator __last, _Bound __m, bool* __restrict __mask, _UnaryPredicate __pred,
+                    /*vector=*/::std::false_type) noexcept
+{
+    auto __count_true = _DifferenceType(0);
+    auto __size = __last - __first;
```
mmichel11, line 1274: Since `__first` and `__last` are forward iterators we should use `std::distance`, correct?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -22,7 +22,7 @@
 #include <tuple>
 #include <algorithm>
 
-#include "../../iterator_impl.h"
+#include "../../iterator_impl.h"
```
SergeyKopienko, line 25: Space-char in the end?

SergeyKopienko:
I prepared one more draft PR https://github.com/uxlfoundation/oneDPL/pull/2163 with the changes from this PR but on current code base just to check CI.

akukanov:
Merged all the changes accumulated in the current main branch. Dropped modifications to the hetero backend done in this branch; it was incomplete anyway, and it seems easier now to re-do manually than to merge.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
```diff
@@ -23,20 +23,28 @@ main()
...
+             }
+        }
+        return ret_type{it_in, it_out};
     };
 
-    test_range_algo<0, int, data_in_out>{big_sz}(dpl_ranges::copy_if,  copy_if_checker, pred);
-    test_range_algo<1, int, data_in_out>{}(dpl_ranges::copy_if,  copy_if_checker, pred, proj);
-    test_range_algo<2, P2, data_in_out>{}(dpl_ranges::copy_if,  copy_if_checker, pred, &P2::x);
-    test_range_algo<3, P2, data_in_out>{}(dpl_ranges::copy_if,  copy_if_checker, pred, &P2::proj);
+    test_range_algo<0, int, data_in_out_lim>{big_sz}(dpl_ranges::copy_if,  copy_if_checker, pred, std::identity{});
```
dmitriy-sobolev: After adding `std::identity`, the interface without a projection is not tested anymore. Let's remove `std::idenity` unless it is necessary (I do not know why it would be, however).

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -71,11 +71,11 @@ __pattern_transform(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r, _O
...
     assert(std::ranges::size(__in_r) <= std::ranges::size(__out_r)); // for debug purposes only
 
-    oneapi::dpl::__internal::__unary_op<_F, _Proj> __unary_op{__op, __proj};
+    oneapi::dpl::__internal::__transform_functor<oneapi::dpl::__internal::__unary_op<_F, _Proj>> __f{{__op, __proj}};
 
     oneapi::dpl::__internal::__pattern_walk2(__tag, std::forward<_ExecutionPolicy>(__exec), std::ranges::begin(__in_r),
-        std::ranges::begin(__in_r) + std::ranges::size(__in_r), std::ranges::begin(__out_r),
-        oneapi::dpl::__internal::__transform_functor<decltype(__unary_op)>{std::move(__unary_op)});
+                                             std::ranges::begin(__in_r) + std::ranges::size(__in_r),
+                                             std::ranges::begin(__out_r), __f);
```
dmitriy-sobolev, line 78: [nitpick] Let's move unrelated refactoring changes into a separate PR.

akukanov: It's not unrelated refactoring but an attempt to fix a warning after adding MSVC configurations to CI.

I guess I have addressed it in another way, most likely with suppression. However, these changes are just 5 lines of modified code (see https://github.com/uxlfoundation/oneDPL/pull/1898/commits/fbbda9cb2f920b93dc9c9b7a5d66d9e887e52093), and honestly, I do not think separating those 5 lines into another PR is worth the time that would be spent by me and reviewers.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1354,6 +1424,47 @@ __pattern_copy_if(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
...
+                        [](_RandomAccessIterator1 __x, _RandomAccessIterator2 __z) { *__z = *__x; }, _IsVector{});
+                    if (__res.second + __initial == __n_out) // Reached the end of the output
+                        __res_in = __i + __res.first;
+                }
+            },
+            [&__res_out](auto __total_out) { // Apex
+                if (__total_out < __res_out) // Output size is bigger than needed
+                    __res_out = __total_out;
+            });
+        return std::make_pair(__first + __res_in, __result + __res_out);
```
SergeyKopienko, line 1480: I believe no reasons to call `std::make_pair` because we know the return type so we can just return `{__first + __res_in, __result + __res_out}`

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -544,29 +545,57 @@ __pattern_copy(__serial_tag</*IsVector*/std::false_type>, _ExecutionPolicy&&, _I
...
 
 template <typename _ExecutionPolicy, typename _InRange, typename _OutRange, typename _Pred, typename _Proj>
 std::ranges::copy_if_result<std::ranges::borrowed_iterator_t<_InRange>, std::ranges::borrowed_iterator_t<_OutRange>>
 __pattern_copy_if_ranges(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy&&, _InRange&& __in_r,
                          _OutRange&& __out_r, _Pred __pred, _Proj __proj)
 {
-    return std::ranges::copy_if(std::forward<_InRange>(__in_r), std::ranges::begin(__out_r), __pred, __proj);
+    auto __it_in = std::ranges::begin(__in_r);
+    auto __it_out = std::ranges::begin(__out_r);
+    for (; __it_in != std::ranges::end(__in_r) && __it_out != std::ranges::end(__out_r); ++__it_in)
```
SergeyKopienko: Looks like multiple calls of `std::ranges::end` inside `for`-loop.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
```diff
@@ -23,20 +23,28 @@ main()
...
-        auto res = std::ranges::copy_if(std::forward<decltype(r_in)>(r_in), std::ranges::begin(r_out),
-            std::forward<decltype(args)>(args)...);
-
         using ret_type = std::ranges::copy_if_result<std::ranges::borrowed_iterator_t<decltype(r_in)>,
             std::ranges::borrowed_iterator_t<decltype(r_out)>>;
-        return ret_type{res.in, res.out};
+
+        auto it_in = std::ranges::begin(r_in);
+        auto it_out = std::ranges::begin(r_out);
+        for(; it_in != std::ranges::end(r_in) && it_out != std::ranges::end(r_out); ++it_in)
```
SergeyKopienko: The same point about `std::ranges::end` calls

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
akukanov: That's a good point that the change degrades test coverage. But I think it might be necessary because of the custom serial checker which does not have the default argument for the projection. I will explore if that can be added to the lambda; if not, the checker will need to be rewritten as a class.

include/oneapi/dpl/pstl/algorithm_impl.h
akukanov: Apparently, `std::make_pair` is better here because the return expression is for the lambda of `__except_handler`, not the function. Adding explicit return type for the lambda would be more verbose.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
akukanov: Changed to use local variables.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
akukanov: Rewritten to use index-based iteration instead - primarily to make the test code sufficiently different from the serial implementation. The repeated calls to `std::ranges::size` are fine I believe, as the standard guarantees constant time complexity.

akukanov:
> Meanwhile, what is the plan regarding the failing tests with the device policies?

If the implementation for device policies takes long time, disabling is the only option I guess.

> may be make sense to introduce some type alias for `std::ranges::copy_if_result<...>`

`copy_if` is not the only algorithm that uses a standard result type in function declarations. I prefer to keep it consistent.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -544,29 +545,59 @@ __pattern_copy(__serial_tag</*IsVector*/std::false_type>, _ExecutionPolicy&&, _I
...
+                         _OutRange&& __out_r, _Pred __pred, _Proj __proj)
 {
-    static_assert(__is_parallel_tag_v<_Tag> || typename _Tag::__is_vector{});
+    auto __first_in = std::ranges::begin(__in_r);
+    auto __first_out = std::ranges::begin(__out_r);
+    auto __sz_in = std::ranges::size(__in_r);
+    auto __sz_out = std::ranges::size(__out_r);
 
-    oneapi::dpl::__internal::__unary_op<_Pred, _Proj> __pred_1{__pred, __proj};
+    if (__sz_in > 0 && __sz_out > 0)
```
dmitriy-sobolev, line 560: Does it make sense to redirect to the ordinary pattern in case of `__sz_out >= __sz_in`?

My conjecture is that  `__sz_out == __sz_in` should be common, and `__pattern_bounded_copy_if` is slower than `__pattern_copy_if` due to additional checks.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1274,6 +1298,52 @@ __brick_copy_by_mask(_RandomAccessIterator1 __first, _RandomAccessIterator1 __la
...
+        _Bound __copied = __unseq_backend::__simd_copy_by_mask(__first, __m, __result, __mask, __assigner);
+        __n -= __m;
+        __first += __m;
+        __m -= __copied;
+        __result += __copied;
+    }
+    // The loop above may not decrease __m or __n below 0
+    if (__m >= n) // enough space left for the rest
+    {
+        __result += __unseq_backend::__simd_copy_by_mask(__first, __n, __result, __mask, __assigner);
```
dmitriy-sobolev: Adding to `__result` seems to be not needed - it will not be used.
I suppose we should subtract the returned value from `__n` and `__m`.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
```diff
@@ -22,21 +22,29 @@ main()
...
     };
 
-    test_range_algo<0, int, data_in_out>{big_sz}(dpl_ranges::copy_if,  copy_if_checker, pred);
-    test_range_algo<1, int, data_in_out>{}(dpl_ranges::copy_if,  copy_if_checker, pred, proj);
-    test_range_algo<2, P2, data_in_out>{}(dpl_ranges::copy_if,  copy_if_checker, pred, &P2::x);
-    test_range_algo<3, P2, data_in_out>{}(dpl_ranges::copy_if,  copy_if_checker, pred, &P2::proj);
+    test_range_algo<0, int, data_in_out_lim>{big_sz}(dpl_ranges::copy_if,  copy_if_checker, pred);
+    test_range_algo<1, int, data_in_out_lim>{}(dpl_ranges::copy_if,  copy_if_checker, pred, proj);
+    test_range_algo<2, P2, data_in_out_lim>{}(dpl_ranges::copy_if,  copy_if_checker, pred, &P2::x);
+    test_range_algo<3, P2, data_in_out_lim>{}(dpl_ranges::copy_if,  copy_if_checker, pred, &P2::proj);
```
dmitriy-sobolev: Do you think that the testing coverage is sufficient given that this copy_if fully implements its patterns and bricks? I'm hesitant because there are still many common parts with the iterator-based copy_if.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
akukanov: We need first to measure and observe that it is notably slower. I can add a TODO for this, if you prefer,

dmitriy-sobolev: Agreed. Let's add a TODO

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
akukanov: I think with added corner-case tests for `data_in_out_lim` in `std_ranges_test.h` it should be sufficient, as various combinations of sizes, as well as a variety of "base" sizes, are tested.

If you have some specific suggestions for more tests, please elaborate.

include/oneapi/dpl/pstl/algorithm_impl.h
akukanov: A good catch; thanks for noticing.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
dmitriy-sobolev:
```suggestion
    test_range_algo<0, int, data_in_out_lim>{big_sz}(dpl_ranges::copy_if,  copy_if_checker, [](auto val){ return val % 42 != 0; });
    test_range_algo<1, int, data_in_out_lim>{}(dpl_ranges::copy_if,  copy_if_checker, pred, proj);
    test_range_algo<2, P2, data_in_out_lim>{}(dpl_ranges::copy_if,  copy_if_checker, pred, &P2::x);
    test_range_algo<3, P2, data_in_out_lim>{}(dpl_ranges::copy_if,  copy_if_checker, pred, &P2::proj);
```

That would be my only suggestion.

The reason is that the data is generated as `0, 1, 2, ... n - 1` and the existing `pred` just checks if the value is equal to `5`. It looks like we do not check a case when the output sequence capacity is not sufficient.

With the update predicate, it will be checked here:
`process_data_in_out(max_n, r_size, r_size/2, CLONE_TEST_POLICY(exec), algo, checker, args...);`

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1465,14 +1455,22 @@ __pattern_bounded_copy_if(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec,
...
-                        __first + __i, __len, __result + __initial, __n_out - __initial, __mask + __i,
+                        __first + __i, __len, __result + __initial, __n_out - __initial, __mask_start,
                         [](_RandomAccessIterator1 __x, _RandomAccessIterator2 __z) { *__z = *__x; }, _IsVector{});
-                    if (__res.second + __initial == __n_out) // Reached the end of the output
-                        __res_in = __i + __res.first;
+                    if (__res.first == __len)
+                        return;
+                    __mask_start += __res.first;
                 }
+                bool* __stop = __brick_find_if(__mask_start, __mask_end, oneapi::dpl::identity{}, _IsVector{});
```
dmitriy-sobolev: Consistency with `__internal::__brick_bounded_copy_by_mask`: `__brick_find_if` -> `__internal::__brick_find_if`.

dmitriy-sobolev (commented):
Still trying to understand the modified `__pattern_bounded_copy_if` in detail. The rest changes look good.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1354,6 +1431,55 @@ __pattern_copy_if(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
...
+                bool* __mask_start = __mask + __i;
+                bool* __mask_end = __mask + (__i + __len);
+                if (__initial < __n_out)
+                {
+                    _DifferenceType __checked = __internal::__brick_bounded_copy_by_mask(
+                        __first + __i, __len, __result + __initial, __n_out - __initial, __mask_start,
+                        [](_RandomAccessIterator1 __x, _RandomAccessIterator2 __z) { *__z = *__x; }, _IsVector{});
+                    if (__checked == __len)
+                        return;
+                    __mask_start += __checked;
```
dmitriy-sobolev, line 1469: Do we need to move `__mask_start` one item further?

As far as I understand, currently, it will point to the last copied element and `__brick_find_if` below will find it, while we want to find the next matching element.

akukanov: No, `__mask_start` points to an element after the last copied.

`__brick_bounded_copy_by_mask` should return the number of checked elements in the mask (please double-check if that's the case). So when we shift `__mask_start` by that number, it will again point to the first unchecked element.

dmitriy-sobolev: Checked. Indeed, it works as expected. Thanks for the explanation.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1197,24 +1197,53 @@ __brick_swap_ranges(_RandomAccessIterator1 __first, _RandomAccessIterator1 __las
...
 #else
-    return ::std::copy_if(__first, __last, __result, __pred);
+    return std::copy_if(__first, __last, __result, __pred);
 #endif
 }
 
+template <class _RandomAccessIterator1, class _RandomAccessIterator2, class _UnaryPredicate>
+std::pair<_RandomAccessIterator1, _RandomAccessIterator2>
+__brick_bounded_copy_if(_RandomAccessIterator1 __first,
+                        typename std::iterator_traits<_RandomAccessIterator2>::difference_type __n,
```
dmitriy-sobolev: `__n` is used with `__first`. Should there be `_RandomAccessIterator1`?

akukanov: Probably a copy-paste error, thanks for catching.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1274,6 +1303,54 @@ __brick_copy_by_mask(_RandomAccessIterator1 __first, _RandomAccessIterator1 __la
...
+    {
+        _Bound __copied = __unseq_backend::__simd_copy_by_mask(__first, __m, __result, __mask, __assigner);
+        __n -= __m;
+        __first += __m;
+        __mask += __m;
+        __m -= __copied;
+        __result += __copied;
+    }
+    // The loop above may not decrease __m or __n below 0
+    if (__m >= n) // enough space left for the rest
```
SergeyKopienko:
```suggestion
    if (__m >= __n) // enough space left for the rest
```

SergeyKopienko: It should generate compile error but hasn't been illuminated. Does this mean tests not covered some code?

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1354,6 +1431,55 @@ __pattern_copy_if(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
...
+    using __backend_tag = typename __parallel_tag<_IsVector>::__backend_tag;
+    using _DifferenceType = typename std::iterator_traits<_RandomAccessIterator1>::difference_type;
+
+    __par_backend::__buffer<bool> __mask_buf(__n);
+    bool* __mask = __mask_buf.get();
+    return __internal::__except_handler([&__exec, __n, __first, __result, __pred, __mask, __n_out]() {
+        _DifferenceType __res_in{__n}, __res_out{__n_out};
+        __par_backend::__parallel_strict_scan(
+            __backend_tag{}, std::forward<_ExecutionPolicy>(__exec), __n, _DifferenceType(0),
+            [=](_DifferenceType __i, _DifferenceType __len) { // Reduce
```
SergeyKopienko, line 1452: In this `reduce` - part we doesn't take into account the amount of output data and process all input data.
I don't understand is it possible or no....

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
```diff
@@ -22,21 +22,34 @@ main()
     using namespace test_std_ranges;
     namespace dpl_ranges = oneapi::dpl::ranges;
 
-    auto copy_if_checker = [](std::ranges::random_access_range auto&& r_in,
-                           std::ranges::random_access_range auto&& r_out, auto&&... args)
+    auto copy_if_checker = []<typename Proj = std::identity>(std::ranges::random_access_range auto&& r_in,
```
SergeyKopienko, line 25: Evidence for this implementation: https://eel.is/c++draft/alg.copy
```
Otherwise, {j, result_last} for the overloads in namespace ranges, where j is the iterator in [first, last) for which E(j) holds and there are exactly N iterators i in [first, j) for which E(i) holds[.](https://eel.is/c++draft/alg.copy#23.3.sentence-1)
```

include/oneapi/dpl/pstl/algorithm_impl.h
akukanov: Thanks for catching. Yes, it is not covered - see Dmitry's comment above that we now do not have a compiler that supports both OpenMP monotonic clause and C++20.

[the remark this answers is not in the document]
dmitriy-sobolev: Already reported.

include/oneapi/dpl/pstl/algorithm_impl.h
akukanov: Since the positions of which elements to copy are not predictable, it is not clear how to utilize the output size for anything good while computing and reducing the masks.

SergeyKopienko: ok

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1354,6 +1431,56 @@ __pattern_copy_if(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
...
+                return __internal::__brick_calc_mask_1<_DifferenceType>(
+                    __first + __i, __first + (__i + __len), __mask + __i, __pred, _IsVector{}).first;
+            },
+            std::plus<_DifferenceType>(), // Combine
+            [=, &__res_in](_DifferenceType __i, _DifferenceType __len, _DifferenceType __initial) { // Scan
```
danhoeflinger, line 1457: [optional nitpick]  I'd prefer to have `__mask, __n_out, __first, __result` listed explicitly rather than `=`, similar for lambda above.

For me this improves clarity and helps show where different variables are coming from more explicitly.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -1354,6 +1431,56 @@ __pattern_copy_if(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
...
+                        __first + __i, __len, __result + __initial, __n_out - __initial, __mask_start,
+                        [](_RandomAccessIterator1 __x, _RandomAccessIterator2 __z) { *__z = *__x; }, _IsVector{});
+                    if (__checked == __len)
+                        return;
+                    __mask_start += __checked;
+                }
+                bool* __stop =
+                    __internal::__brick_find_if(__mask_start, __mask_end, oneapi::dpl::identity{}, _IsVector{});
+                if (__stop != __mask_end) // Found the position of the first element that cannot be copied
+                    __res_in = __stop - __mask;
```
danhoeflinger: Am I correct that exactly one parallel iteration / segment will write to `__res_in` (and there is no race) because we have the following conditions:
1) my segment exists completely beyond the end of the output to be written - exit from first early `return`
2) my segment can write all of its masked elements fully within the output size - exit from second early `return`
3) my segment has found an element which should be written but cant be written to the output (only segment which writes)
4) my segment had elements left to check but found no elements to write in the remaining elements: (false to this if conditional)

A comment here would be useful to clarify that at most a single segment can write here (without saying all of this).

akukanov: Yes, your reasoning is correct.

There is also another way to think of it: this write is essentially no different than any write to the output range. For each element of the output there is exactly one element in the input to be written, by exactly one thread. And the write to `__res_in` is equivalent to an imaginary write to the output at the position `__n_out`.

akukanov: Added a comment.

include/oneapi/dpl/pstl/algorithm_impl.h
akukanov: Sorry, but I will not do that. It would make the function differ (in not an important way) from similar ones like `__pattern_copy_if` and `__pattern_unique_copy`, and so would complicate possible refactoring.

danhoeflinger: ok

PR #1905, opened 2024-10-15
Fixed coverity issue in Philox RNG engine

ElenaTyuleneva:
**What introduces this PR:**
1. Fix for the incorrect counter incrementing in case `w` is not the recommended value (`w!=32` and `w!=64`)
2. `discard_overflow_test` was extended: previously only the first chunk of the counter was overflowing, now introduced the high-level loop, which iterates through the other positions of the counter(2nd and 3rd) and also checks for the overflow.
3. Introduced a new unit test - `counter_management_test`. 
   - The testing is based on comparing the work of two different methods of the engine - simple `set_counter` and more complicated `increase_counter_internal(unsigned long long __z)` which is called by `discard`
   - The test discards the engine by the increment which takes at least 2 chunks of the counter type(`2*w`) - this way `increase_counter_internal` will take `__z` that should be placed properly in the 1st chunk and propagated to the 2nd one. The right calculations are checked by `referenceEngine` which is based on simpler `set_counter` function.

### DIFF
--- a/test/xpu_api/random/unit_tests/philox_unit_test.pass.cpp
+++ b/test/xpu_api/random/unit_tests/philox_unit_test.pass.cpp
@@ -162,7 +162,7 @@ main()
     std::cout << "void counter_management_test() [Engine = philox4x64_w25]";
     err += counter_management_test<philox4x64_w25>();
     std::cout << "void counter_management_test() [Engine = philox4x64_w49]";
-    err += counter_management_test<philox4x64_w49>(); 
+    err += counter_management_test<philox4x64_w49>();
 
     return TestUtils::done();
 }
@@ -367,10 +367,11 @@ discard_overflow_test()
     using T = typename Engine::result_type;
     using scalar_type = typename Engine::scalar_type;
 
-    // Iterate throug the counter's position being overflown 
-    for(int overflown_position = 0; overflown_position < Engine::word_count - 1; ++overflown_position) {
+    // Iterate throug the counter's position being overflown
+    for (int overflown_position = 0; overflown_position < Engine::word_count - 1; ++overflown_position)
+    {
         Engine engine1;
-        std::array<T, Engine::word_count> counter = { 0 };
+        std::array<T, Engine::word_count> counter = {0};
 
         // Overflow of a counter's element. Correspondance is the following:
         // 0 1 2 3 overflow position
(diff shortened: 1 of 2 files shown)


test/xpu_api/random/unit_tests/philox_unit_test.pass.cpp
```diff
@@ -338,42 +362,99 @@ template <typename Engine>
...
+        Engine engine1;
+        std::array<T, Engine::word_count> counter = {0};
 
-    // overflow of the 0th counter element
-    counter[2] = 1;
+        // Overflow of a counter's element. The correspondence is the following:
+        // 0 1 2 3 overflow position
+        // 1 2 3 0 set-up counter position in engine
+        // 2 1 0 3 raw_counter_position
+        int raw_counter_position = (Engine::word_count - overflown_position - 2) % Engine::word_count;
```
aelizaro, line 380: please, check this line if `Engine::word_count - overflown_position - 2` would overflow or is it expected?

also is -2 here only for Engine::word_count = 4?

test/xpu_api/random/unit_tests/philox_unit_test.pass.cpp
```diff
@@ -338,42 +362,99 @@ template <typename Engine>
...
+int
+counter_management_test()
+{
+    using T = typename Engine::result_type;
+    Engine testedEngine;
+    Engine referenceEngine;
+
+    // set the counter which value is 2-chunk bitsize
+    unsigned long long increment = ((unsigned long long)testedEngine.max() << Engine::word_size) | testedEngine.max();
+    unsigned long long counter_increment = increment / 4;
```
aelizaro, line 434:
```suggestion
    unsigned long long counter_increment = increment / Engine::word_count;
```

ElenaTyuleneva: Thanks for catching that! Fixed + all unit tests are also extended for Philox 2x*(except for `set_counter_conformance_test` as it expects `1955073260` and `3409172418970261260` numbers to be generated as 10000th)

test/xpu_api/random/unit_tests/philox_unit_test.pass.cpp
ElenaTyuleneva: > please, check this line if Engine::word_count - overflown_position - 2 would overflow or is it expected?

This question was discussed offline.

> also is -2 here only for Engine::word_count = 4?

I've double-checked, this equation is also valid for `word_count = 2`

aelizaro (approved):
Looks good to me, thank you!

PR #1907, opened 2024-10-16
Modify is_execution_policy test to reflect change in specification

### LINKED ISSUE #1901: Add testing of oneapi::dpl::is_execution_policy
A recent modification of the oneDPL specification https://github.com/uxlfoundation/oneAPI-spec/pull/567 introduced the execution policy trait into the namespace `oneapi::dpl`, in addition to `oneapi::dpl::execution`.

While the corresponding aliases have already been added to the implementation, there are no tests checking that `oneapi::dpl::is_execution_policy` works as expected.

adamfidel:
Closes #1901.

The oneDPL specification has recently changed (https://github.com/uxlfoundation/oneAPI-spec/pull/567) to move `is_execution_policy` and `is_execution_policy_v` from the `oneapi::dpl::execution` namespace to `oneapi::dpl`. This PR changes the tests to use the struct in the `oneapi::dpl` namespace for testing.

### DIFF
--- a/test/general/test_policies.pass.cpp
+++ b/test/general/test_policies.pass.cpp
@@ -46,7 +46,7 @@ void test_policy_instance(const Policy& policy)
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
 template<typename Policy>
-void assert_is_execution_policy()
+constexpr void assert_is_execution_policy()
 {
     static_assert(oneapi::dpl::is_execution_policy<Policy>::value, "wrong result for oneapi::dpl::is_execution_policy");
     static_assert(oneapi::dpl::is_execution_policy_v<Policy>, "wrong result for oneapi::dpl::is_execution_policy_v");
--- a/test/general/test_policies.pass.cpp
+++ b/test/general/test_policies.pass.cpp
@@ -115,6 +115,9 @@ main()
 
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
+    // Test that the policy is not decayed
+    static_assert(!is_execution_policy_v<execution::sequenced_policy&&>, "wrong result for is_execution_policy_v<sequenced_policy>");
+
     return TestUtils::done(TEST_DPCPP_BACKEND_PRESENT);
 }
 
--- a/test/general/test_policies.pass.cpp
+++ b/test/general/test_policies.pass.cpp
@@ -48,23 +48,23 @@ void test_policy_instance(const Policy& policy)
 std::int32_t
 main()
 {
-    using namespace oneapi::dpl::execution;
-    static_assert(is_execution_policy<sequenced_policy>::value, "wrong result for is_execution_policy<sequenced_policy>");
-    static_assert(is_execution_policy<unsequenced_policy>::value, "wrong result for is_execution_policy<unsequenced_policy>");
-    static_assert(is_execution_policy<parallel_policy>::value, "wrong result for is_execution_policy<parallel_policy>");
-    static_assert(is_execution_policy<parallel_unsequenced_policy>::value, "wrong result for is_execution_policy<parallel_unsequenced_policy>");
-    static_assert(is_execution_policy_v<sequenced_policy>, "wrong result for is_execution_policy_v<sequenced_policy>");
-    static_assert(is_execution_policy_v<unsequenced_policy>, "wrong result for is_execution_policy_v<unsequenced_policy>");
-    static_assert(is_execution_policy_v<parallel_policy>, "wrong result for is_execution_policy_v<parallel_policy>");
-    static_assert(is_execution_policy_v<parallel_unsequenced_policy>, "wrong result for is_execution_policy_v<parallel_unsequenced_policy>");
+    using namespace oneapi::dpl;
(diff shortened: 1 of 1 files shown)


danhoeflinger:
I think that this PR is not exactly making the change suggested by the change in the specification. This PR just changes the namespace we are `using` and then inserts `execution` inline. I think we need to remove all the inline `execution` to comply with the "suggestion for new code". However, it probably makes sense to test both `oneapi::dpl::execution` and `oneapi::dpl` since both options are available from the specification.

adamfidel:
> I think that this PR is not exactly making the change suggested by the change in the specification. This PR just changes the namespace we are `using` and then inserts `execution` inline. I think we need to remove all the inline `execution` to comply with the "suggestion for new code". However, it probably makes sense to test both `oneapi::dpl::execution` and `oneapi::dpl` since both options are available from the specification.

The PR is only inserting `execution::` prefixes for the execution policies themselves and not for the two type traits that have moved out of that namespace (`is_execution_policy` and `is_execution_policy_v`). We still need `execution::` for the policies because they are still in the `execution` namespace whereas the utility structs are now not.

Although I agree that it probably makes sense to test both `oneapi::dpl::` and `oneapi::dpl::execution::` versions of the utility structs. Let me see how I can easily do that in the same test file.

danhoeflinger:
> The PR is only inserting `execution::` prefixes for the execution policies themselves and not for the two type traits that have moved out of that namespace (`is_execution_policy` and `is_execution_policy_v`). We still need `execution::` for the policies because they are still in the `execution` namespace whereas the utility structs are now not.
> 
> Although I agree that it probably makes sense to test both `oneapi::dpl::` and `oneapi::dpl::execution::` versions of the utility structs. Let me see how I can easily do that in the same test file.

Oh whoops yes my mistake.  My brain short circuited and made a bad assumption.  Sorry about that.

  It is probably enough to have a small check of the execution namespace just that they are present, perhaps not a full copy of the test again for both namespaces.

test/general/test_policies.pass.cpp
```diff
@@ -77,44 +77,47 @@ main()
...
+    static_assert(is_execution_policy_v<execution::fpga_policy</*unroll_factor =*/ 1, class Kernel_0>>, "wrong result for is_execution_policy_v<fpga_policy>");
     test_policy_instance(dpcpp_fpga);
 
     // make_fpga_policy
     test_policy_instance(TestUtils::make_fpga_policy</*unroll_factor =*/ 1, class Kernel_31>(sycl::queue{TestUtils::default_selector}));
     test_policy_instance(TestUtils::make_fpga_policy</*unroll_factor =*/ 2, class Kernel_32>(sycl::device{TestUtils::default_selector}));
-    test_policy_instance(TestUtils::make_fpga_policy</*unroll_factor =*/ 4, class Kernel_33>(dpcpp_fpga));
+    test_policy_instance(TestUtils::make_fpga_policy</*unroll_factor =*/ 4, class Kernel_33>(execution::dpcpp_fpga));
     // Special case: required to call make_fpga_policy directly from oneapi::dpl::execution namespace
     test_policy_instance(oneapi::dpl::execution::make_fpga_policy</*unroll_factor =*/ 8, class Kernel_34>());
```
danhoeflinger, line 104: It probably makes sense to make this consistent with the rest and just use the `execution` namespace.

danhoeflinger (commented):
I think this mostly LGTM after some small test of presence of the traits directly in the `execution` namespace too.

adamfidel:
> I think this mostly LGTM after some small test of presence of the traits directly in the `execution` namespace too.

I restructured the test a bit and it now checks for the traits in both the `oneapi::dpl` and the `oneapi::dpl::execution` namespaces.

I believe this newer version simplifies the code which results in a smaller set of changes.

danhoeflinger:
> I restructured the test a bit and it now checks for the traits in both the `oneapi::dpl` and the `oneapi::dpl::execution` namespaces.
> 
> I believe this newer version simplifies the code which results in a smaller set of changes.

Yes I like this much better, let me take a closer look but on first inspection it looks good.

test/general/test_policies.pass.cpp
```diff
@@ -45,24 +45,31 @@ void test_policy_instance(const Policy& policy)
...
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
+template<typename Policy>
+void assert_is_execution_policy()
+{
+    static_assert(oneapi::dpl::is_execution_policy<Policy>::value, "wrong result for oneapi::dpl::is_execution_policy");
+    static_assert(oneapi::dpl::is_execution_policy_v<Policy>, "wrong result for oneapi::dpl::is_execution_policy_v");
+    static_assert(oneapi::dpl::execution::is_execution_policy<Policy>::value, "wrong result for oneapi::dpl::execution::is_execution_policy");
+    static_assert(oneapi::dpl::execution::is_execution_policy_v<Policy>, "wrong result for oneapi::dpl::execution::is_execution_policy_v");
+}
```
danhoeflinger, line 55: optional nitpick, this could be `constexpr`.

danhoeflinger (dismissed):
LGTM, pending green CI

danhoeflinger (approved):

Issue #1906, opened 2024-10-15
Upgrade __buffer infrastructure to usm USM where possible


danhoeflinger:
It would be good to upgrade `__buffer_impl` within oneDPL to use USM when it is able to for performance reasons.  
We should be able to just utilize `__result_and_scratch_storage` inside of `__buffer_impl` in `parallel_backend_sycl_utils.h`, but there may be some nuance and complications when handling arbitrary typed value types.  

This has the potential to provide a performance boost to all APIs which use temporary storage internally in the form of a sycl buffer currently.

This issue stems from conversation in https://github.com/oneapi-src/oneDPL/pull/1879#discussion_r1801161138_

PR #1909, opened 2024-10-17
Specify access mode in `__result_and_scratch_storage` methods

SergeyKopienko:
In this PR we specify access mode in the two `__result_and_scratch_storage` methods:
- `__result_and_scratch_storage::__get_result_acc`
- `__result_and_scratch_storage::__get_scratch_acc` 

The main idea - to write `<sycl::access_mode::write>` is more correct then `<sycl::access_mode::read_write>` when we only write some data and to write `<sycl::access_mode::read>` is more correct then `<sycl::access_mode::read_write>` when we only read some data.


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -319,6 +319,8 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         // 1. Local scan on each workgroup
         auto __submit_event = __exec.queue().submit([&](sycl::handler& __cgh) {
             oneapi::dpl::__ranges::__require_access(__cgh, __rng1, __rng2); //get an access to data under SYCL buffer
+            // TODO what is correct access mode here for __get_scratch_acc call?
```
SergeyKopienko, line 322: @danhoeflinger do you know the correct param here and in other TODO's ?

danhoeflinger: i think this can be write

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -340,6 +342,8 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
             auto __iters_per_single_wg = oneapi::dpl::__internal::__dpl_ceiling_div(__n_groups, __wgroup_size);
             __submit_event = __exec.queue().submit([&](sycl::handler& __cgh) {
                 __cgh.depends_on(__submit_event);
+                // TODO what is correct access mode here for __get_scratch_acc call?
+                // Is default sycl::access_mode::read_write is ok?
```
danhoeflinger, line 346: i think read_write.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -362,7 +366,11 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         auto __final_event = __exec.queue().submit([&](sycl::handler& __cgh) {
             __cgh.depends_on(__submit_event);
             oneapi::dpl::__ranges::__require_access(__cgh, __rng1, __rng2); //get an access to data under SYCL buffer
+            // TODO what is correct access mode here for __get_scratch_acc call?
+            // Is default sycl::access_mode::read_write is ok?
```
danhoeflinger, line 370: i believe read.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -362,7 +366,11 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
...
             auto __temp_acc = __result_and_scratch.__get_scratch_acc(__cgh);
+            // TODO what is right access mode here?
+            // sycl::access_mode::read_write is used by default.
```
danhoeflinger, line 373: i believe write

danhoeflinger (commented):
Please instigate these yourself as well but i tried to assess each, mostly looks good with structure. we will need to confirm with some testing of course.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -377,7 +383,8 @@ struct __parallel_transform_reduce_impl
                     [=](sycl::nd_item<1> __item_id) {
                         auto __temp_ptr = __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__temp_acc);
                         auto __res_ptr =
-                            __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__res_acc, 2 * __n_groups);
+                            __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr<sycl::access_mode::write>(
+                                __res_acc, 2 * __n_groups);
```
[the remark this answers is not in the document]
SergeyKopienko: Agree. It's my error.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -358,7 +360,11 @@ struct __parallel_transform_reduce_impl
             __reduce_event = __exec.queue().submit([&, __is_first, __offset_1, __offset_2, __n,
                                                     __n_groups](sycl::handler& __cgh) {
                 __cgh.depends_on(__reduce_event);
+                // TODO what is correct access mode here for __get_scratch_acc call?
+                // Is default sycl::access_mode::read_write is ok?
                 auto __temp_acc = __scratch_container.__get_scratch_acc(__cgh);
+                // TODO what is correct access mode here for __get_result_acc call?
+                // Is default sycl::access_mode::read_write is ok?
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -358,7 +360,11 @@ struct __parallel_transform_reduce_impl
             __reduce_event = __exec.queue().submit([&, __is_first, __offset_1, __offset_2, __n,
                                                     __n_groups](sycl::handler& __cgh) {
                 __cgh.depends_on(__reduce_event);
+                // TODO what is correct access mode here for __get_scratch_acc call?
+                // Is default sycl::access_mode::read_write is ok?
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -478,12 +478,11 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
 template <typename _ContainerOrIterable>
 using __value_t = typename __internal::__memobj_traits<_ContainerOrIterable>::value_type;
 
-template <typename _T>
+template <typename _T, sycl::access_mode _AccessMode = sycl::access_mode::read_write>
```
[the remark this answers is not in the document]
SergeyKopienko: Agree. Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -298,10 +298,11 @@ struct __parallel_reduce_then_scan_reduce_submitter<__sub_group_size, __max_inpu
             __dpl_sycl::__local_accessor<_InitValueType> __sub_group_partials(__num_sub_groups_local, __cgh);
             __cgh.depends_on(__prior_event);
             oneapi::dpl::__ranges::__require_access(__cgh, __in_rng);
-            auto __temp_acc = __scratch_container.__get_scratch_acc(__cgh);
+            auto __temp_acc = __scratch_container.__get_scratch_acc<sycl::access_mode::write>(__cgh);
             __cgh.parallel_for<_KernelName...>(
                     __nd_range, [=, *this](sycl::nd_item<1> __ndi) [[sycl::reqd_sub_group_size(__sub_group_size)]] {
-                _InitValueType* __temp_ptr = _TmpStorageAcc::__get_usm_or_buffer_accessor_ptr(__temp_acc);
+                _InitValueType* __temp_ptr =
+                    _TmpStorageAcc::__get_usm_or_buffer_accessor_ptr<sycl::access_mode::write>(__temp_acc);
```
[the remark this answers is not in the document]
SergeyKopienko: Agree. Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -453,8 +454,8 @@ struct __parallel_reduce_then_scan_scan_submitter<
             __dpl_sycl::__local_accessor<_InitValueType> __sub_group_partials(__num_sub_groups_local + 1, __cgh);
             __cgh.depends_on(__prior_event);
             oneapi::dpl::__ranges::__require_access(__cgh, __in_rng, __out_rng);
-            auto __temp_acc = __scratch_container.__get_scratch_acc(__cgh);
-            auto __res_acc = __scratch_container.__get_result_acc(__cgh);
+            auto __temp_acc = __scratch_container.__get_scratch_acc<sycl::access_mode::read_write>(__cgh);
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -512,6 +512,23 @@ struct __usm_or_buffer_accessor
...
+
+         else if constexpr (_AccessMode == sycl::access::mode::write)
+             return sycl::write;
+
+         else if constexpr (_AccessMode == sycl::access::mode::read_write)
+             return sycl::read_write;
+
+         else
+             static_assert(false, "Unknown _AccessMode state");
+     }
```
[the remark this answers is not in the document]
SergeyKopienko: Function has been declared as `constexpr`.
But I doesn't found any applications for it in any other places so I propose to leave it inside `__usm_or_buffer_accessor`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -478,12 +478,12 @@ using __repacked_tuple_t = typename __repacked_tuple<T>::type;
 template <typename _ContainerOrIterable>
 using __value_t = typename __internal::__memobj_traits<_ContainerOrIterable>::value_type;
 
-template <typename _T>
+template <typename _T, sycl::access_mode _AccessMode>
```
MikeDvorskiy, line 481: Actually, the spec tells:
"4.7.6.9.2. Deduction tags for buffer command accessors
Some accessor constructors take a TagT parameter, which is used to deduce template arguments...."
And the implementation before (line 495) used that tag value - `sycl::read_write`.
`__acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{}))`

If we want to pass information about required access mode outside, lets accept the same tag as an additional argument in `__usm_or_buffer_accessor` constructor and pass one further into `accessor` constructor, like this: 
 ```
 template <typename TagT>
__usm_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf, TagT __tag)
        : __acc(sycl::accessor(*__sycl_buf, __tag, __dpl_sycl::__no_init{}))
```
That's all.

In that case we don't need a template parameter `sycl::access_mode _AccessMode` and don't  need `__get_access_mode_tag()` function.

**Review by MikeDvorskiy — DISMISSED** [review]

**Comment by SergeyKopienko** [comment]
@danhoeflinger, @dmitriy-sobolev 
So in this PR we have as result two places with `<sycl::access_mode::read>` specification:
 - the first in the `__parallel_scan_submitter::operator()` - inside kernel for `__final_event`
 - the second in the `__parallel_transform_reduce_work_group_kernel_submitter::operator()` 

I believe we may have some performance boost at these places.
Could somebody check this point?

File: include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h (hunk 12 above)
Comment (MikeDvorskiy) [reply include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h:481]: Or we can use another accessor constructor w/o tag parameter. 
`__acc(sycl::accessor(*__sycl_buf, __dpl_sycl::__no_init{}))`
An access mode for accessor is defined via accessor type definition, via `sycl::access_mode _AccessMode`

Comment (SergeyKopienko) [reply include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h:481]: Good point, fixed.

File: include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h (hunk 11 above)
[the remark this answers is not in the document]
Comment (SergeyKopienko) [reply include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h]: This function and it's usage has been removed at all - please see details in the https://github.com/oneapi-src/oneDPL/pull/1909#discussion_r1808501820

File: include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h [hunk 13]
```diff
@@ -492,11 +492,11 @@ struct __usm_or_buffer_accessor
   public:
     // Buffer accessor
     __usm_or_buffer_accessor(sycl::handler& __cgh, sycl::buffer<_T, 1>* __sycl_buf)
-        : __acc(sycl::accessor(*__sycl_buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{}))
+        : __acc(*__sycl_buf, __cgh, __dpl_sycl::__no_init{})
```
danhoeflinger, line 495: This isn't a new bug from this PR, but I think that `__no_init` it is wrong to be using at all times here I think this is something that we need to specify each time and only when it is proper.
Well this is temporary memory it's often used to bridge between multiple kernels where the second kernel reads before it writes (if it does write). In this case `__no_init` is wrong I believe.

danhoeflinger:
> @danhoeflinger, @dmitriy-sobolev So in this PR we have as result two places with `<sycl::access_mode::read>` specification:
> 
> * the first in the `__parallel_scan_submitter::operator()` - inside kernel for `__final_event`
> * the second in the `__parallel_transform_reduce_work_group_kernel_submitter::operator()`
> 
> I believe we may have some performance boost at these places. Could somebody check this point?

I think in reality most of the target cases we are targeting for performance have access to USM and this will therefore be no change to performance. We could test it on some devices which don't support USM or force it into non-USM mode for the purposes of benchmarking but I'm unsure how useful that is.

danhoeflinger:
I do think that this can indirectly help with performance in our target cases by getting us closer to #1906. Switching the existing temporary storage that we have in buffers over to usm where it is available generally does provide a performance improvement. This PR improves the infrastructure to better enable using it elsewhere.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: Issue created: https://github.com/oneapi-src/oneDPL/issues/1914

SergeyKopienko: From https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_queue_interface,
Table 52. Properties supported by accessors
"This property is allowed only for accessors with access_mode::write or access_mode::read_write access modes. Attempting to construct an access_mode::read accessor with this property causes an exception with the errc::invalid error code to be thrown"

SergeyKopienko: @danhoeflinger for compatibility with the current code and based on documentation, I propose to set up `__dpl_sycl::__no_init{}` depends on used `_AccessMode` state.

danhoeflinger: I think we should just fix it here (even though it grows beyond the original scope of the PR).  I don't think we should create a new system here to propagate the bug only to later replace it.  I can help advise where `no_init` is and isn't needed.
For `read`, its certainly not supposed to be `no_init`. For `write`, it should be `no_init`. For `read_write`, it will depend how it is used (written before read?).  A good clue is if it is the first kernel using it or not. Generally subsequent kernels should not be `no_init`.

SergeyKopienko: Implemented, please check.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -453,8 +454,10 @@ struct __parallel_reduce_then_scan_scan_submitter<
             __dpl_sycl::__local_accessor<_InitValueType> __sub_group_partials(__num_sub_groups_local + 1, __cgh);
             __cgh.depends_on(__prior_event);
             oneapi::dpl::__ranges::__require_access(__cgh, __in_rng, __out_rng);
-            auto __temp_acc = __scratch_container.__get_scratch_acc(__cgh);
-            auto __res_acc = __scratch_container.__get_result_acc(__cgh);
+            auto __temp_acc = __scratch_container.template __get_scratch_acc<sycl::access_mode::read_write>(
+                __cgh, __dpl_sycl::__no_init{});
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: i think after the fix this looks good.  Thanks for adding this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -611,29 +616,31 @@ struct __result_and_scratch_storage
...
-            return __usm_or_buffer_accessor<_T>(__cgh, __result_buf.get());
+            return __usm_or_buffer_accessor<_T, _AccessMode>(__cgh, __result_buf.get(), __prop_list);
         else if (__supports_USM_device)
-            return __usm_or_buffer_accessor<_T>(__cgh, __scratch_buf.get(), __scratch_n);
-        return __usm_or_buffer_accessor<_T>(__cgh, __sycl_buf.get(), __scratch_n);
+            return __usm_or_buffer_accessor<_T, _AccessMode>(__cgh, __scratch_buf.get(), __scratch_n, __prop_list);
+        return __usm_or_buffer_accessor<_T, _AccessMode>(__cgh, __sycl_buf.get(), __scratch_n, __prop_list);
 #else
-        return sycl::accessor(*__sycl_buf.get(), __cgh, sycl::read_write, __dpl_sycl::__no_init{});
+        return sycl::accessor(*__sycl_buf.get(), __cgh, _AccessMode, __prop_list);
```
mmichel11, line 630: `_AccessMode` should be passed as a template parameter I believe. This path doesn't compile as there is no matching constructor.

The following should work:
```
 return sycl::accessor<_T, 1, _AccessMode>(*__sycl_buf.get(), __cgh, __prop_list);
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -611,29 +616,31 @@ struct __result_and_scratch_storage
...
     {
 #if _ONEDPL_SYCL_UNIFIED_USM_BUFFER_PRESENT
         if (__use_USM_host || __supports_USM_device)
-            return __usm_or_buffer_accessor<_T>(__cgh, __scratch_buf.get());
-        return __usm_or_buffer_accessor<_T>(__cgh, __sycl_buf.get());
+            return __usm_or_buffer_accessor<_T, _AccessMode>(__cgh, __scratch_buf.get(), __prop_list);
+        return __usm_or_buffer_accessor<_T, _AccessMode>(__cgh, __sycl_buf.get(), __prop_list);
 #else
-        return sycl::accessor(*__sycl_buf.get(), __cgh, sycl::read_write, __dpl_sycl::__no_init{});
+        return sycl::accessor(*__sycl_buf.get(), __cgh, _AccessMode, __prop_list);
```
mmichel11, line 643: Same comment as above regarding `_AccessMode`.

mmichel11 (commented):
I noticed a couple of compilation errors in the buffer fallback path. I think it would be good to manually test the following scenarios by temporarily modifying the macros / function returns:

1. The most common case: `_ONEDPL_SYCL_UNIFIED_USM_BUFFER_PRESENT=1`, `__use_USM_host_allocations=true`
2. USM device allocations but no host memory: `_ONEDPL_SYCL_UNIFIED_USM_BUFFER_PRESENT=1`, `__use_USM_host_allocations=false`. Testing on a CUDA / HIP device would also take this path.
3. The buffer fallback case: `_ONEDPL_SYCL_UNIFIED_USM_BUFFER_PRESENT=0`. This is where we would probably find all issues related to access modes.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: Alternatively, I think we could just switch to using the tags here.

SergeyKopienko: Done,
But is it enough?
Or we should write something like 
```C++
sycl::accessor<_T, 1, _AccessMode, __dpl_sycl::__target_device, sycl::access::placeholder::false_t>;
```
as it written at `struct __usm_or_buffer_accessor` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
But see my comment at https://github.com/oneapi-src/oneDPL/pull/1909/files/accfdc36a1403ff4f6a58b55928b1b21cc3801f8#r1809561553.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: Those additional parameters are defaulted to the same values that we explicitly list in `__usm_or_buffer_accessor` in the SYCL spec:
```
template <typename DataT, int Dimensions = 1,
          access_mode AccessMode =
              (std::is_const_v<DataT> ? access_mode::read
                                      : access_mode::read_write),
          target AccessTarget = target::device,
          access::placeholder isPlaceholder = access::placeholder::false_t>
class accessor { ... }
```
so the effect is the same. Although for consistency we should maybe use the same in both places.

danhoeflinger: Yes I think we should use the whole set like the other places.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: And now we have only one definition of `__accessor_t` in the `struct __result_and_scratch_storage`:
```C++
    template <sycl::access_mode _AccessMode>
    using __accessor_t =
        sycl::accessor<_T, 1, _AccessMode, __dpl_sycl::__target_device, sycl::access::placeholder::false_t>;
```

SergeyKopienko:
@danhoeflinger, @mmichel11 Could you please take a look again?

danhoeflinger (approved):
LGTM

mmichel11 (approved):
LGTM

PR #1911, opened 2024-10-17
[oneDPL][ranges][test] + test_range_algo::test_view functionality

MikeDvorskiy:
The PR introduces the checks for following standard view adapters:

- transform_view
- reverse_view
- take_view
- drop_view

NOTICE: std::ranges::views::all, std::ranges::subrange, std::span are tested implicitly within the 'test_range_algo' test engine.
single_view - is a trivial view and it doesn't  make sense for testing in the parallel code.

### DIFF
--- a/test/parallel_api/ranges/std_ranges_test_views.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_test_views.pass.cpp
@@ -1,0 +1,45 @@
+// -*- C++ -*-
+//===----------------------------------------------------------------------===//
+//
+// Copyright (C) Intel Corporation
+//
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+// This file incorporates work covered by the following copyright and permission
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#include "std_ranges_test.h"
+
+std::int32_t
+main()
+{
+#if _ENABLE_STD_RANGES_TESTING
+    using namespace test_std_ranges;
+    namespace dpl_ranges = oneapi::dpl::ranges;
+
+    const int n = 1<<25; //32M
+
+    //transform view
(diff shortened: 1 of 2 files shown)


test/parallel_api/ranges/std_ranges_test_views.pass.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include "std_ranges_test.h"
+
+std::int32_t
+main()
+{
+#if _ENABLE_STD_RANGES_TESTING
+    using namespace test_std_ranges;
+    namespace dpl_ranges = oneapi::dpl::ranges;
+
+    const int n = 1<<25; //32M
```
danhoeflinger, line 25: Does it make sense to test this out for more than just a single value of `n`?  I'm not sure I know the answer to this really.  We aren't testing the views themselves but rather the interoperability of the views with oneDPL algorithms.

Do we have a reason to suspect that different algorithms or even implementations in an algorithm may not handle views appropriately?  

Obviously we don't want to expand our testing to apply these views to all possible combinations of algorithm / `n`, but perhaps there is a way that would provide more coverage without a huge computation time cost.  At least we might want to select something other than `count_if` to not repeat it in this test.

danhoeflinger (approved):
This largely LGTM other than the open question on targeting of test coverage.

test/parallel_api/ranges/std_ranges_test_views.pass.cpp
MikeDvorskiy: @danhoeflinger, sorry, I am not sure that It understand your concern correctly...
A goal of this PR - to have an positive  implementation experience with the mentioned standard views in parallel processing. We don't want to test interoperability of the mentioned views with the all oneDPL algorithm, at least in this PR.

danhoeflinger: Are we not trying to test whether specific standard library views work in combination with the parallel algorithms within oneDPL (specifically that they work equivalently to when used with the serial equivalent)?

The question of coverage is basically the question of if these views may work on some oneDPL ranges algorithms but not others (due to a bug in oneDPL).  If that is possible, then we probably want to "touch" as many of the algorithms as we can to test them out.

PR #1913, opened 2024-10-18
Add missing device copyable specialization for `single_match_pred`

mmichel11:
`zip_iterator_find.pass` which has been recently added fails to compile on Windows. This is caused by a missing device copyable specialization for `oneapi::dpl::unseq_backend::single_match_pred`. As a result, `find_if` does not compile if a `std::tuple` member is in the user-provided predicate.

Adding the specialization resolved the compile error and the test now passes.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
@@ -332,6 +332,9 @@ template <typename _ExecutionPolicy, typename _Pred>
 struct single_match_pred_by_idx;
 
 template <typename _ExecutionPolicy, typename _Pred>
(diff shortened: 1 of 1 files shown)


danhoeflinger (commented):
The implementation looks good tentatively, but we need to add tests. Please look at `device_copyable.pass.cpp` and add some tests. We try to add them in both positive and negative direction.

mmichel11:
> The implementation looks good tentatively, but we need to add tests. Please look at `device_copyable.pass.cpp` and add some tests. We try to add them in both positive and negative direction.
> 
> Thanks for catching this one that was missed.

Thanks, I have added a test case for each direction.

danhoeflinger (approved):
LGTM, thanks!

PR #1915, opened 2024-10-22
Replace SYCL backend `reduce_by_segment` implementation with reduce-then-scan call

mmichel11:
## Summary
This PR implements a SYCL backend `reduce_by_segment` by using higher level calls to reduce-then-scan along with new specialty functors to achieve a segmented reduction. This PR is an initial step of porting the implementation to reduce-then-scan with optimization likely to follow. Future efforts may include additional modification to reduce-then-scan kernels.

Performance improves for all input sizes. For small inputs, we see 3-5x improvements and for very large sizes ~1.25x on GPU Series Max 1550. Please contact me if you would like to see performance data.

## Description of changes
- The SYCL `reduce_by_segment` implementation that was previously handwritten is replaced by a higher level call to our reduce-then-scan kernels. Several new callback functors for the reduce-then-scan kernel have been made to achieve this operation.
- `reduce_by_segment.pass` was encountering linker crashes due to the large number of test cases being compiled growing past the maximum size of the binary's data region. SYCL testing has been trimmed down with regards to USM device and shared testing which resolves this issue. Instead of running each test with a device and shared USM allocation, every other test switches the USM type.
- `ONEDPL_WORKAROUND_FOR_IGPU_64BIT_REDUCTION` has been removed as the SYCL implementation has been replaced, and we are no longer impacted by this issue.
- The legacy `reduce_by_segment` implementation is used as a fallback for when the sub-group size, device, and trivial copyability constraints cannot be satisfied.

## Future work
Future efforts on `reduce_by_segment` may built on top of this implementation and the reduce-then-scan kernels to better handle first and last element cases.


mmichel11:
I have made some design changes based on offline discussion. There is quite a bit of code movement that has happened, so here is a summary of the recently made changes:

- An iterator-based `__pattern_reduce_by_segment` has been added to `algorithm_impl_hetero.h`. Previously, we just had a range-based version. This resolves the issue of calling range-based patterns from iterator-based algorithms.
- The fallback `reduce_by_segment` implementation based on high-level `copy_if` and `parallel_for` calls has been moved down a level from `algorithm_ranges_impl_hetero.h` to `dpcpp/parallel_backend_sycl.h` so that we can fallback on this implementation when reduce-then-scan cannot be used in `__parallel_reduce_by_segment`. This pattern has been implemented synchronously as parallel pattern calls cannot currently depend on each other.
- Due to observed performance issues for compilers prior to icpx 2025.0, the reduce-then-scan path must be disabled in this case. The known-identity based implementation has been added back to avoid introducing a performance regression for `icpx 2024.2.1` and prior. It has been moved into `dpcpp/parallel_backend_sycl_reduce_by_segment.h` so the implementations are not split across several directories.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2124,6 +2275,180 @@ __parallel_partial_sort(oneapi::dpl::__internal::__device_backend_tag __backend_
...
+    if constexpr (oneapi::dpl::unseq_backend::__has_known_identity<_BinaryOperator, __val_type>::value)
+    {
+        return __parallel_reduce_by_segment_known_identity(
+            oneapi::dpl::__internal::__device_backend_tag{}, std::forward<_ExecutionPolicy>(__exec),
+            std::forward<_Range1>(__keys), std::forward<_Range2>(__values), std::forward<_Range3>(__out_keys),
+            std::forward<_Range4>(__out_values), __binary_pred, __binary_op);
+    }
+    else
+    {
+        // Round 1: reduce with extra indices added to avoid long segments
```
danhoeflinger, line 2345: This is perhaps just a preference thing for code organization, but perhaps we can name this implementation and put it in its own function, that way this function can just be the algorithm selection, and it may be more clear.

mmichel11: I think this is a good idea and will make this change.

mmichel11: I have made this change. The two fallback paths I have named to `__parallel_reduce_by_segment_fallback` with a true_type / false_type parameter for whether there is a known identity.

I do not like that one fallback path is in `parallel_backend_sycl.h` and the other is in `parallel_backend_sycl_reduce_by_segment.h`. However, I did this since I did not want to add both fallback paths into `parallel_backend_sycl_reduce_by_segment.h` as it would require forward declarations for the other parallel algorithms calls made, and I also did not want to add another 400+ lines to `parallel_backend_sycl.h`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_by_segment.h
```diff
@@ -0,0 +1,464 @@
...
+#include "../../../internal/scan_by_segment_impl.h"
+
+namespace oneapi
+{
+namespace dpl
+{
+namespace __par_backend_hetero
+{
+
+template <typename... Name>
```
danhoeflinger, line 68: I confirmed in an outside editor that the only changes to this from the previous location are only cosmetic.  I'm not looking deeply into this files changes otherwise since it has already been reviewed and it was just moved.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -786,6 +788,128 @@ struct __gen_transform_input
...
+        const std::size_t __new_seg_mask = __id > 0 && !__binary_pred(__in_keys[__id - 1], __in_keys[__id]);
+        return oneapi::dpl::__internal::make_tuple(__new_seg_mask, _ValueType{__in_vals[__id]});
+    }
+    _BinaryPred __binary_pred;
+};
+
+template <typename _BinaryPred>
+struct __gen_red_by_seg_scan_input
+{
+    // Returns the following tuple:
```
danhoeflinger, line 814: In other reduce_then_scan based API implementations, we are able to reuse the `..._reduce_input` generator inside the `..._scan_input`, but I can see why we may not want to do that here to consolidate branches.  It would be nice to rely on the same code to prevent redundancy and better show connection here.  I was thinking about ways to do it (perhaps with helper functions called by the call operators of each) but it may not be worth it.

I'll let you decide on this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -786,6 +788,128 @@ struct __gen_transform_input
...
+        }
+        else if (__id == __n - 1)
+        {
+            const _KeyType& __prev_key = __in_keys[__id - 1];
+            const std::size_t __new_seg_mask = !__binary_pred(__prev_key, __current_key);
+            return oneapi::dpl::__internal::make_tuple(
+                oneapi::dpl::__internal::make_tuple(__new_seg_mask, _ValueType{__in_vals[__id]}), true, __current_key,
+                __current_key); // Passing __current_key as the next key for the last element is a placeholder
+        }
+        else
```
danhoeflinger, line 848: add a comment here that this is to handle the `__id==0` case.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -786,6 +788,128 @@ struct __gen_transform_input
...
+        {
+            const _KeyType& __next_key = __in_keys[__id + 1];
+            return oneapi::dpl::__internal::make_tuple(
+                oneapi::dpl::__internal::make_tuple(std::size_t{0}, _ValueType{__in_vals[__id]}),
+                !__binary_pred(__current_key, __next_key), __next_key, __current_key);
+        }
```
danhoeflinger, line 854: It may be possible to avoid the `__id == 0` case, in a similar way to unique.  It is a little more complicated because we would need to set up the carry-in appropriately, but I think its possible and could provide some branch avoiding (and tuple shrinking) gains in the helpers.  
If you think its possible to do this, lets leave it as an issue to be explored in a follow up.

adamfidel (commented):
First pass of comments. I have looked at primarily the fallback algorithms and intend to focus on the reduce-then-scan implementation next.

test/parallel_api/numeric/numeric.ops/reduce_by_segment.pass.cpp
```diff
@@ -360,17 +358,17 @@ main()
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
 #if !_PSTL_RED_BY_SEG_WINDOWS_COMPILE_ORDER_BROKEN
-    run_test<MatrixPoint<float>, UserBinaryPredicate<MatrixPoint<float>>, MaxFunctor<MatrixPoint<float>>>();
+    run_test</*use_device_alloc=*/true, MatrixPoint<float>, UserBinaryPredicate<MatrixPoint<float>>, MaxFunctor<MatrixPoint<float>>>();
```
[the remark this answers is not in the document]
mmichel11: Yes, every test is alternated with testing USM shared vs. USM device memory instead of testing both. Prior to doing this, the data section of the executable grew far too large resulting in a linker error, " relocation truncated to fit...".

From looking the error message up, this could be avoided by using special compilation flags to use 64-bit relative addressing, but I think trimming the test down a bit makes the most sense.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -786,6 +788,128 @@ struct __gen_transform_input
...
+template <typename _BinaryOp>
+struct __red_by_seg_op
+{
+    template <typename _Tup1, typename _Tup2>
+    auto
+    operator()(const _Tup1& __lhs_tup, const _Tup2& __rhs_tup) const
+    {
+        using std::get;
+        // The left-hand side has processed elements from the same segment, so update the reduction value.
+        if (std::get<0>(__rhs_tup) == 0)
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -786,6 +788,128 @@ struct __gen_transform_input
...
+        const std::size_t __out_idx = get<0>(get<0>(__tup));
+
+        if (__id == 0)
+            __out_keys[0] = __current_key;
+        if (__is_seg_end)
+        {
+            __out_values[__out_idx] = __current_value;
+            if (__id != __n - 1)
+                __out_keys[__out_idx + 1] = __next_key;
+        }
```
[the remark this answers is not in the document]
mmichel11: Good point. I added a comment explaining that this is done to output the first key in each segment along with the special handling of the first and last element.

test/general/implementation_details/device_copyable.pass.cpp
```diff
@@ -155,6 +155,16 @@ test_device_copyable()
         sycl::is_device_copyable_v<oneapi::dpl::__par_backend_hetero::__gen_transform_input<noop_device_copyable>>,
         "__gen_transform_input is not device copyable with device copyable types");
 
+    //__gen_red_by_seg_reduce_input
+    static_assert(
+        sycl::is_device_copyable_v<oneapi::dpl::__par_backend_hetero::__gen_red_by_seg_reduce_input<noop_device_copyable>>,
```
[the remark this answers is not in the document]
mmichel11: I switched to use `binary_op_device_copyable`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I took a look at doing this. It is possible to add some helper utilities shared between the reduce / scan input generation, but I do not think it improves the simplifies the code too much unless it is written in a way that has some more conditional checks which I want to avoid.

With the small differences in each branch in `__gen_red_by_seg_scan_input`, I think it is clearest to leave it as is.

danhoeflinger (commented):
At a high level I agree with the changes in this PR, but there are still a few remaining nit picks outstanding.  

I have run out of time before my time off to get into the small details like sizes of types and forwarding of references, things like that.  The clang format suggestions can be ignored as of now.

So, I wont hit approve officially but I think this is very close and trust @adamfidel / others to be able to get it across the finish line and have no objections to merging with another approval.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
mmichel11: Thanks for the feedback. I have added some more explanations and a small example showing how these functors achieve a the reduce_by_segment operation.

test/general/implementation_details/device_copyable.pass.cpp
```diff
@@ -155,6 +155,16 @@ test_device_copyable()
...
         "__gen_transform_input is not device copyable with device copyable types");
 
+    //__gen_red_by_seg_reduce_input
+    static_assert(sycl::is_device_copyable_v<
+                      oneapi::dpl::__par_backend_hetero::__gen_red_by_seg_reduce_input<binary_op_device_copyable>>,
+                  "__gen_red_by_seg_reduce_input is not device copyable with device copyable types");
+
+    //__gen_red_by_seg_scan_input
+    static_assert(
+        sycl::is_device_copyable_v<oneapi::dpl::__par_backend_hetero::__gen_red_by_seg_scan_input<noop_device_copyable>>,
```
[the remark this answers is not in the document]
mmichel11: I have made this change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
danhoeflinger: Looks good.  Thanks!

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_by_segment.h
```diff
@@ -0,0 +1,464 @@
...
+
+                // 2e. Update local partial reductions in first segment and write to global memory.
+                bool __apply_aggs = true;
+                std::size_t __item_offset = 0;
+
+                // first item in group does not have any work-group aggregates to apply
+                if (__local_id == 0)
+                {
+                    __apply_aggs = false;
+                    if (__global_id == 0 && __n > 0)
```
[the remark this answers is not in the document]
mmichel11: Good point, it is not needed and I have removed it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_by_segment.h
```diff
@@ -0,0 +1,464 @@
...
+                        __loc_partials[__i - __start] = __accumulator;
+                        ++__item_segments;
+                        __max_end = __local_id;
+                        __accumulator = __identity;
+                    }
+                }
+
+                // 2c. Count the number of prior work segments cooperatively over group
+                std::size_t __prior_segs_in_wg = __dpl_sycl::__exclusive_scan_over_group(
+                    __group, __item_segments, __dpl_sycl::__plus<decltype(__item_segments)>());
```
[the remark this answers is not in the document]
mmichel11: I changed this occurrence and a few lines below in the inclusive scan call as well which was a similar case.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_by_segment.h
```diff
@@ -0,0 +1,464 @@
...
+#endif
+        __cgh.parallel_for<_SegReduceCountKernel>(
+            sycl::nd_range<1>{__n_groups * __wgroup_size, __wgroup_size}, [=](
+#if _ONEDPL_COMPILE_KERNEL && !_ONEDPL_KERNEL_BUNDLE_PRESENT
+                                                                              __seg_reduce_count_kernel,
+#endif
+                                                                              sycl::nd_item<1> __item) {
+                auto __group = __item.get_group();
+                std::size_t __group_id = __item.get_group(0);
+                std::size_t __local_id = __item.get_local_id(0);
```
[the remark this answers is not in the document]
mmichel11: I changed this type to be `std::uin32_t`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -32,12 +32,14 @@
...
 #include "../../execution_impl.h"
 #include "../../utils_ranges.h"
+#include "../../ranges_defs.h"
 
 #include "sycl_defs.h"
 #include "parallel_backend_sycl_utils.h"
 #include "parallel_backend_sycl_reduce.h"
 #include "parallel_backend_sycl_merge.h"
 #include "parallel_backend_sycl_merge_sort.h"
+#include "parallel_backend_sycl_reduce_by_segment.h"
```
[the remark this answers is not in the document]
mmichel11: I took another look at adding all of the implementations to the reduce-by-segment header. It works, but there are a few issues that I think make it cleaner to leave it as is:

- `__parallel_copy_if` has default template / function arguments and providing a forward declaration in `parallel_backend_reduce_by_segment.h` will require having these defaults here and not in `parallel_backend_sycl.h`. This is confusing for code readability in my opinion as the defaults will be in this new header that is unrelated to copy_if itself.
- `__get_zeroth_element` requires a full definition of the struct, so we would need to define a similar struct or move this callback somewhere else. Currently, all of the reduce_then_scan callback functors are in `parallel_backend_sycl.h`
- We have an additional copyright on the fallback implementation currently in the new header. Having only this implementation hear clearly indicates where the copyright applies.

If there are no objections, I think the current approach is best.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I agree and have opened https://github.com/oneapi-src/oneDPL/issues/1958.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -786,6 +788,181 @@ struct __gen_transform_input
...
+        using _ValueType = oneapi::dpl::__internal::__value_t<decltype(__in_vals)>;
+        const _KeyType& __current_key = __in_keys[__id];
+        // Ordering the most common condition first has yielded the best results.
+        if (__id > 0 && __id < __n - 1)
+        {
+            const _KeyType& __prev_key = __in_keys[__id - 1];
+            const _KeyType& __next_key = __in_keys[__id + 1];
+            const std::size_t __new_seg_mask = !__binary_pred(__prev_key, __current_key);
+            return oneapi::dpl::__internal::make_tuple(
+                oneapi::dpl::__internal::make_tuple(__new_seg_mask, _ValueType{__in_vals[__id]}),
```
adamfidel, line 840: Since the expression `_ValueType{__in_vals[__id]}` appears in every branch, would it make sense to hoist the copy before the branches and `std::move` the value into each tuple?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2124,6 +2333,189 @@ __parallel_partial_sort(oneapi::dpl::__internal::__device_backend_tag __backend_
...
                                         __partial_merge_kernel<decltype(__mid_idx)>{__mid_idx}, __comp);
 }
+
+//------------------------------------------------------------------------
+// reduce_by_segment - sync pattern
+//------------------------------------------------------------------------
+
+// TODO: The non-identity fallback path of reduce-by-segment must currently be implemented synchronously due to the
+// inability to create event dependency chains across separate parallel pattern calls. If we ever add support for
+// cross parallel pattern dependencies, then we can implement this as an async pattern.
```
adamfidel, line 2343:
```suggestion
//------------------------------------------------------------------------
// reduce_by_segment - sync pattern
//
// TODO: The non-identity fallback path of reduce-by-segment must currently be implemented synchronously due to the
// inability to create event dependency chains across separate parallel pattern calls. If we ever add support for
// cross parallel pattern dependencies, then we can implement this as an async pattern.
//------------------------------------------------------------------------
```

Suggest that we move the comment about the synchronousity of the fallback up because its current placements looks like it's commenting on `__reduce1_wrapper` instead of the pattern as a whole.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -786,6 +788,181 @@ struct __gen_transform_input
...
+        // With the exception of the first key which is output by index 0, the first key in each segment is written
+        // by the work item that outputs the previous segment's reduction value. This is because the reduce_by_segment
+        // API requires that the first key in a segment is output and is important for when keys in a segment might not
+        // be the same (but satisfy the predicate). The last segment does not output a key as there are no future
+        // segments process.
+        if (__id == 0)
+            __out_keys[0] = __current_key;
+        if (__is_seg_end)
+        {
+            __out_values[__out_idx] = __current_value;
```
adamfidel, line 957: This looks to be assigning a reference from a tuple. I know we have hit snags in the past with references and tuples. Is this safe to do in this instance or do we need to cast to the value type (i.e., `_ValType{__current_value}`)?

adamfidel (commented):
A few more comments. 

I think overall the approach is good and I don't mind approving once these few small comments have been addressed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I made this change by adding a new variable. Instead of copying and moving it into the tuple I did what is being done with the keys and pass it as a const reference instead.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I think `__current_value` is already the correct type without references. This is because its type, `_ValType` is `using _ValType = oneapi::dpl::__internal::__value_t<decltype(__out_values)>;` and the references should already be removed.

Does this make sense?

adamfidel (approved):
LGTM overall. The `clang-format` changes can be ignored in my opinion.

mmichel11:
I have fixed a compilation issue identified in CI. With some configurations, specifying the full namespace `oneapi::dpl::__ranges` when calling `__get_sycl_range` is needed to avoid compilation errors.

I am rerunning CI to make sure there are no more issues.

adamfidel (approved):
LGTM after the recent changes to fix issues discovered in CI.

danhoeflinger (approved):
LGTM after minor changes.  Thanks and sorry for the delay

PR #1918, opened 2024-10-23
Fix test_policies for FPGA

dmitriy-sobolev:
A straightforward fix would be to add missing `>`. However, testing the `fpga_policy` as it is looks a better choice due to being a basic case. I guess the change in #1907 with wrapping  `fpga_policy` into `device_policy` was a typo.


adamfidel (approved):
Good catch. The addition of the `device_policy` was indeed a typo. Thanks!

PR #1920, opened 2024-10-23
[oneDPL][ranges][tests] a fix for std::ranges::<algo> using with Windows C++ standard library.

MikeDvorskiy:
[oneDPL][ranges][tests] a fix for std::ranges::sort using with Windows C++ standard library.
This PR is a sequel for https://github.com/oneapi-src/oneDPL/pull/1885.

### DIFF
--- a/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
+++ b/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
@@ -424,10 +424,9 @@ struct __stable_sort_fn
     operator()(_ExecutionPolicy&& __exec, _R&& __r, _Comp __comp = {}, _Proj __proj = {}) const
     {
         const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
-        return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(__dispatch_tag,
-            std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
-            [](auto&&... __args) { return std::ranges::stable_sort(std::forward<decltype(__args)>(__args)...); }
-            );
+        return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(
+            __dispatch_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
+            [](auto&&... __args) { return std::ranges::stable_sort(std::forward<decltype(__args)>(__args)...); });
     }
 }; //__stable_sort_fn
 }  //__internal
@@ -446,10 +445,9 @@ struct __sort_fn
     operator()(_ExecutionPolicy&& __exec, _R&& __r, _Comp __comp = {}, _Proj __proj = {}) const
     {
         const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
-        return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(__dispatch_tag,
-            std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
-            [](auto&&... __args) { return std::ranges::sort(std::forward<decltype(__args)>(__args)...); }
-            );
+        return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(
+            __dispatch_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
+            [](auto&&... __args) { return std::ranges::sort(std::forward<decltype(__args)>(__args)...); });
     }
 }; //__sort_fn
 }  //__internal
--- a/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
+++ b/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
@@ -426,7 +426,7 @@ struct __stable_sort_fn
         const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
         return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(__dispatch_tag,
             std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
-            _ONEDPL_PREPARE_CALLABLE(std::ranges::stable_sort)
+            [](auto&&... __args) { return std::ranges::stable_sort(std::forward<decltype(__args)>(__args)...); }
             );
     }
 }; //__stable_sort_fn
@@ -448,7 +448,7 @@ struct __sort_fn
         const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
         return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(__dispatch_tag,
             std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
-            _ONEDPL_PREPARE_CALLABLE(std::ranges::sort)
+            [](auto&&... __args) { return std::ranges::sort(std::forward<decltype(__args)>(__args)...); }
             );
     }
 }; //__sort_fn
--- a/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
+++ b/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
@@ -426,11 +426,7 @@ struct __stable_sort_fn
         const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
         return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(__dispatch_tag,
             std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
-#if _ONEDPL_STD_RANGES_ALGO_CPP_FUN
-            [](auto&&... __args) { return std::ranges::stable_sort(std::forward<decltype(__args)>(__args)...); }
-#else
-            std::ranges::stable_sort
-#endif
+            _ONEDPL_ALGO(std::ranges::stable_sort)
             );
     }
 }; //__stable_sort_fn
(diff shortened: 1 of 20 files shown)


test/parallel_api/ranges/std_ranges_sort.pass.cpp
```diff
@@ -21,18 +21,24 @@ main()
 #if _ENABLE_STD_RANGES_TESTING
     using namespace test_std_ranges;
     namespace dpl_ranges = oneapi::dpl::ranges;
+    auto sort_checker = 
+#if _ONEDPL_STD_RANGES_ALGO_CPP_FUN
```
dmitriy-sobolev, line 25: Could you introduce a similar macro in the tests in order not to rely on oneDPL internals?

MikeDvorskiy: I still strong disagree to code duplication.
Rather lets consider a an option to make public that macro....

dmitriy-sobolev: I do not think that having a public macro for that is a good idea given that it is a workaround for a niche issue, not resulting in enablement of any oneDPL feature. 

I guess we need a mediator here.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -333,4 +333,11 @@
 #    define _ONEDPL_STD_RANGES_ALGO_CPP_FUN 0
 #endif
 
+#if _ONEDPL_STD_RANGES_ALGO_CPP_FUN
+#   define _ONEDPL_ALGO(std_algo_name) \
+[](auto&&... __args) { return std_algo_name(std::forward<decltype(__args)>(__args)...); }
```
dmitriy-sobolev, line 338: Will not it be better to move that utility in a separate header, e.g. `utils.h`? 
A "config" usually contains well some settings, e.g. to enable workarounds, rather than workarounds themselves. Also, having a macro (which are generelly avoided) looks unnecessary to resolve that particular issue.

MikeDvorskiy: To tell the truth, practically, many macro wrappers were accumulated in `onedpl_config.h`. F.e. a lot of _ONEDPL_PRAGMA_SIMD wrappers.
`utils.h`, traditionally, contains C/C++ functons/utilities.

dmitriy-sobolev: I  implied creation of a function instead of a macro in `utils.h`, but perhaps it is a matter of taste, so I do not insist.

MikeDvorskiy: No, it is not "it is a matter of taste"...
Of course I prefer function too... But there is no way to write a function here.

dmitriy-sobolev: But it is possible to write a function in `utils.h`, if I understood you correctly.

dmitriy-sobolev (commented):
The PR solves the problem, but I cannot decide whether to approve it or not due to the approach (using internal macro in the tests, using a macro, where it is possible to have an ordinary function). Let's add more reviewers for more opinions.

Update regarding "it is possible to have an ordinary function": I was not able to find such a solution besides providing specializations for signatures of specific algorithms (similar to the existing "checker" entities in the tests).

include/oneapi/dpl/pstl/onedpl_config.h
danhoeflinger: I think I agree that this is more of a "utility" than a "configuration".  I understand the motivation to group preprocessor macros together, and utility functions together, but the concept seems more important to me.  
I think I would be in favor of keeping `_ONEDPL_STD_RANGES_ALGO_CPP_FUN` definition here, and moving the rest to "utils.h", but I also wont insist.

test/parallel_api/ranges/std_ranges_adjacent_find.pass.cpp
```diff
@@ -22,10 +22,12 @@ main()
     using namespace test_std_ranges;
     namespace dpl_ranges = oneapi::dpl::ranges;
 
-    test_range_algo<0>{big_sz}(dpl_ranges::adjacent_find, std::ranges::adjacent_find, binary_pred);
-    test_range_algo<1>{}(dpl_ranges::adjacent_find, std::ranges::adjacent_find, binary_pred, proj);
-    test_range_algo<2, P2>{}(dpl_ranges::adjacent_find, std::ranges::adjacent_find, binary_pred, &P2::x);
-    test_range_algo<3, P2>{}(dpl_ranges::adjacent_find, std::ranges::adjacent_find, binary_pred, &P2::proj);
+    auto adj_find_checker = _ONEDPL_ALGO(std::ranges::adjacent_find);
```
danhoeflinger, line 25: In the past, we have tried to stay away from using internal helpers / utilities / implementation details in our tests.  

If we want to use the same macro here, I think we probably want to re-define it here `test/support/utils.h`, or somewhere similar as `_TEST_PREPARE_CALLABLE`, or something like that, then use that throughout the tests.

test/parallel_api/ranges/std_ranges_sort.pass.cpp
danhoeflinger: Relocating my comment here, I missed this discussion in my initial review.  I agree with Dmitriy here, this has been the convention as established by @akukanov.

We have tried to stay away from using internal helpers / utilities / implementation details in our tests.  

If we want to use the same macro here, I think we probably want to re-define it here `test/support/utils.h`, or somewhere similar as `_TEST_PREPARE_CALLABLE`, or something like that, then use that throughout the tests.

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -333,4 +333,11 @@
 #    define _ONEDPL_STD_RANGES_ALGO_CPP_FUN 0
 #endif
 
+#if _ONEDPL_STD_RANGES_ALGO_CPP_FUN
+#   define _ONEDPL_ALGO(std_algo_name) \
```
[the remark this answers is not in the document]
MikeDvorskiy: I don't mind `_ONEDPL_PREPARE_CALLABLE` macro name

include/oneapi/dpl/pstl/onedpl_config.h
MikeDvorskiy: I don't mind

dmitriy-sobolev: After taking a closer look, and I cannot come up with a solution without macros. Given that, I agree keeping it in `onedpl_config.h`.

danhoeflinger: I'm OK with whatever you decide.

test/parallel_api/ranges/std_ranges_sort.pass.cpp
akukanov: In the essence, here we have two good engineering practices contradicting each other, and we need to choose between "avoid code duplication" and "avoid test dependencies on implementation internals". I choose the second, because most if not all code duplications between the implementation and its tests are simple self-contained utilities or macros and duplicating those does not seem problematic to me.

Moreover, strictly speaking, the tests in this patch do not even have to use `_ONEDPL_PREPARE_CALLABLE` - range algorithms can always be wrapped into a lambda, possibly with a small "syntax sugar" helper.

MikeDvorskiy: I see..
But we missed a way with passing a callable function by reference - `foo(f&& f)`  - https://godbolt.org/z/j1P6MKdvn
Is the implementation of a C++ function/algorithm with doesn't match to `foo(f&& f)`?

include/oneapi/dpl/pstl/onedpl_config.h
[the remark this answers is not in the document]
MikeDvorskiy: Update: currently  - it is only in the tests: `TEST_PREPARE_CALLABLE`

test/parallel_api/ranges/std_ranges_adjacent_find.pass.cpp
```diff
@@ -22,10 +22,12 @@ main()
     using namespace test_std_ranges;
     namespace dpl_ranges = oneapi::dpl::ranges;
 
-    test_range_algo<0>{big_sz}(dpl_ranges::adjacent_find, std::ranges::adjacent_find, binary_pred);
-    test_range_algo<1>{}(dpl_ranges::adjacent_find, std::ranges::adjacent_find, binary_pred, proj);
-    test_range_algo<2, P2>{}(dpl_ranges::adjacent_find, std::ranges::adjacent_find, binary_pred, &P2::x);
-    test_range_algo<3, P2>{}(dpl_ranges::adjacent_find, std::ranges::adjacent_find, binary_pred, &P2::proj);
+    auto adj_find_checker = TEST_PREPARE_CALLABLE(std::ranges::adjacent_find);
```
dmitriy-sobolev, line 25: Could you include `test_macros.h` for `TEST_PREPARE_CALLABLE` (here and in other tests)?

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -452,11 +448,7 @@ struct __sort_fn
         const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
         return oneapi::dpl::__internal::__ranges::__pattern_sort_ranges(__dispatch_tag,
             std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r), __comp, __proj,
-#if _ONEDPL_STD_RANGES_ALGO_CPP_FUN
-            [](auto&&... __args) { std::ranges::sort(std::forward<decltype(__args)>(__args)...); }
-#else
-            std::ranges::sort
-#endif
+            [](auto&&... __args) { return std::ranges::sort(std::forward<decltype(__args)>(__args)...); }
```
dmitriy-sobolev, line 451: Could you apply clang-format? [clang-format.patch](https://github.com/user-attachments/files/17798400/clang-format.patch)

dmitriy-sobolev (commented):
The PR looks good to me leaving aside a couple of issues easy to fix.

dmitriy-sobolev (approved):

PR #1921, opened 2024-10-24
[test] Reduce test input sizes for debug builds with OpenMP in `transform_binary.pass`

mmichel11:
`transform_binary.pass` with the OpenMP backend in debug builds has been timing out. This is caused by testing of forward and bidirectional iterators. When no compiler optimization is present, a large number of re-computation of iterators occurs in this line: https://github.com/oneapi-src/oneDPL/blob/66ead74ff6db1eb7e047227086da836531499a74/include/oneapi/dpl/pstl/omp/parallel_for_each.h#L40 which has an expensive cost for iterators without random access. In release builds, there is no problem due to compiler optimization.

The test is updated to reduce test sizes in this timeout case. I have also updated our `CMakeLists.txt` to set the `PSTL_USE_DEBUG` macro for OpenMP builds as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -180,7 +180,7 @@ if (ONEDPL_BACKEND MATCHES "^(tbb|dpcpp|dpcpp_only)$")
 
     target_compile_definitions(oneDPL INTERFACE
         $<$<CONFIG:Debug>:TBB_USE_DEBUG=1>
-        $<$<CONFIG:Debug>:PSTL_USE_DEBUG>
+        $<$<CONFIG:Debug>:PSTL_USE_DEBUG=1>
         $<$<BOOL:${SET_BACKEND_DPCPP_ONLY}>:ONEDPL_USE_TBB_BACKEND=0>
         $<$<BOOL:${SET_BACKEND_TBB}>:ONEDPL_USE_DPCPP_BACKEND=0>
         )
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -180,7 +180,7 @@ if (ONEDPL_BACKEND MATCHES "^(tbb|dpcpp|dpcpp_only)$")
 
     target_compile_definitions(oneDPL INTERFACE
         $<$<CONFIG:Debug>:TBB_USE_DEBUG=1>
-        $<$<CONFIG:Debug>:PSTL_USE_DEBUG>
+        $<$<CONFIG:Debug>:PSTL_USE_DEBUG=1>
```
dmitriy-sobolev, line 183: `$<$<CONFIG:Debug>:PSTL_USE_DEBUG=1>` should not depend on a backend type. Let's move it outside of the backend-specific branches to apply everywhere.

dmitriy-sobolev (commented):
Let me leave a couple of unsignificant comments. Overall, the PR looks good to me.

test/parallel_api/algorithm/alg.modifying.operations/transform_binary.pass.cpp
```diff
@@ -98,7 +98,15 @@ template <::std::size_t CallNumber, typename In1, typename In2, typename Out, ty
 void
 test(Predicate pred, _IteratorAdapter adap = {})
 {
-    for (size_t n = 0; n <= 100000; n = n <= 16 ? n + 1 : size_t(3.1415 * n))
+    // Testing is restricted for debug build + OpenMp backend as without optimization the compiler generates
```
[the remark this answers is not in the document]
mmichel11: Fixed

CMakeLists.txt
mmichel11: I agree and have moved it outside of backend specific branches.

dmitriy-sobolev (approved):
LGTM

PR #1930, opened 2024-11-01
RFC for histogram CPU implementation

danhoeflinger:
Adds an RFC for histogram CPU implementation.


rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,73 @@
...
+### parallel_for
+`parallel_for` is an interesting pattern in that it is very generic and embarrassingly parallel. This is close to what we need for `histogram`. However, we cannot simply use it without any added infrastructure. If we were to just use `parallel_for` alone, there would be a race condition between threads when incrementing the values in the output histogram. We should be able to use `parallel_for` as a building block for our implementation, but it requires some way to synchronize and accumulate between threads.
+
+## Proposal
+I believe there are two competing options for `histogram`, which may both have utility in the final implementation depending on the use case.
+
+### Implementation One (Embarrassingly Parallel)
+This method uses temporary storage and a pair of embarrassingly parallel `parallel_for` loops to accomplish the `histogram`.
+1) Determine the number of threads that we will use, perhaps adding some method to do this generically based on the backend.
+2) Create temporary data for the number of threads minus one copy of the histogram output sequence. Thread zero can use the user-provided output data.
```
akukanov, line 44: Determining/assuming a certain number of threads is not necessary, and should be avoided for the sake of parallel composability. We need a sort of thread-local storage, maybe the one that is accessible by other threads for data aggregation. It certainly can be done without knowing the number of threads.

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Introduction
+In version 2022.6.0, two `histogram` APIs were added to oneDPL, but implementations were only provided for device policies with the dpcpp backend. `Histogram` was added to the oneAPI specification 1.4 provisional release and should be present in the 1.4 specification. Please see the [oneAPI Specification](https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneDPL/source/parallel_api/algorithms.rst#parallel-algorithms) for a full definition of the semantics of the histogram APIs. In short, they take elements from an input sequence and classify them into either evenly distributed or user-defined bins via a list of separating values and count the number of values in each bin, writing to a user-provided output histogram sequence. Currently, `histogram` is not supported with serial, tbb, or openmp backends in our oneDPL implementation. This RFC aims to propose the implementation of `histogram` for these host-side backends. The serial implementation is straightforward and is not worth discussing in much length here. We will add it, but there is not much to discuss within the RFC, as its implementation will be straightforward.
```
akukanov, line 4: Please apply our general rule about the line width for the markdown files as well.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: There is some precedent for utilizing the maximum number of threads for accumulation in the OpenMP backend with `transform_reduce` [(link)](https://github.com/oneapi-src/oneDPL/blob/bdc037b9335cc18e29c935cef75204272e5d6341/include/oneapi/dpl/pstl/omp/parallel_transform_reduce.h#L42).  This is of course not to say it is correct or best.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: sure, done.

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Implementation Two (Atomics)
+This method uses atomic operations to remove the race conditions during accumulation. With atomic increments of the output histogram data, we can merely run a `parallel_for` pattern.
+
+To deal with atomics appropriately, we have some limitations. We must either use standard library atomics, atomics specific to a backend, or custom atomics specific to a compiler. `C++17` provides `std::atomic<T>`, however, this can only provide atomicity for data which is created with atomics in mind. This means allocating temporary data and then copying it to the output data. `C++20` provides `std::atomic_ref<T>` which would allow us to wrap user-provided output data in an atomic wrapper, but we cannot assume `C++17` for all users. We could look to implement our own `atomic_ref<T>` for C++17, but that would require specialization for individual compilers. OpenMP provides atomic operations, but that is only available for the OpenMP backend.
```
akukanov: > `C++17` provides `std::atomic<T>`, however, this can only provide atomicity for data which is created with atomics in mind. This means allocating temporary data and then copying it to the output data.

I guess you mean that if the output data type is not `std::atomic`, in C++17 there is no standard way to do atomic operations with these data. 

>  We could look to implement our own `atomic_ref<T>` for C++17, but that would require specialization for individual compilers.

I think we do not need a full-fledged `atomic_ref`, we need something like "atomic_cast", and - given that the output data are counters of integral type - that should not be a huge deal.

danhoeflinger: Yes, you are correct. I haven't done a deep look into all compilers in our matrix at this point.  It may be as simple as a reinterpret cast in most cases, but that isn't guaranteed by the standard to work.  I agree this is worth looking at, and may not be so bad.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: On the OpenMP side, perhaps we shouldn't be using `parallel_for` but rather extending some of its building blocks for use by `histogram` and `parallel_for`, and with more control over `__chunk_partitioner` and the parallel tasks / loop is what we need, and where we can understand our number of local temporary copies / threads which will use them.

For TBB, I'm more in the dark since that is more recursively decomposing into chunks , but looking further...

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,73 @@
...
+
+## Open Questions
+* Would it be worthwhile to add our own implementation of `atomic_ref` for C++17? I believe this would require specializations for each of our supported compilers.
+
+* What is the overhead of atomics in general in this case and does the overhead there make them inherently worse than merely having extra copies of the histogram and accumulating?
+
+* Is it worthwhile to have separate implementations for TBB and OpenMP because they may differ in the best-performing implementation? What is the best heuristic for selecting between algorithms (if one is not the clear winner)?
+
+* How will vectorized bricks perform, and in what situations will it be advantageous to use or not use vector instructions?
```
akukanov: A couple of papers that might be of interest:
- https://www.researchgate.net/publication/221131224_SIMD_Vectorization_of_Histogram_Functions
- https://www.researchgate.net/publication/266660552_Versatile_and_scalable_parallel_histogram_construction

danhoeflinger: Thanks, these are interesting, and mainly follow similar lines to the approaches outlined here.  I will look more into the `pcwar` SIMD instruction to understand if that is viable to use here.

danhoeflinger: While its interesting to explore, I believe we basically depend on OpenMP simd to provide our SIMD operations.  We won't be able to serve more complex intrinsic operations if we want to stick to that.

Further, I'm not actually finding anything like the pcwar instruction(s) they refer to here in https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html.  I do find some for conflict detection across lane which _could_ be useful, but again those won't be available through the interface of OpenMP.

I think our options are basically to decline SIMD or to have duplicates of the output for each SIMD lane. I think even that may be more "hands on" with SIMD details than we have done thus far from oneDPL.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: For TBB, does it make sense to do something similar to [this code](https://github.com/oneapi-src/oneDPL/blob/bdc037b9335cc18e29c935cef75204272e5d6341/include/oneapi/dpl/pstl/parallel_backend_tbb.h#L359-L366)?

The plan would be to Isolate then get the max concurrency of the task arena and allocating temporaries based on that. Then using `current_thread_index()` within a function brick which is decomposed directly or via a `parallel_for` pattern followed by a launch of tasks to accumulate those into the output.

It seems like this is not thinking very "tbb-like", because we would be interacting directly with thread indices and number of threads, which is discouraged.  However, I this case I think its important to control the allocation while not sacrificing the ability to use a small grain size for work stealing and load balance.  

Is there some more tbb-way of doing this which would have a similar control on footprint?

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,143 @@
...
+`histogram`.
+1) Determine the number of threads that we will use, perhaps adding some method to do this generically based on the
+2) backend.
+3) Create temporary data for the number of threads minus one copy of the histogram output sequence. Thread zero can
+4) use the user-provided output data.
+5) Run a `parallel_for` pattern which performs a `histogram` on the input sequence where each thread accumulates into
+6) its own copy of the output sequence using the temporary storage to remove any race conditions.
+7) Run a second `parallel_for` over the `histogram` output sequence which accumulates all temporary copies of the
+8) histogram into the output histogram sequence. This step is also embarrassingly parallel.
+9) Deallocate temporary storage.
```
mmichel11, line 92: I think the changes to the line widths interfered with the numbering here.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: Aha, this is what I think I was missing... 
https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneTBB/source/thread_local_storage/enumerable_thread_specific_cls.rst

It may make it difficult to understand where the appropriate heuristic for choosing between algorithms, but I think this is the "tbb way" of doing this which I was missing.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: fixed, thanks

rfcs/proposed/host_backend_histogram/README.md
akukanov: Yes, in TBB there is `enumerable_thread_specific`. But even without TBB, something like this could be done with e.g. a hash map protected by a lock. A thread ID/native handle would be the key to the map; and if it is guaranteed that insertions do not invalidate iterators/references to the map nodes, the lock would only need to be taken for getting such a reference, which could then be used to update the content without data races.

I am not sure if it's better to have a generic implementation like I outlined above, or if a special "interface" with backend-specific implementations is necessary for performance reasons.

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,150 @@
...
+policies with the dpcpp backend. `Histogram` was added to the oneAPI specification 1.4 provisional release and should
+be present in the 1.4 specification. Please see the
+[oneAPI Specification](https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneDPL/source/parallel_api/algorithms.rst#parallel-algorithms)
+for a full definition of the semantics of the histogram APIs. In short, they take elements from an input sequence and
+classify them into either evenly distributed or user-defined bins via a list of separating values and count the number
+of values in each bin, writing to a user-provided output histogram sequence. Currently, `histogram` is not supported
+with serial, tbb, or openmp backends in our oneDPL implementation. This RFC aims to propose the implementation of
+`histogram` for these host-side backends. The serial implementation is straightforward and is not worth discussing in
+much length here. We will add it, but there is not much to discuss within the RFC, as its implementation will be
+straightforward.
```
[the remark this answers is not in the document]
danhoeflinger: For the first point, its true, our approach to SIMD is the key here.  I'll put some more thought into this.  Right now, the only option I see for making use of SIMD is to make more replicates of the temporary storage available.  I think that it is somewhat unique within oneDPL to have such a temporary allocation requirement to use a vector brick. Otherwise, it really maps poorly to these instructions.  

On the second point, my plan is to just support the serial backend.  

I can expand upon this topic a bit in the RFC after some thought, and maybe some experimentation. I'm working through a first draft implementation, which I think is necessary for me to understand all the infrastructure details in full here.

akukanov:
Overall, this all sounds good enough for the "proposed" stage, where it's expected that some details are unknown and need to be determined. I am happy to approve it but will wait for a few days in case @danhoeflinger wants to update the document with some follow-up thoughts on the discussion.

rfcs/proposed/host_backend_histogram/README.md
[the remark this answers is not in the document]
danhoeflinger: I added a small section about serial backend support, and removed the section in the intro.  The questions about SIMD are still open, and will require some experimentation.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: I added a little about my proposed strategy for handling atomics.  I think the best option is to only worry about what we need which is explicitly an atomic increment, and do this using in the best method available.  This allows us the simplest path forward.

danhoeflinger:
> Overall, this all sounds good enough for the "proposed" stage, where it's expected that some details are unknown and need to be determined. I am happy to approve it but will wait for a few days in case @danhoeflinger wants to update the document with some follow-up thoughts on the discussion.

I've made a few more minor adjustments, but I think I'm happy with the state of the RFC now.  The question I have is what is the next best step once we merge.  I will be working on the rough implementation (already in progress), but should I also open a new draft PR moving this from proposed to supported as a location for further discussion? Or should I merely open PRs in-place if there are significant changes motivated by the experience in implementation?

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -110,9 +112,11 @@ To deal with atomics appropriately, we have some limitations. We must either use
 specific to a backend, or custom atomics specific to a compiler. `C++17` provides `std::atomic<T>`, however, this can
 only provide atomicity for data which is created with atomics in mind. This means allocating temporary data and then
 copying it to the output data. `C++20` provides `std::atomic_ref<T>` which would allow us to wrap user-provided output
-data in an atomic wrapper, but we cannot assume `C++17` for all users. We could look to implement our own
-`atomic_ref<T>` for C++17, but that would require specialization for individual compilers. OpenMP provides atomic
-operations, but that is only available for the OpenMP backend.
+data in an atomic wrapper, but we cannot assume `C++17` for all users. OpenMP provides atomic
```
akukanov, line 115: Do you mean "C++20" here?

danhoeflinger: Yes I do, thank you.

akukanov (dismissed):
I approve it for the implementation as a supported functionality, with the following understanding of the proposed design:
- For multithreaded execution with `par` and `par_unseq`, the implementation uses the existing `parallel_for` backend pattern.
- Unsequenced policies do not enable vectorization when thread-local histograms are built (the 1st stage of the algorithm) but do apply for combining those histograms into the final result (the 2nd stage).

The following questions I believe will need to be answered:
- Which extensions are needed in the parallel backend API to facilitate the use of enumerable storage for per-thread temporary buffers?
- Will the serial backend also allocate a temporary buffer, or will it directly accumulate into the output sequence and run just the 1st stage? If the latter - how this difference between backends will be handled on the pattern level?
- Should the 2nd stage always run according to the given execution policy, or should it under certain conditions better run serially by the calling thread?

Specification changes are not required.

I would recommend @MikeDvorskiy as the second approver.

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,165 @@
...
+In version 2022.6.0, two `histogram` APIs were added to oneDPL, but implementations were only provided for device
+policies with the dpcpp backend. `Histogram` was added to the oneAPI specification 1.4 provisional release and should
+be present in the 1.4 specification. Please see the
+[oneAPI Specification](https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneDPL/source/parallel_api/algorithms.rst#parallel-algorithms)
+for a full definition of the semantics of the histogram APIs. In short, they take elements from an input sequence and
+classify them into either evenly distributed or user-defined bins via a list of separating values and count the number
+of values in each bin, writing to a user-provided output histogram sequence. Currently, `histogram` is not supported
+with serial, tbb, or openmp backends in our oneDPL implementation. This RFC aims to propose the implementation of
+`histogram` for these host-side backends.
+
```
MikeDvorskiy, line 13: let me share my thoughts:
In my understanding RFC is not a book... So, I would preferer to have  a short, concise and precise description of what is offered, without frills, like a mathematical theorem. For example:

"The oneDPL library added histogram APIs, currently implemented only for device policies with the DPC++ backend. These APIs are defined in the oneAPI Specification 1.4.  Please see the
[oneAPI Specification](https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneDPL/source/parallel_api/algorithms.rst#parallel-algorithms). The host-side backends (serial, TBB, OpenMP) are not yet supported. This RFC proposes extending histogram support to these backends."

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,165 @@
...
+## Motivations
+Users don't always want to use device policies and accelerators to run their code. It may make more sense in many cases
+to use a serial implementation or a host-side parallel implementation of `histogram`. It's natural for a user to expect
+that oneDPL supports these other backends for all APIs. Another motivation for adding the support is simply to be spec
+compliant with the oneAPI specification.
+
```
MikeDvorskiy, line 19: Due to it is not a story telling, I would suggest omitting introductory expressions like  "It may make more sense" or "It's natural for a user to expect"... Only short and exact information.

For example,
"There are many cases to use a host-side serial or a host-side implementation of histogram. Another motivation for adding the support is simply to be spec compliant with the oneAPI specification."

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,165 @@
...
+Users have a choice of execution policies when calling oneDPL APIs. They also have a number of options of backends
+which they can select from when using oneDPL. It is important that all combinations of these options have support for
+the `histogram` APIs.
+
+### Performance
+As with all algorithms in oneDPL, our goal is to make them as performant as possible. By definition, `histogram` is a
+low computation algorithm which will likely be limited by memory bandwidth, especially for the evenly-divided case.
+Minimizing and optimizing memory accesses, as well as limiting unnecessary memory traffic of temporaries, will likely
+have a high impact on overall performance.
+
```
MikeDvorskiy, line 36: Taking into account my shared thought above, I would propose to re-prahse it keeping the main point shorter:

"A histogram algorithm is a memory-bound algorithm. So, the implementation should care of reducing memory accesses and minimizing temporary memory traffic."

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,165 @@
...
+### Memory Footprint
+There are no guidelines here from the standard library as this is an extension API. However, we should always try to
+minimize memory footprint whenever possible. Minimizing memory footprint may also help us improve performance here
+because, as mentioned above, this will very likely be a memory bandwidth-bound API. In general, the normal case for
+histogram is for the number of elements in the input sequence to be far greater than the number of output histogram
+bins. We may be able to use that to our advantage.
+
+### Code Reuse
```
MikeDvorskiy, line 44: I guess we can this topic omit at all. It tells nothing about 'histogram', just general wording, which can be applied for any new feature in oneDPL...

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,165 @@
...
+Our goal here is to make something maintainable and to reuse as much as we can which already exists and has been
+reviewed within oneDPL. With everything else, this must be balanced with performance considerations.
+
+### unseq Backend
```
MikeDvorskiy, line 48: "unseq Backend"
Basically, we don't have such back end officially. Yes, sometimes we used such term in the internal communication as for "name" for a set of functions with "pragma simd" implementation. But we did not specify and publish API for that.

akukanov: I think this part is about supporting (or not) implementation for unsequenced policies.
I do not mind calling it `unseq backend" in the design docs, but Mikhail is correct that it's rather informal (and parallel backend is somewhat more formal).

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,165 @@
...
+generally available, and certainly not available via OpenMP SIMD. Alternatively, we can multiply our number of temporary
+histogram copies by a factor of the vector width, but we will need to determine if this is worth the overhead, memory
+footprint, and extra accumulation at the end. OpenMP SIMD does provide an `ordered` structured block which we can use to
+exempt the increment from SIMD operations as well. It must be determined if SIMD is beneficial in either API variety. It
+seems only possible to be beneficial for the even bin API, but more investigation is required.
+
+Finally, for our below proposed implementation, there is the task of combining temporary histogram data into the global
+output histogram. This is directly vectorizable via our existing brick_walk implementation.
+
+### Serial Backend
```
MikeDvorskiy, line 75: https://github.com/oneapi-src/oneDPL/pull/1930/files#diff-fb5f6394ad0d350d719b9f31b139fa60c347ec64795c78e56875c4f002aeb0e7R25
We already have the `key requirements` topic where we enumerate all backends that we propose to support.
It is good enough I think, and we also omit this topic "Serial Backend".

Explanation what is  "Serial Backend" means as the others backends mean, is a kind of "oneDPL general description" and not related to RFC for histogram feature, IMHO.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: Yes, I've accepted your language here.  Thanks.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: taken suggestion.  Thanks

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: Taken mostly.  Thanks

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: I've removed some of the general language and added something which is important for histogram in an attempt to answer feedback from @akukanov to clarify where the implementation of the algorithm will live.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: I like the proposed name for the section better anyway for what is discussed.  Thanks.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: With some recent changes, there is some specifics about the serial implementation I wanted to add here so I've kept the section.

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,165 @@
...
+We plan to support a serial backend for histogram APIs in addition to openMP and TBB. This backend will handle all
+policies types, but always provide a serial unvectorized implementation.
+
+## Existing Patterns
+
```
MikeDvorskiy, line 68: If we intent to give some information about OneDPL parallel backend patterns on which histogram can based on, I would notify, there is not "count_if" pattern, there is "reduce"("transform_reduce") pattern.
When a man says "reduce", it becomes more or less obvious that histogram calculation based on reduce is not effective at all.

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,155 @@
...
+When we examine the real world utility of these cases, we find that they are uncommon and unlikely to be the important
+use cases. Histograms generally are used to categorize large images or arrays into a smaller number of bins to
+characterize the result. Cases for which there are similar or more bins than input elements are not very practical in
+practice. The maintenance and complexity cost associated with supporting and maintaining a second implementation to
+serve this subset of cases does not seem to be justified. Therefore, this implementation has been discarded at this
+time.
+
+### Other Unexplored Approaches
+* One could consider some sort of locking approach which locks mutexes for subsections of the output histogram prior to
```
MikeDvorskiy, line 119: BTW, I have a curiosity question. Which approach does NVidia use?

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,155 @@
...
+  percentage of the output histogram may be occupied, even when considering dividing the input amongst multiple
+  threads. This could be explored if we find temporary storage is too large for some cases and the atomic approach
+  does not provide a good fallback.
+
+## Proposal
+After exploring the above implementation for `histogram`, the following proposal better represents the use
+cases which are important, and provides reasonable performance for most cases.
+
+### Embarrassingly Parallel Via Temporary Histograms
+This method uses temporary storage and a pair of embarrassingly parallel `parallel_for` loops to accomplish the
```
MikeDvorskiy, line 135: What does `Embarrassingly Parallel` term mean?

MikeDvorskiy: 1. Update: got it.. https://en.wikipedia.org/wiki/Embarrassingly_parallel

2. Of course, if you are solving a concrete task and you are allowed to use the all machine recourses, and there are no any other workloads on the node, the best way for histogram calculation - to make static dividing of amount of work, each thread is calculating a local histogram, and after  the local histogram are reducing into one.

3. But, talking about parallelism in a kind of general library we have it keep in mind  that a final user's application can work in "different circumstances",  depends on their application type, task, real-time data, other workloads on the same host and other many things..
When we were developing TBB backend we kept in mind that things and preferred to use TBB `auto partitioner` (instead of static f.e).
Also composability reasons make sense here. 

4. BTW, have you considered a "common parallel reduce" (in general) pattern (and `tbb::parallel_reduce` pattern, in particular) for histogram calculation? It seems the parallel histogram calculation matches on the "common reduce": each `Body` calculates a local histogram (bins), `Combiner` summaries the all local bins into final one. 
Additionally, if number of bins  is "big" we can apply the second level of parallelism within `Combiner` code - SIMD or even "parallel_for" and SIMD, if number of bins is "too big".

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,155 @@
...
+`histogram`.
+
+For this algorithm, each parallel backend will add a  `__thread_enumerable_storage<_StoredType>` struct which provides
```
MikeDvorskiy, line 138: 1. Why a TLS is used, not the global memory  with "thread id" as a key? F.e. `bins[thread_id]`?
2. Does TBB guarantee that the same threads finalize the work? "The same threads" means the threads which have started the work?

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: NVidia has a similar API within CUB but not within Thrust, and therefore does not have a CPU implementation that I am aware of, only one specifically for a GPU device.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: 1) There are a number of reasons:
a) My understanding is that in TBB while it may be technically possible to get the the local thread id within an arena, it is an undocumented API and generally discouraged and against the TBB mindset.  Using TLS seems to be the preferred method specifically with TBB.
b) While what you suggest perhaps fits better within OpenMP, we want to create a single implementation and not require a `__parallel_histogram` within every current and future backend, but rather depend upon existing functionality within the backend as much as we can (in this case `__parallel_for`).
c) With smaller values of `n`, `num_bins` and larger number of threads, not all threads *should* be used because of the overhead associated with allocation and initialization of each temporary bin copy.  We can let the partitioner decide how many blocks to employ, but we want to avoid unnecessary allocation and initialization overheads wherever possible.

I will mention a downside for completeness, but it is outweighed here in my opinion:
It requires implementation of a thread local storage class for each backend.  This is only non-trivial for OpenMP. It has been written generically though to serve future patterns though so it is nice to have.

danhoeflinger: 2) I'm not exactly sure what you mean here by "finalize the work".  If you mean the second parallel for, then no, we are explicitly parallelizing over a different dimension (`num_bins`), and accumulating across the temporary histograms which were used from different threads.  TBB does guarantee that each thread will always use its own TLS for each grain of work though, when retrieved through `local()`.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: I clarified the language a little here to make it more clear that copy_if uses reduce internally.  I still think it deserves some text describing it as it may not be immediately obvious to everyone that reduce is not well matched.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: 2) Yes, although I think there is no reason here to do static division of work, but rather rely upon our existing `parallel_for` implementation to provide a good composable implementation.

3) Agree, which is why the intent is to use the existing `parallel_for` structure (including partitioners) to implement the parallelism.  If we were to do it from scratch, we would do it in a similarly composable way, but better to rely upon existing infrastructure

4) Yes, I thought about this.  For TBB and even more for openMP the built in reduction functionality is geared toward very simple lightweight types as the reduction variable where we may have an arbitrarily large array. Especially since we want a unified implementation, it does not seem like these backend are really set up to handle these large reduction variables.  It seems we should take more control to ensure no unnecessary copies are made, and that the final combination is done performantly, based on knowledge we have of the task.  The implementation remains quite simple and unified.

rfcs/proposed/host_backend_histogram/README.md
MikeDvorskiy: > 2\. I'm not exactly sure what you mean here by "finalize the work"

I try to explain with example:
Usage of  a general TBB pattern  `tbb::paralle_for` doesn't suppose using system thread directly. There is only a "Body" which is called (with a part of data(tbb range) by executing thread. Imagine the input range is split into 4 parts. Two threads call 2 parts simultaneously. The Body stores local bin results in TLS, associated with mentioned threads.
After, to "finalize the work", TBB should call  Body two time to process final 2 parts of input range. These final two calls may be done by another threads which have the other associated TLSs. So, it is impossible to make final reduce of local bins, located in TLSs.

danhoeflinger: There are 2 parallel_for calls, each of which is "embarrassingly parallel" where no thread body depends on previous thread bodies.  The first parallel_for must complete before the second one starts though.  The first parallel_for uses the TLS as normal, and just accumulates sections of the input data into each thread's individual TLS.  

The second parallel_for call does not use the TLS as normal, but rather has every thread visit a section of every TLS which was created one by one, processing a section of the histogram bins in parallel, combining the work of different threads from the first loop into the final global histogram.  

The TLS we propose here (that is also implemented in the PR) supports this, and we obtain the correct result.  We will not have perfect cache effects when accessing TLS from different threads than it was created upon but that is just something we have to deal with.

danhoeflinger:
One question I have for the group... 
If we know that a serial implementation will provide better performance up to some threshold (perhaps dependent on num bins, num threads, num input elements),  can / should we dispatch instead to a serial implementation? 

From my reading, it seems the answer is probably no.  Execution policies have semantic meaning, and `par` / `par_unseq` do not simply mean "provide the fastest version" even if that is what the users probably want.

mmichel11:
> One question I have for the group... If we know that a serial implementation will provide better performance up to some threshold (perhaps dependent on num bins, num threads, num input elements), can / should we dispatch instead to a serial implementation?
> 
> From my reading, it seems the answer is probably no. Execution policies have semantic meaning, and `par` / `par_unseq` do not simply mean "provide the fastest version" even if that is what the users probably want.

I agree that we should honor the user's request for a specific policy as opposed to using the serial implementation until some empirically determined cutoff point. I also imagine that the exact cutoff point where the parallel version performs better can highly vary dependent on a user's hardware setup and giving them the freedom to manually choose when to make the switch from the serial to parallel version can result in better performance than any generic decisions we could make.

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,167 @@
...
+
+Users have a choice of execution policies when calling oneDPL APIs. They also have a number of options of backends
+which they can select from when using oneDPL. It is important that all combinations of these options have support for
+the `histogram` APIs.
+
+### Performance
+With little computation, a histogram algorithm is likely a memory-bound algorithm. So, the implementation prioritize
+reducing memory accesses and minimizing temporary memory traffic.
+
+### Memory Footprint
```
mmichel11, line 31: Do we wish to specify space / computational complexity of the proposed algorithm somewhere, or is this too much?

mmichel11 (commented):
I've taken another pass through the document. A single question regarding how technical we want to get when explaining the algorithm. 

The RFC looks ready to me.

rfcs/proposed/host_backend_histogram/README.md
MikeDvorskiy: I don't understand your answer, Dan...
It seems you don't  catch my question/concerns...

danhoeflinger: We are basically using TBB's [enumerable_thread_specific](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.2-rev-1/elements/onetbb/source/thread_local_storage/enumerable_thread_specific_cls) as a model here, and implementing a stripped down version for omp and a trivial version for serial backend.

`enumerable_thread_specific` has two ways of accessing the data.  
a) `local()` which gets the TLS for the current thread, 
b) `begin()` and `end()` which provide iterators to the sequence of all local storage from all threads.  

This allows us to use (a) in the first parallel loop and (b) in the second parallel loop. The second parallel loop does not use the `enumerable_thread_specific` as a "Thread Local Storage" but rather a 2-D array space which it has summing across columns (corresponding to individual histogram bins from different threads). This allows us to accumulate the data from all threads into the global space histogram copy no matter which threads are used and when. 

I'm not sure how else I can explain it.  The code in the implementation is tested, working, and pretty concise, if you want to see the details you can look at the [PR](https://github.com/uxlfoundation/oneDPL/pull/1974).

MikeDvorskiy: I clarified that question with Alexey.
You mean "TBB TLS", not system TLS...  
TBB TLS is a kind of container and allows to iterate the all local bins... Know I got it.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: I've added some sections on computation complexity and temporary memory requirements.

I don't think we should add something like this to the specification though, because I don't want to dictate the algorithm used from the specification, but its good to document it in the current proposal.  Thanks.

akukanov:
> One question I have for the group...
If we know that a serial implementation will provide better performance up to some threshold (perhaps dependent on num bins, num threads, num input elements), can / should we dispatch instead to a serial implementation?
>
> From my reading, it seems the answer is probably no. Execution policies have semantic meaning, and par / par_unseq do not simply mean "provide the fastest version" even if that is what the users probably want.

I believe yes, we can. Generally, a serial implementation is correct for parallel execution policies, it's more of a QoI question. While the policies do not mean "provide the fastest version", they do not mean "always use multiple threads" either. It's a permission to use multiple threads, but not an obligation.

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,181 @@
...
+footprint where possible.
+
+### Code Reuse
+We want to minimize adding requirements for parallel backends to implement, and lift as much as possible to the
+algorithm implementation level. We should be able to avoid adding a `__parallel_histogram` call in the individual
+backends, and instead rely upon `__parallel_for`.
+
+### SIMD/openMP SIMD Implementation
+Currently oneDPL relies upon openMP SIMD to provide its vectorization, which is designed to provide vectorization across
+loop iterations. OneDPL does not directly use any intrinsics which may offer more complex functionality than what is
```
MikeDvorskiy, line 42: The second sentence may be omitted.
Based on the first sentence we can conclude that "OneDPL does not directly use any intrinsics..."

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,181 @@
...
+provided by OpenMP.
+
+There are a few parts of the histogram algorithm to consider. For the calculation to determine which bin to increment
+there are two APIs, even and custom range which have significantly different methods to determine the bin to
+increment. For the even bin API, the calculations to determine selected bin have some opportunity for vectorization as
+each input has the same mathematical operations applied to each. However, for the custom range API, each input element
+uses a binary search through a list of bin boundaries to determine the selected bin. This operation will have a
+different length and control flow based upon each input element and will be very difficult to vectorize.
```
MikeDvorskiy, line 50: But we can calculated bin indexes in SIMD manner. 
After that we can process the result in a serial loop.
No?

danhoeflinger: For now I'll ask that we leave it as described in the RFC, which gives some understanding of how this can be improved in the future, but starts without vectorization for this phase.
We can add an issue to explore using `simd ordered` to get some improvement for histogram even, and leave it out for this RFC and the initial PR implementation.

MikeDvorskiy: I see.. Ok, lets leave it as described.

MikeDvorskiy:
> I believe yes, we can. Generally, a serial implementation is correct for parallel execution policies, so it's more of a QoI question whether to parallelize or not. While the policies do not mean "provide the fastest version", they do not mean "always use multiple threads" either. It's a permission to use multiple threads, but not an obligation.

Agree with Alexey. Especially since, at least, the host patterns "sort"  and "merge" use some thresholds for switching to a serial implementation, depending on number of input elements.

MikeDvorskiy (commented):
Overall, looks good to me.
Just one comment about adding a description that the implementation may do a fallback to a serial implementation.

rfcs/proposed/host_backend_histogram/README.md
akukanov, line 1: I would suggest `algorithms_histogram_cpu_support` or `..._cpu_backends` as the proposal directory name, to follow the recommended pattern.

rfcs/proposed/host_backend_histogram/README.md
```diff
@@ -0,0 +1,185 @@
...
+### Key Requirements
+Provide support for the `histogram` APIs with the following policies and backends:
+- Policies: `seq`, `unseq`, `par`, `par_unseq`
+- Backends: `serial`, `tbb`, `openmp`
+
+Users have a choice of execution policies when calling oneDPL APIs. They also have a number of options of backends which
+they can select from when using oneDPL. It is important that all combinations of these options have support for the
+`histogram` APIs.
+
+### Performance
```
akukanov, line 27: For performance, does it make sense to add what we think are reasonable ranges for the number of bins and the number of elements, and why?

akukanov (approved):
Re-approved.

The questions in [my earlier approval](https://github.com/uxlfoundation/oneDPL/pull/1930#pullrequestreview-2512536551) are now addressed. The last couple of comments I made do not hold the proposal from landing.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: taken, thanks.

rfcs/proposed/host_backend_histogram/README.md
danhoeflinger: Added a paragraph to this end.

danhoeflinger:
Waiting for a second approval to merge, so I added the small asks you made in the mean time @akukanov .

MikeDvorskiy (approved):
LGTM

danhoeflinger:
Merging, with approvals from @akukanov  and @MikeDvorskiy .

PR #1931, opened 2024-11-04
RFC for zip_view implementation, for oneDPL C++20

MikeDvorskiy:


rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,25 @@
...
+the resulting range is a tuple containing one element from each of the input ranges. This can be particularly
+useful for iterating over multiple collections in parallel. `std::ranges::zip_view` is introduced starting with C++23,
+but there are many users who work with C++20 standard yet. So, oneDPL introduces `oneapi::dpl::ranges::zip_view`,
+which the same API and functionality as `std::ranges::zip_view`.
+
+
+### Key Requirements
+`oneapi::dpl::ranges::zip_view` should have:
+- API compliance with `std::ranges::zip_view`
+- `oneapi::dpl::__internal::tuple` underhood, due to `std::tuple` has issues with swapability and device copyability.
```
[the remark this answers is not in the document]
akukanov: Looks like there are two different requirements: the `zip_view` itself should be device copyable, and the `zip_view::iterator` should be value-swappable (https://en.cppreference.com/w/cpp/named_req/ValueSwappable).

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Motivations
+`std::ranges::zip_view` is a convenient way to combine multiple ranges into a single view, where each element of
+the resulting range is a tuple containing one element from each of the input ranges. This can be particularly
+useful for iterating over multiple collections in parallel. `std::ranges::zip_view` is introduced starting with C++23,
+but there are many users who work with C++20 standard yet. So, oneDPL introduces `oneapi::dpl::ranges::zip_view`,
+which the same API and functionality as `std::ranges::zip_view`.
+
+
+### Key Requirements
```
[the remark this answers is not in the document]
MikeDvorskiy: - yes.
- I think it should work (by the requirements)
- I think zip_view::iterator should be convertible to dpl::zip_iterator. I guess the unification (via one type) soesnt make sense due to dpl::zip_iterator requires only C++17, but zip_view and zip_view::iterator - C++20.
- I think any "underlying" views/ranges should/will work. There are no restrictions for the host execution, for a device - rather only device copyability. A general restriction (came from "range-based parallel algorithms paper") - random access "underlying" views/ranges and zip_view itself.
- From current perspective, it seems - no.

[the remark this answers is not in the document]
MikeDvorskiy: Update: In case of device execution  - an additional requirement for "underlying" views: each underlying view should model a `view` (see C++ standard). Basically, it follows from device copyability requirement.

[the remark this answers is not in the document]
akukanov: Good; make sure these questions are discussed in the design document.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,25 @@
...
+`oneapi::dpl::ranges::zip_view` should have:
+- API compliance with `std::ranges::zip_view`
+- `oneapi::dpl::__internal::tuple` underhood, due to `std::tuple` has issues with swapability and device copyability.
+
+
+### Performance
+Combining `std::ranges::zip_view` with data pipelines and kernel fusion enables developers to write expressive,
+efficient code for processing multiple related datasets. This approach not only simplifies the logic but also
+optimizes performance, making it an essential technique in modern C++ development. Whether you're working with
+simple transformations or complex data processing workflows, zip_view can be a valuable utility.
```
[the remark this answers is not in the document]
MikeDvorskiy: To the the truth I find it difficult to answer here...
Currently, I would say we have no (cannot define) any special performance requirements.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+the resulting range is a tuple containing one element from each of the input ranges. This can be particularly
+useful for iterating over multiple collections in parallel. `std::ranges::zip_view` is introduced starting with C++23,
+but there are many users who work with C++20 standard yet. So, oneDPL introduces `oneapi::dpl::ranges::zip_view`,
+which the same API and functionality as `std::ranges::zip_view`.
+
+### Key Requirements
+`oneapi::dpl::ranges::zip_view` should be:
+- compilable with C++20 version (minimum)
+- API compliant with `std::ranges::zip_view`
+- random accessible view; the "underlying" views also should be random accessible
```
dmitriy-sobolev, line 18: It is a stronger requirement compared to the `std::ranges::zip_view`, which "borrows" the access mode from the underlying ranges. Why is it necessary to impose it?

dmitriy-sobolev: Discussed it offline, and agreed that random accessibility is a mandatory requirement.

It is essential for the parallel algorithms. Moreover even algorithms with sequential execution policies will require them according to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3179r2.html, where any algorithm accepting an execution policy is referred as a "parallel range algorithm" with further requirement of random accessibility.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+- in case of a device usage: a device copyable view itself and the "underlying" views also should be device copyable
+  
+`oneapi::dpl::ranges::zip_view::iterator` should be:
+- value-swappable (https://en.cppreference.com/w/cpp/named_req/ValueSwappable)
+- convertible to `oneapi::dpl::zip_iterator`
+- abble to use with the non-range algorithms
+
+### Implementation proposal
+- `oneapi::dpl::ranges::zip_view` is designed as C++ class, which represents a range adaptor (see C++ Range Library).
```
dmitriy-sobolev, line 27:
```suggestion
- `oneapi::dpl::ranges::zip_view` is designed as a C++ class which represents a range adaptor (see C++ Range Library).
```

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Motivations
+`std::ranges::zip_view` is a convenient way to combine multiple ranges into a single view, where each element of
+the resulting range is a tuple containing one element from each of the input ranges. This can be particularly
+useful for iterating over multiple collections in parallel. `std::ranges::zip_view` is introduced starting with C++23,
+but there are many users who work with C++20 standard yet. So, oneDPL introduces `oneapi::dpl::ranges::zip_view`,
+which the same API and functionality as `std::ranges::zip_view`.
+
+### Key Requirements
+`oneapi::dpl::ranges::zip_view` should be:
+- compilable with C++20 version (minimum)
```
dmitriy-sobolev, line 16: Slightly more formal:

```suggestion
- compatible with C++20 version (minimum)
```

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+- API compliant with `std::ranges::zip_view`
```
dmitriy-sobolev, line 17:
```suggestion
- API-compliant with `std::ranges::zip_view`
```

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,27 @@
...
+- random accessible view; the "underlying" views also should be random accessible
+- in case of a device usage: a device copyable view itself and the "underlying" views also should be device copyable
+  
+`oneapi::dpl::ranges::zip_view::iterator` should be:
+- value-swappable (https://en.cppreference.com/w/cpp/named_req/ValueSwappable)
+- convertible to `oneapi::dpl::zip_iterator`
```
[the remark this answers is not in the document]
dmitriy-sobolev: > guess implicit one way conversion (zip_view::iterator to zip_iterator) is enough.

It should be noted, I assume.

However, my understanding of the Alexey's message is that if we decide to implement `dpl::ranges::zip_view::iterator`~> `dpl::zip_iterator` conversion, we should also implement conversions such as `std::iota_view::iterator` ~> `dpl::counting_iterator` for the completeness. That makes me also think if we should keep this requirement... The only (but awkward) example coming into my mind is when a user has a `zip_view` without access to the underlying containers, which prevents the use of `zip_iterator`, with the desire to use an iterator-based algorithm.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+- abble to use with the non-range algorithms
```
dmitriy-sobolev, line 24:
```suggestion
- able to be used with the non-range algorithms
```

This is optional, but I would also recommend starting a bullet with a capital letter.

dmitriy-sobolev (commented):
There is one question regarding the conversions of the iterators. The rest are minor stylistic and wording corrections.

rfcs/proposed/zip_view/README.md
MikeDvorskiy: I would like to write "to be able to compile", not compatible...

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+### Implementation proposal
+- `oneapi::dpl::ranges::zip_view` is designed as C++ class, which represents a range adaptor (see C++ Range Library).
+This class encapsulates a tuple-like type to keep a combination of two or more ranges.
+- The implementation provides all necessary operators to satisfy 'random accessible view' requirement.
+- To provide a device copyability requirement `oneapi::dpl::__internal::tuple` is proposed as tuple-like type underhood.
```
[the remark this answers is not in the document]
MikeDvorskiy: I an not sure that I agree with the proposed change... What does "an internal tuple-like" / "internal"  type mean?
`tuple-like type` is more or less common term,  "internal tuple" - not....

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
MikeDvorskiy: - practically-viable example: A user has some function like `void foo(oneapi::dpl::zip_iterator b, oneapi::dpl::zip_iterator e)` and has some ranges combined into one `zip_view` and is going to call `foo` with `zip_view`.
- `std::iota_view::iterator` ~> `dpl::counting_iterator`, `transform_view::iterator` to `dpl::transform_iterator`, etc... Yes, it makes sense. But it should be considered as the separate features in the another PRs, I think.

rfcs/proposed/zip_view/README.md
MikeDvorskiy: each line is a part of the first sentence above, not a new sentence, which starts with a capital letter.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Introduction
+`std::ranges::zip_view` is a powerful utility that enables developers to combine two or more ranges into a single view,
+where each element is represented as a tuple containing corresponding elements from each input range.
+
+## Motivations
+`std::ranges::zip_view` is a convenient way to combine multiple ranges into a single view, where each element of
+the resulting range is a tuple containing one element from each of the input ranges. This can be particularly
+useful for iterating over multiple collections in parallel. `std::ranges::zip_view` was introduced in C++23,
+but many developers are still using C++20 standard. So, oneDPL introduces `oneapi::dpl::ranges::zip_view`,
```
dmitriy-sobolev, line 11: Is this feature going to be experimental? If yes, I guess the namespace should be  `oneapi::dpl::experimental::ranges::zip_view`.

MikeDvorskiy: Firstly, it is going to be experimental, yes.
Regarding to `experimental` namespace here. I don't know... It depends on RFC process. This document is a proposal in the product or to be productized. So, `experimental` word looks weird.

Talking about the next the next stage of our doing with `zip_view`, yes we would like to have it in the experimental namespace first.

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
dmitriy-sobolev: Let's note that: 

```suggestion
- implicitly convertible to `oneapi::dpl::zip_iterator`
```

rfcs/proposed/zip_view/README.md
dmitriy-sobolev: I guess you expect the feature to be productized fast, similarly to the C++20 range support. Hence, having `experimental` namespace will impair user experience.

Does it make sense to treat is as an extension straight away? Do you feel confident regarding the current design?

danhoeflinger: My understanding here was that we were shooting for full product status for the ranges MVP APIs including `zip_view` after the experience of the existing experimental ranges APIs.  We already have an experimental `zip_view` that has been around for some time now.

dmitriy-sobolev: Ah, indeed, thanks for reminding about  `oneapi::dpl::experimental::ranges::zip_view`. Then, "supported" stage looks better than "experimental" as the next target.


> Talking about the next stage of our doing with zip_view, yes we would like to have it in the experimental namespace first.

Could you elaborate why you prefer "experimental" stage?

MikeDvorskiy: > Could you elaborate why you prefer "experimental" stage?

We don't have `zip_view` in the oneDPL spec yet... It will take additional time to prepare it. But we have very useful ready implemented utility right know. And we would like to have one in the closest oneDPL release.
Next, as far as the spec is ready, we will move `zip_view` from experimental to the product.

MikeDvorskiy: But, talking about existing `oneapi::dpl::experimental::ranges::zip_view`, not compliant with the standard `zip_view`... yes, that proposal rather should explain co-existance of these two implementations, I agree.

MikeDvorskiy: Basically,  there is an internal utility - `oneapi::dpl::__ranges::zip_view` in using everywhere in the internal oneDPL implementation with C++17 support.
Also `oneapi::dpl::__ranges::zip_view` is injected into `oneapi::dpl::experimental::ranges` namespace. 
`oneapi::dpl::experimental::ranges::zip_view` is used in 4 places in oneDPL impl... and probably in a user'code, because it is a public name.

So, I would propose for `_ONEDPL_CPP20_RANGES_PRESENT` mode to have the proposed new `zip_view` in the `oneapi::dpl::experimental::ranges` namespace and  injection of existing `oneapi::dpl::__ranges::zip_view` into `oneapi::dpl::experimental::ranges` otherwise.

@akukanov , @danhoeflinger , @dmitriy-sobolev what do you think?

danhoeflinger: In theory, we have license to change "experimental" things, as you suggest (in C++20 in this case).

What we need to be careful of is that the existing `zip_view` has been stable for a long time, and I think used in non-experimental production code internally.  I think we need to confirm and be sure that we are not introducing breaking changes into non-experimental code, and that any breaking changes to the experimental zip_view with C++20 enabled are well described and kept to a minimum.  

As far as the RFC is concerned, we need to make it clear here the plan for how it will coexist with what already exists (basically what you describe above).

MikeDvorskiy: > the existing `zip_view` has been stable for a long time

The existing zip_view doesn't work with C++ standard ranges. That's why we need new one  for _ONEDPL_CPP20_RANGES_PRESENT mode.

danhoeflinger: Yes that makes sense.  My reason for stating its stability is not to indicate that a new version isn't warranted, but merely that users may have more of an expectation to treat the existing implementation as a "supported" feature rather than an "experimental" feature, because we have basically treated it as such.  
Therefore, changes may take them by surprise.  We are not doing anything wrong by changing it, as it is experimental, we just need to be aware that it may affect users.

akukanov (approved):
I approve the proposal for implementation as an experimental feature.

The questions currently open in the review can be addressed later; just keep track of them please (for example, add to a special section at the end of the document).

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+with the same API and functionality as `std::ranges::zip_view`.
+
+### Key Requirements
+`oneapi::dpl::ranges::zip_view` should be:
+- compilable with C++20 version (minimum)
+- API-compliant with `std::ranges::zip_view`
+- random accessible view; the "underlying" views also should be random accessible
```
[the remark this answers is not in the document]
MikeDvorskiy: In my understanding our intention was to provide `zip_view `like adaptor (for C++20), which "works (is pipeable)" with C++20 ranges/views and can be passed into oneDPL algorithms, because existing `experimental::zip_view`  "doesn't work (is not pipeable)" with C++20 ranges/views.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+- in case of a device usage: a device copyable view itself and the "underlying" views also should be device copyable
```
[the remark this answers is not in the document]
MikeDvorskiy: yes, "a device copyable view itself" -  is not clear ant the sentence should be rephrased.

Yes, we say that the zip_view is device copyable if the base views are. 
The following sentence guarantees that:
`To provide a device copyability requirement `oneapi::dpl::__internal::tuple` is proposed as tuple-like type underhood.`

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+  
+`oneapi::dpl::ranges::zip_view::iterator` should be:
+- value-swappable (https://en.cppreference.com/w/cpp/named_req/ValueSwappable)
+- implicitly convertible to `oneapi::dpl::zip_iterator`
```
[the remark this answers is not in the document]
MikeDvorskiy: As I mentioned above
I have in my mind at least one practically-viable example: A user has some function like `void foo(oneapi::dpl::zip_iterator b, oneapi::dpl::zip_iterator e)` and has some ranges combined into one zip_view and is going to call foo with zip_view.

dmitriy-sobolev (approved):
The RFC also looks good to me.

My additional recommendation is to express the criteria for the feature to become full-supported. I believe, it can be done at a later stage, e.g. when the experimental feature is implemented or thoroughly reviewed for a better grasp of the context.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+- The implementation provides all necessary operators to satisfy 'random accessible view' requirement.
+- To ensure device copyability, `oneapi::dpl::__internal::tuple` is proposed as a tuple-like type for underlying elements.
+- To provide a value-swappable requirement `oneapi::dpl::__internal::tuple` is proposed as a dereferenced value for
+`oneapi::dpl::ranges::zip_view::iterator` due to `std::tuple` not satisfying the value-swappable requirement in C++20.
+- Usage of C++ concepts is desirable to write type requirements for types, methods and members of the class.
+- C++20 is minimum supported version for the class. It allows using modern C++ features such as concepts and others.
+
+### Test coverage
+
+- `oneapi::dpl::ranges::zip_view` is tested itself, base functionality.
```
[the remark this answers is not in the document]
MikeDvorskiy: I think to have at least one call of each method of `oneapi::dpl::ranges::zip_view` with result checking  is quite enough.
As an additional test coverage we may consider to take LLVM tests for `std::ranges::zip_view`, if it is technically possible by C++23/C++20 reasons.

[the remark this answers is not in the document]
MikeDvorskiy: Talking about "the API compliance with std::ranges::zip_view". I mean the public methods and types are equivalent at least for underlying random sized ranges: ones are callable and result the same. And also pipeable feature is supported.

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
akukanov: OK, so there is no intent to make it adaptable to different range categories and usable more widely. It needs to be clear in the documentation & specification.

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
akukanov: OK. Please adjust the text to make the intent clear.

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
akukanov: I tend to think that functions usually take generic iterators rather than specific ones. And as we know, implicit conversions may affect SFINAE, overload resolution sets, etc.

Maybe it is sufficient for the iterator traits to match, including their value and reference types? That makes sense and seems less risky to me. We can consider conversion (implicit or explicit) later, based on real demand and implementation complexity.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+## Introduction
+`std::ranges::zip_view` is a powerful utility that enables developers to combine two or more ranges into a single view,
+where each element is represented as a tuple containing corresponding elements from each input range.
+
+## Motivations
+`std::ranges::zip_view` is a convenient way to combine multiple ranges into a single view, where each element of
+the resulting range is a tuple containing one element from each of the input ranges. This can be particularly
+useful for iterating over multiple collections in parallel. `std::ranges::zip_view` was introduced in C++23,
+but many developers are still using C++20 standard. So, oneDPL introduces `oneapi::dpl::ranges::zip_view`,
+with the same API and functionality as `std::ranges::zip_view`.
```
akukanov, line 12: Thinking more of `zip_view`, I now wonder if we should add it to the specification or rather should provide it as a library extension.

If the goal is only to close the gap between C++20 and C++23, while the standard C++23 `zip_view` is well usable with oneDPL, including with device policies - then maybe it is not worth adding our `zip_view` variation to the specification. Perhaps especially so if it cannot be used as a substitution for the standard `zip_view` in all scenarios (that means, if developers cannot always use the oneDPL variation and may need to choose between that and the standard one).

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
akukanov: And we should also think of implications. If our `zip_view` cannot be used with ranges that are not random access, could it make developers hesitant to use it, possibly impacting oneDPL usage?

rfcs/proposed/zip_view/README.md
akukanov, line 23: Another compatibility/conversion question is with regard to the C++23 `zip_view`. If our view cannot always be used in place of the standard one, perhaps we should consider how developers would switch between the two.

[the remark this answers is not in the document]
MikeDvorskiy: Yes, explicitly conversion is "safer"...  
I don't mind to consider such conversion support "on demand".

MikeDvorskiy: Do you mean the standard `zip_view` C++23 might be not device copyable f.e.? And how developers can use our view instead of?

akukanov: No, I meant if our view is not going to support all requirements/use cases of the standard (e.g. not going to work with bidirectional etc. views), then developers cannot use it all the times. So it makes sense to think if there could be cases to transit from one zip view to another.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,40 @@
...
+- value-swappable (https://en.cppreference.com/w/cpp/named_req/ValueSwappable)
+- implicitly convertible to `oneapi::dpl::zip_iterator`
+- able to be used with the non-range algorithms
+
+### Implementation proposal
+- `oneapi::dpl::ranges::zip_view` is designed as a C++ class which represents a range adaptor (see C++ Range Library).
+This class encapsulates a tuple-like type to keep a combination of two or more ranges.
+- The implementation provides all necessary operators to satisfy 'random accessible view' requirement.
+- To ensure device copyability, `oneapi::dpl::__internal::tuple` is proposed as a tuple-like type for underlying elements.
+- To provide a value-swappable requirement `oneapi::dpl::__internal::tuple` is proposed as a dereferenced value for
```
danhoeflinger, line 31: When it comes to the specification, this will need to be described as an unspecified tuple which satisfies some set of requirements, unless we want to make the move to specify our internal tuple and make it public (which I doubt we want to do).

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
danhoeflinger: Would we want a test which requires c++23 and uses `std::ranges::zip_view` itself as the ground truth here?

rfcs/proposed/zip_view/README.md
danhoeflinger: In the context of this, I was looking at the specification for our current `zip_iterator`, and noticed a couple of discrepancies as compared to our implementation with regard to std::tuple vs our unspecified internal tuple type.  I believe that our implementation is intentional and the "correct" thing here and the spec should be changed, because of issues with device copyability when composing types from `zip_iterators`.  `std::tuple` is not trivially copyable but our internal tuple is.  

https://github.com/uxlfoundation/oneAPI-spec/pull/605

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
MikeDvorskiy: > Would we want a test which requires c++23 and uses `std::ranges::zip_view` itself as the ground truth here?

I guess - No, we don't. We rely on `std::ranges::zip_view` is tested within C++ standard library good enough.
But we may add some test cases with oneDPL algo calls with `std::ranges::zip_view` C++23.

rfcs/proposed/zip_view/README.md
MikeDvorskiy: Yes, the goal is only to close the gap between C++20 and C++23, while the standard C++23 zip_view is well usable with oneDPL.

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
MikeDvorskiy: In my understanding the proposed `zip_view` is just to use with oneDPL parallel algorithms, which accept just RA ranges. The goal was not adding `zip_view`  for C++20 which works with any type ranges (like C++23 `std::zip_view`)

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
MikeDvorskiy: done

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Introduction
+`std::ranges::zip_view` is a powerful utility that enables developers to combine two or more ranges into a single view,
+where each element is represented as a tuple containing corresponding elements from each input range.
+
+## Motivations
+`std::ranges::zip_view` is a convenient way to combine multiple ranges into a single view, where each element of
+the resulting range is a tuple containing one element from each of the input ranges. This can be particularly
+useful for iterating over multiple collections in parallel. `std::ranges::zip_view` was introduced in C++23,
+but many developers are still using C++20 standard. So, oneDPL introduces `oneapi::dpl::ranges::zip_view`,
```
akukanov, line 11: Another part of the story, not yet described in the motivation, is that the implementations of `std::ranges::zip_view` will not work with device policies out of the box. So. our own implementation is necessary not just to provide functionality missing in C++20, but primarily to use it with device policies.

rfcs/proposed/zip_view/README.md
akukanov: I think we now understand that there is more value in our implementation than just closing the gap between standard versions. I believe it should be a specified API. not an extension.

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
akukanov: Am I correct that after evaluation we think we can support `zip_view` beyond our own algorithms, including in the serial code, range-based `for` loops, etc.? Is the requirement of only supporting random-access still important?

rfcs/proposed/zip_view/README.md
akukanov: We are moving towards specifying the tuple-like type better, most likely as a type that satisfies certain concepts / requirements. That should be done in an new proposal though.

rfcs/proposed/zip_view/README.md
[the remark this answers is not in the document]
MikeDvorskiy: Yes, the thread is not actual any more, because the current (the last) `zip_view` implementation supports all access types of input ranges.

rfcs/proposed/zip_view/README.md
MikeDvorskiy: Yes, it's  true.

rfcs/proposed/zip_view/README.md
MikeDvorskiy: I reflected this points (and another points) in the update.

rfcs/proposed/zip_view/README.md
MikeDvorskiy: Yes.. I've mentioned it in the RFC document update.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,56 @@
...
+`std::ranges::zip_view` C++23 still is not device copyable. Any wrapper over `std::tuple` C++23 is not device copyable. (https://godbolt.org/z/brfvcMeM6)
+There are another technical issues with `std::tuple` (see below for the details).
+So, oneDPL tuple-like concept  should be proposed into the oneDPL spec.
+
+### Key Requirements
+`oneapi::dpl::ranges::zip_view` should be:
+- compilable with C++20 version (minimum)
+- API-compliant with `std::ranges::zip_view`
+- in case of a device usage: a device copyable view if the all "underlying" views are device copyable views.
+- To provide a transitive device copyability oneDPL tuple-like type underhood is proposed - `oneapi::dpl::__internal::tuple`.
```
akukanov, line 24: I think which tuple to use is an implementation detail. What we really want is that our `zip_view` is trivially copyable if all the underlying views are - this is the way to have transitive device copyability for view pipelines.

MikeDvorskiy: Yes, I think "to have transitive device copyability for view pipelines" - very strong argument to want that our zip_view is trivially copyable.
BTW, the proposed implementation has such property. The relevant test also was added.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,56 @@
...
+### Other technical reasons not to use std::zip_view C++23 (and std::tuple) with oneDPL algorithms in the future:
+- There is an issue with `std::ranges::sort(zip_view)` with clang 19.0 and older. (https://godbolt.org/z/jKvG9rY5M)
+- There is an issue with `std::ranges::stable_sort(zip_view)` with gcc library 
+- Passing `std::zip_view::iterator` intances to the iterator-based algorithms works only for gcc 14.1 and newer, clang 19.1 and newer or
+  starting 17.01 with libc++ lib (https://godbolt.org/z/To6Mjr9M6)
+- Considiration `std::tuple` as `oneapi::dpl::ranges::zip_view::iterator::value_type`. There are issues, at least, with `sortable`, `permutable`
+  and `indirectly_writable` concepts: const_cast<const std::iter_reference_t<Out>&&>(*o) = std::forward<T>(t) is not compiled till C++23.  (https://godbolt.org/z/zT9qqnjWq)
+
+### Implementation proposal (C++20)
+- `oneapi::dpl::ranges::zip_view` is designed as a C++ class which represents a range adaptor (see C++ Range Library).
```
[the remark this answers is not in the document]
MikeDvorskiy: yes, it is true - the implementation is based on the C++ view_interface.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+### Key Requirements
+`oneapi::dpl::ranges::zip_view` should be:
+- compilable with C++20 version (minimum)
+- API-compliant with `std::ranges::zip_view`
+- in case of a device usage: a device copyable view if the all "underlying" views are device copyable views.
+- To provide a transitive device copyability oneDPL tuple-like type underhood is proposed - `oneapi::dpl::__internal::tuple`.
+  
+`oneapi::dpl::ranges::zip_view::iterator` should be:
+- value-swappable (https://en.cppreference.com/w/cpp/named_req/ValueSwappable)
```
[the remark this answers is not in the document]
MikeDvorskiy: Updated the wording regarding the iterator.

rfcs/proposed/zip_view/README.md
```diff
@@ -0,0 +1,56 @@
...
+- able to be used with the non-range algorithms
+
+### Discrepancies with std::zip_view C++23
+- `oneapi::dpl::ranges::zip_view` is based on oneDPL tuple-like type oneapi::dpl::__internal::tuple instead of std::tuple.
+- `oneapi::dpl::ranges::zip_view::iterator::value_type` is oneDPL tuple-like type oneapi::dpl::__internal::tuple instead of std::tuple.
```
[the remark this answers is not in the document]
MikeDvorskiy: Updated the wording; removed the mention of `oneapi::dpl::__internal::tuple`

rfcs/proposed/zip_view/README.md
MikeDvorskiy: Updated the wording; removed the mention of oneapi::dpl::__internal::tuple

danhoeflinger (approved):
LGTM

PR #1933, opened 2024-11-06
Optimize `merge` algorithm for largest data sizes

SergeyKopienko:
In this PR we optimize `merge` algorithm for data sizes equal or greater then `8M` items (`16M` items for `int` type).

The main idea - we doing two submits:
1) in the first `submit` we find split point in some `"base"` diagonal's subset.
2) in the second `submit` we find split points in all other diagonal and run `serial merge` for each diagonal (as before).
    But when we find split point on the current diagonal, we setup some indexes limits for `rng1` and `rng2`: by this way we decrease the amount of reading data from source data ranges.  For these limits we load split point's data from previous and next `"base"` diagonals, calculated on the step (1).

Applying this approach we have good perf profit for biggest data sizes with  `float` and `int` data types.

### Details:
Probably I should explain why I have create
```C++
struct __result_and_scratch_storage_base
{
    virtual ~__result_and_scratch_storage_base() = default;
};
```
and why I inherit `__result_and_scratch_storage` from `__result_and_scratch_storage_base` :
```C++
struct __result_and_scratch_storage : __result_and_scratch_storage_base
{
    // ...
};
```

Let me explain the reason.
In the `__parallel_merge(oneapi::dpl::__internal::__device_backend_tag, ...)` function we checks in run-time some conditions to make decision which submitter we will call:
1) `__parallel_merge_submitter<std::uint32_t, ...>`
2) `__parallel_merge_submitter_large<std::uint32_t, ...>`
3) `__parallel_merge_submitter_large<std::uint64_t, ...>`

In the cases (2) and (3) we should extend the life-time of different `__result_and_scratch_storage` instances which were created inside `__result_and_scratch_storage<std::uint32_t>::operator()` or `__result_and_scratch_storage<std::uint64_t>::operator()`:
 - in  the case (1) we will not have any `__result_and_scratch_storage`;
 - in  the case (2) we will have `__result_and_scratch_storage<std::uint32_t>`;
 - in  the case (3) we will have `__result_and_scratch_storage<std::uint64_t>`.
 
 So the `std::shared_ptr<__result_and_scratch_storage_base>` is some common interface for the both of them which can extend their life-time. 

About `virtual ~__result_and_scratch_storage_base() = default` - it's needed to properly destroy instances of inherited classes, which are really owned by shared-pointer instance in the cases (2) and (3). Without `virtual` here we will have memory-leaks.


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -166,9 +273,117 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+
+        using __result_and_scratch_storage_t = __result_and_scratch_storage<_ExecutionPolicy, _split_point_t>;
+        __result_and_scratch_storage_t __result_and_scratch{__exec, 0, __base_diag_count + 1};
+
+        sycl::event __event = __exec.queue().submit([&](sycl::handler& __cgh) {
+            oneapi::dpl::__ranges::__require_access(__cgh, __rng1, __rng2);
+            auto __scratch_acc = __result_and_scratch.template __get_scratch_acc<sycl::access_mode::write>(
+                __cgh, __dpl_sycl::__no_init{});
+
+            __cgh.parallel_for<_FindSplitPointsKernelOnMidDiagonal>(
```
SergeyKopienko, line 321: @rarutyun Compile error is here: 
https://github.com/oneapi-src/oneDPL/actions/runs/11722920053/job/32653481992?pr=1933
"C++
D:\a\oneDPL\oneDPL\include\oneapi\dpl\pstl\hetero\dpcpp\parallel_backend_sycl_merge.h(322,64): error: definition with same mangled name '_ZTSN6oneapi3dpl20__par_backend_hetero10__internal11__compositeINS1_50_find_split_points_kernel_on_mid_diagonal_uint32_tIJEEEJLc95ELc90ELc84ELc83ELc78ELc54ELc111ELc110ELc101ELc97ELc112ELc105ELc51ELc100ELc112ELc108ELc50ELc48ELc95ELc95ELc112ELc97ELc114ELc95ELc98ELc97ELc99ELc107ELc101ELc110ELc100ELc95ELc104ELc101ELc116ELc101ELc114ELc111ELc53ELc48ELc95ELc102ELc105ELc110ELc100ELc95ELc115ELc112ELc108ELc105ELc116ELc95ELc112ELc111ELc105ELc110ELc116ELc115ELc95ELc107ELc101ELc114ELc110ELc101ELc108ELc95ELc111ELc110ELc95ELc109ELc105ELc100ELc95ELc100ELc105ELc97ELc103ELc111ELc110ELc97ELc108ELc95ELc117ELc105ELc110ELc116ELc51ELc50ELc95ELc116ELc73ELc74ELc78ELc83ELc48ELc95ELc57ELc101ELc120ELc101ELc99ELc117ELc116ELc105ELc111ELc110ELc53ELc95ELc95ELc100ELc112ELc108ELc49ELc55ELc68ELc101ELc102ELc97ELc117ELc108ELc116ELc75ELc101ELc114ELc110ELc101ELc108ELc78ELc97ELc109ELc101ELc69ELc78ELc83ELc48ELc95ELc56ELc95ELc95ELc114ELc97ELc110ELc103ELc101ELc115ELc49ELc48ELc103ELc117ELc97ELc114ELc100ELc95ELc118ELc105ELc101ELc119ELc73ELc80ELc105ELc69ELc69ELc83ELc57ELc95ELc106ELc78ELc83ELc48ELc95ELc49ELc48ELc95ELc95ELc105ELc110ELc116ELc101ELc114ELc110ELc97ELc108ELc49ELc49ELc95ELc95ELc112ELc115ELc116ELc108ELc95ELc108ELc101ELc115ELc115ELc69ELc69ELc69ELc69EEEE' as another definition
  322 |                 sycl::range</*dim=*/1>(__base_diag_count + 1), [=](sycl::item</*dim=*/1> __item_id)
      |                                                                ^
"

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -166,9 +273,119 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+    {
+        const _IdType __n1 = __rng1.size();
+        const _IdType __n2 = __rng2.size();
+        const _IdType __n = __n1 + __n2;
+
+        assert(__n1 > 0 || __n2 > 0);
+
+        _PRINT_INFO_IN_DEBUG_MODE(__exec);
+
+        using _FindSplitPointsOnMidDiagonalKernel =
```
SergeyKopienko, line 297: @rarutyun I have fixed the error here. Is it correct way?
I am using `__kernel_name_generator` here because I should have two Kernel names: one passed as template parameter pack and the second name I should create inside.

SergeyKopienko: Will discuss with @rarutyun but as far as I understood it's not possible because we should have mandatory type pack as template param for the class like
```C++
template <typename... _Name>
class _find_split_points_kernel_on_mid_diagonal;
```

danhoeflinger: Aha, this is actually an interesting one because the type differs based on a runtime parameter rather than the types the user provides, and so we need both for a single public API call in the code.  I see why its necessary and we cant just add it to the pack now. 

```suggestion
        // Runtime decision to pick the _IdType, so we need both kernels for a single public call, and must embed that in the base name
        using _KernelNameWithType = std::conditional_t<std::is_same_v<_IdType, std::uint32_t>, _find_split_points_kernel_on_mid_diagonal_uint32_t, _find_split_points_kernel_on_mid_diagonal_uint64_t>;
        using _FindSplitPointsOnMidDiagonalKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<_KernelNameWithType,_CustomName, _ExecutionPolicy,
                                   _Range1, _Range2, _Range3, _Compare>;
```

SergeyKopienko: Unfortunately this approach has compile errors: discussed offline.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -166,9 +273,119 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+
+            __cgh.parallel_for<_Name...>(sycl::range</*dim=*/1>(__steps), [=](sycl::item</*dim=*/1> __item_id) {
+                auto __global_idx = __item_id.get_linear_id();
+                const _IdType __i_elem = __global_idx * __chunk;
+
+                auto __scratch_ptr = __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__scratch_acc);
+                auto __scratch_idx = __global_idx / __base_diag_part;
+
+                _split_point_t __start;
+                if (__global_idx % __base_diag_part != 0)
```
danhoeflinger, line 359: We discussed offline about the approach to partition based on SLM size and then work within the partitioned blocks in the second kernel.  

One advantage of this method (beyond working within SLM for all diagonals in this kernel) would be that there would be no work-item divergence with a branch and mod operation like this.  The first partitioning kernel would be lightweight and basically only to establish bounds for the second kernel.  Then the second kernel would work within SLM loaded data and search for all diagonals within that block then serial merge, and all work items could be the same (with possible exception for the zeroth work item).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -177,23 +387,51 @@ __parallel_merge(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy
...
+        if (__n <= std::numeric_limits<std::uint32_t>::max())
+        {
+            using _WiIndex = std::uint32_t;
+            using _MergeKernelName = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
+                __merge_kernel_name<_CustomName, _WiIndex>>;
+            return __parallel_merge_submitter<_WiIndex, _MergeKernelName>()(
+                std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__rng1), std::forward<_Range2>(__rng2),
+                std::forward<_Range3>(__rng3), __comp);
+        }
+        else
```
mmichel11, line 401: Can we remove this else branch and its kernel? `std::uint32_t` has a well-defined maximum, and we know `__n < 4 * 1'048'576` in this branch, so it can always be indexed with this type.

SergeyKopienko: Good point.
But another solution - we may use `std::uint16_t` type too for smaller data sizes.

mmichel11: Do you have an estimate of how much performance benefit the 16-bit indexing kernel brings? I think it would be best to weigh the impact of this kernel against the increase in JIT time. If the performance benefit is significant, then I am in favor of keeping it.

SergeyKopienko: No perf profit, removed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: I has checked suggestion with using of SLM cache in https://github.com/oneapi-src/oneDPL/pull/1937
No additional performance profit achived.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -19,7 +19,7 @@
 #include <limits>    // std::numeric_limits
 #include <cassert>   // assert
 #include <cstdint>   // std::uint8_t, ...
-#include <utility>   // std::make_pair, std::forward
+#include <utility>   // std::forward
```
akukanov, line 22: An accidental change?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -78,6 +80,148 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
...
+//   |   |
+// 0 | 0 | 1  1  1   1
+//   |   ---------->
+// 2 | 0   0  0  0 | 1
+//   |             ---->
+// 3 | 0   0  0  0   0 |
+template <typename _Rng1, typename _Rng2, typename _Index, typename _Compare>
+_split_point_t<_Index>
+__find_start_point_in(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_to, const _Rng2& __rng2,
+                      const _Index __rng2_from, _Index __rng2_to, const _Index __i_elem, _Compare __comp)
```
akukanov, line 98: If a "full" unification of the implementation is not feasible, I would at least try to unify `__find_start_point` and `__find_start_point_in` into a single function. In theory it should be possible, because `const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_to` together really define a subrange of `__rng1`; so perhaps we could create it on the caller side and pass to the function as the range to process.

akukanov (commented):
Would it be reasonable to unify the implementation, by considering smaller sizes to be just a "degenerate" case of big sizes which does not need coarse-grain split points and only executes the second kernel?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Yes. Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: There is one moment.
My checks shows me that `__find_start_point_in` is more effective on middle and large data sizes
and `__find_start_point` is more effective on smaller data sizes.
That was the reason why I save old implementation.
But technically I can use `__find_start_point_in` in all cases: for smaller data sizes some performance degradation will have place.

SergeyKopienko:
> Would it be reasonable to unify the implementation, by considering smaller sizes to be just a "degenerate" case of big sizes which does not need coarse-grain split points and only executes the second kernel?

@akukanov, I prepared PR https://github.com/oneapi-src/oneDPL/pull/1956 and checked perf.
On small sizes the perf is worst then in this PR.
So I propose do not combine two submitters into one.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -78,6 +80,148 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
...
+    assert(__rng1_from <= __rng1_to);
+    assert(__rng2_from <= __rng2_to);
+
+    assert(__rng1_to > 0 || __rng2_to > 0);
+
+    if constexpr (!std::is_pointer_v<_Rng1>)
+        assert(__rng1_to <= __rng1.size());
+    if constexpr (!std::is_pointer_v<_Rng2>)
+        assert(__rng2_to <= __rng2.size());
+
```
MikeDvorskiy, line 109: The lines 105-108 look suspicious by several reasons:
1) it is a mix compile time and runtime conditions.
2) Rng1 (Rng2) is not supposed to be a pointer type. If we want to check that we can add a relevant `static_assert` here.
3) If we want to have some bound checks for rng1 and rng2, I would suggest writing that in a couple of asserts above (lines 100-102)  like that:
```
assert(0 <= __rng1_from && __rng1_from < __rng1_to && __rng1_to < __rng1.size());
assert(0 <= __rng2_from && __rng2_from < __rng2_to && __rng2_to < __rng2.size());
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -78,6 +80,148 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
...
+        const _split_point_t<_Index> __result{ *__res, __index_sum - *__res + 1 };
+        assert(__result.first + __result.second == __i_elem);
+
+        assert(__rng1_from <= __result.first && __result.first <= __rng1_to);
+        assert(__rng2_from <= __result.second && __result.second <= __rng2_to);
+
+        return __result;
+    }
+    else
+    {
```
MikeDvorskiy, line 218: It seems work item number '0' does nothing... 
It would be good if we could changes code design a little bit to utilize the all work items.

SergeyKopienko: Fixed.
About `work item number '0' does nothing...` - it's not true: 0-nt work-item simple call `merge`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: But in the test `sort.pass` we has the case when we pass into this function pointers as `__rng1` and `__rng2`.
So I propose to rewrite:
```C++
    if constexpr (!std::is_pointer_v<_Rng1> && !std::is_pointer_v<_Rng2>)
    {
        assert(0 <= __rng1_from && __rng1_from < __rng1_to && __rng1_to < __rng1.size());
        assert(0 <= __rng2_from && __rng2_from < __rng2_to && __rng2_to < __rng2.size());
    }
    else
    {
        assert(0 <= __rng1_from && __rng1_from < __rng1_to);
        assert(0 <= __rng2_from && __rng2_from < __rng2_to);
    }
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -78,6 +80,139 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
...
+//   |--->
+// 0 | 0 | 1  1  1   1
+//   |   |
+// 0 | 0 | 1  1  1   1
+//   |   ---------->
+// 2 | 0   0  0  0 | 1
+//   |             ---->
+// 3 | 0   0  0  0   0 |
+template <typename _Rng1, typename _Rng2, typename _Index, typename _Compare>
+_split_point_t<_Index>
```
MikeDvorskiy, line 96: I would use `std::pair<_Index>` directly here.
And would use a simpler writing in the returning line:
`return {*__res, __index_sum - *__res + 1};`

SergeyKopienko: I believe you still have this ability, because `_split_point_t` is simple type name.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: Well. So, the function can accept some type `_Rng1` which is an "accessible" or, other words the type must have subscription operator `operator[]`, nothing more.
Basically,  the function code doesn't need `size()` "method". And I think we should not demand it for the sake of "assert".

SergeyKopienko: For debug purposes it's very useful, but I am ready to remove `.size()` calls. Will remove?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -129,45 +264,230 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index _
...
+
+                    auto __base_diagonals_sp_global_ptr =
+                        _Storage::__get_usm_or_buffer_accessor_ptr(__base_diagonals_sp_global_acc);
+                    auto __diagonal_idx = __global_idx / __nd_range_params.steps_between_two_base_diags;
+
+                    _split_point_t<_IdType> __start;
+                    if (__global_idx % __nd_range_params.steps_between_two_base_diags != 0)
+                    {
+                        // Check that we fit into size of scratch
+                        assert(__diagonal_idx + 1 < __nd_range_params.base_diag_count + 1);
```
mmichel11, line 415: General question: do we have a consensus on whether it is okay to use `assert` in SYCL kernels? I know it is done in some places and in tests, but I do not think it is officially supported by the SYCL spec.

It is supported by the DPC++ compiler but what about others?

SergeyKopienko: At least `assert` used in the examples of code in https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
akukanov: If we have full control over all the types of ranges that can be passed as inputs to the code, and so we can ensure that the `size` method is always available, then it would be OK to use it for assertions/debug.

If we do not have such level of control, and ranges can be of user-specified types, then we clearly cannot use this method.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
akukanov: None of the examples in the SYCL specification uses `assert` in device code.

We should consider the use of `assert` in device code as an extension of DPC++ and possibly other SYCL implementations, so it should be guarded by some macros. @dmitriy-sobolev, FYI.

SergeyKopienko: Thanks for this comment: you are right. Will remove assert's from Kernel code.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: These checks has been removed at all.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: So I propose to save both `__find_start_point` and `__find_start_point_in` functions.
The justification - to have the same performance on the sizes less then 4 Mb.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -1,204 +1,508 @@
...
-            std::forward<_Range3>(__rng3), __comp);
-    }
-}
-
-} // namespace __par_backend_hetero
-} // namespace dpl
-} // namespace oneapi
-
-#endif // _ONEDPL_PARALLEL_BACKEND_SYCL_MERGE_H
+// -*- C++ -*-
```
MikeDvorskiy, line 1: Why each line of the file is shown as changed? I guess Windows EOL symbol is used...
Could you please check and clean up it? Otherwise, it makes reviewing impossible.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -518,8 +518,14 @@ struct __usm_or_buffer_accessor
     }
 };
 
+struct __result_and_scratch_storage_base
```
MikeDvorskiy, line 521: Why do we introduce that class with virtual destructor?
On first glance I cannot see a polymorphic behavior... 
Even it there is, I would like to warn  - "virtual functions" don't work within a kernel..

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: This `virtual` destructor in the base class really required because we should return the same result from `__parallel_merge_submitter_large::operator()` for two different specializations of `__parallel_merge_submitter_large` by `std::uint32_t` and `std::uint64_t` types. And for that we return the `std::shared_ptr` instance with the pointer to the base class `__result_and_scratch_storage_base`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: `virtual` required here only on host side.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: It is an answer to the next question...

https://github.com/oneapi-src/oneDPL/pull/1933/files#r1888964591 still requires to be addressed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -1,204 +1,508 @@
...
+                    sycl::range</*dim=*/1>(__steps), [=](sycl::item</*dim=*/1> __item_id) {
+                        const _IdType __i_elem = __item_id.get_linear_id() * __chunk;
+                        const auto __start = __find_start_point(__rng1, __rng2, __i_elem, __n1, __n2, __comp);
+                        __serial_merge(__rng1, __rng2, __rng3, __start.first, __start.second, __i_elem, __chunk, __n1,
+                                       __n2, __comp);
+                    });
+            });
+        // We should return the same thing in the second param of __future for compatibility
+        // with the returning value in __parallel_merge_submitter_large::operator()
+        return __future(__event, __result_and_scratch_storage_base_ptr{});
```
MikeDvorskiy, line 269: It looks like that `class __future` supports only `sycl::buffer`  or `__result_and_scratch_storage` or ...  there is another general version of `__wait_and_get_value` method:

    template <typename _T>
```
    constexpr auto
    __wait_and_get_value(const _T& __val)
    {
        wait();
        return __val;
    }
```
What `future` instance does return in case of 
`return __future(__event, __result_and_scratch_storage_base_ptr{});` ?
`std::shared_ptr` on the dummy `__result_and_scratch_storage_base` instance?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Thanks, will fix.

SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: The answer is in real use case:
```C++
            .__deferrable_wait();
```
So we simple waits.
About your question - probably we will return shared-pointer (`__result_and_scratch_storage_base_ptr`) in my implementation.

SergeyKopienko: > It looks like that `class __future` supports only `sycl::buffer` or `__result_and_scratch_storage` or ... there is another general version of `__wait_and_get_value` method:
> 
> ```c++
>     template <typename _T>
>     constexpr auto
>     __wait_and_get_value(const _T& __val)
>     {
>         wait();
>         return __val;
>     }
> ```
> 
> What `future` instance does return in case of `return __future(__event, __result_and_scratch_storage_base_ptr{});` ? `std::shared_ptr` on the dummy `__result_and_scratch_storage_base` instance?

It should and really return `__result_and_scratch_storage_base_ptr{}` instance.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -518,8 +518,14 @@ struct __usm_or_buffer_accessor
...
+{
+    virtual ~__result_and_scratch_storage_base() = default;
+};
+using __result_and_scratch_storage_base_ptr = std::shared_ptr<__result_and_scratch_storage_base>;
```
SergeyKopienko, line 525: Please see my comments in the description of this PR about what is this and why we need to have it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,251 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+
+template <typename _IdType, typename _CustomName, typename _DiagonalsKernelName, typename _MergeKernelName>
+struct __parallel_merge_submitter_large;
+
+template <typename _IdType, typename _CustomName, typename... _DiagonalsKernelName, typename... _MergeKernelName>
+struct __parallel_merge_submitter_large<_IdType, _CustomName,
+                                        __internal::__optional_kernel_name<_DiagonalsKernelName...>,
+                                        __internal::__optional_kernel_name<_MergeKernelName...>>
+{
+  protected:
```
mmichel11, line 264: Can this just be `private`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,251 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+class __merge_kernel_name_large;
+
+template <typename... _Name>
+class __diagonals_kernel_name;
+
+template <typename _Tp>
+constexpr std::size_t
+__get_starting_size_limit_for_large_submitter()
+{
+    return 4 * 1'048'576; // 4 MB
```
mmichel11, line 429: With this starting size, will the large submitter path be tested in our unit tests? If I recall, we go up to 100k for most cases.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,251 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
 
-        auto __event = __exec.queue().submit([&](sycl::handler& __cgh) {
-            oneapi::dpl::__ranges::__require_access(__cgh, __rng1, __rng2, __rng3);
-            __cgh.parallel_for<_Name...>(sycl::range</*dim=*/1>(__steps), [=](sycl::item</*dim=*/1> __item_id) {
-                const _IdType __i_elem = __item_id.get_linear_id() * __chunk;
-                const auto __start = __find_start_point(__rng1, __rng2, __i_elem, __n1, __n2, __comp);
-                __serial_merge(__rng1, __rng2, __rng3, __start.first, __start.second, __i_elem, __chunk, __n1, __n2,
-                               __comp);
+        auto __event = __exec.queue().submit(
+            [&__rng1, &__rng2, &__rng3, __comp, __chunk, __steps, __n1, __n2](sycl::handler& __cgh) {
```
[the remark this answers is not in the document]
SergeyKopienko: For example let's take a look at `__sycl_reduce_by_segment`: all data captured by `&`.
So this place looks ok for me (about capturing of ranges by value).
@MikeDvorskiy how do you think?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Good point, will fix in the test.

SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: @MikeDvorskiy , can I mask this discussion as resolved?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,251 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+                oneapi::dpl::__ranges::__require_access(__cgh, __rng1, __rng2, __rng3);
+                __cgh.parallel_for<_MergeKernelName...>(
+                    sycl::range</*dim=*/1>(__steps), [=](sycl::item</*dim=*/1> __item_id) {
+                        const _IdType __i_elem = __item_id.get_linear_id() * __chunk;
+                        const auto __start = __find_start_point(__rng1, __rng2, __i_elem, __n1, __n2, __comp);
+                        __serial_merge(__rng1, __rng2, __rng3, __start.first, __start.second, __i_elem, __chunk, __n1,
+                                       __n2, __comp);
+                    });
             });
+        // We should return the same thing in the second param of __future for compatibility
```
MikeDvorskiy, line 250: It seem we should not...
`__parallel_merge_submitter::operator()` and `__parallel_merge_submitter_large::operator()`
return the same type.... 

line 252:  "future(event,  `__result_and_scratch_storage_base_ptr{}`)
line 412: future(event, __p_result_and_scratch_storage_base), 

where the second arg is also an instance of `__result_and_scratch_storage_base_ptr` type, because line 
396 tells: `__result_and_scratch_storage_base_ptr __p_result_and_scratch_storage_base;`

So, we may just writing  `return __future(__event);` in the both cases.

SergeyKopienko: Please take a look into the code, what exactly we return and which types it has:
- in the `__parallel_merge_submitter::operator()` :
```C++
return __future(__event, __result_and_scratch_storage_base_ptr{});
```
- in the `__parallel_merge_submitter_large::operator()` :
```C++
return __future(std::move(__event), std::move(__p_result_and_scratch_storage_base));
```

So what is your question?

SergeyKopienko: About your suggestion "So, we may just writing return __future(__event); in the both cases." - it's absolutely incorrect approach. Please read the description of this PR.

SergeyKopienko: Discussed offline.

MikeDvorskiy: yes, sorry for my mistake..
__p_result_and_scratch_storage_base is defined on line 402 (with other meaning), not 396...

SergeyKopienko: No problems.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,252 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+        const nd_range_params __nd_range_params = eval_nd_range_params(__exec, __rng1, __rng2);
+
+        // Create storage for save split-points on each base diagonal + 1 (for the right base diagonal in the last work-group)
+        auto __p_base_diagonals_sp_global_storage =
+            new __result_and_scratch_storage<_ExecutionPolicy, _split_point_t<_IdType>>(
+                __exec, 0, __nd_range_params.base_diag_count + 1);
+
+        // Save raw-pointer into shared_ptr for return it in __future to exted life-time of the storage
+        __result_and_scratch_storage_base_ptr __p_result_and_scratch_storage_base(
+            static_cast<__result_and_scratch_storage_base*>(__p_base_diagonals_sp_global_storage));
```
MikeDvorskiy, line 403: I guess we don't need  `static_cast` here... 
Ap pointer to a derived class is implicitly convertible to a pointer on a base class. 
https://godbolt.org/z/Gs8Pzhe3a

SergeyKopienko: Technically it's not the problem in any case. I like this `static_cast`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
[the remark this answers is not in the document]
MikeDvorskiy: I think that change may be reverted because there is not serios motivation/reason to change it here.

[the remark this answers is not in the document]
SergeyKopienko: The question was another - should we capture range instance by reference or by value?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: I believe the most of C/C++ developers aware of "degrees of 2",  and "<< operator"
2^10 = 1K or 1<<10
2^20 = 1M or 1<<20
2^22=4M or 1<<22
...
But I dont insist...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: It is a bad style...Because, it might lead to potential runtime errors.

SergeyKopienko: Could you please explain in details, how it can "break" compiler? And which potential run-time issues you mean?
Thanks.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: > That was the reason why I save old implementation.
> But technically I can use `__find_start_point_in` in all cases: for smaller data sizes

In that case should be a comment in the code why we have two implementations.

SergeyKopienko: I never seen these explanations in out code.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
danhoeflinger: While I believe @MikeDvorskiy is correct that it is technically unnecessary, I think in this specific instance, casting it is reasonable if only to indicate in a way that we will basically never use this object in its specialized form in the subsequent host code, and it is only being kept alive.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,252 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+                        _Storage::__get_usm_or_buffer_accessor_ptr(__base_diagonals_sp_global_acc);
+
+                    _split_point_t<_IdType> __sp =
+                        __global_idx == 0 ? _split_point_t<_IdType>{0, 0} : _split_point_t<_IdType>{__n1, __n2};
+
+                    if (0 < __global_idx && __global_idx < __nd_range_params.base_diag_count)
+                    {
+                        const _IdType __i_elem =
+                            __global_idx * __nd_range_params.steps_between_two_base_diags * __nd_range_params.chunk;
+                        if (__i_elem < __n)
```
MikeDvorskiy, line 327: The subexpression `__nd_range_params.steps_between_two_base_diags * __nd_range_params.chunk ` is a invariant and can be calculated outside parallel_for kernel.

SergeyKopienko: Good catch. Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,252 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+                            __sp = __find_start_point(__rng1, __rng2, __i_elem, __n1, __n2, __comp);
+                    }
+
+                    __base_diagonals_sp_global_ptr[__global_idx] = __sp;
+                });
```
MikeDvorskiy, line 333: I would suggest a little bit re-writtig the lambda body to make the code more  readable and understandable at least:

```
{
    auto __global_idx = __item_id.get_linear_id();

    auto __base_diagonals_sp_global_ptr =
        _Storage::__get_usm_or_buffer_accessor_ptr(__base_diagonals_sp_global_acc);

    if(__global_idx == 0)
        __base_diagonals_sp_global_ptr[__global_idx] = split_point_t<_IdType>{0, 0};
    else if (__i_elem < __n)
        __base_diagonals_sp_global_ptr[__global_idx] = __find_start_point(__rng1, __rng2, __i_elem, __n1, __n2, __comp);
    else //__global_idx == __nd_range_params.base_diag_count
        __base_diagonals_sp_global_ptr[__global_idx] = _split_point_t<_IdType>{__n1, __n2};
}
```

SergeyKopienko: Will we have any profit in this case?

SergeyKopienko: Also I believe your propose isn't equal to the current code logic: you lost the condition `if (__i_elem < __n)`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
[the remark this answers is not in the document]
MikeDvorskiy: As we discussed in the past it depends..
 In given case it doesn't matter I think. So, there is not  motivation/reason to change the code here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: > Will we have any profit in this case? May be performance profit?

As I said "to make the code more readable and understandable at least".

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger, line 525: I am inclined to remove this `using` line, to not "hide" the fact that this is a **shared** pointer.  I think this adds to some of the confusion surrounding this mechanism.  
At the very least, we should use "shared_ptr" in the name, but at that point, the name would just as long as the more descriptive original type expression.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,253 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+    operator()(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2, _Range3&& __rng3, _Compare __comp) const
+    {
+        assert(__rng1.size() > 0 || __rng2.size() > 0);
+
+        _PRINT_INFO_IN_DEBUG_MODE(__exec);
+
+        // Calculate nd-range params
+        const nd_range_params __nd_range_params = eval_nd_range_params(__exec, __rng1, __rng2);
+
+        // Create storage for save split-points on each base diagonal + 1 (for the right base diagonal in the last work-group)
```
MikeDvorskiy, line 397: `Calculate nd-range params` => `Calculate nd-range parameters`
`for save `= > `to save`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,253 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+        auto __p_base_diagonals_sp_global_storage =
+            new __result_and_scratch_storage<_ExecutionPolicy, _split_point_t<_IdType>>(
+                __exec, 0, __nd_range_params.base_diag_count + 1);
+
+        // Save raw-pointer into shared_ptr for return it in __future to extend life-time of the storage
```
MikeDvorskiy, line 402: => "// Save the raw pointer into a shared_ptr to return it in __future and extend the lifetime of the storage."

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Fixed!

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Fixed!

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: Ok, the main argument: "is correct that it is technically unnecessary".
If we want to pay developer's attention on that i would prefer writing a comment.

danhoeflinger: I'm fine with either outcome.

SergeyKopienko: Fine, As an author of this PR I prefer to save this place as is too.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,253 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+__get_starting_size_limit_for_large_submitter()
+{
+    return 4 * 1'048'576; // 4 MB
+}
+
+template <>
+constexpr std::size_t
+__get_starting_size_limit_for_large_submitter<int>()
+{
+    return 16 * 1'048'576; // 8 MB
```
MikeDvorskiy, line 438: `// 8 MB` =>  `// 16 MB`

SergeyKopienko: Fixed,

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
danhoeflinger: if `__global_idx * __base_diag_chunk < __n` is `true`, do we know that `__global_idx < __nd_range_params.base_diag_count` is also `true`?

The two if statements here where one only protects the single multiplication seems fishy to me. If we can consolidate it down to a single if statement then we end up with something pretty similar to @MikeDvorskiy's formulation.  

A multiplication should be less expensive than a branch, so I don't see the need to protect it.  If we need both conditions, I still prefer to consolidate to a single if statement and just use `&&`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +236,253 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+        using _RangeValueType = std::conditional_t<(sizeof(_Range1ValueType) > sizeof(_Range2ValueType)),
+                                                   _Range1ValueType, _Range2ValueType>;
+
+        const std::size_t __n = __rng1.size() + __rng2.size();
+
+        // Empirical number of values to process per work-item
+        const std::uint8_t __chunk = __exec.queue().get_device().is_cpu() ? 128 : 4;
+
+        const _IdType __steps = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __chunk);
+        const _IdType __base_diag_count = 32 * 1'024;
```
danhoeflinger, line 289: Similar to our previous discussion with SLM loading between base diagonals, it seems to me that the ideal distance between base diagonals probably has some minimum to provide enough work for the partitioning kernel, but also probably should be tied to the value of `__n`, to optimize the second kernel.  I would expect that there is some explicit fixed distance between base diagonals which is ideal (for a given hardware).  Right now, the distance between base diagonals grows with `__n` since we have a fixed number.    

I guess as written, this could be `constexpr`, and if it is important for optimization reasons for it to be a known quantity, we could perhaps adjust it at compile time based upon `_IdType`? 

At least we have some ballpark understanding of size based upon the type we are using for ids.

This isn't required, but if it makes sense, lets list this as a issue / TODO for the future.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Fixed by another way: please take a look.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -518,8 +518,13 @@ struct __usm_or_buffer_accessor
     }
 };
```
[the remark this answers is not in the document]
SergeyKopienko: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: TODO applied for now. Probably will take a look again after merge into main branch.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -286,6 +286,7 @@ struct __parallel_merge_submitter_large<_IdType, _CustomName,
         const std::uint8_t __chunk = __exec.queue().get_device().is_cpu() ? 128 : 4;
 
         const _IdType __steps = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __chunk);
+        // TODO required to evaluate this value besed on available SLM size for each work-group.
```
danhoeflinger, line 289:
```suggestion
        // TODO required to evaluate this value based on available SLM size for each work-group.
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
danhoeflinger: Sure, looks fine.  A ternary version of what Mikhail proposed I think.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -78,6 +80,104 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
...
+// 0 | 0 | 1  1  1   1
+//   |   |
+// 0 | 0 | 1  1  1   1
+//   |   ---------->
+// 2 | 0   0  0  0 | 1
+//   |             ---->
+// 3 | 0   0  0  0   0 |
+template <typename _Rng1, typename _Rng2, typename _Index, typename _Compare>
+_split_point_t<_Index>
+__find_start_point_in(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_to, const _Rng2& __rng2,
```
danhoeflinger, line 97: It seems like we should be able to use the existing `__find_start_point` simply by restricting the ranges via `drop_view_simple`, by their `__rng*_from` value, passing in an appropriate `_n1` and `__n2` to match the tile size, and fixing `__i_elem` to be relative to the previous bounding base diagonal.  
Similar to what is done in merge sort here https://github.com/oneapi-src/oneDPL/blob/7150d5c26dab8de50052f6c5e5064a229203b989/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h#L270-L278 .

I would be surprised / interested to understand why if that is worse performing than this additional implementation.

danhoeflinger:
Other than the comment about `__find_start_point`, I think this is in good shape and I have no other comments.  
Hopefully we can find resolution there, and without loss of performance.

mmichel11:
I have taken another look through the PR, and I do not see any other issues from my side. Probably someone else who has been more involved in the discussions of this algorithm should be the final approver.

One thing that I think would be good would be to understand why the alternative SLM-based approach in https://github.com/oneapi-src/oneDPL/pull/1937 did not perform as well as expected, but it should probably be looked into after the milestone's completion.

SergeyKopienko:
> Other than the comment about `__find_start_point`, I think this is in good shape and I have no other comments. Hopefully we can find resolution there, and without loss of performance.

I have fixed two moments in the `__find_start_point_in` function and delete `__find_start_point` and it's usages at all.
My checks shown me that there is no performance degradation at this step. only profit.

So I propose to save `__find_start_point_in` only like in mt last commit.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: I don't think so:
1) the `oneapi::dpl::__ranges::drop_view_simple` has his own implementation problems: for example it's returns incorrect `size()` result - it's returns the size of initial range.
2) please see my comment at https://github.com/oneapi-src/oneDPL/pull/1933#issuecomment-2556658545 - I rechecked and think that `__find_start_point_in` is ok now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: @MikeDvorskiy hope you like the new variant.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -45,37 +47,87 @@ namespace __par_backend_hetero
...
+    // Run search of split point on diagonal
+
+    using __it_t = oneapi::dpl::counting_iterator<_Index>;
+
+    __it_t __diag_it_begin(idx1_from);
+    __it_t __diag_it_end(idx1_to);
+
+    constexpr bool kValue = false;
+    const __it_t __res = std::lower_bound(__diag_it_begin, __diag_it_end, kValue,
+                                          [&__rng1, &__rng2, __index_sum, __comp](_Index __idx, const bool __value) mutable {
```
SergeyKopienko, line 126: `mutable` is required here.

SergeyKopienko, line 126: Better to pass `__value` by value (not by reference).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -45,37 +47,87 @@ namespace __par_backend_hetero
...
+                                              return __value == __comp(__rng2[__index_sum - __idx], __rng1[__idx]);
```
SergeyKopienko, line 127: It's enough here to compare the result of `__comp` call with the `__value` for equality.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -45,37 +47,87 @@ namespace __par_backend_hetero
...
-                                 const auto __zero_or_one = __comp(__rng2[__n2 - __i_diag - 1], __rng1[__q + __i_diag]);
-                                 return __zero_or_one < __value;
-                             });
-        return std::make_pair(__q + *__res, __n2 - *__res);
-    }
+    using _IndexSigned = std::make_signed_t<_Index>;
+
+    ////////////////////////////////////////////////////////////////////////////////////
+    // Taking into account the specified constraints of the range of processed data
+    const _IndexSigned __index_sum = __i_elem - 1;
```
SergeyKopienko, line 98: The `__index_sum` variable should have signed type here for the case if this implementation will be called with parameter `__i_elem == 0`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -45,37 +47,88 @@ namespace __par_backend_hetero
...
 
-    if (__i_elem < __n2) //a condition to specify upper or lower part of the merge matrix to be processed
-    {
-        const _Index __q = __i_elem;                         //diagonal index
-        const _Index __n_diag = std::min<_Index>(__q, __n1); //diagonal size
-        auto __res =
-            std::lower_bound(__diag_it, __diag_it + __n_diag, 1 /*value to find*/,
-                             [&__rng2, &__rng1, __q, __comp](const auto& __i_diag, const auto& __value) mutable {
-                                 const auto __zero_or_one = __comp(__rng2[__q - __i_diag - 1], __rng1[__i_diag]);
-                                 return __zero_or_one < __value;
```
SergeyKopienko, line 63: What C++17 standard says about this:
``` If the source type is bool, the value false is converted to zero and
the value true is converted to one.```
But in Microsoft C++ compiler we still have warning in these code for C++17 too.

File: include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h [hunk 33]
```diff
@@ -45,37 +47,88 @@ namespace __par_backend_hetero
...
+    using __it_t = oneapi::dpl::counting_iterator<_Index>;
+
+    __it_t __diag_it_begin(idx1_from);
+    __it_t __diag_it_end(idx1_to);
+
+    constexpr bool kValue = false;
+    const __it_t __res =
+        std::lower_bound(__diag_it_begin, __diag_it_end, kValue,
+                         [&__rng1, &__rng2, __index_sum, __comp](_Index __idx, const bool __value) mutable {
+                             return __value == __comp(__rng2[__index_sum - __idx], __rng1[__idx]);
```
SergeyKopienko, line 128: Based on the note described at https://github.com/oneapi-src/oneDPL/pull/1933/files#r1893974883.
for me the better way simple to call `==` here and compare result of `__comp` with `bool` value.
This approach doesn't' gave us any warnings here anymore.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -45,37 +47,88 @@ namespace __par_backend_hetero
...
+    // Run search of split point on diagonal
+
+    using __it_t = oneapi::dpl::counting_iterator<_Index>;
+
+    __it_t __diag_it_begin(idx1_from);
+    __it_t __diag_it_end(idx1_to);
+
+    constexpr bool kValue = false;
+    const __it_t __res =
+        std::lower_bound(__diag_it_begin, __diag_it_end, kValue,
```
danhoeflinger, line 126: Nitpick, but do we gain anything from having `kValue`?  can we just pass `false`, seems more clear to me?

SergeyKopienko: Fixed

mmichel11 (approved):
LGTM. Great optimization work! Please wait for the final reviews from others as well.

danhoeflinger (approved):
LGTM.  
I think it worked out nicely.
As mentioned by @mmichel11, I still think there are couple places we have room for more optimization with more effort but this is a big step in the right direction.

Thanks @SergeyKopienko !

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -136,47 +188,250 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+                        _Storage::__get_usm_or_buffer_accessor_ptr(__base_diagonals_sp_global_acc);
+
+                    const _IdType __i_elem = __global_idx * __base_diag_chunk;
+
+                    __base_diagonals_sp_global_ptr[__global_idx] =
+                        __i_elem == 0 ? _split_point_t<_IdType>{0, 0}
+                                      : (__i_elem < __n ? __find_start_point(__rng1, _IdType{0}, __n1, __rng2,
+                                                                             _IdType{0}, __n2, __i_elem, __comp)
+                                                        : _split_point_t<_IdType>{__n1, __n2});
+                });
```
MikeDvorskiy, line 282: `? :` operator is used instead of `if.. else` in my proposal. It is by performance reasons? I've just interesting in...

SergeyKopienko: The implementation of `__serial_merge` on ternary operator gave us good profit. So I hope some small profit we may have in this place too. But I am not ready to prepare any precision data about this.

MikeDvorskiy: I see.

MikeDvorskiy (approved):
LGTM

PR #1935, opened 2024-11-14
Remove usage of `__kernel_name_generator` as not required for non-compiled Kernel's

SergeyKopienko:
In this PR we remove usage of `__kernel_name_generator` from `__parallel_find_or` as not required.
We should avoid to use `__kernel_name_generator` for not-compiler Kernels (when absent the call of `__kernel_compiler` in the code) due a lot of char codes in template names.
For example we may see these names with char codes from VTune.

### Attention :
- when you will check the differences for this PR, please **Hide whitespace**.
Otherwise you will see a lot of unsignificant changes due the formatting has been changed.
Thanks.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1837,100 +1837,46 @@ struct __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backe
 };
 #endif // !_ONEDPL_FPGA_EMU
 
-// Base pattern for __parallel_or and __parallel_find. The execution depends on tag type _BrickTag.
-template <typename KernelName, bool __or_tag_check, typename _ExecutionPolicy, typename _BrickTag,
-          typename __FoundStateType, typename _Predicate, typename... _Ranges>
-__FoundStateType
-__parallel_find_or_impl_one_wg(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec,
-                               _BrickTag __brick_tag, const std::size_t __rng_n, const std::size_t __wgroup_size,
-                               const __FoundStateType __init_value, _Predicate __pred, _Ranges&&... __rngs)
-{
-    using __result_and_scratch_storage_t = __result_and_scratch_storage<_ExecutionPolicy, __FoundStateType>;
-    __result_and_scratch_storage_t __result_storage{__exec, 1, 0};
-
-    // Calculate the number of elements to be processed by each work-item.
-    const auto __iters_per_work_item = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __wgroup_size);
-
-    // main parallel_for
-    auto __event = __exec.queue().submit([&](sycl::handler& __cgh) {
-        oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
-        auto __result_acc =
-            __result_storage.template __get_result_acc<sycl::access_mode::write>(__cgh, __dpl_sycl::__no_init{});
-
-        __cgh.parallel_for<KernelName>(
-            sycl::nd_range</*dim=*/1>(sycl::range</*dim=*/1>(__wgroup_size), sycl::range</*dim=*/1>(__wgroup_size)),
-            [=](sycl::nd_item</*dim=*/1> __item_id) {
-                auto __local_idx = __item_id.get_local_id(0);
-
-                // 1. Set initial value to local found state
-                __FoundStateType __found_local = __init_value;
-
-                // 2. Find any element that satisfies pred
-                //  - after this call __found_local may still have initial value:
-                //    1) if no element satisfies pred;
-                //    2) early exit from sub-group occurred: in this case the state of __found_local will updated in the next group operation (3)
-                __pred(__item_id, __rng_n, __iters_per_work_item, __wgroup_size, __found_local, __brick_tag, __rngs...);
-
-                // 3. Reduce over group: find __dpl_sycl::__minimum (for the __parallel_find_forward_tag),
-                // find __dpl_sycl::__maximum (for the __parallel_find_backward_tag)
-                // or update state with __dpl_sycl::__any_of_group (for the __parallel_or_tag)
-                // inside all our group items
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1837,100 +1837,46 @@ struct __parallel_find_or_nd_range_tuner<oneapi::dpl::__internal::__device_backe
...
-    const auto __iters_per_work_item = oneapi::dpl::__internal::__dpl_ceiling_div(__rng_n, __n_groups * __wgroup_size);
-
-    // scope is to copy data back to __result after destruction of temporary sycl:buffer
+template <typename... KernelName, bool __or_tag_check>
+struct __parallel_find_or_impl_one_wg<__internal::__optional_kernel_name<KernelName...>, __or_tag_check>
+{
+    template <typename _ExecutionPolicy, typename _BrickTag, typename __FoundStateType, typename _Predicate,
+              typename... _Ranges>
+    __FoundStateType
+    operator()(_ExecutionPolicy&& __exec, _BrickTag __brick_tag, const std::size_t __rng_n,
```
danhoeflinger, line 1850: I see that we are removing the device backend tag here and below as a side effect of this change.  I understand that it is not technically necessary as device backend is the only option at this point.  
However, we decided to include it originally.  Is it useful as a form of documentation still (this is my only guess as to why it was originally included)?

danhoeflinger (dismissed):
This seems like a reasonable change.  I agree that the `__kernel_name_generator` should not be necessary here because we have no need to pre-compile the kernel and interrogate characteristics of the compiled kernel, and therefore shouldn't be used.  

I'm OK with removing the `device_backend_tag` argument as well, but I want to draw attention to it and make sure we consider that, as it isn't in the original goal of the PR.

It would be good to get another approval here, but I think it looks good.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Originally, I made it myself during tag dispatch implementation. But at fresh view it simple doesn't make sense at all.

SergeyKopienko:
Filed spelling check will fixed in separate PR https://github.com/uxlfoundation/oneDPL/pull/1998

danhoeflinger (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1935,8 +1945,11 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
         // We shouldn't have any restrictions for _AtomicType type here
         // because we have a single work-group and we don't need to use atomics for inter-work-group communication.
 
+        using _KernelName =
```
rarutyun, line 1948: You use the same name in both if and else branches. Today it works but in principle, it's not necessary. I would suggest to rename it to the proper name for easier refactoring in the future.

```suggestion
        using __find_or_one_wg_kernel_name = 
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1935,8 +1945,11 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+            oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__find_or_kernel_one_wg<_CustomName>>;
+
         // Single WG implementation
-        __result = __parallel_find_or_impl_one_wg<_FindOrKernelOneWG, __or_tag_check>(
+        __result = __parallel_find_or_impl_one_wg<_KernelName, __or_tag_check>()(
```
rarutyun, line 1952:
```suggestion
        __result = __parallel_find_or_impl_one_wg<__find_or_one_wg_kernel_name, __or_tag_check>()(
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1945,8 +1958,11 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
         assert("This device does not support 64-bit atomics" &&
                (sizeof(_AtomicType) < 8 || __exec.queue().get_device().has(sycl::aspect::atomic64)));
 
+        using _KernelName =
```
rarutyun, line 1961:
```suggestion
        using __find_or_kernel_name =
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1945,8 +1958,11 @@ __parallel_find_or(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
+            oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<__find_or_kernel<_CustomName>>;
+
         // Multiple WG implementation
-        __result = __parallel_find_or_impl_multiple_wgs<_FindOrKernel, __or_tag_check>(
+        __result = __parallel_find_or_impl_multiple_wgs<_KernelName, __or_tag_check>()(
```
rarutyun, line 1965:
```suggestion
        __result = __parallel_find_or_impl_multiple_wgs<__find_or_kernel_name, __or_tag_check>()(
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Done

rarutyun (approved):
I hope I didn't miss anything, but the PR looks solid to me

PR #1937, opened 2024-11-18
Optimize `merge` algorithm for data sizes equal or greater then 4M items with SLM cache usage

SergeyKopienko:
One more approach for https://github.com/oneapi-src/oneDPL/pull/1933

Unfortunately this approach doesn't gave us performance profit in comparison with https://github.com/oneapi-src/oneDPL/pull/1933

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -299,7 +299,7 @@ struct __parallel_merge_submitter_large<_IdType, _CustomName,
         _PRINT_INFO_IN_DEBUG_MODE(__exec);
 
         // Empirical number of values to process per work-item
-        const _IdType __chunk = __exec.queue().get_device().is_cpu() ? 128 : 8;
+        const _IdType __chunk = __exec.queue().get_device().is_cpu() ? 128 : 4;
         assert(__chunk > 0);
 
         // Pessimistically only use half of the memory to take into account memory used by compiled kernel
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -631,24 +631,12 @@ __parallel_merge(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy
     {
         static_assert(__starting_size_limit_for_large_submitter < std::numeric_limits<std::uint32_t>::max());
 
-        if (__n <= std::numeric_limits<std::uint16_t>::max())
-        {
-            using _WiIndex = std::uint16_t;
-            using _MergeKernelName = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-                __merge_kernel_name<_CustomName, _WiIndex>>;
-            return __parallel_merge_submitter<_WiIndex, _MergeKernelName>()(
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -166,34 +272,394 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+                        __rng2_cache_slm = std::addressof(__rng2_loc_acc[0]) + __offset_to_slm2;
+
+                        // Cooperative data load from __rng1 to __rng1_cache_slm, from __rng2 to __rng1_cache_slm
+                        load_data_into_slm(__rng1, __rng1_cache_slm, __sp_base_left_global.first,  __sp_base_right_global.first,
+                                           __rng2, __rng2_cache_slm, __sp_base_left_global.second, __sp_base_right_global.second,
+                                           __wi_in_one_wg, __local_id);
+                    }
+
+                    // Wait until all the data is loaded (if we have more then one item in work-group
+                    if (__wi_in_one_wg > 1)
```
danhoeflinger, line 579: This is not the issue, but it seems that having a branch here for this degenerate case is probably best avoided at a higher level, or not worth caring about.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -166,34 +272,408 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+            {
+                const auto __slm_idx_begin = __local_id * __loading_data_per_wi;
+                const auto __slm_idx_end = __slm_idx_begin + __loading_data_per_wi;
+
+                std::size_t __slm_idx = __slm_idx_begin;
+                std::size_t __rng_idx = __idx_global_begin + __slm_idx;
+
+                _ONEDPL_PRAGMA_UNROLL
+                for (; __slm_idx < __slm_idx_end && __rng_idx < __idx_global_end; ++__slm_idx, ++__rng_idx)
+                    __slm[__slm_idx] = __rng[__rng_idx];
```
mmichel11, line 389: I believe the global memory access pattern here from `__rng` is poor. This may be a big bottleneck. For a good access pattern, we want all items in a sub-group to access adjacent indices from `__rng` in each iteration of the loop.

My thoughts are that we should ensure good global access patterns first, and then profile and optimize SLM accesses to minimize bank conflicts.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -166,34 +272,408 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
...
+                                              __local_id
+           
+            "+" - load one source data item ito SLM
+        */
+
+        const std::size_t __wg_data_size_rng = __idx_global_end - __idx_global_begin;
+        if (__wg_data_size_rng > 0)
+        {
+            // Calculate the size of the current part of merging data per work-item
+            const std::size_t __loading_data_per_wi = oneapi::dpl::__internal::__dpl_ceiling_div(__rng.size(), __wi_in_one_wg);
```
mmichel11, line 376: Could you clarify this for me if I am not following the code correctly? `__rng` looks to be the input buffer, so `__loading_data_per_wi` will be very large for large input. Is this correct?

mmichel11 (commented):
First comments from a first look. I plan to look deeper tomorrow.

PR #1941, opened 2024-11-19
[Github Config] Launch testing jobs only with code changes 

danhoeflinger:
Skip testing for PRs which don't affect code / testing.

Require that they touch files in one of the following paths:
```
      - '.github/**/*'
      - 'cmake/**/*'
      - 'include/**/*'
      - 'test/**/*'
 ```
Always run codespell and documentation jobs.

### CODE REVIEW

File: .github/workflows/ci-testing.yml [hunk 1]
```diff
@@ -0,0 +1,320 @@
...
+  pull_request:
+    branches:
+      - release_oneDPL
+      - main
+      - 'release/**'
+    paths:
+      - '.github/**/*'
+      - 'cmake/**/*'
+      - 'include/**/*'
+      - 'test/**/*'
```
danhoeflinger, line 15: This is the important addition, everything else is just separating / moving the jobs as is to a new workflow

danhoeflinger:
Please see https://github.com/oneapi-src/oneDPL/actions/runs/11918913730
and https://github.com/oneapi-src/oneDPL/actions 
To see that before the `.github` folder was added to the paths list, that only the documentation workflow was launched (codespell and documentation), once I added the .github folder to the path filter, all jobs were launched (see current CI actions).

.github/workflows/ci-testing.yml
MikeDvorskiy: I guess, we should add `examples `and `make` as well.

MikeDvorskiy: 1) about make:
I don't know reasons why 'make'  would not work.
Make configuration still works, I use it.
But if CI doesn't use make files, we can exclude that folder, of course.

2) "we don't test the examples in the CI."..  
Interesting, it was done intentionally to minimize `CI per commit` time or just we forgot about that?

danhoeflinger: > 1. about make:
>    I don't know reasons why 'make'  would not work.
>    Make configuration still works, I use it.
>    But if CI doesn't use make files, we can exclude that folder, of course.

As far as I understand, the CI generates "fresh" build configuration using `cmake`, then uses that build configuration to build the tests.  Of course, there is nothing wrong with using the make config if it works, it just has not changed much in the last few years and I don't actively use it.  So I'm not aware of its current support for all cases.

> 2. "we don't test the examples in the CI."..
>    Interesting, it was done intentionally to minimize `CI per commit` time or just we forgot about that?

Its possible we build and run the examples and I am unaware of it.  I can try to look into this.  It would be a good idea to do if we aren't already.  I suppose we can add it in here in anticipation of adding these to the CI in the future.

danhoeflinger: Good point about `CMakeLists.txt` ... 
Let me add that.  
I'm OK with leaving out `examples` for now, and adding it if we add it in the future to the CI.

.github/workflows/ci.yml
```diff
@@ -1,4 +1,4 @@
-name: oneDPL CI
+name: oneDPL CI Docs
```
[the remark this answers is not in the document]
danhoeflinger: Updated this.

.github/workflows/ci-testing.yml
```diff
@@ -0,0 +1,320 @@
...
+permissions: read-all
+
+env:
+  BUILD_CONCURRENCY: 4
+  MACOS_BUILD_CONCURRENCY: 3
+  TEST_TIMEOUT: 360
+  WINDOWS_TBB_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/8222de4d-2c5e-41bc-bdd5-f557a9edda28/w_tbb_oneapi_p_2021.13.0.627_offline.exe
+  WINDOWS_ICPX_DOWNLOAD_LINK: https://registrationcenter-download.intel.com/akdlm/IRC_NAS/7991e201-ca0f-4689-bdb6-1ed73a8246fd/w_dpcpp-cpp-compiler_p_2024.2.0.491_offline.exe
+  WINDOWS_ONEAPI_PATH: C:\Program Files (x86)\Intel\oneAPI
+  LINUX_ONEAPI_PATH: /opt/intel/oneapi
```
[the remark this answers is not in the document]
danhoeflinger: Cleaned these, and actually all the env variables are unused in the docs so I removed them

danhoeflinger:
@timmiesmith I updated the badge support on the Readme.md to use the badge from the testing workflow.  Do you want me to add one for the docs as well?  I don't want to clutter it.

dmitriy-sobolev (approved):
LGTM. Could you wait for the CI completion before merging?

PR #1949, opened 2024-11-25
[Cmake] Make SYCL detection more portable

dmitriy-sobolev:
Let's use -fsycl only if it is supported by a compiler.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,8 +89,11 @@ if (NOT DEFINED ONEDPL_BACKEND OR ONEDPL_BACKEND MATCHES "^(dpcpp|dpcpp_only)$")
     if (NOT SYCL_HEADER)
         CHECK_INCLUDE_FILE_CXX("CL/sycl.hpp" SYCL_HEADER_OLD ${FSYCL_OPTION_IF_SUPPORTED})
     endif()
-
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -75,26 +75,35 @@ include(CheckCXXCompilerFlag)
...
+            message(STATUS "SYCL is not supported")
+        endif()
+    endif()
 endif()
 
 # Set the default backend if one has not been explicitly provided
-if (NOT ONEDPL_BACKEND)
-    if (_sycl_support)
+if (NOT DEFINED ONEDPL_BACKEND)
+    if (SYCL_SUPPORT)
```
adamfidel, line 106: I'm confused where this variable `SYCL_SUPPORT` is coming from. Should this condition instead be checking if `FSYCL_OPTION_IF_SUPPORTED` is set to a non-empty value?

dmitriy-sobolev: Good catch. It was rather a refactoring mistake, which could be avoided by proper testing. Now `SYCL_SUPPORT` is set.

mmichel11 (approved):
This approach LGTM.

adamfidel (approved):
LGTM now as well!

PR #1954, opened 2024-12-03
Fall back to SYCL 2020 API for a generic SYCL implementation

dmitriy-sobolev:
Let's assume that a generic SYCL library is compliant to SYCL 2020 and fallback to that API.

Below is a specific example of the issue with `sycl::property::no_init`. Other places fixed in the PR are essentially the same. 

Before:

```c++
#define _ONEDPL_NO_INIT_PRESENT (_ONEDPL_LIBSYCL_VERSION >= 50300)
...
using __no_init =
#if _ONEDPL_NO_INIT_PRESENT
    // SYCL 2020 API; Will be selected only with newer DPC++ versions
    sycl::property::no_init;
#else
    // Specific to older DPC++; Will be selected with older DPC++ and other SYCL implementations
    sycl::property::noinit; 
#endif
```

After:
```c++

#define _ONEDPL_LIBSYCL_VERSION_LESS_THAN(_ONEDPL_SYCL_VERSION_NUMBER)                                                 \
    (_ONEDPL_LIBSYCL_VERSION && _ONEDPL_LIBSYCL_VERSION < _ONEDPL_SYCL_VERSION_NUMBER)
#define _ONEDPL_SYCL2020_NO_INIT_PRESENT (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
...
using __no_init =
#if _ONEDPL_SYCL2020_NO_INIT_PRESENT
    // SYCL 2020 API; Will be selected with newer DPC++ versions and other SYCL implementations
    sycl::property::no_init;
#elif _ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300)
    // Specific to older DPC++; Will be selected only with older DPC++
    sycl::property::noinit;
#endif

```
Meanwhile, the conditions relying on DPC++ extensions remained the same, e.g. FPGA support or ESIMD. 

------------------
Additional changes:

Some macros were renamed to have SYCL2020 or LIBSYCL in their names to denote their purpose (either SYCL 2020 feature or dependency on DPC++ SYCL implementation/extension). 

`_ONEDPL_SYCL2020_UNIFIED_USM_BUFFER_PRESENT` was renamed to `_ONEDPL_SYCL2020_DEFAULT_ACCESSOR_CONSTRUCTOR_PRESENT` to show the exact SYCL 2020 feature needed to enable code section. `_ONEDPL_SYCL_L0_EXT_PRESENT` was introduced for getting L0 backend with different SYCL implementations.


include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -37,36 +37,40 @@
 
 // Combine SYCL runtime library version
 #if defined(__LIBSYCL_MAJOR_VERSION) && defined(__LIBSYCL_MINOR_VERSION) && defined(__LIBSYCL_PATCH_VERSION)
-#    define _ONEDPL_LIBSYCL_VERSION                                                                                    \
+#    define _ONEDPL_DPCPP_LIBSYCL_VERSION                                                                              \
         (__LIBSYCL_MAJOR_VERSION * 10000 + __LIBSYCL_MINOR_VERSION * 100 + __LIBSYCL_PATCH_VERSION)
 #else
-#    define _ONEDPL_LIBSYCL_VERSION 0
+#    define _ONEDPL_DPCPP_LIBSYCL_VERSION 0
 #endif
```
akukanov, line 44: > _ONEDPL_LIBSYCL_VERSION was renamed to _ONEDPL_DPCPP_LIBSYCL_VERSION as the definition uses macros specific to DPC++ SYCL implementation.

Does not the same library come with the intel/llvm SYCL compiler as well?
I personally do not consider "libsycl" as a name for any SYCL runtime, but rather the one of DPC++ and intel/llvm.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort.h
```diff
@@ -346,7 +346,7 @@ enum class __peer_prefix_algo
 template <std::uint32_t __radix_states, typename _OffsetT, __peer_prefix_algo _Algo>
 struct __peer_prefix_helper;
 
-#if (_ONEDPL_LIBSYCL_VERSION >= 50700)
+#if _ONEDPL_DPCPP_LIBSYCL_ZERO_OR_GE(50700)
```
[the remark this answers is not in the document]
dmitriy-sobolev: I've followed the suggestion where I found it to be applicable (see [7fcd1b2](https://github.com/oneapi-src/oneDPL/pull/1954/commits/7fcd1b2b31d062fcab5ca7ae82998a517372c305)).

dmitriy-sobolev:
Turning into draft to make the approach more scalable.

akukanov (commented):
The approach looks good to me in principle, but I have not checked thoroughly that all the new definitions match the old ones.

test/general/test_policies.pass.cpp
```diff
@@ -75,7 +75,7 @@ main()
 
     // make_device_policy
     test_policy_instance(TestUtils::make_device_policy<class Kernel_11>(q));
-#if _ONEDPL_LIBSYCL_VERSION < 60000
+#if _ONEDPL_LIBSYCL_VERSION_LESS_THAN(60000)
```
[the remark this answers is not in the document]
dmitriy-sobolev: I took a deeper look and got rid of the case, creating #1965 to track the related issue. 
This case covers API specified in the documentation, but it works only with the old DPC++ versions, which is suspicious.

dmitriy-sobolev:
@MikeDvorskiy could you also take a look at the PR? The macros can be tricky and prone to subtle issues, so the second reviewer can help a lot.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -39,9 +39,8 @@
 #if defined(__LIBSYCL_MAJOR_VERSION) && defined(__LIBSYCL_MINOR_VERSION) && defined(__LIBSYCL_PATCH_VERSION)
 #    define _ONEDPL_LIBSYCL_VERSION                                                                                    \
         (__LIBSYCL_MAJOR_VERSION * 10000 + __LIBSYCL_MINOR_VERSION * 100 + __LIBSYCL_PATCH_VERSION)
-#else
-#    define _ONEDPL_LIBSYCL_VERSION 0
 #endif
+#define _ONEDPL_LIBSYCL_VERSION_LESS_THAN(_VERSION) (_ONEDPL_LIBSYCL_VERSION && _ONEDPL_LIBSYCL_VERSION < _VERSION)
```
SergeyKopienko, line 43: I propose to define instead of this macro or as additional macros something like
`_ONEDPL_LIBSYCL_VERSION_GREAT_EQ_THAN`

The goal - to avoid constructions like
`#define _ONEDPL_SYCL2020_KNOWN_IDENTITY_PRESENT (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))`
(I mean the negotiation).

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -51,23 +50,48 @@
...
+    (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50700) || __SYCL_COMPILER_VERSION > 20201214)
+#define _ONEDPL_SYCL2020_BUFFER_ALLOCATOR_PRESENT (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60000))
+#define _ONEDPL_SYCL2020_LOCAL_ACCESSOR_PRESENT (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60000))
+// The unified future supporting USM memory and buffers is only supported after DPC++ 2023.1 but not by 2023.2.
+#define _ONEDPL_SYCL2020_DEFAULT_ACCESSOR_CONSTRUCTOR_PRESENT                                                          \
+    (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60100) || _ONEDPL_LIBSYCL_VERSION != 60200)
+#define _ONEDPL_SYCL2020_HOST_TARGET_PRESENT (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60200))
+#define _ONEDPL_SYCL2020_HOST_ACCESSOR_PRESENT (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60200))
+#define _ONEDPL_SYCL2020_GET_HOST_ACCESS_PRESENT (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60200))
+#define _ONEDPL_SYCL2020_LOCAL_ACC_GET_MULTI_PTR_PRESENT (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(70000))
```
SergeyKopienko, line 82:
```suggestion
// SYCL 2020 feature macros. They are enabled by default:
// A SYCL implementation is assumed to support the features unless specified otherwise.
// This is controlled by extendable logic: !(A && A < SUPPORTED_VER_A) && !(B && B < SUPPORTED_VER_B) && ...,
// where A, B, etc., are macros representing the version of a specific SYCL implementation.
#define _ONEDPL_SYCL2020_BITCAST_PRESENT                      (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
#define _ONEDPL_SYCL2020_NO_INIT_PRESENT                      (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
#define _ONEDPL_SYCL2020_COLLECTIVES_PRESENT                  (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
#define _ONEDPL_SYCL2020_BUFFER_SIZE_PRESENT                  (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
#define _ONEDPL_SYCL2020_ACCESSOR_SIZE_PRESENT                (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
#define _ONEDPL_SYCL2020_KERNEL_BUNDLE_PRESENT                (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
#define _ONEDPL_SYCL2020_KNOWN_IDENTITY_PRESENT               (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
#define _ONEDPL_SYCL2020_FUNCTIONAL_OBJECTS_PRESENT           (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
#define _ONEDPL_SYCL2020_REQD_SUB_GROUP_SIZE_PRESENT          (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300))
#define _ONEDPL_SYCL2020_TARGET_PRESENT                       (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50400))
#define _ONEDPL_SYCL2020_TARGET_DEVICE_PRESENT                (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50400))
#define _ONEDPL_SYCL2020_ATOMIC_REF_PRESENT                   (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50500))
#define _ONEDPL_SYCL2020_SUB_GROUP_PRESENT                    (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50700))
#define _ONEDPL_SYCL2020_SUBGROUP_BARRIER_PRESENT             (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50700))

// 20201214 value corresponds to DPC++ 2021.1.2
#define _ONEDPL_SYCL2020_KERNEL_DEVICE_API_PRESENT            (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50700) || __SYCL_COMPILER_VERSION > 20201214)
#define _ONEDPL_SYCL2020_BUFFER_ALLOCATOR_PRESENT             (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60000))
#define _ONEDPL_SYCL2020_LOCAL_ACCESSOR_PRESENT               (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60000))
// The unified future supporting USM memory and buffers is only supported after DPC++ 2023.1 but not by 2023.2.
#define _ONEDPL_SYCL2020_DEFAULT_ACCESSOR_CONSTRUCTOR_PRESENT (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60100) || _ONEDPL_LIBSYCL_VERSION != 60200)
#define _ONEDPL_SYCL2020_HOST_TARGET_PRESENT                  (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60200))
#define _ONEDPL_SYCL2020_HOST_ACCESSOR_PRESENT                (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60200))
#define _ONEDPL_SYCL2020_GET_HOST_ACCESS_PRESENT              (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(60200))
#define _ONEDPL_SYCL2020_LOCAL_ACC_GET_MULTI_PTR_PRESENT      (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(70000))
```

SergeyKopienko: The idea - readability here is more significant then formatting.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
akukanov: Unfortunately, "libsycl-greater-or-equal" only applies to the libsycl library, while "not libsycl-less-than" also covers all other libraries.

SergeyKopienko: May be.
But from my point of view this negotiation make defines a little bit heavy for uderstanding.

akukanov: The comment above the block of macros is added to clarify the logic.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -78,9 +102,9 @@
 #    endif
 #endif // _ONEDPL_DETECT_SPIRV_COMPILATION
 
-#if _ONEDPL_LIBSYCL_VERSION >= 50300
+#if _ONEDPL_SYCL2020_REQD_SUB_GROUP_SIZE_PRESENT
 #    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(SIZE) sycl::reqd_sub_group_size(SIZE)
-#else
+#elif _ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300)
```
MikeDvorskiy, line 107: What if both macros evaluated to zero?
1) If it is possible, we don't control the all cases and get UB/error...
2) If it is not possible, I don't understand why do we write the second if  -  `#elif`...

akukanov: Of course we do not control all the cases, because we cannot predict what an arbitrary SYCL implementation can or cannot do, and what workaround it would require.

The optimistic expectation is that fully SYCL-conformant implementations go into the "present" branch with no effort from our side. However if for some implementation that's not true, the case (1) will happen and we will need to analyze the failure and provide a proper workaround, likely in another #elif branch.

akukanov: Specifically for `_ONEDPL_SYCL_REQD_SUB_GROUP_SIZE`, we can define it as no-op in the #else branch. But this might result in runtime errors for the kernels that really require certain WG size, which would be even worse than the guaranteed compilation error.

akukanov: Another approach is to switch from "present" macros to "workaround" macros. I think Dmitry tried something of that kind with "absent" macros in a previous revision of the patch. It would be something like this:

```
#if _ONEDPL_LIBSYCL_REQD_SUB_GROUP_SIZE_WA // == _ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300)
#    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(SIZE) intel::reqd_sub_group_size(SIZE)
#else // SYCL 2020 conformance expected
#    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(SIZE) sycl::reqd_sub_group_size(SIZE)
#endif
```

For an arbitrary implementation of SYCL it is no better than the proposed way, because if `sycl::reqd_sub_group_size` is not supported, there will also be a compilation error. And while the code appears to be slightly simpler, a downside is that the conformant code comes last, not first.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: Applied

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: Let me add more details on top of Alexey's answer.

> What if both macros are evaluated to zero... If it is possible

It is possible.

Let's consider such a case: older versions of FOOSYCL, some SYCL library, do not support `sycl::reqd_sub_group_size`. There is no workaround, or it was simply forgotten to be added. 

```c++
#define _ONEDPL_SYCL2020_REQD_SUB_GROUP_SIZE_PRESENT \
    (!_ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300) && !_ONEDPL_FOOSYCL_VERSION_LESS_THAN(42))

// Current approach: older versions of FOOSYCL will have undefined macro, which may have unpredictable consequences
#if _ONEDPL_SYCL2020_REQD_SUB_GROUP_SIZE_PRESENT
#    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(SIZE) sycl::reqd_sub_group_size(SIZE)
#elif _ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300)
#    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(SIZE) intel::reqd_sub_group_size(SIZE)
#endif

// Possible approach: error will be emitted with the older FOOSYCL versions, but it is more verbose
#if _ONEDPL_SYCL2020_REQD_SUB_GROUP_SIZE_PRESENT
#    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(SIZE) sycl::reqd_sub_group_size(SIZE)
#elif _ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300)
#    define _ONEDPL_SYCL_REQD_SUB_GROUP_SIZE(SIZE) intel::reqd_sub_group_size(SIZE)
#else
#    error "No implementation for reqd_sub_group_size"
#endif
```

Was it your concern, Mikhail? If so, would you prefer "Possible approach" above?

MikeDvorskiy: > we can define it as no-op in the #else branch

We can also define `#pragma error` in the #else branch... 
I think compile time error is better  then runtime error.

MikeDvorskiy: > if `sycl::reqd_sub_group_size` is not supported, there will also be a compilation error

Ok, this behavior is acceptable for me.

MikeDvorskiy: > ```c++
> #    error "No implementation for reqd_sub_group_size"
> ```

Yes, this approach is acceptable to me

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -163,9 +163,9 @@ inline constexpr all_view_fn<sycl::access::mode::read, __dpl_sycl::__target_devi
 inline constexpr all_view_fn<sycl::access::mode::write, __dpl_sycl::__target_device, sycl::access::placeholder::true_t>
     all_write;
 
-#if _ONEDPL_SYCL_PLACEHOLDER_HOST_ACCESSOR_DEPRECATED
+#if _ONEDPL_SYCL2020_HOST_ACCESSOR_PRESENT
 inline constexpr all_host_view_fn
-#else
+#elif _ONEDPL_LIBSYCL_VERSION_LESS_THAN(60200)
 inline constexpr all_view_fn<sycl::access::mode::read_write, __dpl_sycl::__host_target,
                              sycl::access::placeholder::false_t>
```
MikeDvorskiy, line 170: `#else # error ...` ?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
MikeDvorskiy: > Was it your concern, Mikhail? If so, would you prefer "Possible approach" above? It's an edge-case, I am reluctant to do that across the PR due to the verbosity.

Sorry, but it seems I didn't catch what you want to say...

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -93,21 +117,13 @@
...
 {
 
 using __no_init =
-#if _ONEDPL_NO_INIT_PRESENT
+#if _ONEDPL_SYCL2020_NO_INIT_PRESENT
     sycl::property::no_init;
-#else
+#elif _ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300)
     sycl::property::noinit;
 #endif
```
MikeDvorskiy, line 128: `#else #error ..... ` branch ?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -149,19 +165,19 @@ template <typename _T>
 using __minimum = sycl::ONEAPI::minimum<_T>;
 #endif // _ONEDPL_SYCL2020_FUNCTIONAL_OBJECTS_PRESENT
 
-#if _ONEDPL_SYCL_SUB_GROUP_PRESENT
+#if _ONEDPL_SYCL2020_SUB_GROUP_PRESENT
 using __sub_group = sycl::sub_group;
-#else
+#elif _ONEDPL_LIBSYCL_VERSION_LESS_THAN(50700)
 using __sub_group = sycl::ONEAPI::sub_group;
 #endif
```
MikeDvorskiy, line 172: `#else #error ..... ` branch ?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -149,19 +165,19 @@ template <typename _T>
...
 constexpr auto
 __get_buffer_size(const _Buffer& __buffer)
 {
-#if _ONEDPL_LIBSYCL_VERSION >= 50300
+#if _ONEDPL_SYCL2020_BUFFER_SIZE_PRESENT
     return __buffer.size();
-#else
+#elif _ONEDPL_LIBSYCL_VERSION_LESS_THAN(50300)
     return __buffer.get_count();
 #endif
```
MikeDvorskiy, line 182: `#else #error ..... ` branch ?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: > We can also define #pragma error in the #else branch...
I think compile time error is better then runtime error.

Done.

> It's an edge-case, I am reluctant to do that across the PR due to the verbosity.

It was about proliferation of "#else... #error ..." parts, which I assumed as non-necessary. It is not relevant, I changed my mind: it is better to have these parts to avoid UB in some cases.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: Added

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: Added

MikeDvorskiy (approved):
LGTM

PR #1957, opened 2024-12-05
Update version to 2022.7.1

masterleinad:
`main` still shows 2022.7.0 and the same holds for the 2022.7.1 release tag, see https://github.com/oneapi-src/oneDPL/blob/oneDPL-2022.7.1-release/include/oneapi/dpl/internal/version_impl.h. In particular, this hinders me from testing the changes related to the fixes in  https://github.com/oneapi-src/oneDPL/pull/1927 and https://github.com/oneapi-src/oneDPL/pull/1932 (easily).

### DIFF
--- a/include/oneapi/dpl/internal/version_impl.h
+++ b/include/oneapi/dpl/internal/version_impl.h
@@ -13,7 +13,7 @@
 // The library version
 #define ONEDPL_VERSION_MAJOR 2022
 #define ONEDPL_VERSION_MINOR 7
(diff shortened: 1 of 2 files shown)


masterleinad:
It would be great if the version update could at least be incorporated in the corresponding `oneAPI` release.

timmiesmith:
Apologies for this miss.  It is unfortunately too late for the correction to make it into the oneAPI patch release.

PR #1959, opened 2024-12-06
Fix capture modes in `submit` calls

SergeyKopienko:
In this PR we fix capture modes in `submit` calls.

The goal is to avoid extra captured by value variables.

**Justification.**
1) `sycl::submit` always executed synchronously till the call of `parallel_for` inside.
This mean that no reasons to capture `by value` any variables used only inside `submit` call but NOT inside `parallel_for`
2) `sycl::parallel_for` function always captures all variables `by copy` (`[=]`).
This mean that inside lambda,  passed into `sycl::parallel_for` call all references and so on will be in real new local copies of source variables,

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -235,7 +235,7 @@ struct __parallel_for_submitter<__internal::__optional_kernel_name<_Name...>>
     {
         assert(oneapi::dpl::__ranges::__get_first_range_size(__rngs...) > 0);
         _PRINT_INFO_IN_DEBUG_MODE(__exec);
-        auto __event = __exec.queue().submit([&__rngs..., &__brick, __count](sycl::handler& __cgh) {
+        auto __event = __exec.queue().submit([&](sycl::handler& __cgh) {
             //get an access to data under SYCL buffer:
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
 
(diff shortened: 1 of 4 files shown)


SergeyKopienko:
@danhoeflinger you wrote `this` capturing:
```C++
return __exec.queue().submit([&, this](sycl::handler& __cgh) {
```
It's technically correct but a little bit tricky.
Could you please fix it too in this PR?

mmichel11:
What is the benefit of capturing types with a size less than or equal to the size of a pointer by reference? Throughout oneDPL, we accept small types by value instead of by reference.

If the type is large, then this change makes sense to me.

SergeyKopienko:
> What is the benefit of capturing types with a size less than or equal to the size of a pointer by reference? Throughout oneDPL, we accept small types by value instead of by reference.
> 
> If the type is large, then this change makes sense to me.

Exactly in this place? Inside `parallel_for` we will work with the copies, I mean.

mmichel11:
> > What is the benefit of capturing types with a size less than or equal to the size of a pointer by reference? Throughout oneDPL, we accept small types by value instead of by reference.
> > If the type is large, then this change makes sense to me.
> 
> Exactly in which place? Inside `parallel_for` we will work with the copies, I mean.

In the outer lambda passed to submit. I assume the capture by reference pulls in the address of the variable. If the size of the type is smaller than the size of the type that holds the address (likely a pointer), then I do not see the benefit.

MikeDvorskiy:
> > > What is the benefit of capturing types with a size less than or equal to the size of a pointer by reference? Throughout oneDPL, we accept small types by value instead of by reference.
> > > If the type is large, then this change makes sense to me.
> > 
> > 
> > Exactly in which place? Inside `parallel_for` we will work with the copies, I mean.
> 
> In the outer lambda passed to submit. I assume the capture by reference pulls in the address of the variable. If the size of the type is smaller than the size of the type that holds the address (likely a pointer), then I do not see the benefit. Although I assume the compiler optimizes either case and there is no real performance concern.

Yes, I believe in a compiler too.

PR #1960, opened 2024-12-09
[RNG] Shortening some test file names to avoid warnings on Windows

egrabovskaya:


test/xpu_api/random/statistics_tests/extreme_value_distribution_dp_portion_test.pass.cpp
```diff
@@ -0,0 +1,183 @@
+// -*- C++ -*-
+//===-- extreme_value_distribution_dp_tests_set_portion_test.cpp ---------------------------------===//
```
aelizaro, line 2: Please, check that new filenames are matched with the comments.
Also, a suggestion: maybe we can shorted `distribution` to `distr` as well, as it still seems very long

akukanov:
The name of the PR suggests that existing tests are going to be changed, but the patch adds a bunch of new tests without removing/renaming any old ones. Is that on purpose?

Generally, it would be better to have a short description of what the patch is supposed to do and why.

PR #1962, opened 2024-12-09
[oneDPL] oneDPL '__future' class  code cleanup and API fix

MikeDvorskiy:
A little bit refactoring and class API fix for oneDPL `future` type.

1. In case of compiler time parameter set: there was no way to get the second and following parameters from `__future` instance via `__future::get` method.
2. In case of runtime parameter set (array/buffer): there was no way to get the second and following elements from buffer , which is incapsulated into `__future` instance via `__future::get` method.
3. There was a `code coupling` between following types: `__future`, 'sycl::buffer', '__result_and_scratch_storage'

The PR fixes the 1,2,3 mentioned issues.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -713,7 +713,8 @@ __wait_and_get_value(_Event&&, const sycl::buffer<_T>& __buf, std::size_t __idx 
 // An overload of __wait_and_get_value for '__result_and_scratch_storage'
 template <typename _Event, typename _ExecutionPolicy, typename _T>
 constexpr auto
-__wait_and_get_value(_Event&& __event, const __result_and_scratch_storage<_ExecutionPolicy, _T>& __storage, std::size_t __idx = 0)
+__wait_and_get_value(_Event&& __event, const __result_and_scratch_storage<_ExecutionPolicy, _T>& __storage,
+                     std::size_t __idx = 0)
 {
     return __storage.__wait_and_get_value(__event, __idx);
 }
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -764,13 +775,14 @@ class __future : private std::tuple<_Args...>
 #endif
     }
 
+    template <std::size_t _Idx = 0>
     auto
-    get()
+    get(std::size_t __idx = 0)
```
SergeyKopienko, line 780: Everybody who will take a look at this function will ask - what is template param and run-time param here,
Explanation required, I think.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -701,31 +701,42 @@ struct __deferrable_mode
 {
 };
 
-//A contract for future class: <sycl::event or other event, a value, sycl::buffers..., or __usm_host_or_buffer_storage>
-//Impl details: inheritance (private) instead of aggregation for enabling the empty base optimization.
+// An overload of __wait_and_get_value for 'sycl::buffer'
```
SergeyKopienko, line 704: Can we modify these function without move them outside?

MikeDvorskiy: The change also solver a `code coupling` issue.
There was a `code coupling` between following types: __future, 'sycl::buffer', '__result_and_scratch_storage'.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: Better to understand this from code...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: But why should move them outside of `class __future` ?

MikeDvorskiy:
In general, the the change doesn't solve a use case when we need to return/get a couple of values from a future objects. 
And there is simpler way - to use, for example,  `std::pair` as T type for `__result_and_scratch_storage` instead of specifying  `2` elements in `__result_and_scratch_storage` constructor. So, I don't see significant sense with the proposed changes and intend close the PR.

PR #1967, opened 2024-12-13
Show relevant CONTRIBUTORS.md on the main page

dmitriy-sobolev:
There previous one, https://github.com/oneapi-src/oneDPL/blob/release_oneDPL/CONTRIBUTING.md, is outdated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ for information on how to report a potential security issue or vulnerability.
 You can also view the [Security Policy](SECURITY.md).
 
 ## Contributing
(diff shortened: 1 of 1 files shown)


danhoeflinger:
I agree with this change.  
I did a search for other instances of `https://github.com/oneapi-src/oneDPL/blob/release_oneDPL`
in the repo for other instances of stale links and found several with the `LICENSE.txt`.  Should we also update those to use "main"?

danhoeflinger (approved):
Feel free to merge this as is.  The changes for License.md are optional (I'm not sure if it has actually changed)

dmitriy-sobolev:
#1968 is a follow-up with `License.md`.

PR #1970, opened 2024-12-17
Optimization of `__serial_merge` function

SergeyKopienko:
In this PR we make optimization of `__serial_merge` function.
The justification: previous implementation is too complicated:
- a lot of conditions;
- a lot of for-loops (and one inside another).

With this new implementation of `__serial_merge` function we have significant performance profit for `merge` and `merge_sort` algorithms on all data sizes.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -98,13 +98,13 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index _
 
     for (_Index __rng3_idx = __start3; __rng3_idx < __rng3_idx_end; ++__rng3_idx)
     {
-        const bool __rng1_idx_less__n1 = __rng1_idx < __rng1_idx_end;
-        const bool __rng2_idx_less__n2 = __rng2_idx < __rng2_idx_end;
+        const bool __rng1_idx_less_n1 = __rng1_idx < __rng1_idx_end;
+        const bool __rng2_idx_less_n2 = __rng2_idx < __rng2_idx_end;
 
         __rng3[__rng3_idx] =
-            __rng1_idx_less__n1 && __rng2_idx_less__n2
+            __rng1_idx_less_n1 && __rng2_idx_less_n2
                 ? (__comp(__rng2[__rng2_idx], __rng1[__rng1_idx]) ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++])
-                : (__rng1_idx_less__n1 ? __rng1[__rng1_idx++] : __rng2[__rng2_idx++]);
+                : (__rng1_idx_less_n1 ? __rng1[__rng1_idx++] : __rng2[__rng2_idx++]);
     }
 }
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -80,7 +80,7 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
 
 // Do serial merge of the data from rng1 (starting from start1) and rng2 (starting from start2) and writing
 // to rng3 (starting from start3) in 'chunk' steps, but do not exceed the total size of the sequences (n1 and n2)
-template <typename _Rng1, typename _Rng2, typename _Rng3, typename _Index, typename _Compare>
+template <unsigned int _UnrollFactor = 4, typename _Rng1, typename _Rng2, typename _Rng3, typename _Index, typename _Compare>
 void
 __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index __start1, _Index __start2,
                const _Index __start3, const std::uint8_t __chunk, const _Index __n1, const _Index __n2, _Compare __comp)
@@ -99,7 +99,7 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index _
     
     bool __rng1_idx_less__n1, __rng2_idx_less__n2;
 
-    #pragma unroll 4
+    #pragma unroll _UnrollFactor
     for (_Index __rng3_idx = __start3; __rng3_idx < __rng3_idx_end; ++__rng3_idx)
     {
         __rng1_idx_less__n1 = __rng1_idx < __rng1_idx_end;
@@ -119,6 +119,8 @@ struct __parallel_merge_submitter;
 template <typename _IdType, typename... _Name>
 struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_Name...>>
 {
+    static constexpr std::uint32_t __gpu_chunk = 4;
+
     template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3, typename _Compare>
     auto
     operator()(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2, _Range3&& __rng3, _Compare __comp) const
@@ -132,7 +134,7 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
         _PRINT_INFO_IN_DEBUG_MODE(__exec);
 
         // Empirical number of values to process per work-item
-        const std::uint8_t __chunk = __exec.queue().get_device().is_cpu() ? 128 : 4;
+        const std::uint8_t __chunk = __exec.queue().get_device().is_cpu() ? 128 : __gpu_chunk;
 
         const _IdType __steps = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __chunk);
 
@@ -141,8 +143,8 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
             __cgh.parallel_for<_Name...>(sycl::range</*dim=*/1>(__steps), [=](sycl::item</*dim=*/1> __item_id) {
                 const _IdType __i_elem = __item_id.get_linear_id() * __chunk;
                 const auto __start = __find_start_point(__rng1, __rng2, __i_elem, __n1, __n2, __comp);
-                __serial_merge(__rng1, __rng2, __rng3, __start.first, __start.second, __i_elem, __chunk, __n1, __n2,
-                               __comp);
+                __serial_merge<__gpu_chunk>(__rng1, __rng2, __rng3, __start.first, __start.second, __i_elem, __chunk,
+                                            __n1, __n2, __comp);
             });
         });
         return __future(__event);
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -83,11 +83,11 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
 template <typename _Rng1, typename _Rng2, typename _Rng3, typename _Index, typename _Compare>
 void
 __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index __start1, _Index __start2,
-               const _Index __start3, const std::uint8_t __chunk, const _Index __n1, const _Index __n2, _Compare __comp)
+               const _Index __start3, const _Index __chunk, const _Index __n1, const _Index __n2, _Compare __comp)
 {
-    const _Index __rng1_size = std::min<_Index>(__n1 > __start1 ? __n1 - __start1 : _Index{0}, (_Index)__chunk);
-    const _Index __rng2_size = std::min<_Index>(__n2 > __start2 ? __n2 - __start2 : _Index{0}, (_Index)__chunk);
-    const _Index __rng3_size = std::min<_Index>(__rng1_size + __rng2_size, (_Index)__chunk);
+    const _Index __rng1_size = std::min<_Index>(__n1 > __start1 ? __n1 - __start1 : _Index{0}, __chunk);
+    const _Index __rng2_size = std::min<_Index>(__n2 > __start2 ? __n2 - __start2 : _Index{0}, __chunk);
+    const _Index __rng3_size = std::min<_Index>(__rng1_size + __rng2_size, __chunk);
 
     const _Index __rng1_idx_end = __start1 + __rng1_size;
     const _Index __rng2_idx_end = __start2 + __rng2_size;
@@ -128,7 +128,7 @@ struct __parallel_merge_submitter<_IdType, __internal::__optional_kernel_name<_N
         _PRINT_INFO_IN_DEBUG_MODE(__exec);
 
         // Empirical number of values to process per work-item
-        const std::uint8_t __chunk = __exec.queue().get_device().is_cpu() ? 128 : 4;
+        const _IdType __chunk = __exec.queue().get_device().is_cpu() ? 128 : 4;
 
         const _IdType __steps = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __chunk);
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -101,6 +101,9 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index _
         const bool __rng1_idx_less_n1 = __rng1_idx < __rng1_idx_end;
         const bool __rng2_idx_less_n2 = __rng2_idx < __rng2_idx_end;
 
+        // Logically, one of __rng1_idx_less_n1 and __rng2_idx_less_n2 should be true here
+        // because 1) we should fill output data with elements from one of the input ranges
+        // 2) we calculate __rng3_idx_end as std::min<_Index>(__rng1_size + __rng2_size, __chunk);
         __rng3[__rng3_idx] =
             __rng1_idx_less_n1 && __rng2_idx_less_n2
                 ? (__comp(__rng2[__rng2_idx], __rng1[__rng1_idx]) ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++])
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -95,7 +95,6 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index _
 
     _Index __rng1_idx = __start1;
     _Index __rng2_idx = __start2;
-    _Index __rng3_idx = __start3;
 
     for (_Index __rng3_idx = __start3; __rng3_idx < __rng3_idx_end; ++__rng3_idx)
     {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -97,13 +97,11 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index _
     _Index __rng2_idx = __start2;
     _Index __rng3_idx = __start3;
     
-    bool __rng1_idx_less__n1, __rng2_idx_less__n2;
-
     #pragma unroll _UnrollFactor
     for (_Index __rng3_idx = __start3; __rng3_idx < __rng3_idx_end; ++__rng3_idx)
     {
-        __rng1_idx_less__n1 = __rng1_idx < __rng1_idx_end;
-        __rng2_idx_less__n2 = __rng2_idx < __rng2_idx_end;
+        const bool __rng1_idx_less__n1 = __rng1_idx < __rng1_idx_end;
+        const bool __rng2_idx_less__n2 = __rng2_idx < __rng2_idx_end;
 
         __rng3[__rng3_idx] =
             __rng1_idx_less__n1 && __rng2_idx_less__n2
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -105,8 +105,8 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index _
         // because 1) we should fill output data with elements from one of the input ranges
         // 2) we calculate __rng3_idx_end as std::min<_Index>(__rng1_size + __rng2_size, __chunk).
         __rng3[__rng3_idx] =
-            (__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])) ||
-                    !__rng1_idx_less_n1
+            ((__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])) ||
+             !__rng1_idx_less_n1)
                 ? __rng2[__rng2_idx++]
                 : __rng1[__rng1_idx++];
     }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -85,47 +85,30 @@ void
 __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index __start1, _Index __start2,
                const _Index __start3, const std::uint8_t __chunk, const _Index __n1, const _Index __n2, _Compare __comp)
 {
-    if (__start1 >= __n1)
+    const _Index __rng1_size = std::min<_Index>(__n1 > __start1 ? __n1 - __start1 : (_Index)0, (_Index)__chunk);
+    const _Index __rng2_size = std::min<_Index>(__n2 > __start2 ? __n2 - __start2 : (_Index)0, (_Index)__chunk);
+    const _Index __rng3_size = std::min<_Index>(__rng1_size + __rng2_size, (_Index)__chunk);
+
+    const auto __rng1_idx_end = __start1 + __rng1_size;
+    const auto __rng2_idx_end = __start2 + __rng2_size;
+    const auto __rng3_idx_end = __start3 + __rng3_size;
+
+    _Index __rng1_idx = __start1;
+    _Index __rng2_idx = __start2;
+    _Index __rng3_idx = __start3;
+    
+    bool __rng1_idx_less__n1, __rng2_idx_less__n2;
+
+    #pragma unroll 4
+    for (_Index __rng3_idx = __start3; __rng3_idx < __rng3_idx_end; ++__rng3_idx)
     {
-        //copying a residual of the second seq
-        const _Index __n = std::min<_Index>(__n2 - __start2, __chunk);
-        for (std::uint8_t __i = 0; __i < __n; ++__i)
-            __rng3[__start3 + __i] = __rng2[__start2 + __i];
-    }
-    else if (__start2 >= __n2)
-    {
-        //copying a residual of the first seq
-        const _Index __n = std::min<_Index>(__n1 - __start1, __chunk);
-        for (std::uint8_t __i = 0; __i < __n; ++__i)
-            __rng3[__start3 + __i] = __rng1[__start1 + __i];
-    }
-    else
-    {
-        for (std::uint8_t __i = 0; __i < __chunk && __start1 < __n1 && __start2 < __n2; ++__i)
-        {
-            const auto& __val1 = __rng1[__start1];
-            const auto& __val2 = __rng2[__start2];
-            if (__comp(__val2, __val1))
-            {
-                __rng3[__start3 + __i] = __val2;
-                if (++__start2 == __n2)
-                {
-                    //copying a residual of the first seq
-                    for (++__i; __i < __chunk && __start1 < __n1; ++__i, ++__start1)
-                        __rng3[__start3 + __i] = __rng1[__start1];
-                }
-            }
-            else
-            {
-                __rng3[__start3 + __i] = __val1;
-                if (++__start1 == __n1)
-                {
-                    //copying a residual of the second seq
-                    for (++__i; __i < __chunk && __start2 < __n2; ++__i, ++__start2)
-                        __rng3[__start3 + __i] = __rng2[__start2];
-                }
-            }
-        }
+        __rng1_idx_less__n1 = __rng1_idx < __rng1_idx_end;
+        __rng2_idx_less__n2 = __rng2_idx < __rng2_idx_end;
+
+        __rng3[__rng3_idx] =
+            __rng1_idx_less__n1 && __rng2_idx_less__n2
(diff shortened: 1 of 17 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -80,52 +80,35 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
...
+
+    const auto __rng1_idx_end = __start1 + __rng1_size;
+    const auto __rng2_idx_end = __start2 + __rng2_size;
+    const auto __rng3_idx_end = __start3 + __rng3_size;
+
+    _Index __rng1_idx = __start1;
+    _Index __rng2_idx = __start2;
+    _Index __rng3_idx = __start3;
+    
+    bool __rng1_idx_less__n1, __rng2_idx_less__n2;
```
MikeDvorskiy, line 100: The uninitialized variables...  I guess we can move the variables definition  to the place where ones are initialized within the loop, w/o performance any penalty.

MikeDvorskiy (commented):
Looks very good to me, excepting one minor comment, see above/

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -272,17 +274,17 @@ struct __merge_sort_global_submitter<_IndexT, __internal::__optional_kernel_name
                             const oneapi::dpl::__ranges::drop_view_simple __rng2(__dst, __offset + __n1);
 
                             const auto start = __find_start_point(__rng1, __rng2, __i_elem_local, __n1, __n2, __comp);
-                            __serial_merge(__rng1, __rng2, __rng /*__rng3*/, start.first, start.second, __i_elem,
-                                           __chunk, __n1, __n2, __comp);
+                            __serial_merge<__gpu_chunk>(__rng1, __rng2, __rng /*__rng3*/, start.first, start.second,
```
mmichel11, line 277: I think renaming `__gpu_chunk` to something like `__unroll_factor` is more clear since there is a different `__chunk` and the device may not be a GPU.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
SergeyKopienko: Semantically in this code the variable exactly mean the value of `chunk` for any device that can also be used to accelerate a 3D graphics API. In other words - the chunk for non-CPU device.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -96,8 +97,8 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index _
     _Index __rng1_idx = __start1;
     _Index __rng2_idx = __start2;
     _Index __rng3_idx = __start3;
-    
-    #pragma unroll _UnrollFactor
+
+#pragma unroll _UnrollFactor
```
MikeDvorskiy, line 101: `#pragma` is not portable.. OneDPL has a macro
// Portability "#pragma" definition
```
#ifdef _MSC_VER
#    define _ONEDPL_PRAGMA(x) __pragma(x)
#else
#    define _ONEDPL_PRAGMA(x) _Pragma(#    x)
#endif

```

SergeyKopienko: Could you please explain how to apply this for unroll?

SergeyKopienko: We made decision together with @dmitriy-sobolev to remove `unroll` here at all: no perf profit has been detected.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -85,47 +85,27 @@ void
...
+    const _Index __rng2_size = std::min<_Index>(__n2 > __start2 ? __n2 - __start2 : (_Index)0, (_Index)__chunk);
+    const _Index __rng3_size = std::min<_Index>(__rng1_size + __rng2_size, (_Index)__chunk);
+
+    const auto __rng1_idx_end = __start1 + __rng1_size;
+    const auto __rng2_idx_end = __start2 + __rng2_size;
+    const auto __rng3_idx_end = __start3 + __rng3_size;
+
+    _Index __rng1_idx = __start1;
+    _Index __rng2_idx = __start2;
+    _Index __rng3_idx = __start3;
```
dmitriy-sobolev, line 98: `__rng3_idx` is redefined in the loop below.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -85,47 +85,27 @@ void
...
-            __rng3[__start3 + __i] = __rng1[__start1 + __i];
-    }
-    else
+    const _Index __rng1_size = std::min<_Index>(__n1 > __start1 ? __n1 - __start1 : (_Index)0, (_Index)__chunk);
+    const _Index __rng2_size = std::min<_Index>(__n2 > __start2 ? __n2 - __start2 : (_Index)0, (_Index)__chunk);
+    const _Index __rng3_size = std::min<_Index>(__rng1_size + __rng2_size, (_Index)__chunk);
+
+    const auto __rng1_idx_end = __start1 + __rng1_size;
+    const auto __rng2_idx_end = __start2 + __rng2_size;
+    const auto __rng3_idx_end = __start3 + __rng3_size;
```
dmitriy-sobolev, line 94: Let's use `_Index` instead of auto, because the type is known.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -85,47 +85,27 @@ void
...
-                __rng3[__start3 + __i] = __val1;
-                if (++__start1 == __n1)
-                {
-                    //copying a residual of the second seq
-                    for (++__i; __i < __chunk && __start2 < __n2; ++__i, ++__start2)
-                        __rng3[__start3 + __i] = __rng2[__start2];
-                }
-            }
-        }
+        const bool __rng1_idx_less__n1 = __rng1_idx < __rng1_idx_end;
```
dmitriy-sobolev, line 102: A nitpick: `__rng1_idx_less__n1` -> `__rng1_idx_less_n1`. Another option is `__rng1_idx_valid`.

MikeDvorskiy (approved):
Looks very good to me.

dmitriy-sobolev (commented):
It also looks good to me. Below are some minor suggestions.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -85,47 +85,27 @@ void
...
+        const bool __rng2_idx_less__n2 = __rng2_idx < __rng2_idx_end;
```
akukanov: Can it happen that both these conditions are false at the same time? The code that follows does not seem to handle such case correctly.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Fixed

dmitriy-sobolev (approved):

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -85,47 +85,26 @@ void
...
-    {
-        //copying a residual of the first seq
-        const _Index __n = std::min<_Index>(__n1 - __start1, __chunk);
-        for (std::uint8_t __i = 0; __i < __n; ++__i)
-            __rng3[__start3 + __i] = __rng1[__start1 + __i];
-    }
-    else
+    const _Index __rng1_size = std::min<_Index>(__n1 > __start1 ? __n1 - __start1 : (_Index)0, (_Index)__chunk);
+    const _Index __rng2_size = std::min<_Index>(__n2 > __start2 ? __n2 - __start2 : (_Index)0, (_Index)__chunk);
+    const _Index __rng3_size = std::min<_Index>(__rng1_size + __rng2_size, (_Index)__chunk);
```
danhoeflinger, line 90: Is there a reason to do C style casting here rather than doing explicit construction of a `_Index` type rvalue (or perhaps a `static_cast`)?

`(_Index)0` -> `_Index{0}`

I don't think we do a lot of c style casting in the repository.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -85,47 +85,26 @@ void
...
-                        __rng3[__start3 + __i] = __rng2[__start2];
-                }
-            }
-        }
+        const bool __rng1_idx_less_n1 = __rng1_idx < __rng1_idx_end;
+        const bool __rng2_idx_less_n2 = __rng2_idx < __rng2_idx_end;
+
+        __rng3[__rng3_idx] =
+            __rng1_idx_less_n1 && __rng2_idx_less_n2
+                ? (__comp(__rng2[__rng2_idx], __rng1[__rng1_idx]) ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++])
```
MikeDvorskiy, line 106: Postfix increment leads to a temporary variable creation usually...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -85,47 +85,27 @@ void
...
-    {
-        //copying a residual of the first seq
-        const _Index __n = std::min<_Index>(__n1 - __start1, __chunk);
-        for (std::uint8_t __i = 0; __i < __n; ++__i)
-            __rng3[__start3 + __i] = __rng1[__start1 + __i];
-    }
-    else
+    const _Index __rng1_size = std::min<_Index>(__n1 > __start1 ? __n1 - __start1 : (_Index)0, (_Index)__chunk);
+    const _Index __rng2_size = std::min<_Index>(__n2 > __start2 ? __n2 - __start2 : (_Index)0, (_Index)__chunk);
+    const _Index __rng3_size = std::min<_Index>(__rng1_size + __rng2_size, (_Index)__chunk);
```
akukanov: Since `uint8_t` is no more used for iteration, maybe is it better to change the `__chunk` parameter type to `_Index` instead of explicit conversions?
Also, `_Index(0)` is slightly better than the C-style cast.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Post-increment here required from logic side of implementation: we should get value from range and then increment index.  And exactly increment index for that range from which we got value.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
danhoeflinger: I agree they should result in the same code, I point it out because it is against what I understand to be our convention, but its a nitpick and I wont insist.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
akukanov: Both `__startX` and `__rngX_size` are of type `_Index`, so what is `auto` for?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: `_Index{0}` looks good.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: > Post-increment here required from logic side of implementation: we should get value from range and then increment index. And exactly increment index for that range from which we got value.

Of course I understand the logic here and I am aware of difference between Post-increment  and Pre-increment  operations in C/C++...

I am talking about possible additional temporary variables creation. Usually it is bad style in C/C++ by performance reasons. So, do we have / might have any performance penalty comparing with code w/o temporary variable in given context?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: I believe it's not possible due 
1) condition in the for-loop: `__rng3_idx < __rng3_idx_end`
2) and how we define the value of `__rng3_idx_end` :
```C++
    const _Index __rng1_size = std::min<_Index>(__n1 > __start1 ? __n1 - __start1 : _Index{0}, (_Index)__chunk);
    const _Index __rng2_size = std::min<_Index>(__n2 > __start2 ? __n2 - __start2 : _Index{0}, (_Index)__chunk);
    const _Index __rng3_size = std::min<_Index>(__rng1_size + __rng2_size, (_Index)__chunk);
```

So, logically we should fill all required data and this mean we have them in the first of in the second ranges.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
danhoeflinger: @MikeDvorskiy  I suppose the suggestion would then be to convert the ternary operators into a couple `if`/`else` statements where we can group the increment in a separate expression in the if block rather than embedding it in the assignment.
Is that your suggestion?

akukanov: Since `operator[]` here is likely a function call, and the expressions for the arguments of function calls are evaluated before the call, a temporary variable is technically needed. On the other hand, arguments are passed to the function calls either on the stack or in registers (depending on the calling convention), so the place for that temporary variable is likely to exist anyway.

Overall, I would not bother with such micro-optimizations, given that even explicit loop unrolling turned out to have no notable impact.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -85,47 +85,27 @@ void
...
+
+    const auto __rng1_idx_end = __start1 + __rng1_size;
+    const auto __rng2_idx_end = __start2 + __rng2_size;
+    const auto __rng3_idx_end = __start3 + __rng3_size;
+
+    _Index __rng1_idx = __start1;
+    _Index __rng2_idx = __start2;
+    _Index __rng3_idx = __start3;
+
+    for (_Index __rng3_idx = __start3; __rng3_idx < __rng3_idx_end; ++__rng3_idx)
```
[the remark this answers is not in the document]
SergeyKopienko: My personal preferences - to define in the for-loop only one variable to make for-loop more clear.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
akukanov: Perhaps add a comment that at least one of the conditions is (expectedly) true.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: > Is that your suggestion?

Not necessary  `if/else`.... for example: `(++__rng2_idx, __rng2[__rng2_idx-1])`

```
__rng3[__rng3_idx] =
            __rng1_idx_less__n1 && __rng2_idx_less__n2
                ? (__comp(__rng2[__rng2_idx], __rng1[__rng1_idx]) ? (++__rng2_idx, __rng2[__rng2_idx-1]) : (++__rng1_idx, __rng1[__rng1_idx-1]))
                : (__rng1_idx_less__n1 ? (++__rng1_idx, __rng1[__rng1_idx-1]) : (++__rng2_idx, __rng2[__rng2_idx-1]));
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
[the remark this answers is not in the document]
akukanov: OK, I have no desire to confront your personal preferences :)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: > Overall, I would not bother with such micro-optimizations, given that even explicit loop unrolling turned out to have no notable impact.

I just ask a question about post-increment, nothing more :)
Additionally, I've just compared the asm files with -03. The are identical. 

So, Q.E.D. `we do believe in a compiler.`, This is something like what I expected to hear as an answer, not explanation how serial merge works :)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
akukanov: > for example: (++__rng2_idx, __rng2[__rng2_idx-1])
... where `__rng2_idx-1` would be I believe an rvalue similar to the one created by `__rng2_idx++` :)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -83,49 +83,28 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
...
-                }
-            }
-        }
+        const bool __rng1_idx_less_n1 = __rng1_idx < __rng1_idx_end;
+        const bool __rng2_idx_less_n2 = __rng2_idx < __rng2_idx_end;
+
+        __rng3[__rng3_idx] =
+            __rng1_idx_less_n1 && __rng2_idx_less_n2
+                ? (__comp(__rng2[__rng2_idx], __rng1[__rng1_idx]) ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++])
+                : (__rng1_idx_less_n1 ? __rng1[__rng1_idx++] : __rng2[__rng2_idx++]);
```
danhoeflinger, line 110: So, I don't know if its worth the cost that may make it harder to read, but technically there are only 2 outcomes:  
use rng1 and increment its index 
use rng2 and increment its index

We should be able to form a single ternary / if statement to get the result we want.  Let me give it a try (obviously it'll need formatting):
```suggestion
        __rng3[__rng3_idx] =
            ((__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])) || __rng2_idx_less_n2) ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++]);
```

danhoeflinger: Its possible the compiler can do this for us but if not... we have an extra unnecessary branch which is costly.

SergeyKopienko: I believe technically you suggestion is absolutely correct. but a little bit not so readable.
@akukanov, @MikeDvorskiy, @dmitriy-sobolev what do you think?

danhoeflinger: I think at the benefit of removing a branch in an inner loop of the main computation loop of an algorithm, we can sacrifice readability.  If we feel its confusing, lets add a comment to clarify.

SergeyKopienko: One more variant:
```C++
        __rng3[__rng3_idx] =
            ((__rng1_idx_less_n1 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])) && __rng2_idx_less_n2)
                ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++]);
```

MikeDvorskiy: As for me, any variant readable...
But I would prefer variant w/o redundant braces, it makes the code lines less readable for me.
Regarding formatting - it doesn't matter for me with readable perspective.

danhoeflinger: But `__rng1_idx_less_n1==false` and `__rng2_idx_less_n2== true`, this variant will attempt to use `__rng1` and crash, wont it?

akukanov: I think Dan's code would not work as needed when `__rng1_idx_less_n1 == true && __rng2_idx_less_n2 == true && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx]) == false`.

danhoeflinger:
```
// If we are within both range 1 and 2, and the element from range2 is "less than" the element from range2, use it.  
// If we are not within both ranges, but we are still within range2, use the element from range2
// Otherwise, we know we are within range1 because we are in the loop, so use the element from range1.
```

danhoeflinger: @akukanov Yeah, I think you are right. Sorry let me fix the check, it has to get a bit more complex.

danhoeflinger: Second try:
```suggestion
        __rng3[__rng3_idx] =
            ((__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])) || !__rng1_idx_less_n1) ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++]);
```
This is what I originally had but thought I could get clever to avoid the negation.  I think this resolves the issues though.

SergeyKopienko: One comma pairs aren't required I believe:
```C++
        __rng3[__rng3_idx] =
            (__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])) || !__rng1_idx_less_n1
                ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++];
```

danhoeflinger: Yeah that is fine.  I personally like to wrap the condition in ternary operations in parens just to make the order of operations by the compiler clear both to the reader and the compiler, but I don't think it is necessary for correctness.

`!__rng1_idx_less_n1 ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++]` alone is a "valid" expression, which could then be combined via `||` with the first half of the conditional, so for me it is less confusing to just explicitly group the condition in parens to remove any doubt.

SergeyKopienko: Applied.

SergeyKopienko: > Yeah that is fine. I personally like to wrap the condition in ternary operations in parens just to make the order of operations clear both to the reader and the compiler, but I don't think it is necessary for correctness.
> 
> `!__rng1_idx_less_n1 ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++]` alone is a "valid" expression, which could then be combined via `||` with the first half of the conditional, so for me it is less confusing to just explicitly group the condition in parens to remove any doubt.

ok, applied too.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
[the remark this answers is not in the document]
SergeyKopienko: May I resolve this conversation?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -63,7 +63,7 @@ struct __group_merge_path_sorter
     template <typename _StorageAcc, typename _Compare>
     bool
     sort(const sycl::nd_item<1>& __item, const _StorageAcc& __storage_acc, _Compare __comp, std::uint32_t __start,
-         std::uint32_t __end, std::uint32_t __sorted, std::uint16_t __data_per_workitem,
+         std::uint32_t __end, std::uint32_t __sorted, std::uint32_t __data_per_workitem,
```
danhoeflinger, line 66: Last minor question:  
What changed here to need this type change?  Is it a bug fix?  it seems unrelated to the rest of the PR.

mmichel11 (dismissed):
LGTM on my side.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -82,50 +82,33 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
...
-                    //copying a residual of the second seq
-                    for (++__i; __i < __chunk && __start2 < __n2; ++__i, ++__start2)
-                        __rng3[__start3 + __i] = __rng2[__start2];
-                }
-            }
-        }
+        const bool __rng1_idx_less_n1 = __rng1_idx < __rng1_idx_end;
+        const bool __rng2_idx_less_n2 = __rng2_idx < __rng2_idx_end;
+
+        // One of __rng1_idx_less_n1 and __rng2_idx_less_n2 should be true here
```
MikeDvorskiy, line 104: Do we whether want to write an `assert` here?

SergeyKopienko: It's Kernel code, so no asserts.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -82,50 +82,33 @@ __find_start_point(const _Rng1& __rng1, const _Rng2& __rng2, const _Index __i_el
...
+        // because 1) we should fill output data with elements from one of the input ranges
+        // 2) we calculate __rng3_idx_end as std::min<_Index>(__rng1_size + __rng2_size, __chunk).
+        __rng3[__rng3_idx] =
+            ((__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])) ||
+             !__rng1_idx_less_n1)
+                ? __rng2[__rng2_idx++]
+                : __rng1[__rng1_idx++];
```
MikeDvorskiy, line 111: common! it is too much.. Let's not writting the code in one column :)

`!__rng1_idx_less_n1)  ? __rng2[__rng2_idx++]  : __rng1[__rng1_idx++];`

is absolutely readable.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: 1) CPU device?
2) https://intel.github.io/llvm/design/Assert.html
3) I guess we are going to share the function with the host code as well, probably...

SergeyKopienko: As noted @akukanov earlier in https://github.com/oneapi-src/oneDPL/pull/1933 conversation, we can't use `assert` calls inside Kernel code. So what you propose exactly?

danhoeflinger (approved):
LGTM pending green CI, thanks

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
danhoeflinger: After offline discussion, we see that `__data_per_workitem` is only used in 32 bit unsigned context, and we have made things more explicit by removing `auto` in the return from `__find_start_point` for the other parameters with `_Index` type going into `__serial_merge` so that it is clear that we will end up with a consistent type.  
Thanks for the changes here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: "disagree and commit"

mmichel11 (dismissed):

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: > It's automatic format.

What kind of "automatic" ? by IDE/text redactor or by clang format (form CI) ?

SergeyKopienko: It's someone IDE with our clang-format setting.
Is it really important?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
akukanov: For the sake of accuracy, I have said this:
> We should consider the use of assert in device code as an extension of DPC++ and possibly other SYCL implementations, so it should be guarded by some macros.

Practically, I would perhaps recommend to use the assertion during development to ensure that it does not fail for our tests, but then remove it. The rationale is that any later changes to the code would likely require the assertion to be modified as well, so it does not seem much useful beyond the initial testing.

dmitriy-sobolev (dismissed):

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: Based on [1933](https://github.com/oneapi-src/oneDPL/pull/1933#discussion_r1873731880) discussion,  I agree with following conclusion: 

`We should consider the use of assert in device code as an extension of DPC++ and possibly other SYCL implementations, so it should be guarded by some macros`

I don't insist doing it in this PR but I will create an GitHub issue.

SergeyKopienko:
Unfortunately clang-formatting has been fixed. So approves required again.
Sorry for inconveniences.

danhoeflinger (approved):

PR #1971, opened 2024-12-18
RFC process clarification

akukanov:
Following an internal team discussion, this PR updates the RFC documentation to reflect that
- at any state proposals can be developed iteratively in several revisions;
- there is no single strictly defined flow, and each state can be skipped;
- "umbrella" documents might be simultaneously in a few states, as necessitated by the progress of sub-features.

### DIFF
--- a/rfcs/experimental/README.md
+++ b/rfcs/experimental/README.md
@@ -15,10 +15,11 @@ These conditions might include demonstrated performance improvements, demonstrat
 interest from the community, acceptance of the required specification changes, etc.
 
 A document here needs to be updated if the corresponding feature undergoes
-modifications while remaining experimental.
+modifications while remaining experimental. Other changes, such as updates on the
+exit conditions or on the implementation and usage experience are also welcome.
 
-For features that require specification changes, the document might include
-wording for those changes or a link to any PRs opened against the specification.
+For features that require specification changes prior to production, the document might
+include wording for those changes or a link to any PRs opened against the specification.
 
 Proposals in the `rfcs/experimental` directory do not remain there indefinitely.
 They should move either to `rfcs/supported` when they become fully supported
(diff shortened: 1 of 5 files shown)


rfcs/experimental/README.md
```diff
@@ -5,19 +5,20 @@ released as experimental features in the library. An experimental
...
-demonstrated interest from the community,
-acceptance of the required specification changes, etc. 
+These conditions might include demonstrated performance improvements, demonstrated
+interest from the community, acceptance of the required specification changes, etc.
 
-For features that require specification changes, the document might
-include wording for those changes or a link to any PRs opened
-against the specification.
+A document here needs to be updated if the corresponding feature undergoes
+modifications while remaining experimental.
```
vossmjp, line 18: Should you add that the document may also be updated as the status of the listed exit conditions change. For example, if performance improvements have been demonstrated.

rfcs/template.md
```diff
@@ -24,6 +24,8 @@ Depending on the kind of the proposal, the description should cover:
 - New use cases supported by the extension.
 - The expected performance benefit for a modification, supported with the data, if available.
 - The API of extensions such as class definitions and function declarations.
+- Key technical design aspects, sufficient to understand how the functionality should work
```
vossmjp, line 27: Does this section sound too strong now given the relaxing of the process. For example, maybe "Depending on the kind of proposal, the description may cover the following. And these sections may be iteratively clarified as the proposal matures." Or something like this,

rfcs/experimental/README.md
akukanov: Thanks; I added a sentence for this.

rfcs/template.md
akukanov: Updated in line with your suggestion.

vossmjp (approved):
Looks good to me

danhoeflinger (approved):
LGTM, I think this reflects the changes we wanted to make.

PR #1974, opened 2024-12-18
Host Implementation of Histogram APIs

danhoeflinger:
Implementation of histogram APIs for host backends.  

Implementations are provided for `serial`, `tbb`, and `openMP` backends.  We add a generic `__thread_enumerable_storage` struct to add a generic thread local storage for our host backends.  We use the new TLS (Thread local storage) with `parallel_for` to implement histogram.  Testing is also added, and some minor adjustments are made to cmake.

Please see the RFC documentation / discussion [here](https://github.com/oneapi-src/oneDPL/pull/1930) for more details.


include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -4289,6 +4289,87 @@ __pattern_shift_right(_Tag __tag, _ExecutionPolicy&& __exec, _BidirectionalItera
...
+    _DiffType __n = __last - __first;
+    if (__n > 0)
+    {
+        __par_backend::__thread_enumerable_storage<std::vector<_HistogramValueT>> __tls{__num_bins,
+                                                                                        _HistogramValueT{0}};
+
+        //main histogram loop
+        //TODO: add defaulted grain-size option for __parallel_for and use larger one here to account for overhead
+        __par_backend::__parallel_for(
+            __backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec), __first, __last,
```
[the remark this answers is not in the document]
danhoeflinger: thanks, fixed.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+        {
+            // create temporary storage on first usage to avoid extra parallel region and unnecessary instantiation
+            __thread_specific_storage[__i] = __construct_helper->construct();
+            __num_elements.fetch_add(1);
+        }
+        return *__thread_specific_storage[__i];
+    }
+
+    std::vector<std::unique_ptr<_StorageType>> __thread_specific_storage;
+    std::atomic<std::uint32_t> __num_elements;
```
SergeyKopienko, line 237: May be better to use `std::atomic<std::size_t>` or  `atomic_size_t` ?

include/oneapi/dpl/pstl/parallel_backend_serial.h
```diff
@@ -42,6 +42,35 @@ __cancel_execution(oneapi::dpl::__internal::__serial_backend_tag)
...
+template <typename _StorageType>
+struct __thread_enumerable_storage
+{
+    template <typename... Args>
+    __thread_enumerable_storage(Args&&... __args) : __storage(std::forward<Args>(__args)...)
+    {
+    }
+
+    std::uint32_t
+    size() const
```
SergeyKopienko, line 54: I think better to use `std::size_t` for size-linked stuff.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+            {
+                __count++;
+            }
+        }
+        // Need to back up one once we have found a valid storage object
+        return *__thread_specific_storage[__j - 1];
+    }
+
+    _StorageType&
+    get_for_current_thread()
```
SergeyKopienko, line 224: I think this function isn't thread-safe. Is it ok?

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
     __f(__first, __last);
 }
 
+// abstract class to allow inclusion in __thread_enumerable_storage as member without requiring explicit template
+// instantiation of param types
+template <typename _StorageType>
+class __construct_by_args_base
+{
+  public:
+    virtual ~__construct_by_args_base() {}
```
SergeyKopienko, line 163:
```suggestion
    virtual ~__construct_by_args_base() = default;
```

SergeyKopienko: This change usually strongly recommended for `virtual` destructors in the base class.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+
+// Helper class to allow construction of _StorageType from a stored argument pack
+template <typename _StorageType, typename... _P>
+class __construct_by_args : public __construct_by_args_base<_StorageType>
+{
+  public:
+    std::unique_ptr<_StorageType>
+    construct() override
+    {
+        return std::move(
```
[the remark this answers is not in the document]
SergeyKopienko: I feel it's not needed too.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+            std::apply([](auto... __arg_pack) { return std::make_unique<_StorageType>(__arg_pack...); }, __pack));
```
SergeyKopienko, line 176: I think we know the real type of `__arg_pack` here, so should we need to use `auto` ?

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+    }
+    __construct_by_args(_P&&... __args) : __pack(std::forward<_P>(__args)...) {}
+
+  private:
+    std::tuple<_P...> __pack;
```
SergeyKopienko, line 181: May be better to declare `__pack` field as `const` ?

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+};
+
+template <typename _StorageType>
+struct __thread_enumerable_storage
+{
+    template <typename... Args>
+    __thread_enumerable_storage(Args&&... __args) : __num_elements(0)
+    {
+        __construct_helper =
```
SergeyKopienko, line 190: The proper name for this field - something like `__storage_factory`, it's a real some kind of `factory` pattern.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+        }
+        // Need to back up one once we have found a valid storage object
+        return *__thread_specific_storage[__j - 1];
+    }
+
+    _StorageType&
+    get_for_current_thread()
+    {
+        std::uint32_t __i = omp_get_thread_num();
+        if (!__thread_specific_storage[__i])
```
SergeyKopienko, line 227: What if we using `std::call_once` flag for every thread?
And declare it as `thread_local`...
In this case I believe we didn't need `__num_elements` anymore and may safe use `std::vector::size()` in the `size()` method. The justification - in the `__pattern_histogram` implementation we see two separate phases:
- in the first phase we create storages for each thread and in the second phase we simple accumulate data from them.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -4289,6 +4289,86 @@ __pattern_shift_right(_Tag __tag, _ExecutionPolicy&& __exec, _BidirectionalItera
...
+                    _RandomAccessIterator2 __histogram_first)
+{
+    using __backend_tag = typename __parallel_tag<_IsVector>::__backend_tag;
+    using _HistogramValueT = typename std::iterator_traits<_RandomAccessIterator2>::value_type;
+    using _DiffType = typename std::iterator_traits<_RandomAccessIterator2>::difference_type;
+
+    _DiffType __n = __last - __first;
+    if (__n > 0)
+    {
+        __par_backend::__thread_enumerable_storage<std::vector<_HistogramValueT>> __tls{__num_bins,
```
SergeyKopienko, line 4334: May be we have some way to create common data dtorage for all threads in all place?
And then each thread will will his own part of common data storage...

include/oneapi/dpl/pstl/parallel_backend_serial.h
```diff
@@ -42,6 +42,35 @@ __cancel_execution(oneapi::dpl::__internal::__serial_backend_tag)
...
+    }
+
+    _StorageType&
+    get_for_current_thread()
+    {
+        return __storage;
+    }
+
+    _StorageType&
+    get_with_id(std::uint32_t __i)
```
SergeyKopienko, line 66: Probably we will have some warning here about unused variable `__i`.

SergeyKopienko, line 66:
```suggestion
    get_with_id(std::uint32_t /*__i*/)
```

include/oneapi/dpl/pstl/omp/util.h
danhoeflinger: We could do this, but it would be at a cost to performance, especially for small input sizes and for machines with many threads.  

For histogram at least, it is a large overhead cost to allocate, initialize and then finally accumulate the temporary storage.  So if we have an input size which would only warrant a few threads, we would rather not deal with the overhead for all threads.  This is why we prefer a TLS which only allocates as threads encounter the need for it.  

On the other hand, for small values of `n` our performance is already not very good for the "even bins" API. If we choose to abandon performance of the small case, this would certainly simplify the openMP implementation.

include/oneapi/dpl/pstl/omp/util.h
danhoeflinger: `__num_elements` corresponds to the number of threads.  If we have more than 4 billion threads,  I think something has gone terribly wrong :) .

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+        std::uint32_t __j = 0;
+
+        for (; __j < __thread_specific_storage.size() && __count <= __i; ++__j)
+        {
+            // Only include storage from threads which have instantiated a storage object
+            if (__thread_specific_storage[__j])
+            {
+                __count++;
+            }
+        }
```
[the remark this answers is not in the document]
danhoeflinger: Good idea, done.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +154,90 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+
+    _StorageType&
+    get_with_id(std::uint32_t __i)
+    {
+        assert(__i < size());
+
+        std::uint32_t __count = 0;
+        std::uint32_t __j = 0;
+
+        for (; __j < __thread_specific_storage.size() && __count <= __i; ++__j)
```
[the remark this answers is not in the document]
danhoeflinger: thanks, done.

include/oneapi/dpl/pstl/omp/util.h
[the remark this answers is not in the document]
danhoeflinger: Agreed, thanks.

include/oneapi/dpl/pstl/omp/util.h
danhoeflinger: Good point.  fixed.

include/oneapi/dpl/pstl/omp/util.h
danhoeflinger: Thanks, done.

include/oneapi/dpl/pstl/omp/util.h
danhoeflinger: Thanks, done.

danhoeflinger:
One question I have for the group... 
If we know that a serial implementation will provide better performance up to some threshold (perhaps dependent on num bins, num threads, num input elements),  can / should we dispatch instead to a serial implementation? 

From my reading, it seems the answer is probably no.  Execution policies have semantic meaning, and `par` / `par_unseq` do not simply mean "provide the fastest version" even if that is what the users probably want.

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/omp/util.h
SergeyKopienko: OK

include/oneapi/dpl/pstl/omp/util.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/parallel_backend_serial.h
SergeyKopienko: From my point of view, if we describe anything about the size, better to use `std::size_t` type.
The exception - only one case when we save size in the memory and want to reduce the amount of memory.
But of course, I don't insist, it's not one place with another approach.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -4289,6 +4289,86 @@ __pattern_shift_right(_Tag __tag, _ExecutionPolicy&& __exec, _BidirectionalItera
...
+                __internal::__brick_histogram(__first_local, __last_local, __func,
+                                              __tls.get_for_current_thread().begin(), _IsVector{});
+            });
+        // now accumulate temporary storage into output global histogram
+        __par_backend::__parallel_for(
+            __backend_tag{}, std::forward<_ExecutionPolicy>(__exec), __histogram_first, __histogram_first + __num_bins,
+            [__histogram_first, &__tls](auto __global_histogram_first, auto __global_histogram_last) {
+                _DiffType __local_n = __global_histogram_last - __global_histogram_first;
+                std::uint32_t __num_temporary_copies = __tls.size();
+                _DiffType __range_begin_id = __global_histogram_first - __histogram_first;
```
SergeyKopienko, line 4351:
```suggestion
                const _DiffType __local_n = __global_histogram_last - __global_histogram_first;
                const std::uint32_t __num_temporary_copies = __tls.size();
                const _DiffType __range_begin_id = __global_histogram_first - __histogram_first;
```

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -4289,6 +4289,86 @@ __pattern_shift_right(_Tag __tag, _ExecutionPolicy&& __exec, _BidirectionalItera
...
+                    //accumulate into output global histogram with other local histogram via += operator
+                    __internal::__brick_walk2_n(
+                        __tls.get_with_id(__i).begin() + __range_begin_id, __local_n, __global_histogram_first,
+                        [](_HistogramValueT __x, _HistogramValueT& __y) { __y += __x; }, _IsVector{});
+                }
+            });
+    }
+    else
+    {
+        __pattern_fill(__parallel_tag<_IsVector>{}, std::forward<_ExecutionPolicy>(__exec), __histogram_first,
```
SergeyKopienko, line 4367: Dan, could you please note, which code will fill by `0' bins in the case when `__n > 0' ?

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +155,95 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+{
+    template <typename... Args>
+    __thread_enumerable_storage(Args&&... __args) : __num_elements(0)
+    {
+        __storage_factory = std::make_unique<__construct_by_args<_StorageType, Args...>>(std::forward<Args>(__args)...);
+        _PSTL_PRAGMA(omp parallel)
+        _PSTL_PRAGMA(omp single) { __thread_specific_storage.resize(omp_get_num_threads()); }
+    }
+
+    // Note: Size should not be used concurrantly with parallel loops which may instantiate storage objects, as it may
```
mmichel11, line 195:
```suggestion
    // Note: Size should not be used concurrently with parallel loops which may instantiate storage objects, as it may
```

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -4289,6 +4289,86 @@ __pattern_shift_right(_Tag __tag, _ExecutionPolicy&& __exec, _BidirectionalItera
...
+            [__func, &__tls](_RandomAccessIterator1 __first_local, _RandomAccessIterator1 __last_local) {
+                __internal::__brick_histogram(__first_local, __last_local, __func,
+                                              __tls.get_for_current_thread().begin(), _IsVector{});
+            });
+        // now accumulate temporary storage into output global histogram
+        __par_backend::__parallel_for(
+            __backend_tag{}, std::forward<_ExecutionPolicy>(__exec), __histogram_first, __histogram_first + __num_bins,
+            [__histogram_first, &__tls](auto __global_histogram_first, auto __global_histogram_last) {
+                _DiffType __local_n = __global_histogram_last - __global_histogram_first;
+                std::uint32_t __num_temporary_copies = __tls.size();
```
mmichel11, line 4350: After the first parallel_for, `__tls.size()` should no longer grow, correct? What about moving `__num_temporary_copies` outside of this parallel_for and capturing it into the lambda? This way we only perform a single atomic load.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -4289,6 +4289,86 @@ __pattern_shift_right(_Tag __tag, _ExecutionPolicy&& __exec, _BidirectionalItera
...
+__pattern_histogram(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomAccessIterator1 __first,
+                    _RandomAccessIterator1 __last, _Size __num_bins, _IdxHashFunc __func,
+                    _RandomAccessIterator2 __histogram_first)
+{
+    using __backend_tag = typename __parallel_tag<_IsVector>::__backend_tag;
+    using _HistogramValueT = typename std::iterator_traits<_RandomAccessIterator2>::value_type;
+    using _DiffType = typename std::iterator_traits<_RandomAccessIterator2>::difference_type;
+
+    _DiffType __n = __last - __first;
+    if (__n > 0)
```
MikeDvorskiy, line 4332: I don't insist... But there is a common good practice to process a shorter branch of `if..else` operator first. It has better readability. Also, in case of immediate return from the shorter branch we have the "non indented" code of the bigger branch:

```
if(n ==0)
{
    //shorter branch
   return;
}
//else: bigger branch - no indents.
.....
```

danhoeflinger: I rearranged if and else, but did not return early and remove braces for else.

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: good idea, done.

include/oneapi/dpl/pstl/omp/util.h
danhoeflinger: done, thanks.

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: thanks, done.

include/oneapi/dpl/pstl/parallel_backend_serial.h
danhoeflinger: I've done this.  You are right that there is no need to squeeze space out of types like these at this level.  Also, for TBB we rely upon their size type, so its possible in some situations it could have been a narrowing type conversion to uint32_t.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -4289,6 +4289,88 @@ __pattern_shift_right(_Tag __tag, _ExecutionPolicy&& __exec, _BidirectionalItera
...
 
+template <class _ForwardIterator, class _IdxHashFunc, class _RandomAccessIterator, class _IsVector>
+void
+__brick_histogram(_ForwardIterator __first, _ForwardIterator __last, _IdxHashFunc __func,
+                  _RandomAccessIterator __histogram_first, _IsVector) noexcept
+{
+    for (; __first != __last; ++__first)
+    {
+        std::int32_t __bin = __func.get_bin(*__first);
+        if (__bin >= 0)
```
MikeDvorskiy, line 4300: 1) Can `get_bin` return negative value?
2) If yes, what is correct behavior  for `__brick_histogram`?

danhoeflinger: Yes, `-1` is returned when the input element does not fall within any histogram bin.  The correct behavior is to do nothing and skip this input element.
[Specification](https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneDPL/source/parallel_api/algorithms.rst)
"Input values that do not map to a defined bin are skipped silently."

I recently looked into expanding the bin helper interface to include a separate function to check bounds, and another to get the bin which assumes it is in bounds.  I thought this might provide benefit by reducing the number of branches by 1, but I saw no performance benefit from this change for CPU or GPU.  It is still something we could pursue in the future.

mmichel11 (dismissed):
The implementation LGTM, but we should probably hold off on merging until the RFC is first merged.

I agree with ignoring clang-format for the single difference.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -4289,6 +4289,88 @@ __pattern_shift_right(_Tag __tag, _ExecutionPolicy&& __exec, _BidirectionalItera
...
+__pattern_histogram(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomAccessIterator1 __first,
+                    _RandomAccessIterator1 __last, _Size __num_bins, _IdxHashFunc __func,
+                    _RandomAccessIterator2 __histogram_first)
+{
+    using __backend_tag = typename __parallel_tag<_IsVector>::__backend_tag;
+    using _HistogramValueT = typename std::iterator_traits<_RandomAccessIterator2>::value_type;
+    using _DiffType = typename std::iterator_traits<_RandomAccessIterator2>::difference_type;
+
+    _DiffType __n = __last - __first;
+    if (__n <= 0)
```
akukanov, line 4332: It's rather not typical for STL-like algorithms to do such range checks. Instead, it is assumed (required) that `first` and `last` belong to the same sequence and `last` can be reached when iteration starts from `first`. Perhaps it can be explained by STL working with non-random-access iterators. Anyway, for oneDPL algorithms there is no such checks in most cases, and passing iterators in a wrong order or totally unrelated is undefined behavior.

So, unless this is needed to match the semantics of other histogram implementations, I would not bother with this check.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +155,95 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+
+  private:
+    const std::tuple<_P...> __pack;
+};
+
+template <typename _StorageType>
+struct __enumerable_thread_local_storage
+{
+    template <typename... Args>
+    __enumerable_thread_local_storage(Args&&... __args) : __num_elements(0)
```
akukanov, line 188: I wonder what downsides would be (if any) if `Args` here were the parameters of the class, and not of the constructor. Obviously it would simplify and reduce overheads in the OpenMP backend implementation, and there would be virtually no difference for the serial and TBB backend implementations. Would it be more complicated to use? It would lose flexibility to construct in different ways (with different sets of arguments) - but is that important?

CMakeLists.txt
```diff
@@ -173,7 +173,7 @@ if (ONEDPL_BACKEND MATCHES "^(tbb|dpcpp|dpcpp_only)$")
     set(SET_BACKEND_${ONEDPL_BACKEND_NAME} TRUE)
 
     if (ONEDPL_BACKEND MATCHES "^(tbb|dpcpp)$")
-        find_package(TBB 2021 REQUIRED tbb OPTIONAL_COMPONENTS tbbmalloc)
+        find_package(TBB 2021 REQUIRED tbb tbbmalloc)
```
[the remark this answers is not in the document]
danhoeflinger: If I remember correctly, I believe `tbbmalloc` was necessary for `tbb::enumerable_thread_specific`.  I can confirm this tomorrow...

include/oneapi/dpl/pstl/omp/util.h
danhoeflinger: The idea here was to not require users of `__enumerable_thread_local_storage` to know and verbosely write out the types of all the arguments they are using to initialize each element of thread local storage. 

```__par_backend::__enumerable_thread_local_storage<std::vector<_HistogramValueT>> __tls{__num_bins, _HistogramValueT{0}};``` 

is a lot cleaner than 
```__par_backend::__enumerable_thread_local_storage<std::vector<_HistogramValueT>, _Size, _HistogramValueT> __tls{__num_bins, _HistogramValueT{0}};``` 

in my opinion.

Do you think that it is slowing down the openMP backend?  I'm not understanding how in any nontrivial way, but maybe I'm missing something.

File: include/oneapi/dpl/pstl/algorithm_impl.h (hunk 21 above)
Comment (danhoeflinger) [reply include/oneapi/dpl/pstl/algorithm_impl.h:4332]: This is basically to cover the case where no input elements are provided, but we still need to clear the output histogram by the semantics of the algorithm (as I remember it).  I suppose we could just make it `__n== 0`...

File: include/oneapi/dpl/pstl/omp/util.h (hunk 22 above)
Comment (akukanov) [reply include/oneapi/dpl/pstl/omp/util.h:188]: Well, maybe it does not slow down much, but you would not need the "storage factory" and the associated memory allocations and virtual calls, you'd just store the construction argument pack as a tuple in the class instance, and directly call `std::apply` whenever a new thread specific value is needed.

The usability concern could be mitigated by CTAD if all template parameters would be deducible, but also a simple "make" utility function would do the trick without much trouble: `__par_backend::__enumerable_thread_local_storage __tls = __make_etls<std::vector<_HistogramValueT>>(__num_bins, _HistogramValueT{0});`

File: include/oneapi/dpl/pstl/algorithm_impl.h (hunk 21 above)
Comment (akukanov) [reply include/oneapi/dpl/pstl/algorithm_impl.h:4332]: OK, if the semantics for an empty input range are to return all zeros and we cannot totally eliminate the check, I do not mind it either way.

File: CMakeLists.txt (hunk 23 above)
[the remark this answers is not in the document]
Comment (akukanov) [reply CMakeLists.txt:176]: It should not be, unless there is a bug in oneTBB.

File: include/oneapi/dpl/pstl/omp/util.h (hunk 22 above)
Comment (danhoeflinger) [reply include/oneapi/dpl/pstl/omp/util.h:188]: Yes, this is what I based the implementation on.  I agree the make function could do the trick.  I'll play around with that in order to remove some of the complexity, and see if there are any downsides, thanks.

File: include/oneapi/dpl/pstl/omp/util.h [hunk 24]
```diff
@@ -153,6 +155,95 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+    {
+        __storage_factory = std::make_unique<__construct_by_args<_StorageType, Args...>>(std::forward<Args>(__args)...);
+        _PSTL_PRAGMA(omp parallel)
+        _PSTL_PRAGMA(omp single) { __thread_specific_storage.resize(omp_get_num_threads()); }
```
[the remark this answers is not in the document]
danhoeflinger: fixed.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +155,95 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+    {
+        return std::apply([](_P... __arg_pack) { return std::make_unique<_StorageType>(__arg_pack...); }, __pack);
+    }
+    __construct_by_args(_P&&... __args) : __pack(std::forward<_P>(__args)...) {}
+
+  private:
+    const std::tuple<_P...> __pack;
+};
+
+template <typename _StorageType>
```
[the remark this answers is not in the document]
danhoeflinger: fixed.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -4289,6 +4289,88 @@ __pattern_shift_right(_Tag __tag, _ExecutionPolicy&& __exec, _BidirectionalItera
...
+                const _DiffType __local_n = __global_histogram_last - __global_histogram_first;
+                const _DiffType __range_begin_id = __global_histogram_first - __histogram_first;
+                //initialize output global histogram with first local histogram via assign
+                __internal::__brick_walk2_n(__tls.get_with_id(0).begin() + __range_begin_id, __local_n,
+                                            __global_histogram_first, oneapi::dpl::__internal::__pstl_assign(),
+                                            _IsVector{});
+                for (std::size_t __i = 1; __i < __num_temporary_copies; ++__i)
+                {
+                    //accumulate into output global histogram with other local histogram via += operator
+                    __internal::__brick_walk2_n(
```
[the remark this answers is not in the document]
danhoeflinger: I have not yet tried, but the intention is to allow travel in the inner most loop along the cachelines (within a single temporary / global), rather than skipping from temporary copy to temporary copy for each vector or scalar element.  

I will see if we can get any benefit out of unrolling, and check on some alternatives swapping the loop orders.

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: I've swapped it to `==` for more clarity of intention.

CMakeLists.txt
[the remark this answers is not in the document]
danhoeflinger: I must've been mistaken or in some intermediate state which required it...  Thanks, removed.

include/oneapi/dpl/pstl/algorithm_impl.h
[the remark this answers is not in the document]
danhoeflinger: I tested this within our revised matrix of benchmarking space.

Unrolling doesn't really have any measurable effect when outside the brick (both full unrolling an unrolling to 8).  

I tried switching to a `std::accumulate` call within the brick because of the TLS. It has to be a little more complex than expected, using a counting iterator and transform iterator to properly access the TLS.  I also send in the zeroth element in as init, and start the sequence at index 1.  In TBB, this is approximately the same performance, but it is FAR worse in OpenMP.  This makes sense with my previous comment because the minimum grainsize in TBB is 1, but in OpenMP it is 2048.  My understanding is that TBB is not able to take advantage of access along the cacheline either way since it is processing a single element at a time, but OpenMP does get an advantage here and loses it when we exchange the loop order to `std::accumulate` within the brick.  I did not try it as a loop, but if my understanding is correct, there is no need to look at it. 

I'd recommend keeping it as is.  Then as previously discussed,  working to improve the `parallel_for` pattern's expressivity for control from this level over grain size, we are likely missing performance in TBB until we address this.

include/oneapi/dpl/pstl/omp/util.h
danhoeflinger: I took your suggestion for the `__make_etls` function.  It greatly simplifies the OpenMP implementation, at the cost of an unused template parameter pack for serial and TBB.   I think its an overall benefit, the previous implementation was overkill for the benefit it provided.

test/parallel_api/numeric/numeric.ops/histogram.pass.cpp
```diff
@@ -107,20 +124,21 @@ test_histogram(T min_boundary, T max_boundary, T overflow, Size jitter, Size tra
             test_range_and_even_histogram<CallNumber>(n, min_boundary, max_boundary, overflow, jitter, bin_size, trash);
         }
     }
+
+    //single large case for coverage
+    Size n = 1100000;
+    Size bin_size = 1100000;
```
SergeyKopienko, line 130:
```suggestion
    const Size n = 1100000;
    const Size bin_size = 1100000;
```

danhoeflinger: This case actually isnt necessary for coverage anymore without the atomic version of the algorithm, so I've removed it.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +155,70 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+        {
+            return *__thread_specific_storage[__i];
+        }
+
+        for (std::size_t __count = 0; __j < __thread_specific_storage.size() && __count <= __i; ++__j)
+        {
+            // Only include storage from threads which have instantiated a storage object
+            if (__thread_specific_storage[__j])
+            {
+                __count++;
```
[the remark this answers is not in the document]
danhoeflinger: applied, thanks.

include/oneapi/dpl/pstl/algorithm_impl.h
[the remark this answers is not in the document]
akukanov: OK. Consider adding some info about this into the RFC update.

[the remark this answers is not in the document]
danhoeflinger: Sure, I can do that.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +155,87 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
     __f(__first, __last);
 }
 
+namespace __detail
+{
+
+template <typename _ValueType, typename... _Args>
+struct __enumerable_thread_local_storage
+{
```
akukanov, line 163: Even though these class templates are now internal, the API description for them is still a part of the backend API - you should know what can be done with the object obtained from the make function. We will therefore need to document it somewhere in sufficient detail - perhaps in the RFC for a start.

include/oneapi/dpl/pstl/histogram_impl.h
```diff
@@ -18,7 +18,10 @@
 
 #include "histogram_extension_defs.h"
 #include "histogram_binhash_utils.h"
+#include "execution_impl.h"
 #include "iterator_impl.h"
+#include "algorithm_impl.h"
```
akukanov, line 23: Would it be sufficient to have just forward declarations for other patterns and bricks here, i.e. include `algorithm_fwd.h`? And would it be preferable to do it this way - e.g. to avoid potential circular dependencies?

include/oneapi/dpl/pstl/histogram_impl.h
```diff
@@ -18,7 +18,10 @@
...
+#include "parallel_backend_utils.h"
```
akukanov, line 24: I do not see anything from `oneapi::dpl::__utils` used in the code; looks like an artifact of some experimental code.

include/oneapi/dpl/pstl/histogram_impl.h
```diff
@@ -32,16 +35,86 @@ namespace dpl
...
+                                              __tls.get_for_current_thread().begin(), _IsVector{});
+            });
+        // now accumulate temporary storage into output global histogram
+        const std::size_t __num_temporary_copies = __tls.size();
+        __par_backend::__parallel_for(
+            __backend_tag{}, std::forward<_ExecutionPolicy>(__exec), __histogram_first, __histogram_first + __num_bins,
+            [__num_temporary_copies, __histogram_first, &__tls](auto __global_histogram_first,
+                                                                auto __global_histogram_last) {
+                const _DiffType __local_n = __global_histogram_last - __global_histogram_first;
+                const _DiffType __range_begin_id = __global_histogram_first - __histogram_first;
```
akukanov, line 104: Possibly this can be simplified a bit.
As far as I remember, `__parallel_for` can work with indexes, not just with random access iterators. So if you pass the iteration space begin/end indexes when calling it, you do not have to compute indexes like `__range_begin_id` inside the lambda.

include/oneapi/dpl/pstl/histogram_impl.h
```diff
@@ -32,16 +35,86 @@ namespace dpl
...
+            [__func, &__tls](_RandomAccessIterator1 __first_local, _RandomAccessIterator1 __last_local) {
+                __internal::__brick_histogram(__first_local, __last_local, __func,
+                                              __tls.get_for_current_thread().begin(), _IsVector{});
+            });
+        // now accumulate temporary storage into output global histogram
+        const std::size_t __num_temporary_copies = __tls.size();
+        __par_backend::__parallel_for(
+            __backend_tag{}, std::forward<_ExecutionPolicy>(__exec), __histogram_first, __histogram_first + __num_bins,
+            [__num_temporary_copies, __histogram_first, &__tls](auto __global_histogram_first,
+                                                                auto __global_histogram_last) {
```
akukanov, line 102: The parameter names are confusing, specifically the `__global` prefix. Semantically, the parameters are "local first" and "local last" I believe,
I understand that you mean the histogram is global (as opposed to per-thread), but nevertheless it's confusing.

include/oneapi/dpl/pstl/omp/util.h
danhoeflinger: Yes, I believe this interface is described in the RFC to some extent already.  I can make that more specific and explicit.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -153,6 +155,87 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+
+}// namespace __detail
+
+// enumerable thread local storage should only be created from make function
+template <typename _ValueType, typename... Args>
+__detail::__enumerable_thread_local_storage<_ValueType, Args...>
+__make_etls(Args&&... __args)
+{
+    return __detail::__enumerable_thread_local_storage<_ValueType, Args...>(std::forward<Args>(__args)...);
+}
```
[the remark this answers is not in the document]
danhoeflinger: done.

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: Yes, I believe `algorithm_impl.h` will always be included alongside `histogram_impl.h` from all paths, so we can do just include the forward declarations.

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: You are correct,  removed.

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: I've simplified it using indices.

include/oneapi/dpl/pstl/histogram_impl.h
danhoeflinger: Refactored with the indices, and renamed to `__hist_start_id` and `__hist_end_id`.

akukanov (approved):

mmichel11 (approved):
I have looked through the changes since my last approval, and it LGTM as well.

PR #1979, opened 2024-12-22
Do not use deprecated sub-group load/store extension

dmitriy-sobolev:
Code sections which use `sycl::sub_group::<load()|store()>` DPC++ extension are not entirely disabled. Non-template expressions are still checked by a compiler even if they are within a discarded `if constexpr` branch. 

This makes the code non-compatible with compilers other than DPC++, and it results in a deprecation warning with the most recent DPC++ builds.

A mock-up example: https://godbolt.org/z/svhnY14r5

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -492,7 +492,7 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
                     const ::std::uint16_t __subgroup_id = __subgroup.get_group_id();
                     const ::std::uint16_t __subgroup_size = __subgroup.get_local_linear_range();
 
-#if _ONEDPL_SYCL_SUB_GROUP_LOAD_STORE_PRESENT
+#if _ONEDPL_LIBSYCL_SUB_GROUP_LOAD_STORE_PRESENT
                     constexpr bool __can_use_subgroup_load_store =
                         _IsFullGroup && oneapi::dpl::__internal::__range_has_raw_ptr_iterator_v<::std::decay_t<_InRng>>;
 #else
@@ -502,6 +502,7 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
                     auto __lacc_ptr = __dpl_sycl::__get_accessor_ptr(__lacc);
                     if constexpr (__can_use_subgroup_load_store)
                     {
+#if _ONEDPL_LIBSYCL_SUB_GROUP_LOAD_STORE_PRESENT
                         _ONEDPL_PRAGMA_UNROLL
                         for (::std::uint16_t __i = 0; __i < _ElemsPerItem; ++__i)
                         {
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -598,7 +602,7 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
                     const ::std::uint16_t __subgroup_id = __subgroup.get_group_id();
                     const ::std::uint16_t __subgroup_size = __subgroup.get_local_linear_range();
 
-#if _ONEDPL_SYCL_SUB_GROUP_LOAD_STORE_PRESENT
+#if _ONEDPL_LIBSYCL_SUB_GROUP_LOAD_STORE_PRESENT
```
MikeDvorskiy, line 605: 1. This is a "logic mix" of preprocessor code and C++ code.... It looks a mess..
I would recommend "have a look" on C++ code which remains after the preprocessor keeps one branch "#if" and the other branch "#else" and make the code clean-up (here and everywhere in that PR).

2. BTW, `_ONEDPL_LIBSYCL_SUB_GROUP_LOAD_STORE_PRESENT` is defined as identical zero.
https://github.com/oneapi-src/oneDPL/pull/1979/files?diff=split&w=0#diff-521f7db06a55567c1a1dffa855dde585fb6bc5fbe8633d19b528356e3a501ea0R91

M.b. it makes sense to remove the branches `_ONEDPL_LIBSYCL_SUB_GROUP_LOAD_STORE_PRESENT` at all?

dmitriy-sobolev: I've removed all the branches. Initially, they were there to preserve the logic if we decide to add newer versions of sub-group load/store. If we settle with this, I can create an issue to remember about this optimization opportunity.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -492,59 +492,26 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
-                        _ONEDPL_PRAGMA_UNROLL
-                        for (::std::uint16_t __i = 0; __i < _ElemsPerItem; ++__i)
-                        {
-                            auto __idx = __i * _WGSize + __subgroup_id * __subgroup_size;
-                            auto __val = __unary_op(__subgroup.load(__in_rng.begin() + __idx));
-                            __subgroup.store(__lacc_ptr + __idx, __val);
-                        }
-                    }
-                    else
+                    for (::std::uint16_t __idx = __item_id; __idx < __n; __idx += _WGSize)
```
mmichel11, line 496: Can we change all of the `::std` instances to `std` in the affected lines of code?

mmichel11:
Removing the sub-group load store logic seems fine to me as it is currently unused. It may be worth creating an issue to investigate the performance difference and determine if it is worth it to adopt.

I suggest we wait for @adamfidel's input when he returns since I believe he introduced sub-group load / store path here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
dmitriy-sobolev: Done

adamfidel (approved):
Thanks for cleaning this up. The idea was to come back and re-enable this code path at a later point, but since this has been dead code for some time now, I agree it makes sense to just remove it for now.

mmichel11 (approved):

PR #1986, opened 2025-01-02
Fix DEFAULT_ACCESSOR_CONSTRUCTOR_PRESENT macro

dmitriy-sobolev:
The feature is mistakenly enabled with older compilers, e.g. with icpx 2022.2.1, leading to:

> error: constructor for 'oneapi::dpl::__par_backend_hetero::__usm_or_buffer_accessor<sycl::accessor<unsigned long, 1, sycl::access::mode::write, sycl::access::target::global_buffer, sycl::access::placeholder::false_t>>' must explicitly initialize the member '__acc' which does not have a default constructor


danhoeflinger (approved):
LGTM,  assuming the original assessment from #1410 is accurate.  (Supported by hotfixes of 2023.1 and 2023.2 but not by 2023.2.0)
This fixes the mistake in the recent changes. which flipped the `&&` to `||`.

PR #1987, opened 2025-01-03
Bugfix for reduce-by-segment with differing input and output types

mmichel11:
For differing input and output key / value types, we cannot store an input element into a variable that is the output type. This is especially important for discard iterators in the output where we are met with a compile-time error if we attempt to store an input into this type as it can hold no value.

A test case has been added which combines zip and discard iterators. This catches the compilation issue before the fix.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -939,12 +939,10 @@ struct __write_red_by_seg
         using std::get;
         auto __out_keys = get<0>(__out_rng.tuple());
         auto __out_values = get<1>(__out_rng.tuple());
-        using _KeyType = oneapi::dpl::__internal::__value_t<decltype(__out_keys)>;
-        using _ValType = oneapi::dpl::__internal::__value_t<decltype(__out_values)>;
 
-        const _KeyType& __next_key = get<2>(__tup);
-        const _KeyType& __current_key = get<3>(__tup);
-        const _ValType& __current_value = get<1>(get<0>(__tup));
+        const auto& __next_key = get<2>(__tup);
+        const auto& __current_key = get<3>(__tup);
+        const auto& __current_value = get<1>(get<0>(__tup));
         const bool __is_seg_end = get<1>(__tup);
         const std::size_t __out_idx = get<0>(get<0>(__tup));
 
--- a/test/parallel_api/numeric/numeric.ops/reduce_by_segment_zip.pass.cpp
+++ b/test/parallel_api/numeric/numeric.ops/reduce_by_segment_zip.pass.cpp
@@ -124,8 +124,7 @@ test_zip_with_discard()
     int values1[n] = {1, 1, 1, 1, 1};
     int values2[n] = {2, 2, 2, 2, 2};
     int output_keys[n] = {};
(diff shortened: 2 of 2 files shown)


test/parallel_api/numeric/numeric.ops/reduce_by_segment_zip.pass.cpp
```diff
@@ -108,6 +108,65 @@ test_with_usm()
...
+    auto d_output_keys = dt_helper5.get_data();
+    auto d_output_values1 = dt_helper6.get_data();
+    auto d_output_values2 = dt_helper7.get_data();
+
+    //make zip iterators
+    auto begin_keys_in = oneapi::dpl::make_zip_iterator(d_keys1, d_keys2);
+    auto end_keys_in = oneapi::dpl::make_zip_iterator(d_keys1 + n, d_keys2 + n);
+    auto begin_vals_in = oneapi::dpl::make_zip_iterator(d_values1, d_values2);
+    auto begin_keys_out = oneapi::dpl::make_zip_iterator(d_output_keys, oneapi::dpl::discard_iterator());
+    auto begin_vals_out = oneapi::dpl::make_zip_iterator(d_output_values1, d_output_values2);
```
danhoeflinger, line 151: Does it make sense to include a discard iterator in the values as well?  I think right now we are just hitting a compilation issue just with the key side of the issue, not the value side, right?

mmichel11: Yes, it makes sense to add this for completeness as the same issue would occur with a discard iterator in the output values. I have added it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -939,12 +939,10 @@ struct __write_red_by_seg
...
         auto __out_values = get<1>(__out_rng.tuple());
-        using _KeyType = oneapi::dpl::__internal::__value_t<decltype(__out_keys)>;
-        using _ValType = oneapi::dpl::__internal::__value_t<decltype(__out_values)>;
 
-        const _KeyType& __next_key = get<2>(__tup);
-        const _KeyType& __current_key = get<3>(__tup);
-        const _ValType& __current_value = get<1>(get<0>(__tup));
+        const auto& __next_key = get<2>(__tup);
+        const auto& __current_key = get<3>(__tup);
+        const auto& __current_value = get<1>(get<0>(__tup));
```
danhoeflinger, line 945: I'm not sure if there is any advantage, but we could get the type with `std::tuple_element_t<I, _Tup>` rather than using `auto`.  
Effectively I think we end up with the same code, without much more information about the types were using.  Just mentioning it to see if you find reason to use it.

danhoeflinger (approved):
LGTM, Good catch.  If you want to make a `tuple_element` change, I'll re-approve then.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I think I agree.  The only thing I stopped to think about is if somehow with the lvalue ref `auto`, we are possibly being less specific than we intend to be, but I definitely agree that what you have now is much cleaner.

I'm good with how you have it, I think.

PR #1990, opened 2025-01-06
Fix ambiguous call to sycl::min

dmitriy-sobolev:
`__local_mem_size` may differ from `_Size` in some cases leading to "error: call to 'min' is ambiguous".


adamfidel (approved):
LGTM.

Another alternative would be to change the `auto __local_mem_size = /*...*/` to `_Size __local_mem_size = /*...*/`, but what you have is good as well!

PR #1991, opened 2025-01-06
Make kernel compilation logic more portable

dmitriy-sobolev:
Disable kernel compilation if it is not supported. For instance, ACPP does not support kernel bundles, nor `sycl::program`, which appears to be a former DPC++ feature or an artifact from previous SYCL revisions.


MikeDvorskiy (approved):
LGTM

Issue #1805, opened 2024-08-29
Issue with make_zip_view with certain input types


danhoeflinger:
I don't like that this change is required, but it mirrors how it is currently handled in the ranges implementation. https://github.com/oneapi-src/oneDPL/blob/aec12a4c55ac5de94753f8e3259b50cd6befb0c3/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h#L364
Its needed here because we are unifying the ranges API and main API handling of `__parallel_copy_if` to enable the performance improvements to be felt in both cases, and also to remove redundant code.

This is to avoid an issue with making a `zip_view` from a non const lvalue reference to a zip view.  There may be changes we can make to our internal tuple class to resolve this, but I think that lies outside the scope of this PR.

_Originally posted by @danhoeflinger in https://github.com/oneapi-src/oneDPL/pull/1763#discussion_r1707551425_
           
An issue arises in reduce_by_key_ranges_sycl.pass when using `make_zip_view` and forwarding instead of using `zip_view` constructor here with a copy of the range.  This seems wrong, and may imply changes with zip_view or our internal tuple.

danhoeflinger:
This should be fixed with #2391.  There are a few places where we are using zip_view constructor when we should use make_zip_view, that we can switch back.

PR #1997, opened 2025-01-10
Work around crashes and incorrect results in scan-based algorithms when compiling with -O0

mmichel11:
On certain integrated graphics architectures, sub-group sizes of 32 are not supported for kernels with certain properties when compiled with -O0 using the `icpx` compiler. The compiler is normally able to workaround this issue by compiling to a sub-group size of 16 instead. However, in cases in which an explicit sub-group size is required, then the compiler throws an exception at JIT time. This issue directly affects our reduce-then-scan implementation which has a required sub-group size of 32.

To properly work around this issue, several things must be done. Firstly, exception handling is implemented to catch this synchronous exception while re-throwing any other exceptions back to the user. Secondly, after discussion with compiler developers, kernel compilation must be separated from execution of the kernel to prevent corruption of the underlying `sycl::queue` that occurs when this exception is thrown after implicit buffer accessor dependencies around the kernel have been established. To do this, kernel bundles are used to first compile the kernel before executing.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -2498,7 +2498,7 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, cons
     }
 #endif
     return __parallel_reduce_by_segment_fallback(
-        oneapi::dpl::__internal::__device_backend_tag{}, std::forward<_ExecutionPolicy>(__exec),
+        oneapi::dpl::__internal::__device_backend_tag{}, __exec,
         std::forward<_Range1>(__keys), std::forward<_Range2>(__values), std::forward<_Range3>(__out_keys),
         std::forward<_Range4>(__out_values), __binary_pred, __binary_op,
         oneapi::dpl::unseq_backend::__has_known_identity<_BinaryOperator, __val_type>{});
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1442,9 +1442,9 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag __backend_tag, 
     using _CopyOp = unseq_backend::__copy_by_mask<_ReduceOp, _Assign,
                                                   /*inclusive*/ std::true_type, 1>;
 
-    return __parallel_scan_copy(__backend_tag, std::forward<_ExecutionPolicy>(__exec),
-                                std::forward<_InRng>(__in_rng), std::forward<_OutRng>(__out_rng), __n,
-                                _CreateOp{__pred}, _CopyOp{_ReduceOp{}, __assign});
+    return __parallel_scan_copy(__backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_InRng>(__in_rng),
+                                std::forward<_OutRng>(__out_rng), __n, _CreateOp{__pred},
+                                _CopyOp{_ReduceOp{}, __assign});
 }
 
 #if _ONEDPL_COMPILE_KERNEL
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -2491,7 +2491,6 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, cons
                 auto __res = oneapi::dpl::__par_backend_hetero::__parallel_reduce_by_segment_reduce_then_scan(
                     oneapi::dpl::__internal::__device_backend_tag{}, __exec, __keys, __values, __out_keys, __out_values,
                     __binary_pred, __binary_op);
-                __res.wait();
                 // Because our init type ends up being tuple<std::size_t, ValType>, return the first component which is the write index. Add 1 to return the
                 // past-the-end iterator pair of segmented reduction.
                 return std::get<0>(__res.get()) + 1;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1134,17 +1134,18 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
             using _ScanInputTransform = oneapi::dpl::__internal::__no_op;
             using _WriteOp = oneapi::dpl::__par_backend_hetero::__simple_write_to_id;
 
-            auto __opt_return = __handle_sync_sycl_exception(
-                [=, &__exec] {
-                    _GenInput __gen_transform{__unary_op};
-                    return __parallel_transform_reduce_then_scan(
-                        __backend_tag, __exec, std::move(__in_rng), std::move(__out_rng), __gen_transform, __binary_op,
-                        __gen_transform, _ScanInputTransform{}, _WriteOp{}, __init, _Inclusive{},
-                        /*_IsUniquePattern=*/std::false_type{});
-                },
-                __bypass_sycl_kernel_not_supported{});
-            if (__opt_return)
-                return __opt_return.value();
+            try
+            {
+                _GenInput __gen_transform{__unary_op};
+                return __parallel_transform_reduce_then_scan(__backend_tag, __exec, __in_rng, __out_rng,
+                                                             __gen_transform, __binary_op, __gen_transform,
+                                                             _ScanInputTransform{}, _WriteOp{}, __init, _Inclusive{},
+                                                             /*_IsUniquePattern=*/std::false_type{});
+            }
+            catch (const sycl::exception& e)
+            {
+                __bypass_sycl_kernel_not_supported(e);
+            }
         }
 #endif
     }
@@ -1295,17 +1296,18 @@ __parallel_unique_copy(oneapi::dpl::__internal::__device_backend_tag __backend_t
 #if _ONEDPL_COMPILE_KERNEL
     if (oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec))
     {
-        auto __opt_return = __handle_sync_sycl_exception(
-            [=, &__exec] {
-                using _GenMask = oneapi::dpl::__par_backend_hetero::__gen_unique_mask<_BinaryPredicate>;
-                using _WriteOp = oneapi::dpl::__par_backend_hetero::__write_to_id_if<1, _Assign>;
-                return __parallel_reduce_then_scan_copy(__backend_tag, __exec, std::move(__rng), std::move(__result),
-                                                        __n, _GenMask{__pred}, _WriteOp{_Assign{}},
-                                                        /*_IsUniquePattern=*/std::true_type{});
-            },
-            __bypass_sycl_kernel_not_supported{});
-        if (__opt_return)
-            return __opt_return.value();
+        try
+        {
+            using _GenMask = oneapi::dpl::__par_backend_hetero::__gen_unique_mask<_BinaryPredicate>;
+            using _WriteOp = oneapi::dpl::__par_backend_hetero::__write_to_id_if<1, _Assign>;
+            return __parallel_reduce_then_scan_copy(__backend_tag, __exec, __rng, __result, __n, _GenMask{__pred},
+                                                    _WriteOp{_Assign{}},
+                                                    /*_IsUniquePattern=*/std::true_type{});
+        }
+        catch (const sycl::exception& e)
+        {
+            __bypass_sycl_kernel_not_supported(e);
+        }
     }
 #endif
     using _ReduceOp = std::plus<decltype(__n)>;
@@ -1344,8 +1346,8 @@ __parallel_reduce_by_segment_reduce_then_scan(oneapi::dpl::__internal::__device_
     assert(__n > 1);
     return __parallel_transform_reduce_then_scan(
         __backend_tag, std::forward<_ExecutionPolicy>(__exec),
-        oneapi::dpl::__ranges::make_zip_view(std::forward<_Range1>(__keys), std::forward<_Range2>(__values)),
-        oneapi::dpl::__ranges::make_zip_view(std::forward<_Range3>(__out_keys), std::forward<_Range4>(__out_values)),
+        oneapi::dpl::__ranges::zip_view(std::forward<_Range1>(__keys), std::forward<_Range2>(__values)),
+        oneapi::dpl::__ranges::zip_view(std::forward<_Range3>(__out_keys), std::forward<_Range4>(__out_values)),
         _GenReduceInput{__binary_pred}, _ReduceOp{__binary_op}, _GenScanInput{__binary_pred, __n},
         _ScanInputTransform{}, _WriteOp{__binary_pred, __n},
         oneapi::dpl::unseq_backend::__no_init_value<oneapi::dpl::__internal::tuple<std::size_t, _ValueType>>{},
@@ -1362,18 +1364,19 @@ __parallel_partition_copy(oneapi::dpl::__internal::__device_backend_tag __backen
 #if _ONEDPL_COMPILE_KERNEL
     if (oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec))
     {
-        auto __opt_return = __handle_sync_sycl_exception(
-            [=, &__exec] {
-                using _GenMask = oneapi::dpl::__par_backend_hetero::__gen_mask<_UnaryPredicate>;
-                using _WriteOp =
-                    oneapi::dpl::__par_backend_hetero::__write_to_id_if_else<oneapi::dpl::__internal::__pstl_assign>;
-                return __parallel_reduce_then_scan_copy(__backend_tag, __exec, std::move(__rng), std::move(__result),
-                                                        __n, _GenMask{__pred}, _WriteOp{},
-                                                        /*_IsUniquePattern=*/std::false_type{});
-            },
-            __bypass_sycl_kernel_not_supported{});
-        if (__opt_return)
-            return __opt_return.value();
+        try
+        {
+            using _GenMask = oneapi::dpl::__par_backend_hetero::__gen_mask<_UnaryPredicate>;
+            using _WriteOp =
+                oneapi::dpl::__par_backend_hetero::__write_to_id_if_else<oneapi::dpl::__internal::__pstl_assign>;
+            return __parallel_reduce_then_scan_copy(__backend_tag, __exec, __rng, __result, __n, _GenMask{__pred},
+                                                    _WriteOp{},
+                                                    /*_IsUniquePattern=*/std::false_type{});
+        }
+        catch (const sycl::exception& e)
+        {
+            __bypass_sycl_kernel_not_supported(e);
+        }
     }
 #endif
     using _ReduceOp = std::plus<decltype(__n)>;
@@ -1418,17 +1421,18 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag __backend_tag, 
 #if _ONEDPL_COMPILE_KERNEL
     else if (oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec))
     {
-        auto __opt_return = __handle_sync_sycl_exception(
-            [=, &__exec] {
-                using _GenMask = oneapi::dpl::__par_backend_hetero::__gen_mask<_Pred>;
-                using _WriteOp = oneapi::dpl::__par_backend_hetero::__write_to_id_if<0, _Assign>;
-                return __parallel_reduce_then_scan_copy(__backend_tag, __exec, std::move(__in_rng),
-                                                        std::move(__out_rng), __n, _GenMask{__pred}, _WriteOp{__assign},
-                                                        /*_IsUniquePattern=*/std::false_type{});
-            },
-            __bypass_sycl_kernel_not_supported{});
-        if (__opt_return)
-            return __opt_return.value();
+        try
+        {
+            using _GenMask = oneapi::dpl::__par_backend_hetero::__gen_mask<_Pred>;
+            using _WriteOp = oneapi::dpl::__par_backend_hetero::__write_to_id_if<0, _Assign>;
+            return __parallel_reduce_then_scan_copy(__backend_tag, __exec, __in_rng, __out_rng, __n, _GenMask{__pred},
+                                                    _WriteOp{__assign},
+                                                    /*_IsUniquePattern=*/std::false_type{});
+        }
+        catch (const sycl::exception& e)
+        {
+            __bypass_sycl_kernel_not_supported(e);
+        }
     }
 #endif
     using _ReduceOp = std::plus<_Size>;
@@ -1466,7 +1470,7 @@ __parallel_set_reduce_then_scan(oneapi::dpl::__internal::__device_backend_tag __
 
     return __parallel_transform_reduce_then_scan(
         __backend_tag, std::forward<_ExecutionPolicy>(__exec),
-        oneapi::dpl::__ranges::make_zip_view(
+        oneapi::dpl::__ranges::zip_view(
             std::forward<_Range1>(__rng1), std::forward<_Range2>(__rng2),
             oneapi::dpl::__ranges::all_view<std::int32_t, __par_backend_hetero::access_mode::read_write>(
                 __mask_buf.get_buffer())),
@@ -1537,14 +1541,15 @@ __parallel_set_op(oneapi::dpl::__internal::__device_backend_tag __backend_tag, c
 #if _ONEDPL_COMPILE_KERNEL
     if (oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec))
     {
-        auto __opt_return = __handle_sync_sycl_exception(
-            [=, &__exec] {
-                return __parallel_set_reduce_then_scan(__backend_tag, __exec, std::move(__rng1), std::move(__rng2),
-                                                       std::move(__result), __comp, __is_op_difference);
-            },
-            __bypass_sycl_kernel_not_supported{});
-        if (__opt_return)
-            return __opt_return.value();
+        try
+        {
+            return __parallel_set_reduce_then_scan(__backend_tag, __exec, __rng1, __rng2, __result, __comp,
+                                                   __is_op_difference);
+        }
+        catch (const sycl::exception& e)
+        {
+            __bypass_sycl_kernel_not_supported(e);
+        }
     }
 #endif
     return __parallel_set_scan(__backend_tag, __exec, std::forward<_Range1>(__rng1), std::forward<_Range2>(__rng2),
@@ -2481,19 +2486,20 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, cons
     {
         if (oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec))
         {
-            auto __opt_return = __handle_sync_sycl_exception(
-                [=, &__exec] {
-                    auto __res = oneapi::dpl::__par_backend_hetero::__parallel_reduce_by_segment_reduce_then_scan(
-                        oneapi::dpl::__internal::__device_backend_tag{}, __exec, std::move(__keys), std::move(__values),
-                        std::move(__out_keys), std::move(__out_values), __binary_pred, __binary_op);
-                    __res.wait();
-                    // Because our init type ends up being tuple<std::size_t, ValType>, return the first component which is the write index. Add 1 to return the
-                    // past-the-end iterator pair of segmented reduction.
-                    return std::get<0>(__res.get()) + 1;
-                },
-                __bypass_sycl_kernel_not_supported{});
-            if (__opt_return)
-                return __opt_return.value();
+            try
+            {
+                auto __res = oneapi::dpl::__par_backend_hetero::__parallel_reduce_by_segment_reduce_then_scan(
+                    oneapi::dpl::__internal::__device_backend_tag{}, __exec, __keys, __values, __out_keys, __out_values,
+                    __binary_pred, __binary_op);
+                __res.wait();
+                // Because our init type ends up being tuple<std::size_t, ValType>, return the first component which is the write index. Add 1 to return the
+                // past-the-end iterator pair of segmented reduction.
+                return std::get<0>(__res.get()) + 1;
+            }
+            catch (const sycl::exception& e)
+            {
+                __bypass_sycl_kernel_not_supported(e);
+            }
         }
     }
 #endif
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1143,7 +1143,7 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
                         _ScanInputTransform{}, _WriteOp{}, __init, _Inclusive{},
                         /*_IsUniquePattern=*/std::false_type{});
                 },
-                oneapi::dpl::__par_backend_hetero::__bypass_sycl_kernel_not_supported{});
+                __bypass_sycl_kernel_not_supported{});
             if (__opt_return)
                 return __opt_return.value();
         }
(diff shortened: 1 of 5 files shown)


SergeyKopienko:
I never seen before in our code the examples of usage `std::optional'.
From my point of view we may simple write `try / catch` instead of `__handle_sync_sycl_exception` and don't use `std::optional' at all.

mmichel11:
> I never seen before in our code the examples of usage `std::optional'. From my point of view we may simple write `try / catch`instead of`__handle_sync_sycl_exception`and don't use`std::optional' at all.

I don't have a strong preference if we choose to go with the way I implemented or just directly add try ... catch throughout this header. I initially did this to avoiding having many try catch statements here.

Let me leave this open first for others' opinions to see if they prefer a more functional approach or just directly adding try...catch throughout.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -756,12 +756,17 @@ __parallel_transform_reduce_then_scan(oneapi::dpl::__internal::__device_backend_
...
-        __reduce_then_scan_scan_kernel<_CustomName>>;
     using _ValueType = typename _InitType::__value_type;
+    using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
+    using _ReduceKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
+        __reduce_then_scan_reduce_kernel, _CustomName, _InRng, _OutRng, _GenReduceInput, _ReduceOp, _InitType,
+        _Inclusive, _IsUniquePattern>;
+    using _ScanKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
+        __reduce_then_scan_scan_kernel, _CustomName, _InRng, _OutRng, _GenScanInput, _ReduceOp, _ScanInputTransform,
+        _WriteOp, _InitType, _Inclusive, _IsUniquePattern>;
+    static auto __kernels = __internal::__kernel_compiler<_ReduceKernel, _ScanKernel>::__compile(__exec);
```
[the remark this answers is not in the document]
dmitriy-sobolev: > I wanted to point this unique case out and ensure there are no issues with this approach.

I am thinking about possible issues with uniqueness of the kernels in some corner cases. Is it possible to end up with different kernels with the same instantiation of `__parallel_transform_reduce_then_scan`? I can imagine that, for example: 

```
// no kernel specified: unnamed lambda case
sycl::queue queue_a{selector_vendor_a{});
sycl::queue queue_b{selector_vendor_b{});

// policy_a and policy_b have the same type
dpl::execution::device_policy policy_a(queue_a);
dpl::execution::device_policy policy_b(queue_b);

// ... containers and predicates have the same types
dpl::copy_if(policy_a, ...);
dpl::copy_if(policy_b, ...); // will it use the kernels compiled for the queue_a (and thus device "a")?
```
Such cases are highly unlikely, and I cannot think of any others. I would rather document this as a known limitation than compromise performance (if it is large enough overhead).

dmitriy-sobolev (commented):
I agree with the approach, but I need to review the PR more thoroughly. Indeed, according to the SYCL 2020 spec, launching a kernel decorated with `reqd_sub_group_size` attribute may throw an exception depending on what inside that kernel...

I have a question regarding that part:

> Secondly, after discussion with compiler developers, kernel compilation must be separated from execution of the kernel to prevent corruption of the underlying sycl::queue

Is it a limitation/bug of the DPC++ SYCL implementation?

mmichel11:
> I agree with the approach, but I need to review the PR more thoroughly. Indeed, according to the SYCL 2020 spec, launching a kernel decorated with `reqd_sub_group_size` attribute may throw an exception depending on what inside that kernel...
> 
> I have a question regarding that part:
> 
> > Secondly, after discussion with compiler developers, kernel compilation must be separated from execution of the kernel to prevent corruption of the underlying sycl::queue
> 
> Is it a limitation/bug of the DPC++ SYCL implementation?

I think it is more of a limitation with the existing implementation. From my discussion, it seems as if there is no easy fix which is why this workaround was recommended. From their perspective, it is not a bug as the SYCL specification does not specify any behavior regarding resubmissions to a queue after an exception such as this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1099,7 +1099,7 @@ struct __write_to_id_if_else
 template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _UnaryOperation, typename _InitType,
           typename _BinaryOperation, typename _Inclusive>
 auto
-__parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _ExecutionPolicy&& __exec,
+__parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backend_tag, const _ExecutionPolicy& __exec,
```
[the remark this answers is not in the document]
mmichel11: There's a tricky scenario with aligning the future return type from the different potential paths when we use the forwarding reference. I will take another look and see if it can be implemented with the forwarding reference.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -89,6 +89,15 @@
 
 #define _ONEDPL_SYCL_DEVICE_COPYABLE_SPECIALIZATION_BROKEN (_ONEDPL_LIBSYCL_VERSION_LESS_THAN(70100))
 
+// Macro to check if the exception thrown when a kernel cannot be ran on a device does not align with
+// sycl::errc::kernel_not_supported as required by the SYCL spec. Detects the Intel DPC++ and open-source intel/llvm
+// compilers.
+#ifdef _ONEDPL_LIBSYCL_VERSION
```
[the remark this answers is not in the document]
mmichel11: Limited it up to `20250200` for now with the `icpx` compiler. With an open-source compiler build, I believe this will be enabled until we manually turn it off since there is no `__INTEL_LLVM_COMPILER` macro there.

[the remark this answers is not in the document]
dmitriy-sobolev: What about defining the macro like that:

```
// Future version, conservative estimation
#define _ONEDPL_SYCL_KERNEL_NOT_SUPPORTED_EXCEPTION_BROKEN (_ONEDPL_LIBSYCL_VERSION_LESS_THAN(90000))
```

Which is simpler than:

```
#if defined(_ONEDPL_LIBSYCL_VERSION) && (!defined(__INTEL_LLVM_COMPILER) || __INTEL_LLVM_COMPILER < 20250200)
#    define _ONEDPL_ICPX_KERNEL_NOT_SUPPORTED_EXCEPTION_BROKEN 1
#else
#    define _ONEDPL_ICPX_KERNEL_NOT_SUPPORTED_EXCEPTION_BROKEN 0
#endif
```
It sticks to the version of the library, which has that bug (not a compiler).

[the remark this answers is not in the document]
mmichel11: I agree. There is no estimate when this issue will be fixed, so I just used the next libsycl major version as suggested.

dmitriy-sobolev (commented):
Let me suggest some adjustments according the latest changes in main branch

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -756,12 +756,17 @@ __parallel_transform_reduce_then_scan(oneapi::dpl::__internal::__device_backend_
...
                                       _Inclusive, _IsUniquePattern)
 {
-    using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
-    using _ReduceKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-        __reduce_then_scan_reduce_kernel<_CustomName>>;
-    using _ScanKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
-        __reduce_then_scan_scan_kernel<_CustomName>>;
     using _ValueType = typename _InitType::__value_type;
+    using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
+    using _ReduceKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
```
dmitriy-sobolev, line 761: A nitpick. I am curious if constants such as `__sub_group_size` and `__max_inputs_per_item` should also be included into the kernel names. I think they are not strictly necessary right now since `__sub_group_size` is the same, and `__max_inputs_per_item` depends on the input type, which is already indirectly included into the name. I would suggest leaving a comment about it...

dmitriy-sobolev (commented):
Here is a couple of minor suggestions.

Overall, the PR looks good to me. I have no strong preference regarding the previous topic about how to pass an execution policy (const lref or forwarding reference): both options are fine by me.

dmitriy-sobolev:
> I think it is more of a limitation with the existing DPC++ implementation. From my discussion, it seems as if there is no easy fix which is why this workaround was recommended. From their perspective, it is not a bug as the SYCL specification does not specify any behavior regarding resubmissions to a queue after an exception such as this.

I think that It would be more natural for the SYCL implementation to preserve valid state of the queue after throwing an exception: these exceptions and queues are a part of one framework (SYCL). I guess we can ask SYCL spec maintainers for the clarification. I hope they will confirm it, which can be used as an argument to motivate the compiler folks.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: You are correct that these are not necessary right now. I did not include them, so we do not further increase the size of the kernel name which is user unfriendly (especially when profiling). However, if this ever changes, then we need to be careful here, so I have added a comment.

mmichel11:
> > I never seen before in our code the examples of usage `std::optional'. From my point of view we may simple write `try / catch`instead of`__handle_sync_sycl_exception`and don't use`std::optional' at all.
> 
> I don't have a strong preference if we choose to go with the way I implemented or just directly add try ... catch throughout this header. I initially did this to avoiding having many try catch statements here.
> 
> Let me leave this open first for others' opinions to see if they prefer a more functional approach or just directly adding try...catch throughout.

@dmitriy-sobolev After looking through the PR, do you have any thoughts on Sergey's question here? Is the functional exception handler approach preferred in your opinion or just directly adding try-catch throughout parallel_backend_sycl.h?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1117,24 +1122,31 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
...
             using _WriteOp = oneapi::dpl::__par_backend_hetero::__simple_write_to_id;
 
-            _GenInput __gen_transform{__unary_op};
-
-            return __parallel_transform_reduce_then_scan(
-                __backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__in_rng),
-                std::forward<_Range2>(__out_rng), __gen_transform, __binary_op, __gen_transform, _ScanInputTransform{},
-                _WriteOp{}, __init, _Inclusive{}, /*_IsUniquePattern=*/std::false_type{});
+            auto __opt_return = __handle_sync_sycl_exception(
+                [=, &__exec] {
```
adamfidel, line 1138: Is the omission of parenthesis for this lambda a deliberate choice?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -442,7 +437,7 @@ struct __parallel_reduce_then_scan_scan_submitter<
     operator()(_ExecutionPolicy&& __exec, const sycl::nd_range<1> __nd_range, _InRng&& __in_rng, _OutRng&& __out_rng,
                _TmpStorageAcc& __scratch_container, const sycl::event& __prior_event,
                const std::uint32_t __inputs_per_sub_group, const std::uint32_t __inputs_per_item,
-               const std::size_t __block_num) const
+               const std::size_t __block_num, sycl::kernel& __scan_kernel) const
```
adamfidel, line 440: Can the kernel be const here or does that cause issues?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Yes, since we don't pass any parameters to the lambda I just omitted them.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: I checked the SYCL spec and the `get_kernel_bundle` member function is const and there are no compilation issues, so I have made these references all const.

dmitriy-sobolev:
> > > I never seen before in our code the examples of usage `std::optional'. From my point of view we may simple write `try / catch`instead of`__handle_sync_sycl_exception`and don't use`std::optional' at all.
> > 
> > 
> > I don't have a strong preference if we choose to go with the way I implemented or just directly add try ... catch throughout this header. I initially did this to avoiding having many try catch statements here.
> > Let me leave this open first for others' opinions to see if they prefer a more functional approach or just directly adding try...catch throughout.
> 
> @dmitriy-sobolev After looking through the PR, do you have any thoughts on Sergey's question here? Is the functional exception handler approach preferred in your opinion or just directly adding try-catch throughout parallel_backend_sycl.h?

I slightly prefer Sergey's option. `__handle_sync_sycl_exception` does not reuse much as an abstraction, but adds complexity. In contrast, `__bypass_sycl_kernel_not_supported` is helpful here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -842,6 +843,44 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
...
+    void
+    operator()(const sycl::exception& __e) const
+    {
+        // The SYCL spec compliant solution would be to compare __e.code() and sycl::errc::kernel_not_supported
+        // and rethrow the encountered exception if the two do not compare equal. However, the icpx compiler currently
+        // returns a generic error code in violation of the SYCL spec which has a value of 7. If we are using the Intel
+        // compiler, then compare the value of the error code. Otherwise, assume the implementation is spec compliant.
+#if _ONEDPL_SYCL_KERNEL_NOT_SUPPORTED_EXCEPTION_BROKEN
+        if (__e.code().value() != 7)
+            throw;
```
dmitriy-sobolev, line 860: One more question which is not trivial to me. Should we rethrow that exception? oneDPL algorithm selection causes it by using an optional feature, handles it and choses a backup option. I see no value by passing it to the user, but I am not very experienced here, so I would seek for @MikeDvorskiy advice.

mmichel11:
@SergeyKopienko @dmitriy-sobolev I have removed `__handle_sync_sycl_exception`, turned `__bypass_sycl_kernel_not_supported` into a function, and made some other necessary changes to directly use try...catch.

I will run some internal testing to fully verify the change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
mmichel11: To clarify this issue, there are some instances where forwarding the execution policy in one path and not forwarding it in the other could result in differing future return types as the `_ExecutionPolicy` class template in `__result_and_scratch_storage` which is bundled in the future return may or may not be references of `device_policy`. Making this a const l-value reference aligns the return.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1329,45 +1346,51 @@ __parallel_reduce_by_segment_reduce_then_scan(oneapi::dpl::__internal::__device_
     assert(__n > 1);
     return __parallel_transform_reduce_then_scan(
         __backend_tag, std::forward<_ExecutionPolicy>(__exec),
-        oneapi::dpl::__ranges::make_zip_view(std::forward<_Range1>(__keys), std::forward<_Range2>(__values)),
-        oneapi::dpl::__ranges::make_zip_view(std::forward<_Range3>(__out_keys), std::forward<_Range4>(__out_values)),
+        oneapi::dpl::__ranges::zip_view(std::forward<_Range1>(__keys), std::forward<_Range2>(__values)),
```
mmichel11, line 1349: This change was necessary in a few places. When the range forwarding was removed in some places, `make_zip_view` resulted in a compiler error. It seemed to be caused by being unable to process l-value references to ranges. With the previous approach, we captured by value in a lambda and moved it in so this was a non-issue.

It looks like it is an instance of this issue: https://github.com/uxlfoundation/oneDPL/issues/1805.

mmichel11:
clang-format wants to add an additional 4 space indentation throughout both the scan kernel implementations due to some of the changes. I disagree with this as we already have a lot of indentation here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1329,45 +1346,51 @@ __parallel_reduce_by_segment_reduce_then_scan(oneapi::dpl::__internal::__device_
...
-        using _GenMask = oneapi::dpl::__par_backend_hetero::__gen_mask<_UnaryPredicate>;
-        using _WriteOp =
-            oneapi::dpl::__par_backend_hetero::__write_to_id_if_else<oneapi::dpl::__internal::__pstl_assign>;
-
-        return __parallel_reduce_then_scan_copy(__backend_tag, std::forward<_ExecutionPolicy>(__exec),
-                                                std::forward<_Range1>(__rng), std::forward<_Range2>(__result), __n,
-                                                _GenMask{__pred}, _WriteOp{}, /*_IsUniquePattern=*/std::false_type{});
+        try
+        {
+            using _GenMask = oneapi::dpl::__par_backend_hetero::__gen_mask<_UnaryPredicate>;
```
SergeyKopienko, line 1369: I think better to move typedefs at this place and in other try / catch blocks outside of try / catch.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2449,24 +2481,30 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
+                    oneapi::dpl::__internal::__device_backend_tag{}, __exec, __keys, __values, __out_keys, __out_values,
+                    __binary_pred, __binary_op);
+                __res.wait();
+                // Because our init type ends up being tuple<std::size_t, ValType>, return the first component which is the write index. Add 1 to return the
+                // past-the-end iterator pair of segmented reduction.
+                return std::get<0>(__res.get()) + 1;
+            }
+            catch (const sycl::exception& e)
+            {
+                __bypass_sycl_kernel_not_supported(e);
```
SergeyKopienko, line 2501: What we are going to return from this `__parallel_reduce_by_segment` function if we suppress exception?
The same question make sense for other calls of `__bypass_sycl_kernel_not_supported` too...

SergeyKopienko: Understood, it's OK

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2449,24 +2481,30 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
-            return std::get<0>(__res.get()) + 1;
+            try
+            {
+                auto __res = oneapi::dpl::__par_backend_hetero::__parallel_reduce_by_segment_reduce_then_scan(
+                    oneapi::dpl::__internal::__device_backend_tag{}, __exec, __keys, __values, __out_keys, __out_values,
+                    __binary_pred, __binary_op);
+                __res.wait();
+                // Because our init type ends up being tuple<std::size_t, ValType>, return the first component which is the write index. Add 1 to return the
+                // past-the-end iterator pair of segmented reduction.
+                return std::get<0>(__res.get()) + 1;
```
SergeyKopienko, line 2497: Probably this call `__res.wait();` isn't required because the same call should have place inside `__res.get()`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -842,6 +842,25 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
...
+    // and rethrow the encountered exception if the two do not compare equal. However, the icpx compiler currently
+    // returns a generic error code in violation of the SYCL spec which has a value of 7. If we are using the Intel
+    // compiler, then compare the value of the error code. Otherwise, assume the implementation is spec compliant.
+#if _ONEDPL_SYCL_KERNEL_NOT_SUPPORTED_EXCEPTION_BROKEN
+    if (__e.code().value() != 7)
+        throw;
+#else // Generic SYCL compiler. Assume it is spec compliant.
+    if (__e.code() != sycl::errc::kernel_not_supported)
+        throw;
+#endif
```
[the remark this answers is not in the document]
mmichel11: The issue with this approach is that `7` is not the error code itself but the value. The error code is some generic, undocumented error code which is why we call `__e.code().value()` which compares equal to `7`.

We could unify it with your approach by using the value of `sycl::errc::kernel_not_supported` by static casting it to an integer (`sycl::errc` is an enum class), but this is less clean in my opinion and I prefer the small code duplication here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Either way is not a big deal to me, but my thought is constraining the scope of the alias as much as possible makes the most sense. Since it's only needed in this try...catch, it makes sense to limit it to this scope to me.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Good catch, fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1506,22 +1534,26 @@ __parallel_set_scan(oneapi::dpl::__internal::__device_backend_tag __backend_tag,
...
-    {
-        return __parallel_set_scan(__backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__rng1),
-                                   std::forward<_Range2>(__rng2), std::forward<_Range3>(__result), __comp,
-                                   __is_op_difference);
+        try
+        {
+            return __parallel_set_reduce_then_scan(__backend_tag, __exec, __rng1, __rng2, __result, __comp,
+                                                   __is_op_difference);
+        }
+        catch (const sycl::exception& e)
```
dmitriy-sobolev, line 1549: Uglification: `e` -> `__e`.

dmitriy-sobolev (commented):
A minor comment. I am ready to approve the PR once it has been addressed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: The opposite opinion - to include into `try` block only the code which really may generate exception.
Writing aliases inside it you make me read what was written here and understand that it's only aliases.
But up to you.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: ...or `__exc`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I see your point here. I have made these changes.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I think `__e` is sufficiently clear in this context. I made this change.

mmichel11: I think `__e` is sufficiently clear in this context. I made this change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -842,6 +842,25 @@ class __static_monotonic_dispatcher<::std::integer_sequence<::std::uint16_t, _X,
...
+// a required sub-group size of 32.
+inline void
+__bypass_sycl_kernel_not_supported(const sycl::exception& __e)
+{
+    // The SYCL spec compliant solution would be to compare __e.code() and sycl::errc::kernel_not_supported
+    // and rethrow the encountered exception if the two do not compare equal. However, the icpx compiler currently
+    // returns a generic error code in violation of the SYCL spec which has a value of 7. If we are using the Intel
+    // compiler, then compare the value of the error code. Otherwise, assume the implementation is spec compliant.
+#if _ONEDPL_SYCL_KERNEL_NOT_SUPPORTED_EXCEPTION_BROKEN
+    if (__e.code().value() != 7)
```
SergeyKopienko, line 856: 1) What is `code()` ?
"
const std::error_code& code() const noexcept
Returns the error code stored inside the exception.
"
2) What is `std::error_code` ?
https://en.cppreference.com/w/cpp/error/error_code :
"std::error_code represents a **platform-dependent** error code value."

So the first question: is it really SAFE to process error code with numeric const value `7` ?
The second question: my be are able experimentally detect the corresponding `__e.code()` state for it?

SergeyKopienko, line 856: 1) What is `code()` ?
"
const std::error_code& code() const noexcept
Returns the error code stored inside the exception.
"
2) What is `std::error_code` ?
https://en.cppreference.com/w/cpp/error/error_code :
"std::error_code represents a **platform-dependent** error code value."

So the first question: is it really SAFE to process error code with numeric const value `7` ?
The second question: my be are able experimentally detect the corresponding `__e.code()` state for it?

dmitriy-sobolev (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: 1. In the case of the SYCL spec, the returned enum from `code()` aligns with the `errc` enum class here: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:exception.class. In the else of this preprocessor check, this is perfectly safe.
2. It is not truly safe, but this is because the compiler is violating the SYCL spec by returning the incorrect error code. I took another look, and the error code prints as `sycl:7`. I was told this is some generic code, but looking at the SYCL spec it does compare equal to `sycl::errc::build` which is guaranteed to have a value of `7` per the spec. If it alleviates your concern here, I can compare to this error code instead of the value.

SergeyKopienko: (2) For my point of view it will be much more safe.

mmichel11: I adopted a similar suggestion to what you initially recommended here. I do think the `switch` statement is a bit overkill since there are only two possibilities, so I just used a single `if`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -850,15 +850,16 @@ __bypass_sycl_kernel_not_supported(const sycl::exception& __e)
 {
     // The SYCL spec compliant solution would be to compare __e.code() and sycl::errc::kernel_not_supported
     // and rethrow the encountered exception if the two do not compare equal. However, the icpx compiler currently
-    // returns a generic error code in violation of the SYCL spec which has a value of 7. If we are using the Intel
-    // compiler, then compare the value of the error code. Otherwise, assume the implementation is spec compliant.
+    // returns a sycl::errc::build in violation of the SYCL spec. If we are using the Intel compiler, then compare
+    // to this error code. Otherwise, assume the implementation is spec compliant.
+    const std::error_code __kernel_not_supported_ec =
```
dmitriy-sobolev, line 855: Let's include `<system_error>` to make sure `std::error_code` is available.

mmichel11: Done

dmitriy-sobolev (approved):
Reapproving

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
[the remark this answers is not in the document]
akukanov: We will need to have a proper fix.
Generally, caching kernels is not our business, but the one of the SYCL runtime. Caching memory is not our business, but the one of a memory manager.
If we do caching ourselves, we need to check if the cached value is valid for the call - for example, if it was compiled for the device for which the algorithm is called. Documenting it as a limitation is not a long-term substitute for a proper, reliable implementation.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
akukanov: The `ExecutionPolicy` type should not be used in templates etc. without first clearing from cv qualifiers and references. According to the standard, even the check of the `is_execution_policy` trait when the overload set is determined should use `decay`/`remove_cvref`. Sounds like `__result_and_scratch_storage` implementation or usage has a bug that needs to be fixed, instead of changing the pattern signatures.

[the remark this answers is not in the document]
danhoeflinger: Yes, I can make a fix for this to be included for this release.  My thought was to figure out some way to specify get the proper type in a way that didn't require the decay to make usage easier, but we should first fix the bug.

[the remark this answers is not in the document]
danhoeflinger: I believe #2031 should resolve any issues.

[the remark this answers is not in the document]
akukanov: Does it also revert the changes made by this patch? That should be a good test :)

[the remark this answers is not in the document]
danhoeflinger: No, but I can make that change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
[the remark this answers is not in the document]
mmichel11: Thank you for the response. Since this does not seem like a good option to keep, it might be worth considering to incur the performance hit and remove the `static` usage. The performance impact I've measured to be around a constant `3-4 us` with this caching approach for small inputs after the first setup. When incurring the additional overhead each call, it adds about another `4-5` us for for a total hit of `8-9 us` on Intel GPU Series Max 1550 for an `inclusive_scan` call.

Once we scale our inputs past cache sizes the impact of this change is minimized with a lower threshold if we are loading data from global memory in the cold cache case. If this performance impact seems acceptable, then I can open a PR to make this change.

The true overarching issue here is that the SYCL specification does not specify recovery behavior after handling this sort of exception which has led to the need for this workaround. I've been told that it has been discussed at the Khronos working group before but there was no consensus.

PR #1999, opened 2025-01-13
RFC for "Passed Directly" Customization Point

danhoeflinger:
Provides a proposal for a public customization point for users to define to indicate if their types are passed directly to sycl kernels.

This RFC is intended to address and resolve #1939.

Here is a working proof of concept to play with: https://godbolt.org/z/jvo1esoeb (updated)


danhoeflinger:
I have a working proof of concept for this in godbolt, but needs some cleanup before sharing.  Let me know if that would be useful to provide here, but likely any needed details can and should be provided in the text.

rfcs/proposed/passed_directly_API/README.md
```diff
@@ -0,0 +1,144 @@
...
+have expressed that this is undesirable.
+
+### Testing
+We will need a detailed test checking both positive and negative responses to `is_passed_directly_to_sycl_kernels` come
+as expected, with custom types and combinations of iterators, usm pointers etc.
+
+## Open Questions
+
+Is there a better / more concise name than `is_passed_directly_to_sycl_kernels` we can use which properly conveys the
+meaning to the users?
```
masterleinad, line 141: I would prefer if the name somewhere includes `onedpl`. The function is not generally relevant for `SYCL`. Maybe
- `is_passed_directly_to_onedpl`, or
-  `is_passed_directly_to_onedpl_kernels`

rfcs/proposed/passed_directly_API/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+Should we be targeting Experimental or fully supported with this proposal?
+ (Do we think user feedback is required to solidify an interface / experience?)
```
masterleinad, line 144: I guess the general design of passed directly has been tested internally pretty well at least.

masterleinad (commented):
Apart from the name, the approach looks pretty reasonable to me.

rfcs/proposed/passed_directly_API/README.md
danhoeflinger: Yes, its a fair point that since the customization point in user's code wont be in the `oneapi::dpl` namespace, there is little connecting it to oneDPL unless it is in the name.  I was trying to be more descriptive about the semantic meaning as this is only relevant for the SYCL based dpcpp backend however that is probably less important than including oneDPL.

Thanks, I think the first of your suggestions is probably the best option so far.

rfcs/proposed/passed_directly_API/README.md
danhoeflinger: Agreed.  From my perspective the only reason to keep it in experimental to start with is if we are uncertain of the exact API specifics or to find any unexpected gotchas with the approach of using a customization point generally as opposed to some other option.  

Lets see what others have to say, but I'm leaning toward targeting supported, and just adding it to the specification directly.

rfcs/proposed/passed_directly_API/README.md
adamfidel: As this property only really makes sense for the SYCL backend of oneDPL, there is a sense that `is_passed_directly_to_onedpl` doesn't quite tell the full story of what it is for. It won't be used in the TBB or OpenMP backends of oneDPL, which you wouldn't understand from just the name alone.

On the other hand, something like `is_passed_directly_to_sycl` is too broad and could be confusing because the SYCL implementation won't use the function directly. Perhaps the second suggestion of `is_passed_directly_to_onedpl_kernels` is closer, but one could take issue with its verbosity.

danhoeflinger: Perhaps `is_passed_directly_to_onedpl_dpcpp` is the right choice (I hate how wordy it is but I'm not sure we have a choice).
A normal user of oneDPL may better recognize "dpcpp" than "kernels", saves a couple characters too.

akukanov: Actially I would prefer not to introduce more names with `dpcpp`, to avoid the impression that our implementation is based on DPC++ (as opposed to a generic SYCL).

danhoeflinger: `is_passed_directly_to_onedpl_device_backend`
`is_passed_directly_to_onedpl_sycl`
`is_passed_directly_to_onedpl_kernels`

We could also drop the "to" and shorten some words...  

`is_pass_direct_onedpl_sycl`?

masterleinad: I like the first one (`is_passed_directly_to_onedpl_device_backend`). Maybe, you would rather want to talk about policies than backends, though, so `is_passed_directly_in_onedpl_device_policies` to make a clear connection to https://uxlfoundation.github.io/oneDPL/parallel_api/pass_data_algorithms.html#pass-data-to-algorithms. I don't mind the verbosity too much.

danhoeflinger: I've updated to use `is_passed_directly_in_onedpl_device_policies`, but we will see what others think.

rfcs/proposed/passed_directly_API/README.md
```diff
@@ -0,0 +1,144 @@
...
+
+    struct my_passed_directly_type
+    {
+        /* unspecified user definition */
+    };
+
+    template <typename It1, typename It2>
+    constexpr
+    bool
+    is_passed_directly_to_onedpl(const my_passed_directly_type&)
```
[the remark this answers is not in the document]
danhoeflinger: Ah good catch.  I originally was using the template types rather than the member fields themselves to make the logic here.

rfcs/proposed/passed_directly_API/README.md
```diff
@@ -0,0 +1,144 @@
...
+under some circumstances (based usually on their base types).
+
+Internally, these rules are defined with a trait `oneapi::dpl::__ranges::is_passed_directly<T>` which evaluates to
+`std::true_type` or `std::false_type` to indicate whether the type `T` should be passed directly to sycl kernels.
+There exists a unofficial legacy `is_passed_directly` trait which types can define like this:
+`using is_passed_directly = std::true_type;` which is supported within oneDPL. This method is currently used for a
+number of helper types within the SYCLomatic compatibility headers, (`device_pointer`, `device_iterator`,
+`tagged_pointer`, `constant_iterator`, `iterator_adaptor`). There is no official public API for users who want to
+create their own types which could be passed directly to SYCL kernels, this is a gap we should fill in with an official
+public API.
```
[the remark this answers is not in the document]
danhoeflinger: Yes, that is true.  This is only for iterator types, not callables or scalars etc.  I'll update it.

rfcs/proposed/passed_directly_API/README.md
```diff
@@ -0,0 +1,144 @@
...
+This option can exist in concert with existing methods, the legacy `is_passed_directly` typedef in types, the internal
+`oneapi::dpl::__ranges::is_passed_directly` trait specializations. It would be possible to simplify the internal
+implementation away from explicit specializations of the trait to the customization point, but that is not required
+at first implementation.
+
+### Implementation details
+To make this robust, we will follow an C++17 updated version of what is discussed in
+[Eric Niebler's Post](https://ericniebler.com/2014/10/21/customization-point-design-in-c11-and-beyond/), using a
+callable, and using an `inline constexpr` to avoid issues with ODR and to avoid issues with resolving customization
+points when not separating the call to two steps with a `using` statement first.
```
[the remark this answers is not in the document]
danhoeflinger: Perhaps I am mistaken, but my understanding is that Niebler is describing a way using function pointers to allow _qualified_ calls to also pick up the default implementation rather than just _unqualified_ calls with the `using` statement.  
We don't need his more complex strategy for ODR because of changes in C++17, but I think the function pointer strategy is still a benefit.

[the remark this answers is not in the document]
masterleinad: > Perhaps I am mistaken, but my understanding is that Niebler is describing a way using function pointers to allow qualified calls to also pick up the default implementation rather than just unqualified calls with the using statement.

See https://godbolt.org/z/TM914E6fv for an example.

rfcs/proposed/passed_directly_API/README.md
```diff
@@ -0,0 +1,144 @@
...
+### Require specifically named typedef / using in user's type
+We could simply make official our requirements for user's types to include a typedef or using statement to define if the
+type is passed directly like `using is_passed_directly = std::true_type;`, where the absence of this would be equivalent
+to a `std::false_type`. 
+
+However, this clutters the user type definitions with specifics of oneDPL. It also may not be as clear what this
+signifies for maintenance of user code without appropriate comments describing the details of oneDPL and SYCL. Users
+have expressed that this is undesirable.
+
+### Testing
```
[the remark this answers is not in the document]
danhoeflinger: fixed.

rfcs/proposed/passed_directly_API/README.md
```diff
@@ -0,0 +1,144 @@
+# Passed Directly Customiation Point for User Defined Types
+
+## Introduction
+
+OneDPL handles some types of input data automatically as input to its dpcpp (sycl-based) backend as described
```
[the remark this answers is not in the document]
danhoeflinger: fixed in a handful of places.

rfcs/proposed/passed_directly_API/README.md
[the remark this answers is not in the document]
danhoeflinger: I cleaned up and added my proof of concept to the description as well for another example.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,151 @@
...
+
+Create a customization point `oneapi::dpl::is_passed_directly_in_onedpl_device_policies` free function which allows
+users to mark their types as passed directly:
+
+```
+template <typename T>
+constexpr bool is_passed_directly_in_onedpl_device_policies(const T&);
+```
+
+oneDPL will provide a default implementation which will defer to the existing trait:
```
mmichel11, line 34: What header should a user include to get the default implementation? Execution? I think it would be good to specify it here.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,151 @@
...
+    }
+} //namespace user
+```
+
+Users can use any constexpr logic based on their type to determine if the type can be passed directly into a SYCL kernel
+without any processing. Below is an example of a type which contains a pair of iterators, and should be treated as
+passed directly if and only if both base iterators are also passed directly. oneDPL will use this customization point
+internally when determining how to handle incoming data, picking up any user defined customizations.
+
+When using device policies, oneDPL will check iterator types by calling `is_passed_directly_in_onedpl_device_policies`
```
akukanov, line 71: Is it supposed to be a compile-time evaluation or a run-time function call?

rfcs/proposed/passed_directly_API/README.md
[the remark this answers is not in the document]
akukanov: > my understanding is that Niebler is describing a way using function pointers to allow qualified calls to also pick up the default implementation rather than just unqualified calls with the using statement.

I do not see where he uses any function pointers. His `std::begin` is a reference to an instance of  `struct std::__detail::__begin_fn` which has a function call operator, so that `std::begin(X)` is a valid code. This operator internally uses an unqualified call to `begin`, which "default" implementation is in `std::__detail`. As I said, it's more or less matches the CPO design in C++20.

And it's not what you proposed so far, as far as I can see. In this proposal, the default implementation is a free function in the `oneapi::dpl` namespace, and customization is a function in the user's type namespace (it is not said explicitly that the namespace should be the same, but it should for ADL to work). So I do not see how a qualified call will take specializations, neither how an unqualified call will take the default implementation without a using declaration.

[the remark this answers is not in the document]
akukanov: > my understanding is that Niebler is describing a way using function pointers to allow qualified calls to also pick up the default implementation rather than just unqualified calls with the using statement.

I do not see where he uses any function pointers. His std::begin is a reference to an instance of struct std::__detail::__begin_fn which has a function call operator, so that std::begin(X) is a valid code. This operator internally uses an unqualified call to begin, which "default" implementation is in std::__detail. As I said, it's more or less matches the CPO design in C++20.

And it's not what you proposed so far, as far as I can see. In this proposal, the default implementation is a free function in the oneapi::dpl namespace, and customization is a function in the user's type namespace (it is not said explicitly that the namespace should be the same, but it should for ADL to work). So I do not see how a qualified call will take specializations, neither how an unqualified call will take the default implementation without a using declaration.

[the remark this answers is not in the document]
danhoeflinger: I think I need to improve the language to be more specific and accurate about the proposed implementation details, but I think what you describe in your first paragraph is correct and accurate to my intentions.  

You are correct that an unqualified call does require a `using` declaration, but a qualified call should take the user customizations because a qualified call will use the function object.  The function object internally makes a unqualified call from the namespace of the default implementation on behalf of the user, allowing it to either find the more specific user customization if it exists, or end up in the default otherwise.  You should be able to see this tested in the proof of concept [here](https://godbolt.org/z/K71osE8Te) on line 121.

Hopefully I'm not missing something, but I may be.

rfcs/proposed/iterators_pass_directly_customization/README.md
danhoeflinger: Its a good question, and I think I agree that execution is the appropriate place.

rfcs/proposed/iterators_pass_directly_customization/README.md
danhoeflinger: As I experiment a little more, I'm realizing that perhaps we can't combine the customization point with a compile-time evaluation requirement, or some other requirements on the iterator types themselves like default constructible.

I'm running into issues attempting to use it within `enable_if` / SFINAE without some way to take the type alone and turn it into a const lvalue ref argument for ADL to work.  If I can call a default constructor, it works, but I'm not sure what other option there is.  

At first I thought `std::declval` was the answer, but it is not.  It is, in fact, both not `constexpr`, and it would be being used in an evaluated expression here. I'll continue experimenting...

danhoeflinger: I think its only a roadblock for using it in SFINAE, and not in general for `constexpr`. I believe we should be able to refactor our code to lean instead on `if constexr` for "passed directly" checks.

danhoeflinger: I found a way to deal use it with SFINAE as well, it required returning `std::true_type{}` or `std::false_type` rather than `bool`.  This seems like the right way to go. I've updated the PR accordingly.

rfcs/proposed/passed_directly_API/README.md
akukanov: I agree that there is not much need for an experimental phase. But a POC with practical usage outside of oneDPL is necessary I think.

rfcs/proposed/passed_directly_API/README.md
rarutyun: I (honestly) would say `is_passed_directly_to_sycl_backend`. We use 'backend' term in our documentation, so it should be clear enough. Please don't repeat `_onedpl_` in the name, because it's already in `oneapi::dpl` namespace. Duplication doesn't bring clarity. I am 100% agree with the intent to _not_ use `dpcpp` in the name. The rarer we use it, the better

rfcs/proposed/iterators_pass_directly_customization/README.md
rarutyun: `<oneapi/dpl/traits>`?

rfcs/proposed/passed_directly_API/README.md
danhoeflinger: After some development from the original idea, the only public API is the trait which will include the `oneapi::dpl` namespace.  However, users will still be overriding the customization point by defining a function in their own namespace, and I believe that name should be associated with the trait (its name + `_v`).  Without repeating onedpl in the name of the customization point, there will be nothing tying it to oneDPL in the users code.  This is the motivation for including `_onedpl_`.  

We could consider different names for the customization point and the trait, but that may also be confusing.

rfcs/proposed/iterators_pass_directly_customization/README.md
danhoeflinger: Now that the RFC has evolved a bit to make the public API a trait rather than directly calling the customization point, I think `oneapi/dpl/type_traits` does make more sense than `oneapi/dpl/execution`.

mmichel11: `oneapi/dpl/type_traits` makes sense to me as well.

rfcs/proposed/passed_directly_API/README.md
akukanov: We do not use the word 'backend" in the public API (actually, not at all in the specification), and I think we should not.

There are two elements of the API to name - the trait value (also a class?) and the user-defined customization function. Their names should be related, but I am not sure if almost exact match is needed. I agree that the name of the function should refer to oneDPL; that can be achieved by adding a prefix to the function name.

Also, as far as I understand, the trait is not so much (if at all) about the iterator itself "passed" to a device, as it is about the data "underneath" the iterator being accessible from that device, so that no data copying/no intermediate buffer is required.

Can it be something like `onedpl_is_iterator_device_ready()` for the function and `oneapi::dpl::is_iterator_device_ready[_v]` for the trait?

rfcs/proposed/iterators_pass_directly_customization/README.md
akukanov: I think it should better go to `<oneapi/dpl/iterator>`.

`type_traits` is OK in principle, but, as far as I can tell, this thing is closer semantically to a concept representing an iterator category (like `forward_iterator`) or a requirement (like `sortable`) than to a generic type trait.

danhoeflinger:
[This section of the specification](https://github.com/uxlfoundation/oneAPI-spec/blob/cfa6ee354f0b38d78ec7623380a8845097c9d4c9/source/elements/oneDPL/source/parallel_api/execution_policies.rst?plain=1#L97C1-L104C36) makes me think that the API we end up with for this RFC should be an extension.  

The specification leaves the way data is passed as unspecified, so unless we want to significantly expand that section of the specification, I think an extension is probably the correct place for this.

danhoeflinger:
> [This section of the specification](https://github.com/uxlfoundation/oneAPI-spec/blob/cfa6ee354f0b38d78ec7623380a8845097c9d4c9/source/elements/oneDPL/source/parallel_api/execution_policies.rst?plain=1#L97C1-L104C36) makes me think that the API we end up with for this RFC should be an extension rather than a full part of the specification.
> 
> The specification leaves the way data is passed as unspecified, so unless we want to significantly expand that section of the specification, I think an extension is probably the correct place for this.

Let me walk that back, after further consideration and discussion with @rarutyun. We do already specify some of the context we need in the specification, and it would not be too much more to add enough to describe using our provided iterator types, and define passed directly in that context as well as how it could be leveraged by users.

I do think it can end up in the full specification, and will add something valuable to for others who may implement oneDPL.

The undefined part of the specification is about host data, which is orthogonal here really.

rfcs/proposed/iterators_pass_directly_customization/README.md
danhoeflinger: I don't have a strong opinion between these two options.  You are correct that it specifically will augment a custom iterator type, so I don't object to `oneapi/dpl/iterator`.  The inclusion there may perhaps also help users find some combination of our provided custom iterators which may serve their purposes without requiring their own customizations in the first place.

.gitignore
```diff
@@ -1,5 +1,5 @@
 # -------- CMake --------
-build/*
+build*/*
```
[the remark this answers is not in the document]
danhoeflinger: Good catch, this was an accidental check-in.  Its something I keep in my working copy which is nice for me, but I haven't attempted to impose it on others.

rfcs/proposed/passed_directly_API/README.md
danhoeflinger: I've added an open question about the naming, which can be discussed in the specification PR when we add it soon.

rfcs/proposed/iterators_pass_directly_customization/README.md
danhoeflinger: I've added an open question about where this should live.  We can discuss in the soon to come specification PR.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,170 @@
...
+
+## Testing
+We will need a detailed test checking both positive and negative responses to
+`oneapi::dpl::is_passed_directly_in_onedpl_device_policies_v` to ensure they have the expected result. This should
+include tests with custom types, combinations of iterators, USM pointers, etc.
+
+## Open Questions
+
+* Is there a better, more concise name than `oneapi::dpl::is_passed_directly_in_onedpl_device_policies[_v]` that
+* properly conveys the meaning to the users?
```
masterleinad, line 164:
```suggestion
* Is there a better, more concise name than `oneapi::dpl::is_passed_directly_in_onedpl_device_policies[_v]` that
  properly conveys the meaning to the users?
```

masterleinad:
> [This section of the specification](https://github.com/uxlfoundation/oneAPI-spec/blob/cfa6ee354f0b38d78ec7623380a8845097c9d4c9/source/elements/oneDPL/source/parallel_api/execution_policies.rst?plain=1#L97C1-L104C36) makes me think that the API we end up with for this RFC should be an extension rather than a full part of the specification.

In my opinion, it's important to clarify when the user can pass in device data and when the data is expected to be in host memory. That section doesn't really do that.

rfcs/proposed/iterators_pass_directly_customization/README.md
danhoeflinger: thanks, fixed.

danhoeflinger:
> > [This section of the specification](https://github.com/uxlfoundation/oneAPI-spec/blob/cfa6ee354f0b38d78ec7623380a8845097c9d4c9/source/elements/oneDPL/source/parallel_api/execution_policies.rst?plain=1#L97C1-L104C36) makes me think that the API we end up with for this RFC should be an extension rather than a full part of the specification.
> 
> In my opinion, it's important to clarify when the user can pass in device data and when the data is expected to be in host memory. That section doesn't really do that.

We can try to improve the language of the specification either in the context of the specification PR related to this feature, or in a separate specification PR.

akukanov (approved):
I approve the proposed design. The API needs to be finalized and approved for a oneDPL specification update.

@masterleinad does the proposed design meet your needs and expectations?

@rarutyun @dmitriy-sobolev @MikeDvorskiy - please take a look. The goal is to settle on the design and make sure it does not have significant flaws.

masterleinad:
> @masterleinad does the proposed design meet your needs and expectations?

Yes, the design looks reasonable to me.

danhoeflinger:
I have a tentative implementation at #2126.  This is not ready for review and should not be seriously considered until we have this and a specification PR merged.  

However, since I have it, I thought I'd link it in case its useful.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,170 @@
...
+### High Level Proposal
+
+We will create an Argument-Dependent Lookup (ADL) customization point which defines if iterators are "passed directly"
+`is_passed_directly_in_onedpl_device_policies`. Users may define if their iterator types should be "passed directly" by
+oneDPL by defining a constexpr function in the same namespace where the iterator is defined.
+`is_passed_directly_in_onedpl_device_policies` must accept a const lvalue reference to the iterator type being
+specialized and return `std::true_type{}` if the iterator type is "passed directly" and `std::false_type{}` otherwise.
+
+Additionally, oneDPL will provide a public trait,
+`inline constexpr bool oneapi::dpl::is_passed_directly_in_onedpl_device_policies_v<T>`, indicating if the iterator type
```
MikeDvorskiy, line 34: 1) `oneapi::dpl::is_passed_directly_in_onedpl_device_policies_v` has double  info about "oneDPL":
I mean "oneapi::dpl" and substring "_onedpl_".
2) A suffix "_device_policies" is too long.... I would vote to omit that.
So, I would suggest name `oneapi::dpl::is_passed_directly_v`

danhoeflinger: For now, I think the idea is to merge this RFC with the naming and location as an open question, to be decided in the discussion of the specification PR.  All the same, thanks for your input here.

Also some context for the individual points:
1) The idea behind this is that the public trait follows the naming convention of the ADL customization point function.  The ADL customization point function needs to identify itself as `onedpl` because it user's customizations will be in their own namespaces.  However, we could decide not to have the public trait share naming with the ADL function, and I would agree with removing this redundancy then.  

2) There was an earlier discussion in this PR https://github.com/uxlfoundation/oneDPL/pull/1999#discussion_r1913701632 , I think its a difficult decision with many opinions.  I don't really have strong opinions here and am happy to go with the group's decision.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,170 @@
+# Passed Directly Customization Point for User Defined Types
+
+## Introduction
+
+oneDPL handles some types of input data automatically as input to its device backend (SYCL-based) as described
```
MikeDvorskiy, line 5: If I were a user (not oneDPL developer) I would had not understood what phrase "handles some types of input data automatically as input to its device backend" mean... And also I would had not understood what is "passed directly" and why it might be helpful.
I think we have to  explain it more clear by saying about copying data (under an iterator) between the host and device in implicit way in case of "passed directly" is false.

danhoeflinger: Take a look and see if the changes help clarify this section.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,174 @@
...
+## Testing
+We will need a detailed test checking both positive and negative responses to
+`oneapi::dpl::is_passed_directly_in_onedpl_device_policies_v` to ensure they have the expected result. This should
+include tests with custom types, combinations of iterators, USM pointers, etc.
+
+## Open Questions
+
+* Is there a better, more concise name than `oneapi::dpl::is_passed_directly_in_onedpl_device_policies[_v]` that
+properly conveys the meaning to the users?
+    * Other names proposed:
```
MikeDvorskiy, line 169: `oneapi::dpl::is_passed_directly_to_device` ?

MikeDvorskiy (commented):
Excepting the names, I am OK with that proposal,

rfcs/proposed/iterators_pass_directly_customization/README.md
danhoeflinger: I added the two proposed names you raised and a note to the open questions section.

dmitriy-sobolev (approved):
I've also checked the proposal with the example, and it looks good to me since it is well defined and implementable. 

Let me approve it. Note that I have not participated in the discussions, so I may lack some intricate details and it may be better to get an additional approval.

akukanov (approved):
I have reviewed the changes since my approval, and I confirm it.

rfcs/proposed/iterators_pass_directly_customization/README.md
danhoeflinger: >Are we also going to introduce a struct as a trait?
 
Its a good question, and probably needs to be handled in connection with the question of naming, and if the ADL function should be named the same as the trait.  

Now I am thinking that it would be better to make their names distinct, and then to offer both the struct and value version of the trait
`oneapi::dpl::is_passed_directly_in_onedpl_device_policies[_v]` as you describe.  This is the more normal convention, but requires divorcing the names between the ADL customization point and the trait.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,179 @@
...
+    constexpr
+    auto
+    is_passed_directly_in_onedpl_device_policies(const my_passed_directly_iterator&)
+    {
+        return std::true_type{};
+    }
+} //namespace user
+```
+
+Users can use any `constexpr` logic based on their iterator to determine if the iterator can be passed directly into a
```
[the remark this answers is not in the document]
akukanov: "Users can use any logic" it says.

It is a requirement that the function is `constexpr`, and it is stated earlier. Remember that this is not a formal specification but design description.

The proposal says "When using device policies, oneDPL will run **compile-time checks on argument iterator types** by calling is_passed_directly_in_onedpl_device_policies.". It might be not 100% accurate but the intent is clear (at least to me).

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,179 @@
...
+    {
+        /* unspecified user definition */
+    };
+
+    constexpr
+    auto
+    is_passed_directly_in_onedpl_device_policies(const my_passed_directly_iterator&)
+    {
+        return std::true_type{};
+    }
```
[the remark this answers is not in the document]
akukanov: I would consider this not a typical example. Typically there should be some logic, like in the next example.

rfcs/proposed/iterators_pass_directly_customization/README.md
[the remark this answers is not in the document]
danhoeflinger: I've improved the language a bit to make it clear that any logic may be used, as long as it is `constexpr` logic.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,179 @@
...
+    constexpr 
+    auto is_passed_directly_in_onedpl_device_policies(const iterator_pair<It1, It2>&)
+    {
+        if constexpr (oneapi::dpl::is_passed_directly_in_onedpl_device_policies_v<It1> &&
+                      oneapi::dpl::is_passed_directly_in_onedpl_device_policies_v<It2>)
+            return std::true_type{};
+        else 
+            return std::false_type{}; 
+    }
+} //namespace user
```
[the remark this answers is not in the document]
akukanov: It is more concise but not really simpler. Brevity does not equal to clarity. I personally think that the "decltype" variant is harder to comprehend (and so to write properly).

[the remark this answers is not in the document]
danhoeflinger: While I'm not sure that your alternative is "simpler", I've included it as another example for those who do. 

I prefer the first version for readability, however I may be missing some advantage the second provides technically.

rfcs/proposed/iterators_pass_directly_customization/README.md
danhoeflinger: I've separated the names of the ADL customization point and the public trait.  This allows us to provide both the struct and value version of the trait, which is the norm.

rfcs/proposed/iterators_pass_directly_customization/README.md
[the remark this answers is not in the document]
danhoeflinger: Thinking about the way we use it for our own internal iterator types, I think normally there will be some logic, but there are also some iterator types which have signatures like this.  (`counting_iterator`, `discard_iterator`)

akukanov (approved):

danhoeflinger:
The draft implementation #2126 has been updated with the current state of the RFC.  I added the body-less ADL customization point function to the tests, and added the struct version of the public trait (and use it in a few places where it allows for more concise code).  I don't want to focus on that implementation in this phase, but I wanted to ensure there was nothing problematic about the design, and I don't find anything so far.

rfcs/proposed/iterators_pass_directly_customization/README.md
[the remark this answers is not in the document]
rarutyun: In our private discussion with Dan we figured out that we are not talking about `constexpr` function but we are talking about necessity to get the result at compile time. So, a function return type is always a compile-time information, so we should be fine even if users don't put `constexpr` on their customizations. I trust Dan can improve the wording

rfcs/proposed/iterators_pass_directly_customization/README.md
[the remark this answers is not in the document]
rarutyun: When I said "simpler", I referred to the first example that I've provided. I called it "simpler" because it's "logically simpler" in my opinion. In other words, doesn't have redundant branches. In the original RFC text if was written in the style:
```cpp
if (condition)
    return true;
else
    return false;
```
But it seems that you, guys, disagree with me. You might rewrite it with `std::conjuction` if it makes it clearer...

Speaking about a `decltype`: The hole point is when we allow `decltype` scenario we don't force users to write a function definition when having a declaration is sufficient. Some (including myself) would choose to write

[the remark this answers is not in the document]
danhoeflinger: So, this has evolved a little.  
In an earlier iteration, we only had the `_v` boolean value typed public trait, not the `std::true_type`/`std::false type` public trait as well.  There are definitely places in the draft implementation where it makes more sense to `return condition` as you mention.  

I can update the RFC to reflect this better, and mention usage in "hidden friends" and/or "no body" functions.

rfcs/proposed/iterators_pass_directly_customization/README.md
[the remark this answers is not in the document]
danhoeflinger: I adjusted the language here to remove mention of `constexpr` as that is a red herring but rather focus on defining at compile time the return type.

rfcs/proposed/iterators_pass_directly_customization/README.md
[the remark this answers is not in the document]
danhoeflinger: I've adjusted the examples as mentioned.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,201 @@
...
+specialized and return `std::true_type{}` if the iterator type is "passed directly" and `std::false_type{}` otherwise.
+
+Additionally, oneDPL will provide a public trait,
+```cpp
+namespace oneapi
+{
+namespace dpl
+{
+template <typename T>
+struct is_passed_directly_to_device; // std::true_type or std::false_type
```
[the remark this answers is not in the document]
danhoeflinger: I've adjusted this to adopt close to the language of the specification.

danhoeflinger:
I've added some comments about derived classes or non-exact matches for implemented ADL functions, after doing some exploration on my own.  So far, I don't see any big dangers but its something worth thinking more about.  Generally, ambiguity issues can be resolved by implementing a more specific ADL function for the type in question.  It just must be communicated to the user to be aware of this in documentation / specification.
To me, as long as it is documented, matching to derived-from classes is a potentially useful feature without too much of a risk.

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -117,40 +124,55 @@ namespace user
...
+
+```cpp
+    template <typename It1, typename It2>
+    struct iterator_pair
+    {
+        It1 first;
+        It2 second;
+        auto is_passed_directly_in_onedpl_device_policies(const iterator_pair&) ->
+            std::conjunction<oneapi::dpl::is_passed_directly_to_device<It1>,
+                             oneapi::dpl::is_passed_directly_to_device<It2>>;
```
akukanov, line 162: This is not a "hidden friend" function but a method.

danhoeflinger: good catch.  I left out the friend keyword.  I'll fix in a second.

rfcs/proposed/iterators_pass_directly_customization/README.md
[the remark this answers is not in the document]
rarutyun: Don't mind about this wording. It's perhaps even better compared to what I proposed. I just wanted to make sure that we don't create a wrong impression that is_passed_directly **is** a `true_type` or `false_type` (because it's impossible in C++ with structs and possible with aliases, which is not our case)

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,229 @@
...
+of their base iterator(s) using only the base iterator type(s) rather than a named instance. It may also be used to
+confirm user iterator types "passed directly" traits are as intended to prevent unnecessary overhead in oneDPL calls.
+
+### Additional Information
+
+The default implementation of the customization point `is_passed_directly_in_onedpl_device_policies` will be used to
+explicitly mark the following iterators as "passed directly":
+* Pointers (assumes USM pointers)
+* Iterators containing the legacy `using is_passed_directly = true` trait defined within the type definition
+* Iterators to USM shared allocated `std::vectors` (when knowable)
```
rarutyun, line 64:
```suggestion
* Iterators containing the legacy `using is_passed_directly = std::true_type` trait defined within the type definition
* Iterators to USM shared allocated `std::vector`-s (when knowable)
```

rfcs/proposed/iterators_pass_directly_customization/README.md
```diff
@@ -0,0 +1,229 @@
...
+    auto is_passed_directly_in_onedpl_device_policies(const iterator_pair<It1, It2>&)
+    {
+        return std::conjunction<oneapi::dpl::is_passed_directly_to_device<It1>,
+                                oneapi::dpl::is_passed_directly_to_device<It2>>{};
+    }
+} //namespace user
+```
+
+It is also possible to write overloads without a body using an `auto` trailing return type. The following is an
+alternative written in this way for the previous example:
```
rarutyun, line 139: It does not matter whether you use a trailing return type of not. It will still work. So, the suggestion is
```suggestion
It is also possible to write overloads without a body. The following is an
alternative written in this way for the previous example:
```

rarutyun (approved):

PR #2000, opened 2025-01-15
Update broken macro for XPU tests

dmitriy-sobolev:
Updating the macro version, since these tests fail with newer compiler versions (icpx 2025.0.4 and internal builds for the next release).

Some values were also changes:
 - `_PSTL_LAMBDA_PTR_TO_MEMBER_WINDOWS_BROKEN`: set to <20250200 since the update is expected with that compiler version.
 - `_PSTL_TEST_COMPARISON_BROKEN` and  `_PSTL_RED_BY_SEG_WINDOWS_COMPILE_ORDER_BROKEN`: set <20250100 since the issues do not reproduce with icpx 2025.0.
 - `_PSTL_ICPX_FPGA_TEST_USM_VECTOR_ITERATOR_BROKEN` was set to <20250100 since it is expected to be fixed with that version (not-verified yet, but we can raise the version in the future if needed).

### DIFF
--- a/test/support/test_config.h
+++ b/test/support/test_config.h
@@ -258,7 +258,7 @@
 // building for an FPGA device.  This prevents fpga compilation of usm-allocated std vector wrapped in zip, transform,
 // and permutation iterators (as a map).
 #if (TEST_DPCPP_BACKEND_PRESENT && defined(ONEDPL_FPGA_DEVICE) && defined(__INTEL_LLVM_COMPILER) &&                   \
-        __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILE)
+        __INTEL_LLVM_COMPILER < 20250100)
 #    define _PSTL_ICPX_FPGA_TEST_USM_VECTOR_ITERATOR_BROKEN 1
 #else
 #    define _PSTL_ICPX_FPGA_TEST_USM_VECTOR_ITERATOR_BROKEN 0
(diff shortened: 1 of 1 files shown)


SergeyKopienko (dismissed):
LGTM

dmitriy-sobolev:
> LGTM

I've corrected the PR to limit it to the XPU tests only.

test/support/test_config.h
```diff
@@ -209,7 +209,7 @@
 // within a sycl kernel which MSVC uses to allow comparisons with literal zero without warning
 #define _PSTL_TEST_COMPARISON_BROKEN                                                                                   \
     ((__cplusplus >= 202002L || _MSVC_LANG >= 202002L) && _MSVC_STL_VERSION >= 143 && _MSVC_STL_UPDATE >= 202303L &&   \
-    __INTEL_LLVM_COMPILER > 0 && __INTEL_LLVM_COMPILER <= _PSTL_TEST_LATEST_INTEL_LLVM_COMPILER)
+    __INTEL_LLVM_COMPILER > 0 && __INTEL_LLVM_COMPILER <= 20250000)
```
mmichel11, line 212: Should all of these be `20250004` since we had some patch releases? I have not checked, but I would assume the same issues probably apply to the patch release.

mmichel11 (commented):
I took a look at the `reduce_by_segment` case and it LGTM as it passes with 2025.1.0.

test/support/test_config.h
danhoeflinger: what about `< 20250100`

mmichel11: > what about `< 20250100`

That sounds better to me than my suggestion.

SergeyKopienko: I also have thinks about this variant and prefer it.

dmitriy-sobolev: Agree, done

mmichel11 (approved):
Thanks! LGTM

SergeyKopienko (approved):
LGTM

PR #2002, opened 2025-01-16
Ensure `__max_sg_size` is at least 4 in sort

carbotaniuman:
On some implementations, the max subgroup size is 1, which is too small for the sorting algorithm to produce correct results. Make the value at least 4.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
@@ -384,7 +384,7 @@ __submit_selecting_leaf(_ExecutionPolicy&& __exec, _Range&& __rng, _Compare __co
     if (__is_cpu)
     {
         const auto __sg_sizes = __device.template get_info<sycl::info::device::sub_group_sizes>();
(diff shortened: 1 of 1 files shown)


dmitriy-sobolev:
@carbotaniuman, thanks for noting that issue.
The issue is not limited to a CPU device only: I would propose a slightly different solution to fix all device types by correcting the chunk calculation (line 244). I hope you do not mind.

```
// The chunk size must not exceed two sorted sub-sequences to be merged,
// ensuring that at least one work-item processes them.
const _IndexT __chunk = std::min<_IndexT>(__is_cpu ? 32 : 4, __n_sorted * 2);
```

carbotaniuman:
I tested the change on my local installation and it produces the correct results with the new change.

dmitriy-sobolev (approved):
LGTM, but I would ask @MikeDvorskiy to check the changes as well. I may be biased because I also contributed.

MikeDvorskiy (approved):
LGTM

dmitriy-sobolev:
@carbotaniuman, thank you once again. If you want your contributions to be mentioned, you can create a PR with their brief descriptions in https://github.com/uxlfoundation/oneDPL/blob/main/CREDITS.txt.

PR #2008, opened 2025-01-18
Modernize Tested API example, replace cl_int

dmitriy-sobolev:
The initial motivation was to replace `cl_int`, which should not be used, at least in a generic code parts due to being specific to OpenCL backend. Since it involves changing the Tested API example, I prefer to update other parts of it because it is outdated.


SergeyKopienko (approved):
LGTM

SergeyKopienko (approved):
LGTM

PR #2012, opened 2025-01-20
Fix timeouts in std_ranges tests

dmitriy-sobolev:
Processing 32M elements is too much for a debug mode. Let's stick to 256K elements which is enough to use multiple work-group implementation in case of device offloading.

---
std_ranges_sort.pass takes >10min on my laptop with 32M elements and ~11sec with 256K elements.

### DIFF
--- a/test/parallel_api/ranges/std_ranges_test.h
+++ b/test/parallel_api/ranges/std_ranges_test.h
@@ -38,7 +38,12 @@ static_assert(ONEDPL_HAS_RANGE_ALGORITHMS >= 202409L);
 namespace test_std_ranges
 {
 
-inline constexpr std::size_t big_sz = (1<<25) + 10; //32M
+inline constexpr std::size_t big_sz =
+#if PSTL_USE_DEBUG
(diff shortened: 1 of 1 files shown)


test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -38,7 +38,12 @@ static_assert(ONEDPL_HAS_RANGE_ALGORITHMS >= 202409L);
 namespace test_std_ranges
 {
 
-inline constexpr std::size_t big_sz = (1<<25) + 10; //32M
+inline constexpr std::size_t big_sz =
+#if PSTL_USE_DEBUG
+    (1<<18) + 10; //256K
```
MikeDvorskiy, line 43: Actually, we are reducing the test coverage here, because some parallel patterns have code differentiation by data sizes, for example `__parallel_merge`...
I would suggest discussing it offline on one of the next technical meeting.

dmitriy-sobolev:
Closing the PR to reconsider the approach.

PR #2016, opened 2025-01-22
Update version number for the coming release

timmiesmith:
The PR updates the oneDPL version macro and test verifying it.  This PR should not be merged until code freeze for the coming release.

### DIFF
--- a/include/oneapi/dpl/internal/version_impl.h
+++ b/include/oneapi/dpl/internal/version_impl.h
@@ -12,8 +12,8 @@
 
 // The library version
 #define ONEDPL_VERSION_MAJOR 2022
-#define ONEDPL_VERSION_MINOR 7
-#define ONEDPL_VERSION_PATCH 1
+#define ONEDPL_VERSION_MINOR 8
+#define ONEDPL_VERSION_PATCH 0
 
 #if _ONEDPL___cplusplus >= 202002L && __has_include(<version>)
 #    include <version> // The standard C++20 header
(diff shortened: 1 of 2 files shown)


dmitriy-sobolev:
Let's also update this place:

https://github.com/uxlfoundation/oneDPL/blob/9a1dada8a0201f2b2275df540b5c0d1a8b3fe329/documentation/library_guide/conf.py#L40

timmiesmith:
> Let's also update this place:
> 
> https://github.com/uxlfoundation/oneDPL/blob/9a1dada8a0201f2b2275df540b5c0d1a8b3fe329/documentation/library_guide/conf.py#L40

Done.  Thanks for the pointer.

danhoeflinger (approved):
Looks good to me.  It may be a good idea to separate addition `ONEDPL_SPEC_VERSION` and the version update to different commits if it is not too much trouble.  
I can see us wanting to be able to find both of these in the git log history in the future.

danhoeflinger:
> Would mention of both in the commit line be sufficient? I'm not sure how to split a PR into two separate commits when we merge.

Yeah that's probably fine as long as they are both in the top line.  

If you want to split to 2 commits it would probably involve manually rebasing it locally to the commit separation you want and then force pushing (reapproving), then using "rebase and merge" from github rather than "squash and merge". 

As I write this I actually see that we have lost the ability to "rebase and merge" in the change to UXL.  I've used it in the past to do similar things.  Its rare but useful. We may want to look into reenabling that feature.

akukanov:
> As I write this I actually see that we have lost the ability to "rebase and merge" in the change to UXL. I've used it in the past to do similar things. Its rare but useful. We may want to look into reenabling that feature.

Unfortunately, it was a regular source of troubles, with somebody accidentally using it instead of squashing and by that adding multiple intermediate feature development commits into the main branch history.

PR #2013, opened 2025-01-22
Fix compile error in `__serial_merge`

SergeyKopienko:
In this PR we fix compile errors in `__serial_merge` function for case when the types of `__rng2` and `__rng1` aren't the same:
```C++
/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h:162:17: error: conditional expression is ambiguous; '__return_t' (aka 'const oneapi::dpl::__internal::tuple<unsigned long, unsigned long>') can be converted to 'decltype(this->make_reference(::std::declval<_tuple_ranges_t>(), __i, ::std::make_index_sequence<__num_ranges>()))' (aka 'tuple<const unsigned long &, const unsigned long &>') and vice versa
  162 |                 ? __rng2[__rng2_idx++]
      |                 ^ ~~~~~~~~~~~~~~~~~~~~
  163 |                 : __rng1[__rng1_idx++];
      |                   ~~~~~~~~~~~~~~~~~~~~
/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h:198:21: note: in instantiation of function template specialization 'oneapi::dpl::__par_backend_hetero::__serial_merge<oneapi::dpl::__ranges::zip_view<oneapi::dpl::__ranges::all_view<unsigned long, sycl::access::mode::read>, oneapi::dpl::__ranges::all_view<unsigned long, sycl::access::mode::read>>, oneapi::dpl::__ranges::all_view<oneapi::dpl::__internal::tuple<unsigned long, unsigned long>, sycl::access::mode::read>, const oneapi::dpl::__ranges::zip_view<oneapi::dpl::__ranges::all_view<unsigned long, sycl::access::mode::write>, oneapi::dpl::__ranges::all_view<unsigned long, sycl::access::mode::write>>, unsigned int, dpct::internal::compare_key_fun<>>' requested here
  198 |                     __serial_merge(__rng1, __rng2, __rng3, __start.first, __start.second, __i_elem, __chunk, __n1, __n2,
      |                     ^
[...]

PR #2022, opened 2025-01-27
Fix performance issue in `__serial_merge`

SergeyKopienko:
This is the second attempt of https://github.com/uxlfoundation/oneDPL/pull/2013: the source compile error has been introduced in the PR https://github.com/uxlfoundation/oneDPL/pull/1970
Example of the source compile error: https://godbolt.org/z/x6z7GMv1W

In the PR https://github.com/uxlfoundation/oneDPL/pull/2013 I detected performance issue.

The goal of this PR - to use ternary operator when it's possible.

Checks on GodBold: 
1) https://godbolt.org/z/6zPcWEMsn - implementation with `__assing_impl`
2) https://godbolt.org/z/fdsPdoPz3 - implementation without `__assing_impl`

Thanks to @MikeDvorskiy and @dmitriy-sobolev for their implementations of `__can_use_ternary_op_v` !

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -146,6 +146,7 @@ __can_use_ternary_op(...) -> std::false_type
     return {};
 }
 
+// This implementation of __assing_impl is required for performance optimisation
 template <typename _Rng1, typename _Rng2, typename _Rng3, typename _Index>
 std::enable_if_t<__can_use_ternary_op<_Rng1, _Rng2>().value, void>
 __assing_impl(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index& __rng1_idx, _Index& __rng2_idx,
@@ -154,6 +155,7 @@ __assing_impl(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index& _
     __rng3[__rng3_idx] = __use_rng2_val ? __rng2[__rng2_idx++] : __rng1[__rng1_idx++];
 }
 
+// TODO required to understand why the usual if-else is slower then ternary operator
 template <typename _Rng1, typename _Rng2, typename _Rng3, typename _Index>
 std::enable_if_t<!__can_use_ternary_op<_Rng1, _Rng2>().value, void>
 __assing_impl(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index& __rng1_idx, _Index& __rng2_idx,
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -141,13 +141,13 @@ __can_use_ternary_op(int) -> decltype(true ? std::declval<_value_t_rng1>() : std
 
 template <typename _Rng1, typename _Rng2>
 constexpr auto
-__can_use_ternary_op(...) -> std::false_type
+__can_use_ternary_op(long) -> std::false_type
 {
     return {};
 }
 
 template <typename _Rng1, typename _Rng2>
-constexpr static bool __can_use_ternary_op_v = decltype(__can_use_ternary_op<_Rng1, _Rng2>(0))::value;
+constexpr static bool __can_use_ternary_op_v = decltype(__can_use_ternary_op<_Rng1, _Rng2>(int{0}))::value;
 
 // Do serial merge of the data from rng1 (starting from start1) and rng2 (starting from start2) and writing
 // to rng3 (starting from start3) in 'chunk' steps, but do not exceed the total size of the sequences (n1 and n2)
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -130,19 +130,28 @@ __find_start_point(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_
     return _split_point_t<_Index>{*__res, __index_sum - *__res + 1};
 }
 
-template <typename _Rng1, typename _Rng2,
+template <typename _Rng1, typename _Rng2, typename _Rng3,
           typename _value_t_rng1 = oneapi::dpl::__internal::__value_t<_Rng1>,
-          typename _value_t_rng2 = oneapi::dpl::__internal::__value_t<_Rng2>>
-using __may_use_ternary_op = 
-    typename std::disjunction<
-        typename std::is_same<_value_t_rng1, _value_t_rng2>::type,
-        typename std::conjunction<
-            typename std::is_arithmetic<_value_t_rng1>::type,
-            typename std::is_arithmetic<_value_t_rng2>::type>::type
-    >::type;
+          typename _value_t_rng2 = oneapi::dpl::__internal::__value_t<_Rng2>,
+          typename _value_t_rng3 = oneapi::dpl::__internal::__value_t<_Rng3>>
+constexpr auto
+__can_use_ternary_op(int)
+    -> decltype(std::declval<std::reference_wrapper<_value_t_rng3>>() ? std::declval<_value_t_rng1>()
+                                                                      : std::declval<_value_t_rng2>(),
+                std::true_type{})
+{
+    return {};
+}
+
+template <typename _T1, typename _T2>
+constexpr auto
+__can_use_ternary_op(...) -> std::false_type
+{
+    return {};
+}
 
 template <typename _Rng1, typename _Rng2, typename _Rng3, typename _Index>
-std::enable_if_t<__may_use_ternary_op<_Rng1, _Rng2>::value, void>
+std::enable_if_t<__can_use_ternary_op<_Rng1, _Rng2>().value, void>
 __assing_impl(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index& __rng1_idx, _Index& __rng2_idx,
               const _Index __rng3_idx, const bool __use_rng2_val)
 {
@@ -150,7 +159,7 @@ __assing_impl(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index& _
 }
 
 template <typename _Rng1, typename _Rng2, typename _Rng3, typename _Index>
-std::enable_if_t<!__may_use_ternary_op<_Rng1, _Rng2>::value, void>
+std::enable_if_t<!__can_use_ternary_op<_Rng1, _Rng2>().value, void>
 __assing_impl(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, _Index& __rng1_idx, _Index& __rng2_idx,
               const _Index __rng3_idx, const bool __use_rng2_val)
 {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -132,12 +132,17 @@ __find_start_point(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_
 }
 
 template <typename _Rng1, typename _Rng2, typename = void>
-struct __can_use_ternary_op : std::false_type {};
+struct __can_use_ternary_op : std::false_type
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -156,11 +172,22 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, const _I
         // One of __rng1_idx_less_n1 and __rng2_idx_less_n2 should be true here
         // because 1) we should fill output data with elements from one of the input ranges
         // 2) we calculate __rng3_idx_end as std::min<_Index>(__rng1_size + __rng2_size, __chunk).
-        if (__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx]) ||
-            !__rng1_idx_less_n1)
-            __rng3[__rng3_idx] = __rng2[__rng2_idx++];
+        if constexpr (__can_use_ternary_op<_Rng1, _Rng2>().value)
```
akukanov, line 175: Please add a comment that the use of ternary operator is an optimization, and perhaps a TODO to understand why the usual if-else is slower.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -130,6 +130,43 @@ __find_start_point(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_
...
+template <typename _T1, typename _T2>
+constexpr auto
+__can_use_ternary_op(...) -> std::false_type
+{
+    return {};
+}
+
+// This implementation of __assing_impl is required for performance optimisation
+template <typename _Rng1, typename _Rng2, typename _Rng3, typename _Index>
+std::enable_if_t<__can_use_ternary_op<_Rng1, _Rng2>().value, void>
```
danhoeflinger, line 151: I think to use this SFINAE trick, you need to add something which could be interpreted as an integer as the argument. 

```suggestion
std::enable_if_t<__can_use_ternary_op<_Rng1, _Rng2>(0).value, void>
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -130,6 +130,22 @@ __find_start_point(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_
...
+    return {};
+}
+
+template <typename _T1, typename _T2>
+constexpr auto
+__can_use_ternary_op(...) -> std::false_type
+{
+    return {};
+}
+
```
danhoeflinger, line 148: I suggest adding the following, then using it, to hide some more of the ugly details.

```
template <typename T1, typename T2>
constexpr static bool __can_use_ternary_op_v = decltype(__can_use_ternary_op<T1,T2>(0))::value;
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -156,11 +172,24 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, const _I
         // One of __rng1_idx_less_n1 and __rng2_idx_less_n2 should be true here
         // because 1) we should fill output data with elements from one of the input ranges
         // 2) we calculate __rng3_idx_end as std::min<_Index>(__rng1_size + __rng2_size, __chunk).
-        if (__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx]) ||
-            !__rng1_idx_less_n1)
-            __rng3[__rng3_idx] = __rng2[__rng2_idx++];
+        if constexpr (__can_use_ternary_op<_Rng1, _Rng2>(0).value)
+        {
+            // This implementation is required for performance optimisation
```
danhoeflinger, line 177:
```suggestion
            // This implementation is required for performance optimization
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -130,6 +130,22 @@ __find_start_point(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_
...
+constexpr auto
+__can_use_ternary_op(int) -> decltype(true ? std::declval<_value_t_rng1>() : std::declval<_value_t_rng2>(),
+                                      std::true_type{})
+{
+    return {};
+}
+
+template <typename _T1, typename _T2>
+constexpr auto
+__can_use_ternary_op(...) -> std::false_type
```
MikeDvorskiy, line 144: Whether should be long instead of `...` because "error : SYCL kernel cannot call a variadic function"?

SergeyKopienko: It was on caller code side because I wrote `__can_use_ternary_op<_Rng1, _Rng2>()` without any arguments

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: So what you propose for now&
To rewrite as
```C++
template <typename _Rng1, typename _Rng2>
constexpr static bool __can_use_ternary_op_v = decltype(__can_use_ternary_op<_Rng1, _Rng2>(int{0}))::value;
```
?

danhoeflinger: Yes I think so, and also use `long` instead of `...`.

SergeyKopienko: Done.

danhoeflinger (approved):
I think this looks good, but should have a second set of eyes on it.  

Also, we should confirm that the reported bug is still fixed, and the performance regression disappears for the normal case.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -130,6 +130,25 @@ __find_start_point(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_
...
+
+template <typename _Rng1, typename _Rng2>
+constexpr auto
+__can_use_ternary_op(long) -> std::false_type
+{
+    return {};
+}
+
+template <typename _Rng1, typename _Rng2>
+constexpr static bool __can_use_ternary_op_v = decltype(__can_use_ternary_op<_Rng1, _Rng2>(int{0}))::value;
```
MikeDvorskiy, line 150: Why `int{0}` instead of `0`?

danhoeflinger: *I think* normally with this pattern you don't need to worry about what type you pass in because the `...` overload will be a lower priority than all other options, but with the variadic function limitation we make the "default" another typed parameter instead.  If we were to pass in a `long` then it just will take the "default" even if the `int` overload is available, so we have to make sure that the optionally enabled overload has priority.

Thinking more about this, what about systems where `int` is defined the same as `long`?  Should we use more explicitly different types here?

danhoeflinger: I agree `int{0}` is the same as `0`, but the `int{0}` provides more intention in the code.

MikeDvorskiy: I think "int" is more preferable in overloads choosing context. We have many places in the code where we rely on that.
For example, https://github.com/uxlfoundation/oneDPL/blob/main/include/oneapi/dpl/pstl/utils_ranges.h#L47

danhoeflinger: Perhaps we should use `std::int32_t` and std::int16_t`, and make sure our argument aligns with the type in the "true" overload, just to guarantee they are distinct types.

MikeDvorskiy: > but the `int{0}` provides more intention in the code.

which "intention" do we have?

SergeyKopienko: I hope the same like you wrote here:
```C++
    //SFINAE iterator type checks
    template <typename It>
    static constexpr auto
    __is_addressable(int) -> decltype(std::addressof(*::std::declval<It&>()), std::true_type{});
    template <typename It>
    static constexpr std::false_type
    __is_addressable(...);
    template <typename It>
    static constexpr bool __is_addressable_v = decltype(__is_addressable<It>(0))::value;
```

danhoeflinger: > which "intention" do we have?

Exactly this:
> make sure our argument aligns with the type in the "true" overload

If the type aligns with the default, it will just use the default no matter the validity of the SFINAE check in the "true" overload.  If the type doesn't align with either, then I think it is ambiguous, and will end up as an error.

SergeyKopienko: You have also passed `0` as argument.

SergeyKopienko: Let's finalize: my proposal - ask colleagues to check their compile error in our code after merge.
So right now - merge as is.

danhoeflinger: https://godbolt.org/z/1ae3s3T8o

Hopefully this shows why it changes when we have `long` as the "default" overload as compared to `...`.

Switch between the alternatives there.  It makes it more important what type we pass in.  As I understand it `int`, `long`, etc.  can be defined by the implementation and may sometimes be exactly the same type.  
There is nothing magic about `int` and `long` here, so I believe we should instead pick more explicitly defined and distinct types.

SergeyKopienko: One more variant from @dmitriy-sobolev : https://godbolt.org/z/1GP5Y68cM

danhoeflinger (approved):
LGTM, wait for green CI

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: The PR https://github.com/uxlfoundation/oneDPL/pull/2024 has been merged into this branch.

SergeyKopienko: @MikeDvorskiy is this conversation resolved now?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -130,6 +131,22 @@ __find_start_point(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_
...
+template <typename _Rng1DataType, typename _Rng2DataType>
+struct __can_use_ternary_op<_Rng1DataType, _Rng2DataType,
+                            std::void_t<decltype(true ? std::declval<_Rng1DataType>()
+                                                      : std::declval<_Rng2DataType>())>>
+    : std::true_type
+{
+};
+
+template <typename _Rng1, typename _Rng2>
+constexpr static bool __can_use_ternary_op_v = __can_use_ternary_op<_Rng1, _Rng2>::value;
```
danhoeflinger, line 148: nitpick:  
Naming of `_Rng[1/2]`.  Honestly I wouldn't mind all these just being `_T1`, `_T2`.  They don't need to care about how they are called.

danhoeflinger (approved):
LGTM :)  third time is a charm

danhoeflinger (approved):

MikeDvorskiy (approved):
Looks reasonable.
But before merging into the main I suggest being sure that CI and a case with tuple type issue (which was before) are passed.

PR #2029, opened 2025-01-30
Fix error `aspect fp64 is not supported` in tests of `std::complex`

SergeyKopienko:
In this PR we fix the error `Required aspect fp64 is not supported on the device` in some tests of `std::complex`.

### DIFF
--- a/test/support/test_complex.h
+++ b/test/support/test_complex.h
@@ -25,6 +25,7 @@
 
 #include <type_traits>
 #include <cassert>
+#include <limits>
 
 #if !_PSTL_MSVC_LESS_THAN_CPP20_COMPLEX_CONSTEXPR_BROKEN
 #    define STD_COMPLEX_TESTS_STATIC_ASSERT(arg) static_assert(arg)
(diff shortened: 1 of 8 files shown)


dmitriy-sobolev (dismissed):
LGTM

danhoeflinger:
As discussed offline, I'm not sure how this should help this error.  The numeric literals should all be integers, which shouldn't cause problems being confused for doubles.  We already seem to protect against testing cases with doubles with our macros, so its only floats that we cast to.  These should already be implicitly converting due to constexpr constructor of `std::complex`, so I'd expect no change to the issue.

I see these changes as harmless, but I don't understand how they help either.

danhoeflinger:
So, I think your changes will avoid the problem.  
Here is the comment from the definition of `TestUtils::infinity_val`:
```
    // This constant was declared to avoid the substitution and evaluation of the macro INFINITY
    // inside Kernel code. Since the implementation of INFINITY macro is not regulated, there are
    // cases where temporary expressions of type double are used to calculate the value of INFINITY.
    // This lead to an error on devices where the double type is not supported.
```
However, it effectively just removes any testing of this for devices not supporting `double`.  Can we pick a different baseline check for these with floating point which doesn't use TestUtils::infinity_val`?

akukanov:
This is clearly not a proper fix. If the test is supposed is supposed to check the use of floats on a device that does not support doubles, obviously the test itself should not use doubles either.

SergeyKopienko:
> So, I think your new changes will avoid the problem. Here is the comment from the definition of `TestUtils::infinity_val`:
> 
> ```
>     // This constant was declared to avoid the substitution and evaluation of the macro INFINITY
>     // inside Kernel code. Since the implementation of INFINITY macro is not regulated, there are
>     // cases where temporary expressions of type double are used to calculate the value of INFINITY.
>     // This lead to an error on devices where the double type is not supported.
> ```
> 
> However, it effectively just removes any testing of this for devices not supporting `double`. Can we pick a different baseline check for these with floating point which doesn't use TestUtils::infinity_val`?

All other significant changes are packed into `test_edges()` which also required `double` type support.

PR #2030, opened 2025-01-30
[Docs] Explicit python 3.12 for sphinx and imghdr

danhoeflinger:
Python 3.13 deprecates `imghdr`, which is still used until Sphinx 6.2. https://github.com/python/cpython/issues/104818

We could upgrade Sphinx from 4.4->6.2 but it seems like we have limited Sphinx at 4.4 it there for some other reason.  For now, lets just limit python to a version (3.12) which works.

### DIFF
--- a/.github/workflows/ci-docs.yml
+++ b/.github/workflows/ci-docs.yml
@@ -33,7 +33,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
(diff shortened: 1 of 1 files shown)


dmitriy-sobolev (approved):
LGTM as a fast fix.

There are many libraries with fixed versions in `requirements.txt`: updating Sphinx version may result in a cascade of dependency issues.

PR #2031, opened 2025-01-30
Bugfix to decay Policy for `__result_and_scratch_storage`

danhoeflinger:
Always decay the policy before use in `__result_and_scratch_storage`.

This also reverts some of #1997 which was working around this bug in the usage of `__result_and_scratch_storage`.  
Also, there are some minor improvements to the template arguments of the reduce submitter to require the decayed policy used in `__result_and_scratch_storage` to match the policy used to submit.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1093,7 +1093,7 @@ struct __write_to_id_if_else
 template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _UnaryOperation, typename _InitType,
           typename _BinaryOperation, typename _Inclusive>
 auto
-__parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backend_tag, const _ExecutionPolicy& __exec,
+__parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _ExecutionPolicy&& __exec,
                           _Range1&& __in_rng, _Range2&& __out_rng, std::size_t __n, _UnaryOperation __unary_op,
                           _InitType __init, _BinaryOperation __binary_op, _Inclusive)
 {
@@ -1122,9 +1122,9 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
             std::size_t __single_group_upper_limit = __use_reduce_then_scan ? 2048 : 16384;
             if (__group_scan_fits_in_slm<_Type>(__exec.queue(), __n, __n_uniform, __single_group_upper_limit))
             {
-                return __parallel_transform_scan_single_group(__backend_tag, __exec, std::forward<_Range1>(__in_rng),
-                                                              std::forward<_Range2>(__out_rng), __n, __unary_op, __init,
-                                                              __binary_op, _Inclusive{});
+                return __parallel_transform_scan_single_group(
+                    __backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__in_rng),
+                    std::forward<_Range2>(__out_rng), __n, __unary_op, __init, __binary_op, _Inclusive{});
             }
         }
 #if _ONEDPL_COMPILE_KERNEL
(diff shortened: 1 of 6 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1137,10 +1137,10 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
             _GenInput __gen_transform{__unary_op};
             try
             {
-                return __parallel_transform_reduce_then_scan(__backend_tag, __exec, __in_rng, __out_rng,
-                                                             __gen_transform, __binary_op, __gen_transform,
-                                                             _ScanInputTransform{}, _WriteOp{}, __init, _Inclusive{},
-                                                             /*_IsUniquePattern=*/std::false_type{});
+                return __parallel_transform_reduce_then_scan(
+                    __backend_tag, std::forward<_ExecutionPolicy>(__exec), __in_rng, __out_rng, __gen_transform,
```
mmichel11, line 1141: We should not forward `__exec` here and in similar places. If ownership is transferred to the function and then an exception is thrown, `__exec` may no longer be valid when passed to the fallback call.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1543,17 +1546,18 @@ __parallel_set_op(oneapi::dpl::__internal::__device_backend_tag __backend_tag, c
     {
         try
         {
-            return __parallel_set_reduce_then_scan(__backend_tag, __exec, __rng1, __rng2, __result, __comp,
-                                                   __is_op_difference);
+            return __parallel_set_reduce_then_scan(__backend_tag, std::forward<_ExecutionPolicy>(__exec), __rng1,
```
SergeyKopienko, line 1549: The same

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2492,8 +2497,8 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, cons
             try
             {
                 auto __res = oneapi::dpl::__par_backend_hetero::__parallel_reduce_by_segment_reduce_then_scan(
-                    oneapi::dpl::__internal::__device_backend_tag{}, __exec, __keys, __values, __out_keys, __out_values,
-                    __binary_pred, __binary_op);
+                    oneapi::dpl::__internal::__device_backend_tag{}, std::forward<_ExecutionPolicy>(__exec), __keys,
```
SergeyKopienko, line 2500: The same

SergeyKopienko:
@danhoeflinger , you wrote in description of this PR:
"This also reverts some of https://github.com/uxlfoundation/oneDPL/pull/1997 which was working around this bug in the usage of __result_and_scratch_storage."

May you explain what was the bug inside `__result_and_scratch_storage` and how it has been fixed on `__result_and_scratch_storage` level? Looks like it hasn't been fixed...

danhoeflinger:
> @danhoeflinger , you wrote in description of this PR: "This also reverts some of #1997 which was working around this bug in the usage of __result_and_scratch_storage."
> 
> May you explain how this bug has been fixed on `__result_and_scratch_storage` level? Looks like it hasn't been fixed...

The bug is in the **usage** of `__result_and_scratch_storage`, not the type itself. That part is fixed with this PR, because we decay the policy type everywhere we are specifying the type of `__result_and_scratch_storage` (not during overloads where we are instead deducing the type).  We can still think about if there is a better way to "get" the type of `__result_and_scratch_storage`, without needing to always think about decaying of a policy.

Perhaps we can add a using statement to define `__result_and_scratch_storage_t` which automatically decays the policy type...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1369,8 +1371,8 @@ __parallel_partition_copy(oneapi::dpl::__internal::__device_backend_tag __backen
             oneapi::dpl::__par_backend_hetero::__write_to_id_if_else<oneapi::dpl::__internal::__pstl_assign>;
         try
         {
-            return __parallel_reduce_then_scan_copy(__backend_tag, __exec, __rng, __result, __n, _GenMask{__pred},
-                                                    _WriteOp{},
+            return __parallel_reduce_then_scan_copy(__backend_tag, std::forward<_ExecutionPolicy>(__exec), __rng,
```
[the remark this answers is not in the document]
danhoeflinger: fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed.

danhoeflinger:
@mmichel11 @SergeyKopienko  please look at #2033, do we like that additional change?

It makes usage easier, but obfuscates what is going on.

mmichel11 (approved):
LGTM

PR #2033, opened 2025-01-31
Improve usage Pattern for result_and_scratch_storage

danhoeflinger:
This adds a using statement to define `__result_and_scratch_storage`   which does the decay of the execution policy for the caller, and makes the actual type `__result_and_scratch_storage_impl`.

I'm not sure I like this, because it "hides" the decay of the policy.  It certainly makes usage easier, but also obfuscates what is really happening here. 

What do you guys think?

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -315,8 +315,8 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_wg);
         // Storage for the results of scan for each workgroup
 
-        using __result_scratch_t = __result_and_scratch_storage_t<_ExecutionPolicy, _Type>;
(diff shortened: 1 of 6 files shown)


mmichel11:
There is some limited precedent in our enable ifs I see: https://github.com/uxlfoundation/oneDPL/blob/528c0f0db52fa46c507cd9b8d8ba7359460ed261/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h#L379-L381

However, I think I agree with keeping the decay. Since `std::is_execution_policy` requires it, I think designing all of our utilities similarly makes sense as it keeps a consistent behavior.

danhoeflinger:
@SergeyKopienko also pointed me to `__par_backend::__buffer` which has a similar pattern with decaying the policy as compared to the one proposed here.

I think I like this change.  The only instance where we need to reach into the `_impl` type is in the definition of `__future`, and I've added a comment to describe why.

danhoeflinger:
Alright, applying this into the other PR.  I think its the better option because it prevents future bugs where users fail to decay policies.  
We have an example of how to overcome type deduction issues in `__future` if they occur.  (Or in reduce, where it just works because we have the policy elsewhere in the signature)

PR #2034, opened 2025-02-01
Bugfix to avoid `__out` for MSVC

danhoeflinger:
__out is a macro in Microsoft C++ SAL (Source Annotation Language).
We cant use it as a variable name.  Causes compiler errors for microsoft platforms.


mmichel11 (approved):
LGTM

SergeyKopienko (commented):
LGTM

PR #2037, opened 2025-02-03
[oneDPL][rfc][ranges] proposal for implementation of the second part of range based API for oneDPL

MikeDvorskiy:
[oneDPL][rfc][ranges] proposal for implementation of the second part of range based API for oneDPL


rfcs/proposed/ranges_api_m/README.md
akukanov, line 1: Please rename to `range_algorithms/README.md`

rfcs/proposed/ranges_api_m/README.md
```diff
@@ -0,0 +1,28 @@
+# Support the second portion of the oneDPL Range APIs
+
+## Introduction
+Based on statistics (observing C++ code within github.com) for the usage of popular algorithms, the following
+range-based APIs are supposed to be implemented in oneDPL.
```
akukanov, line 5:
```suggestion
range-based APIs are suggested to be implemented next in oneDPL.
```

rfcs/proposed/ranges_api_m/README.md
```diff
@@ -0,0 +1,28 @@
...
+`fill`, `generate`, `move`, `replace`, `replace_if`, `remove`, `remove_if`, `mismatch`, `minmax_element`, `minmax`,
+`min`, `max`, `find_first_of`, `find_end`, `is_sorted_until`
+
+## Motivations
+The feature is proposed as the next step of range-based API support for oneDPL.
+
+### Key Requirements
+- The range-based signatures for the mentioned API should correspond to the proposal for C++ parallel range algorithms, P3179.
+(https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3179r4.html)
```
akukanov, line 14:
```suggestion
(https://wg21.link/p3179)
```

rfcs/proposed/ranges_api_m/README.md
MikeDvorskiy: done

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,28 @@
...
+range-based APIs are suggested to be implemented next in oneDPL.
+`fill`, `generate`, `move`, `replace`, `replace_if`, `remove`, `remove_if`, `mismatch`, `minmax_element`, `minmax`,
+`min`, `max`, `find_first_of`, `find_end`, `is_sorted_until`
+
+## Motivations
+The feature is proposed as the next step of range-based API support for oneDPL.
+
+### Key Requirements
+- The range-based signatures for the mentioned API should correspond to the proposal for C++ parallel range algorithms, P3179.
+(https://wg21.link/p3179)
```
akukanov, line 14: As an exception to this rule, we might have `mismatch`  to follow `equal` and `transform` in the oneDPL specification and allow only one of the two ranges to be sized.

I am still thinking of the best way to handle this.

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,28 @@
...
+- The proposed implementation should support all oneDPL execution policies: `seq`, `unseq`, `par`, `par_unseq` and `a device policy`.
```
akukanov, line 15:
```suggestion
- The proposed implementation should support all oneDPL execution policies: `seq`, `unseq`, `par`, `par_unseq`, and `device_policy`.
```

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+### Implementation proposal
+- The implementation is supposed to rely on existing range-based or iterator-based algorithm patterns, which are already
+implemented in oneDPL.
+- Several algorithms described in P3179 have slightly different semantics. To implement these, some existing algorithm patterns
+might require modifications or new versions.
+
+### Test coverage
+
+- It should be called with both small and large data sizes and with all the policies mentioned above.
```
akukanov, line 25: Does it make sense to also mention what kinds of input and output ranges will be tested?

MikeDvorskiy: Yes, it does.
I think we need to write down it here explicitly...

rfcs/proposed/range_algorithms/README.md
akukanov: Addressed in the specification draft:
> - As a rule, both input and output ranges must be sized.
    - Exceptions are binary ``transform``, ``equal``, and ``mismatch``, where at least one of the input ranges must be sized, and if a range is not sized it is supposed to be infinite. [*Note*: An example of an infinite range is ``std::views::repeat`` with no bound. -- *end note*]

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,35 @@
...
+non device copyabiability reason. To support  `std::vector` with `USM` allocator `std::vector` should be wrapped into std::ranges::subrange.
+
+### Test coverage
+- The algorithms should be called with both small and large data sizes and with all the policies mentioned above.
+- Output data, return type, and value should be checked/compared with the reference result
+computed by the corresponding serial std::ranges algorithm or by a custom implemented serial version
+in case of different semantics.
+- The tests should also call the algorithms with following standard range adapters: `std::ranges::subrange`, `std::span`, `std::views::all`
+- The tests should also call the algorithms with non-trivial projections.
+- In case of a device policy and std::vector with USM allocator,
```
akukanov, line 34: The last item is incomplete.

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,35 @@
...
+### Implementation limitation
+- In case of a device policy and `std::vector` with `USM` allocator, `std::vector` cannot be passed into algorithm directly by
+non device copyabiability reason. To support  `std::vector` with `USM` allocator `std::vector` should be wrapped into std::ranges::subrange.
+
+### Test coverage
+- The algorithms should be called with both small and large data sizes and with all the policies mentioned above.
+- Output data, return type, and value should be checked/compared with the reference result
+computed by the corresponding serial std::ranges algorithm or by a custom implemented serial version
+in case of different semantics.
+- The tests should also call the algorithms with following standard range adapters: `std::ranges::subrange`, `std::span`, `std::views::all`
```
akukanov, line 32: Would it make sense to test some algorithms with view chains as well? As a reminder, `iota`, `transform`, `reverse` are documented as supported views.

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,35 @@
...
+range-based APIs are suggested to be implemented next in oneDPL.
+`fill`, `generate`, `move`, `replace`, `replace_if`, `remove`, `remove_if`, `mismatch`, `minmax_element`, `minmax`,
+`min`, `max`, `find_first_of`, `find_end`, `is_sorted_until`
+
+## Motivations
+The feature is proposed as the next step of range-based API support for oneDPL.
+
+### Key Requirements
+- The range-based signatures for the mentioned API should correspond to the proposal for C++ parallel range algorithms, P3179.
+(https://wg21.link/p3179), but the all input and output ranges should be sized and random access ranges.
```
akukanov, line 14: I think here we can now refer to the specification update PR https://github.com/uxlfoundation/oneAPI-spec/pull/614
```suggestion
- The range-based signatures for the mentioned API should correspond to the [proposed specification](https://github.com/uxlfoundation/oneAPI-spec/pull/614)
that is based on the [C++ standardization proposal P3179](https://wg21.link/p3179).
```

rfcs/proposed/range_algorithms/README.md
MikeDvorskiy: yes.. and  `take`, `drop` as well.

akukanov (approved):
Seems good enough to me.

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,36 @@
...
+### Key Requirements
+- The range-based signatures for the mentioned API should correspond to the [proposed specification](https://github.com/uxlfoundation/oneAPI-spec/pull/614)
+that is based on the [C++ standardization proposal P3179](https://wg21.link/p3179).
+- The proposed implementation should support all oneDPL execution policies: `seq`, `unseq`, `par`, `par_unseq`, and `device_policy`.
+
+### Implementation proposal
+- The implementation is supposed to rely on existing range-based or iterator-based algorithm patterns, which are already
+implemented in oneDPL.
+- Several algorithms described in P3179 have slightly different semantics. To implement these, some existing algorithm patterns
+might require modifications or new versions.
```
danhoeflinger, line 21: I don't think we need to go into the details of each, I think we should at least list the APIs with semantic differences which may need modifications

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Implementation limitation
+- In case of a device policy and `std::vector` with `USM` allocator, `std::vector` cannot be passed into algorithm directly by
+non device copyabiability reason. To support  `std::vector` with `USM` allocator `std::vector` should be wrapped into std::ranges::subrange.
```
danhoeflinger, line 25:
```suggestion
- In case of a `device_policy` and `std::vector` with `USM` allocator, `std::vector` cannot be passed into algorithm directly by because a `std::vector` is not [SYCL device_copyable](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable). To support  `std::vector` with `USM` allocator `std::vector` should be wrapped into `std::ranges::subrange`.
```

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Test coverage
+- The algorithms should be called with both small and large data sizes and with all the policies mentioned above.
+- Output data, return type, and value should be checked/compared with the reference result
+computed by the corresponding serial std::ranges algorithm or by a custom implemented serial version
```
danhoeflinger, line 30:
```suggestion
computed by the corresponding serial `std::ranges` algorithm or by a custom implemented serial version
```

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,36 @@
...
+in case of different semantics.
+- The tests should also call the algorithms with following standard range adapters: `std::ranges::subrange`, `std::span`, `std::views::all`,
+  `std::views::iota`, `std::views::transform`, `std::views::reverse`, `std::views::take`, `std::views::drop`
+- The tests should also call the algorithms with non-trivial projections.
+- In case of a device policy and std::vector with USM allocator, the algorithms accept the vector wrapped into `std::ranges::subrange` or `std::span`.
```
danhoeflinger, line 35:
```suggestion
- In case of a `device_policy` and `std::vector` with USM allocator, the algorithms accept the vector wrapped into `std::ranges::subrange` or `std::span`.
```

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,36 @@
...
+- The proposed implementation should support all oneDPL execution policies: `seq`, `unseq`, `par`, `par_unseq`, and `device_policy`.
+
+### Implementation proposal
+- The implementation is supposed to rely on existing range-based or iterator-based algorithm patterns, which are already
+implemented in oneDPL.
+- Several algorithms described in P3179 have slightly different semantics. To implement these, some existing algorithm patterns
+might require modifications or new versions.
+
+### Implementation limitation
+- In case of a `device_policy` and `std::vector` with `USM` allocator, `std::vector` cannot be passed into algorithm directly by because a `std::vector` is not [SYCL device_copyable](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable). To support  `std::vector` with `USM` allocator `std::vector` should be wrapped into `std::ranges::subrange`.
```
danhoeflinger, line 24:
```suggestion
- In case of a `device_policy` and `std::vector` with `USM` allocator, `std::vector` cannot be passed into algorithm directly because a `std::vector` is not [SYCL device_copyable](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable). To support  `std::vector` with `USM` allocator `std::vector` should be wrapped into `std::ranges::subrange`.
```

danhoeflinger (dismissed):
After minor change, generally looks fine to me.

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,36 @@
...
+- In case of a `device_policy` the projections pointer-to-member and pointer-to-function are not supported, for SYCL backend at least.
+
+### Test coverage
```
dmitriy-sobolev, line 27: I suggest testing the most representative cases, which may depend on a certain algorithm, to make sure that underlying patterns are called the right way, and to rule out critical issues. 

For example, `is_sorted_until` should be tested with partially sorted data, or `find_end` should find the last occurrence.
I and Mikhail have different opinions on that (e.g. in https://github.com/uxlfoundation/oneDPL/pull/2100#discussion_r2037981134), so I opinions of others would be very useful.

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,36 @@
...
+implemented in oneDPL.
+- Several algorithms described in P3179 have slightly different semantics. To implement these, some existing algorithm patterns
+might require modifications or new versions.
+
+### Implementation limitation
+- In case of a `device_policy` and `std::vector` with `USM` allocator, `std::vector` cannot be passed into algorithm directly because a `std::vector` is not [SYCL device_copyable](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable). To support  `std::vector` with `USM` allocator `std::vector` should be wrapped into `std::ranges::subrange`.
+- In case of a `device_policy` the projections pointer-to-member and pointer-to-function are not supported, for SYCL backend at least.
+
+### Test coverage
+- The algorithms should be called with both small and large data sizes and with all the policies mentioned above.
```
dmitriy-sobolev, line 28: In #2100, we agreed that we should also test trivial cases, such as 0 and 1 elements, since they are usually handled on a "dispatch" level, which is not shared between iterator and range-based algorithms.

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,38 @@
+# Support the second portion of the oneDPL Range APIs
+
+## Introduction
+Based on statistics (observing C++ code within github.com) for the usage of popular algorithms, the following
+range-based APIs are suggested to be implemented next in oneDPL.
+`fill`, `generate`, `move`, `replace`, `replace_if`, `remove`, `remove_if`, `mismatch`, `minmax_element`, `minmax`,
```
dmitriy-sobolev, line 6: Let's remove `generate` at this time. There is a concern that the interface may change in p3179 in further revisions.

rfcs/proposed/range_algorithms/README.md
MikeDvorskiy: probably, it is reasonable..

rfcs/proposed/range_algorithms/README.md
```diff
@@ -0,0 +1,36 @@
...
+implemented in oneDPL.
+- Several algorithms described in P3179 have slightly different semantics. To implement these, some existing algorithm patterns
+might require modifications or new versions.
+
+### Implementation limitation
+- In case of a `device_policy` and `std::vector` with `USM` allocator, `std::vector` cannot be passed into algorithm directly by because a `std::vector` is not [SYCL device_copyable](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable). To support  `std::vector` with `USM` allocator `std::vector` should be wrapped into `std::ranges::subrange`.
+- In case of a `device_policy` the projections pointer-to-member and pointer-to-function are not supported, for SYCL backend at least.
+
+### Test coverage
+- The algorithms should be called with both small and large data sizes and with all the policies mentioned above.
```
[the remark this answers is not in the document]
MikeDvorskiy: "small" an "big" sizes should enable differentiated implementation with level of parallelism perspective.

"small" means that such data sizes are processed by differentiated implementation for "small" data sizes. Usually it bounded by a work group size (for  example,  < 256). Other words. A small size sequence is processed by a kind of "one group algorithm in case  of a device policy. And a small size sequence is processed by a kind SIMD algorithm in kind of seq/unseq policy.

 "big" sizes means that such data sizes are processed by differentiated implementation for such sizes where parallelism is, or where parallelism is maximum. Tens of millions for device policy and  millions  for the host policies.

rfcs/proposed/range_algorithms/README.md
dmitriy-sobolev: @akukanov or @danhoeflinger, what is your opinion here? I've had an offline discussion with Mike, but we cannot agree. His position is that it is overtesting.

Here is a case from the implementation. The testing framework generates a monotonically increasing sequence (0, 1,... n-1) for `is_sorted_until`. We have different opinions whether it is enough or not.

dmitriy-sobolev: @akukanov and @danhoeflinger, what is your opinion? I and Mikhail cannot agree after an offline discussion.

Here is a more concrete case for the context. `is_sorted_until` from #2100 is tested with (0, 1,...n-1) sequence with less and greater predicates. I think it is not sufficient. Mikhail thinks that adding a partially sorted case is overtesting.

rfcs/proposed/range_algorithms/README.md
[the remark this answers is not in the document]
dmitriy-sobolev:
```suggestion   
- The algorithms should be called with both small and large sizes.
  Usually, the algorithms have a general implementation, and a specialized implementation for small sizes.
  The definition of small and large depends on a certain algorithm and a policy it is used with.
  Small can be 256 elements or less.
  The magnitude of large can be 10^3 elements for `seq` and `unseq`, 10^6 for `par` and `par_unseq`, and 10^7 for `device_policy` to maximize parallelism.

File: rfcs/proposed/range_algorithms/README.md (hunk 14 above)
Comment (danhoeflinger) [reply rfcs/proposed/range_algorithms/README.md:27]: I think there are multiple questions here:  
1) Can we improve our test coverage with some targeted tests for common use cases?  
  My answer: Absolutely, yes.
2) Is it the responsibility of the ranges API feature PRs to provide these additional tests?
  My answer: Probably not.

As far as I understand, these are not new algorithms, but rather new interfaces with slightly different semantics in some cases.  

My suggestion is that we allow ranges testing to be somewhat lighter weight, to check the interface in situations where the semantics have not changed from the iterator version of the same algorithm.  Separately, if there are places with the iterator API testing is insufficient, lets improve that but with the right priority and not added on to an already large effort to add ranges API.  

However, in cases where we have different semantics for the new ranges API (size limited output ranges, additional return values, etc.), it *should* be the responsibility of the ranges PRs to add testing for those systems, as they are not present in the iterator API.  We should have testing to confirm that output range limited calls stop at the appropriate place, and that all return values are appropriate.
---
TLDR:  My suggestion would be to make the iterator tests fully test the iterator API and semantics (adding PRs where to fill gaps we notice), and to have the ranges API tests verify the aspects which are unique to the ranges API.

Comment (akukanov) [reply rfcs/proposed/range_algorithms/README.md:27]: I tend to agree with Dan's opinion that the tests for range algorithms should primarily focus on the specifics of those. It's not merely the API testing though, because in many cases there is enough of semantical differences as well - such as handling the limited output, returning proper iterators to all the sequences, using projections, etc.

However, this concerns me as well:
> The testing framework generates a monotonically increasing sequence (0, 1,... n-1) for `is_sorted_until` called with less and greater predicates.

I do not think this is enough. And improving this test it's not really hard - just swap elements in two random positions in the sorted sequence, and check that the algorithm returns the "smallest" of these positions.

File: rfcs/proposed/range_algorithms/README.md (hunk 17 above)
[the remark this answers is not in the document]
Comment (akukanov) [reply rfcs/proposed/range_algorithms/README.md:33]: I would instead say that testing should be done for a variety of the range sizes, from several hundreds to several millions.

For me, the main motivation of this is to make sure that the test is really executed in parallel. A typical test that operates with a dozen of elements is often sufficient to check the semantics and even the corner cases, but it does not create enough parallel slack and essentially will execute in a single thread. Ensuring that we tested the parallel execution is specifically important for us as the authors of P3179 to claim that there is reasonable implementation experience.

File: rfcs/proposed/range_algorithms/README.md (hunk 14 above)
Comment (MikeDvorskiy) [reply rfcs/proposed/range_algorithms/README.md:27]: > I think there are
Thank you Dan for expressing your thoughts.  It is absolutely matching with my thoughts.

Comment (MikeDvorskiy) [reply rfcs/proposed/range_algorithms/README.md:27]: > The testing framework generates a monotonically increasing sequence (0, 1,... n-1) for `is_sorted_until` called with less and greater predicates

But it is question of a quality of implementation and should be addressed in a corresponding implementation PR, not RFC... Telling about API, the corner cases and cases when the range based impl has different semantic - of course this RFC should mention about that.

**Review by dmitriy-sobolev — COMMENTED** [review]
I am ready to approve once the suggestion below has been applied.

File: rfcs/proposed/range_algorithms/README.md [hunk 18]
```diff
@@ -0,0 +1,45 @@
...
+- If a range-based algorithm shares its implementation with an iterator-based variant that is sufficiently tested,
+  the range of input sizes, element types, and data distributions tested for the range-based version can be reduced.
+  However, if the range-based algorithm performs any additional processing,
+  such as handling trivial cases before delegating to the shared implementation,
+  those scenarios must be tested explicitly.
+- The algorithms should be called with both small and large sizes.
+  Usually, the algorithms have a general implementation, and a specialized implementation for small sizes.
+  The definition of small and large depends on a certain algorithm and a policy it is used with.
+  Small can be 256 elements or less.
+  The magnitude of large can be 10^3 elements for `seq` and `unseq`, 10^6 for `par` and `par_unseq`, and 10^7 for `device_policy` to maximize parallelism.
```
dmitriy-sobolev, line 37: Less details as suggested in https://github.com/uxlfoundation/oneDPL/pull/2037#discussion_r2068670479:

```suggestion
- Testing should cover a range of input sizes, from hundreds to millions of elements, to ensure parallel execution (where applicable, based on the execution policy). Smaller sizes may be used to verify semantics.
```

rfcs/proposed/range_algorithms/README.md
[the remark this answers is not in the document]
dmitriy-sobolev: Done. See https://github.com/uxlfoundation/oneDPL/pull/2037#discussion_r2069356745.

rfcs/proposed/range_algorithms/README.md
dmitriy-sobolev: Thanks for the input. My initial suggestion is excessive for RFC format.

dmitriy-sobolev (approved):
LGTM.

https://github.com/uxlfoundation/oneDPL/pull/2037/files#r2035529923 suggestion is optional in my opinion.

PR #2038, opened 2025-02-03
Bugfix: define NOMINMAX in a limited way around TBB less than 2021.12

danhoeflinger:
In 2021.12, TBB fixed an issue with `tbb/enumerable_thread_specific.h` which includes `windows.h` by wrapping it in a `NOMINMAX` define.  
https://github.com/uxlfoundation/oneTBB/pull/1291

To work with versions of TBB prior to 2021.12, we need to do the same prior to including `tbb/enumerable_thread_specific.h` on windows.

### DIFF
--- a/include/oneapi/dpl/pstl/parallel_backend_tbb.h
+++ b/include/oneapi/dpl/pstl/parallel_backend_tbb.h
@@ -48,14 +48,14 @@
 #include <tbb/tbb_allocator.h>
 #include <tbb/enumerable_thread_specific.h>
 
+#if TBB_INTERFACE_VERSION > 12000
+#    include <tbb/task.h>
+#endif
+
 #if __ONEDPL_DEFINED_NOMINMAX
 // Avoid affecting outside code with NOMINMAX
 #    undef NOMINMAX
 #    undef __ONEDPL_DEFINED_NOMINMAX
-#endif
-
-#if TBB_INTERFACE_VERSION > 12000
-#    include <tbb/task.h>
 #endif
 
 #if TBB_INTERFACE_VERSION < 10000
--- a/include/oneapi/dpl/pstl/parallel_backend_tbb.h
+++ b/include/oneapi/dpl/pstl/parallel_backend_tbb.h
@@ -36,23 +36,23 @@
 #include <tbb/tbb_allocator.h>
 
 #if _WIN32 || _WIN64
-    // For windows, TBB prior to version 2021.12 requires NOMINMAX to be defined before including
-    // tbb/enumerable_thread_specific.h, which includes windows.h. This is not required for TBB 2021.12 and later due
-    // to an TBB fix.
-    #if TBB_VERSION_MAJOR < 2021 || (TBB_VERSION_MAJOR == 2021 && TBB_VERSION_MINOR < 12)
-        #ifndef NOMINMAX
-            #define NOMINMAX
-            #define __ONEDPL_DEFINED_NOMINMAX 1
-        #endif
-    #endif
+// For windows, TBB prior to version 2021.12 requires NOMINMAX to be defined before including
+// tbb/enumerable_thread_specific.h, which includes windows.h. This is not required for TBB 2021.12 and later due
+// to an TBB fix.
+#    if TBB_VERSION_MAJOR < 2021 || (TBB_VERSION_MAJOR == 2021 && TBB_VERSION_MINOR < 12)
+#        ifndef NOMINMAX
+#            define NOMINMAX
+#            define __ONEDPL_DEFINED_NOMINMAX 1
+#        endif
+#    endif
 #endif
 
 #include <tbb/enumerable_thread_specific.h>
 
 #if __ONEDPL_DEFINED_NOMINMAX
-    // Avoid affecting outside code with NOMINMAX
-    #undef NOMINMAX
-    #undef __ONEDPL_DEFINED_NOMINMAX
+// Avoid affecting outside code with NOMINMAX
+#    undef NOMINMAX
+#    undef __ONEDPL_DEFINED_NOMINMAX
 #endif
 
 #if TBB_INTERFACE_VERSION > 12000
--- a/include/oneapi/dpl/pstl/parallel_backend_tbb.h
+++ b/include/oneapi/dpl/pstl/parallel_backend_tbb.h
@@ -25,6 +25,18 @@
 #include "parallel_backend_utils.h"
 #include "execution_impl.h"
 
+#if _WIN32 || _WIN64
+// For windows, TBB prior to version 2021.12 requires NOMINMAX to be defined before including
+// tbb/enumerable_thread_specific.h, which includes windows.h. This is not required for TBB 2021.12 and later due
+// to an TBB fix.
+#    if TBB_INTERFACE_VERSION < 12120
+#        ifndef NOMINMAX
+#            define NOMINMAX
+#            define __ONEDPL_DEFINED_NOMINMAX 1
+#        endif
+#    endif
+#endif
+
 // Bring in minimal required subset of Intel(R) Threading Building Blocks (Intel(R) TBB)
 #include <tbb/blocked_range.h>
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -34,7 +34,27 @@
 #include <tbb/parallel_invoke.h>
 #include <tbb/task_arena.h>
 #include <tbb/tbb_allocator.h>
+
+#if _WIN32 || _WIN64
```
SergeyKopienko, line 38: May be better to move this block before `tbb` includes?

mmichel11: In my opinion, the current approach is best since it targets just the problematic header and limits where we use this macro to only where it is needed.

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/parallel_backend_tbb.h
danhoeflinger: I can see the point about wrapping all tbb headers, as we dont know if another TBB header might include `tbb/enumerable_thread_specific.h`.  

However, I see the requirement for this define as undesirable, so I'd rather limit it in our code by version to allow TBB to handle it as they see fit in the future. 

I'm OK with whatever the consensus is here, though.

include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -35,6 +47,13 @@
...
+
+#if __ONEDPL_DEFINED_NOMINMAX
+// Avoid affecting outside code with NOMINMAX
+#    undef NOMINMAX
+#    undef __ONEDPL_DEFINED_NOMINMAX
+#endif
+
 #if TBB_INTERFACE_VERSION > 12000
 #    include <tbb/task.h>
 #endif
```
SergeyKopienko, line 59:
```suggestion

#if TBB_INTERFACE_VERSION > 12000
#    include <tbb/task.h>
#endif

#if __ONEDPL_DEFINED_NOMINMAX
// Avoid affecting outside code with NOMINMAX
#    undef NOMINMAX
#    undef __ONEDPL_DEFINED_NOMINMAX
#endif
```

danhoeflinger: done, thanks

SergeyKopienko: I mean better to restore macro state after `include <tbb/task.h>`

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -25,6 +25,18 @@
 #include "parallel_backend_utils.h"
 #include "execution_impl.h"
 
+#if _WIN32 || _WIN64
+// For windows, TBB prior to version 2021.12 requires NOMINMAX to be defined before including
+// tbb/enumerable_thread_specific.h, which includes windows.h. This is not required for TBB 2021.12 and later due
+// to an TBB fix.
```
akukanov, line 31:
```suggestion
// Prior to version 2021.12, TBB requires NOMINMAX to be defined before including
// tbb/enumerable_thread_specific.h, which includes windows.h.
// This is no more required since TBB 2021.12 due to a TBB fix.
```

include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -35,10 +47,17 @@
...
 #if TBB_INTERFACE_VERSION > 12000
 #    include <tbb/task.h>
 #endif
 
+#if __ONEDPL_DEFINED_NOMINMAX
+// Avoid affecting outside code with NOMINMAX
+#    undef NOMINMAX
+#    undef __ONEDPL_DEFINED_NOMINMAX
+#endif
+
```
akukanov, line 60: This is actually pretty useless, since `windows.h` has been included already.

akukanov: Maybe that's the reason why TBB did not have the "fix" for a while.

akukanov: I do not really know what we should do with that.

a) Ideally, I would just say that oneDPL requires `NOMINMAX` to be defined, if `windows.h` is included. But (a) it might be included indirectly, without users knowing of that, and (b) what's worse, now we just indirectly include it ourselves.

b) The other way I can see is to document that oneDPL always defines `NOMINMAX`, period. And change the implementation to just do it always on Windows, early enough for the TBB headers and `windows.h` there to see it. That, however, will not help if our headers are included too late.

c) We can keep it as a "workaround" for the known problematic case. This, however, looks rather insufficient, as it still requires to document both that `NOMINMAX` is required and that we can include `windows,h` while that macro is defined.

I think I am inclining to a combination of both (a) and (b) - documenting that this macro is required if `windows.h` can be included prior to oneDPL headers and that we define it in our headers.

akukanov: Or, we can drop the use `enumerable_thread_specific` and instead implement something similar as in the OpenMP backend.

danhoeflinger: > Or, we can drop the use `enumerable_thread_specific` and instead implement something similar as in the OpenMP backend.

I'd prefer not to have to avoid using their more robust version in TBB for something like this.

--- 

Can we just state that `NOMINMAX` is required on windows in documentation + release notes, and add it to our build rules in cmake for windows?

I think we need such a rule either way for user who might first include windows.h before any tbb or oneDPL headers, so why not let that be our fix entirely, and not add a workaround like this which is messy and unreliable.

akukanov: > Can we just state that NOMINMAX is required on windows in documentation + release notes, and add it to our build rules in cmake for windows?

Do you think adding the macro to oneDPL cmake rules is better than adding to the headers? I do not see it why. If someone integrates oneDPL into their code with our cmake files, they will get the macro defined for the whole translation unit, right? So we still need to document that oneDPL can define the macro. At the same time, if someone does not use our cmake files and just includes the headers, they could see a compile error which would be perceived as oneDPL problem (and rightfully so, I would say). So only defining the macro in the cmake rules is not sufficient, I am afraid.

Probably defining it in both the headers and the cmake files, and documenting clearly, would minimize the impact on users.

danhoeflinger: The "all of the above" approach works for me. (doc + cmake + header definition)

I just don't think we get that much benefit from adding it to the headers because of the following case:
``` 
#include<windows.h>
#include<oneapi/dpl/algorithm>
///... user code
```  

In this case, the macros have already been defined in the previous inclusion of "windows.h" before oneDPL has the chance to add `NOMINMAX`.  Also, can defining `NOMINMAX` result in some ODR violations if it is not project wide? I've seen some comments mentioning that but I'm not sure.

If someone uses our headers without cmake, then they need to adhere to documented build rules / requirement. 

---

There is also another option:
We can wrap any instances of `*::min` `*::max` in parens.  This is what MSVC STL does, because it does not want to define `NOMINMAX` for all users.

`(std::min)(a,b)`
`(std::numeric_limits<int>::max)()`

Its not something I want to do, but if the other options are more undesirable, then we should consider it.

danhoeflinger: https://github.com/boostorg/stacktrace/issues/37#issuecomment-347997174
discussion of essentially the same topic in boost.

danhoeflinger: I've added #2042 as a draft for discussion.

danhoeflinger:

PR #2043, opened 2025-02-04
[Draft] Protecting min and max usages from macro collisions

danhoeflinger:
Another option for resolving NOMINMAX issue.  
Similar to MSVC STL, we can protect usages of min and max against colliding with the macros defined in windows.h. 

The upside for this is that there is no need for /DNOMINMAX with this option.  The downside is cluttering the code, and maintenance in the future where we must continue to protect such instances similarly.

Perhaps we can include windows.h in our tests from windows to define these macros and detect these errors in our per commit CI going forward.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
+++ b/include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_kernels.h
@@ -44,7 +44,7 @@ __one_wg_kernel(sycl::nd_item<1> __idx, ::std::uint32_t __n, _RngPack1&& __rng_p
 
     // max SLM is 256 * 4 * 64 + 256 * 2 * 64 + 257*2, 97KB, when  __data_per_work_item = 256, __bin_count = 256
     // to support 512 processing size, we can use all SLM as reorder buffer with cost of more barrier
-    __dpl_esimd::__ns::slm_init<::std::max(__reorder_slm_size, __bin_hist_slm_size + __incoming_offset_slm_size)>();
+    __dpl_esimd::__ns::slm_init<(std::max)(__reorder_slm_size, __bin_hist_slm_size + __incoming_offset_slm_size)>();
 
     const ::std::uint32_t __local_tid = __idx.get_local_linear_id();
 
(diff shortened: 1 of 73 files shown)


SergeyKopienko:
I think the usage of `(std::max)` in the a lot of places is too much. Let's create our internal implementations?

SergeyKopienko:
Are we able to change something in the [include/oneapi/dpl/pstl/ranges/nanorange.hpp](https://github.com/uxlfoundation/oneDPL/pull/2043/files#diff-4843812df860c30ae5cac28a81322cddd71da483498e02f621a3bbc5ad143412) ?

danhoeflinger:
> I think the usage of `(std::max)` in the a lot of places is too much. Let's create our internal implementations?

It is possible for us to do this for `std::max`, `std::min`, `std::numeric_limits<T>::min`, `std::numeric_limits<T>::max`.
It would require us to have our own `oneapi::dpl::__internal::__min`, `oneapi::dpl::__internal::__max`, `oneapi::dpl::__internal::__numeric_limits_min<T>`. `oneapi::dpl::__internal::__numeric_limits_max<T>` which is a lot more clumsy than extra parens in my opinion, but its is possible.
We cannot name them with simple `min(` or `max(` because those would be substituted with macros as well.

There is also the `min()` and `max()` functions we define in the RNG code which will still need an interrupting dummy macro like we use `_ONEDPL_PREVENT_MACRO_SUBSTITUTION` for now, and usages of other namespace `dpl::min` and `max` functions, which will still require paren wrapping or similar replacement.

danhoeflinger:
> Are we able to change something in the [include/oneapi/dpl/pstl/ranges/nanorange.hpp](https://github.com/uxlfoundation/oneDPL/pull/2043/files#diff-4843812df860c30ae5cac28a81322cddd71da483498e02f621a3bbc5ad143412) ?

We have done so in the past, but it is a good question.
https://github.com/uxlfoundation/oneDPL/commits/main/include/oneapi/dpl/pstl/ranges/nanorange.hpp

danhoeflinger:

PR #2045, opened 2025-02-04
[Draft] Undefine and redefine min and max macros

danhoeflinger:
Undefine and later redefine windows.h `min` and `max` macros in public headers and after TBB inclusion to protect against conflicts with `std` and other APIs.

Requires defining NOMINMAX in test_config.h (or some similar solution) to prevent similar conflict issues in test code.


danhoeflinger:

PR #2046, opened 2025-02-04
Implement alternative `-O0` workaround for reduce-then-scan with compiler optimization detection

mmichel11:
GCC defines the `__OPTIMIZE__` macro which evaluates to `1` if compiler optimizations are enabled, and is `0` otherwise. Similarly, clang does not seem to explicitly document it, but it exposes an `__OPTIMIZE__` macro with the same semantics as GCC's. This same macro is defined in the Intel LLVM based compiler. We can use this knowledge to implement the reduce-then-scan `-O0` workaround with no performance penalty by switching the required sub-group size to `16` when optimizations are not present.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1057,12 +1057,7 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
     // work-group implementation requires a fundamental type which must also be trivially copyable.
     if constexpr (std::is_trivially_copyable_v<_Type>)
     {
-        bool __use_reduce_then_scan =
-#if _ONEDPL_COMPILE_KERNEL
-            oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec);
-#else
-            false;
-#endif
+        bool __use_reduce_then_scan = oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec);
 
         // TODO: Consider re-implementing single group scan to support types without known identities. This could also
         // allow us to use single wg scan for the last block of reduce-then-scan if it is sufficiently small.
@@ -1081,7 +1076,6 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
                     std::forward<_Range2>(__out_rng), __n, __unary_op, __init, __binary_op, _Inclusive{});
             }
         }
-#if _ONEDPL_COMPILE_KERNEL
         if (__use_reduce_then_scan)
         {
             using _GenInput = oneapi::dpl::__par_backend_hetero::__gen_transform_input<_UnaryOperation>;
@@ -1089,19 +1083,12 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
             using _WriteOp = oneapi::dpl::__par_backend_hetero::__simple_write_to_id;
 
             _GenInput __gen_transform{__unary_op};
-            try
-            {
-                return __parallel_transform_reduce_then_scan(__backend_tag, __exec, __in_rng, __out_rng,
-                                                             __gen_transform, __binary_op, __gen_transform,
-                                                             _ScanInputTransform{}, _WriteOp{}, __init, _Inclusive{},
-                                                             /*_IsUniquePattern=*/std::false_type{});
-            }
-            catch (const sycl::exception& __e)
-            {
-                __bypass_sycl_kernel_not_supported(__e);
-            }
+
+            return __parallel_transform_reduce_then_scan(
+                __backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__in_rng),
+                std::forward<_Range2>(__out_rng), __gen_transform, __binary_op, __gen_transform, _ScanInputTransform{},
+                _WriteOp{}, __init, _Inclusive{}, /*_IsUniquePattern=*/std::false_type{});
         }
-#endif
     }
 
     //else use multi pass scan implementation
@@ -1176,7 +1163,6 @@ struct __invoke_single_group_copy_if
     }
 };
 
-#if _ONEDPL_COMPILE_KERNEL
 template <typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Size, typename _GenMask,
           typename _WriteOp, typename _IsUniquePattern>
 auto
@@ -1195,7 +1181,6 @@ __parallel_reduce_then_scan_copy(oneapi::dpl::__internal::__device_backend_tag _
         _ScanInputTransform{}, __write_op, oneapi::dpl::unseq_backend::__no_init_value<_Size>{},
         /*_Inclusive=*/std::true_type{}, __is_unique_pattern);
 }
-#endif
 
 template <typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Size, typename _CreateMaskOp,
           typename _CopyByMaskOp>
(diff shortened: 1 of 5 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -726,14 +726,25 @@ struct __parallel_reduce_then_scan_scan_submitter
...
-// reduce_then_scan requires subgroup size of 32, and performs well only on devices with fast coordinated subgroup
-// operations.  We do not want to run this scan on CPU targets, as they are not performant with this algorithm.
+// With optimization enabled, reduce-then-scan requires a sub-group size of 32. Without optimization, we must compile
+// to a sub-group size of 16 to workaround a hardware bug on certain Intel integrated graphics architectures.
+constexpr inline std::uint8_t __reduce_then_scan_sg_sz =
+#if _ONEDPL_DETECT_COMPILER_OPTIMIZATIONS_ENABLED
+    std::uint8_t{32};
+#else
+    std::uint8_t{16};
+#endif
```
danhoeflinger, line 736: Perhaps this is just my preference, but I think I would prefer this as a constexpr function we could call, rather than inline `std::uint8_t` just floating in the namespace.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -756,23 +767,15 @@ __parallel_transform_reduce_then_scan(oneapi::dpl::__internal::__device_backend_
...
-
-    constexpr std::uint8_t __sub_group_size = 32;
+    using _ReduceKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
+        __reduce_then_scan_reduce_kernel<_CustomName>>;
+    using _ScanKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
+        __reduce_then_scan_scan_kernel<_CustomName>>;
+    using _ValueType = typename _InitType::__value_type;
+
+    // To workaround hardware bugs on older integrated graphics, we need to use a sub-group size of 16 with -O0
+    // compilation.
```
danhoeflinger, line 778: If we do change `__reduce_then_scan_sg_sz` to a constexpr function, we could call it here and assign it to `constexpr __sub_group_size`, and leave `__sub_group_size` as the variable in the rest.

danhoeflinger (commented):
Other than the small nit picks, this largely looks good to me, assuming we are satisfied with the way we detect -O0.  I did not verify this independently.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: Sure, I made this change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1371,37 +1346,32 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag __backend_tag,
...
 
         return __par_backend_hetero::__static_monotonic_dispatcher<_SizeBreakpoints>::__dispatch(
-            _SingleGroupInvoker{}, __n, __exec, __n, std::forward<_InRng>(__in_rng), std::forward<_OutRng>(__out_rng),
-            __pred, __assign);
+            _SingleGroupInvoker{}, __n, std::forward<_ExecutionPolicy>(__exec), __n, std::forward<_InRng>(__in_rng),
+            std::forward<_OutRng>(__out_rng), __pred, __assign);
     }
-#if _ONEDPL_COMPILE_KERNEL
-    else if (oneapi::dpl::__par_backend_hetero::__is_gpu_with_sg_32(__exec))
+    else if (oneapi::dpl::__par_backend_hetero::__is_gpu_with_reduce_then_scan_sg_sz(__exec))
```
SergeyKopienko, line 1352:
```suggestion
    if (oneapi::dpl::__par_backend_hetero::__is_gpu_with_reduce_then_scan_sg_sz(__exec))
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1495,23 +1464,18 @@ __parallel_set_op(oneapi::dpl::__internal::__device_backend_tag __backend_tag, _
...
+        return __parallel_set_reduce_then_scan(__backend_tag, std::forward<_ExecutionPolicy>(__exec),
+                                               std::forward<_Range1>(__rng1), std::forward<_Range2>(__rng2),
+                                               std::forward<_Range3>(__result), __comp, __is_op_difference);
+    }
+    else
+    {
+        return __parallel_set_scan(__backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__rng1),
+                                   std::forward<_Range2>(__rng2), std::forward<_Range3>(__result), __comp,
+                                   __is_op_difference);
     }
```
SergeyKopienko, line 1478:
```suggestion

        return __parallel_set_scan(__backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__rng1),
                                   std::forward<_Range2>(__rng2), std::forward<_Range3>(__result), __comp,
                                   __is_op_difference);
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1427,7 +1397,7 @@ __parallel_set_reduce_then_scan(oneapi::dpl::__internal::__device_backend_tag __
 
     return __parallel_transform_reduce_then_scan(
         __backend_tag, std::forward<_ExecutionPolicy>(__exec),
-        oneapi::dpl::__ranges::zip_view(
+        oneapi::dpl::__ranges::make_zip_view(
```
SergeyKopienko, line 1400: Does this change really needed?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -726,14 +726,29 @@ struct __parallel_reduce_then_scan_scan_submitter
...
+constexpr inline std::uint8_t
+__get_reduce_then_scan_sg_sz()
+{
+    return
+#if _ONEDPL_DETECT_COMPILER_OPTIMIZATIONS_ENABLED
+        std::uint8_t{32};
+#else
+        std::uint8_t{16};
+#endif
+}
```
SergeyKopienko, line 740:
```suggestion
constexpr inline std::uint8_t
__get_reduce_then_scan_sg_sz()
{
#if _ONEDPL_DETECT_COMPILER_OPTIMIZATIONS_ENABLED
        return 32;
#else
        return 16;
#endif
}
```

SergeyKopienko:
One note about this approach.
At whole this mean that in debug (usually without optimization) and in release (usually with optimization) our code has different behavior. Potentially it may bring to the users some difficult for debugging cases.
May be also some documented `broken macro` needed also?
To inform users about this broken case and ability to make through this macro make the same behavior (optionally).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I had this same question to Matt offline.  His response:

"In the first PR I went from make_zip_view -> zip_view: https://github.com/uxlfoundation/oneDPL/pull/1997#discussion_r1933086592, so the revert undid this."

It goes back to some issue with `make_zip_view` being able to process creation of a zip_view of zip_views I believe.  This returns a change that was necessary in the previous workaround PR. We still have that issue mentioned #1805, but this restores us to before the first workaround commit.

akukanov:
> May be also some documented `broken macro` needed also? To inform users about this broken case and ability to make through this macro make the same behavior (optionally).

Is not `_ONEDPL_DETECT_COMPILER_OPTIMIZATIONS_ENABLED` already a macro that indicates this difference of "debug" vs "release" in this case? If someone wants, they can change its definition when debugging.

Or do you suggest making a public documented macro that users would set? Well, I do not think it's deserved, as the code is the same, just parameterized differently, and also I do not think users really want/need to go deep into oneDPL internals during the debug.

SergeyKopienko:
> > May be also some documented `broken macro` needed also? To inform users about this broken case and ability to make through this macro make the same behavior (optionally).
> 
> Is not `_ONEDPL_DETECT_COMPILER_OPTIMIZATIONS_ENABLED` already a macro that indicates this difference of "debug" vs "release" in this case? If someone wants, they can change its definition when debugging.
> 
> Or do you suggest making a public documented macro that users would set? Well, I do not think it's deserved, as the code is the same, just parameterized differently, and also I do not think users really want/need to go deep into oneDPL internals during the debug.

ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: OK

SergeyKopienko:
@mmichel11 I understood that this PR is partially return to some previous code state plus your new fixes.
So I don't insist on mine proposals for code change, up to you in any case.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: Done

mmichel11:
> @mmichel11 I understood that this PR is partially return to some previous code state plus your new fixes. So I don't insist on mine proposals for code change, up to you in any case.

My thought is just to leave the reverted code "as-is". The suggestions I think are a matter of developer preference, so let's minimize the reshuffling we are doing here.

danhoeflinger (approved):
LGTM, may want to wait for @akukanov to weigh in but I think this looks good and is a better solution than before.

akukanov:
At glance I see no issues.

PR #2047, opened 2025-02-05
Fix exception `Required aspect fp64 is not supported on the device` in tests of `std::complex`

SergeyKopienko:
Some tests of `std::compex` throws exception `Required aspect fp64 is not supported on the device`.
It's happen on Windows + GPU (without support `double` type) + MS Standard Library.

The reason of that - MS Standard Library implement some operations on `std::complex<float>` through `double` type inside.

### DIFF
--- a/documentation/CHANGES.rst
+++ b/documentation/CHANGES.rst
@@ -440,9 +440,7 @@ Known Issues and Limitations
 - ``std::array`` member function swap cannot be used in DPC++ kernels on Windows platform.
 - ``std::swap`` for ``std::array`` cannot work in DPC++ kernels on Windows platform.
 - Not all functions in <cmath> are supported currently, please refer to `DPC++ library guide <https://www.intel.com/content/www/us/en/docs/onedpl/developer-guide/current/overview.html>`_ for detail list.
-- Due to specifics of Microsoft Visual C++ implementation, some standard math functions for float (including: ``std::ldexp``, ``std::frexp``, ``std::acosh(std::complex<float>)``,``std::asin(std::complex<float>)``,
-``std::asinh(std::complex<float>)``, ``std::asoc(std::complex<float>)``, ``std::log(std::complex<float>)``, ``std::log10(std::complex<float>)``, ``std::pow(std::complex<float>, std::complex<float>)``,
-``std::sqrt(std::complex<float>)``) require device support for double precision.
+- Due to specifics of Microsoft Visual C++ implementation, some standard math functions for float (including: ``std::ldexp``, ``std::frexp``, ``std::sqrt(std::complex<float>)``) require device support for double precision.
 - There is a known issue on Windows platform with trying to use clGetPlatformInfo and ClGetDeviceInfo when using a graphics driver older than 27.20.100.8280.
   If you run into this issue, please upgrade to the latest driver of at least version 27.20.100.8280 from the `Download Center <https://downloadcenter.intel.com/product/80939/Graphics>`_.
 
--- a/documentation/CHANGES.rst
+++ b/documentation/CHANGES.rst
@@ -440,7 +440,9 @@ Known Issues and Limitations
 - ``std::array`` member function swap cannot be used in DPC++ kernels on Windows platform.
 - ``std::swap`` for ``std::array`` cannot work in DPC++ kernels on Windows platform.
 - Not all functions in <cmath> are supported currently, please refer to `DPC++ library guide <https://www.intel.com/content/www/us/en/docs/onedpl/developer-guide/current/overview.html>`_ for detail list.
-- Due to specifics of Microsoft Visual C++ implementation, some standard math functions for float (including: ``std::ldexp``, ``std::frexp``, ``std::sqrt(std::complex<float>)``) require device support for double precision.
+- Due to specifics of Microsoft Visual C++ implementation, some standard math functions for float (including: ``std::ldexp``, ``std::frexp``, ``std::acosh(std::complex<float>)``,``std::asin(std::complex<float>)``,
+``std::asinh(std::complex<float>)``, ``std::asoc(std::complex<float>)``, ``std::log(std::complex<float>)``, ``std::log10(std::complex<float>)``, ``std::pow(std::complex<float>, std::complex<float>)``,
+``std::sqrt(std::complex<float>)``) require device support for double precision.
 - There is a known issue on Windows platform with trying to use clGetPlatformInfo and ClGetDeviceInfo when using a graphics driver older than 27.20.100.8280.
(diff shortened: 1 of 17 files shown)


documentation/library_guide/api_for_sycl_kernels/tested_standard_cpp_api.rst
```diff
@@ -459,7 +459,7 @@ libc++ (LLVM)                                 Provided with Clang*-11.0
 --------------------------------------------- ---------------------------------------------
 Microsoft Visual C++* (MSVC) Standard Library Provided with Microsoft Visual Studio* 2017;
```
SergeyKopienko, line 460: @dmitriy-sobolev Which tested versions of C++ we should specify here?

documentation/CHANGES.rst
```diff
@@ -440,7 +440,9 @@ Known Issues and Limitations
 - ``std::array`` member function swap cannot be used in DPC++ kernels on Windows platform.
 - ``std::swap`` for ``std::array`` cannot work in DPC++ kernels on Windows platform.
 - Not all functions in <cmath> are supported currently, please refer to `DPC++ library guide <https://www.intel.com/content/www/us/en/docs/onedpl/developer-guide/current/overview.html>`_ for detail list.
-- Due to specifics of Microsoft Visual C++ implementation, some standard math functions for float (including: ``std::ldexp``, ``std::frexp``, ``std::sqrt(std::complex<float>)``) require device support for double precision.
+- Due to specifics of Microsoft Visual C++ implementation, some standard math functions for float (including: ``std::ldexp``, ``std::frexp``, ``std::acosh(std::complex<float>)``,``std::asin(std::complex<float>)``,
+``std::asinh(std::complex<float>)``, ``std::asoc(std::complex<float>)``, ``std::log(std::complex<float>)``, ``std::log10(std::complex<float>)``, ``std::pow(std::complex<float>, std::complex<float>)``,
+``std::sqrt(std::complex<float>)``) require device support for double precision.
```
danhoeflinger, line 445: It seems like old release notes (2021.1-beta07) is not the appropriate place to document this.  Perhaps "Known Limitations" within `documentation/library_guide/introduction.rst`

danhoeflinger:
So does this mean that the changes to "infinity" were not the problem / solution?  #2029

Do we have reason to revert those changes?

documentation/CHANGES.rst
SergeyKopienko: Done

test/xpu_api/numerics/complex.number/complex.transcendentals/asinh.pass.cpp
```diff
@@ -126,7 +126,12 @@ void test_edges()
 
 ONEDPL_TEST_NUM_MAIN
 {
+#if !_PSTL_TEST_COMPLEX_OP_ASINH_USING_DOUBLE
     test<float>();
+#else
+    IF_DOUBLE_SUPPORT(test<float>())
+#endif
```
danhoeflinger, line 133: On windows we are not testing anything here, I think the appropriate thing to do would be to report "skipped" rather than "passed" for these tests.

SergeyKopienko: I think it's not very simple exercise now. Required to change a lot of functions.
Also we always test code on host.

SergeyKopienko:
> So does this mean that the changes to "infinity" were not the problem / solution? #2029
> 
> Do we have reason to revert those changes?

I am not sure and propose to save as is.

test/xpu_api/numerics/complex.number/complex.transcendentals/asinh.pass.cpp
danhoeflinger: I guess I just don't know what utility this xpu_api test suite provides if we return success from this test when there is no version which works with the environment in question.

Perhaps we need to re-evaluate how we treat these tests within oneDPL. They probably need to be more separated from our oneDPL functional testing, and it is OK if they fail as long as they are an accurate depiction of the "status" of the standard library call within a kernel with a certain environment / hardware.

I wont insist here on changes, but we may need to address this at a higher level.

SergeyKopienko: May be you are right, but not in this PR.

documentation/library_guide/introduction.rst
```diff
@@ -142,8 +142,10 @@ Known Limitations
 * ``std::array::at`` member function cannot be used in kernels because it may throw an exception;
   use ``std::array::operator[]`` instead.
 * Due to specifics of Microsoft* Visual C++, some standard floating-point math functions
-  (including ``std::ldexp``, ``std::frexp``, ``std::sqrt(std::complex<float>)``) require device support
-  for double precision.
+  (including: ``std::ldexp``, ``std::frexp``, ``std::acosh(std::complex<float>)``,``std::asin(std::complex<float>)``,
+``std::asinh(std::complex<float>)``, ``std::asoc(std::complex<float>)``, ``std::log(std::complex<float>)``,
+``std::log10(std::complex<float>)``, ``std::pow(std::complex<float>, std::complex<float>)``,
+``std::sqrt(std::complex<float>)``) require device support for double precision.
```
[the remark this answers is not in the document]
SergeyKopienko: For note: `std::pow` has two arguments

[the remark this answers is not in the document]
SergeyKopienko: I am OK with any of variants.

documentation/library_guide/api_for_sycl_kernels/tested_standard_cpp_api.rst
dmitriy-sobolev: Let's review the versions to note here separately. We definitely should update them, e.g. VS 2017, which I see here, is not supported anymore.

documentation/library_guide/introduction.rst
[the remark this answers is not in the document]
dmitriy-sobolev: "which" in the suggestion looks not necessary:

> Due to specifics of Microsoft* Visual C++, some standard floating-point math functions (...) which require device support for double precision.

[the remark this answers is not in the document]
SergeyKopienko: Fixed

dmitriy-sobolev (approved):
LGTM.

I would slightly change the description for better flow:


Due to specifics of Microsoft* Visual C++, certain standard floating-point math functions (such as std::ldexp and std::frexp) and several complex number functions (std::acosh, std::asin, std::asinh, std::asoc, std::log10, std::log, std::pow, and std::sqrt when used with std::complex<float>) require device support for double precision.


but it might be a matter of taste.

danhoeflinger (approved):
LGTM,  I'm fine with either formulation of the description,  Dmitriy's looks good too.

PR #2048, opened 2025-02-05
Implement unified ETLS without TBB and use it for all host backends

danhoeflinger:
`tbb/enumerable_thread_specific.h` includes an indirect inclusion of windows.h on windows.  
This is undesirable, and therefore we would like to avoid it.  Without enumerable_thread_specific to depend upon in TBB, we can instead unify all implementations into a shared one which uses a pair of helpers: 
`__get_thread_num()`
and  
`__get_num_threads()` 
which are passed to the shared implementation from the parallel backends as template arguments during creation in the make function.  

We implement a make function for each backend to properly pass in the appropriate helper functions for the parallel backend. We no longer need to bury the __enumerable_thread_local_storage in a `__details` namespace, since all backends could in theory directly use the constructor if they fill out all the appropriate template arguments.  Variadic `Args...` are present in the type itself for all backends including TBB now.  However, make functions are beneficial to keep verbosity out of the local usage of `__enumerable_thread_local_storage`. 

---
Sadly, I don't see a way to automatically pick up the appropriate helper routines from the parallel backend from within `include/oneapi/dpl/pstl/parallel_backend_utils.h` because of a circular dependency without adding another utils header.  

We could consider adding another header `include/oneapi/dpl/pstl/parallel_unified_utils.h` which depends on the parallel backend rather than vice-versa.  This could allow us to depend upon a generic `__par_backend::` call for the two helpers. We could also unify all make functions into this header. The downside to this would be that we sacrifice the ability to use ETLS from the backends themselves, which we dont do now, but we may want to do in the future.

### DIFF
--- a/include/oneapi/dpl/pstl/omp/util.h
+++ b/include/oneapi/dpl/pstl/omp/util.h
@@ -153,13 +153,14 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
 namespace __detail
 {
 
-template <typename _ValueType, typename... Args>
+template <typename _ValueType, typename... _Args>
 struct __enumerable_thread_local_storage
-    : public oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage_base<
-          _ValueType, __enumerable_thread_local_storage<_ValueType, Args...>, Args...>
+    : public oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage_base<__enumerable_thread_local_storage,
+                                                                                    _ValueType, _Args...>
 {
-    using base_type = oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage_base<
-        _ValueType, __enumerable_thread_local_storage<_ValueType, Args...>, Args...>;
+    using base_type =
+        oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage_base<__enumerable_thread_local_storage,
+                                                                               _ValueType, _Args...>;
 
     template <typename... _LocalArgs>
     __enumerable_thread_local_storage(_LocalArgs&&... __args) : base_type(std::forward<_LocalArgs>(__args)...)
@@ -182,12 +183,12 @@ struct __enumerable_thread_local_storage
 } // namespace __detail
 
 // enumerable thread local storage should only be created from make function
-template <typename _ValueType, typename... Args>
-oneapi::dpl::__omp_backend::__detail::__enumerable_thread_local_storage<_ValueType, Args...>
-__make_enumerable_tls(Args&&... __args)
+template <typename _ValueType, typename... _Args>
+oneapi::dpl::__omp_backend::__detail::__enumerable_thread_local_storage<_ValueType, _Args...>
+__make_enumerable_tls(_Args&&... __args)
 {
-    return oneapi::dpl::__omp_backend::__detail::__enumerable_thread_local_storage<_ValueType, Args...>(
-        std::forward<Args>(__args)...);
+    return oneapi::dpl::__omp_backend::__detail::__enumerable_thread_local_storage<_ValueType, _Args...>(
+        std::forward<_Args>(__args)...);
 }
 } // namespace __omp_backend
 } // namespace dpl
--- a/include/oneapi/dpl/pstl/omp/util.h
+++ b/include/oneapi/dpl/pstl/omp/util.h
@@ -158,7 +158,7 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
 struct __get_num_threads
 {
     std::size_t
-    operator()()
+    operator()() const
     {
         return omp_in_parallel() ? omp_get_num_threads() : omp_get_max_threads();
     }
@@ -167,7 +167,7 @@ struct __get_num_threads
 struct __get_thread_num
 {
     std::size_t
-    operator()()
+    operator()() const
     {
         return omp_get_thread_num();
     }
--- a/include/oneapi/dpl/pstl/omp/util.h
+++ b/include/oneapi/dpl/pstl/omp/util.h
@@ -31,7 +31,7 @@
 #ifdef _MSC_VER
 #    define _PSTL_PRAGMA(x) __pragma(x)
 #else
-#    define _PSTL_PRAGMA(x) _Pragma(#    x)
+#    define _PSTL_PRAGMA(x) _Pragma(#x)
 #endif
 
 namespace oneapi
@@ -152,19 +152,28 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
 
 namespace __detail
 {
-struct __get_num_threads
+
+template <typename _ValueType, typename... Args>
+struct __enumerable_thread_local_storage
+    : public oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage_base<
+          _ValueType, __enumerable_thread_local_storage<_ValueType, Args...>, Args...>
 {
-    std::size_t
-    operator()() const
+    using base_type = oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage_base<
+        _ValueType, __enumerable_thread_local_storage<_ValueType, Args...>, Args...>;
+
+    template <typename... _LocalArgs>
+    __enumerable_thread_local_storage(_LocalArgs&&... __args) : base_type(std::forward<_LocalArgs>(__args)...)
+    {
+    }
+
+    static std::size_t
+    get_num_threads()
     {
         return omp_in_parallel() ? omp_get_num_threads() : omp_get_max_threads();
     }
-};
 
-struct __get_thread_num
-{
-    std::size_t
-    operator()() const
+    static std::size_t
+    get_thread_num()
     {
         return omp_get_thread_num();
     }
@@ -174,14 +183,11 @@ struct __get_thread_num
 
 // enumerable thread local storage should only be created from make function
 template <typename _ValueType, typename... Args>
-oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage<
-    _ValueType, oneapi::dpl::__omp_backend::__detail::__get_num_threads,
-    oneapi::dpl::__omp_backend::__detail::__get_thread_num, Args...>
+oneapi::dpl::__omp_backend::__detail::__enumerable_thread_local_storage<_ValueType, Args...>
 __make_enumerable_tls(Args&&... __args)
 {
-    return oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage<
-        _ValueType, oneapi::dpl::__omp_backend::__detail::__get_num_threads,
-        oneapi::dpl::__omp_backend::__detail::__get_thread_num, Args...>(std::forward<Args>(__args)...);
+    return oneapi::dpl::__omp_backend::__detail::__enumerable_thread_local_storage<_ValueType, Args...>(
+        std::forward<Args>(__args)...);
 }
 } // namespace __omp_backend
 } // namespace dpl
--- a/include/oneapi/dpl/pstl/omp/util.h
+++ b/include/oneapi/dpl/pstl/omp/util.h
@@ -16,11 +16,11 @@
 #ifndef _ONEDPL_INTERNAL_OMP_UTIL_H
 #define _ONEDPL_INTERNAL_OMP_UTIL_H
 
-#include <iterator> //std::iterator_traits, std::distance
-#include <cstddef> //std::size_t
-#include <memory> //std::allocator
+#include <iterator>    //std::iterator_traits, std::distance
+#include <cstddef>     //std::size_t
+#include <memory>      //std::allocator
 #include <type_traits> // std::decay, is_integral_v, enable_if_t
-#include <utility> // std::forward
+#include <utility>     // std::forward
 #include <omp.h>
 
 #include "../parallel_backend_utils.h"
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,74 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
...
+            }
+        }
+        // Need to back up one once we have found a valid storage object
+        return *__thread_specific_storage[__j - 1];
+    }
+
+    _ValueType&
+    get_for_current_thread()
+    {
+        std::size_t __i = _GetThreadNum{}();
```
SergeyKopienko, line 361:
```suggestion
        const std::size_t __i = _GetThreadNum{}();
```

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,74 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
...
 }
 
+template <typename _ValueType, typename _GetNumThreads, typename _GetThreadNum, typename... _Args>
+struct __enumerable_thread_local_storage
+{
+    template <typename... _LocalArgs>
+    __enumerable_thread_local_storage(_LocalArgs&&... __args)
+        : __num_elements(0), __args(std::forward<_LocalArgs>(__args)...)
+    {
+        std::size_t __num_threads = _GetNumThreads{}();
```
SergeyKopienko, line 316:
```suggestion
        const std::size_t __num_threads = _GetNumThreads{}();
```

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: fixed

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: fixed

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,74 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
...
+        __thread_specific_storage.resize(__num_threads);
```
SergeyKopienko, line 317: May be better to initialize this vector not in constructor body but in the member initialization list passing into vector size?

danhoeflinger: good point, fixed.

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,72 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
...
+    __enumerable_thread_local_storage(_LocalArgs&&... __args)
+        : __thread_specific_storage(_GetNumThreads{}()), __num_elements(0), __args(std::forward<_LocalArgs>(__args)...)
+    {
+    }
+
+    // Note: size should not be used concurrently with parallel loops which may instantiate storage objects, as it may
+    // not return an accurate count of instantiated storage objects in lockstep with the number allocated and stored.
+    // This is because the count is not atomic with the allocation and storage of the storage objects.
+    std::size_t
+    size() const
```
SergeyKopienko, line 322: Does it make sense to declare  this method as `noexcept' ?

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,72 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
...
+    {
+        // only count storage which has been instantiated
+        return __num_elements.load();
```
SergeyKopienko, line 325: I am not sure but probably we have small reserve here: may be the default param value [std::memory_order_seq_cst](http://en.cppreference.com/w/cpp/atomic/memory_order) may be changed.

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,72 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
...
+    {
+        assert(__i < size());
+
+        std::size_t __j = 0;
+
+        if (size() == __thread_specific_storage.size())
+        {
+            return *__thread_specific_storage[__i];
+        }
+
```
SergeyKopienko, line 346:
```suggestion
        if (size() == __thread_specific_storage.size())
        {
            return *__thread_specific_storage[__i];
        }

        std::size_t __j = 0;
```

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,72 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
...
+        }
+        // Need to back up one once we have found a valid storage object
+        return *__thread_specific_storage[__j - 1];
+    }
+
+    _ValueType&
+    get_for_current_thread()
+    {
+        const std::size_t __i = _GetThreadNum{}();
+        if (!__thread_specific_storage[__i])
```
SergeyKopienko, line 363: Does it make sense to replace access to item of `std::vector` through save refence to it into some local variable?

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: I suppose here we could use `memory_order_relaxed`.  Do you agree?

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,72 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
...
+            __thread_specific_storage[__i] =
+                std::apply([](_Args... __arg_pack) { return std::make_unique<_ValueType>(__arg_pack...); }, __args);
+            __num_elements.fetch_add(1);
+        }
+        return *__thread_specific_storage[__i];
+    }
+
+    std::vector<std::unique_ptr<_ValueType>> __thread_specific_storage;
+    std::atomic_size_t __num_elements;
+    std::tuple<_Args...> __args;
```
SergeyKopienko, line 375:
```suggestion
    const std::tuple<_Args...> __args;
```
May we declare this member as `const` ?

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: sure.

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: done.

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: I assume the compiler can likely do this, but I've done it here anyway.

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: sure, done.

include/oneapi/dpl/pstl/parallel_backend_utils.h
akukanov: Is there need to declare it `noexcept` though? Does it impact exception-related expectations of any classes or functions?
Generally, we do not use `noexcept` "just in case" (otherwise it would have been over all the code.

include/oneapi/dpl/pstl/parallel_backend_serial.h
```diff
@@ -171,6 +171,24 @@ __parallel_for_each(oneapi::dpl::__internal::__serial_backend_tag, _ExecutionPol
...
+
+struct __get_thread_num
+{
+    std::size_t
+    operator()() const
+    {
+        return 0;
+    }
+};
+
```
akukanov, line 191: A leftover artifact?

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: I don't know of a direct need.  My guess is that @SergeyKopienko Sergey noticed it on atomic `load()`, and since that is all we do here, to propagate it up.  No strong opinion here from my side.

SergeyKopienko: In this case my idea was that we simple use one operation - `std::atomic::load` which declared as `noexcept`
So we can declare this method too.

include/oneapi/dpl/pstl/parallel_backend_serial.h
danhoeflinger: yes, fixed.

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,76 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
...
+    get_for_current_thread()
+    {
+        const std::size_t __i = _GetThreadNum{}();
+        std::unique_ptr<_ValueType>& __thread_local_storage = __thread_specific_storage[__i];
+        if (!__thread_local_storage)
+        {
+            // create temporary storage on first usage to avoid extra parallel region and unnecessary instantiation
+            __thread_local_storage =
+                std::apply([](_Args... __arg_pack) { return std::make_unique<_ValueType>(__arg_pack...); }, __args);
+            __num_elements.fetch_add(1);
```
akukanov, line 368: We might as well add `memory_order_relaxed` here, as atomicity is still guaranteed, while both local and global order of operations do not seem important for counting.

include/oneapi/dpl/pstl/parallel_backend_utils.h
SergeyKopienko: I think so. @akukanov how do you think?

include/oneapi/dpl/pstl/parallel_backend_utils.h
akukanov: As I said, we can do it in many cases, but usually we do not. Let's not clutter the code with extra details that serve no purpose.

SergeyKopienko: ok

include/oneapi/dpl/pstl/parallel_backend_utils.h
akukanov: Given that we "require" `size()` to only be used in serial parts of code, and so there is no need to establish happens-before relations of particular expressions, I believe all operations with the atomic counter can be `relaxed`, including increments.

SergeyKopienko:
@danhoeflinger, I haven't any additional questions or comments from my side.

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -158,84 +153,42 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
 namespace __detail
 {
 
-template <typename _ValueType, typename... _Args>
+template <typename _ValueType, typename... Args>
```
akukanov, line 156: Probably we should add uglification in other places, not remove it here.

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -301,6 +306,79 @@ __set_symmetric_difference_construct(_ForwardIterator1 __first1, _ForwardIterato
     return __cc_range(__first2, __last2, __result);
 }
 
+namespace __detail
+{
+
+template <template <typename...> typename _Concrete, typename _ValueType, typename... _Args>
```
mmichel11, line 312:
```suggestion
template <template <typename, typename...> typename _Concrete, typename _ValueType, typename... _Args>
```
Optional suggestion, but would this be a better match to `template <typename _ValueType, typename... _Args>`?

akukanov: I believe it is not technically necessary. Even class templates that do not have variadic parameter packs at all, only at least one template parameter, would be suitable derived classes. That's how the variadic templates work.

But I would not object either, if that simplifies understanding for somebody.

danhoeflinger: Yeah, either way is technically correct.  Your proposed method is more specific / restrictive and perhaps more simple to understand, so why not,  I'll make the change.

akukanov: Just to clarify, it's not even more restrictive :) It says that at least one template parameter is required - but that was already a requirement before, due to the definition of `derived_type` that passed `_ValueType` as the argument.

akukanov (approved):
Looks good.

include/oneapi/dpl/pstl/parallel_backend_utils.h
danhoeflinger: Fair enough, in combination with the other parts of the definition, I agree.

mmichel11 (approved):
LGTM

danhoeflinger:
I had to revert CRTP implementation due to an issue on CL2017. Back to passing callables details from the individual parallel backends.  Otherwise, nothing has changed since approval.  CI run is going now.

mmichel11 (approved):
Reapproving

akukanov:
Restored the branch to try some improvements.

PR #2049, opened 2025-02-06
Improve the implementation of __enumerable_thread_local_storage

akukanov:
This patch returns to the CRTP-based implementation, with a workaround applied to make the VS 2017 C++ compiler happy.
But more importantly, it addresses two other issues: stores the construction arguments as a tuple of values (by removing references from the deduced types) and replaces `unique_ptr` with `optional` to avoid memory allocations.

### DIFF
--- a/include/oneapi/dpl/pstl/omp/util.h
+++ b/include/oneapi/dpl/pstl/omp/util.h
@@ -31,7 +31,7 @@
 #ifdef _MSC_VER
 #    define _PSTL_PRAGMA(x) __pragma(x)
 #else
-#    define _PSTL_PRAGMA(x) _Pragma(#    x)
+#    define _PSTL_PRAGMA(x) _Pragma(#x)
 #endif
 
 namespace oneapi
@@ -152,19 +152,28 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
 
 namespace __detail
 {
-struct __get_num_threads
+
+template <typename _ValueType, typename... Args>
+struct __enumerable_thread_local_storage
+    : public oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage_base<
+          _ValueType, __enumerable_thread_local_storage<_ValueType, Args...>, Args...>
 {
-    std::size_t
-    operator()() const
+    using base_type = oneapi::dpl::__utils::__detail::__enumerable_thread_local_storage_base<
+        _ValueType, __enumerable_thread_local_storage<_ValueType, Args...>, Args...>;
+
+    template <typename... _LocalArgs>
+    __enumerable_thread_local_storage(_LocalArgs&&... __args) : base_type(std::forward<_LocalArgs>(__args)...)
+    {
+    }
+
+    static std::size_t
+    get_num_threads()
     {
         return omp_in_parallel() ? omp_get_num_threads() : omp_get_max_threads();
     }
-};
 
-struct __get_thread_num
-{
-    std::size_t
-    operator()() const
+    static std::size_t
+    get_thread_num()
     {
         return omp_get_thread_num();
     }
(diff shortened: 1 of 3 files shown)


include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -1308,35 +1308,46 @@ __parallel_for_each(oneapi::dpl::__internal::__tbb_backend_tag, _ExecutionPolicy
...
+
+template <typename _ValueType, typename... _Args>
+struct __enumerable_thread_local_storage : public __etls_base<_ValueType, _Args...>
 {
-    std::size_t
-    operator()() const
+
+    template <typename... _LocalArgs>
+    __enumerable_thread_local_storage(_LocalArgs&&... __args)
+        : __etls_base<_ValueType, _Args...>({std::forward<_LocalArgs>(__args)...})
```
danhoeflinger, line 1325: I don't understand the benefit of wrapping the `__args` pack in braces on the way in to `__etls_base`. 
Can you explain why this is a benefit?

danhoeflinger: Nevermind, I see how the base has changed to accept a tuple of `__args...` rather than a pack of `__args` and then constructing a tuple.  I missed that change.  Now it makes sense.

danhoeflinger (approved):
LGTM, thanks for catching this problem and fixing it. (And also the workaround for vs2017)

include/oneapi/dpl/pstl/omp/util.h
```diff
@@ -152,36 +152,46 @@ __process_chunk(const __chunk_metrics& __metrics, _Iterator __base, _Index __chu
...
+
+// Workaround for VS 2017: declare an alias to the CRTP base template
+template <typename _ValueType, typename... _Args>
+struct __enumerable_thread_local_storage;
+
+template <typename... _Ts>
+using __etls_base = __utils::__enumerable_thread_local_storage_base<__enumerable_thread_local_storage, _Ts...>;
+
+template <typename _ValueType, typename... _Args>
+struct __enumerable_thread_local_storage : public __etls_base<_ValueType, _Args...>
```
SergeyKopienko, line 164:
```suggestion
struct __enumerable_thread_local_storage : __etls_base<_ValueType, _Args...>
```

include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -1308,35 +1308,46 @@ __parallel_for_each(oneapi::dpl::__internal::__tbb_backend_tag, _ExecutionPolicy
...
+
+// Workaround for VS 2017: declare an alias to the CRTP base template
+template <typename _ValueType, typename... _Args>
+struct __enumerable_thread_local_storage;
+
+template <typename... _Ts>
+using __etls_base = __utils::__enumerable_thread_local_storage_base<__enumerable_thread_local_storage, _Ts...>;
+
+template <typename _ValueType, typename... _Args>
+struct __enumerable_thread_local_storage : public __etls_base<_ValueType, _Args...>
```
SergeyKopienko, line 1320:
```suggestion
struct __enumerable_thread_local_storage : __etls_base<_ValueType, _Args...>
```

include/oneapi/dpl/pstl/omp/util.h
akukanov: I prefer to be explicit, since it's important there that the API of the base class is publicly available in the derived class.

SergeyKopienko: ok

include/oneapi/dpl/pstl/parallel_backend_tbb.h
SergeyKopienko: ok

SergeyKopienko (approved):
LGTM

PR #2050, opened 2025-02-06
Fix one more broken test case for std::exp(std::complex<double>)

SergeyKopienko:
In this PR we fix one more broken test case for std::exp(std::complex<double>)


test/xpu_api/numerics/complex.number/complex.transcendentals/exp.pass.cpp
```diff
@@ -83,7 +83,7 @@ void test_edges()
         else if (std::isinf(testcases[i].real()) && testcases[i].real() > 0 && std::isnan(testcases[i].imag()))
         {
             assert(std::isinf(r.real()));
-#if !_PSTL_ICC_TEST_COMPLEX_EXP_BROKEN_TEST_EDGES
+#if !_PSTL_ICC_TEST_COMPLEX_EXP_BROKEN_TEST_EDGES && !_PSTL_ICC_TEST_COMPLEX_EXP_BROKEN_TEST_EDGES_1
```
dmitriy-sobolev, line 86: Renaming: _PSTL_ICC_TEST_COMPLEX_EXP_BROKEN_TEST_EDGES_1 -> _PSTL_ICC_TEST_COMPLEX_EXP_BROKEN_TEST_EDGES_IMAG, as only imag() is affected here.

Also, remove unnecessary condiiton.

```suggestion
#if !_PSTL_ICC_TEST_COMPLEX_EXP_BROKEN_TEST_EDGES_IMAG
```

SergeyKopienko: Fixed

dmitriy-sobolev (approved):

test/xpu_api/numerics/complex.number/complex.transcendentals/exp.pass.cpp
dmitriy-sobolev: Discussed offline not to attach the name of the macro to a particular line as other similar issues may occur.

PR #2051, opened 2025-02-06
Update C++ tested versions in documentation

SergeyKopienko:


documentation/library_guide/api_for_sycl_kernels/tested_standard_cpp_api.rst
```diff
@@ -460,11 +460,4 @@ libc++ (LLVM)                                 Provided with Clang*-11.0
 Microsoft Visual C++* (MSVC) Standard Library Provided with Microsoft Visual Studio* 2017;
```
SergeyKopienko: @dmitriy-sobolev Do you know versions which we should specify here?

documentation/library_guide/api_for_sycl_kernels/tested_standard_cpp_api.rst
```diff
@@ -457,14 +457,6 @@ libstdc++ (GNU)                               Provided with GCC*-7.5.0, GCC*-9.3
 --------------------------------------------- ---------------------------------------------
 libc++ (LLVM)                                 Provided with Clang*-11.0
```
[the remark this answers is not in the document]
timmiesmith: I'll see what I can find.  It's been several years, but I have seen a mapping of gcc to libstdc++ versions.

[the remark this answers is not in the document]
akukanov: The major version of libstdc++ can be found as `_GLIBCXX_RELEASE` in `bits/c++config.h`, and generally, it should match the GCC version (`g++ -dumpfullversion`).

The version of libc++ can be found as `_LIBCPP_VERSION` in the `__config` header file, and is also said to match the version of LLVM/Clang (`clang++ -dumpfullversion`).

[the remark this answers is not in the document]
timmiesmith: I've updated the document based on the compiler versions used to test the HOST device.

documentation/library_guide/api_for_sycl_kernels/tested_standard_cpp_api.rst
```diff
@@ -453,18 +453,12 @@ C++ Standard API                      libstdc++  libc++     MSVC
 These tests were done for the following versions of the standard C++ library:
 
 ============================================= =============================================
-libstdc++ (GNU)                               Provided with GCC*-7.5.0, GCC*-9.3.0
+libstdc++ (GNU)                               Provided with GCC*-8.4.0, GCC*-9.3.0,
+                                              GCC*-11.4.0, GCC*-13.2.0
 --------------------------------------------- ---------------------------------------------
-libc++ (LLVM)                                 Provided with Clang*-11.0
+libc++ (LLVM)                                 Provided with Clang*-11.0, Clang*-14.0,
+                                              Clang*-17.0, Clang*-18.1
```
akukanov, line 460: @dcbenito Are the asterisks really necessary for the compiler names? I do not think these are trademarks, really.

akukanov: @timmiesmith I also think that hyphens between the name and the version are rather unusual, though I do not mind it either.

dcbenito: An asterisk is only needed on the first instance of a non-Intel product, so we only need a single asterisk. The hyphens, in my opinion, are not needed, as this is just a version number and we don't follow that format for any other version numbers.

timmiesmith: I've removed the hyphens and limited the asterisk to the first use as suggested.

akukanov (approved):

documentation/library_guide/api_for_sycl_kernels/tested_standard_cpp_api.rst
dcbenito: LGTM.

Issue #1679, opened 2024-07-08
`sycl::group_barrier` did not used in `__group_barrier(_Item __item)`


SergeyKopienko:
https://github.com/oneapi-src/oneDPL/blob/470df99e5f27ab8da3ee55941ff9d6a9e0aa9730/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h#L170

dmitriy-sobolev:
There are two major reasons of using the outdated barrier API:

1. Performance. You can find the details here: https://github.com/intel/llvm/issues/12531.
2. Non-compatible semantics, as mentioned in the code comments, which may require some work: https://github.com/oneapi-src/oneDPL/blob/470df99e5f27ab8da3ee55941ff9d6a9e0aa9730/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h#L171-L174

I suppose that the most important reason is 1. However, oneDPL is claimed to be SYCL 2020 conformant, so SYCL 2020 should be supported, at least for the compilers other than oneAPI DPC++ compiler.

al42and:
> All in all, it seems that the new group_barrier can be used in the same way / with the same effect as the old one.

As far as I understand, another difference (and the reason for performance drop mentioned in the linked issues) is that the old version "Executes a work-group barrier with memory ordering on the local address _space_", while the new version affects all memory operations, in both local and global address _spaces_ (but both only within work-group/local _scope_):

> sycl::memory_scope::work_group The ordering constraint applies only to work-items in the same [work-group](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#work-group) as the calling work-item;

akukanov:
Thanks @al42and - indeed, this is an important difference that I missed, and that impacts performance.

If that difference is important for oneDPL code, then we should make it visible - either in the oneDPL wrapper name or maybe with a template parameter - that this barrier orders operations only for data in local memory but does not order global data accesses.

akukanov:
The comment https://github.com/intel/llvm/issues/12531#issuecomment-2370736542 refers to a device compiler bug that has been fixed (https://github.com/intel/intel-graphics-compiler/commit/ed639f68d142bc963a7b626badc207a42fb281cb) and that should improve performance.

Despite the subtle semantical difference, we still need to drop the use of the outdated API.

dmitriy-sobolev:
[...]

PR #2054, opened 2025-02-07
Fix synchronization issues in radix sort and histogram

dmitriy-sobolev:
Value (`__buf_val`) and Count (`__buf_count`) buffers store data in either local or global memory. Let's use group barriers with proper fences (local or global memory fences) to avoid memory contention issues, which are observed on Xe2 architectures.

The fix changes how `__group_barrier` is defined, so the PR additionally includes what is done in #1988 to avoid conflicts and fix another issue related to the barriers: we use SYCL 1.2.1 barriers, but oneDPL claims to be SYCL 2020 compatible.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
@@ -173,8 +173,14 @@ struct __subgroup_radix_sort
                         uint16_t __begin_bit = 0;
                         constexpr uint16_t __end_bit = sizeof(_KeyT) * ::std::numeric_limits<unsigned char>::digits;
 
-                        auto __val_mem_adjusted_barrier = []() {
+                        auto __val_mem_adjusted_barrier = [__it]() {
                             if constexpr (_SLM_tag_val::value)
+                                __dpl_sycl::__group_barrier(__it);
+                            else
+                                __dpl_sycl::__group_barrier(__it, __dpl_sycl::__fence_space_global_and_local{});
+                        };
+                        auto __count_mem_adjusted_barrier = [__it]() {
+                            if constexpr (_SLM_counter::value)
                                 __dpl_sycl::__group_barrier(__it);
                             else
                                 __dpl_sycl::__group_barrier(__it, __dpl_sycl::__fence_space_global_and_local{});
@@ -182,7 +188,7 @@ struct __subgroup_radix_sort
 
                         //copy(move) values construction
                         __block_load<_ValT>(__wi, __src, __values.__v, __n);
-                        __dpl_sycl::__group_barrier(__it);
+                        __val_mem_adjusted_barrier();
 
                         while (true)
                         {
@@ -212,7 +218,7 @@ struct __subgroup_radix_sort
                                     __indices[__i] = *__counters[__i];
                                     *__counters[__i] = __indices[__i] + 1;
                                 }
-                                __dpl_sycl::__group_barrier(__it);
+                                __count_mem_adjusted_barrier();
 
                                 //2. scan phase
                                 {
@@ -225,8 +231,7 @@ struct __subgroup_radix_sort
                                     _ONEDPL_PRAGMA_UNROLL
                                     for (uint16_t __i = 1; __i < __bin_count; ++__i)
                                         __bin_sum[__i] = __bin_sum[__i - 1] + __counter_lacc[__wi * __bin_count + __i];
-
-                                    __dpl_sycl::__group_barrier(__it);
+                                    __count_mem_adjusted_barrier();
 
                                     //exclusive scan local sum
                                     uint16_t __sum_scan = __dpl_sycl::__exclusive_scan_over_group(
@@ -238,7 +243,7 @@ struct __subgroup_radix_sort
 
                                     if (__wi == 0)
                                         __counter_lacc[0] = 0;
-                                    __dpl_sycl::__group_barrier(__it);
+                                    __count_mem_adjusted_barrier();
                                 }
 
                                 _ONEDPL_PRAGMA_UNROLL
@@ -309,7 +314,7 @@ struct __subgroup_radix_sort
                                 if (__idx < __n)
                                     __values.__v[__i] = ::std::move(__exchange_lacc[__idx]);
                             }
-                            __dpl_sycl::__group_barrier(__it);
+                            __val_mem_adjusted_barrier();
                         }
                     }));
             });
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
@@ -175,8 +175,9 @@ struct __subgroup_radix_sort
 
                         //copy(move) values construction
                         __block_load<_ValT>(__wi, __src, __values.__v, __n);
+                        // __dpl_sycl::__group_barrier(__it);
+                        sycl::group_barrier(__it.get_group(), sycl::memory_scope::work_group);
 
-                        __dpl_sycl::__group_barrier(__it);
                         while (true)
                         {
                             uint16_t __indices[__block_size]; //indices for indirect access in the "re-order" phase
@@ -206,6 +207,7 @@ struct __subgroup_radix_sort
                                     *__counters[__i] = __indices[__i] + 1;
                                 }
                                 __dpl_sycl::__group_barrier(__it);
+                                // sycl::group_barrier(__it.get_group(), sycl::memory_scope::work_group);
 
                                 //2. scan phase
                                 {
@@ -220,6 +222,8 @@ struct __subgroup_radix_sort
                                         __bin_sum[__i] = __bin_sum[__i - 1] + __counter_lacc[__wi * __bin_count + __i];
 
                                     __dpl_sycl::__group_barrier(__it);
+                                    // sycl::group_barrier(__it.get_group(), sycl::memory_scope::work_group);
+
                                     //exclusive scan local sum
                                     uint16_t __sum_scan = __dpl_sycl::__exclusive_scan_over_group(
                                         __it.get_group(), __bin_sum[__bin_count - 1], __dpl_sycl::__plus<uint16_t>());
@@ -231,6 +235,7 @@ struct __subgroup_radix_sort
                                     if (__wi == 0)
                                         __counter_lacc[0] = 0;
                                     __dpl_sycl::__group_barrier(__it);
+                                    // sycl::group_barrier(__it.get_group(), sycl::memory_scope::work_group);
                                 }
 
                                 _ONEDPL_PRAGMA_UNROLL
@@ -245,6 +250,7 @@ struct __subgroup_radix_sort
 
                             //3. "re-order" phase
                             __dpl_sycl::__group_barrier(__it);
+                            // sycl::group_barrier(__it.get_group(), sycl::memory_scope::work_group);
                             if (__begin_bit >= __end_bit)
                             {
                                 // the last iteration - writing out the result
@@ -294,7 +300,8 @@ struct __subgroup_radix_sort
                                 }
                             }
 
-                            __dpl_sycl::__group_barrier(__it);
+                            // __dpl_sycl::__group_barrier(__it);
+                            sycl::group_barrier(__it.get_group(), sycl::memory_scope::work_group);
 
                             _ONEDPL_PRAGMA_UNROLL
                             for (uint16_t __i = 0; __i < __block_size; ++__i)
@@ -304,7 +311,8 @@ struct __subgroup_radix_sort
                                     __values.__v[__i] = ::std::move(__exchange_lacc[__idx]);
                             }
 
-                            __dpl_sycl::__group_barrier(__it);
+                            // __dpl_sycl::__group_barrier(__it);
+                            sycl::group_barrier(__it.get_group(), sycl::memory_scope::work_group);
                         }
                     }));
             });
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
@@ -80,6 +80,12 @@ struct __subgroup_radix_sort
         {
             return __dpl_sycl::__local_accessor<_KeyT>(__buf_size, __cgh);
         }
+
+        inline static constexpr auto
+        get_fence()
+        {
+            return __dpl_sycl::__fence_space_local;
+        }
     };
 
     template <typename _KeyT>
@@ -93,6 +99,12 @@ struct __subgroup_radix_sort
         get_acc(sycl::handler& __cgh)
         {
             return sycl::accessor(__buf, __cgh, sycl::read_write, __dpl_sycl::__no_init{});
+        }
+
+        inline constexpr static auto
+        get_fence()
+        {
+            return __dpl_sycl::__fence_space_global;
         }
     };
 
@@ -173,16 +185,9 @@ struct __subgroup_radix_sort
                         uint16_t __begin_bit = 0;
                         constexpr uint16_t __end_bit = sizeof(_KeyT) * ::std::numeric_limits<unsigned char>::digits;
 
-                        auto __mem_adjusted_barrier = [__it](auto __is_slm) {
-                            if constexpr (decltype(__is_slm)::value)
-                                __dpl_sycl::__group_barrier(__it);
-                            else
-                                __dpl_sycl::__group_barrier(__it, __dpl_sycl::__fence_space_global);
-                        };
-
                         //copy(move) values construction
                         __block_load<_ValT>(__wi, __src, __values.__v, __n);
-                        __mem_adjusted_barrier(_SLM_tag_val{}); // TODO: check if the barrier can be removed
+                        __dpl_sycl::__group_barrier(__it, decltype(__buf_val)::get_fence()); // TODO: check if the barrier can be removed
 
                         while (true)
                         {
@@ -212,7 +217,7 @@ struct __subgroup_radix_sort
                                     __indices[__i] = *__counters[__i];
                                     *__counters[__i] = __indices[__i] + 1;
                                 }
-                                __mem_adjusted_barrier(_SLM_counter{});
+                                __dpl_sycl::__group_barrier(__it, decltype(__buf_count)::get_fence());
 
                                 //2. scan phase
                                 {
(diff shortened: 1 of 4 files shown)


SergeyKopienko:
Taking into account the discussion from https://github.com/uxlfoundation/oneDPL/issues/1679
I think that probably the switch which barrier exactly we should use in all oneAPI code - may be incorrect approach...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -173,10 +173,23 @@ struct __subgroup_radix_sort
...
+                            else
+                                __dpl_sycl::__group_barrier(__it, __dpl_sycl::__fence_space_global_and_local{});
+                        };
+                        auto __count_mem_adjusted_barrier = [__it]() {
+                            if constexpr (_SLM_counter::value)
+                                __dpl_sycl::__group_barrier(__it);
+                            else
+                                __dpl_sycl::__group_barrier(__it, __dpl_sycl::__fence_space_global_and_local{});
+                        };
+
```
adamfidel, line 188: This just a thought, but could we use a single lambda here and pass the boolean constant as a parameter to it? Something like:
```c++
    auto __mem_adjusted_barrier = [__it](auto __is_slm) {
        if constexpr (decltype(__is_slm)::value)
            __dpl_sycl::__group_barrier(__it);
        else
            __dpl_sycl::__group_barrier(__it, __dpl_sycl::__fence_space_global_and_local{});
    };
```
And when used:
```c++
   __mem_adjusted_barrier(_SLM_tag_val{});
```
```c++   
  __mem_adjusted_barrier(_SLM_counter{});
```

See https://godbolt.org/z/GMx64dTb5.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -173,10 +173,23 @@ struct __subgroup_radix_sort
...
+                            if constexpr (_SLM_tag_val::value)
+                                __dpl_sycl::__group_barrier(__it);
+                            else
+                                __dpl_sycl::__group_barrier(__it, __dpl_sycl::__fence_space_global_and_local{});
+                        };
+                        auto __count_mem_adjusted_barrier = [__it]() {
+                            if constexpr (_SLM_counter::value)
+                                __dpl_sycl::__group_barrier(__it);
+                            else
+                                __dpl_sycl::__group_barrier(__it, __dpl_sycl::__fence_space_global_and_local{});
```
mmichel11, line 186: Since `__counters` is entirely through global memory in this case, is `__dpl_sycl::__fence_space_global` sufficient here?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -225,17 +226,45 @@ __get_accessor_size(const _Accessor& __accessor)
...
+#    if _ONEDPL_LIBSYCL_VERSION
+#        define ONEDPL_SYCL121_GROUP_BARRIER 1
+#    else
+#        define ONEDPL_SYCL121_GROUP_BARRIER 0
+#    endif
+#endif
+
+#if ONEDPL_SYCL121_GROUP_BARRIER
+template <sycl::access::fence_space _Space>
+struct __fence_space
```
MikeDvorskiy, line 245: I would suggest writing 3floor C++ code....
Actually? we already have the simple approach for wrapping sycl constants. For example,
see `__target_device` usage.

SergeyKopienko:
My assumptions that the changes from this PR is a little bit overcomplicated.
Let me propose alternative approach for this fix: https://github.com/uxlfoundation/oneDPL/pull/2055

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
MikeDvorskiy: Why do we need a compile time tag here?
Whether is it simpler to call with a constant, wrapped into `__dpl_sycl::name_xx`, like?

`__dpl_sycl::__group_barrier(__it, __dpl_sycl::fence_global_and_local);`

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -225,17 +226,45 @@ __get_accessor_size(const _Accessor& __accessor)
...
-    //TODO: usage of sycl::group_barrier: probably, we have to revise SYCL parallel patterns which use a group_barrier.
-    // 1) sycl::group_barrier() implementation is not ready
-    // 2) sycl::group_barrier and sycl::item::group_barrier are not quite equivalent
+#if ONEDPL_SYCL121_GROUP_BARRIER
+    __item.barrier(_Space::__value);
+#elif _ONEDPL_SYCL2020_GROUP_BARRIER_PRESENT
     sycl::group_barrier(__item.get_group(), sycl::memory_scope::work_group);
 #else
-    __item.barrier(sycl::access::fence_space::local_space);
+#    error "sycl::group_barrier is not supported, and no alternative is available"
```
MikeDvorskiy, line 267: Whether do we get potential issues here with user's code in case when neither `ONEDPL_SYCL121_GROUP_BARRIER `neither `_ONEDPL_SYCL2020_GROUP_BARRIER_PRESENT `is defined?

SergeyKopienko:
About
"The fix changes how __group_barrier is defined, so the PR additionally includes what is done in https://github.com/uxlfoundation/oneDPL/pull/1988 to avoid conflicts and fix another issue related to the barriers: we use SYCL 1.2.1 barriers, but oneDPL claims to be SYCL 2020 compatible."
- I believe it's incorrect moment to change that before release. Let's do this change later in the next release.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: The problem is that I need to provide some no-op alternative. Defining structures helps to achieve it. Could you provide an example in this context?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
dmitriy-sobolev: @MikeDvorskiy, I do not understand how you suggesting is supposed to work. We need to use different fence arguments. Could you be more specific? Perhaps I just do not get your idea.

dmitriy-sobolev: Meanwhile, I've implemented what Adam suggested.

dmitriy-sobolev:
> Taking into account the discussion from #1679 I think that probably the switcher which barrier exactly we should use in all `oneDPL` code - may be incorrect approach...

I briefly checked the other places: they all seem to be using local memory. But let me delve deeper.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
dmitriy-sobolev: Indeed, `__fence_space_global` is sufficient. I've applied your suggestion.

dmitriy-sobolev:
> About "The fix changes how __group_barrier is defined, so the PR additionally includes what is done in #1988 to avoid conflicts and fix another issue related to the barriers: we use SYCL 1.2.1 barriers, but oneDPL claims to be SYCL 2020 compatible."
> 
> * I believe it's incorrect moment to change that before release. Let's do this change later in the next release.

Well, technically, the changes and fixes should be as isolated as possible, especially when they are close to a release.
But, when I implemented the fix with barriers, keeping in mind that it should not conflict with #1988, I implemented the major part of #1988, which had already been approved. Why not to fix that as well... Anyway, the alternative version you provided is helpful if others think the opposite.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
danhoeflinger: Perhaps you can do this a little simpler...
```
#if ONEDPL_SYCL121_GROUP_BARRIER
constexpr sycl::access::fence_space __fence_space_local = sycl::access::fence_space::local_space;
constexpr sycl::access::fence_space __fence_space_global = sycl::access::fence_space::global_space;
constexpr sycl::access::fence_space __fence_space_global_and_local = sycl::access::fence_space::global_and_local;
#else
constexpr int __fence_space_local = 0;
constexpr int __fence_space_global = 0;
constexpr int __fence_space_global_and_local = 0;
#endif // ONEDPL_SYCL121_GROUP_BARRIER


template <typename _Item, typename _Space = decltype(__fence_space_local)>
void
__group_barrier(_Item __item, [[maybe_unused]] _Space __space = __fence_space_local)
{
#if ONEDPL_SYCL121_GROUP_BARRIER
    __item.barrier(__space);
#elif _ONEDPL_SYCL2020_GROUP_BARRIER_PRESENT
    sycl::group_barrier(__item.get_group(), sycl::memory_scope::work_group);
#else
#    error "sycl::group_barrier is not supported, and no alternative is available"

#endif
}
```
Does this work?  maybe there is a better "dummy" type / value than int, but otherwise I think it would work.  

You would just need to remove the `{}` from `__dpl_sycl::__fence_space_global{}` above I think.

danhoeflinger: I guess if we are following the example of `__target_device` we can do a similar trick to create a typename `__fence_space_t` which is either `sycl::access::fence_space` or `int` depending on `ONEDPL_SYCL121_GROUP_BARRIER`.

mmichel11:
> > About "The fix changes how __group_barrier is defined, so the PR additionally includes what is done in #1988 to avoid conflicts and fix another issue related to the barriers: we use SYCL 1.2.1 barriers, but oneDPL claims to be SYCL 2020 compatible."
> > 
> > * I believe it's incorrect moment to change that before release. Let's do this change later in the next release.
> 
> Well, technically, the changes and fixes should be as isolated as possible, especially when they are done close to a release. But, when I implemented the fix with barriers, keeping in mind that it should not conflict with #1988, I implemented the major part of #1988, which had already been approved. Why not to fix that as well... Anyway, the alternative version you provided is helpful if others think the opposite.

My opinion is the approach here is fine for the release. Since `ONEDPL_SYCL121_GROUP_BARRIER` defaults to 1 for the Intel SYCL implementation, we still get the same SYCL 1.2.1 barrier by default. Only if the user sets the variable themselves or a separate SYCL compiler is used, we get `sycl::group_barrier`. Using `sycl::group_barrier` in the case of a separate SYCL compiler is likely much safer than what we do now which is assume it supports the SYCL 1.2.1 barrier.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -225,17 +226,45 @@ __get_accessor_size(const _Accessor& __accessor)
...
+// SYCL 1.2.1 version is used due to having an API with a local memory fence,
+// which gives better performance on Intel GPUs.
+// The performance gap is negligible since
+// https://github.com/intel/intel-graphics-compiler/commit/ed639f68d142bc963a7b626badc207a42fb281cb (Aug 20, 2024)
+// But the fix is not a part of the LTS GPU drivers (Linux) yet.
+#if !defined(ONEDPL_SYCL121_GROUP_BARRIER)
+#    if _ONEDPL_LIBSYCL_VERSION
+#        define ONEDPL_SYCL121_GROUP_BARRIER 1
+#    else
+#        define ONEDPL_SYCL121_GROUP_BARRIER 0
```
danhoeflinger, line 239: This is the only think I see as "risky" in the PR now that we are past code freeze.  

This changes the default behavior for non-intel LLVM compilers to assume sycl 2020 group barrier support, where previous to this PR, we were assuming sycl 1.2.1 support.

Would it make sense to have a `SYCL_LANGUAGE_VERSION >= 202000` check here and decide for non-intel compilers that way?

(This remains a change which may not be "required" at a time where we should resist making changes).

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
```diff
@@ -241,26 +241,22 @@ __get_accessor_size(const _Accessor& __accessor)
...
-using __fence_space_global_and_local = __fence_space<sycl::access::fence_space::global_and_local>;
-#else
-struct __fence_space_local {};
-struct __fence_space_global {};
-struct __fence_space_global_and_local {};
+inline constexpr sycl::access::fence_space __fence_space_local = sycl::access::fence_space::local_space;
+inline constexpr sycl::access::fence_space __fence_space_global = sycl::access::fence_space::global_space;
+inline constexpr sycl::access::fence_space __fence_space_global_and_local = sycl::access::fence_space::global_and_local;
+#else
+struct __fence_space_dummy{}; // No-op dummy type since SYCL 2020 does specify memory fence spaces in group barriers
```
danhoeflinger, line 248:
```suggestion
struct __fence_space_dummy{}; // No-op dummy type since SYCL 2020 does not specify memory fence spaces in group barriers
```

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: The option with global constants looks good. I've implemented it with adding a dummy type.

SergeyKopienko:
If somebody will have interest for https://github.com/uxlfoundation/oneDPL/pull/2055, please reopen it.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
danhoeflinger: I think we have the following decision to make for the else branch here:
1) Minimize change post code freeze
`#define ONEDPL_SYCL121_GROUP_BARRIER 1`

2) Assume sycl 2020 for non intel llvm compiler 
`#define ONEDPL_SYCL121_GROUP_BARRIER 0`

3) Check sycl spec version macro and decide based on that
`#define ONEDPL_SYCL121_GROUP_BARRIER SYCL_LANGUAGE_VERSION >= 202000`


I don't have as good a grasp on what should be expected for non intel-llvm sycl compilers, so I'm interested what others think here.

danhoeflinger:
> I briefly checked the other places: they all seem to be using local memory. 
>But let me delve deeper.

It looks like one implementation of histogram may have a similar issue (my fault):

`__clear_wglocal_histograms` when called from `__histogram_general_private_global_atomics_submitter` is operating on global data. 

I can come up with a fix if you want.

danhoeflinger:
I've added https://github.com/uxlfoundation/oneDPL/pull/2056 in case we want to include it here.  The rest of the usages I believe should be in the local memory space.  If we have any question here with the changes, lets wait because this implementation requires a very large number of bins to be used.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: I suggest making a decision on Monday as it is late. In the meantime, I will test the PR using icx and acpp.

My personal preference is option 2: oneDPL declares SYCL 2020 compatibility, so this change can be treated as a bug fix, while there is a backup option to manually set `ONEDPL_SYCL121_GROUP_BARRIER` for those who rely on these barriers.

danhoeflinger: Does acpp correctly define `SYCL_LANGUAGE_VERSION`?

If so, would (3) be a more sycl specification compliant version of basically the same as (2)?

dmitriy-sobolev: > Does acpp correctly define SYCL_LANGUAGE_VERSION?

It does: https://github.com/AdaptiveCpp/AdaptiveCpp/blob/e1c6c6ede4c17877a2005f876afc869f11fba5be/include/hipSYCL/sycl/sycl.hpp#L27C1-L27C38

> If so, would (3) be a more sycl specification compliant version of basically the same as (2) at least for acpp?

Agree (3) looks better than (2) with this additional check. The check should be slightly different, though. If it is SYCL 2020, then define `ONEDPL_SYCL121_GROUP_BARRIER` to `0`.

danhoeflinger: yes you are correct.  Thanks for the correction.

SergeyKopienko:
I think as far as @dmitriy-sobolev introduced three new types here, probably required to declare that they are device-copyable:
- `__fence_space_local`
- '__fence_space_global'
- `__fence_space_global_and_local'

SergeyKopienko:
Additional question: in the current state of this PR the value `__fence_space_global_and_local` is never used.
So should we really declare this value `__fence_space_global_and_local` ?

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: I checked that ACPP supports SYCL 2020 barriers. I am now slightly inclined towards keeping option (2) as we want to have only SYCL 2020 barriers in the future, and we do not now other implementations still having only SYCL 1.2.1 barriers, but which are SYCL 2020 compliant in other places.

danhoeflinger: I'm OK with that option, but lets leave a comment explaining it.

dmitriy-sobolev:
> Additional question: in the current state of this PR the value `__fence_space_global_and_local` is never used. So should we really declare this value `__fence_space_global_and_local` ?

They were indeed unnecessary, I've removed them.

dmitriy-sobolev:
> I think as far as @dmitriy-sobolev introduced three new types here, probably required to declare that they are device-copyable (`is_device_copyable`):
> 
> * `__fence_space_local`
> * `__fence_space_global`
> * `__fence_space_global_and_local`

I see a couple of problems here:
 - We should not copy these object onto a device.
 - `sycl::access::fence_space` is a trivially copyable type (enum), so should also be device copyable.

We discussed it offline, and the issue was work-arounded by declaring `get_fence` as static (a function this issue occurred), but I still have no understanding why the lack of `static` triggered that issue...

danhoeflinger (dismissed):
I agree with the first clang formatting complaint, not the others.  Otherwise, LGTM.

Probably makes sense to move the TODO above the line.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_defs.h
dmitriy-sobolev: Added:
> // For safety, assume that other SYCL implementations comply with SYCL 2020, which is a oneDPL requirement.

mmichel11 (approved):
LGTM

PR #2055, opened 2025-02-07
Alternative approach of working with barriers

SergeyKopienko:
The initial issue has been detected and investigated by @dmitriy-sobolev  in the PR https://github.com/uxlfoundation/oneDPL/pull/2054

This PR simple make the fit for that issue by a little bit another way.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
@@ -183,8 +183,8 @@ struct __subgroup_radix_sort
                 oneapi::dpl::__ranges::__require_access(__cgh, __src);
 
                 // Get accessors to sycl::buffer (global memory) or sycl::local_accessor (local memory)
-                auto __acc_buf_val   = __buf_val.get_acc  (__cgh);
-                auto __acc_buf_count = __buf_count.get_acc(__cgh);
+                auto __exchange_lacc = __buf_val.get_acc  (__cgh);
+                auto __counter_lacc = __buf_count.get_acc(__cgh);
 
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -74,6 +74,9 @@ struct __subgroup_radix_sort
         uint16_t __buf_size;
 
       public:
+
+        static constexpr sycl::access::fence_space _space = sycl::access::fence_space::local_space;
```
mmichel11, line 78: I think the problem with this approach is that `sycl::access::fence_space` is not a part of SYCL2020 at all, so we cannot expect a compiler to implement it.

SergeyKopienko: I think it's simple not documented now, but not deprecated too.
And we haven't any warnings about that.

SergeyKopienko: Practically we are using this enum in oneDPL code in main branch too, but only in one place.

mmichel11: I see, so I guess the current state is probably not compatible with non-dpcpp SYCL compilers unless they also have this SYCL 1.2.1 support.

SergeyKopienko: May be. But in any case, we are not making anything worse at this place.

SergeyKopienko:
As far as I understood this implementation is not preferable, so I close the PR.

PR #2056, opened 2025-02-07
Global barrier fence histogram fix

danhoeflinger:
This changes the global barrier's fence spaces for histogram  catch all which uses global atomics and no SLM.

I'm not sure if we want to add this in right now, or take care of this after the release.  
This is only used when the number of histogram bins is very large as compared to the memory space of the device.

If we have any question marks here, lets wait on these changes.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
@@ -127,10 +127,12 @@ class __histo_kernel_local_atomics;
 template <typename... _Name>
 class __histo_kernel_private_glocal_atomics;
 
-template <typename _HistAccessor, typename _OffsetT, typename _Size>
+template <typename _HistAccessor, typename _OffsetT, typename _Size,
+          typename _FenceSpace = decltype(__dpl_sycl::__fence_space_local)>
 void
 __clear_wglocal_histograms(const _HistAccessor& __local_histogram, const _OffsetT& __offset, _Size __num_bins,
-                           const sycl::nd_item<1>& __self_item)
+                           const sycl::nd_item<1>& __self_item,
+                           _FenceSpace __fence_space = __dpl_sycl::__fence_space_local)
 {
     using _BinUint_t =
         ::std::conditional_t<(sizeof(_Size) >= sizeof(::std::uint32_t)), ::std::uint64_t, ::std::uint32_t>;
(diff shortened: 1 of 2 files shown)


akukanov (dismissed):

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
```diff
@@ -127,10 +127,12 @@ class __histo_kernel_local_atomics;
 template <typename... _Name>
 class __histo_kernel_private_glocal_atomics;
 
-template <typename _HistAccessor, typename _OffsetT, typename _Size>
+template <typename _HistAccessor, typename _OffsetT, typename _Size,
+          typename _FenceSpace = decltype(__dpl_sycl::__fence_space_local)>
```
akukanov, line 131: Though... should we decay the type here?

danhoeflinger: `__dpl_sycl::__fence_space_local` is either a `constexpr inline`  `sycl::access::fence_space` or `struct __fence_space_dummy{};

Does it change anything to decay `decltype(__dpl_sycl::__fence_space_local)`?  I would think it doesn't.  I suppose it doesn't hurt, but I'm not seeing the real danger of not decaying right now.

danhoeflinger: I should say it will remove `const`, but I'm not sure why we want to do that here.

danhoeflinger: One way to clear all this up would be to define a `__dpl__sycl::__fence_space_t` in the main PR and then using it in places where we are otherwise using `decltype(__dpl_sycl::__fence_space_local)`.  (here and in one place in the main PR I think)

danhoeflinger: I added the definition of `__dpl__sycl::__fence_space_t` and associated changes.  I like it better, and it removes this question about decay.

@akukanov @dmitriy-sobolev what do you think?

mmichel11:
The approach here LGTM once the conflicts are resolved.

danhoeflinger:
I've rebased (a couple times).  @dmitriy-sobolev feel free to merge if you agree with the changes.

mmichel11 (approved):

dmitriy-sobolev (approved):
LGTM. I like an addition simplification.

PR #2058, opened 2025-02-10
[test] Reduce for algorithm testing with debug builds to prevent timeouts

mmichel11:
## Description
The input sizes used for testing for-based algorithms is dependent on the number of execution units on the SYCL queue. This approach was used to ensure we generate large enough inputs to test the large for submitter path in oneDPL. A hard cap is set at 10,000,000 elements which is sufficient to test the large submitter path on high-performance GPUs with many EUs.

However, we recently spotted that this approach results in test timeouts on client GPUs such as BMG and Arc in debug mode. To fix this issue, a new approach has been added for debug mode to avoid such timeouts while ensuring a single test case for the large submitter path. The "normal" set of inputs is used (factors of PI up to 100k elements) followed by a single test case above the threshold to invoke the large submitter.  For release builds, testing remains unchanged and is more extensive. Test times on BMG with debug builds now in several hundred seconds (as opposed to timing out at 1200) and CI is able to complete with several hours before the timeout threshold.
## Details
The previous utility `get_pattern_for_max_n` utility which returned a single maximum input is replaced with `get_pattern_for_test_sizes` which returns a monotonically increasing sequence of elements to test with. Each for-based test is touched to accommodate these changes.

### DIFF
--- a/test/support/utils.h
+++ b/test/support/utils.h
@@ -1031,7 +1031,7 @@ get_pattern_for_max_n()
     // threshold for Debug (-O0 + -g) builds where runtimes are significantly longer.
     std::size_t cap =
 #if PSTL_USE_DEBUG
-        500000;
+        2000000;
 #else
         10000000;
 #endif
--- a/test/support/utils.h
+++ b/test/support/utils.h
@@ -1022,6 +1022,9 @@ inline std::vector<std::size_t>
 get_pattern_for_test_sizes()
 {
     std::size_t max_n = 0;
+    // We do not enable large input size testing for FPGA devices as __parallel_for_submitter_fpga only has a single
+    // implementation with the standard input sizes providing full coverage, and testing large inputs is slow with the
+    // FPGA emulator.
 #if TEST_DPCPP_BACKEND_PRESENT && !ONEDPL_FPGA_DEVICE
     sycl::queue q = TestUtils::get_test_queue();
     sycl::device d = q.get_device();
@@ -1034,7 +1037,8 @@ get_pattern_for_test_sizes()
 #if TEST_DPCPP_BACKEND_PRESENT && !PSTL_USE_DEBUG && !ONEDPL_FPGA_DEVICE
     std::size_t cap = 10000000;
     max_n = multiplier * large_submitter_limit;
-    max_n = std::min(cap, max_n);
+    // Ensure that TestUtils::max_n <= max_n <= cap
+    max_n = std::max(TestUtils::max_n, std::min(cap, max_n));
 #else
     max_n = TestUtils::max_n;
 #endif
--- a/test/support/utils.h
+++ b/test/support/utils.h
@@ -1014,34 +1014,39 @@ generate_arithmetic_data(T* input, std::size_t size, std::uint32_t seed)
     }
 }
 
-// Utility that models __estimate_best_start_size in the SYCL backend parallel_for
-// to ensure large enough inputs are used to test the large submitter path.
-// A multiplier to the max n is added to ensure we get a few separate test inputs for
-// this path.
-std::size_t
-get_pattern_for_max_n()
+// Utility that models __estimate_best_start_size in the SYCL backend parallel_for to ensure large enough inputs are
+// used to test the large submitter path. A multiplier to the max n is added to ensure we get a few separate test inputs
+// for this path. For debug testing, only test with a single large n to avoid timeouts. Returns a monotonically increasing
+// sequence for use in testing.
+inline std::vector<std::size_t>
+get_pattern_for_test_sizes()
 {
-#if TEST_DPCPP_BACKEND_PRESENT
+    std::size_t max_n = 0;
+#if TEST_DPCPP_BACKEND_PRESENT && !ONEDPL_FPGA_DEVICE
     sycl::queue q = TestUtils::get_test_queue();
     sycl::device d = q.get_device();
     constexpr std::size_t max_iters_per_item = 16;
     constexpr std::size_t multiplier = 4;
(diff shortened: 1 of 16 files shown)


danhoeflinger:
Are we losing coverage on full kernels by shrinking this space or are we still at least hitting all the same kernels on debug?

mmichel11:
> Are we losing coverage on full kernels by shrinking this space or are we still at least hitting all the same kernels on debug?

On integrated graphics, we will be hitting all of the kernels in debug mode which make up most of our CI testing. However, on PVC we will not. I had experimented with a cap of 2 million which would ensure this for PVC, but I saw very long times with debug builds on battlemage (500 - 600 seconds per test).

mmichel11:
> > Are we losing coverage on full kernels by shrinking this space or are we still at least hitting all the same kernels on debug?
> 
> On integrated graphics, we will be hitting all of the kernels in debug mode which make up most of our CI testing. However, on PVC / BMG we will not. I had experimented with a cap of 2 million which would ensure this for PVC, but I saw very long times with debug builds on battlemage (500 - 600 seconds per test).

Moving to draft while I test a new strategy on the following branch: https://github.com/uxlfoundation/oneDPL/compare/dev/mmichel11/pfor_debug_input_size_reduce_tmp

test/support/utils.h
```diff
@@ -1014,26 +1014,39 @@ generate_arithmetic_data(T* input, std::size_t size, std::uint32_t seed)
...
+// Utility that models __estimate_best_start_size in the SYCL backend parallel_for to ensure large enough inputs are
+// used to test the large submitter path. A multiplier to the max n is added to ensure we get a few separate test inputs
+// for this path. For debug testing, only test with a single large n to avoid timeouts. Returns a monotonically increasing
+// sequence for use in testing.
+inline std::vector<std::size_t>
+get_pattern_for_test_sizes()
 {
-#if TEST_DPCPP_BACKEND_PRESENT
+    std::size_t max_n = 0;
+#if TEST_DPCPP_BACKEND_PRESENT && !ONEDPL_FPGA_DEVICE
```
dmitriy-sobolev, line 1025: What is the purpose of having a separate flow for an FPGA device? Is it because the emulator returns some unusual device properties?

test/support/utils.h
```diff
@@ -1014,26 +1014,39 @@ generate_arithmetic_data(T* input, std::size_t size, std::uint32_t seed)
...
     sycl::queue q = TestUtils::get_test_queue();
     sycl::device d = q.get_device();
     constexpr std::size_t max_iters_per_item = 16;
     constexpr std::size_t multiplier = 4;
     constexpr std::size_t max_work_group_size = 512;
-    std::size_t __max_n = multiplier * max_iters_per_item * max_work_group_size *
-                          d.get_info<sycl::info::device::max_compute_units>();
-    __max_n = std::min(std::size_t{10000000}, __max_n);
-    return __max_n;
+    const std::size_t large_submitter_ub =
```
dmitriy-sobolev, line 1031: What does _ub stand on?

dmitriy-sobolev (commented):
The PR looks good enough to me leaving aside some minor comments.

test/support/utils.h
mmichel11: We have a separate parallel-for implementation for FPGA: https://github.com/uxlfoundation/oneDPL/blob/dbf9c08048e144e294d29ae8e288d2b0e35667df/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h#L56 which does not have a small and large submitter. As a result, we do not need to test these large input sizes here for full coverage, and it is likely very slow if we do.

test/support/utils.h
mmichel11: Upper bound. The maximum iterations per work-item is 16 for any data type, so the `large_submitter_ub` is the maximum possible start point to invoke the large submitter for any type.

dmitriy-sobolev: I see. I thought it was undefined behaviour. Let it be _limit for peace of mind of other developers.

test/support/utils.h
dmitriy-sobolev: Thanks. Let's leave a small note about the intention...

test/support/utils.h
```diff
@@ -1014,26 +1014,39 @@ generate_arithmetic_data(T* input, std::size_t size, std::uint32_t seed)
...
+        max_iters_per_item * max_work_group_size * d.get_info<sycl::info::device::max_compute_units>();
+#endif
+#if TEST_DPCPP_BACKEND_PRESENT && !PSTL_USE_DEBUG && !ONEDPL_FPGA_DEVICE
+    std::size_t cap = 10000000;
+    max_n = multiplier * large_submitter_ub;
+    max_n = std::min(cap, max_n);
```
dmitriy-sobolev, line 1037: Do we also want to make sure that we at least `TestUtils::max_n`?

test/support/utils.h
mmichel11: Done

test/support/utils.h
mmichel11: Yes, let's avoid this confusion :) I switched to `large_submitter_limit`.

test/support/utils.h
mmichel11: Sure, on some devices we could potentially run into this. I switched to:
`max_n = std::max(TestUtils::max_n, std::min(cap, max_n));`

dmitriy-sobolev (approved):
LGTM

danhoeflinger (approved):
This seems like a good strategy with some flexibility going forward, allowing us to keep general coverage on debug builds.  LGTM

PR #1976, opened 2024-12-19
Re-implement SYCL backend `parallel_for` to improve bandwidth utilization

mmichel11:
**High Level Description**
This PR improves hardware bandwidth utilization of oneDPL's SYCL backend parallel for pattern through two ideas:
- Process multiple input iterations per work-item which involves a switch to a nd_range kernel combined with a sub / work group strided indexing approach.
- To generate wide loads for small types, implement a path that vectorizes loads / stores by processing adjacent indices within a single work item. This is combined with the above approach to maximize hardware bandwidth utilization. Vectorization is only applied to fundamental types of size less than 4 (e.g. uint16_t, uint8_t) under a contiguous container.
 
**Implementation Details**
- Parallel for bricks have been reworked in the following manner:
    - Each brick contains a pack of ranges within its template parameters to define tuning parameters.
    - The following static integral members are defined (implemented with inheritance):
        - __can_vectorize
        - __preferred_vector_size (1 if __can_vectorize is false)
        - __preferred_iters_per_item
    - The following public member functions are defined
        - __scalar_path (for small input sizes this member function is explicitly called)
        - __vector_path (optional for algorithms that are not vectorizable e.g. `binary_search`)
        - An overloaded function call operator which dispatches to the appropriate strategy

To implement this approach, the parallel for kernel rewrite from https://github.com/oneapi-src/oneDPL/pull/1870 was adopted with additional changes to handle vectorization paths. Additionally, generic vectorization and strided loop utilities have been defined with the intention for these to be applicable in other portions of the codebase as well. Tests have been expanded to ensure coverage of vectorization paths.

This PR will supersedes https://github.com/oneapi-src/oneDPL/pull/1870. Initially, the plan was to merge this PR into 1870 but after comparing the diff, I believe the most straightforward approach will be to target this directly to main.

PR #2065, opened 2025-02-12
Fix circular dependency between utils_ranges_sycl.h and ranges_defs.h

mmichel11:
PR #1976 introduced a circular dependency between `oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h` and `oneapi/dpl/pstl/ranges_defs.h`. This shows up as an issue in kernel templates where the following produces a compiler error:

```
#include <oneapi/dpl/experimental/kernel_templates>

int main()
{
}
```
In `utils_ranges_sycl.h`, the include was added to be able to use `__nanorange::nano::ranges::contiguous_range`, so we can fix this by directly including the nanorange header. This resolves the circular dependency and the compilation issue.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
@@ -23,7 +23,7 @@
 #endif
 
 #include "../../utils_ranges.h"
-#include "../../ranges_defs.h" // contiguous_range and contiguous_iterator from nanorange
+#include "../../ranges/nanorange.hpp" // contiguous_range and contiguous_iterator from nanorange
 #include "../../iterator_impl.h"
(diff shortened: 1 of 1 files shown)


dmitriy-sobolev (approved):
LGTM.

`utils_ranges_sycl.h` defines what `ranges_defs.h` uses (such as `all_view`), so `utils_ranges_sycl.h` should not include `ranges_defs.h`. Including `nanorange.hpp` directly looks good.

akukanov:
The patch is OK as a quick fix,

But in the long run, the fix should be different. Namely, I believe `ranges_defs.h` should NOT include any utility headers, because its purpose is to provide forward declarations, not implementation parts.

timmiesmith:
@mmichel11 would you please open an issue capturing @akukanov's direction for the correct fix?

PR #2069, opened 2025-02-14
Fix CTAD in merge sort

dmitriy-sobolev:
This fixes issue with template argument deduction of `DropViews` nested structure. The issue happens with clang 17 and earlier. icpx 2023.2, which is based on clang 17, is also prone to that issue.

Reproducer: https://godbolt.org/z/WhMb5W8on

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
@@ -330,6 +330,8 @@ struct __merge_sort_global_submitter<_IndexT, __internal::__optional_kernel_name
         {
         }
     };
+    template <typename Rng>
+    DropViews(Rng&, const WorkDataArea&) -> DropViews<Rng>;
 
     template <typename _ExecutionPolicy>
     std::size_t


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h
```diff
@@ -330,6 +330,8 @@ struct __merge_sort_global_submitter<_IndexT, __internal::__optional_kernel_name
         {
         }
     };
+    template <typename Rng>
+    DropViews(Rng&, const WorkDataArea&) -> DropViews<Rng>;
```
mmichel11, line 334: Do we want to leave a comment explaining why the deduction guide is necessary to workaround a compiler issue?

dmitriy-sobolev: Yes, that would be helpful.  I've left a comment about the reason to add the deduction guide.

However, I still do not understand whether this is a fix of oneDPL issue (relying on template argument deduction in a context where it is not expected to work), or a compiler issue for a better comment.

dmitriy-sobolev: I've updated the comment. It is indeed a clang++ bug.

danhoeflinger (approved):
LGTM

mmichel11 (approved):
LGTM

SergeyKopienko:
LGTM

PR #2070, opened 2025-02-17
Fix various warnings in the library

dmitriy-sobolev:
The PR fixes warnings, where the fix is the most straightforward one:
 - -Wreorder-ctor
 - -Wunused-variable
 - -Wlogical-op-parentheses
 - -Wunused-lambda-capture
 - -Wunused-parameter

Examples of the messages:

1. dpl/pstl/algorithm_impl.h(387): warning C4100: '__exec': unreferenced formal parameter
3. oneDPL/include/oneapi/dpl/pstl/histogram_binhash_utils.h:68:29: warning: field '__num_bins' will be initialized after field '__range_size' [-Wreorder-ctor]
4. oneDPL/include/oneapi/dpl/pstl/parallel_backend_tbb.h:868:14: warning: unused variable '__ky' [-Wunused-variable]
5. oneDPL\include\oneapi\dpl\pstl\parallel_backend_tbb.h(887): warning C4189: '__ny': local variable is initialized but not referenced
6. oneDPL/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h:182:99: warning: '&&' within '||' [-Wlogical-op-parentheses]
7. oneDPL/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h:772:14: warning: unused variable '__n' [-Wunused-variable]
8. oneDPL/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge_sort.h:218:36: warning: lambda capture '__comp' is not used [-Wunused-lambda-capture]
9. oneDPL\include\oneapi\dpl\pstl\hetero\algorithm_ranges_impl_hetero.h(172,42): warning: unused parameter '__tag' [-Wunused-parameter]

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
@@ -179,8 +179,8 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, const _I
         if constexpr (__can_use_ternary_op_v<decltype(__rng1[__rng1_idx]), decltype(__rng2[__rng2_idx])>)
         {
             // This implementation is required for performance optimization
-            __rng3[__rng3_idx] = (!__rng1_idx_less_n1 || __rng1_idx_less_n1 && __rng2_idx_less_n2 &&
-                                                             __comp(__rng2[__rng2_idx], __rng1[__rng1_idx]))
+            __rng3[__rng3_idx] = (!__rng1_idx_less_n1 || (__rng1_idx_less_n1 && __rng2_idx_less_n2 &&
+                                                          __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])))
                                      ? __rng2[__rng2_idx++]
                                      : __rng1[__rng1_idx++];
         }
@@ -188,7 +188,7 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, const _I
         {
             // TODO required to understand why the usual if-else is slower then ternary operator
             if (!__rng1_idx_less_n1 ||
-                __rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx]))
+                (__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])))
                 __rng3[__rng3_idx] = __rng2[__rng2_idx++];
             else
                 __rng3[__rng3_idx] = __rng1[__rng1_idx++];
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -179,7 +179,7 @@ __pattern_walk1(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomAcc
 
 template <class _Tag, class _ExecutionPolicy, class _ForwardIterator, class _Brick>
 void
-__pattern_walk_brick(_Tag, _ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last,
+__pattern_walk_brick(_Tag, _ExecutionPolicy&&, _ForwardIterator __first, _ForwardIterator __last,
                      _Brick __brick) noexcept
 {
     static_assert(__is_serial_tag_v<_Tag> || __is_parallel_forward_tag_v<_Tag>);
@@ -241,7 +241,7 @@ __pattern_walk1_n(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __exec, _R
 
 template <class _Tag, class _ExecutionPolicy, class _ForwardIterator, class _Size, class _Brick>
 _ForwardIterator
-__pattern_walk_brick_n(_Tag, _ExecutionPolicy&& __exec, _ForwardIterator __first, _Size __n, _Brick __brick) noexcept
+__pattern_walk_brick_n(_Tag, _ExecutionPolicy&&, _ForwardIterator __first, _Size __n, _Brick __brick) noexcept
 {
     static_assert(__is_serial_tag_v<_Tag> || __is_parallel_forward_tag_v<_Tag>);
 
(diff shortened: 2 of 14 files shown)


danhoeflinger (commented):
It probably makes sense to merge #1996 first (which I think is very close), then this will be easier to review.  
(or rebase this on it)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -179,16 +179,16 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, const _I
         if constexpr (__can_use_ternary_op_v<decltype(__rng1[__rng1_idx]), decltype(__rng2[__rng2_idx])>)
         {
             // This implementation is required for performance optimization
-            __rng3[__rng3_idx] = (!__rng1_idx_less_n1 || __rng1_idx_less_n1 && __rng2_idx_less_n2 &&
-                                                             __comp(__rng2[__rng2_idx], __rng1[__rng1_idx]))
+            __rng3[__rng3_idx] = (!__rng1_idx_less_n1 || (__rng1_idx_less_n1 && __rng2_idx_less_n2 &&
+                                                          __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])))
```
MikeDvorskiy, line 183: Why the redundant parentheses were added?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -179,16 +179,16 @@ __serial_merge(const _Rng1& __rng1, const _Rng2& __rng2, _Rng3& __rng3, const _I
...
                                      ? __rng2[__rng2_idx++]
                                      : __rng1[__rng1_idx++];
         }
         else
         {
             // TODO required to understand why the usual if-else is slower then ternary operator
             if (!__rng1_idx_less_n1 ||
-                __rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx]))
+                (__rng1_idx_less_n1 && __rng2_idx_less_n2 && __comp(__rng2[__rng2_idx], __rng1[__rng1_idx])))
```
MikeDvorskiy, line 191: Why the redundant parentheses were added?

MikeDvorskiy (commented):
Looks OK to me, excepting the redundant parentheses in several places.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
dmitriy-sobolev: It silences this warning:

```
oneDPL/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h:182:99: warning: '&&' within '||' [-Wlogical-op-parentheses]
```

The compiler is a bit paranoid - we know the priority of && and ||. Etherealness, I prefer not to introduce even such warnings because it is a library code...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
dmitriy-sobolev: See the answer below

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
danhoeflinger: I also think this improves clarity for someone reading the code, but I know that may just be personal preference.

MikeDvorskiy: > The compiler is a bit paranoid

Absolutely agree. It is the same  if the compiler would warn about  `a*b + 7` and force to write `(a*b) + 7`.
I would suggest suppressing the warning.

danhoeflinger: I agree with Adam and this is why I always defer to the idea that explicit intention should be expressed by the programmer in any situation which is close to ambiguous.  
Even if the person reading the code knows the precedence, the extra parens confirm that the original code was written with this intention in mind (and it is not a bug lurking).  

In the code, its cost is just a couple of characters.  For the person writing the code, its cost the small inconvenience of another commit if we someday enforce reducing warnings to a minimum.

akukanov: Unfortunately, it's not the compiler being paranoid, it's users asking for this. I had the same sort of argument some time ago when worked on TBB, and I was on the position that programmers should know the precedence of `&&` to `||` (after all, noone would put parenthesis around multiplication to "disambiguate" precedence with addition. But it seems too many programmers do it wrong.

So if we want to enable -Wall or alike as part of our testing strategy, to avoid excessive warnings for users who use it in their projects (and it's quite common these days), we, as library developers, have to bite the bullet and do things we do not like.

mmichel11: An alternative option that I want to mention would be to selectively suppress warnings we disagree with via compiler pragmas: https://clang.llvm.org/docs/UsersManual.html#controlling-diagnostics-via-pragmas. My understanding is that via the push / pop mechanism we can target code blocks in our headers without affecting warnings in user code. I think TBB does this in some places.

I do not like this option though and think that we should directly fix these warnings as others suggest.

akukanov: Right, header-scope suppression with pragmas is an option which we can consider. Thanks @mmichel11 for mentioning this.

dmitriy-sobolev:
> It probably makes sense to merge #1996 first (which I think is very close), then this will be easier to review. (or rebase this on it)

It's been rebased.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
dmitriy-sobolev: It seems the majority prefers to keep paratheses. Let me close the discussion.

danhoeflinger (approved):
LGTM, great work.  We can see the warnings going down in the new reports, and all the changes look fine to me.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
MikeDvorskiy: The democracy is great thing :-)

PR #2071, opened 2025-02-17
Silence negators deprecation warnings

dmitriy-sobolev:
It suppresses the warning:

oneDPL\include\oneapi\dpl\pstl../functional(51): warning C4996: 'std::binary_negate': warning STL4008: std::not1(), std::not2(), std::unary_negate, and std::binary_negate are deprecated in C++17. They are superseded by std::not_fn(). You can define _SILENCE_CXX17_NEGATORS_DEPRECATION_WARNING or _SILENCE_ALL_CXX17_DEPRECATION_WARNINGS to suppress this warning.

### DIFF
--- a/include/oneapi/dpl/functional
+++ b/include/oneapi/dpl/functional
@@ -17,7 +17,18 @@
 #define _ONEDPL_FUNCTIONAL
 
 #include "oneapi/dpl/internal/common_config.h"
+
+// Silence the deprecation warnings of the negators injected from <functional>
+// The macro definition is contained in this header to avoid redefinition errors on the user side
+#if _ONEDPL___cplusplus < 202002L && defined(_MSC_VER) && !defined(_SILENCE_CXX17_NEGATORS_DEPRECATION_WARNING)
+#    define _SILENCE_CXX17_NEGATORS_DEPRECATION_WARNING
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/functional
```diff
@@ -17,7 +17,18 @@
 #define _ONEDPL_FUNCTIONAL
 
 #include "oneapi/dpl/internal/common_config.h"
+
+// Silence the deprecation warnings of the negators injected from <functional>
+// The macro definition is contained in this header to avoid redefinition errors on the user side
```
akukanov, line 22: In my understanding - please correct if I am wrong - the "silence" macro effectively does not add the deprecation attribute to the negation functions. That means that a) if the standard header has been included before oneDPL headers, the warnings will still be there, and b) if we include the header with the "silence" macro defined, it will impact the whole translation unit anyway.

The direct conclusion from (b) is that undefining the macro makes no good whatsoever, and maybe even harms, by disabling the warnings and leaving no clue it has been made.

akukanov: Overall, I think we should not do this suppression. If we position our headers as substitutes for the standard ones (i.e., these can be used instead), in most if not all cases we should not change the behavior of the standard headers.

PR #2073, opened 2025-02-20
[Tests] Remove broken macro for MSVC array swap

danhoeflinger:
We added a workaround for an issue with swapping with MSVC in #1560, but neglected to remove the test's "broken" macro.

This PR removes the broken macro, which allows us to resume testing on windows with the workaround.


timmiesmith (approved):
LGTM

PR #2077, opened 2025-02-24
Sort test: split and make it more flexible to reduce execution time

dmitriy-sobolev:
The main motivation is to reduce the execution time of the test as it takes too much time.
The PR splits sort.pass into four files: stable/unstable sort, interfaces with/without a comparator.

It also more flexibility through filtering to avoid exponential growth of configurations. See `SortTestConfig` uses, and the table below showing execution times.  
It also improves diagnostics, for example:
 - Before: `sort.pass.cpp:216 - wrong result from sort without predicate #2` (note: it is misleading - it can be with a predicate).
 - After: `utils_sort.h:307 - float, device, custom greater, total size = 50718, mismatch with reference at index 38847`

The test was also refactored:
 - `::` was removed before `std::`
 - `_` prefix was removed (no uglification needed in the tests)
 - `test_default_name_gen` function was simplified
 - some functions were regrouped. 

----
Below are measurements of ctest times of the sort test(s). `main` includes `sort.pass` time, and `this PR` combines the times of the added four tests. 

Device | configuration | main, s | this PR, s
-- | -- | -- | --
GPU, Intel® Data Center GPU Max 1100 | release | 310 | 8
GPU, Intel® Data Center GPU Max 1100 | debug | 1170 | 40
  |   |   |  
CPU (Intel(R) Xeon(R) Platinum 8480+, 2   sockets) | release | 286 | 20
CPU (Intel(R) Xeon(R) Platinum 8480+, 2   sockets) | debug | segfault at 127s | segfault at 14s in sort_comp.pass, 47s the rest combined
  |   |   |  
CPU (Intel(R) Core(TM) i7-1270P) | release | 373 | 186
CPU (Intel(R) Core(TM) i7-1270P) | debug | >3600 | 390

Note: `utils_sort.h` preserves the history of `sort.pass`, but GitHub shows it as a new file (which is probably because of large number of changes). Let me know if the refactoring changes should be rolled back to simplify revewing.


test/parallel_api/algorithm/alg.sorting/alg.stable.sort/stable_sort_comp.pass.cpp
```diff
@@ -0,0 +1,67 @@
...
+    test_sort<std::uint16_t>(cfg.msg("uint16_t, device, ConstGreater"), extended_sizes, Device<1>{},
+                             Converter<std::uint16_t>{}, ConstGreater{});
+#if __SYCL_UNNAMED_LAMBDA__
+    // Test potentially clashing naming for radix sort descending / ascending with minimal timing impact
+    test_default_name_gen(cfg.msg("default name generation"));
+#endif
+    // TODO: add a test for stability
+#endif
+
+    return TestUtils::done();
```
SergeyKopienko, line 66: Looking to used defines I think it's possible then this test doing nothing.
In this case important to return `return TestUtils::done(0);`

It's make sense for this test and probably for other added in this PR tests too.

dmitriy-sobolev: This combination skips the tests: "test only device policies" (`TEST_ONLY_HETERO_POLICIES`) and "device backend is not available" (`!TEST_DPCPP_BACKEND_PRESENT`). I treat it as invalid. I've added a check to stop testing if it happens.

test/support/utils_sort.h
```diff
@@ -0,0 +1,473 @@
...
+               OutputIterator2 /* expected_first */, OutputIterator2 /* expected_last */, InputIterator /* first */,
+               InputIterator /* last */, Size /* n */, Compare .../*compare*/)
+    {
+    }
+};
+
+template <typename T, typename Invoker, typename Converter, typename... Compare>
+void
+test_sort(SortTestConfig config, const std::vector<std::size_t> sizes, Invoker invoker, Converter converter,
+          Compare... compare)
```
SergeyKopienko, line 439:
```suggestion
test_sort(SortTestConfig config, const std::vector<std::size_t>& sizes, Invoker invoker, Converter converter,
          Compare... compare)
```
- please avoid to pass `std::vector` by value

test/support/utils_sort.h
```diff
@@ -0,0 +1,473 @@
...
+    // Ignored for host policies
+    bool test_usm_device;
+    bool test_usm_shared;
+
+    std::string err_msg_prefix;
+
+    SortTestConfig(bool is_stable, bool usm_device = false, bool usm_shared = false, const std::string& msg = "")
+        : is_stable(is_stable), test_usm_device(usm_device), test_usm_shared(usm_shared), err_msg_prefix(msg) {}
+
+    SortTestConfig msg(const std::string& msg) const
```
SergeyKopienko, line 54: Looks like UB: you pass 3 uninitialized fields into constructor: `is_stable, test_usm_device, test_usm_shared`

test/support/utils_sort.h
```diff
@@ -0,0 +1,473 @@
...
+#endif
+
+// The struct allows filtering configurations to avoid exponential growth of test cases
+// and to pass an error message prefix for better diagnostics
+struct SortTestConfig
+{
+    bool is_stable;
+    // Ignored for host policies
+    bool test_usm_device;
+    bool test_usm_shared;
```
SergeyKopienko, line 47:
```suggestion
    bool is_stable = false;
    // Ignored for host policies
    bool test_usm_device = false;
    bool test_usm_shared = false;
```

test/support/utils_sort.h
SergeyKopienko, line 54: This design looks a little bit strange from my point of view.
It's some non-static factory method, declared as `const`.
I think we just needs one more constructor like
```C++
    SortTestConfig(const SortTestConfig& cfg, const std::string& msg)
        : SortTestConfig(cfg.is_stable, cfg.test_usm_device, cfg.test_usm_shared, msg)
    {
    }
```

test/support/utils_sort.h
```diff
@@ -0,0 +1,473 @@
...
+
+    std::string err_msg_prefix;
+
+    SortTestConfig(bool is_stable, bool usm_device = false, bool usm_shared = false, const std::string& msg = "")
+        : is_stable(is_stable), test_usm_device(usm_device), test_usm_shared(usm_shared), err_msg_prefix(msg) {}
```
SergeyKopienko, line 52:
```suggestion
    SortTestConfig(bool is_stable, bool test_usm_device = false, bool test_usm_shared= false, const std::string& err_msg_prefix= "")
        : is_stable(is_stable), test_usm_device(test_usm_device), test_usm_shared(test_usm_shared), err_msg_prefix(msg) {}
```

test/support/utils_sort.h
dmitriy-sobolev: Done

test/support/utils_sort.h
dmitriy-sobolev: https://github.com/uxlfoundation/oneDPL/pull/2077/files#r1967600533 allows adding `SortTestConfig() = default;`, what I did.

test/support/utils_sort.h
dmitriy-sobolev: Done, thanks.

test/support/utils_sort.h
dmitriy-sobolev: Done

test/parallel_api/algorithm/alg.sorting/alg.stable.sort/stable_sort_comp.pass.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+#if TEST_DPCPP_BACKEND_PRESENT
+    test_sort<TestUtils::float32_t>(SortTestConfig{cfg, "float, device, custom less comparator"}, sizes, Device<0>{},
+                                    Converter<TestUtils::float32_t>{}, ConstLess{});
+
+    // Test merge-path sort specialization for large sizes, see __get_starting_size_limit_for_large_submitter
+    std::vector<std::size_t> extended_sizes = test_sizes(8'000'000);
+    test_sort<std::uint16_t>(SortTestConfig{cfg, "uint16_t, device, custom greater comparator"}, extended_sizes,
+                             Device<1>{}, Converter<std::uint16_t>{}, ConstGreater{});
+#if __SYCL_UNNAMED_LAMBDA__
```
danhoeflinger, line 59: Is there a reason we only care about this name clashing for stable sort with a comparator?
Perhaps this is enough to prevent regression on all APIs, but it relies upon some knowledge of the implementation, and could change (will add JIT time to add this to the other tests).

dmitriy-sobolev: Now this case is also in `sort_comp.pass`.

test/parallel_api/algorithm/alg.sorting/alg.stable.sort/stable_sort_comp.pass.cpp
SergeyKopienko: ok

test/parallel_api/algorithm/alg.sorting/alg.stable.sort/stable_sort_comp.pass.cpp
```diff
@@ -0,0 +1,67 @@
...
+    // Test potentially clashing naming for radix sort descending / ascending with minimal timing impact
+    test_default_name_gen(SortTestConfig{cfg, "default name generation"});
+#endif
+    // TODO: add a test for stability
```
SergeyKopienko, line 63: ???

SergeyKopienko (approved):
LGTM
But I seen one TODO. Is it ok?

test/parallel_api/algorithm/alg.sorting/alg.stable.sort/stable_sort_comp.pass.cpp
dmitriy-sobolev: We have not tested stability in the original test. This note simply highlights it.

PR #2078, opened 2025-02-24
Ranges sort tests: filter out policies to reduce execution time

dmitriy-sobolev:
`std_ranges_sort.pass` and `std_ranges_stable_sort.pass` tests take to much time to complete. Let's filter out cases which do not have much sense, for example testing millions elements with host policies and thousands with serial ones.
Testing 32M elements with device policies also is too much. There are no specializations for more than 16M elements.

### DIFF
--- a/test/parallel_api/ranges/std_ranges_sort.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_sort.pass.cpp
@@ -24,9 +24,9 @@ main()
 
     auto sort_checker = TEST_PREPARE_CALLABLE(std::ranges::sort);
 
-    test_range_algo<0, int, data_in, SkipHostPolicies>{big_sz}(dpl_ranges::sort, sort_checker);
+    test_range_algo<0, int, data_in, DevicePolicy>{big_sz}(dpl_ranges::sort, sort_checker);
 
-    test_range_algo<1, int, data_in, ParHostPolicies>{medium_sz}(dpl_ranges::sort, sort_checker, std::ranges::less{});
+    test_range_algo<1, int, data_in, DeviceAndParPolicies>{medium_sz}(dpl_ranges::sort, sort_checker, std::ranges::less{});
 
     test_range_algo<2>{}(dpl_ranges::sort, sort_checker, std::ranges::less{}, proj);
     test_range_algo<3>{}(dpl_ranges::sort, sort_checker, std::ranges::greater{}, proj);
--- a/test/parallel_api/ranges/std_ranges_sort.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_sort.pass.cpp
@@ -24,9 +24,8 @@ main()
 
     auto sort_checker = TEST_PREPARE_CALLABLE(std::ranges::sort);
 
-    test_range_algo<0, int, data_in, DevicePolicy>{big_sz}(dpl_ranges::sort, sort_checker);
-
-    test_range_algo<1, int, data_in, DeviceAndParPolicies>{medium_sz}(dpl_ranges::sort, sort_checker, std::ranges::less{});
+    test_range_algo<0>{}(dpl_ranges::sort, sort_checker);
+    test_range_algo<1>{}(dpl_ranges::sort, sort_checker, std::ranges::less{});
 
     test_range_algo<2>{}(dpl_ranges::sort, sort_checker, std::ranges::less{}, proj);
     test_range_algo<3>{}(dpl_ranges::sort, sort_checker, std::ranges::greater{}, proj);
@@ -36,6 +35,12 @@ main()
 
     test_range_algo<6, P2>{}(dpl_ranges::sort, sort_checker, std::ranges::less{}, &P2::proj);
     test_range_algo<7, P2>{}(dpl_ranges::sort, sort_checker, std::ranges::greater{}, &P2::proj);
+
+    // Check larger specializations, use a custom comparator for a comparison-based sort (e.g. merge-sort)
+    auto custom_less = [](const auto& a, const auto& b){ return a < b;};
+    test_range_algo<8, std::uint16_t, data_in, DevicePolicy>{big_sz}(dpl_ranges::sort, sort_checker, custom_less);
+    test_range_algo<9, int, data_in, DeviceAndParPolicies>{medium_sz}(dpl_ranges::sort, sort_checker, custom_less);
+
 #endif //_ENABLE_STD_RANGES_TESTING
 
     return TestUtils::done(_ENABLE_STD_RANGES_TESTING);
--- a/test/parallel_api/ranges/std_ranges_sort.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_sort.pass.cpp
@@ -24,8 +24,9 @@ main()
 
     auto sort_checker = TEST_PREPARE_CALLABLE(std::ranges::sort);
 
-    test_range_algo<0>{big_sz}(dpl_ranges::sort, sort_checker);
-    test_range_algo<1>{}(dpl_ranges::sort, sort_checker, std::ranges::less{});
+    test_range_algo<0, int, data_in, SkipHostPolicies>{big_sz}(dpl_ranges::sort, sort_checker);
+
+    test_range_algo<1, int, data_in, ParHostPolicies>{medium_sz}(dpl_ranges::sort, sort_checker, std::ranges::less{});
 
     test_range_algo<2>{}(dpl_ranges::sort, sort_checker, std::ranges::less{}, proj);
     test_range_algo<3>{}(dpl_ranges::sort, sort_checker, std::ranges::greater{}, proj);
--- a/test/parallel_api/ranges/std_ranges_stable_sort.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_stable_sort.pass.cpp
@@ -24,7 +24,7 @@ main()
 
     auto sort_stable_checker = TEST_PREPARE_CALLABLE(std::ranges::stable_sort);
 
-    test_range_algo<0>{big_sz}(dpl_ranges::stable_sort, sort_stable_checker);
+    test_range_algo<0>{}(dpl_ranges::stable_sort, sort_stable_checker);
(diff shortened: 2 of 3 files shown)


test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -144,22 +143,34 @@ template<typename T>
...
+    {
+        operator()(oneapi::dpl::execution::par, algo, checker,  args...);
+        operator()(oneapi::dpl::execution::par_unseq, algo, checker, args...);
+    }
+    void
+    operator()(SkipHostPolicies, auto algo, auto& checker, auto... args) {}
 
     template<typename Policy, typename Algo, typename Checker, typename TransIn, typename TransOut, TestDataMode mode = test_mode>
-    std::enable_if_t<!std::is_same_v<Policy, std::true_type> && mode == data_in>
+    std::enable_if_t<mode == data_in && !std::is_same_v<Policy, AllHostPolicies> &&
```
SergeyKopienko, line 171: May be better to describe this rule in `if constexpr` ?

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -225,15 +236,19 @@ struct test
 
 public:
     template<typename Policy, typename Algo, typename Checker, TestDataMode mode = test_mode>
-    std::enable_if_t<!std::is_same_v<Policy, std::true_type> && mode == data_in_out>
+    std::enable_if_t<mode == data_in_out && !std::is_same_v<Policy, AllHostPolicies> &&
```
SergeyKopienko, line 239: The same

test/parallel_api/ranges/std_ranges_test.h
SergeyKopienko, line 171: The condition
```C++
!std::is_same_v<Policy, AllHostPolicies> &&
                                               !std::is_same_v<Policy, ParHostPolicies> &&
                                               !std::is_same_v<Policy, SkipHostPolicies>
```
has used 5 times in your changes.

What about to pack it into some useful type?

dmitriy-sobolev: Thanks, I've added a trait

dmitriy-sobolev: That's arguable. `if contexpr` is more readable than `enable_if`, but the former one will lead to larger functions and more indents. I am inclined towards keeping the original `enable_if` design.

SergeyKopienko: ok

test/parallel_api/ranges/std_ranges_sort.pass.cpp
```diff
@@ -24,8 +24,9 @@ main()
 
     auto sort_checker = TEST_PREPARE_CALLABLE(std::ranges::sort);
 
-    test_range_algo<0>{big_sz}(dpl_ranges::sort, sort_checker);
-    test_range_algo<1>{}(dpl_ranges::sort, sort_checker, std::ranges::less{});
+    test_range_algo<0, int, data_in, DevicePolicy>{big_sz}(dpl_ranges::sort, sort_checker);
+
+    test_range_algo<1, int, data_in, DeviceAndParPolicies>{medium_sz}(dpl_ranges::sort, sort_checker, std::ranges::less{});
```
danhoeflinger, line 29: To me, it seems a little strange to sort a big size of 16M, a medium of 128K, and then a small of max 10 elements for tests 2-7.

I think we could increase the small (default) to something like 2k perhaps without any real effect on timing.

test/parallel_api/ranges/std_ranges_stable_sort.pass.cpp
```diff
@@ -24,8 +24,10 @@ main()
 
     auto sort_stable_checker = TEST_PREPARE_CALLABLE(std::ranges::stable_sort);
 
-    test_range_algo<0>{big_sz}(dpl_ranges::stable_sort, sort_stable_checker);
-    test_range_algo<1>{}(dpl_ranges::stable_sort, sort_stable_checker, std::ranges::less{});
+    test_range_algo<0, int, data_in, DevicePolicy>{big_sz}(dpl_ranges::stable_sort, sort_stable_checker);
```
danhoeflinger, line 27: Effectively, these are 7 different signatures / options for the ranges API.  
Should we be fully eliminating the host policies test for this simplest 2 signatures?

It seems what we actually want might be to just limit / scale the max elements tested based on policy type.

test/parallel_api/ranges/std_ranges_sort.pass.cpp
dmitriy-sobolev: Thanks, I've made a change to keep the coverage.

test/parallel_api/ranges/std_ranges_stable_sort.pass.cpp
dmitriy-sobolev: Thanks, I've made a change to keep the coverage.

test/parallel_api/ranges/std_ranges_sort.pass.cpp
dmitriy-sobolev: Done

test/parallel_api/ranges/std_ranges_stable_sort.pass.cpp
dmitriy-sobolev: An approach with scaling the sizes for the policies also looks good.

It must be simpler. For example we can pass multiple sizes instead of single `max_n`: one for serial policies, one for par policies and one for device policies. Or we can pass coefficients for it to be "scaling" as suggested.

test/parallel_api/ranges/std_ranges_stable_sort.pass.cpp
```diff
@@ -24,10 +24,8 @@ main()
 
     auto sort_stable_checker = TEST_PREPARE_CALLABLE(std::ranges::stable_sort);
 
-    test_range_algo<0, int, data_in, DevicePolicy>{big_sz}(dpl_ranges::stable_sort, sort_stable_checker);
-
-    test_range_algo<1, int, data_in, DeviceAndParPolicies>{medium_sz}(dpl_ranges::stable_sort, sort_stable_checker,
-                                                                      std::ranges::less{});
+    test_range_algo<0>{big_sz}(dpl_ranges::stable_sort, sort_stable_checker);
```
danhoeflinger: Did you mean to remove this `big_sz` here?

test/parallel_api/ranges/std_ranges_stable_sort.pass.cpp
danhoeflinger: I'm OK with either strategy.  What you have now works for me, but I agree that it may result in simpler code to merely pass different max_n for different policies.

test/parallel_api/ranges/std_ranges_stable_sort.pass.cpp
dmitriy-sobolev: Yep, removed

danhoeflinger (approved):
LGTM

test/parallel_api/ranges/std_ranges_stable_sort.pass.cpp
dmitriy-sobolev: Could you check #2081? @MikeDvorskiy, you may be interested too.
Let's then decide what which way to follow.

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -50,7 +52,16 @@ auto dpcpp_policy()
 }
 #endif //TEST_DPCPP_BACKEND_PRESENT
 
-auto host_policies() { return std::true_type{};}
+struct AllPolicies{};
+struct DeviceAndParPolicies{};
+struct DevicePolicy{};
+
+template <typename T>
+struct is_host_policy_selector: std::bool_constant<std::is_same_v<T, AllPolicies> ||
```
SergeyKopienko, line 60: Probably some construction with using `std::disjunction` like usen in the implementation of `__has_known_identity` may be more clear....

SergeyKopienko: Also I think the usage of `std::bool_constant` really isn't required here at all.

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -50,7 +52,16 @@ auto dpcpp_policy()
...
+                                                   std::is_same_v<T, DeviceAndParPolicies> ||
+                                                   std::is_same_v<T, DevicePolicy>> {};
```
SergeyKopienko, line 62:
```suggestion
template <typename T>
using is_host_policy_selector = std::disjunction<std::is_same<T, AllPolicies>,
                                                 std::is_same<T, DeviceAndParPolicies>,
                                                 std::is_same<T, DevicePolicy>>;
```

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -511,13 +530,13 @@ using  usm_span = usm_subrange_impl<T, std::span<T>>;
...
-template<int call_id = 0, typename T = int, TestDataMode mode = data_in>
+template<int call_id = 0, typename T = int, TestDataMode mode = data_in, typename PolicySelector = AllPolicies>
 struct test_range_algo
 {
-    const int max_n = 10;
+    const int max_n = small_sz;
     void test_view(auto view, auto algo, auto& checker, auto... args)
     {
-        test<T, host_subrange<T>, mode>{max_n}(host_policies(), algo, checker, view, std::identity{}, args...);
+        test<T, host_subrange<T>, mode>{max_n}(PolicySelector{}, algo, checker, view, std::identity{}, args...);
```
SergeyKopienko, line 539: This approach requires to have all implementations of `PolicySelector` types but inside `test` function you still using only it's type check. I think this mean that better to pass `PolicySelector` type not through the instance but through additional template param.

This approach gives us ability to rewrite
```C++
struct AllPolicies{};
struct DeviceAndParPolicies{};
struct DevicePolicy{};
```

as
```C++
struct AllPolicies;
struct DeviceAndParPolicies;
struct DevicePolicy;
```

dmitriy-sobolev:
Closing this PR to focus on https://github.com/uxlfoundation/oneDPL/pull/2081.

PR #2081, opened 2025-02-25
Range tests: scale sizes according to a policy

dmitriy-sobolev:
An approach alternative to #2078 (suggested in https://github.com/uxlfoundation/oneDPL/pull/2078#discussion_r1970040064)

### DIFF
--- a/test/parallel_api/ranges/std_ranges_adjacent_find.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_adjacent_find.pass.cpp
@@ -24,7 +24,7 @@ main()
 
     auto adj_find_checker = TEST_PREPARE_CALLABLE(std::ranges::adjacent_find);
 
-    test_range_algo<0>{policy_scaled_sizes}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred);
+    test_range_algo<0>{default_sizes_per_policy}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred);
     test_range_algo<1>{}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred, proj);
     test_range_algo<2, P2>{}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred, &P2::x);
     test_range_algo<3, P2>{}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred, &P2::proj);
--- a/test/parallel_api/ranges/std_ranges_adjacent_find.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_adjacent_find.pass.cpp
@@ -24,7 +24,7 @@ main()
 
     auto adj_find_checker = TEST_PREPARE_CALLABLE(std::ranges::adjacent_find);
 
-    test_range_algo<0>{big_sz}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred);
+    test_range_algo<0>{policy_scaled_sizes}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred);
     test_range_algo<1>{}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred, proj);
     test_range_algo<2, P2>{}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred, &P2::x);
     test_range_algo<3, P2>{}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred, &P2::proj);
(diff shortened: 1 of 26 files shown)


danhoeflinger:
I do think I like this version better, it does things more concisely in the tests themselves, without having to call the same API multiple times.  

This PR also touches all the ranges tests, which is nice (though perhaps is something which could be done for the other strategy as well.

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -34,11 +34,15 @@ static_assert(ONEDPL_HAS_RANGE_ALGORITHMS >= 202409L);
 #include <ranges>
 #include <algorithm>
 #include <memory>
+#include <array>
 
 namespace test_std_ranges
 {
 
-inline constexpr std::size_t big_sz = (1<<25) + 10; //32M
+inline constexpr int big_sz = (1<<24) + 10; //16M
```
MikeDvorskiy, line 42: It was 32M before.... What's a reason to changes it to 16M?
Don't we whether make test coverage worse?

dmitriy-sobolev: Testing 32M elements with device policies also is too much. There are no specializations for more than 16M elements.

test/parallel_api/ranges/std_ranges_adjacent_find.pass.cpp
```diff
@@ -24,7 +24,7 @@ main()
 
     auto adj_find_checker = TEST_PREPARE_CALLABLE(std::ranges::adjacent_find);
 
-    test_range_algo<0>{big_sz}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred);
+    test_range_algo<0>{big_sizes_per_policy}(dpl_ranges::adjacent_find, adj_find_checker, binary_pred);
```
MikeDvorskiy, line 27: I propose keeping `big_sz` name. In that case we have got only one ".h' modified.

dmitriy-sobolev: `big_sz` is already used in `std_ranges_transform_iota_sycl.pass.cpp` as `int`. Let's keep `big_sizes_per_policy` name for an array. The change is already done and big_sz for an array would be confising.

danhoeflinger: I disagree.  Why should we prioritize minimizing change in a PR at the cost of quality of code (confusing mis-named variable).

The change is straightforward, and already implemented and basically reviewed.

dmitriy-sobolev: > big_sz as a name of an array would be confusing.

This one is stronger :)

I prefer a trivial replacement in 25 files to a confusing name.

danhoeflinger: To clarify, I certainly understand minimizing change when that change introduces risk, but I really don't think this does.

SergeyKopienko: But what is self-documented code in this case? Comments instead it?

MikeDvorskiy: 4. If everyone doesnt mind, as I am an author of these testes, I can make the final commit with reverting 25 test files.

danhoeflinger: > 2\. A good comment next to `big_sz` eliminates confusing.

So, should we add a comment to each file? :)  
I don't expect future readers of the code will chase down every constant definition to another file try to understand the code rather than taking it's name at face value.

In the end, its not a hill I will die on, I won't insist.

dmitriy-sobolev: I also do not insist. Mike, you can add this commit

MikeDvorskiy: > So, should we add a comment to each file? :)

in one file, of course., where `big_sz` is defined.

danhoeflinger: I would argue that when we have semantic changes to the tests (which we do here), that it is good for the commit to touch the test's file.  
If I'm root causing a change in the pass/fail status of a test and looking for the change which caused it, I look at the git history of a test file, and perhaps the algorithm and its nice to see the changes which impact the test I'm looking at.

You can still of course find the change by looking at changes to dependencies like where this variable is defined, but this kind of obfuscates a real change we are making to the tests.

That being said, I don't think its worth much more of our time to continue discussion and I'm OK with any decision here.

MikeDvorskiy (dismissed):

MikeDvorskiy (approved):

PR #2082, opened 2025-02-26
Avoid specializations of the same submitters with the some policy type but with different qualifiers

SergeyKopienko:
@mmichel11, @danhoeflinger  In this draft PR I describes my ideas how to avoid that.

I am just started, so it's not final state now.

Also I want to fix in this PR the issue https://github.com/uxlfoundation/oneDPL/issues/2083


SergeyKopienko:
Closed because the approach from https://github.com/uxlfoundation/oneDPL/pull/2093 look more preferable

PR #2084, opened 2025-02-27
Remove direct usage of `kernel_name` type

SergeyKopienko:
In this PR we remove direct usage of `kernel_name` type and using `oneapi::dpl::__internal::__policy_kernel_name` trait instead.


danhoeflinger (approved):
LGTM, thanks Sergey!

Issue #1945, opened 2024-11-21
Fix tests relying on assert (release builds)


dmitriy-sobolev:
**Describe the Bug:**
Some tests rely on `assert` for issuing errors, but there is no any handling of `NDEBUG` e.g. on CMake side. That may result in a worse test coverage for release builds.

**To Reproduce:**
The steps to reproduce the behavior (Windows as an example):

```bash

# Comment line in test_config.h to trigger an error:
// #define _PSTL_TEST_COMPLEX_OP_BROKEN (_MSVC_STL_VERSION && _MSVC_STL_VERSION <= _PSTL_TEST_LATEST_MSVC_STL_VERSION)

# The error is detected in the debug build
cmake -GNinja -DCMAKE_CXX_COMPILER=icx -DCMAKE_BUILD_TYPE=debug -DONEDPL_DEVICE_TYPE=CPU path\to\oneDPL
ninja run-asin.pass
1/1 Test #468: asin.pass ........................Exit code 0xc0000409
***Exception:   0.99 sec
Assertion failed: std::isinf(r.imag()), file C:\repos\oneDPL\test\xpu_api\numerics\complex.number\complex.transcendentals\asin.pass.cpp, line 89

# The error is overlooked in the release build
cmake -GNinja -DCMAKE_CXX_COMPILER=icx -DCMAKE_BUILD_TYPE=release -DONEDPL_DEVICE_TYPE=CPU C:\repos\oneDPL
ninja run-asin.pass
1/1 Test #468: asin.pass ........................   Passed    0.84 sec

# Adding "#undef NDEBUG" to the top of asin.pass.cpp results in the triggering of the assertion in the release build as well.

```

**Expected Behavior:**
The tests are equally covered in the release and debug builds.

MikeDvorskiy:
@dmitriy-sobolev , can we add `"#undef NDEBUG"` on the "cmake level" for the all oneDPL tests?

dmitriy-sobolev:
@MikeDvorskiy, see #2089. I added `ReleaseWithAsserts` configuration, to keep `Release` as it is.

danhoeflinger:
@dmitriy-sobolev  can we mark this as resolved or do we intend to make other changes to get away from assert in the tests?

dmitriy-sobolev:
We still need these changes. Having a couple of ReleaseWithAsserts configurations is not enough - we need to have the same coverage for the tests currently relying on asserts compared to other tests.

PR #2089, opened 2025-03-01
Add RelWithAsserts cmake config for testing

dmitriy-sobolev:
Addresses #1945 by adding `RelWithAsserts` cmake configuration.

### Technical details 

https://cmake.org/cmake/help/v3.31/manual/cmake-buildsystem.7.html#default-and-custom-configurations

Each configuration defines a set of compiler and linker flag variables for the language in use. These variables follow the convention `CMAKE_<LANG>_FLAGS_<CONFIG>`, where `<CONFIG>` is always the uppercase configuration name. When defining a custom configuration type, make sure these variables are set appropriately, typically as cache variables.

Additionally, I've noticed more configuration specific variables: `CMAKE_EXE_LINKER_FLAGS`, `CMAKE_MODULE_LINKER_FLAGS` and `CMAKE_SHARED_LINKER_FLAGS` investigating cache after ninja run (a single-config generator).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -69,6 +69,18 @@ if (FIND_GXX_EXE)
     execute_process(COMMAND ${FIND_GXX_EXE} -dumpfullversion OUTPUT_VARIABLE _onedpl_gxx_version)
 endif()
 
+# RelWithAsserts has the same flags as Release, but removes NDEBUG.
+# This build type can be enabled via CMAKE_BUILD_TYPE=RelWithAsserts.
+# It is useful to fully cover tests relying on asserts, such as those exported from libc++, or Tested Standard C++ API tests.
+foreach(FLAGS CMAKE_CXX_FLAGS CMAKE_EXE_LINKER_FLAGS CMAKE_MODULE_LINKER_FLAGS CMAKE_SHARED_LINKER_FLAGS
+        CMAKE_STATIC_LINKER_FLAGS)
+    if (FLAGS STREQUAL "CMAKE_CXX_FLAGS")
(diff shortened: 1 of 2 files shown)


MikeDvorskiy (approved):
LGTM

CMakeLists.txt
```diff
@@ -69,6 +69,18 @@ if (FIND_GXX_EXE)
     execute_process(COMMAND ${FIND_GXX_EXE} -dumpfullversion OUTPUT_VARIABLE _onedpl_gxx_version)
 endif()
 
+# RelWithAsserts has the same flags as Release, but removes NDEBUG.
+# This build type can be enabled via CMAKE_BUILD_TYPE=RelWithAsserts.
+# It is useful to fully cover tests relying on asserts, such as those exported from libc++, or Tested Standard C++ API tests.
+foreach(FLAGS CMAKE_CXX_FLAGS CMAKE_EXE_LINKER_FLAGS CMAKE_MODULE_LINKER_FLAGS CMAKE_SHARED_LINKER_FLAGS
+        CMAKE_STATIC_LINKER_FLAGS)
+    if (FLAGS STREQUAL "CMAKE_CXX_FLAGS")
+        string(REGEX REPLACE "(^| )[/-]DNDEBUG( |$)" " " BORROWED_FLAGS "${${FLAGS}_RELEASE}")
```
danhoeflinger, line 78: It doesn't really matter I suppose, but could we just use the regex replace for all cases?

danhoeflinger (approved):
LGTM, my comment is optional.

CMakeLists.txt
dmitriy-sobolev: What is here is more precise, but it is more descriptive. I think both ways work. I am going to leave it as is.

PR #2090, opened 2025-03-03
Replace sycl::min with std::min in host code

dmitriy-sobolev:
Using sycl::min in this context leads to " error: no matching function for call to 'min'" with `acpp` for some reason. Let's switch to std::min for host code. Using sycl::min is not necessary.


SergeyKopienko (approved):
LGTM

PR #2091, opened 2025-03-03
Fix constructors of two `__transform_if_`-functors

SergeyKopienko:
In this PR we fix two functors: `__transform_if_unary_functor` and `__transform_if_binary_functor` :
1) remove `&&` params qualifiers in constructor;
2) using `std::move` instead of `std::forward`.

### DIFF
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -303,7 +303,7 @@ class __transform_if_unary_functor
 
   public:
     explicit __transform_if_unary_functor(_UnaryOper __op, _UnaryPred __pred)
-        : _M_oper(std::move<_UnaryOper>(__op)), _M_pred(std::move(__pred))
+        : _M_oper(std::move(__op)), _M_pred(std::move(__pred))
     {
     }
 
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -302,8 +302,8 @@ class __transform_if_unary_functor
     mutable _UnaryPred _M_pred;
 
   public:
-    explicit __transform_if_unary_functor(_UnaryOper&& __op, _UnaryPred&& __pred)
-        : _M_oper(::std::forward<_UnaryOper>(__op)), _M_pred(::std::forward<_UnaryPred>(__pred))
+    explicit __transform_if_unary_functor(_UnaryOper __op, _UnaryPred __pred)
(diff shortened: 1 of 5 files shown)


MikeDvorskiy:
@SergeyKopienko , you are writing " we fix..."
Which error we are fixing?

include/oneapi/dpl/pstl/utils.h
```diff
@@ -302,8 +302,8 @@ class __transform_if_unary_functor
     mutable _UnaryPred _M_pred;
 
   public:
-    explicit __transform_if_unary_functor(_UnaryOper&& __op, _UnaryPred&& __pred)
-        : _M_oper(::std::forward<_UnaryOper>(__op)), _M_pred(::std::forward<_UnaryPred>(__pred))
+    explicit __transform_if_unary_functor(_UnaryOper __op, _UnaryPred __pred)
```
danhoeflinger, line 305: I agree this is incorrect as it was written before this PR.
Would a better fix be to make this constructor templated with its own local universal references, and keep the forward?

We already explicitly use explicit template parameters whenever we instantiate these types.

SergeyKopienko: > I agree this is incorrect as it was written before this PR. Would a better fix be to make this constructor templated with its own local universal references, and keep the forward?
> 
> We already explicitly use explicit template parameters whenever we instantiate these types.

We discussed with @rarutyun about this implementations and he proposed do not use `&&` in constructor params of these functors.

danhoeflinger:
It seems this is causing some errors in the testing.  I think if we remove the rvalue references, we should just pass these functors by copy.  Also, we can probably remove the `std::move` in their invocations.

SergeyKopienko:
> It seems this is causing some errors in the testing. I think if we remove the rvalue references, we should just pass these functors by copy. Also, we can probably remove the `std::move` in their invocations.

No. it was just compile error, I think:
```C++
    explicit __transform_if_unary_functor(_UnaryOper __op, _UnaryPred __pred)
        : _M_oper(std::move<_UnaryOper>(__op)), _M_pred(std::move(__pred))
    {
    }
```
We should simple delete `<_UnaryOper>` here.

SergeyKopienko:
> Also, we can probably remove the `std::move` in their invocations.

Done.

danhoeflinger (approved):
LGTM, thanks for fixing this.

MikeDvorskiy (approved):
LGTM

Issue #2083, opened 2025-02-26
Some `__kernel_name_generator` usages doesn't depends from  `_ExecutionPolicy` type


SergeyKopienko:
**Describe the Bug:**
Some `__kernel_name_generator` usages doesn't includes  `_ExecutionPolicy` type,
but the code where it's implemented contains `_ExecutionPolicy` inside template params.
This situation allows to write the code with the different `_ExecutionPolicy` types but to have the same KernelName inside - it's will generate an error in unnamed lambdas switched off.

**To Reproduce:**
No reproducers.

**Expected Behavior:**
Probably every `__kernel_name_generator` should depends on  `_ExecutionPolicy` type if it is used for code specialization.

**Additional Context:**

Examples where  `_ExecutionPolicy` type is not used in `__kernel_name_generator`  :
https://github.com/uxlfoundation/oneDPL/blob/84da6aca02ed5fce710529a5b358d98fb4df90be/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h#L242
https://github.com/uxlfoundation/oneDPL/blob/84da6aca02ed5fce710529a5b358d98fb4df90be/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h#L244
https://github.com/uxlfoundation/oneDPL/blob/84da6aca02ed5fce710529a5b358d98fb4df90be/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h#L321

SergeyKopienko:
During implementation of https://github.com/uxlfoundation/oneDPL/pull/2093 with fix for this issue we found another option: to use `__kernel_name_generator` without `_ExecutionPolicy` but also changing the type of `_ExecutionPolicy` param in submitters (and one level up functions - in submitter callers) from `_ExecutionPolicy&&` to `const _ExecutionPolicy&`.
This approach will give us only one specialization for every execution policy.

SergeyKopienko:
@rarutyun could you please comment this issue? Is it really an error?

rarutyun:
This issue should be considered together with #2041. While they look different they are tightly connected. Furthermore, #2041 is the first level question, and this issue is a second level question.

Issue #2041, opened 2025-02-03
Reconsider the use of forwarding references throughout all SYCL backend kernel submitters


mmichel11:
**Describe the Bug:**
Almost all of oneDPL's SYCL backend kernel submitters rely on forwarding references of the execution policy and any ranges used within the kernel. Here is an example from our reduce submitter:
```cpp
template <typename _ExecutionPolicy, typename _Size, typename _ReduceOp, typename _TransformOp, typename... _Ranges>
auto
operator()(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, const _Size __n,
    const _Size __work_group_size, const _Size __iters_per_work_item, _ReduceOp __reduce_op,
    __transform_op,
    const __result_and_scratch_storage<_ExecutionPolicy, _Tp>& __scratch_container,
    _Ranges&&... __rngs) ...
```
With these forwarding references, different cv / ref qualifiers on the same execution policy will lead to separate function template instantiations by the compiler. Similarly, the same can be said for the ranges.

With these separate submitter function template instantiations, a new kernel is compiled **per** instantiation. However, `_ExecutionPolicy&& __exec` is only used for kernel submission, so logically no new kernel is needed. Similarly, for ranges, these will be passed in as lightweight views, and we may be able to just accept these by-value.

If the user is compiling with unnamed lambda naming and the submitter is using our internal "kernel name provider", then I think there is no risk of a compilation error here. However, we may compile more kernels than what is logically necessary depending on how the user passes policies / ranges. This will lead to long JIT / AOT compile times.

If the user is trying to name kernels themselves or the underlying submitter is using our "kernel compiler" internal API for kernel bundles, then we may see compilation errors regarding duplicate kernel names.

**To Reproduce:**
The following results in compilation error using no unnamed lambdas (icpx 2025.0.0 and oneDPL 2022.7.0) despite the same underlying policy being used:
```cpp
// icpx -fsycl -fno-sycl-unnamed-lambda reduce.cpp
// We will see a compilation error.
#include <oneapi/dpl/execution>
#include <oneapi/dpl/algorithm>

#include <iostream>

int main()
{
    sycl::queue q;
    int n = 10;
    int *ptr = sycl::malloc_shared<int>(n, q);
    q.fill(ptr, 1, n).wait();

    oneapi::dpl::execution::device_policy<class kernel> policy{q};

    auto res1 = oneapi::dpl::reduce(policy, ptr, ptr + n);
[...]

PR #2093, opened 2025-03-03
Avoid specializations of the same submitters with the some policy type but with different type qualifiers (`l-value`, `r-value`)

SergeyKopienko:
In this PR we fix the issue https://github.com/uxlfoundation/oneDPL/issues/2083 and https://github.com/uxlfoundation/oneDPL/issues/2041

## The problem statement.
Usually, `oneDPL` submitters which do `sycl::queue::submit()` calls are implemented in three ways:
- `operator()` implementation inside some `class` / `struct`;
- `submit()` function implementation inside some `class` / `struct`;
- some free function.

In all cases, these functions are template functions parametrized by `_ExecutioPolicy` type. Also they have `_ExecutioPolicy&& __exec` in the list of functions parameters.
This approach means that we may have two specializations of these functions:
- one specialization for the case when `__exec` is `r-value`;
- another specialization for other cases.

### The problems of this approach which already implemented in `oneDPL`:
- we should generate unique Kernel-names using not only `_ExecutionPolicy` type but also their qualifiers (`l-value`, `r-value`);
- we increase the amount of sycl-kernel code and their compile-time;
- we increase the amount of host code in compile-time.

## The solution
As solution in this PR proposed the idea of @akukanov - to declare `_ExecutionPolicy` in submitters and on one upper level as `const _ExecutionPolicy&` (const reference).
This approach fixes all point of the problem described below.

### Some key changes in the code
* the usage of `_ExecutionPolicy` is not required inside `__kernel_name_generator` template parameters anymore:
-- before:
```C++
    template <typename... _Name>
    using _SegScanPrefixPhase = __seg_scan_prefix_kernel<__is_inclusive, _Name...>;

    template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3,
              typename _BinaryPredicate, typename _BinaryOperator, typename _T>
    void
    operator()(_BackendTag, _ExecutionPolicy&& __exec, _Range1&& __keys, _Range2&& __values, _Range3&& __out_values,
               _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init, _T __identity)
    {

        using _SegScanWgKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
            _SegScanWgPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _BinaryPredicate,
            _BinaryOperator>;

        // ...
    }
```
-- now:
```C++
    template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3,
              typename _BinaryPredicate, typename _BinaryOperator, typename _T>
    void
    operator()(_BackendTag, const _ExecutionPolicy& __exec, _Range1&& __keys, _Range2&& __values, _Range3&& __out_values,
               _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init, _T __identity)
    {

        // We should avoid using _ExecutionPolicy in __kernel_name_generator template params
        // because we always specialize this operator() calls only by _ExecutionPolicy as "const reference".
        // So, from this template param point of view, only one specialization is possible.
        using _SegScanWgKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
            _SegScanWgPhase, _CustomName, _Range1, _Range2, _Range3, _BinaryPredicate, _BinaryOperator>;

        // ...
    }
```

Now we able to detect a lot of errors as described in https://github.com/uxlfoundation/oneDPL/issues/2041 - this test coverage implemented in the PR https://github.com/uxlfoundation/oneDPL/pull/2102

Another approach has been prepared in https://github.com/uxlfoundation/oneDPL/pull/2082

## Please review this PR under switched off option "File changed - Gear button - Hide whitespace"

### DIFF
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -108,8 +108,9 @@ struct __sycl_scan_by_segment_impl
     template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3,
               typename _BinaryPredicate, typename _BinaryOperator, typename _T>
     void
-    operator()(_BackendTag, const _ExecutionPolicy& __exec, _Range1&& __keys, _Range2&& __values, _Range3&& __out_values,
-               _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init, _T __identity)
+    operator()(_BackendTag, const _ExecutionPolicy& __exec, _Range1&& __keys, _Range2&& __values,
+               _Range3&& __out_values, _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init,
+               _T __identity)
     {
         using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
 
@@ -393,9 +394,9 @@ __scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, const Polic
 
     constexpr iter_value_t identity = unseq_backend::__known_identity<Operator, iter_value_t>;
 
-    __sycl_scan_by_segment_impl<Inclusive::value>()(_BackendTag{}, policy, key_buf.all_view(),
-                                                    value_buf.all_view(), value_output_buf.all_view(), binary_pred,
-                                                    binary_op, init, identity);
+    __sycl_scan_by_segment_impl<Inclusive::value>()(_BackendTag{}, policy, key_buf.all_view(), value_buf.all_view(),
+                                                    value_output_buf.all_view(), binary_pred, binary_op, init,
+                                                    identity);
     return result + n;
 }
 
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -117,7 +117,7 @@ struct __sycl_scan_by_segment_impl
         // We should avoid using _ExecutionPolicy in __kernel_name_generator template params
         // because we always specialize this operator() calls only by _ExecutionPolicy as "const reference".
         // So, from this template param point of view, only one specialization is possible per concrete _ExecutionPolicy type.
-        // _ExecutionPolicy type information is encoded in _CustomName to distinguish between concrete policy types.
+        // _ExecutionPolicy type information is embedded in _CustomName to distinguish between concrete policy types.
         using _SegScanWgKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
             _SegScanWgPhase, _CustomName, _Range1, _Range2, _Range3, _BinaryPredicate, _BinaryOperator>;
         using _SegScanPrefixKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -116,7 +116,8 @@ struct __sycl_scan_by_segment_impl
 
         // We should avoid using _ExecutionPolicy in __kernel_name_generator template params
         // because we always specialize this operator() calls only by _ExecutionPolicy as "const reference".
-        // So, from this template param point of view, only one specialization is possible.
+        // So, from this template param point of view, only one specialization is possible per concrete _ExecutionPolicy type.
+        // _ExecutionPolicy type information is encoded in _CustomName to distinguish between concrete policy types.
         using _SegScanWgKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
             _SegScanWgPhase, _CustomName, _Range1, _Range2, _Range3, _BinaryPredicate, _BinaryOperator>;
         using _SegScanPrefixKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -370,7 +370,7 @@ struct __sycl_scan_by_segment_impl
 template <typename _BackendTag, typename Policy, typename InputIterator1, typename InputIterator2,
           typename OutputIterator, typename T, typename BinaryPredicate, typename Operator, typename Inclusive>
 OutputIterator
-__scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIterator1 first1,
+__scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, const Policy& policy, InputIterator1 first1,
                               InputIterator1 last1, InputIterator2 first2, OutputIterator result, T init,
                               BinaryPredicate binary_pred, Operator binary_op, Inclusive)
 {
@@ -393,7 +393,7 @@ __scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, Policy&& po
 
     constexpr iter_value_t identity = unseq_backend::__known_identity<Operator, iter_value_t>;
 
-    __sycl_scan_by_segment_impl<Inclusive::value>()(_BackendTag{}, ::std::forward<Policy>(policy), key_buf.all_view(),
+    __sycl_scan_by_segment_impl<Inclusive::value>()(_BackendTag{}, policy, key_buf.all_view(),
                                                     value_buf.all_view(), value_output_buf.all_view(), binary_pred,
                                                     binary_op, init, identity);
     return result + n;
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1107,7 +1107,7 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
     _NoOpFunctor __get_data_op;
 
     return __parallel_transform_scan_base(
-        __backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__in_rng),
+        __backend_tag, __exec, std::forward<_Range1>(__in_rng),
         std::forward<_Range2>(__out_rng), __init,
         // local scan
         unseq_backend::__scan<_Inclusive, _ExecutionPolicy, _BinaryOperation, _UnaryFunctor, _Assigner, _Assigner,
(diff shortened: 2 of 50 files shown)


SergeyKopienko:
About CLang format for this PR:
 - I propose to not apply some changes:
```C++
 {
@@ -503,7 +507,8 @@ __copy_kernel_for_radix_sort(::std::size_t __segments, const ::std::size_t __ele
 template <typename _KernelName, ::std::uint32_t __radix_bits, bool __is_ascending, __peer_prefix_algo _PeerAlgo,
           typename _ExecutionPolicy, typename _InRange, typename _OutRange, typename _OffsetBuf, typename _Proj
 #if _ONEDPL_COMPILE_KERNEL
-          , typename _Kernel
+          ,
+          typename _Kernel
 #endif
           >
 sycl::event
@@ -511,7 +516,8 @@ __radix_sort_reorder_submit(const _ExecutionPolicy& __exec, ::std::size_t __segm
                             ::std::uint32_t __radix_offset, _InRange&& __input_rng, _OutRange&& __output_rng,
                             _OffsetBuf& __offset_buf, sycl::event __dependency_event, _Proj __proj
 #if _ONEDPL_COMPILE_KERNEL
-                            , _Kernel& __kernel
+                            ,
+                            _Kernel& __kernel
 #endif
 )
```

mmichel11:
I like the approach here, and it should also reduce the number of kernels for the unnamed lambda case. One issue I see in the case of `reduce` is that we still embed the forward reference `_ExecutionPolicy` into the our walk functor here:
https://github.com/uxlfoundation/oneDPL/blob/70a57d85d4d1ff70c640a72d5fb4bbb032f3faaa/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h#L83 and
https://github.com/uxlfoundation/oneDPL/blob/70a57d85d4d1ff70c640a72d5fb4bbb032f3faaa/include/oneapi/dpl/pstl/hetero/numeric_impl_hetero.h#L50
which leads to unnecessary kernel compilation with the reproducer in https://github.com/uxlfoundation/oneDPL/issues/2041.

I think the policy can either be decayed or just removed from `walk_n` (I am not sure if it has a purpose, but it appears unused). I checked the first option and doing so only leads to the compilation of a single small submitter kernel for this case.

include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -371,7 +373,7 @@ struct __sycl_scan_by_segment_impl
 template <typename _BackendTag, typename Policy, typename InputIterator1, typename InputIterator2,
           typename OutputIterator, typename T, typename BinaryPredicate, typename Operator, typename Inclusive>
 OutputIterator
-__scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIterator1 first1,
+__scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, const Policy& policy, InputIterator1 first1,
```
danhoeflinger, line 376: Why do we need to change this signature as well?  A const lvalue ref can accept whatever policy type is forwarded along from this function.

include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -108,17 +108,19 @@ struct __sycl_scan_by_segment_impl
...
-               _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init, _T __identity)
+    operator()(_BackendTag, const _ExecutionPolicy& __exec, _Range1&& __keys, _Range2&& __values,
+               _Range3&& __out_values, _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init,
+               _T __identity)
     {
         using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
 
+        // We should avoid using _ExecutionPolicy in __kernel_name_generator template params
+        // because we always specialize this operator() calls only by _ExecutionPolicy as "const reference".
+        // So, from this template param point of view, only one specialization is possible.
```
danhoeflinger, line 119: While it is true that we wont have an rvalue and a lvalue reference version of any specific `_ExecutionPolicy`, it is still possible that we may differently typed policies at this level, right?

` device_policy` vs `fpga_policy`, or even different `_KernelName`.  Before this change, these were embedded in the kernel name, should they still be?

It seems like perhaps yes, since `_ExecutionPolicy` is used in the `__buffer` and therefore accessors used in the kernel?  

Still thinking about if policy is necessary, but I think it may be.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -232,13 +232,17 @@ struct __parallel_scan_submitter;
...
     auto
-    operator()(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2, _InitType __init,
+    operator()(const _ExecutionPolicy& __exec, _Range1&& __rng1, _Range2&& __rng2, _InitType __init,
                _LocalScan __local_scan, _GroupScan __group_scan, _GlobalScan __global_scan) const
     {
         using _Type = typename _InitType::__value_type;
+
+        // We should avoid using _ExecutionPolicy in __kernel_name_generator template params
+        // because we always specialize this operator() calls only by _ExecutionPolicy as "const reference".
+        // So, from this template param point of view, only one specialization is possible.
```
danhoeflinger, line 245: Same comment as above, we don't have different typed based on reference flavors, but the policy type itself is not fixed.

In this case, we use `__result_and_scratch_storage<_ExecutionPolicy, _Type>` in our kernel, so it seem like a difference in this type would change the kernel generated, and a different name is required.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -554,7 +558,7 @@ struct __parallel_copy_if_static_single_group_submitter<_Size, _ElemsPerItem, _W
 template <typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _UnaryOperation, typename _InitType,
           typename _BinaryOperation, typename _Inclusive>
 auto
-__parallel_transform_scan_single_group(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec,
```
danhoeflinger, line 557: Same question about why this change is necessary.  `__parallel_transform_scan_static_single_group_submitter` with a forwarded policy should work and `const _ExecutionPolicy&` should bind to forwarded policy type.

danhoeflinger (commented):
I believe I will have the same questions throughout the PR.  I may be incorrect about them, but before going further, we should discuss them.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: I changed this signature because the initial idea was to change signature of submitters and on one level up (their callers).
This `__scan_by_segment_impl_common` function calls some submitter `__sycl_scan_by_segment_impl` :
```C++
    __sycl_scan_by_segment_impl<Inclusive::value>()(_BackendTag{}, policy, key_buf.all_view(), value_buf.all_view(),
                                                    value_output_buf.all_view(), binary_pred, binary_op, init,
                                                    identity);
```

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: Yes, exactly, in this code we may have `_ExecutionPolicy` defined as `device_policy` or `fpga_policy`.
And they has different `KernelName` inside, we return in through
```C++
```
For `device_policy` - it's defined as `DefaultKernelName` and for `fpga_policy` it's defiend as `DefaultKernelNameFPGA` (by default).

So owe still using policy `kernel_name` here:
```C++
        using _SegScanWgKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
            _SegScanWgPhase, _CustomName, _Range1, _Range2, _Range3, _BinaryPredicate, _BinaryOperator>;
        using _SegScanPrefixKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
            _SegScanPrefixPhase, _CustomName, _Range1, _Range2, _Range3, _BinaryPredicate, _BinaryOperator>;
```
But the usage of `_ExecutionPolicy` now really is not required because the different policies gives us different `_CustomName`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: The same - this code calls some submitter `__parallel_transform_scan_static_single_group_submitter`

SergeyKopienko:
> using _Functor = unseq_backend::walk_n<_ExecutionPolicy, _UnaryOperation>;

Thanks, @mmichel11, you present here very good example!
And really we able to fix this case if we replace the code
```C++
using _Functor = unseq_backend::walk_n<_ExecutionPolicy, _UnaryOperation>;
```
by
```C++
using _Functor = unseq_backend::walk_n<std::decay_t<_ExecutionPolicy>, _UnaryOperation>;
```

But I think that we are unable to fix all places like that analytically, reading the code.
And what I propose as common solution as fix for ALL cases like that: I propose to replace `_ExecutionPolicy&&` in parameters by `const _ExecutionPolicy&` in all internal `oneDPL` namespaces:
- in `oneapi::dpl::internal` namespace;
- in `oneapi::dpl::__internal` namespace;
- in `oneapi::dpl::__internal::__ranges` namespace;
- in `oneapi::dpl::__serial_backend` namespace;
- in `oneapi::dpl::__tbb_backend` namespace;
- in `oneapi::dpl::__omp_backend` namespace;
- in `oneapi::dpl::__par_backend_hetero` namespace.

But on `oneDPL` API level, in public namespaces like `oneapi::dpl`, `oneapi::dpl::experimental` we save `_ExecutionPolicy&&` in parameters without changes.

I am already did this exercise in the separate PR https://github.com/uxlfoundation/oneDPL/pull/2101 for merge into this PR.

include/oneapi/dpl/internal/scan_by_segment_impl.h
danhoeflinger: Aha, yes.  Thanks for the explanation, that makes sense to me now.

danhoeflinger, line 119:
```suggestion
        // We should avoid using _ExecutionPolicy in __kernel_name_generator template params
        // because we always specialize this operator() calls only by _ExecutionPolicy as "const reference".
        // So, from this template param point of view, only one specialization is possible per concrete _ExecutionPolicy type.
        // _ExecutionPolicy type information is embedded in _CustomName to distinguish between concrete policy types.
```

To avoid confusion similar to what I had question about, perhaps we can add to this comment something like what I have above.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: Please see my comment at https://github.com/uxlfoundation/oneDPL/pull/2093#issuecomment-2700963832

SergeyKopienko:
> > I am already did this exercise in the separate PR #2101 for merge into this PR.
> 
> @SergeyKopienko Is the scenario motivating the #2101 PR only due to similar instances where `_ExecutionPolicy` is passed as a class template to some functor?
> 
> We have many instances of this particularly in `unseq_backend_sycl.h`. I think the higher level question to first consider is why do we need `_ExecutionPolicy` as a class template for these functors if it is never used. Is it relevant for some kernel naming related reason? If we think this is not needed now, then the proper solution may be to keep the approach in this PR where we use the const l-value reference only one level up and remove these unneeded class template parameters.

Implementation of this your approach in the current PR technically will mean a lot of changes and big chance to make some error in these functor fixing. So I proposed some approach how to avoid these problems more-less automatically - simple changing types in internal oneDPL APIs.

PR #2096, opened 2025-03-04
[oneDPL][ranges] A fix for ranges::transform algo - it requires all input and output ranges to be sized.

MikeDvorskiy:
A fix according to p3179r4.
It requires all input and output ranges to be sized for ranges::transform algo.


include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -102,32 +102,28 @@ struct __transform_fn
...
 
         using _Size = std::common_type_t<oneapi::dpl::__internal::__range_size_t<_R1>,
             oneapi::dpl::__internal::__range_size_t<_R2>, std::ranges::range_size_t<_OutRange>>;
-        _Size __size = std::ranges::size(__out_r);
-        if constexpr(std::ranges::sized_range<_R1>)
-            __size = std::ranges::min(__size, (_Size)std::ranges::size(__r1));
-        if constexpr(std::ranges::sized_range<_R2>)
-            __size = std::ranges::min(__size, (_Size)std::ranges::size(__r2));
+        const _Size __size = std::ranges::min({(_Size)std::ranges::size(__r1), (_Size)std::ranges::size(__r2),
+            (_Size)std::ranges::size(__out_r)});
```
dmitriy-sobolev, line 121: Could you make the same adjustment in `std_ranges_transform.pass.cpp`? https://github.com/uxlfoundation/oneDPL/blob/f49feb459eb0fc19524a2cd0b5789e5b96575b5e/test/parallel_api/ranges/std_ranges_transform.pass.cpp#L51-L57

Could you also use `take_view` instead of a subrange in `transform_binary_checker` like it is done below in the library?

dmitriy-sobolev (commented):
Looks good to me leaving aside a suggestion for `std_ranges_transform.pass`.

Let me leave a link to the paper and a relevant section for other reviewers and my future self: https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3179r4.html#require_bounded_ranges

> In the case of two input ranges or sequences, for a few algorithms - namely, mismatch, equal, and transform - it could be sufficient for just one range to be bounded and the other assumed to have at least as many elements as the bounded one.
...
However, SG9 [decided](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p3179r4.html#sg9_wroclaw_2024) to require sized_range for all inputs, with the plan to relax these constraints for transform once there is a way to statically detect infinite ranges like views::repeat (as opposed to finite unsized ranges, such as null terminated strings).

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
MikeDvorskiy: done

test/parallel_api/ranges/std_ranges_transform.pass.cpp
```diff
@@ -50,15 +50,12 @@ main()
...
             std::ranges::range_size_t<decltype(r_out)>>;
-        Size size = std::ranges::size(r_out);
-        if constexpr(std::ranges::sized_range<decltype(r_1)>)
-            size = std::ranges::min(size, (Size)std::ranges::size(r_1));
-        if constexpr(std::ranges::sized_range<decltype(r_2)>)
-            size = std::ranges::min(size, (Size)std::ranges::size(r_2));
+        Size size = std::ranges::min({(Size)std::ranges::size(r_1), (Size)std::ranges::size(r_2), 
+                                      (Size)std::ranges::size(r_out)});
 
         auto res = std::ranges::transform(std::ranges::subrange(std::ranges::begin(r_1), std::ranges::begin(r_1) + size),
```
dmitriy-sobolev, line 56: The first subrange is still in the test. Could you replace it with take view as well?

dmitriy-sobolev (dismissed):
LGTM.

This PR changes public API, though.
I would vote for merging it as is, because:
 - It makes the algorithm compliant with the recent developments in C++ standard. 
 - This algorithm is new (it's been in oneDPL for 2 releases), which lowers the risk.

However, lets wait for the opinion of others, as we discussed separately.

akukanov:
Unfortunately, our specification diverges from recent P3179 updates in this regard. Also in the long run it's likely that there will be a variation of `||` logic when the second range is infinite.

So I suggest not to proceed with this patch immediately and discuss what should be our best way forward. The same applies to `equal` and possibly `mismatch`.

PR #2099, opened 2025-03-04
Reduce the number of the generated KT tests

dmitriy-sobolev:
These tests consume disproportionate amount of time.
 
esimd radix sort: ~50 cases, 49 are random -> 10 cases, 8 are random.
single pass scan: 180 cases -> 5 cases, 2 are random.


test/kt/CMakeLists.txt
```diff
@@ -162,12 +165,18 @@ function(_generate_gpu_scan_tests)
...
     endforeach()
 endfunction()
 
 if (ONEDPL_TEST_ENABLE_KT_SYCL)
     _generate_gpu_scan_tests()
+
+    # Pin some cases to track them
+    _generate_gpu_scan_test("8" "512" "uint32_t" 1000)
+    _generate_gpu_scan_test("4" "256" "float" 1000)
+    _generate_gpu_scan_test("16" "64" "int64_t" 1000)
```
dmitriy-sobolev, line 181: @adamfidel, could you suggest the most important cases?

adamfidel: I think these three are a good representative sample for the scan kernel template.

mmichel11 (approved):
LGTM

PR #2102, opened 2025-03-05
Extend test coverage for different policy qualifiers

SergeyKopienko:
In this PR we extend test coverage to detect compile-time errors in oneDPL code like described at https://github.com/uxlfoundation/oneDPL/issues/2041.
We extend test coverage under `false`-run-time condition check so we only compile additional Kernels without their runs.
This idea was proposed by @danhoeflinger. 

Also in this PR we modified a lot of tests.
The goal - to pass temporary created execution policies inside `oneDPL` algorithms as `l-value` / `r-value` depends on qualifiers of execution policy passed from test environment. This mandatory needed for detect compile errors `...mangled name already used...` inside `oneDPL` sycl-submitters.

## Implementation details.
To achieve this goal we implements two things in tests:
1. Create helper for return temporary execution policy as `l-value` / `r-value` :
```C++
// struct test_policy_container - a container for policy which return saved policy
// as l-value or r-value depends on source policy type qualifiers
template <typename _PolicySource, typename _PolicyNew>
struct test_policy_container
{
    using _PolicyNewDecayed = std::decay_t<_PolicyNew>;

    // Define TestingPolicyType as a type which is either r-value or l-value reference
    using TestingPolicyType = std::conditional_t<
        std::is_reference_v<_PolicySource>,
        std::conditional_t<std::is_rvalue_reference_v<_PolicySource>, _PolicyNewDecayed&&, const _PolicyNewDecayed&>,
        _PolicyNewDecayed>;

    _PolicyNewDecayed __policy_new;
    bool __policy_new_returned = false;

    test_policy_container(_PolicyNewDecayed&& __policy_new)
        : __policy_new(std::move(__policy_new))
    {
    }

    // Delete copy constructor to avoid copying of test_policy_container
    test_policy_container(const test_policy_container&) = delete;
    test_policy_container(test_policy_container&&) = delete;

    // Delete assignment operator to avoid copying of test_policy_container
    test_policy_container& operator=(const test_policy_container&) = delete;
    test_policy_container& operator=(test_policy_container&&) = delete;

    // Return testing policy
    // Attention: this method can be called only once, because it can returns saved policy as r-value too.
    TestingPolicyType get()
    {
        if (__policy_new_returned)
            throw std::logic_error("The second call of test_policy_container::get() function");

        __policy_new_returned = true;

        return static_cast<TestingPolicyType>(__policy_new);
    }
};
```
2. Create new `CLONE_TEST_POLICY`, `CLONE_TEST_POLICY_IDX` and `CLONE_TEST_POLICY_NAME` macros to simplify creation and passing into calling algorithm new temporary execution policy as `l-value` / `r-value` :
```C++
#define CLONE_TEST_POLICY(policy_src)                                                                                  \
        TestUtils::test_policy_container<                                                                              \
            decltype(policy_src),                                                                                      \
            std::decay_t<decltype(policy_src)>                                                                         \
        >(                                                                                                             \
            std::decay_t<decltype(policy_src)>(policy_src)                                                             \
         ).get()

// Create copy of hetero policy with new name (appends idx) and pass copied instance as l-value / r-value
// depends on qualifiers of source policy type.
// The source policy state is not changed, so it can be used in the same test
// ATTENTION: new Kernel name generation depends on TEST_EXPLICIT_KERNEL_NAMES macro state
#    define CLONE_TEST_POLICY_IDX(policy_src, idx)                                                                     \
        TestUtils::test_policy_container<                                                                              \
            decltype(policy_src),                                                                                      \
            decltype(TestUtils::make_new_policy<TestUtils::new_kernel_name<decltype(policy_src), idx>>(policy_src))    \
        >(                                                                                                             \
            TestUtils::make_new_policy<TestUtils::new_kernel_name<decltype(policy_src), idx>>(policy_src)              \
         ).get()

// Create copy of hetero policy with new name and pass copied instance as l-value / r-value
// depends on qualifiers of source policy type.
// The source policy state is not changed, so it can be used in the same test.
#    define CLONE_TEST_POLICY_NAME(policy_src, NewKernelName)                                                          \
        TestUtils::test_policy_container<                                                                              \
            decltype(policy_src),                                                                                      \
            decltype(TestUtils::make_new_policy<NewKernelName>(policy_src))                                            \
        >(                                                                                                             \
            TestUtils::make_new_policy<NewKernelName>(policy_src)                                                      \
         ).get()

#else

// Clone policy and pass cloned instance as l-value / r-value
// depends on qualifiers of source policy type.
// The source policy state is not changed, so it can be used in the same test.
#    define CLONE_TEST_POLICY_IDX(policy_src, idx) CLONE_TEST_POLICY(policy_src)

// Clone policy and pass cloned instance as l-value / r-value
// depends on qualifiers of source policy type.
// The source policy state is not changed, so it can be used in the same test.
#    define CLONE_TEST_POLICY_NAME(policy_src, NewKernelName) CLONE_TEST_POLICY(policy_src)

```

As result, now in `oneDPL` tests we are simplified a lot of code:
- before:
```C++
Iterator result = ::std::adjacent_find(make_new_policy<new_kernel_name<Policy, 0>>(exec), first, last, comp);
```
in this code we passed new temporary execution policy into `oneDPL` algorithm as `ExecutionPolicy&&` (`r-value`) and it's hide a lot of compile errors in `oneDPL` Kernels when we going to `move` execution policy into `oneDPL` algorithms in test environment.
- now:
```C++
Iterator result = ::std::adjacent_find(CLONE_TEST_POLICY_IDX(exec, 0), first, last, comp);
```
in this code we passed new temporary execution policy into `oneDPL` algorithm with qualifiers which depends on source execution policy (`exec`) type qualifiers (`l-value` / `r-value`).

## Detected errors :
- https://github.com/uxlfoundation/oneDPL/pull/2329 before this PR.
- https://github.com/uxlfoundation/oneDPL/pull/2330 before this PR.

### DIFF
--- a/test/general/dpl_namespace.pass.cpp
+++ b/test/general/dpl_namespace.pass.cpp
@@ -85,9 +85,10 @@ int main()
     auto policy = TestUtils::get_dpcpp_test_policy();
     test_impl(policy);
 
+#if TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY
     TestUtils::check_compilation(policy, [](auto&& policy) { test_impl(std::forward<decltype(policy)>(policy)); });
-
 #endif
+#endif // TEST_DPCPP_BACKEND_PRESENT
 
     return TestUtils::done(TEST_DPCPP_BACKEND_PRESENT);
 }
--- a/test/general/dpl_namespace.pass.cpp
+++ b/test/general/dpl_namespace.pass.cpp
@@ -23,27 +23,13 @@
 #include <iostream>
 #include <tuple>
 
-#if TEST_DPCPP_BACKEND_PRESENT
-
 class ForEach;
 class Transform;
 class Scan;
 
-template <typename T>
-struct ForEachOp
+int main()
 {
-    const int n;
-
-    void operator()(std::tuple<T, T> x) const
-    {
-        std::get<1>(x) = (2 * std::get<0>(x)) / n;
-    }
-};
-
-template <typename Policy>
-void
-test_impl(Policy&& exec)
-{
+#if TEST_DPCPP_BACKEND_PRESENT
     const int n = 1000;
     const int k = 1000;
     using T = std::uint64_t;
@@ -59,36 +45,20 @@ test_impl(Policy&& exec)
     auto zip_first = dpl::make_zip_iterator(counting_first, key_first);
 
     // key_buf = {0,0,...0,1,1,...,1}
-    std::for_each(
-        CLONE_TEST_POLICY_NAME(exec, ForEach),
-        zip_first, zip_first + n,
-        ForEachOp<T>{n});
-
+    std::for_each(TestUtils::make_device_policy<ForEach>(dpl::execution::dpcpp_default),
+		zip_first, zip_first + n,
+        [](std::tuple<T, T> x){
+            std::get<1>(x) = (2 * std::get<0>(x)) / n;
+        });
     // val_buf = {0,1,2,...,n-1}
-    std::transform(
-        CLONE_TEST_POLICY_NAME(exec, Transform),
-        counting_first, counting_first + n, val_first, dpl::identity());
-
+    std::transform(TestUtils::make_device_policy<Transform>(dpl::execution::dpcpp_default),
+		counting_first, counting_first + n, val_first, dpl::identity());
     auto result = dpl::inclusive_scan_by_segment(
-        CLONE_TEST_POLICY_NAME(exec, Scan),
-        key_first, key_first + n, val_first, res_first);
+		TestUtils::make_device_policy<Scan>(dpl::execution::dpcpp_default),
+		key_first, key_first + n, val_first, res_first);
 
     EXPECT_EQ(k, result - res_first, "size of keys output is not valid");
-}
-
 #endif
-
-int main()
-{
-#if TEST_DPCPP_BACKEND_PRESENT
-
-    auto policy = TestUtils::get_dpcpp_test_policy();
-    test_impl(policy);
-
-#if TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY
-    TestUtils::check_compilation(policy, [](auto&& policy) { test_impl(std::forward<decltype(policy)>(policy)); });
-#endif
-#endif // TEST_DPCPP_BACKEND_PRESENT
 
     return TestUtils::done(TEST_DPCPP_BACKEND_PRESENT);
 }
--- a/test/general/dpl_namespace.pass.cpp
+++ b/test/general/dpl_namespace.pass.cpp
@@ -23,13 +23,27 @@
 #include <iostream>
 #include <tuple>
 
+#if TEST_DPCPP_BACKEND_PRESENT
+
 class ForEach;
 class Transform;
 class Scan;
 
-int main()
+template <typename T>
+struct ForEachOp
 {
-#if TEST_DPCPP_BACKEND_PRESENT
+    const int n;
+
+    void operator()(std::tuple<T, T> x) const
+    {
+        std::get<1>(x) = (2 * std::get<0>(x)) / n;
+    }
+};
+
+template <typename Policy>
+void
+test_impl(Policy&& exec)
+{
     const int n = 1000;
     const int k = 1000;
     using T = std::uint64_t;
@@ -45,19 +59,34 @@ int main()
     auto zip_first = dpl::make_zip_iterator(counting_first, key_first);
 
     // key_buf = {0,0,...0,1,1,...,1}
-    std::for_each(TestUtils::make_device_policy<ForEach>(dpl::execution::dpcpp_default),
-		zip_first, zip_first + n,
-        [](std::tuple<T, T> x){
-            std::get<1>(x) = (2 * std::get<0>(x)) / n;
-        });
+    std::for_each(
+        CLONE_TEST_POLICY_NAME(exec, ForEach),
+        zip_first, zip_first + n,
+        ForEachOp<T>{n});
+
     // val_buf = {0,1,2,...,n-1}
-    std::transform(TestUtils::make_device_policy<Transform>(dpl::execution::dpcpp_default),
-		counting_first, counting_first + n, val_first, dpl::identity());
+    std::transform(
+        CLONE_TEST_POLICY_NAME(exec, Transform),
+        counting_first, counting_first + n, val_first, dpl::identity());
+
     auto result = dpl::inclusive_scan_by_segment(
-		TestUtils::make_device_policy<Scan>(dpl::execution::dpcpp_default),
-		key_first, key_first + n, val_first, res_first);
+        CLONE_TEST_POLICY_NAME(exec, Scan),
+        key_first, key_first + n, val_first, res_first);
 
     EXPECT_EQ(k, result - res_first, "size of keys output is not valid");
+}
+
+#endif
+
+int main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+
+    auto policy = TestUtils::get_dpcpp_test_policy();
+    test_impl(policy);
+
+    TestUtils::check_compilation(policy, [](auto&& policy) { test_impl(std::forward<decltype(policy)>(policy)); });
+
 #endif
 
     return TestUtils::done(TEST_DPCPP_BACKEND_PRESENT);
--- a/test/general/test_policies.pass.cpp
+++ b/test/general/test_policies.pass.cpp
@@ -29,9 +29,9 @@ class Kernel;
 #if TEST_DPCPP_BACKEND_PRESENT
 
 template<typename Policy>
-void test_policy_instance(Policy&& policy)
+void test_policy_instance(Policy&& exec)
 {
-    sycl::queue queue = policy.queue();
+    sycl::queue queue = exec.queue();
 
     auto __max_work_group_size = queue.get_device().template get_info<sycl::info::device::max_work_group_size>();
     EXPECT_TRUE(__max_work_group_size > 0, "policy: wrong work group size");
@@ -42,11 +42,11 @@ void test_policy_instance(Policy&& policy)
     static ::std::vector<int> a(n);
 
     ::std::fill(a.begin(), a.end(), 0);
-    std::fill(std::forward<Policy>(policy), a.begin(), a.end(), -1);
+    std::fill(std::forward<Policy>(exec), a.begin(), a.end(), -1);
 #if _PSTL_SYCL_TEST_USM
     queue.wait_and_throw();
 #endif
-    EXPECT_TRUE(::std::all_of(a.begin(), a.end(), [](int i) { return i == -1; }), "wrong result of ::std::fill with policy");
+    EXPECT_TRUE(::std::all_of(a.begin(), a.end(), [](int i) { return i == -1; }), "wrong result of ::std::fill with v");
 }
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
--- a/test/general/test_policies.pass.cpp
+++ b/test/general/test_policies.pass.cpp
@@ -46,7 +46,7 @@ void test_policy_instance(Policy&& exec)
 #if _PSTL_SYCL_TEST_USM
     queue.wait_and_throw();
 #endif
-    EXPECT_TRUE(::std::all_of(a.begin(), a.end(), [](int i) { return i == -1; }), "wrong result of ::std::fill with v");
+    EXPECT_TRUE(::std::all_of(a.begin(), a.end(), [](int i) { return i == -1; }), "wrong result of ::std::fill with policy");
 }
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
--- a/test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
+++ b/test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
@@ -288,7 +288,9 @@ test_all_policies(StabilityTag stability_tag)
     auto policy = TestUtils::get_dpcpp_test_policy();
     test_device_policy(policy, stability_tag);
 
+#if TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY
     TestUtils::check_compilation(policy, [stability_tag](auto&& policy) { test_device_policy(std::forward<decltype(policy)>(policy), stability_tag); });
+#endif
 #endif // TEST_DPCPP_BACKEND_PRESENT
 
     test_std_polcies<int, int>(large_size, stability_tag);
--- a/test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
+++ b/test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
@@ -107,16 +107,16 @@ generate_data(KeyIt keys_begin, ValIt vals_begin, Size keys_n, Size vals_n, std:
 
 template<typename Policy, typename KeyIt, typename ValIt, typename Size, typename... Compare>
 void
-call_sort(Policy&& policy, KeyIt keys_begin, ValIt vals_begin, Size n, StableSortTag, Compare... compare)
+call_sort(Policy&& exec, KeyIt keys_begin, ValIt vals_begin, Size n, StableSortTag, Compare... compare)
 {
-    oneapi::dpl::stable_sort_by_key(std::forward<Policy>(policy), keys_begin, keys_begin + n, vals_begin, compare...);
+    oneapi::dpl::stable_sort_by_key(std::forward<Policy>(exec), keys_begin, keys_begin + n, vals_begin, compare...);
 }
 
 template<typename Policy, typename KeyIt, typename ValIt, typename Size, typename... Compare>
 void
-call_sort(Policy&& policy, KeyIt keys_begin, ValIt vals_begin, Size n, UnstableSortTag, Compare... compare)
+call_sort(Policy&& exec, KeyIt keys_begin, ValIt vals_begin, Size n, UnstableSortTag, Compare... compare)
 {
-    oneapi::dpl::sort_by_key(std::forward<Policy>(policy), keys_begin, keys_begin + n, vals_begin, compare...);
+    oneapi::dpl::sort_by_key(std::forward<Policy>(exec), keys_begin, keys_begin + n, vals_begin, compare...);
 }
 
 template<typename KeyIt, typename ValIt, typename Size, typename Compare = std::less<>>
@@ -176,7 +176,7 @@ check_sort(const KeyIt& keys_begin, const ValIt& vals_begin,
 
 template<typename KeyT, typename ValT, typename Size, typename Policy, typename StabilityTag, typename... Compare>
 void
-test_with_std_policy(Policy&& policy, Size n, StabilityTag stability_tag, Compare... compare)
+test_with_std_policy(Policy&& exec, Size n, StabilityTag stability_tag, Compare... compare)
 {
     Size keys_n = n;
     Size vals_n = n + 5; // to test that the remaining values are not touched
@@ -186,15 +186,17 @@ test_with_std_policy(Policy&& policy, Size n, StabilityTag stability_tag, Compar
     std::vector<KeyT> keys(origin_keys);
     std::vector<ValT> vals(origin_vals);
 
-    call_sort(std::forward<Policy>(policy), keys.begin(), vals.begin(), keys_n, stability_tag, compare...);
+    call_sort(std::forward<Policy>(exec), keys.begin(), vals.begin(), keys_n, stability_tag, compare...);
     check_sort(keys.begin(), vals.begin(), origin_keys.begin(), origin_vals.begin(), keys_n, vals_n, stability_tag, compare...);
 }
 
 #if TEST_DPCPP_BACKEND_PRESENT
+struct TagUSM;
+
 template <typename KeyT, typename ValT, sycl::usm::alloc alloc_type, std::uint32_t KernelNameID,
-          typename Size, typename StabilityTag, typename ...Compare>
+          typename Size, typename StabilityTag, typename Policy, typename ...Compare>
 void
-test_with_usm(sycl::queue& q, Size n, StabilityTag stability_tag, Compare... compare)
+test_with_usm(Policy&& exec, Size n, StabilityTag stability_tag, Compare... compare)
 {
     Size keys_n = n;
     Size vals_n = n + 5; // to test that the remaining values are not touched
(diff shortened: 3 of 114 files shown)


test/support/utils_invoke.h
```diff
@@ -219,6 +219,15 @@ struct invoke_on_all_hetero_policies
             auto my_policy = make_new_policy<kernel_name>(queue);
             iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(
                 my_policy, op, ::std::forward<Args>(rest)...);
+
+#if !__SYCL_UNNAMED_LAMBDA__
+            if (std::sin(0) == std::cos(0))
```
SergeyKopienko, line 224: Some runtime `false` condition check just to compile Kernel code without their run.

test/support/utils_invoke.h
```diff
@@ -219,6 +219,16 @@ struct invoke_on_all_hetero_policies
             auto my_policy = make_new_policy<kernel_name>(queue);
             iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(
                 my_policy, op, ::std::forward<Args>(rest)...);
+
+#if !__SYCL_UNNAMED_LAMBDA__
+            if (false)
```
SergeyKopienko, line 224: Some run-time `false`-condition check just to compile Kernel's code.

test/support/utils_invoke.h
```diff
@@ -219,6 +219,16 @@ struct invoke_on_all_hetero_policies
             auto my_policy = make_new_policy<kernel_name>(queue);
             iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(
                 my_policy, op, ::std::forward<Args>(rest)...);
+
+#if !__SYCL_UNNAMED_LAMBDA__
```
SergeyKopienko, line 223: With `ONEDPL_USE_UNNAMED_LAMBDA`==`ON` option this check simple doesn't make sense.

mmichel11:
The approach looks good to me to detect these sorts of issues assuming the compiler does not optimize out kernel compilation due to the `if (false)`.

test/support/utils_invoke.h
danhoeflinger: Yeah, this is the question I have as well.  One way I proposed we could do this is to read a value from a file at runtime which we know to be "false", but the compiler does not. 

The reason I prefer something like this where we can guarantee it is a real thing only knowable at runtime, but also deterministically `false`, is that dead code elimination may depend on compiler internals, and may change over time, so even if it works now it may not work in the future.

mmichel11: I think something like this would work: https://godbolt.org/z/9n7GW7s7q

Just define some variable `volatile bool always_false = false;` Reads to the volatile variable cannot be optimized out, so I think the contents within the if statement will always be compiled.

danhoeflinger: Thats a good idea.

SergeyKopienko: Fixed.

test/support/utils_invoke.h
```diff
@@ -219,6 +219,17 @@ struct invoke_on_all_hetero_policies
...
+            volatile bool always_false = false;
+            if (always_false)
```
SergeyKopienko, line 225: Please see explanation here: https://github.com/uxlfoundation/oneDPL/pull/2102#discussion_r1981878460

test/support/utils_invoke.h
```diff
@@ -219,6 +219,17 @@ struct invoke_on_all_hetero_policies
...
+            {
+                // We just need to compile some Kernel code and we don't need to run this code in run-time
+                // so we can move the rest of params again
+                iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(
+                    std::move(my_policy), op, ::std::forward<Args>(rest)...);
```
dmitriy-sobolev, line 230: Have you considered introduction of a dedicated test which calls ~3-5 algorithms and resembles a reproducer from #2041, or extending `lambda_naming.pass`?

What bothers me in that approach is that this change is hidden inside the testing framework and it affects the majority of tests.

SergeyKopienko: I think this change exactly inside the testing framework gives us big benefit: we will check a lot of tests on Kernel compile phase.
About your suggestion of extending `lambda_naming.pass` and etc. - this mean we should rewrite all tests, I think.
Looks overcomplicated.

SergeyKopienko: We doesn't need have additional tests I think because we just not execute them in run-time, only compile.

dmitriy-sobolev: I see your point: this approach covers a lot and costs almost nothing.
As far as I understand, it will result in extra code to compile, and with your other PR merged - only host code.

My point: this is a niche problem, it is enough to have a dedicated test, even a compile-time test.

test/support/utils_invoke.h
dmitriy-sobolev: Could you add a comment describing the purpose of the check? I mean the whole new branch with `iterator_invoker`.
It is very hard to guess it without exploring the PR on GitHub.

SergeyKopienko: Done.

test/support/utils_invoke.h
danhoeflinger: I think the goal of this is not only to test regression of a specific problem, but also protect against an easy to introduce type of bug: Using different flavors of undecayed ExecutionPolicy in kernel submitters / kernel names, resulting in clashing names and extra JIT compilation time.

This is to lower future maintenance, and hopefully to allow us to keep perfect forwarding of policies where possible in the library, since we have this bug preventative measure.  For me, I see this as an alternative to #2101.  Although, I believe @SergeyKopienko may disagree and want to go forward with both.

test/support/utils_invoke.h
```diff
@@ -266,8 +408,14 @@ struct invoke_on_all_hetero_policies
             // For example, param<int*>. In this case the runtime interpreters it as a memory object and
             // performs some checks that fail. As a workaround, define for functors which have this issue
             // __functor_type(see kernel_type definition) type field which doesn't have any pointers in it's name.
-            iterator_invoker<::std::random_access_iterator_tag, /*IsReverse*/ ::std::false_type>()(
-                my_policy, op, ::std::forward<Args>(rest)...);
+            iterator_invoker<std::random_access_iterator_tag, /*IsReverse*/ std::false_type>()(
+                my_policy, op, std::forward<Args>(rest)...);
+
+            // Check compilation of the kernel with different policy type qualifiers
+            check_compilation(my_policy, [&](auto&& __policy) {
```
mmichel11, line 425: Do you have a rough estimate of how much longer build / runtimes are with this compilation check in debug and release builds? This should tell us if we need a macro to turn this on and off in CI.

I don't think JIT time should increase since we use `-fsycl-device-code-split=per_kernel`, but I think it's worth verifying.

test/parallel_api/experimental/asynch-scan.pass.cpp
```diff
@@ -27,78 +27,104 @@
...
 
 int main()
 {
 #if TEST_DPCPP_BACKEND_PRESENT
-    test_with_buffers();
+
+    auto policy = TestUtils::get_dpcpp_test_policy();
+    test_with_buffers(policy);
+
+    //TestUtils::check_compilation(policy, [](auto&& policy) { test_with_buffers(std::forward<decltype(policy)>(policy)); });
```
mmichel11: Should this line be uncommented?

test/parallel_api/numeric/numeric.ops/exclusive_scan_by_segment.pass.cpp
```diff
@@ -154,8 +153,8 @@ DEFINE_TEST_2(test_exclusive_scan_by_segment, BinaryPredicate, BinaryOperation)
         initialize_data(host_keys.get(), host_vals.get(), host_val_res.get(), n);
         update_data(host_keys, host_vals, host_val_res);
 
-        auto new_policy3 = make_new_policy<new_kernel_name<Policy, 2>>(exec);
-        auto res3 = oneapi::dpl::exclusive_scan_by_segment(new_policy3, keys_first, keys_last, vals_first,
+        auto binary_op = [](ValT first, ValT second) { return first + second; };
```
mmichel11: I think this is an accidental addition. It doesn't look to be used.

test/parallel_api/numeric/numeric.ops/reduce_by_segment_zip.pass.cpp
```diff
@@ -87,14 +85,6 @@ test_with_usm(BinaryOp binary_op)
     dt_helper7.retrieve_data(output_values1);
     dt_helper8.retrieve_data(output_values2);
 
-//Dump
-#if 0
```
mmichel11: You may wish to revert this change based on the previous discussion(s) in https://github.com/uxlfoundation/oneDPL/pull/2310#discussion_r2163984846. I do not have a strong preference though.

test/support/utils_invoke.h
```diff
@@ -165,6 +177,89 @@ get_dpcpp_test_policy()
...
+    test_policy_container(test_policy_container&&) = delete;
+
+    // Delete assignment operator to avoid copying of test_policy_container
+    test_policy_container& operator=(const test_policy_container&) = delete;
+    test_policy_container& operator=(test_policy_container&&) = delete;
+
+    // Return testing policy
+    decltype(auto) get()
+    {
+        using TestingPolicyType = std::conditional_t<
```
mmichel11: suggestion: Is it cleaner to move this `TestingPolicyType` outside of the function and use it as the return type in `get()`?

mmichel11 (commented):
I suggest we do a full internal CI run to catch any issues prior to merging since there are are a large number of test changes.

test/parallel_api/experimental/asynch-scan.pass.cpp
SergeyKopienko: Yes, sure: uncommented.

test/parallel_api/numeric/numeric.ops/exclusive_scan_by_segment.pass.cpp
SergeyKopienko: You are right, fixed.

test/parallel_api/numeric/numeric.ops/reduce_by_segment_zip.pass.cpp
SergeyKopienko: Done

test/support/utils_invoke.h
SergeyKopienko: Done

test/support/utils_invoke.h
SergeyKopienko: Good catch.
I think make sense to check compilation with the different policy value category when `__SYCL_UNNAMED_LAMBDA__` switched off or `TEST_EXPLICIT_KERNEL_NAMES` switched on.
So `TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY` has been introduced and defined as 
```C++
// Enable check compilation with different policy value category
#ifndef TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY
#    define TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY (!__SYCL_UNNAMED_LAMBDA__ || TEST_EXPLICIT_KERNEL_NAMES)
#endif
```

mmichel11 (approved):
The approach LGTM in catching issues due to differing policy qualifiers, and the pass rate looks good after running internal CI. I recall there was some differing opinions on the use of macros with this patch. In my opinion, it is the most understandable.

I recommend confirming with others involved in the discussion of this work that they have no objections with the current approach and state of the PR before merging.

test/support/utils_invoke.h
```diff
@@ -165,6 +178,97 @@ get_dpcpp_test_policy()
...
+template <typename _PolicySource, typename _PolicyNew>
+struct test_policy_container
+{
+    using _PolicyNewDecayed = std::decay_t<_PolicyNew>;
+
+    // Define TestingPolicyType as a type which is either r-value or l-value reference
+    using TestingPolicyType = std::conditional_t<
+        std::is_reference_v<_PolicySource>,
+        std::conditional_t<std::is_rvalue_reference_v<_PolicySource>, _PolicyNewDecayed&&, const _PolicyNewDecayed&>,
+        _PolicyNewDecayed>;
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

test/support/utils_invoke.h
```diff
@@ -165,6 +178,97 @@ get_dpcpp_test_policy()
         throw;
     }
 }
+#endif // TEST_DPCPP_BACKEND_PRESENT
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

test/support/utils_invoke.h
```diff
@@ -165,6 +178,97 @@ get_dpcpp_test_policy()
...
+    using TestingPolicyType = std::conditional_t<
+        std::is_reference_v<_PolicySource>,
+        std::conditional_t<std::is_rvalue_reference_v<_PolicySource>, _PolicyNewDecayed&&, const _PolicyNewDecayed&>,
+        _PolicyNewDecayed>;
+
+    _PolicyNewDecayed __policy_new;
+    bool __policy_new_returned = false;
+
+    test_policy_container(_PolicyNewDecayed&& __policy_new)
+        : __policy_new(std::move(__policy_new))
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

mmichel11 (approved):

test/general/dpl_namespace.pass.cpp
```diff
@@ -45,20 +59,36 @@ int main()
...
+
+int main()
+{
+#if TEST_DPCPP_BACKEND_PRESENT
+
+    auto policy = TestUtils::get_dpcpp_test_policy();
+    test_impl(policy);
+
+#if TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY
+    TestUtils::check_compilation(policy, [](auto&& policy) { test_impl(std::forward<decltype(policy)>(policy)); });
```
danhoeflinger: I guess I don't object to testing compilations of alternative value categories even in tests like this (given that the work is done), but I'm a bit surprised to see it here.  Do you think its necessary to do this here rather than merely in the individual `parallel_api` and perhaps sycl_iterator tests?

test/general/test_policies.pass.cpp
danhoeflinger, line 1: Nitpick, but I dont understand the point of these name / assertion text changes...

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -107,16 +107,16 @@ generate_data(KeyIt keys_begin, ValIt vals_begin, Size keys_n, Size vals_n, std:
 
 template<typename Policy, typename KeyIt, typename ValIt, typename Size, typename... Compare>
 void
-call_sort(Policy&& policy, KeyIt keys_begin, ValIt vals_begin, Size n, StableSortTag, Compare... compare)
+call_sort(Policy&& exec, KeyIt keys_begin, ValIt vals_begin, Size n, StableSortTag, Compare... compare)
```
danhoeflinger: Same question about renaming.  I suppose it may be for the future in to make it easier to make changes via scripts / regex?  
I don't object, but I suppose we shouldn't necessarily expect that all policies are named `exec` throughout the test for maintenance, and this is already a very large PR.

test/parallel_api/iterator/input_data_sweep_host_iter.pass.cpp
```diff
@@ -29,51 +29,54 @@
...
     else
     {
-        TestUtils::unsupported_types_notifier(policy.queue().get_device());
+        TestUtils::unsupported_types_notifier(exec.queue().get_device());
     }
 }
 
+template <typename Policy>
+void
+test_impl(Policy&& exec)
```
danhoeflinger, line 54: Can we name this and the function above differently?  Its very strange to have both of these `test_impl` distinguished by different parameter lists.

test/parallel_api/iterator/input_data_sweep_sycl_iter.pass.cpp
```diff
@@ -29,52 +29,56 @@
...
     else
     {
-        TestUtils::unsupported_types_notifier(policy.queue().get_device());
+        TestUtils::unsupported_types_notifier(exec.queue().get_device());
     }
 }
 
+template <typename Policy>
+void
+test_impl(Policy&& exec)
```
danhoeflinger, line 56: change name of one layer of function, same as above.

test/parallel_api/iterator/input_data_sweep_counting_iter.pass.cpp
```diff
@@ -29,63 +29,66 @@
...
     }
 }
 
-#endif //TEST_DPCPP_BACKEND_PRESENT
-
-int
-main()
+template <typename Policy>
+void
+test_impl(Policy&& exec)
```
danhoeflinger, line 58: change name of one layer of function (same comment)

test/parallel_api/iterator/input_data_sweep_usm_device.pass.cpp
```diff
@@ -29,50 +29,54 @@
...
     else
     {
-        TestUtils::unsupported_types_notifier(policy.queue().get_device());
+        TestUtils::unsupported_types_notifier(exec.queue().get_device());
     }
 }
 
+template <typename Policy>
+void
+test_impl(Policy&& exec)
```
danhoeflinger, line 54: change name, (same name as next layer).

test/parallel_api/iterator/input_data_sweep_usm_shared.pass.cpp
```diff
@@ -29,65 +29,69 @@
...
     }
 }
 
-#endif //TEST_DPCPP_BACKEND_PRESENT
-
-int
-main()
+template <typename Policy>
+void
+test_impl(Policy&& exec)
```
danhoeflinger, line 57: same name complaint.

test/parallel_api/numeric/numeric.ops/reduce_by_segment_zip.pass.cpp
```diff
@@ -170,23 +169,38 @@ test_zip_with_discard(BinaryOp binary_op)
...
+{
+    // Run tests for USM shared/device memory
+    test_with_usm<sycl::usm::alloc::shared, 0>(CLONE_TEST_POLICY(exec), binary_op);
+    test_with_usm<sycl::usm::alloc::device, 1>(CLONE_TEST_POLICY(exec), binary_op);
+
+    test_zip_with_discard<2>(CLONE_TEST_POLICY(exec), binary_op);
+}
+
+template <typename Policy>
+void test_impl(Policy&& exec)
```
danhoeflinger, line 184: similar to above, lets come up with different names for the different layers.

test/general/dpl_namespace.pass.cpp
SergeyKopienko: My initial idea was to test everywhere...
Otherwise later we will have some tests with compilation check and some tests without it.
About the tests from `test/general` :
 - they may during compile checking sign early about possible problems inside^ usually they compiled at first.
 - so of course, we may not include this compile check into `test/general` but I think better to have this compilation check in `test/general` too.

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
SergeyKopienko: Previously in tests we had `Policy&& policy` and `Policy&& exec` together:
 - `Policy&& exec` - 373 entries in tests the `main` branch now;
 - `Policy&& policy` -  14 (16) entries in tests the `main` branch now.
It was very inconvenient for search in tests and etc. (when I made this PR).
So I had tried to rename all variants into one form - `Policy&& exec`

For simplification I am ready to prepare separate PR with this renames, and when we merge in, these changes will be absent here.

UPD: the PR prepared: https://github.com/uxlfoundation/oneDPL/pull/2338

SergeyKopienko: Also I am ready to rename all `exec` to `policy`, it doesn't matter.

test/general/test_policies.pass.cpp
SergeyKopienko: What exactly do you mean?

test/parallel_api/iterator/input_data_sweep_sycl_iter.pass.cpp
SergeyKopienko: Done

test/parallel_api/iterator/input_data_sweep_counting_iter.pass.cpp
SergeyKopienko: Done

test/parallel_api/iterator/input_data_sweep_usm_shared.pass.cpp
SergeyKopienko: Done

test/parallel_api/iterator/input_data_sweep_usm_device.pass.cpp
SergeyKopienko: Done

test/parallel_api/iterator/input_data_sweep_host_iter.pass.cpp
SergeyKopienko: Fixed (renamed)

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
SergeyKopienko: The PR https://github.com/uxlfoundation/oneDPL/pull/2338 has been merged into main branch.
So we shouldn't have any changes with renames in this PR anymore.

test/parallel_api/numeric/numeric.ops/reduce_by_segment_zip.pass.cpp
SergeyKopienko: Fixed

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
danhoeflinger: Thanks,  I guess I would just be careful not to depend upon the fact that `exec` is always the name of the policy in the tests, especially if this is not in some sort of mandated style guide for the tests.

test/general/test_policies.pass.cpp
danhoeflinger: I think you addressed this, thanks

test/general/dpl_namespace.pass.cpp
danhoeflinger: Yeah, I don't have a problem with it, I just see it as a bit of overkill with complexity. I'm ok to approve whichever choice you make here.

mmichel11 (approved):
The latest state LGTM as well.

Issue #2110, opened 2025-03-06
Brick `__brick_copy` specialized by `_ExecutionPolicy` without `std::decay_t`


SergeyKopienko:
Brick `__brick_copy` specialized by `_ExecutionPolicy` without `std::decay_t`
at:
https://github.com/uxlfoundation/oneDPL/blob/417f118fb2363908962e3871bf7ce970c32ca264/include/oneapi/dpl/pstl/glue_algorithm_impl.h#L275
and at:
https://github.com/uxlfoundation/oneDPL/blob/e11ac890599fc6e630cf10cae22977ec24d94e26/include/oneapi/dpl/internal/async_impl/glue_async_impl.h#L85
and at:
https://github.com/uxlfoundation/oneDPL/blob/417f118fb2363908962e3871bf7ce970c32ca264/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h#L801

SergeyKopienko:
The PR https://github.com/uxlfoundation/oneDPL/pull/2148 has been merged into main branch

PR #2112, opened 2025-03-06
Brick `__brick_copy` specialized by `_ExecutionPolicy` without `std::decay_t`

SergeyKopienko:
In this PR we fix the issue https://github.com/uxlfoundation/oneDPL/issues/2110

### DIFF
--- a/include/oneapi/dpl/internal/async_impl/glue_async_impl.h
+++ b/include/oneapi/dpl/internal/async_impl/glue_async_impl.h
@@ -79,10 +79,12 @@ copy_async(_ExecutionPolicy&& __exec, _ForwardIterator1 __first, _ForwardIterato
 {
     auto __dispatch_tag = oneapi::dpl::__internal::__select_backend(__exec, __first, __result);
 
+    using _DecayedExecutionPolicy = std::decay_t<_ExecutionPolicy>;
+
     wait_for_all(::std::forward<_Events>(__dependencies)...);
     auto ret_val = oneapi::dpl::__internal::__pattern_walk2_brick_async(
         __dispatch_tag, ::std::forward<_ExecutionPolicy>(__exec), __first, __last, __result,
-        oneapi::dpl::__internal::__brick_copy<decltype(__dispatch_tag), _ExecutionPolicy>{});
+        oneapi::dpl::__internal::__brick_copy<decltype(__dispatch_tag), _DecayedExecutionPolicy>{});
     return ret_val;
 }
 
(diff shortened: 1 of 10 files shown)


mmichel11:
How about just removing `_ExecutionPolicy` from the class templates of `__brick_copy` and `__brick_copy_n`? If we take a look at the our specializations of `__brick_copy`, first on the host:
https://github.com/uxlfoundation/oneDPL/blob/417f118fb2363908962e3871bf7ce970c32ca264/include/oneapi/dpl/pstl/algorithm_impl.h#L1079-L1105 and secondly in the hetero backend:
https://github.com/uxlfoundation/oneDPL/blob/417f118fb2363908962e3871bf7ce970c32ca264/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h#L416-L425 We see we use the tag to define these specializations whereas `_ExecutionPolicy` is used.

Maybe I am misunderstanding something, but to me it looks like all we need is a tag.

SergeyKopienko:
I don't know the history why `_ExecutionPolicy` is used as brick's template parameter, so right now I don't think that their remove is correct. May be someone else explain these reasons.

SergeyKopienko:
> > I don't know the history why `_ExecutionPolicy` is used as brick's template parameter, so right now I don't think that their remove is correct. May be someone else explain these reasons.
> 
> My guess is that when we added tag dispatching, we replaced the needed functionality from the policy with functionality from the Tag, but neglected to remove the policy parameter then. I agree that we can just remove this template parameter.

Yes, @danhoeflinger, you are right:
 - as example I found the previous state of
 ```C++
struct __brick_copy<_ExecutionPolicy,
                    oneapi::dpl::__internal::__enable_if_hetero_execution_policy<_ExecutionPolicy, void>>
{
    // ...
}
```

MikeDvorskiy:
> ```c++
> oneapi::dpl::__internal::__enable_if_hetero_execution_policy<_ExecutionPolicy, void>>
> ```

Yes, before the tag dispatching policy type is used for type specialization..
But, now, when we have type specialization based on tag type trait,  we don't need `_ExecutionPolicy` parameter.

SergeyKopienko:
> > ```c++
> > oneapi::dpl::__internal::__enable_if_hetero_execution_policy<_ExecutionPolicy, void>>
> > ```
> 
> Yes, before the tag dispatching policy type is used for type specialization.. But, now, when we have type specialization based on tag type trait, we don't need `_ExecutionPolicy` parameter.

As far as I understood @rarutyun on our offline discussion with him, he has opposite opinion....

SergeyKopienko:
Continue in the PR https://github.com/uxlfoundation/oneDPL/pull/2129

PR #2113, opened 2025-03-06
Remove extra `std::move` call

SergeyKopienko:
In this PR we remove extra `std::move` call in the code:
```C++
    template <typename _ValueT, typename _Wi, typename _Src, typename _Values>
    static void
    __block_load(const _Wi __wi, const _Src& __src, _Values& __values, const uint32_t __n)
    {
        _ONEDPL_PRAGMA_UNROLL
        for (uint16_t __i = 0; __i < __block_size; ++__i)
        {
            const uint16_t __idx = __wi * __block_size + __i;
            if (__idx < __n)
                new (&__values[__i]) _ValueT(::std::move(__src[__idx]));
        }
    }
```
The justification - `__src` parameter declared `const reference`


MikeDvorskiy (approved):
LGTM

danhoeflinger (approved):
LGTM

PR #2114, opened 2025-03-06
Fix warnings in tests: unused-but-set-variable

dmitriy-sobolev:


SergeyKopienko:
I propose do not touch dynamic selection tests at all due this is experimental mechanic.

test/parallel_api/dynamic_selection/test_customizations.pass.cpp
```diff
@@ -78,8 +78,8 @@ test_all_customizations()
 int
 main()
 {
-    auto actual = test_no_customizations();
```
dmitriy-sobolev, line 81: @egfefey, is it fine if I remove unused `actual` this way throughout the dynamic selection tests?

dmitriy-sobolev:
> I propose do not touch dynamic selection tests at all due this is experimental mechanic.

I asked Elvis, who participated during development of that feature. I hope he does not mind fixing warnings this way...

I see nothing wrong with fixing warnings in tests whether they are for an experimental feature or not.

test/parallel_api/dynamic_selection/test_customizations.pass.cpp
egfefey: @dmitriy-sobolev Yes that should be fine. I'll have to update them later with explicit checks for each line but it should be fine to remove the unused actual. Thanks.

mmichel11 (approved):
LGTM

PR #2118, opened 2025-03-10
Removing invalid link to Yocto instructions.

timmiesmith:
@dcbenito notified me that the Yocto instructions have been removed from the IoT Toolkit and there's not a redirect or replacement available.  Removing the dead link from the getting started guide.


dcbenito (approved):
LGTM

PR #2119, opened 2025-03-11
Move min/max XPU tests into XPU folder

dmitriy-sobolev:
These are the only tests for the C++ standard tested API residing in `test/parallel_api` instead of `test/xpu_api`. Let's move them into `test/xpu_api`.

This change will make these tests building and running only when the device backend is available.


SergeyKopienko (approved):
LGTM

PR #2120, opened 2025-03-11
Fix int-in-bool-context warning in all_of test

dmitriy-sobolev:
It fixes warning:
> all_of.pass.cpp:72:19: warning: ‘*’ in boolean context, suggest ‘&&’ instead [-Wint-in-bool-context]


mmichel11 (approved):
LGTM

PR #2121, opened 2025-03-11
Forward execution policy in __pattern_walk1 of hetero backend

mmichel11:
We only use `__exec` once here, so we can forward the forwarding reference to the `__parallel_for` call from `__pattern_walk1`.


SergeyKopienko (approved):
LGTM

PR #2122, opened 2025-03-11
Refactor argument list with an unused tag to simplify search

dmitriy-sobolev:
The change is suggested by @SergeyKopienko to simplify search of the patters using `__hetero_tag` tags.


akukanov:
Honestly, I do not quite understand how adding an attribute in front of the parameter would complicate any search for `__hetero_tag`. Could you or Sergey give an example?

SergeyKopienko:
> Honestly, I do not quite understand how adding an attribute in front of the parameter would complicate any search for `__hetero_tag`. Could you or Sergey give an example?

Sure.
The common search to find something is search by string like "(__hetero_tag"
In this search now we will miss all results with "([[maybe_unused]] __hetero_tag"

akukanov:
`__hetero_tag` - 172 hits in the header files.

`(__hetero_tag` - 120 hits. By this search, you have missed 50+ cases, including in patterns like `(__internal::__hetero_tag` and others. In other words, your search pattern is already unreliable. Adding code clutter to catch just one of many other missed matches seems wrong to me.

SergeyKopienko:
> Adding code clutter

It's not "code clutter" - I just propose to fix this warning inside function.

dmitriy-sobolev:
I suggest leaving the code as is and close the PR.

I do see much value in moving the warning suppression code into the function body - it adds more code and needs an explanation.
As for the search, one can use regular expressions to find the necessary cases, such as `\(.*__hetero_tag`, it will also cover the additional cases highlighted by Alexey.

SergeyKopienko:
@MikeDvorskiy, I see that you wrote `__process_host_iter_impl` function which also has been modified by this way.
What do you think about this approach?

MikeDvorskiy:
> `__process_host_iter_impl`

Actually, it was not me.... Rather, Dmitry.

MikeDvorskiy:
> this approach

I am not delighted form this approach, but likely there not  another way to eliminate this warning.

SergeyKopienko:
> > this approach
> 
> I am not delighted form this approach, but likely there not another way to eliminate this warning.

https://stackoverflow.com/questions/1486904/how-do-i-best-silence-a-warning-about-unused-variables :
 - a lot of any approaches for for warning like this.

dmitriy-sobolev:
> Probably, Sergey gave not quite appropriate search pattern. I guess he meant "__pattern_adjacent_difference(__hetero_tag" searching pattern, for example. In that case I cannot find the line `__pattern_adjacent_difference([[maybe_unused]] __hetero_tag`

You can use a simple regex here as well. `__pattern_adjacent_difference\(.*__hetero_tag` will match both `__pattern_adjacent_difference(__hetero_tag` and `__pattern_adjacent_difference([[maybe_unused]] __hetero_tag`. Just expect that there may be `[[maybe_unused]]` before an argument (what the language allows), and use such a regexp rather than a plain search pattern.

SergeyKopienko (approved):
LGTM

PR #2123, opened 2025-03-12
Fix int-in-bool-context warning in any_of and none_of

dmitriy-sobolev:
It fixes the same warning as in #2120, but in other tests which slipped my attention.


mmichel11 (approved):
LGTM

PR #2124, opened 2025-03-12
Remove unused class template from __parallel_for_large_submitter

mmichel11:
The `_RangeTypes` variadic class template in `__parallel_for_large_submitter` is a remnant of an older prototype of the current implementation and unused, so we are able to remove it. Thanks @SergeyKopienko for pointing this out.


SergeyKopienko (approved):
LGTM

PR #2125, opened 2025-03-12
Fix missing-field-initializers warning

dmitriy-sobolev:
The PR fixes:
> parallel_backend_sycl.h(1308,64): warning: missing field '__rng_transform' initializer [-Wmissing-field-initializers]

I believe that the original behavior was also correct, which relied on that[ aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization) rule:
> Otherwise, if the element is not a reference, the element is [copy-initialized](https://en.cppreference.com/w/cpp/language/copy_initialization) from an empty initializer list.

Hence, `__gen_mask{pred}` was an equivalent of `__gen_mask{pred, {}}` (but with the use of a copy constructor). It resulted in the warning, though.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -995,7 +995,7 @@ struct __gen_expand_count_mask
         return std::tuple(mask ? _SizeType{1} : _SizeType{0}, mask, ele);
     }
     _GenMask __gen_mask;
-    _RangeTransform __rng_transform;
+    _RangeTransform __rng_transform{};
 };
 
 struct __get_zeroth_element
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -884,7 +884,7 @@ struct __gen_mask
         return __pred((__rng_transform(std::forward<_InRng>(__in_rng)))[__id]);
     }
     _Predicate __pred;
-    _RangeTransform __rng_transform;
+    _RangeTransform __rng_transform{};
```
MikeDvorskiy, line 887: I guess we have got problem with lambda before C++20...

dmitriy-sobolev: That's right.

I've changed the approach to pass `__no_op` explicitly. Also, I removed a default template parameter to make sure it `__no_op` is explicitly passed in the future.

danhoeflinger (approved):
Yes I think this is a better approach, and LGTM

MikeDvorskiy (approved):
LGTM

PR #2126, opened 2025-03-12
Indirectly Device Accessible Iterator Trait and ADL Customization Point

danhoeflinger:
Types are "Indirectly Device Accessible" when they can represent data which can inherently be dereferenced in a SYCL kernel.  Marking types as such can help oneDPL avoid unnecessary data movement surrounding sycl kernels.

Provides a default implementation for the Argument-Dependant Lookup (ADL) customization point `is_onedpl_indirectly_device_accessible`, as well as defined ADL overloads for oneDPL's provided iterators to provide rules for when they should be considered "indirectly device accessible".

Provides a public trait, `is_indirectly_device_accessible[_v]` which can be used to query whether a type is "indirectly device accessible".  This can be used in combination with `sycl::is_device_copyable` and a check that it is a random access iterator to determine if it can be passed directly into sycl kernels without extra logic to copy the data to the device.

---
PR implementing the proposed design in #1999, and the current state of specification PR https://github.com/uxlfoundation/oneAPI-spec/pull/620.

Will not merge until specification has been merged.


include/oneapi/dpl/iterator
```diff
@@ -56,6 +56,10 @@ using ::std::rbegin;
 using ::std::rend;
 using ::std::reverse_iterator;
 using ::std::size;
+
+using oneapi::dpl::__internal::is_device_accessible_content_iterator;
```
SergeyKopienko, line 60: This `using` instructions looks strange in this file because (from my point of view) it's some public header file and it should contains only public things... Isn't it?

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -75,6 +76,56 @@ struct __make_references
...
+template <typename _Iter>
+struct __is_legacy_passed_directly<_Iter, ::std::enable_if_t<_Iter::is_passed_directly::value>> : std::true_type
+{
+};
+
+template <typename _T>
+struct __is_reverse_it_to_device_accessible_content;
+
+template <typename T>
+constexpr auto is_onedpl_device_accessible_content_iterator(T)
```
SergeyKopienko, line 93: What if we define this by a little bit another way:
```C++
template <typename T>
struct __is_onedpl_device_accessible_content_iterator_t : std::disjunction<
#if _ONEDPL_BACKEND_SYCL
    oneapi::dpl::__internal::__is_known_usm_vector_iter<std::decay_t<T>>,                   // USM vector iterator
#endif // _ONEDPL_BACKEND_SYCL
    std::is_pointer<std::decay_t<T>>,                                                       // USM pointer
    oneapi::dpl::__internal::__is_legacy_passed_directly<std::decay_t<T>>,                  // legacy passed directly iter
    oneapi::dpl::__internal::__is_reverse_it_to_device_accessible_content<std::decay_t<T>>> // reverse iterator
{
};
```

And for example, for `counting_iterator` then something like
```C++
template <>
struct __is_onedpl_device_accessible_content_iterator_t<counting_iterator> : std::true_type
{
};
```

In this case `friend` inside `class counting_iterator` wouldn't required...

danhoeflinger: If I understand correctly, you are proposing using a trait rather than an ADL based customization point, is that correct?

This has already been discussed in some amount of depth in #1999 and #1939.  The idea behind the ADL approach is to provide a way for users to customize their own types without needing to inject something into the `oneapi::dpl` namespace to specialize a trait.  If we force users to specialize in `oneapi::dpl` it requires either:
(1) separating their customization from the type definition (maintenance issue), or
(2)  closing and opening namespaces in their code, (code clutter).

If you want to propose an alternative design, that may better belong in 
 the specification PR: https://github.com/uxlfoundation/oneAPI-spec/pull/620 .

danhoeflinger: To say a little more, the ADL version allows some freedom in how to define this as a hidden friend,  as a function in the same namespace as the class definition, etc. according to the ADL rules.

SergeyKopienko: ok

include/oneapi/dpl/iterator
danhoeflinger: This is making `is_device_accessible_content_iterator_v` publicly available intentionally.  However, perhaps it should be defined directly in a public namespace though... I'm not sure.  

The closest comparison I've seen thus far is `is_execution_policy[_v]`, which we do use internally from its public namespace(s), `oneapi::dpl` and `oneapi::dpl::execution`.  

So, maybe we should directly define it in a non-uglified namespace.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -110,7 +110,7 @@ inline constexpr __is_onedpl_device_accessible_content_iterator_fn __is_onedpl_d
 
 template <typename T>
 struct is_device_accessible_content_iterator
-    : decltype(oneapi::dpl::__internal::__is_onedpl_device_accessible_content_iterator(std::declval<T>())){};
+    : std::bool_constant<bool(decltype(oneapi::dpl::__internal::__is_onedpl_device_accessible_content_iterator(std::declval<T>()))::value)>{};
```
SergeyKopienko: Am I correct that the result of `oneapi::dpl::__internal::__is_onedpl_device_accessible_content_iterator(std::declval<T>())` is the value of `true_type` or `false_type` ?
If so I don't understood why we need conversion to `bool` type here and then `std::bool_constant`...

SergeyKopienko: OK, it was just my question, I am not going to argue with this approach.

include/oneapi/dpl/iterator
danhoeflinger: I've adjusted this to be defined directly in `oneapi::dpl`

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -266,6 +315,9 @@ class counting_iterator
         return !(*this < __it);
     }
 
+    friend std::true_type
+    is_onedpl_indirectly_device_accessible_iterator(const counting_iterator&);
```
SergeyKopienko, line 319: Does it make sense to define this function here as `constexpr` ?
I mean really their result doesn't depends from `counting_iterator` state...

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -398,6 +450,10 @@ class zip_iterator
         return !(*this < __it);
     }
 
+    friend auto
+    is_onedpl_indirectly_device_accessible_iterator(const zip_iterator&)
```
SergeyKopienko, line 454: The same

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -574,6 +630,9 @@ class transform_iterator
     {
         return __my_unary_func_;
     }
+    friend auto
+    is_onedpl_indirectly_device_accessible_iterator(const transform_iterator&)
```
SergeyKopienko, line 634: The same

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -765,6 +824,12 @@ class permutation_iterator
         return !(*this < it);
     }
 
+    friend auto
+    is_onedpl_indirectly_device_accessible_iterator(const permutation_iterator&)
```
SergeyKopienko, line 828: The same

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -927,6 +992,9 @@ class discard_iterator
         return !(*this < __it);
     }
 
+    friend std::true_type
+    is_onedpl_indirectly_device_accessible_iterator(const discard_iterator&);
```
SergeyKopienko, line 996: The same

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -162,21 +166,32 @@ template <typename Iter, typename ValueType = std::decay_t<typename std::iterato
...
-constexpr bool __vector_iter_distinguishes_by_allocator_v =
-    !std::is_same_v<__default_alloc_vec_iter<Iter>, __usm_shared_alloc_vec_iter<Iter>> &&
-    !std::is_same_v<__default_alloc_vec_iter<Iter>, __usm_host_alloc_vec_iter<Iter>> &&
-    !std::is_same_v<__usm_host_alloc_vec_iter<Iter>, __usm_shared_alloc_vec_iter<Iter>>;
+struct __vector_iter_distinguishes_by_allocator
+    : std::conjunction<std::negation<std::is_same<__default_alloc_vec_iter<Iter>, __usm_shared_alloc_vec_iter<Iter>>>,
+                       std::negation<std::is_same<__default_alloc_vec_iter<Iter>, __usm_host_alloc_vec_iter<Iter>>>,
+                       std::negation<std::is_same<__usm_host_alloc_vec_iter<Iter>, __usm_shared_alloc_vec_iter<Iter>>>>
+{
+};
```
SergeyKopienko, line 179: What about redefine `__vector_iter_distinguishes_by_allocator` as
```C++
template <typename Iter>
using __vector_iter_distinguishes_by_allocator_t = 
       std::conjunction<std::negation<std::is_same<__default_alloc_vec_iter<Iter>, __usm_shared_alloc_vec_iter<Iter>>>,
                       std::negation<std::is_same<__default_alloc_vec_iter<Iter>, __usm_host_alloc_vec_iter<Iter>>>,
                       std::negation<std::is_same<__usm_host_alloc_vec_iter<Iter>, __usm_shared_alloc_vec_iter<Iter>>>>;
```

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
```diff
@@ -162,21 +166,32 @@ template <typename Iter, typename ValueType = std::decay_t<typename std::iterato
...
+inline constexpr bool __vector_iter_distinguishes_by_allocator_v =
+    __vector_iter_distinguishes_by_allocator<Iter>::value;
+
+template <typename Iter>
+struct __is_known_usm_vector_iter
+    : std::conjunction<__vector_iter_distinguishes_by_allocator<Iter>,
+                       std::disjunction<std::is_same<Iter, oneapi::dpl::__internal::__usm_shared_alloc_vec_iter<Iter>>,
+                                        std::is_same<Iter, oneapi::dpl::__internal::__usm_host_alloc_vec_iter<Iter>>>>
+{
+};
```
SergeyKopienko, line 191: ...then we may reimplement this as
```C++
template <typename Iter>
using __is_known_usm_vector_iter_t =
      std::conjunction<__vector_iter_distinguishes_by_allocator<Iter>,
                       std::disjunction<std::is_same<Iter, oneapi::dpl::__internal::__usm_shared_alloc_vec_iter<Iter>>,
                                        std::is_same<Iter, oneapi::dpl::__internal::__usm_host_alloc_vec_iter<Iter>>>>;
```

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -277,6 +221,16 @@ struct is_hetero_legacy_trait<Iter, ::std::enable_if_t<Iter::is_hetero::value>>
 template <typename Iter>
 inline constexpr bool is_sycl_iterator_v = is_sycl_iterator<Iter>::value || is_hetero_legacy_trait<Iter>::value;
 
+template <typename _Iter>
+struct __is_passed_directly_device_ready
+    : std::conjunction<oneapi::dpl::is_indirectly_device_accessible_iterator<_Iter>, sycl::is_device_copyable<_Iter>,
+                       oneapi::dpl::__internal::__is_random_access_iterator<_Iter>>
+{
+};
```
SergeyKopienko, line 229: ...then
```C++
template <typename _Iter>
using __is_passed_directly_device_ready_t =
      std::conjunction<oneapi::dpl::is_indirectly_device_accessible_iterator<_Iter>, sycl::is_device_copyable<_Iter>,
                       oneapi::dpl::__internal::__is_random_access_iterator<_Iter>>;
```

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -285,8 +239,9 @@ struct is_temp_buff : ::std::false_type
 };
 
 template <typename _Iter>
-struct is_temp_buff<_Iter, ::std::enable_if_t<!is_sycl_iterator_v<_Iter> && !::std::is_pointer_v<_Iter> &&
-                                              !is_passed_directly_v<_Iter>>> : ::std::true_type
+struct is_temp_buff<_Iter, std::enable_if_t<!is_sycl_iterator_v<_Iter> && !std::is_pointer_v<_Iter> &&
```
SergeyKopienko, line 242: Should we write here full qualified name for `is_sycl_iterator_v` (include namespace)?

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -546,8 +501,10 @@ struct __get_sycl_range
     }
 
     //specialization for permutation_iterator using USM pointer or direct pass object as source
-    template <sycl::access::mode _LocalAccMode, typename _Iter, typename _Map,
-              ::std::enable_if_t<!is_sycl_iterator_v<_Iter> && is_passed_directly_v<_Iter>, int> = 0>
+    template <
+        sycl::access::mode _LocalAccMode, typename _Iter, typename _Map,
+        std::enable_if_t<
+            !is_sycl_iterator_v<_Iter> && oneapi::dpl::__ranges::__is_passed_directly_device_ready_v<_Iter>, int> = 0>
```
SergeyKopienko, line 507: The same

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -563,9 +520,11 @@ struct __get_sycl_range
...
 
     // specialization for general case, permutation_iterator with base iterator that is not sycl_iterator or
-    // passed directly.
-    template <sycl::access::mode _LocalAccMode, typename _Iter, typename _Map,
-              ::std::enable_if_t<!is_sycl_iterator_v<_Iter> && !is_passed_directly_v<_Iter>, int> = 0>
+    // device accessible content iterators.
+    template <
+        sycl::access::mode _LocalAccMode, typename _Iter, typename _Map,
+        std::enable_if_t<
+            !is_sycl_iterator_v<_Iter> && !oneapi::dpl::__ranges::__is_passed_directly_device_ready_v<_Iter>, int> = 0>
```
SergeyKopienko, line 527: The same

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -277,6 +221,16 @@ struct is_hetero_legacy_trait<Iter, ::std::enable_if_t<Iter::is_hetero::value>>
 template <typename Iter>
 inline constexpr bool is_sycl_iterator_v = is_sycl_iterator<Iter>::value || is_hetero_legacy_trait<Iter>::value;
```
SergeyKopienko, line 222: Does it make sense to reimplement this like
```C++
template <typename Iter>
using is_sycl_iterator_t = std::conjunction<is_sycl_iterator<Iter>, is_hetero_legacy_trait<Iter>>;

template <typename Iter>
inline constexpr bool is_sycl_iterator_v = is_sycl_iterator_t<Iter>::value;
```

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -75,6 +76,38 @@ struct __make_references
     }
 };
 
+template <typename _Iter, typename _Void = void>
```
SergeyKopienko, line 79: Probably good enough to write this like
```suggestion
template <typename _Iter, typename = void>
```

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -157,6 +190,22 @@ namespace oneapi
...
 {
+
+template <typename T>
+struct is_indirectly_device_accessible_iterator
+    : decltype(oneapi::dpl::__internal::__is_onedpl_indirectly_device_accessible_iterator(std::declval<T>()))
+{
+    static_assert(
+        std::is_same_v<decltype(decltype(oneapi::dpl::__internal::__is_onedpl_indirectly_device_accessible_iterator(
+                           std::declval<T>()))::value),
+                       const bool>,
```
SergeyKopienko, line 201: Should we really use here `const bool` or simple `bool` is ok too?

https://stackoverflow.com/questions/1443659/does-it-make-sense-to-return-fundamental-types-by-const-value
"
To be a little more specific, only "objects" can be const. The C++ standard's definition of "object" includes everything an lvalue refers to ("has a name") and class-type temporaries. A boolean return value is an rvalue of a non-class type which is why a standards-compliant compiler will just ignore "const" in this case. As others said already, it's useless in this context.
"

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -157,6 +190,22 @@ namespace oneapi
...
+        "Return type of is_onedpl_indirectly_device_accessible_iterator does not have the characteristics of a "
+        "bool_constant");
+};
+
+template <typename T>
+inline constexpr bool is_indirectly_device_accessible_iterator_v = is_indirectly_device_accessible_iterator<T>::value;
```
SergeyKopienko, line 207: Does it make sense to redefine it like
```suggestion
template <typename T>
using is_indirectly_device_accessible_iterator_t =
    std::integral_constant<bool, decltype(oneapi::dpl::__internal::__is_onedpl_indirectly_device_accessible_iterator(
                                     std::declval<T>()))::value>;
```

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -157,6 +190,22 @@ namespace oneapi
...
+struct is_indirectly_device_accessible_iterator
+    : decltype(oneapi::dpl::__internal::__is_onedpl_indirectly_device_accessible_iterator(std::declval<T>()))
+{
+    static_assert(
+        std::is_same_v<decltype(decltype(oneapi::dpl::__internal::__is_onedpl_indirectly_device_accessible_iterator(
+                           std::declval<T>()))::value),
+                       const bool>,
+        "Return type of is_onedpl_indirectly_device_accessible_iterator does not have the characteristics of a "
+        "bool_constant");
+};
```
SergeyKopienko, line 204: Does it make sense to redefine it like
```suggestion
template <typename T>
using is_indirectly_device_accessible_iterator_t =
    std::integral_constant<bool, decltype(oneapi::dpl::__internal::__is_onedpl_indirectly_device_accessible_iterator(
                                     std::declval<T>()))::value>;
```

include/oneapi/dpl/pstl/iterator_impl.h
SergeyKopienko, line 207: ...then
```suggestion
template <typename T>
inline constexpr bool is_indirectly_device_accessible_iterator_v = is_indirectly_device_accessible_iterator_t<T>::value;
```

SergeyKopienko: And the second point.
How we going to use this const in our code? Or in the user code?
Looks like it's just for our tests. If so, should we really want to have this const here?

SergeyKopienko:
Should we apply clang format suggestions here?

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: I originally had this as `constexpr` and removed it.  We deliberately don't define a body for the function because it is only used in a `decltype` context to determine the return type (which must be known / available at compile-time inherently).  

It may seem to be more descriptive with `constexpr`, but tend to I think it may be somewhat misleading.  This function should never be called outside a `decltype` context, and `constexpr` doesn't **require** that a function is only used at compile-time, rather it says that this is available to be called at compile-time.  In fact we aren't ever calling the function even at compile-time,  we are merely deducing the return type, which you can do with any function whether it is labelled  `constexpr` or not.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
danhoeflinger: Do you have an example of a `true/false` trait where we use `_t` suffix?  
My understanding is that the convention is to use `_t` suffix when it is meant to describe a type rather than a binary switch (bool_constant).  I'm looking at the examples `is_execution_policy[_v]`, `__is_sycl_accessor[_v]`, `__is_*_tag[_v]` and following that convention.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: yes. fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: yes. fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: yes. fixed.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Here we are attempting to detect if something is not a `bool_constant` as a return type from the ADL customization point.  
`bool_constant`-s are defined to be `integral_constant<bool, v>`, and those must have have a `constexpr bool value` member constant. `constexpr` variables are `const`, so when we are checking `is_same_v` we need to match or decay the `const` away.  Since we know that these should be `const bool` by the specification, we need to check against the same. I've tried this out in practice, and at least with icpx we need `const bool` or a `std::decay_t`, I believe any valid implementation would be the same. https://en.cppreference.com/w/cpp/types/integral_constant

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
SergeyKopienko: https://github.com/uxlfoundation/oneDPL/blob/87ad0a609ef250e4a36957a127e0b6d48f021646/include/oneapi/dpl/pstl/hetero/dpcpp/execution_sycl_defs.h#L365
 - but it's without `_t` postfix...
https://github.com/uxlfoundation/oneDPL/blob/87ad0a609ef250e4a36957a127e0b6d48f021646/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L54
 - the same..
 - So I don't insist on `_t` postfix of course, but what is more correct - to use `struct` or template alias here?

include/oneapi/dpl/pstl/iterator_impl.h
SergeyKopienko: Understood your point, ok.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: As stated before, I disagree with the `_t` part, but as for an alias instead of a `struct`...

It removes our ability to `static_assert`.  After a discussion offline with @reble , I believe its best to force people into returning types with base characteristics of `bool_constant` rather than permitting `integral_constant` with `value` 
constant convertible to `bool`.

This allows the specification to be simple (and only discuss `bool_constant`). We don't have a compelling use case for needing to support `integral_constant` with `value` constant convertible to `bool`. With the `static_assert`, it provides a useful error which will correct any reliance on undefined behavior outside the simpler specification.

However, what you describe is a possibility, and if we have a compelling use case of an possible user type which wants to use an `integral_constant` directly I'm open to changing my mind here (which probably means changing the spec PR as well).

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: This is a part of the public API, provided for convenience as it common with this sort of `bool_constant` trait.  
We use the `bool_constant` in our code, not just our tests, but could just as easily use the value version of it now or in the future.  I think it is useful.

SergeyKopienko: I mean that I found only one usage of `is_indirectly_device_accessible_iterator_v` in our includes:
```C++
template <typename T>
inline constexpr bool is_indirectly_device_accessible_iterator_v = is_indirectly_device_accessible_iterator<T>::value;
```
It's ok?

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: I don't know if I agree with your exact suggestion, but I definitely don't like that `is_sycl_iterator` and  `is_sycl_iterator_v` don't mean **exactly** the same thing.
My suggestion would probably be to just rename `is_sycl_iterator_v` to be `is_sycl_or_hetero_iterator_v`. If you agree I can include it in this PR.  If we get more involved than that, it is probably outside the scope of this PR.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Yes.  We use `oneapi::dpl::is_indirectly_device_accessible<_Iter>` in the definition of `__is_passed_directly_device_ready`, but the value version of the trait is provide for convenience and flexibility of user code as part of the public API.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
SergeyKopienko: Dan, about any renames - up to you.
My idea was simple define some `constexpr bool` const state as `value` from other type, nothing more.

include/oneapi/dpl/pstl/iterator_impl.h
SergeyKopienko: Ok

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: I'll make a type and value version of `is_sycl_or_hetero_iterator[_v]`.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
danhoeflinger: Adjusted to be an alias.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
danhoeflinger: Adjusted to be an alias.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: Adjusted to be an alias.

include/oneapi/dpl/pstl/iterator_impl.h
danhoeflinger: Both versions force the public trait to be a `bool_constant`, regardless of what users return from the ADL based customization point, which I think is important.  This version in the PR with the static assertion requires explicitly that users return `bool_constant` from their ADL based customization point functions.  If we decide to switch to a more permissive specification, then your suggestion should be what we use for the implementation.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -940,6 +1008,16 @@ namespace dpl
 {
 namespace __internal
 {
+template <typename _T>
+struct __is_reversed_indirectly_device_accessible_it : std::false_type
+{
+};
+
+template <typename _BaseIter>
+struct __is_reversed_indirectly_device_accessible_it<std::reverse_iterator<_BaseIter>>
```
mmichel11, line 1017: Do we support `std::move_iterator`? If so, do we need a specialization similar to this one for `std::reverse_iterator`?

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -275,7 +219,18 @@ struct is_hetero_legacy_trait<Iter, ::std::enable_if_t<Iter::is_hetero::value>>
...
-inline constexpr bool is_sycl_iterator_v = is_sycl_iterator<Iter>::value || is_hetero_legacy_trait<Iter>::value;
+using is_sycl_or_hetero_iterator = std::disjunction<is_sycl_iterator<Iter>, is_hetero_legacy_trait<Iter>>;
+
+template <typename Iter>
+inline constexpr bool is_sycl_or_hetero_iterator_v = is_sycl_or_hetero_iterator<Iter>::value;
+
+template <typename _Iter>
+using __is_passed_directly_device_ready =
+    std::conjunction<oneapi::dpl::is_indirectly_device_accessible<_Iter>, sycl::is_device_copyable<_Iter>,
+                     oneapi::dpl::__internal::__is_random_access_iterator<_Iter>>;
```
mmichel11, line 230: Do we want to require that the iterator is random access in the spec? Here's the corresponding comment: https://github.com/uxlfoundation/oneAPI-spec/pull/620#discussion_r2069304471

Probably, we should choose to change it in the spec PR or remove it here.

mmichel11 (commented):
I have went through the current specification PR and verified that the requirements are satisfied here. I just have one question in the spec about specifying that iterators must be random access to use on the device.

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -765,6 +824,12 @@ class permutation_iterator
         return !(*this < it);
     }
 
+    friend auto
+    is_onedpl_indirectly_device_accessible_iterator(const permutation_iterator&)
+        -> std::conjunction<oneapi::dpl::is_indirectly_device_accessible<SourceIterator>,
+                            oneapi::dpl::is_indirectly_device_accessible<
```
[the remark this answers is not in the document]
danhoeflinger: good catch, yes this works.

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: I think its something we want.  However, I'm not so confident that checking for `std::random_access_iterator_tag` is a super reliable way of doing this, but I think it is the best we have `pre-c++20` concepts.

I worry a little about this cutting off custom types which are in practice random access iterators, but do not add the random access iterator tag.  

@akukanov, Should we enforce this requirement, or only state it in the specification without enforcement?

akukanov: It's really orthogonal to the iterators, device accessibility etc. - it's a question of the requirements set for/by parallel algorithms, specifically when invoked with a device policy. It probably deserves clarification, but not within the scope of this patch nor the corresponding spec patch.

danhoeflinger: I suppose so.  We are already covered for `sycl_iterator` by a lack of device copyability of a `sycl::queue`, so maybe this check is frivolous and just liable to get us into trouble.  I think I will remove it.

test/general/implementation_details/indirectly_device_accessible.pass.cpp
```diff
@@ -0,0 +1,461 @@
+// -*- C++ -*-
+//===-- indirectly_device_accessible.pass.cpp -----------------------------------------------===//
```
MikeDvorskiy, line 2: Minor comment. I would recommend don't write file name here(Essentially, we already do it almost everywhere in OneDPL). 
Reason: it duplicates a filename of the file. And such duplication might lead to an issues with filename discrepancy when somebody wants to rename the filename, and rather forget rename this string as well.

danhoeflinger: sure, done

MikeDvorskiy: `indirectly_device_accessible.pass.cpp` is too long filename.... I believe  we can use more compact filename w/o sense loss... 
BTW, probably, I missed something from offline discussion about "indirectly access" C++ term. Or I could not explain my point of view on that.
Yes, I understand that 
```
int val;
val = 6;
```
is a direct access.

```int* pval = &val;
*pval = 6;
```
is indirect access in C/C++ (in host memory model) term.

But, we are talking about  a trait with another meaning: 
it has a device execution/data access specific and means: an instance of such type should be passed  "AS IS" (by value, automatically with SYCL API), and within a kernel we have an access (directly or indirectly) to the data. 
In contrast, when we cannot pass data AS IS sycl buffer, for example.

So, I would suggest just "device accessible" term.

test/general/implementation_details/indirectly_device_accessible.pass.cpp
```diff
@@ -0,0 +1,461 @@
...
+using simple_implicitly_not_indirectly_device_accessible_iterator = simple_iterator;
+
+struct simple_indirectly_device_accessible_iterator : public simple_iterator
+{
+    using is_passed_directly = std::true_type;
+
+    simple_indirectly_device_accessible_iterator(int start = 0) : simple_iterator(start) {}
+};
+
+struct simple_explicitly_not_indirectly_device_accessible_iterator : public simple_iterator
```
MikeDvorskiy, line 160: Is an extremely long filename....

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
```diff
@@ -275,7 +219,17 @@ struct is_hetero_legacy_trait<Iter, ::std::enable_if_t<Iter::is_hetero::value>>
 };
 
 template <typename Iter>
-inline constexpr bool is_sycl_iterator_v = is_sycl_iterator<Iter>::value || is_hetero_legacy_trait<Iter>::value;
+using is_sycl_or_hetero_iterator = std::disjunction<is_sycl_iterator<Iter>, is_hetero_legacy_trait<Iter>>;
```
MikeDvorskiy, line 222: Am I right that "sycl iterator" is a special case of more general term "hetero iterator"?
If so, I would suggest more general name `is_hetero_iterator`.

test/general/implementation_details/indirectly_device_accessible.pass.cpp
danhoeflinger: Well, to clarify, it is not the same semantically as the previous "passed directly" trait.  
"indirectly device accessible" only means that it (indirectly) represents data on the device.  `sycl_iterator` also is "indirectly device accessible" but it is not "passed directly".  

We combine the check with "sycl device-copyable" to achieve "passed directly" and filter out `sycl_iterator`, as it requires additional processing to convert the buffer and offset to an accessor during the submit call.

It is a necessary but not sufficient condition for "passed directly".

include/oneapi/dpl/pstl/iterator_impl.h
```diff
@@ -75,6 +76,38 @@ struct __make_references
...
+template <typename _Iter>
+struct __is_legacy_passed_directly<_Iter, ::std::enable_if_t<_Iter::is_passed_directly::value>> : std::true_type
+{
+};
+
+template <typename _T>
+struct __is_reversed_indirectly_device_accessible_it;
+
+template <typename T>
+constexpr auto is_onedpl_indirectly_device_accessible_iterator(T)
```
MikeDvorskiy, line 93: So too long name...
BTW, due to "indirectly" is confusing term... may be we omit that?)

test/general/implementation_details/indirectly_device_accessible.pass.cpp
danhoeflinger: shortened

include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h
danhoeflinger: suggestion taken.

test/general/implementation_details/indirectly_device_accessible.pass.cpp
danhoeflinger: Also, the example you give is basically exactly what we are describing, only on the device, which brings us to the current name.

"Device_accessible" is too similar to "sycl device-copyable" in my opinion and is somewhat ambiguous, where I believe "indirectly device accessible" properly puts the focus on the content that the variable refers to.

include/oneapi/dpl/pstl/iterator_impl.h
MikeDvorskiy, line 454: Just a couple of questions.
Do we have some risk that we use not fully qualified name for `zip_iterator`?
Do we have some risk that something  `zip_iterator` value (with qualificators) doesn't
match to `const zip_iterator&`?

test/general/implementation_details/indirectly_device_accessible.pass.cpp
MikeDvorskiy: Ok, thank you for clarification.

test/parallel_api/iterator/indirectly_device_accessible.pass.cpp
```diff
@@ -0,0 +1,464 @@
...
+
+} // namespace custom_user
+
+template <bool base_indirectly_device_accessible, typename BaseIter>
+void
+test_with_base_iterator()
+{
+    //test assumption about base iterator device accessible content iterator
+    static_assert(oneapi::dpl::is_indirectly_device_accessible_v<BaseIter> == base_indirectly_device_accessible,
+                  "is_onedpl_indirectly_device_accessible_iterator is not working correctly for base iterator");
```
MikeDvorskiy, line 350: It is normal that we check `oneapi::dpl::is_indirectly_device_accessible_v` but in the assert message we have another name - `is_onedpl_indirectly_device_accessible_iterator`?

test/parallel_api/iterator/indirectly_device_accessible.pass.cpp
```diff
@@ -0,0 +1,464 @@
...
+    // custom iter type implicitly not device accessible content iterator
+    test_with_base_iterator<false, implicit_non_IDA_iter>();
+
+    // std vector with normal allocator
+    std::vector<int> vec2(10);
+    test_with_base_iterator<false, decltype(vec2.begin())>();
+
+    // test discard_iterator
+    static_assert(oneapi::dpl::is_indirectly_device_accessible_v<oneapi::dpl::discard_iterator> == true,
+                  "is_onedpl_indirectly_device_accessible_iterator is not working correctly for discard iterator");
```
MikeDvorskiy: In this couple of places (here and below), is it normal that we check a specialization for `oneapi::dpl::is_indirectly_device_accessible_v`, but assert message tells about another trait name - `oneapi::dpl::is_indirectly_device_accessible_v` ?

test/general/implementation_details/indirectly_device_accessible.pass.cpp
```diff
@@ -0,0 +1,464 @@
...
+
+    // test discard_iterator
+    static_assert(oneapi::dpl::is_indirectly_device_accessible_v<oneapi::dpl::discard_iterator> == true,
+                  "is_onedpl_indirectly_device_accessible_iterator is not working correctly for discard iterator");
+
+    // test buffer_wrapper
+    sycl::buffer<int, 1> buf(10);
+    auto buffer_wrapper = oneapi::dpl::begin(buf);
+    static_assert(oneapi::dpl::is_indirectly_device_accessible_v<decltype(buffer_wrapper)> == true,
+                  "is_onedpl_indirectly_device_accessible_iterator is not working correctly for sycl iterator");
```
MikeDvorskiy: Actually we have not officially "sycl iterator" term. In the doc it is an unspecified type which is returned from `onedpl::begin()`.
Yes, we have `sycl_iterator` type but it internal type.

test/general/implementation_details/indirectly_device_accessible.pass.cpp
```diff
@@ -0,0 +1,464 @@
...
+    // custom iter type implicitly not device accessible content iterator
+    test_with_base_iterator<false, implicit_non_IDA_iter>();
+
+    // std vector with normal allocator
+    std::vector<int> vec2(10);
+    test_with_base_iterator<false, decltype(vec2.begin())>();
+
+    // test discard_iterator
+    static_assert(oneapi::dpl::is_indirectly_device_accessible_v<oneapi::dpl::discard_iterator> == true,
+                  "is_onedpl_indirectly_device_accessible_iterator is not working correctly for discard iterator");
```
MikeDvorskiy: We have a public type  - `discard_iterator`,
Probably, should we use  ".....is not working correctly for `discard_iterator`" ?

MikeDvorskiy (dismissed):
LGTM

mmichel11 (dismissed):
I have reviewed the commits since my last review. Everything LGTM.

mmichel11 (approved):
I reviewed the test change and it LGTM.

PR #2129, opened 2025-03-13
Some bricks specialized by `ExecutionPolicy` without apply `std::decay_t`

SergeyKopienko:
In this PR we fix the issue https://github.com/uxlfoundation/oneDPL/issues/2110

As we discussed with @rarutyun we still need to have `EcecutionPolicy` in bricks from architecture point of view.

This PR is a new copy of https://github.com/uxlfoundation/oneDPL/pull/2112

## Implementation details
Now in all `struct __brick_...` implementations we have new `static_assert` :
```C++
static_assert(std::is_same_v<_ExecutionPolicy, std::decay_t<_ExecutionPolicy>>);
```
So we will have compile error if we use `brick` with non-decayed policy.

### DIFF
--- a/include/oneapi/dpl/internal/async_impl/glue_async_impl.h
+++ b/include/oneapi/dpl/internal/async_impl/glue_async_impl.h
@@ -79,10 +79,12 @@ copy_async(_ExecutionPolicy&& __exec, _ForwardIterator1 __first, _ForwardIterato
 {
     auto __dispatch_tag = oneapi::dpl::__internal::__select_backend(__exec, __first, __result);
 
+    using _DecayedExecutionPolicy = std::decay_t<_ExecutionPolicy>;
+
     wait_for_all(::std::forward<_Events>(__dependencies)...);
     auto ret_val = oneapi::dpl::__internal::__pattern_walk2_brick_async(
         __dispatch_tag, ::std::forward<_ExecutionPolicy>(__exec), __first, __last, __result,
-        oneapi::dpl::__internal::__brick_copy<decltype(__dispatch_tag), _ExecutionPolicy>{});
+        oneapi::dpl::__internal::__brick_copy<decltype(__dispatch_tag), _DecayedExecutionPolicy>{});
     return ret_val;
 }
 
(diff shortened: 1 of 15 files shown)


danhoeflinger (commented):
My first reaction to this is that it is very likely a good change.
However, I think we need to wait to consider this in the context of a larger discussion about differing reference types of arguments (Policy and ranges) and SYCL kernels, which I hope will happen in the next couple weeks.

We first need to decide what the expected behavior is when a user passes different flavored reference types of policies and ranges, and if that changes for different contexts:

1. no unnamed lambda vs unnamed lambda
2.  same kernel name specified vs different kernel name specified vs none specified

Once we have some conclusions about expected behavior there, we can figure out the best steps to achieve that.  

I believe something like this will be part of the solution. However, I would resist merging any related changes until we can take a step back and consider it at from a higher level.

danhoeflinger:
(If a user is waiting on a specific fix for a compiler error, that may change the calculus here, but I don't think that is the case)

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
```diff
@@ -79,10 +79,12 @@ copy_async(_ExecutionPolicy&& __exec, _ForwardIterator1 __first, _ForwardIterato
 {
     auto __dispatch_tag = oneapi::dpl::__internal::__select_backend(__exec, __first, __result);
 
+    using _DecayedExecutionPolicy = std::decay_t<_ExecutionPolicy>;
+
     wait_for_all(::std::forward<_Events>(__dependencies)...);
     auto ret_val = oneapi::dpl::__internal::__pattern_walk2_brick_async(
         __dispatch_tag, ::std::forward<_ExecutionPolicy>(__exec), __first, __last, __result,
-        oneapi::dpl::__internal::__brick_copy<decltype(__dispatch_tag), _ExecutionPolicy>{});
+        oneapi::dpl::__internal::__brick_copy<decltype(__dispatch_tag), _DecayedExecutionPolicy>{});
```
akukanov, line 87: In this and many other functions, `_DecayedExecutionPolicy` is only used once in its definition scope. Meanwhile, it saves just 7 symbols, comparing to `std::decay_t<_ExecutionPolicy>`, and adds nothing to readability.

I see little value in the alias even where it is reused, and absolutely zero (that is, -273 :)) reasons to add it when used only once.

SergeyKopienko: I am ready to any variant, it's not problem....

SergeyKopienko: Fixed

PR #2133, opened 2025-03-17
Correct reduce-then-scan -O0 workaround behavior with different host and device optimization levels

mmichel11:
https://github.com/uxlfoundation/oneDPL/pull/2046 introduced a workaround for a hardware bug that prevents sub-group sizes of 32 from being used with `-O0` compilation on certain devices based on the detection of optimization via the `__OPTIMIZE__` macro. However, this approach does not work when the host and device compiler use different optimization levels, specifically when only one of the two phases uses `-O0` compilation. The integral sub-group size constant is embedded in the kernel name causing the host to call a kernel not compiled by the device which results in "Kernel not found" errors.

To support these cases where the host and device compiler optimization levels may differ, we need to make the following changes:

1. Remove the sub-group size from the kernel name. This can be done through removing it as an integral template parameter to the submitters which gets reflected in the kernel naming scheme.
2. Allocate sufficient temporary storage on the host to handle both sub-group sizes of 16 and 32. We store one partial result per sub-group, so the sub-group size of 16 case requires more temporary storage.
3. Compute all sub-group specific fields (e.g. `__num_sub_groups_local`, `__num_sub_groups_global`, etc.) on the device itself as the true sub-group size can only be determined when on the device.
4.  Conditionally, enable sub-group sizes of 16 when optimization is not enabled **AND** we are compiling to a SPIR-V (Intel) target GPU. This enables other targets to use a sub-group size of 32 for all cases.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -338,6 +338,9 @@ struct __parallel_reduce_then_scan_reduce_submitter<__max_inputs_per_item, __is_
                 __cgh, __dpl_sycl::__no_init{});
             __cgh.parallel_for<_KernelName...>(
                     __nd_range, [=, *this](sycl::nd_item<1> __ndi) [[sycl::reqd_sub_group_size(__sub_group_size)]] {
+                // Compute work distribution fields dependent on sub-group size within the kernel. This is because we
+                // can only rely on the value of __sub_group_size provided in the device compilation phase within the
+                // kernel itself.
                 const auto [__num_sub_groups_local, __num_sub_groups_global, __inputs_per_sub_group,
                             __inputs_per_item] =
                     __get_reduce_then_scan_sub_group_params(__work_group_size, __sub_group_size, __max_num_work_groups,
@@ -907,6 +910,7 @@ __parallel_transform_reduce_then_scan(oneapi::dpl::__internal::__device_backend_
         // 2. Scan step - Compute intra-wg carries, determine sub-group carry-ins, and perform full input block scan.
         __event = __scan_submitter(__exec, __kernel_nd_range, __in_rng, __out_rng, __result_and_scratch, __event,
                                    __inputs_remaining, __b);
+        __inputs_remaining -= std::min(__inputs_remaining, __block_size);
         if (__b + 2 == __num_blocks)
         {
             __inputs_per_item = __inputs_remaining >= __max_inputs_per_block
@@ -915,7 +919,6 @@ __parallel_transform_reduce_then_scan(oneapi::dpl::__internal::__device_backend_
                                                       oneapi::dpl::__internal::__dpl_bit_ceil(__inputs_remaining),
                                                       __num_work_groups * __work_group_size);
         }
-        __inputs_remaining -= std::min(__inputs_remaining, __block_size);
     }
     return __future(__event, __result_and_scratch);
 }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -307,16 +307,14 @@ class __reduce_then_scan_reduce_kernel;
 template <typename... _Name>
 class __reduce_then_scan_scan_kernel;
 
-template <std::uint16_t __max_inputs_per_item, bool __is_inclusive,
-          bool __is_unique_pattern_v, typename _GenReduceInput, typename _ReduceOp, typename _InitType,
-          typename _KernelName>
+template <std::uint16_t __max_inputs_per_item, bool __is_inclusive, bool __is_unique_pattern_v,
+          typename _GenReduceInput, typename _ReduceOp, typename _InitType, typename _KernelName>
 struct __parallel_reduce_then_scan_reduce_submitter;
 
-template <std::uint16_t __max_inputs_per_item, bool __is_inclusive,
-          bool __is_unique_pattern_v, typename _GenReduceInput, typename _ReduceOp, typename _InitType,
-          typename... _KernelName>
-struct __parallel_reduce_then_scan_reduce_submitter<__max_inputs_per_item, __is_inclusive,
-                                                    __is_unique_pattern_v, _GenReduceInput, _ReduceOp, _InitType,
+template <std::uint16_t __max_inputs_per_item, bool __is_inclusive, bool __is_unique_pattern_v,
+          typename _GenReduceInput, typename _ReduceOp, typename _InitType, typename... _KernelName>
+struct __parallel_reduce_then_scan_reduce_submitter<__max_inputs_per_item, __is_inclusive, __is_unique_pattern_v,
+                                                    _GenReduceInput, _ReduceOp, _InitType,
                                                     __internal::__optional_kernel_name<_KernelName...>>
 {
     static constexpr std::uint8_t __sub_group_size = __get_reduce_then_scan_sg_sz();
@@ -326,8 +324,7 @@ struct __parallel_reduce_then_scan_reduce_submitter<__max_inputs_per_item, __is_
     sycl::event
     operator()(_ExecutionPolicy&& __exec, const sycl::nd_range<1> __nd_range, _InRng&& __in_rng,
                _TmpStorageAcc& __scratch_container, const sycl::event& __prior_event,
-               const std::size_t __inputs_remaining,
-               const std::size_t __block_num) const
+               const std::size_t __inputs_remaining, const std::size_t __block_num) const
     {
         using _InitValueType = typename _InitType::__value_type;
         return __exec.queue().submit([&, this](sycl::handler& __cgh) {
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -266,42 +266,84 @@ __scan_through_elements_helper(const __dpl_sycl::__sub_group& __sub_group, _GenI
...
                 __cgh, __dpl_sycl::__no_init{});
             __cgh.parallel_for<_KernelName...>(
                     __nd_range, [=, *this](sycl::nd_item<1> __ndi) [[sycl::reqd_sub_group_size(__sub_group_size)]] {
+                // Compute work distribution fields dependent on sub-group size within the kernel. This is because we
+                // can only rely on the value of __sub_group_size provided in the device compilation phase within the
+                // kernel itself.
+                const auto [__num_sub_groups_local, __num_sub_groups_global, __inputs_per_sub_group,
+                            __inputs_per_item] =
+                    __get_reduce_then_scan_sub_group_params(__work_group_size, __sub_group_size, __max_num_work_groups,
+                                                            __max_block_size, __inputs_remaining);
```
dmitriy-sobolev, line 344: This function contains a substantial number of operations. Does it affect the performance?

mmichel11: I did not see any negative performance impact here on PVC. I will verify on a few different devices. Our alternative I think would be to pre-compute values for both sub-group sizes on the host and select the correct set on the difference.  We will be transferring twice as much data to the device and use more registers, so I will evaluate this option as well.

mmichel11: I took a second look here on PVC, BMG, and A100. The performance impact is minimal as all differences I measured were close or within a single standard deviation. I can send you the data if you would like.

I think the current approach is best as it is easily adaptable to the next future addition where we remove `reqd_sub_group_size` entirely (see comments in https://github.com/uxlfoundation/oneDPL/issues/2134 where I have found how to do this performantly). The current alternative would be to precompute both sub-group size 16 and 32 paths on the host. However, with the next future work we will support any sub-group size and would need to precompute potential options based on a runtime query of the device's supported sub-group sizes and then copy to the device via a SYCL buffer or SYCL host memory. I expect this would likely perform worse than the current approach.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -266,42 +266,84 @@ __scan_through_elements_helper(const __dpl_sycl::__sub_group& __sub_group, _GenI
...
+// certain Intel graphics with older drivers.
+//
+// When querying from the host compiler, return a value of 32 always to ensure a generic non-Intel device supports this
+// sub-group size. For the case of Intel GPUs, no additional check for sub-group sizes of 16 is performed on the host as
+// we know this is supported.
+constexpr inline std::uint8_t
+__get_reduce_then_scan_sg_sz()
+{
+    return
+#if !defined(__SYCL_DEVICE_ONLY__) || _ONEDPL_DETECT_COMPILER_OPTIMIZATIONS_ENABLED || !_ONEDPL_DETECT_SPIRV_COMPILATION
```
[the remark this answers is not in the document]
adamfidel: I think it's understandable in its current form, but requires a lot of context. I think the mixing of "called from host" vs "called from device" might be introducing some confusion and it might clear things up by splitting the function into a host version and device version as you suggest.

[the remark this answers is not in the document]
dmitriy-sobolev: I think I understand it now. It is complicated.
`!_ONEDPL_DETECT_SPIRV_COMPILATION` condition properly works only during the device compilation phase.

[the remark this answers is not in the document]
dmitriy-sobolev: > Would splitting this up into __get_reduce_then_scan_sg_sz_host and __get_reduce_then_scan_sg_sz_device functions be clearer?

It will still look as it is called from host for host. I am not sure.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: Your suggestion is the simplest solution which covers the case you mention, but I prefer the current for several reasons:

1. Ultimately, the long term solution with no `reqd_sub_group_size` will need to compute these mentioned fields on the device, so `__get_reduce_then_scan_sub_group_params` will stay as is along with allocation of temporary storage to cover the case with the smallest sub-group size. Most of these changes will be present in the long-term solution with the exception that `__get_reduce_then_scan_sg_sz` will be removed and the min / max sub-group size on the host will need to be generalized based on the device's supported sub-groups.
2. Your suggestion covers the most likely scenario (clang++ default options), but the guidance I received from the compiler team is to not make any assumptions regarding optimization levels between host-and-device and vice versa. It is unlikely, but I'd rather play it safe in that regard so we do not have any other unexpected issues like this.

Feel free to let me know if you disagree, or if anyone else has thoughts on this.

dmitriy-sobolev: My main concern was that there is too much complexity in what we will not need soon, but your first point refutes it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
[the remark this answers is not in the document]
mmichel11: > I think it's understandable in its current form, but requires a lot of context. I think the mixing of "called from host" vs "called from device" might be introducing some confusion and it might clear things up by splitting the function into a host version and device version as you suggest.

I've made this change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -266,42 +266,89 @@ __scan_through_elements_helper(const __dpl_sycl::__sub_group& __sub_group, _GenI
...
+    const std::uint32_t __num_sub_groups_local = __work_group_size / __sub_group_size;
+    const std::uint32_t __num_sub_groups_global = __num_sub_groups_local * __max_num_work_groups;
+    const std::uint32_t __max_inputs_per_subgroup = __max_block_size / __num_sub_groups_global;
+    const std::uint32_t __evenly_divided_remaining_inputs =
+        std::max(std::size_t{__sub_group_size},
+                 oneapi::dpl::__internal::__dpl_bit_ceil(__inputs_remaining) / __num_sub_groups_global);
+    const std::uint32_t __inputs_per_sub_group =
+        __inputs_remaining >= __max_block_size ? __max_inputs_per_subgroup : __evenly_divided_remaining_inputs;
+    const std::uint32_t __inputs_per_item = __inputs_per_sub_group / __sub_group_size;
+    return std::make_tuple(__num_sub_groups_local, __num_sub_groups_global, __inputs_per_sub_group, __inputs_per_item);
```
dmitriy-sobolev, line 306: Could you include `<tuple>` for `std::make_tuple`?

dmitriy-sobolev (commented):
I've left a small suggestion. Other than that, the PR looks good to me.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: Thanks, done.

dmitriy-sobolev (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -777,7 +825,8 @@ __parallel_transform_reduce_then_scan(oneapi::dpl::__internal::__device_backend_
         __reduce_then_scan_scan_kernel<_CustomName>>;
     using _ValueType = typename _InitType::__value_type;
 
-    constexpr std::uint8_t __sub_group_size = __get_reduce_then_scan_sg_sz();
+    constexpr std::uint8_t __min_sub_group_size = 16;
+    constexpr std::uint8_t __max_sub_group_size = 32;
```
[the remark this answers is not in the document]
mmichel11: Yes, added.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -460,9 +507,19 @@ struct __parallel_reduce_then_scan_scan_submitter<
 
             __cgh.parallel_for<_KernelName...>(
                     __nd_range, [=, *this] (sycl::nd_item<1> __ndi) [[sycl::reqd_sub_group_size(__sub_group_size)]] {
+                // Compute work distribution fields dependent on sub-group size within the kernel. This is because we
+                // can only rely on the value of __sub_group_size provided in the device compilation phase within the
+                // kernel itself.
+                const auto [__num_sub_groups_local, __num_sub_groups_global, __inputs_per_sub_group,
+                            __inputs_per_item] =
+                    __get_reduce_then_scan_sub_group_params(__work_group_size, __sub_group_size, __max_num_work_groups,
+                                                            __max_block_size, __inputs_remaining);
```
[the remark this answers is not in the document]
mmichel11: There is no reason to do things the way I have them now. I have converted it to a struct and switched to use its members throughout reduce-then-scan.

danhoeflinger (approved):
LGTM assuming green CI follows

mmichel11:
Will run through internal CI once more to confirm there are no issues and will then merge.

Issue #2023, opened 2025-01-27
Reconsider the use of inheritance for compile time constants in SYCL backend parallel-for bricks


mmichel11:
@MikeDvorskiy made a very good design suggestion in the review of https://github.com/uxlfoundation/oneDPL/pull/1976 in the following comment: https://github.com/uxlfoundation/oneDPL/pull/1976/files#r1929054989. Due to the closeness of the 2022.8.0 milestone and amount of change this entails, I suggested we defer this potential design change as a refactoring in the next milestone if it proves beneficial which was agreed upon offline.

For context, the PR introduces the use of two tuning classes, `walk_scalar_base` and `walk_vector_or_scalar_base`, which define the following compile-time parameters, `__can_vectorize`, `__preferred_iters_per_item`, `__preferred_vector_size` based on a set of ranges provided through the class' template parameters. These compile-time parameters are queried by the parallel-for implementation to define strides and access patterns for use in the kernel. Individual bricks inherit from these base classes and provide a set of ranges, so they do not need to manually define these tuning parameters themselves.

This design approach has a downside of requiring that the types of the underlying ranges are provided through the class-template when they were previously deduced from the function call operator. This results in more code at the hetero pattern level as more template parameters must be provided by the caller. For instance, `__pattern_walk_3`'s call to parallel for has changed from this:
```
    auto __future =
        oneapi::dpl::__par_backend_hetero::__parallel_for(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
                                                          unseq_backend::walk_n<_ExecutionPolicy, _Function>{__f}, __n,
                                                          __buf1.all_view(), __buf2.all_view(), __buf3.all_view());
```
to
```
    auto __future = oneapi::dpl::__par_backend_hetero::__parallel_for(
        _BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
        unseq_backend::walk3_vectors_or_scalars<_ExecutionPolicy, _Function, decltype(__buf1.all_view()),
                                                decltype(__buf2.all_view()), decltype(__buf3.all_view())>{
            __f, static_cast<size_t>(__n)},
        __n, __buf1.all_view(), __buf2.all_view(), __buf3.all_view());
```
which results in more code and decreased readability.

[...]

PR #2136, opened 2025-03-18
Remove use of inheritance from parallel-for bricks in SYCL backend

mmichel11:
This PR addresses https://github.com/uxlfoundation/oneDPL/issues/2023.

In https://github.com/uxlfoundation/oneDPL/pull/1976, parallel-for bricks were redesigned to have an inheritance relationship with a either a scalar or vector tuning base class in order to inherit compile-time tuning constants. This came with a downside of adding the user-provided range types to the class template of parallel-for bricks, complicating usage from the caller side as more templates must be provided when creating a brick.

This PR removes the use of inheritance from parallel-for bricks along with the range class templates with the following approach:

- Each parallel-for brick defines two compile-time boolean constants, `__can_vectorize` and `__can_process_multiple_iters`. `__can_vectorize` defines whether a vectorization path is provided. For some bricks (e.g. `__custom_brick` for binary search) there is no benefit of providing a vectorization path. `__can_process_multiple_iters` specifies whether a single item can safely process multiple strided iterations of input.
- The parallel-for SYCL implementation makes use of a new class `__pfor_params` which queries the brick's static members to define vectorization and iters per item tuning parameters.
- An instance of `__pfor_params` is passed to the brick when invoking the function call operator which includes vector sizes and whether vectorization should be enabled.

Additionally, the `__scalar_path_impl`, `__vector_path_impl`, and dispatch `operator()` design of the bricks have been replaced with two `operator()` definitions using SFINAE to define scalar and vector implementations dependent on whether the provided parameter instance is vectorizable.

### DIFF
--- a/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
+++ b/include/oneapi/dpl/internal/async_impl/async_impl_hetero.h
@@ -44,7 +44,7 @@ __pattern_walk1_async(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _For
 
     auto __future_obj = oneapi::dpl::__par_backend_hetero::__parallel_for(
         _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
-        unseq_backend::walk1_vector_or_scalar<_ExecutionPolicy, _Function, decltype(__buf.all_view())>{
+        unseq_backend::walk1_vector_or_scalar<_ExecutionPolicy, _Function>{
             __f, static_cast<std::size_t>(__n)},
         __n, __buf.all_view());
     return __future_obj;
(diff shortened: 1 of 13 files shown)


SergeyKopienko:
@mmichel11 Does we really have some benefits of using `std::enable_if_t<_Params::__b_vectorize` in 9 places in comparison `if constexpr` ?

mmichel11:
> @mmichel11 Does we really have some benefits of using `std::enable_if_t<_Params::__b_vectorize` in 9 places in comparison `if constexpr` ?

I will take a look at using `if constexpr` instead and see if that is cleaner than the current approach. For the brick's with a small function body, it should be.

mmichel11:
Closing, as this PR is no longer needed.

Issue #2139, opened 2025-03-24
Incorrect Kernel name generation for `__parallel_reduce_then_scan_reduce_submitter` and for `__parallel_reduce_then_scan_scan_submitter`


SergeyKopienko:
I think have place incorrect Kernel name generation for `__parallel_reduce_then_scan_reduce_submitter` and for `__parallel_reduce_then_scan_scan_submitter` here:

https://github.com/uxlfoundation/oneDPL/blob/148d91f55f9add15cf507a2d54b9f1b2b57bcbc1/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h#L774

Justification: these Kernel names should depends on all used template parameters of ` struct __parallel_reduce_then_scan_reduce_submitter` and for `struct __parallel_reduce_then_scan_scan_submitter` (accordingly) and also from template params of their `operator()` - so they should describe all available instantiations of ` struct __parallel_reduce_then_scan_reduce_submitter` and `struct __parallel_reduce_then_scan_scan_submitter`.

SergeyKopienko:
@mmichel11 what do you think?

SergeyKopienko:
It's not an error: some things like `_ReduceKernel` in this example may depends on only from `oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>`

mmichel11:
> [@mmichel11](https://github.com/mmichel11) what do you think?

I am in agreement that there is no issue here.

PR #2140, opened 2025-03-24
Incorrect Kernel name generation for `__parallel_reduce_then_scan_reduce_submitter` and for `__parallel_reduce_then_scan_scan_submitter`

SergeyKopienko:
In this PR we fixing the issue https://github.com/uxlfoundation/oneDPL/issues/2139

Previously we had next Kernel names generation for these structures:
```C++
    using _ReduceKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
        __reduce_then_scan_reduce_kernel<_CustomName>>;
    using _ScanKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
```
It's not enough to describe all possible instantiations of `struct __parallel_reduce_then_scan_reduce_submitter`, `struct __parallel_reduce_then_scan_scan_submitter` and their `operator()` instantiations.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -417,15 +417,8 @@ struct __parallel_reduce_then_scan_reduce_submitter
 
 template <std::uint8_t __sub_group_size, std::uint16_t __max_inputs_per_item, bool __is_inclusive,
           bool __is_unique_pattern_v, typename _ReduceOp, typename _GenScanInput, typename _ScanInputTransform,
-          typename _WriteOp, typename _InitType, typename _KernelName>
-struct __parallel_reduce_then_scan_scan_submitter;
-
-template <std::uint8_t __sub_group_size, std::uint16_t __max_inputs_per_item, bool __is_inclusive,
-          bool __is_unique_pattern_v, typename _ReduceOp, typename _GenScanInput, typename _ScanInputTransform,
-          typename _WriteOp, typename _InitType, typename... _KernelName>
-struct __parallel_reduce_then_scan_scan_submitter<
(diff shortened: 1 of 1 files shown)


SergeyKopienko:
@mmichel11, should we also include this fix into already released version of `oneDPL` ?

SergeyKopienko:
Probably this issue has been introduced in the PR https://github.com/uxlfoundation/oneDPL/pull/2046

danhoeflinger:
Why do we need kernel name generator for this case (rather than kernel name provider)?

If someone passes in a policy with the same external kernel name in to two different kernels then its their error not ours, right?

SergeyKopienko:
@danhoeflinger Please just find in the code how `_ReduceKernel` and `_ScanKerne` later used. They passed into `parallel_for` as Kernel names without any changes.

danhoeflinger:
> @danhoeflinger Please just find in the code how `_ReduceKernel` and `_ScanKerne` later used. They passed into `parallel_for` as Kernel names without any changes.

OK, so why is that a problem?  
We separate the reduce and scan kernels using different wrappers on the user provided custom name.  


Lets go through the options:
1) A user provides a unique kernel name, and unnamed lambdas are permitted:  Works
2) A user provides a unique kernel name, and unnamed lambdas are NOT permitted:  Works
3) A user uses default policy name, and unnamed lambdas are permitted: Works, because `__kernel_name_provider` uses `__optional_kernel_name<>`, and then the submitters partially specialize to get `_KernelName...` pack, and provide that to the kernel, where it is an empty pack, so it uses an unnamed lambda
4) A user uses the default policy name and unnamed lambda are NOT permitted:  Fails, (correctly) because it is the user's responsibility to provide a unique kernel name to the policy for each call when using no-unamed-lambdas.
5) A user uses the same custom named policy for two oneDPL API calls:  Fails, (correctly) because if custom names are provided they must be unique for each call.

Are there any other options, or any options which you disagree with the expected outcome?

SergeyKopienko:
It's not a bug: it's by design.

PR #2146, opened 2025-03-26
Remove `ExecutionPolicy` template param from internal oneDPL buffers

SergeyKopienko:
In this PR we remove `ExecutionPolicy` template param from internal oneDPL buffers as not required:
- from `__buffer` template aliases;
- from `class __buffer_impl` (both implementations);
- from `__buffer` instances in oneDPL code.


danhoeflinger (approved):
Yeah, this LGTM.  
I think this in line with the offline decision to remove Policy from touching bricks, etc.  where it is not needed.  
If a future parallel backend requires `ExecutionPolicy` to be passed in, then we will need to figure out how to handle that then.

MikeDvorskiy (approved):
LGTM

PR #2148, opened 2025-03-27
Removal `ExecutionPolicy` template parameter from `brick`s

SergeyKopienko:
In this PR we remove `ExecutionPolicy` template parameter from bricks according to our offline discussion.

Other linked PR's: https://github.com/uxlfoundation/oneDPL/pull/2152, https://github.com/uxlfoundation/oneDPL/pull/2153

Issue: https://github.com/uxlfoundation/oneDPL/issues/2110

PR #2152, opened 2025-03-31
Removal `ExecutionPolicy` template parameter from sycl utilities

SergeyKopienko:
In this PR we remove `ExecutionPolicy` template parameter from bricks according to our offline discussion.

As part of these changes, in this PR we modified some functions in the ` class __result_and_scratch_storage_impl` - we remove `const` declarations from them (https://github.com/uxlfoundation/oneDPL/pull/2152/files#r2020767693, https://github.com/uxlfoundation/oneDPL/pull/2152/files#r2020768635 and https://github.com/uxlfoundation/oneDPL/pull/2152/files#r2020772661) because they are using non-const methods of `sycl::queue` - `sycl::queue::memcpy`.

Other linked PR's: https://github.com/uxlfoundation/oneDPL/pull/2148, https://github.com/uxlfoundation/oneDPL/pull/2153

### Note
Multiple `__exec.queue()` calls in some functions isn't important here because later we going to remove them at all and replace `policy` by `sycl::queue`.


include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -377,7 +377,7 @@ __radix_sort(sycl::queue __q, _RngPack1&& __pack_in, _RngPack2&& __pack_out, _Ke
     const auto __n = __pack_in.__keys_rng().size();
     assert(__n > 0);
 
-    // _PRINT_INFO_IN_DEBUG_MODE(__exec); TODO: extend the utility to work with queues
+    _PRINT_INFO_IN_DEBUG_MODE(__q);
```
SergeyKopienko, line 380: @dmitriy-sobolev, is this change here OK from your point of view?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -657,7 +643,7 @@ struct __result_and_scratch_storage_impl : __result_and_scratch_storage_base
     }
 
     _T
-    __wait_and_get_value(sycl::event __event) const
+    __wait_and_get_value(sycl::event __event)
```
SergeyKopienko, line 646: We remove `const` here because `__get_value()` function is non-const now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -712,7 +698,7 @@ struct __result_and_scratch_storage_impl : __result_and_scratch_storage_base
     }
 
     virtual std::size_t
-    __get_data(sycl::event __event, std::size_t* __p_buf) const override
+    __get_data(sycl::event __event, std::size_t* __p_buf) override
```
SergeyKopienko, line 701: We remove `const` here because we calls `__get_value()` function which is non-cost now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -678,7 +664,7 @@ struct __result_and_scratch_storage_impl : __result_and_scratch_storage_base
         else if (__supports_USM_device)
         {
             _T __tmp;
-            __exec.queue().memcpy(&__tmp, __scratch_buf.get() + __scratch_n + idx, 1 * sizeof(_T)).wait();
+            __q.memcpy(&__tmp, __scratch_buf.get() + __scratch_n + idx, 1 * sizeof(_T)).wait();
```
MikeDvorskiy, line 667: I would propose to use 
`const_cast<__result_and_scratch_storage*>(this)->__q.memcpy(&__tmp, __scratch_buf.get() + __scratch_n + idx, 1 * sizeof(_T)).wait();` 
here... and keep `const` modificatory in the several other places, where the const was removed.

SergeyKopienko: One more variant - to rewrite this as
```C++
            const_cast<sycl::queue&>(__q)
                .memcpy(&__tmp, __scratch_buf.get() + __scratch_n + idx, 1 * sizeof(_T))
                .wait();
```
and restore `const` methods

MikeDvorskiy: My point is that the methods 

1) `__get_value, __wait_and_get_value, __get_data` are data reading function and don't modify the instance of `__result_and_scratch_storage`. With that perspective the mentioned methods should be marked as const.

2) But, `sycl::queue::memcpy` method is defined as non-const according to the SYCl spec. Probably, it is because `memcpy` does submit a task to a queue and returns a sycl event. But it a kind of the SYCL implementation derails.

So, we have 3 ways:
1) To Remove "const" from  `__get_value, __wait_and_get_value, __get_data` as in this PR.
2) To make `__q` field as mutable.
3) To use `const_cast` just in the place where we really need it
`const_cast<queue&>(__q).memcpy(...)`

Let's discuss it.
I would vote for (3).

SergeyKopienko: @akukanov, @danhoeflinger, @mmichel11 what do you think about this?

danhoeflinger: I don't have a very passionate opinion here, but unless we currently need `__get_value`, `__wait_and_get_value`, `__get_data` to be `const` functions, or unless we foresee a need for them to be I would just vote for (1).  

My reasoning being that I have a general preference against modifying constness via `mutable` / `const_cast` unless there is a very good reason to do so.

SergeyKopienko: So this looks like option (1) is our common decision and I save this code as is (without `const`-declared methods).
@MikeDvorskiy , @danhoeflinger, @mmichel11 could you please continue your reviews of this PR?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -48,84 +48,77 @@ namespace __internal
...
 
 #if _ONEDPL_DEBUG_SYCL
-template <typename _ExecutionPolicy>
-::std::string
-__device_info(const _ExecutionPolicy& __policy)
+inline std::string
+__device_info(const sycl::queue& __q)
 {
-    return __policy.queue().get_device().template get_info<sycl::info::device::name>();
+    return __q.get_device().template get_info<sycl::info::device::name>();
```
mmichel11, line 54: Since we no longer have the `_ExecutionPolicy` template, I think the scenario is no longer ambiguous and `template` can be removed here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -48,84 +48,77 @@ namespace __internal
...
 #endif
 
-template <typename _ExecutionPolicy>
-std::size_t
-__max_work_group_size(const _ExecutionPolicy& __policy, std::size_t __wg_size_limit = 8192)
+inline std::size_t
+__max_work_group_size(const sycl::queue& __q, std::size_t __wg_size_limit = 8192)
 {
-    std::size_t __wg_size = __policy.queue().get_device().template get_info<sycl::info::device::max_work_group_size>();
+    std::size_t __wg_size = __q.get_device().template get_info<sycl::info::device::max_work_group_size>();
```
mmichel11, line 61: Same comment with regards to the `template` keyword.

SergeyKopienko: https://github.com/uxlfoundation/oneDPL/issues/2173

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
akukanov: My preference is for (2) (declare the queue as mutable) if the class has some other state and we want to ensure that state is not modified.

If the queue is the only field of the class, we might just go with (1) and say its methods are non-const.

However if `__result_and_scratch_storage` is in turn included into some other types as a field, and by making its methods non-const we would just "move" the issue to these other types, then I would again keep the methods const and change the queue to be mutable.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: `const` restored by declaring `sycl::queue` as `mutable`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: `const` restored by declaring `sycl::queue` as `mutable`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -536,7 +524,7 @@ struct __result_and_scratch_storage_impl : __result_and_scratch_storage_base
     using __accessor_t =
         sycl::accessor<_T, 1, _AccessMode, __dpl_sycl::__target_device, sycl::access::placeholder::false_t>;
 
-    _ExecutionPolicy __exec;
+    mutable sycl::queue __q;
```
SergeyKopienko, line 527: I declared this `sycl::queue __q` as `mutable` because their non-const method `sycl::queue::memcpy` using in const-methods of this `struct __result_and_scratch_storage`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -668,7 +654,7 @@ struct __result_and_scratch_storage_impl : __result_and_scratch_storage_base
     // Note: this member function assumes the result is *ready*, since the __future has already
     // waited on the relevant event.
     _T
-    __get_value(size_t idx = 0) const
+    __get_value(size_t idx = 0)
```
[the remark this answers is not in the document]
SergeyKopienko: `const` has been restored by declaring `sycl::queue __q` as `mutable`

MikeDvorskiy (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -547,8 +535,8 @@ struct __result_and_scratch_storage_impl : __result_and_scratch_storage_base
     bool __supports_USM_device;
 
     // Only use USM host allocations on L0 GPUs. Other devices show significant slowdowns and will use a device allocation instead.
-    inline bool
-    __use_USM_host_allocations(sycl::queue __queue)
+    static bool
```
danhoeflinger, line 538: Its a little strange to me to have these two private `static` functions.  Can we just call these from within the constructor and use the queue member field rather than passing in a queue as an argument?

(I understand this is not the purpose of this PR, but marking them `static` shows how odd this is...)

mmichel11 (approved):
LGTM as well.

PR #2155, opened 2025-03-31
[Prototype] Fix scan perf regression and improve set performance

mmichel11:
I noticed 5-6% perf regressions affecting scalability of scan. For `__max_inputs_per_item`, we used to process 128 elements (for 4-byte case), but with new approach only 64 are processed.

Bumping the magic constant from 256 to 512 also improves set performance. I observed ~9% improvements for `set_intersection`.


danhoeflinger (approved):
LGTM, feel free to merge.

PR #2156, opened 2025-03-31
[RNG] Small refactoring of Philox engine implementation and updates in the testing 

ElenaTyuleneva:
**The scope of the changes:**
1. [Philox] Corrected `set_counter` API with setting `idx` to `n-1` according to the C++ proposal's wording:
{70E40BC6-4D83-4B8C-9737-5B9DA35BBF99}
2. [Testing] Generated reference numbers of the 10000th element for Philox with `word_count = 2`  and extended `philox_test.pass`.
3. [Testing] Changed the `philox2x32` and `philox2x64` aliases in the tests.

### DIFF
--- a/include/oneapi/dpl/internal/random_impl/philox_engine.h
+++ b/include/oneapi/dpl/internal/random_impl/philox_engine.h
@@ -134,6 +134,7 @@ class philox_engine
             // all counters are set in reverse order
             state_.X[word_count - __i - 1] = __counter[__i] & in_mask;
         }
+        state_.idx = word_count - 1;
     }
 
     /* Generating functions */
@@ -460,18 +461,15 @@ operator<<(std::basic_ostream<__CharT, __Traits>& __os,
     __CharT __sp = __os.widen(' ');
     __os.fill(__sp);
 
-    for (auto x_elm : __engine.state_.X)
-    {
-        __os << x_elm << __sp;
-    }
     for (auto k_elm : __engine.state_.K)
     {
         __os << k_elm << __sp;
     }
-    for (auto y_elm : __engine.state_.Y)
+    for (auto x_elm : __engine.state_.X)
     {
-        __os << y_elm << __sp;
+        __os << x_elm << __sp;
     }
+
     __os << __engine.state_.idx;
 
     return __os;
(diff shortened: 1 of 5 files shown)


include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -508,7 +512,12 @@ operator>>(std::basic_istream<__CharT, __Traits>& __is, philox_engine<__UIntType
 
     __is.setf(std::ios_base::dec);
 
-    const std::size_t __state_size = 2 * __n + __n / 2 + 1;
+    /* Number of elements in X, K and Y */
+    constexpr std::size_t __x_elm_num = __n;
+    constexpr std::size_t __k_elm_num = __n / 2;
+    constexpr std::size_t __y_elm_num = __n;
+
+    constexpr std::size_t __state_size = __x_elm_num + __k_elm_num + __y_elm_num + 1;
```
andreyfe1, line 520: There are 4 registers instead of one. Do we really need to take more register memory here? The problem with oneMKL Device API Philox now is that it takes too much register memory

include/oneapi/dpl/internal/random_impl/philox_engine.h
```diff
@@ -508,7 +512,12 @@ operator>>(std::basic_istream<__CharT, __Traits>& __is, philox_engine<__UIntType
 
     __is.setf(std::ios_base::dec);
 
-    const std::size_t __state_size = 2 * __n + __n / 2 + 1;
+    /* Number of elements in X, K and Y */
+    constexpr std::size_t __x_elm_num = __n;
```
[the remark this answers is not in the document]
ElenaTyuleneva: Hm, I identified this method as the only one that uses  `__n` in several places in a way that may cause errors - for calculating `__state_size` and below in the write loop. With the variables, size assignment for each field of the state is done only once and the code becomes clearer, but it's IMHO, I can revert it, since as @andreyfe1 specified below, it may cause problems with performance.

include/oneapi/dpl/internal/random_impl/philox_engine.h
ElenaTyuleneva: > There are 4 registers instead of one. Do we really need to take more register memory here? The problem with oneMKL Device API Philox now is that it takes too much register memory. 

Explained the reasoning in the comment above, thanks for pointing this out!

andreyfe1: Actually, variables are `constexpr`. So, currently I don't think it's a real issue. Now I see no concerns with such variables

andreyfe1 (approved):

include/oneapi/dpl/internal/random_impl/philox_engine.h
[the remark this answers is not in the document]
ElenaTyuleneva: Done in [ecbefa9](https://github.com/uxlfoundation/oneDPL/pull/2156/commits/ecbefa9e7802af58122efee5a3c912c2276afa3e)

aelizaro (approved):

ElenaTyuleneva:
> Have you measured the performance on CPU/GPU? What is the speedup/slowdown?


The main computation functions(e.g., permutation and multiplication) are not changed, so I don't expect the perf will differ.

andreyfe1 (approved):

Issue #2006, opened 2025-01-17
Consider combining implementations of `walk*_vectors_or_scalars` with fold expressions


mmichel11:
I *think* it should be possible to combine `walk*_vectors_or_scalars` together with some complicated fold instructions, lambdas, tuples, and std::apply.  

Take a look at the first answer of https://stackoverflow.com/questions/7230621/how-can-i-iterate-over-a-packed-variadic-template-argument-list.  I think you should do something similar, chaining together instructions by returning tuples and then with `std::apply`.

Here is an example I was playing with.  
https://godbolt.org/z/vc8dK4ed6

In the end, I'm not sure if (1) its actually possible and (2) its worth the complexity to consolidate these structs, but its worth considering...

_Originally posted by @danhoeflinger in https://github.com/uxlfoundation/oneDPL/pull/1976#discussion_r1900203587_

The current implementation of `walk*_vectors_or_scalars` introduced in https://github.com/uxlfoundation/oneDPL/pull/1976 manually implements each path separately. This is due to the difficulty of combining the separate vector paths in a single unified implementation. However, there is a common structure shared among all three functors and @danhoeflinger's recommendation looks implementable.

When exploring combining the implementations, we should consider readability / complexity of the fold instruction approach when determining if we should adopt it. Additionally, the `walk1_vector_or_scalar` implementation should avoid explicit copying (as currently implemented), so the user's unary functor in `std::for_each` can still directly modify the reference.

mmichel11:
We can definitely do this with a combination of fold expressions and integer sequences. However, it should not be done for the in-place (walk1) case as it is unique in that we must preserve the in-place modification semantics.

Here is a prototype in how it can be done: https://godbolt.org/z/o1nscnn6s. I have also started a branch which removes `walk1_vectors_or_scalars`, `walk2_vectors_or_scalars`, and `walk3_vectors_or_scalars` and replaces them with a single `walk_n_vectors_or_scalars`: https://github.com/uxlfoundation/oneDPL/tree/dev/mmichel11/pfor_walk_n_vectors_or_scalars.

The question here is if adopting this more generic approach is worth it over two simpler overloads for the case of two and three range parameters. I will think a bit more.

mmichel11:
I have two branches open:

https://github.com/uxlfoundation/oneDPL/tree/dev/mmichel11/pfor_walk_n_vectors_or_scalars
[...]

PR #2100, opened 2025-03-05
[onedpl][ranges] oneDPL range based API: M algorithms set implementation

MikeDvorskiy:
[onedpl][ranges] oneDPL range based API: M algorithms set: fill, ~generate~, move, replace, replace_if, remove, remove_if, mismatch, minmax_element, minmax, min, max, find_first_of, find_end, is_sorted_until

PR #2165, opened 2025-04-06
Refactor parallel for brick functors in the SYCL backend

mmichel11:
This PR removes the use of inheritance from parallel-for bricks along with the range class templates with the following approach:

Each parallel-for brick defines two compile-time boolean constants, `__can_vectorize` and `__can_process_multiple_iters`. `__can_vectorize` defines whether a vectorization path is provided. For some bricks (e.g. `__custom_brick` for binary search) there is no benefit of providing a vectorization path. `__can_process_multiple_iters` specifies whether a single item can safely process multiple strided iterations of input.

The parallel-for SYCL implementation makes use of a new class __pfor_params which queries the brick's static members to define vectorization and iters per item tuning parameters.
An instance of `__pfor_params` is passed to the brick when invoking the function call operator which includes vector sizes and whether vectorization should be enabled.

Additionally, the following three walk functors:
- `walk1_vectors_or_scalars`
- `walk2_vectors_or_scalars`
- `walk3_vectors_or_scalars`

are replaced with a single:
- `walk_n_vectors_or_scalars`

This simplifies the implementation of the walk patterns as a single brick now supports the three different cases and reduces code duplication in the unseq SYCL backend. To implement `walk_n_vectors_or_scalars`, the cases of 1, 2, and 3 range packs are implemented in the brick itself.

An alternative, generic approach of implementing the case of 2 range packs and greater using fold expressions was explored (see details in https://github.com/uxlfoundation/oneDPL/issues/2006 and [the proof-of-concept branch](https://github.com/uxlfoundation/oneDPL/tree/dev/mmichel11/pfor_walk_n_vectors_or_scalars_v2). It is implementable but in my opinion was overly complex and did not gain us much, so I favor the simpler approach in this PR.


SergeyKopienko:
Probably some data in the [struct __brick_shift_left](https://github.com/uxlfoundation/oneDPL/blob/b34453d0510733fd8415842c2ae87128b7cedb48/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L1420) not required anymore.
At least - `constexpr static std::uint8_t __max_vector_size = 4;`.

mmichel11:
> Probably some data in the [struct __brick_shift_left](https://github.com/uxlfoundation/oneDPL/blob/b34453d0510733fd8415842c2ae87128b7cedb48/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h#L1420) not required anymore. At least - `constexpr static std::uint8_t __max_vector_size = 4;`.

I fixed it in #2136 and rebased this branch on top of it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
```diff
@@ -67,10 +90,11 @@ struct __parallel_for_small_submitter<__internal::__optional_kernel_name<_Name..
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
 
             __cgh.parallel_for<_Name...>(sycl::range</*dim=*/1>(__count), [=](sycl::item</*dim=*/1> __item_id) {
+                // Disable vectorization and multiple iterations per item within the brick to evenly spread work across
+                // compute units.
+                __pfor_params<false, false, _Ranges...> __params;
```
adamfidel, line 95:
```suggestion
                __pfor_params<false /*__can_vectorize_brick*/, false /*__can_brick_process_multiple_iters*/, _Ranges...> __params;
```
Maybe this will make it a little clearer.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h
```diff
@@ -67,10 +68,12 @@ struct __parallel_for_fpga_submitter<__internal::__optional_kernel_name<_Name...
             oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
 
             __cgh.single_task<_Name...>([=]() {
+                // Disable vectorization and multiple iterations per item.
+                __pfor_params<false, false, _Ranges...> __params;
```
adamfidel, line 72:
```suggestion
                __pfor_params<false /*__can_vectorize_brick*/, false /*__can_brick_process_multiple_iters*/, _Ranges...> __params;
```
```

File: include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h (hunk 1 above)
Comment (mmichel11) [reply include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h]: Done here and in 2136.

File: include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h (hunk 2 above)
Comment (mmichel11) [reply include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_fpga.h]: Done here and in 2136.

File: include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h [hunk 3]
```diff
@@ -113,160 +114,62 @@ struct walk_n
...
+        _ValueType1 __rng1_vector[_Params::__vector_size];
+        _ValueType2 __rng2_vector[_Params::__vector_size];
 
-        typename __base_t::__vec_load_t __vec_load{__n};
-        typename __base_t::__vec_store_t __vec_store{__n};
+        __vector_load<_Params::__vector_size> __vec_load{__n};
+        __vector_store<_Params::__vector_size> __vec_store{__n};
         oneapi::dpl::__par_backend_hetero::__scalar_load_op __load_op;
 
         // 1. Load inputs into vectors
```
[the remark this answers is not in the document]
mmichel11: Sure, I have made this change throughout the PR as well.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -113,160 +114,62 @@ struct walk_n
...
-
-  public:
-    walk2_vectors_or_scalars(_F __f, std::size_t __n) : __f(std::move(__f)), __n(__n) {}
-
-    template <typename _IsFull>
+    template <typename _IsFull, typename _Params, typename _Range1, typename _Range2,
+              std::enable_if_t<_Params::__b_vectorize, int> = 0>
     void
-    __vector_path_impl(_IsFull __is_full, const std::size_t __idx, _Range1 __rng1, _Range2 __rng2) const
+    __vector_impl(_IsFull __is_full, const std::size_t __idx, _Params, _Range1&& __rng1, _Range2&& __rng2) const
```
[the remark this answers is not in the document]
mmichel11: I did this renaming.

mmichel11:
Many of the suggestions apply to the base PR #2136, which I added there and rebased to get in this branch. I recommend looking at 2136 first, or I can close that PR and just use this larger one if it is preferred.

mmichel11:
@SergeyKopienko @MikeDvorskiy @adamfidel @danhoeflinger @dmitriy-sobolev 
There are many conflicts with the main branch and rebasing on top of #2136 after rebasing that branch on main has become a bit tricky.

Since most of the review has occurred here, are there any objections with closing the intermediate #2136 PR, and reviewing all the refactoring here? This will also reduce the unnecessary complexity of refactoring walk*_vectors_or_scalars only to replace it again in this patch.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
```diff
@@ -37,6 +37,28 @@ namespace dpl
 namespace __par_backend_hetero
 {
 
+template <bool __can_vectorize_brick, bool __can_brick_process_multiple_iters, typename... _Ranges>
```
danhoeflinger, line 40: Does it make sense to instead accept a `typename __brick` and a `bool __attempt_vectorization`?
instead of 2 booleans?

It seems like the vectorization is either dictated directly by the brick, or turned off wholesale by the implementation path.  Then here within `__pfor_params` you can assume `__can_vectorize` and `__can_process_multiple_iters` exist and use them.  If we really want to get fancy we could even make a defaulted `brick_traits<brick>` struct which could provide defaults (`false`) for bricks which don't define these themselves.

I'm not sure this option is better, but it may better follow the logic of what is going on in the caller.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -565,9 +544,9 @@ struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::unseq_backen
 {
 };
 
-template <typename _BinaryOperator, typename _Size, typename _Range>
+template <typename _BinaryOperator, typename _Size>
 struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::unseq_backend::__brick_reduce_idx, _BinaryOperator,
```
danhoeflinger, line 548: I'm glad we are getting rid of the dependence of these structs on the `Ranges` and transferring that to the call operator instead.  It better avoids errors like this one where we were making device copyability depend on `_Range` which was wrong.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
mmichel11: I think passing the brick and a single boolean the most sense. I have renamed it `__enable_tuning` because it applies to both vectorization and processing multiple iterations per work-item. In the case `__enable_tuning` is false, we process one input per work-item to evenly distribute work for small problem sizes.

I also considered having a brick trait, but I favor the current approach where the these boolean variables are defined near the scalar and vectorized implementations in the brick itself.

SergeyKopienko:
@mmichel11, I think probably it's time to merge main branch into the branch for this PR - there are a lot of new commits in the main branch. You may have conflicts...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
```diff
@@ -127,45 +151,48 @@ struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name..
...
             const std::size_t __num_groups = oneapi::dpl::__internal::__dpl_ceiling_div(
                 __count, (__work_group_size * __vector_size * __iters_per_work_item));
             __cgh.parallel_for<_Name...>(
                 sycl::nd_range(sycl::range<1>(__num_groups * __work_group_size), sycl::range<1>(__work_group_size)),
                 [=](sycl::nd_item</*dim=*/1> __item) {
+                    __params_t __params;
                     const auto [__idx, __stride, __is_full] =
                         __stride_recommender(__item, __count, __iters_per_work_item, __vector_size, __work_group_size);
                     __strided_loop<__iters_per_work_item> __execute_loop{static_cast<std::size_t>(__count)};
                     if (__is_full)
```
SergeyKopienko, line 188: What about to add one more overload into `struct __strided_loop` for the case where nothing to do and calculate here proper param value in ternary operator + have only one call of `__execute_loop` without `if` calls?

include/oneapi/dpl/internal/binary_search_impl.h
```diff
@@ -74,21 +77,15 @@ struct __custom_brick : oneapi::dpl::unseq_backend::walk_scalar_base<_Range>
             get<2>(acc[idx]) = (value != end_orig) && (get<1>(acc[idx]) == get<0>(acc[value]));
         }
     }
-    template <typename _IsFull, typename _ItemId, typename _Acc>
+    template <typename _IsFull, typename _Params, typename _Acc>
     void
-    __scalar_path_impl(_IsFull, _ItemId idx, _Acc acc) const
+    operator()(_IsFull, const std::size_t idx, _Params, _Acc acc) const
     {
         if (use_32bit_indexing)
```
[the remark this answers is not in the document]
mmichel11: Initially, when tuning the binary-search algorithms this was done, but we decided the performance impact was small enough for this additional kernel and JIT time to not be worth it. Please see this discussion: 

https://github.com/uxlfoundation/oneDPL/pull/1446#discussion_r1566328996

I had taken another look as well during the first stage of parallel-for refactoring and still have the same thoughts.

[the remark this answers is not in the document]
SergeyKopienko: OK, up to you.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
mmichel11: I added a similar suggestion to this. I think it is better to improve the safety of the `false_type{}` case of `__strided_loop` to avoid underflow for unsigned types so we do not need the outside `else if (__idx < __count)` check. With this change, we can use the ternary as suggested which is cleaner and future uses will be protected from the dangerous scenario where `__idx > __count`.

There is no change in performance.

SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
```diff
@@ -127,46 +151,42 @@ struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name..
...
-                    {
-                        __execute_loop(std::true_type{}, __idx, __stride, __brick, __rngs...);
-                    }
-                    // If we are not full, then take this branch only if there is work to process.
-                    else if (__idx < __count)
-                    {
-                        __execute_loop(std::false_type{}, __idx, __stride, __brick, __rngs...);
-                    }
+                    __is_full ? __execute_loop(std::true_type{}, __idx, __stride, __brick, __params, __rngs...)
+                              : __execute_loop(std::false_type{}, __idx, __stride, __brick, __params, __rngs...);
```
danhoeflinger, line 189: Is it measurably faster to use a ternary operator here or something?
I find it less readable personally to write it this way rather than `if`, `else`.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -113,260 +114,154 @@ struct walk_n
...
-    // To achieve full bandwidth utilization, multiple iterations need to be processed by a work item
-    constexpr static std::uint8_t __preferred_iters_per_item =
-        __bytes_per_item / (__min_type_size * __preferred_vector_size);
-};
-
-template <typename _F, typename _Range>
-struct walk1_vector_or_scalar : public walk_vector_or_scalar_base<_Range>
+// If read accessor returns temporary value then __no_op returns lvalue reference to it.
+// After temporary value destroying it will be a reference on invalid object.
+// So let's don't call functor in case of __no_op
```
danhoeflinger, line 119: Is this true?
It looks like `__no_op` forwards the reference type of its argument exactly as the return value.  What am I missing?
```
struct __no_op
{
    template <typename _Tp>
    _Tp&&
    operator()(_Tp&& __a) const
    {
        return ::std::forward<_Tp>(__a);
    }
};
```

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -113,260 +114,154 @@ struct walk_n
...
-    }
-
-    // _IsFull is ignored here. We assume that boundary checking has been already performed for this index.
-    template <typename _IsFull>
-    void
-    __scalar_path_impl(_IsFull, const std::size_t __idx, _Range __rng) const
-    {
-        __f(__rng[__idx]);
-    }
+    oneapi::dpl::__internal::__no_op __f;
```
danhoeflinger, line 123: If there is a reason to have this specialization, and we don't plan to use `__f`, can't we just omit it?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
danhoeflinger: oh, whoops, I see now that this is *not* new code.... just an artifact of the git diff.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
danhoeflinger: same here, not new code...

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
mmichel11: The diff is unfortunately pretty confusing in some places. We can open an issue if something in the `walk_n` implementation seems suspicious.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
mmichel11: It is not faster, I had just adopted the suggestion in https://github.com/uxlfoundation/oneDPL/pull/2165#discussion_r2065811416.

`if...else` is probably more conventional than a ternary with functions that have no return values, so I can switch back.

SergeyKopienko:
@mmichel11 may be to late, but what I thinks...
1) Your current approach mandatory requires from all bricks passed into `parallel_for` to have some members like `__can_process_multiple_iters`. Is this a legitimate demand? May we have have some bricks without this two params? I don't know, may we have bricks from user-level code?
2) now we have 9 lines with definitions of `constexpr static bool __can_vectorize = true/false;` and `constexpr static bool __can_process_multiple_iters = true/false; which shoul br public and etc.
What about to have common base structure with two bool template params and inherit it in each required brick?

danhoeflinger:
>May we have have some bricks without this two params? I don't know, may we have bricks from user-level code?

Looking through the code, this PR covers all hetero bricks used in parallel for, and we don't just pass in user defined bricks.

> 2. now we have 9 lines with definitions of `constexpr static bool __can_vectorize = true/false;` and `constexpr static bool __can_process_multiple_iters = true/false; which shoul br public and etc.
> What about to have common base structure with two bool template params and inherit it in each required brick?

Its a good point that this creates this "hidden" requirement on bricks for the hetero parallel for.  (Though I think it would result in compile time errors if the booleans were missing...)  We could require some base struct, and have the function signature for parallel for accept a type of the base struct.  Another option could be some `brick_traits` struct which provides defaults if nothing is specified within the bricks.  I'm not sure what is best.

mmichel11:
@SergeyKopienko @danhoeflinger I think there was some initial opposition to using inheritance just to obtain compile-time brick members from previous discussion with @MikeDvorskiy. With the current approach, we do have a hidden requirement for these the bricks to define these members with a compile error when that's not the case, although the compile-time error may not be the most clear.

In terms of having the least amount of code, I think the `brick_traits` struct Dan suggests would be the best with defaults to true for both fields. My concern is that it may be more error prone for someone to define a new brick without examining the defaults and not providing the proper vectorization path which would lead to correctness issues. Since we vectorize for certain types, this could be missed in testing.  Requiring the implementor to define the fields manually would likely reduce the chance of this happening.

As an alternative to the inheritance, we could add some static assert to `__parallel_for` which checks that the members are defined:
```
static_assert(has_pfor_brick_members_v<_Brick>);
```
I have an implementation that is able to ensure `__can_vectorize` and `__can_process_multiple_iters` are defined as members and are convertible to bool.

SergeyKopienko:
> In terms of having the least amount of code, I think the `brick_traits` struct Dan suggests would be the best with defaults to true for both fields. My concern is that it may be more error prone for someone to define a new brick without examining the defaults and not providing the proper vectorization path which would lead to correctness issues. Since we vectorize for certain types, this could be missed in testing. Requiring the implementor to define the fields manually would likely reduce the chance of this happening.

In this case we may just check in `static_assert` that `brick` has been inherited from some required struct.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
```diff
@@ -200,13 +200,36 @@ struct __parallel_for_large_submitter<__internal::__optional_kernel_name<_Name..
     }
 };
 
+template <typename _Brick>
+struct __has_pfor_brick_members
+{
+  private:
+    template <typename _F>
+    static auto
+    test(int) -> decltype(std::bool_constant<_F::__can_vectorize>{},
```
SergeyKopienko, line 209: As far as I understand here you just check that `bool __can_vectorize` and `bool __can_process_multiple_iters` exists but doesn't check that they are declared as `constexpr`.

danhoeflinger (dismissed):
LGTM assuming green CI

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
mmichel11: It does check that the members can be evaluated at compile time as the substitution will fail if the members cannot be passed to the `bool_constant` template.

If the compiler can figure out the members at compile time, then it will still work if the members are `const static bool`. In my opinion, we should recommend the members be `constexpr` in the error static_assert, so it's guaranteed to be known at compile time. Is there a way to determine if a variable is declared `constexpr`?

mmichel11:
There may be some small conflicts with the ranges PR: https://github.com/uxlfoundation/oneDPL/pull/2100. I suggest we merge that one first, and then I will resolve them in a commit here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_for.h
rarutyun: I have no idea about such a way because both `constexpr` variable and `const` variable that is initialized by a constant expression can be used in other constant expressions. Thus, to me it looks undistinguishable.

timmiesmith (approved):
Reapproving now that CI is green with conflicts resolved.

mmichel11:
No issues spotted in internal CI but I had to resolve some minor conflicts with https://github.com/uxlfoundation/oneDPL/pull/2225 (adding `__checked_deferrable_wait` in several places).

danhoeflinger (approved):
LGTM after conflict resolution

PR #2167, opened 2025-04-07
Define the amount of results in `__result_and_scratch_storage` in compile-time

SergeyKopienko:
In this PR we define amount of results in `__result_and_scratch_storage` in compile-time.

Now we control in compile-time that the `value` exist inside `__result_and_scratch_storage_impl` object: in the  `__result_and_scratch_storage_impl::__get_value()` and `__result_and_scratch_storage_impl::__get_data(...)` functions. 

This approach also gives us ability later implement returns of all existed results from `__result_and_scratch_storage_impl` as `std::tuple`.


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -656,34 +654,39 @@ struct __result_and_scratch_storage_impl : __result_and_scratch_storage_base
...
+    auto
     __wait_and_get_value(sycl::event __event) const
     {
+        static_assert(_NResults == 1);
+
         if (is_USM())
             __event.wait_and_throw();
 
-        return __get_value();
+        if constexpr (_NResults == 1)
```
SergeyKopienko, line 665: I believe we fix here some important error that has place for every instance of `__result_and_scratch_storage` without results: we tried to load some value from memory (and so on) ever if it's absent.

@danhoeflinger, @MikeDvorskiy what do you thinks?

danhoeflinger: Are there any places where we attempt to `__wait_and_get_value` when there is no result part of `__result_and_scratch_storage`?  It seems that in and of itself is flawed.

I agree that this is safer, but are there actually spots in the code where we are attempting to get a value that doesn't exist?

SergeyKopienko: Looking to `static_assert(_NResults == 1);` at line 660 I think it's extra condition.
Fixed (removed).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -570,7 +571,8 @@ __parallel_transform_scan_single_group(oneapi::dpl::__internal::__device_backend
 
     // Although we do not actually need result storage in this case, we need to construct
     // a placeholder here to match the return type of the non-single-work-group implementation
-    __result_and_scratch_storage<_ExecutionPolicy, _ValueType> __dummy_result_and_scratch{__exec, 0, 0};
+    __result_and_scratch_storage<_ExecutionPolicy, _ValueType, /* _NResults = */ 1> __dummy_result_and_scratch{__exec,
```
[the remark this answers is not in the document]
mmichel11: In `__parallel_copy_if_static_single_group_submitter::operator()`, the returned result is used. However, in the transform scan case it is not.

Reduce-then-scan always uses result-and-scratch with a single element return. In the case of a transform scan, the result is written to by the kernel but never actually used when it's returned. I think we have three options:

1. Leave it as is. The performance impact is likely minimal which we can verify.
2. Correct reduce-then-scan to have no extra result type for transform-scan based algorithms and use `__result_and_scratch_storage<_ExecutionPolicy, _ValueType, /* _NResults = */ 0>` for this case instead. This would align return types for the issue you encountered and would avoid unnecessary allocation where it is not needed. It probably requires the most changes and would further complicate reduce-then-scan.
3. Do what you suggest and return some pointer to `__result_and_scratch_storage_base`. We definitely need to keep the object alive until the event is finished.

I expect the performance impacts of 1 & 3 are minimal either way. Since 1 is what is currently implemented, I recommend to check the performance of this branch versus main for these small single work-group cases. Assuming there is no performance impact, I think a comment here explaining this would be nice.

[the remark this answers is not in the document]
SergeyKopienko: Hope I found one better variant - to return from `__parallel_transform_scan_single_group` simple `sycl::event` and then create `__future` instance inside `__parallel_transform_scan` implementation.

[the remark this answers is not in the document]
mmichel11: This approach LGTM if you think it makes the overall implementation more understandable.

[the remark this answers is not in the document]
SergeyKopienko: Yes, I think so...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1077,9 +1065,16 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag __backen
             std::size_t __single_group_upper_limit = __use_reduce_then_scan ? 2048 : 16384;
             if (__group_scan_fits_in_slm<_Type>(__exec.queue(), __n, __n_uniform, __single_group_upper_limit))
             {
-                return __parallel_transform_scan_single_group(
+                auto __event = __parallel_transform_scan_single_group(
                     __backend_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_Range1>(__in_rng),
                     std::forward<_Range2>(__out_rng), __n, __unary_op, __init, __binary_op, _Inclusive{});
+
+                // Although we do not actually need result storage in this case, we need to construct
```
MikeDvorskiy, line 1072: I am afraid we have got performance degradation here, because this code adds copying to the host `__result_and_scratch_storage` instance, at least 32K data copying, unfortunately.
Moreover, this perf penalty might be rather big relatively a single group execution time.

SergeyKopienko: @mmichel11 have checked this performance, performance penalty isn't so significant here

SergeyKopienko: please see https://github.com/uxlfoundation/oneDPL/pull/2167#discussion_r2032098626 for details

SergeyKopienko: @MikeDvorskiy I think I should explain additionally this place. Еhe thing is that
```C++
                // Although we do not actually need result storage in this case, we need to construct
                // a placeholder here to match the return type of the non-single-work-group implementation
                __result_and_scratch_storage<_ExecutionPolicy, _Type, /* _NResults */ 1> __dummy_result_and_scratch{
                    __exec, 0};
```
now created just for alignment with return type of this function.
Nowhere in the code is a value set to or retrieved from this container.
From API side of this function `__parallel_transform_scan`, this `__result_and_scratch_storage` instance returns in the `__future` just for extend life-time of some other data. 

As already discussed in the https://github.com/uxlfoundation/oneDPL/pull/2167#discussion_r2032098626, we can return `std::shared_ptr<__result_and_scratch_storage_base>` instead of them in this place.

MikeDvorskiy: Regarding performance penalty. 
A meant that it might lead to get_data call, which calls `queue::memcpy` or implicit copying via sycl::buffer: 

```
__get_value() const
    {
        static_assert(0 <= _Idx && _Idx < _NResults);

        if (__use_USM_host && __supports_USM_device)
        {
            return *(__result_buf.get() + _Idx);
        }
        {
            _T __tmp;
            __q.memcpy(&__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
            return __tmp;
        }
        else
        {
            return __sycl_buf->get_host_access(sycl::read_only)[__scratch_n + _Idx];
        }
    }
```

But, @SergeyKopienko persuaded me that `__get_value()` is not called in a caller for this scenario.
In that case should not be significant perf penalty.

SergeyKopienko: @MikeDvorskiy, @danhoeflinger, @mmichel11 Alternative approach is implemented here: https://github.com/uxlfoundation/oneDPL/pull/2182

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -514,7 +514,7 @@ struct __result_and_scratch_storage_base
     __get_data(sycl::event, std::size_t* __p_buf) const = 0;
 };
 
-template <typename _T>
+template <typename _T, std::size_t _NResults>
```
danhoeflinger, line 517: Can we default this to `1` and get rid of a significant number of these changes?
(perhaps it is not worth losing the clarity when looking at the type, but I think it might be.)

SergeyKopienko: Of course, it's implementable.
But as far as I remembering initially we likes mandatory specify the amount of results and the amount of scratch-data.
But I will do.

danhoeflinger: I'm OK with either option really.  Do what you think is best here.

SergeyKopienko: Done.

MikeDvorskiy (approved):
I have not detected any issues with the changes while I a reviewing it.
It is more or less OK.
But, in general, one remarks  - as I told in the past I don't like the code coupling  - future/__result_and_scratch_storage/places with_NResults, where we rely on 0 or 1.
In the future these classes should be totally re-designed, I guess.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I think what you have here is fine, and an improvement from what it was, because it limits the exposure of the dummy temp return to where we branch.  It does not change the story otherwise from the previous code as far as I understand.

danhoeflinger (approved):
LGTM

mmichel11 (approved):
LGTM

PR #2171, opened 2025-04-08
Enable vectorization of reverse iterators in SYCL backend parallel-for

mmichel11:
The vectorization utilities in https://github.com/uxlfoundation/oneDPL/pull/1976 enforced a requirement that the underlying range / iterator type must be contiguous in order to enable vectorization paths along with other requirements we enforce. While this was not needed for correctness reasons, it was intended to enable good performance for the `reverse_iterator` case. Hardware vector instructions support vector load / stores only on increasing addresses, so in the case where addresses decrease with iterator increments, this is not directly vectorizable.

After further experimentation and study of generated assembly, I see that the oneAPI compiler is able to optimize the loads / stores in the case of `oneapi::dpl::reverse_iterator` to enable the generation of vector instructions / wider loads & stores in parallel-for based algorithms. All experiments I performed on the types we vectorize (single byte and two byte) resulted in better performance when enabling vectorization paths.

Due to this, we can remove the `__is_vectorizable_range` utility in its entirety and always take the vectorization path in fundamental 1 and 2 byte types. This will allow us to vectorize in the contiguous and reverse contiguous case. For types where the access pattern is already poor (e.g. permutation iterator with scattered accesses), the change in access pattern is not a concern here.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
@@ -128,7 +128,6 @@ struct walk_vector_or_scalar_base
 
   public:
     constexpr static bool __can_vectorize =
-        (oneapi::dpl::__ranges::__is_vectorizable_range<std::decay_t<_Ranges>>::value && ...) &&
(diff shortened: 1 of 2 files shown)


MikeDvorskiy (approved):

danhoeflinger (approved):
LGTM

PR #2175, opened 2025-04-10
Remove erroneously const scratch space

danhoeflinger:
We pass scratch space as a const lvalue ref, but it's contents are actually modified.
We only call `const` functions `__get_scratch_acc`, but it is incorrect to label this `const`.

Introduced in #1773.


SergeyKopienko (dismissed):
LGTM

SergeyKopienko (approved):
LGTM

PR #2176, opened 2025-04-10
Fix example CMakes, add build+run into CI

danhoeflinger:
Fix CMake files for examples, and run through build and run of each for dpcpp per-commit CI.


timmiesmith (approved):
LGTM. Thank you.

PR #2178, opened 2025-04-11
Fix philox_test.pass

dmitriy-sobolev:
Let's fix:
> philox_test.pass.cpp:23:24: error: no template named 'philox_engine' in namespace 'oneapi::dpl::experimental'

error, when the unnamed lambda mode if off.


ElenaTyuleneva (approved):
LGTM, thanks for providing the fix!

PR #2183, opened 2025-04-14
Replacing CPATH usage with CPLUS_INCLUDE_PATH

timmiesmith:
oneDPL and other oneAPI components had been using CPATH incorrectly, resulting in the header files not being treated as system header files, which is the expectation for oneAPI installations.  The impact is that on Windows warnings were not being suppressed, which is the default for system headers.

This PR corrects the issue by replacing usage of 'CPATH' with 'CPLUS_INCLUDE_PATH' as recommended by the oneAPI program.  The copyright statement is also updated to remove the year to bring it inline with the copyright statement in other oneDPL files.


danhoeflinger:
I'm not sure about what this is used for but there is also `env/onedplvars.sh`.  Do we need to update that?

If not, can we remove it and its parent folder entirely?

timmiesmith:
> I'm not sure about what this is used for but there is also `env/onedplvars.sh`. Do we need to update that?
> 
> If not, can we remove it and its parent folder entirely?

Good catch,  The file is unused and looks like the initial script we used for configuration in the oneAPI beta period. I've removed the env/ directory.

danhoeflinger (approved):
LGTM

Issue #2184, opened 2025-04-15
Invalid predicate format in `test_adjacent_difference`


SergeyKopienko:
https://github.com/uxlfoundation/oneDPL/blob/6b0b97e61ca60e78f063397634c036eb98ab087a/test/general/sycl_iterator/sycl_iterator_for.pass.cpp#L985

Invalid predicate format has place:
```C++
auto __f = [](Iterator1ValueType& a, Iterator1ValueType& b) -> Iterator2ValueType { return a + b; };
```

It doesn't complaint with requirement for https://en.cppreference.com/w/cpp/algorithm/adjacent_difference - `op` parameter:

op | - | binary operation function object that will be applied.The signature of the function should be equivalent to the following: Ret fun(const Type1 &a, const Type2 &b);The signature does not need to have const &.The types  Type1 and  Type2 must be such that an object of type iterator_traits<InputIt>::value_type can be implicitly converted to both of them. The type Ret must be such that an object of type OutputIt can be dereferenced and assigned a value of type Ret.​

SergeyKopienko:
PR created: https://github.com/uxlfoundation/oneDPL/pull/2185

PR #2185, opened 2025-04-15
Fix predicate in `test_adjacent_difference`

SergeyKopienko:
In this PR we fix issue https://github.com/uxlfoundation/oneDPL/issues/2184

### DIFF
--- a/test/general/sycl_iterator/sycl_iterator_for.pass.cpp
+++ b/test/general/sycl_iterator/sycl_iterator_for.pass.cpp
@@ -982,7 +982,7 @@ DEFINE_TEST(test_adjacent_difference)
         Iterator1ValueType fill_value{1};
         Iterator2ValueType blank_value{0};
 
-        auto __f = [](Iterator1ValueType& a, Iterator1ValueType& b) -> Iterator2ValueType { return a + b; };
+        auto __f = [](const Iterator1ValueType& a, const Iterator1ValueType& b) -> Iterator2ValueType { return a + b; };
 
         // init
         ::std::for_each(host_keys.get(), host_keys.get() + n,


test/general/sycl_iterator/sycl_iterator_for.pass.cpp
```diff
@@ -982,7 +982,7 @@ DEFINE_TEST(test_adjacent_difference)
         Iterator1ValueType fill_value{1};
         Iterator2ValueType blank_value{0};
 
-        auto __f = [](Iterator1ValueType& a, Iterator1ValueType& b) -> Iterator2ValueType { return a + b; };
+        auto __f = [](const Iterator1ValueType& a, const Iterator1ValueType& b) -> Iterator2ValueType { return a + b; };
```
MikeDvorskiy, line 985: or we can write shorter
`auto __f = [](auto& a, auto& b)`

dmitriy-sobolev: I suppose the idea is to have a test which imposes the requirements for an algorithm as described in #2184. We need that "const" to test a more restrictive case.

dmitriy-sobolev (approved):
LGTM

test/general/sycl_iterator/sycl_iterator_for.pass.cpp
dmitriy-sobolev: I guess Mikhail rather suggested writing `auto` instead of `Iterator1ValueType` - I am indifferent here.

MikeDvorskiy: I meant "auto" instead of "const Iterator1ValueType"...
Because "auto& a" wil be deduced into "const int& a" in case of passing "const int", for example.

But I don't insist on.

MikeDvorskiy (approved):
LGTM

PR #709, opened 2022-12-02
[oneDPL] radix sort change: + sorting in one group for small data size

MikeDvorskiy:
[oneDPL] radix sort change: + group_radix_sort; it is faster for small data sizes.

The all sort iterations are processed in one kernel, one submit.
A barrier within a work group is a kind of "global" barrier.
Each work item processes one data block in serial. A  block size is 1 to 64, it is various depending on data size. 

The small size here are:
-  n < 128: block size 1
-  n < 256: block size 2
-  n < 512: block size 4
-  n < 2048: block size 8
-  n < 8192: block size 16
-  n < 16384: block size 32
-  n < 32768: block size 64

PR #2194, opened 2025-04-17
Fix an error in  __one_group_submitter

SergeyKopienko:
.... for me the approach from https://github.com/uxlfoundation/oneDPL/pull/709 looks like a little bit strange:
```C++
//The file is an internal file and the code of that file is included by a major file into the following namespaces:
//namespace oneapi
//{
//namespace dpl
//{
//namespace __par_backend_hetero
//{
```

So in someone case I received warning and in this PR I am trying to fix it.

In ideal case probably the approach from https://github.com/uxlfoundation/oneDPL/pull/709 should be reimplemented in we should place this code into normal `oneapi::dpl::__par_backend_hetero` namespace.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
@@ -152,7 +152,7 @@ struct __subgroup_radix_sort
     struct __one_group_submitter;
 
     template <typename... _Name>
-    struct __one_group_submitter<__internal::__optional_kernel_name<_Name...>>
+    struct __one_group_submitter<oneapi::dpl::__par_backend_hetero::__internal::__optional_kernel_name<_Name...>>
     {
         template <typename _RangeIn, typename _Proj, typename _SLM_tag_val, typename _SLM_counter>
         sycl::event


danhoeflinger (approved):
LGTM.

mmichel11 (approved):

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_radix_sort_one_wg.h
```diff
@@ -152,7 +152,7 @@ struct __subgroup_radix_sort
     struct __one_group_submitter;
 
     template <typename... _Name>
-    struct __one_group_submitter<__internal::__optional_kernel_name<_Name...>>
+    struct __one_group_submitter<oneapi::dpl::__par_backend_hetero::__internal::__optional_kernel_name<_Name...>>
```
SergeyKopienko, line 155: This is only one place inside all this file where according to comments from this file
```C++
//The file is an internal file and the code of that file is included by a major file into the following namespaces:
//namespace oneapi
//{
//namespace dpl
//{
//namespace __par_backend_hetero
//{
```
we simple specify `__internal` for namespace `oneapi::dpl::__par_backend_hetero::__internal`

In some another lines from this file we using full namespace.
Example:
```C++
        using _SortKernelLoc = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
            __radix_sort_one_wg_kernel<_KernelNameBase, __wg_size_t, __block_size_t, __call_0_t>>;
```

SergeyKopienko:
I don't remember exactly in which case I receive error for this string, but according to my comment in https://github.com/uxlfoundation/oneDPL/pull/2194/files#r2050337702 I believe that better to apply this PR in any case.

@MikeDvorskiy how do you think?

MikeDvorskiy:
> @MikeDvorskiy how do you think?

Actually, the comment in the code "//The file is an internal file and the code of that file is included by a major file into the following namespaces:"
explain why we don't need full qualified namespace, only `__internal`. So, it should not lead to an error...
But, you you really faced to an issue in some cases, I don't have objections to use  full qualified namespace here.

PR #2196, opened 2025-04-17
Resolve C++20 extension warning in histogram_impl_hetero.h

mmichel11:
The addition of an explicitly provided template parameter in https://github.com/uxlfoundation/oneDPL/pull/2161 when calling `__parallel_histogram` causes a warning in C++17 as this is a C++20 feature when there is no prior declaration. We are trying to call `__parallel_histogram` which is in `oneapi::dpl::__par_backend_hetero` from `oneapi::dpl::__internal`, so adding the full namespace to the call resolves the warning.

Here is the original warning:
```
 warning: use of function template name with no prior declaration in function call with explicit template arguments is a C++20 extension [-Wc++20-extensions]
  164 |             __parallel_histogram<_CustomName>(_BackendTag{}, __q_local, __init_event, __input_buf.all_view(),
```

I also cherry picked a small commit @SergeyKopienko recommended from his branch.


mmichel11:
Closing since @danhoeflinger opened a similar PR 2 minutes before me.

PR #2197, opened 2025-04-17
Revert: Removal ExecutionPolicy from submitters and etc. stuff #2161

SergeyKopienko:
In this PR we revert https://github.com/uxlfoundation/oneDPL/pull/2161

The reason: https://github.com/uxlfoundation/oneDPL/pull/2161#issuecomment-2813669387

Also the change in `kt` namespace (from `sycl::queue` to `sycl::queue&` in params) breaks our experimental API and looks as incorrect.


adamfidel (dismissed):
LGTM.

I verified that this patch is essentially `git revert 14c97e05fd8cf3a646d69584614ddff17340f7ad` with some minor fix ups to reconcile conflicts with `main`.

danhoeflinger (approved):
LGTM, I made a similar look at the diff and main~2 locally and can confirm that this should fully revert the issue, preserving the fixes from the subsequent PR.

danhoeflinger (approved):

mmichel11 (approved):
LGTM. I compared the diffs on the commit before the original into main and the latest on this branch. Everything looks as expected from the revert.

adamfidel (approved):

danhoeflinger:
The clang format difference looks like mostly intentional decisions from the suggestions, and as we mentioned, we have confirmed that this cleanly reverts the commit, so we are not introducing new clang format issues with this reverting commit.

PR #2192, opened 2025-04-17
Remove `ExecutionPolicy` from `parallel_for`- and `parallel_for`-based submitters

SergeyKopienko:
In this PR we remove `ExecutionPolicy` from `parallel_for`-submitter and from `parallel_for`-based submitters.

#### Examples of submitters with `oneapi::dpl::__internal::__device_backend_tag` and `ExecutionPolicy` in parameters:
```C++
template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3, typename _Range4,
          typename _BinaryPredicate, typename _BinaryOperator>
oneapi::dpl::__internal::__difference_t<_Range3>
__parallel_reduce_by_segment_fallback(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec,
                                      _Range1&& __keys, _Range2&& __values, _Range3&& __out_keys,
                                      _Range4&& __out_values, _BinaryPredicate __binary_pred,
                                      _BinaryOperator __binary_op,
                                      /*known_identity=*/std::false_type);

template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3, typename _Range4,
          typename _BinaryPredicate, typename _BinaryOperator>
oneapi::dpl::__internal::__difference_t<_Range3>
__parallel_reduce_by_segment_fallback(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec,
                                      _Range1&& __keys, _Range2&& __values, _Range3&& __out_keys,
                                      _Range4&& __out_values, _BinaryPredicate __binary_pred,
                                      _BinaryOperator __binary_op,
                                      /*known_identity=*/std::true_type);

template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3, typename _Range4,
          typename _BinaryPredicate, typename _BinaryOperator>
oneapi::dpl::__internal::__difference_t<_Range3>
__parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Range1&& __keys,
                             _Range2&& __values, _Range3&& __out_keys, _Range4&& __out_values,
                             _BinaryPredicate __binary_pred, _BinaryOperator __binary_op);

template <typename _ExecutionPolicy, typename _Fp, typename _Index, typename... _Ranges>
__future<sycl::event>
[...]

PR #2153, opened 2025-03-31
Replacement of internal `lambda` in `oneDPL` code by functional objects

SergeyKopienko:
In this PR we reimplements `lambda` code as functional objects according to our offline discussion.

Other linked PR's: https://github.com/uxlfoundation/oneDPL/pull/2148, https://github.com/uxlfoundation/oneDPL/pull/2152

PR #2198, opened 2025-04-17
Removal `ExecutionPolicy` from submitters and etc. stuff (attemp #2)

SergeyKopienko:
In this PR we fix errors introduced in the https://github.com/uxlfoundation/oneDPL/pull/2161

### Attention!
Now this PR still include all changes from 
- https://github.com/uxlfoundation/oneDPL/pull/2153

UPD: https://github.com/uxlfoundation/oneDPL/pull/2153 has been merged into main branch and this PR now contains actual state of main branch.

Before future work with this PR required to merge into main branch the next PRs (otherwise we will have a lot of differences in this PR):
- https://github.com/uxlfoundation/oneDPL/pull/2152
- https://github.com/uxlfoundation/oneDPL/pull/2153
- https://github.com/uxlfoundation/oneDPL/pull/2148

The goal of this PR - pass into submitters `sycl::queue& __q` instead of `_ExecutionPolicy&& __exec` - so we remove one template parameter and decrease possible instantiations of template submitter classes / functions.

This is the main and final step to use `ExecutionPolicy` in `oneDPL` algorithm calls independently from their category values (when unnamed lambda is switched off or `Kernel` name is specified in the users code directly).

Also this RP contains now some changes from https://github.com/uxlfoundation/oneDPL/pull/2201 till it will be merged into main branch.

### Note about `parallel_for`- and `parallel_for`-based submitters
Now we still need to have `ExecutionPolicy` in `parallel_for`- and in `parallel_for`-based submitters.
I going to fix that in the separate PR https://github.com/uxlfoundation/oneDPL/pull/2192

### Note about clang-format check for this PR
- I propose do not apply the rest of clang-format changes into this PR: please take a loot to them at https://github.com/uxlfoundation/oneDPL/actions/runs/14240611126/job/39909462269?pr=2161

Example (1) of incorrect clang-format:
```C++
         // 2. Apply work group carry outs, calculate output indices, and load results into correct indices.
-        __q.submit([&](sycl::handler& __cgh) {
+        __q
+            .submit([&](sycl::handler& __cgh) {
                 oneapi::dpl::__ranges::__require_access(__cgh, __keys, __out_values);
```

Example (2) of incorrect clang-format:
```C++
 template <typename _KernelName, ::std::uint32_t __radix_bits, bool __is_ascending, typename _ValRange,
           typename _CountBuf, typename _Proj
 #if _ONEDPL_COMPILE_KERNEL
-          , typename _Kernel
+          ,
+          typename _Kernel
 #endif
           >
```

Example (3) of incorrect clang-format:
```C++
 sycl::event
__radix_sort_count_submit(sycl::queue& __q, ::std::size_t __segments, ::std::siz
                           ::std::uint32_t __radix_offset, _ValRange&& __val_rng, _CountBuf& __count_buf,
                           sycl::event __dependency_event, _Proj __proj
 #if _ONEDPL_COMPILE_KERNEL
-                          , _Kernel& __kernel
+                          ,
+                          _Kernel& __kernel
 #endif
 )
```

### DIFF
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -374,9 +374,9 @@ __parallel_scan_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Execu
     sycl::queue __q_local = __exec.queue();
 
     __sycl_scan_by_segment_impl<_CustomName, __is_inclusive>()(
-        oneapi::dpl::__internal::__device_backend_tag{},
-        __q_local, std::forward<_Range1>(__keys), std::forward<_Range2>(__values), std::forward<_Range3>(__out_values),
-        __binary_pred, __binary_op, __init, __identity);
+        oneapi::dpl::__internal::__device_backend_tag{}, __q_local, std::forward<_Range1>(__keys),
+        std::forward<_Range2>(__values), std::forward<_Range3>(__out_values), __binary_pred, __binary_op, __init,
+        __identity);
 }
 } //namespace __par_backend_hetero
 
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -143,7 +143,7 @@ struct __sycl_scan_by_segment_impl
                       oneapi::dpl::__internal::__kernel_work_group_size(__q, __seg_scan_prefix_kernel)});
 #endif
 
-        ::std::size_t __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __wgroup_size * __vals_per_item);
+        std::size_t __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __wgroup_size * __vals_per_item);
 
         auto __partials = oneapi::dpl::__par_backend_hetero::__buffer<__val_type>(__n_groups).get_buffer();
 
@@ -374,13 +374,13 @@ __parallel_scan_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Execu
     sycl::queue __q_local = __exec.queue();
 
     __sycl_scan_by_segment_impl<_CustomName, __is_inclusive>()(
-        oneapi::dpl::__internal::__device_backend_tag{}, __q_local, ::std::forward<_Range1>(__keys),
-        ::std::forward<_Range2>(__values), ::std::forward<_Range3>(__out_values), __binary_pred, __binary_op, __init,
-        __identity);
+        oneapi::dpl::__internal::__device_backend_tag{},
+        __q_local, std::forward<_Range1>(__keys), std::forward<_Range2>(__values), std::forward<_Range3>(__out_values),
+        __binary_pred, __binary_op, __init, __identity);
 }
-
 } //namespace __par_backend_hetero
-namespace internal
+
+namespace __internal
 {
 template <typename _BackendTag, typename Policy, typename InputIterator1, typename InputIterator2,
           typename OutputIterator, typename T, typename BinaryPredicate, typename Operator, typename Inclusive>
@@ -409,12 +409,12 @@ __pattern_scan_by_segment(__internal::__hetero_tag<_BackendTag>, Policy&& policy
     constexpr iter_value_t identity = unseq_backend::__known_identity<Operator, iter_value_t>;
 
     __bknd::__parallel_scan_by_segment<Inclusive::value>(
-        _BackendTag{}, ::std::forward<Policy>(policy), key_buf.all_view(), value_buf.all_view(),
+        _BackendTag{}, std::forward<Policy>(policy), key_buf.all_view(), value_buf.all_view(),
         value_output_buf.all_view(), binary_pred, binary_op, init, identity);
     return result + n;
 }
 
-} // namespace internal
+} // namespace __internal
 } // namespace dpl
 } // namespace oneapi
 #endif
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -108,8 +108,9 @@ struct __sycl_scan_by_segment_impl
     template <typename _BackendTag, typename _Range1, typename _Range2, typename _Range3, typename _BinaryPredicate,
               typename _BinaryOperator, typename _T>
     void
-    operator()(_BackendTag, sycl::queue& __q, _Range1&& __keys, _Range2&& __values, _Range3&& __out_values,
-               _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init, _T __identity)
+    operator()(oneapi::dpl::__internal::__device_backend_tag, sycl::queue& __q, _Range1&& __keys, _Range2&& __values,
+               _Range3&& __out_values, _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init,
+               _T __identity)
     {
         using _SegScanWgKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
             _SegScanWgPhase, _CustomName, _Range1, _Range2, _Range3, _BinaryPredicate, _BinaryOperator>;
@@ -361,10 +362,10 @@ struct __sycl_scan_by_segment_impl
     }
 };
 
-template <typename _BackendTag, typename Policy, typename InputIterator1, typename InputIterator2,
-          typename OutputIterator, typename T, typename BinaryPredicate, typename Operator, typename Inclusive>
-OutputIterator
-__scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIterator1 first1,
+template <typename Policy, typename InputIterator1, typename InputIterator2, typename OutputIterator, typename T,
+          typename BinaryPredicate, typename Operator, typename Inclusive>
+oneapi::dpl::__internal::__enable_if_hetero_execution_policy<Policy, OutputIterator>
+__scan_by_segment_impl_common(oneapi::dpl::__internal::__device_backend_tag, Policy&& policy, InputIterator1 first1,
                               InputIterator1 last1, InputIterator2 first2, OutputIterator result, T init,
                               BinaryPredicate binary_pred, Operator binary_op, Inclusive)
 {
@@ -391,9 +392,9 @@ __scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, Policy&& po
 
     sycl::queue __q_local = policy.queue();
 
-    __sycl_scan_by_segment_impl<_CustomName, Inclusive::value>{}(_BackendTag{}, __q_local, key_buf.all_view(),
-                                                                 value_buf.all_view(), value_output_buf.all_view(),
-                                                                 binary_pred, binary_op, init, identity);
+    __sycl_scan_by_segment_impl<_CustomName, Inclusive::value>{}(
+        oneapi::dpl::__internal::__device_backend_tag{}, __q_local, key_buf.all_view(), value_buf.all_view(),
+        value_output_buf.all_view(), binary_pred, binary_op, init, identity);
 
     return result + n;
 }
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -362,6 +362,7 @@ struct __sycl_scan_by_segment_impl
     }
 };
 
+// TODO required to move into oneapi::dpl::__par_backend_hetero namesace
 template <typename Policy, typename InputIterator1, typename InputIterator2, typename OutputIterator, typename T,
           typename BinaryPredicate, typename Operator, typename Inclusive>
 oneapi::dpl::__internal::__enable_if_hetero_execution_policy<Policy, OutputIterator>
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -96,7 +96,7 @@ class __seg_scan_wg_kernel;
 template <bool __is_inclusive, typename... Name>
 class __seg_scan_prefix_kernel;
 
-template <bool __is_inclusive>
+template <typename _CustomName, bool __is_inclusive>
 struct __sycl_scan_by_segment_impl
 {
     template <typename... _Name>
@@ -105,20 +105,16 @@ struct __sycl_scan_by_segment_impl
     template <typename... _Name>
     using _SegScanPrefixPhase = __seg_scan_prefix_kernel<__is_inclusive, _Name...>;
 
-    template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3,
-              typename _BinaryPredicate, typename _BinaryOperator, typename _T>
+    template <typename _BackendTag, typename _Range1, typename _Range2, typename _Range3, typename _BinaryPredicate,
+              typename _BinaryOperator, typename _T>
     void
-    operator()(_BackendTag, _ExecutionPolicy&& __exec, _Range1&& __keys, _Range2&& __values, _Range3&& __out_values,
+    operator()(_BackendTag, sycl::queue& __q, _Range1&& __keys, _Range2&& __values, _Range3&& __out_values,
                _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init, _T __identity)
     {
-        using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<_ExecutionPolicy>;
-
         using _SegScanWgKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-            _SegScanWgPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _BinaryPredicate,
-            _BinaryOperator>;
+            _SegScanWgPhase, _CustomName, _Range1, _Range2, _Range3, _BinaryPredicate, _BinaryOperator>;
         using _SegScanPrefixKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-            _SegScanPrefixPhase, _CustomName, _ExecutionPolicy, _Range1, _Range2, _Range3, _BinaryPredicate,
-            _BinaryOperator>;
+            _SegScanPrefixPhase, _CustomName, _Range1, _Range2, _Range3, _BinaryPredicate, _BinaryOperator>;
 
         using __val_type = oneapi::dpl::__internal::__value_t<_Range2>;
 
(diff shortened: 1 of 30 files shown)


include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -366,10 +363,11 @@ struct __sycl_scan_by_segment_impl
     }
 };
 
-template <typename _BackendTag, typename Policy, typename InputIterator1, typename InputIterator2,
-          typename OutputIterator, typename T, typename BinaryPredicate, typename Operator, typename Inclusive>
-OutputIterator
-__scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, Policy&& policy, InputIterator1 first1,
+// TODO required to move into oneapi::dpl::__par_backend_hetero namesace
```
danhoeflinger, line 366: @mmichel11 Does this actually belong in `__par_backend_hetero`? I'm not so sure.
We may just want / need  to rename `__sycl_scan_by_segment_impl` to be 
`__parallel_scan_by_segment` or something similar.  Move that to `oneapi::dpl::__par_backend_hetero`, and add a fully qualified name here in this function.  
This would also mean that the related tags would stay as is (not include the changes in this PR).

include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -366,10 +363,11 @@ struct __sycl_scan_by_segment_impl
...
+template <typename Policy, typename InputIterator1, typename InputIterator2, typename OutputIterator, typename T,
+          typename BinaryPredicate, typename Operator, typename Inclusive>
+oneapi::dpl::__internal::__enable_if_hetero_execution_policy<Policy, OutputIterator>
+__scan_by_segment_impl_common(oneapi::dpl::__internal::__device_backend_tag, Policy&& policy, InputIterator1 first1,
```
danhoeflinger, line 370: This probably should be named `__pattern_hetero_scan_by_segment` or something like that.  
In any case, this reshuffling / naming changes belong in a different PR.  

The important part is to come to an agreement about what is in the `hetero` space, and what is in the `device_backend` space.  I believe this function stays in the `hetero` space.

include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -392,9 +390,14 @@ __scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, Policy&& po
...
-                                                    value_buf.all_view(), value_output_buf.all_view(), binary_pred,
-                                                    binary_op, init, identity);
+    using _CustomName = oneapi::dpl::__internal::__policy_kernel_name<Policy>;
+
+    sycl::queue __q_local = policy.queue();
+
+    __sycl_scan_by_segment_impl<_CustomName, Inclusive::value>{}(
+        oneapi::dpl::__internal::__device_backend_tag{}, __q_local, key_buf.all_view(), value_buf.all_view(),
+        value_output_buf.all_view(), binary_pred, binary_op, init, identity);
+
```
danhoeflinger, line 400: Assuming we all agree this stays in the `hetero` space, this change should be reverted.

include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -96,7 +96,8 @@ class __seg_scan_wg_kernel;
 template <bool __is_inclusive, typename... Name>
 class __seg_scan_prefix_kernel;
 
-template <bool __is_inclusive>
+// TODO required to move into oneapi::dpl::__par_backend_hetero namesace
```
danhoeflinger, line 99: I think this should probably have a function that is another layer above this struct,  `oneapi::dpl::__par_backend_hetero::__parallel_scan_by_segment`, which then pulls the Name and queue out of the `ExecutionPolicy` then uses this struct.  With that in mind, the namespace of the struct itself is less important if it is not the entry point to the sycl device backend.

include/oneapi/dpl/pstl/glue_memory_impl.h
```diff
@@ -224,6 +224,7 @@ struct __destroy_fn
     void
     operator()(_ReferenceType __val) const
     {
+        static_assert(std::is_reference_v<_ReferenceType>);
```
danhoeflinger, line 227: I don't understand why this is in this PR.  It is already large, lets limit the scope to what is important for this PR.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -605,7 +605,7 @@ __pattern_minmax_element(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _
 
     // This operator doesn't track the lowest found index in case of equal min. values and the highest found index in
     // case of equal max. values. Thus, this operator is not commutative.
-    __pattern_minmax_element__reduce_fn<_Compare, _ReduceValueType> __reduce_fn{__comp};
+    __pattern_minmax_element_reduce_fn<_Compare, _ReduceValueType> __reduce_fn{__comp};
```
danhoeflinger, line 608: Seems to be an unrelated change.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -129,7 +129,7 @@ template <typename _ReduceValueType, typename _Compare>
 struct __pattern_min_element_reduce_fn;
 
 template <typename _Compare, typename _ReduceValueType>
-struct __pattern_minmax_element__reduce_fn;
+struct __pattern_minmax_element_reduce_fn;
```
danhoeflinger, line 132: another component of the unrelated change.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -314,7 +314,7 @@ struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::
 };
 
 template <typename _Compare, typename _ReduceValueType>
-struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::__pattern_minmax_element__reduce_fn,
+struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::__pattern_minmax_element_reduce_fn,
```
danhoeflinger, line 317: unrelated change

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2251,6 +2258,27 @@ __parallel_partial_sort(oneapi::dpl::__internal::__device_backend_tag __backend_
...
+                        _Iterator __first, _Iterator __mid, _Iterator __last, _Compare __comp)
+{
+    const auto __mid_idx = __mid - __first;
+
+    auto __keep = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read_write, _Iterator>();
+    auto __buf = __keep(__first, __last);
+
+    return __parallel_partial_sort_impl<_CustomName>(__backend_tag, __q, __buf.all_view(),
+                                                     __partial_merge_kernel<decltype(__mid_idx)>{__mid_idx}, __comp);
+}
```
danhoeflinger, line 2280: It looks like this is meant to replace the previous function, but instead adds an overload.  I think we need to remove the old version.

danhoeflinger (commented):
At the low level, other than my comments, the rest of the changes look pretty good to me.
I want to come back and consider it again at a high level, and also confirm that this gets ALL of the locations where we want to eliminate ExecutionPolicy in favor of Name and queue reference.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: We have only one implementation of `__scan_by_segment_impl_common`, `__sycl_scan_by_segment_impl` - and it's for sycl backend.
Also, the `__sycl_scan_by_segment_impl` is exactly submitter, so it shouldn't work with `_ExecutionPolicy`.
So I believe that this change looks correct but these two things just should be moved into `sycl` backend.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: Also, it was the comment from @adamfidel 
I think this small change make sense.
It's just incorrect structure name from one of my previous PRs.
I propose to save as is.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
SergeyKopienko: The same.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
SergeyKopienko: The same.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: I think that would be great.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: @danhoeflinger I don't understood exactly what you mean here.
But usually the structures like that has placed in the `oneapi::dpl::__par_backend_hetero` namespace.
So in my opinion this structure now is placed in the wrong namespace.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: Probably earlier before one of mine PRs they were duplicated lambdas.
I propose to fix it in separate PR and just create issue now for that.

SergeyKopienko: Or I can move the into `utils_hetero.h` file in this PR...

danhoeflinger: I agree, it should be an issue, and perhaps we can address this naming thing when we remove these redundant callables.  We need to take a look through #2153, because it seems there are more than just the ones mentioned.

include/oneapi/dpl/internal/scan_by_segment_impl.h
danhoeflinger: I described it in more detail in this comment: https://github.com/uxlfoundation/oneDPL/pull/2198#discussion_r2050818264 
In any case, lets make those adjustments in a separate PR, then rebase.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
SergeyKopienko: Reverted.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: Reverted

SergeyKopienko: Issue created: https://github.com/uxlfoundation/oneDPL/issues/2199

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
SergeyKopienko: Reverted

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: PR created: https://github.com/uxlfoundation/oneDPL/pull/2200

include/oneapi/dpl/internal/scan_by_segment_impl.h
danhoeflinger: I've created this PR #2201

include/oneapi/dpl/internal/scan_by_segment_impl.h
danhoeflinger: #2201

include/oneapi/dpl/internal/scan_by_segment_impl.h
danhoeflinger: I've dealt with this in #2201

include/oneapi/dpl/internal/scan_by_segment_impl.h
danhoeflinger: Please see #2201. If / when we merge that we can adjust this code accordingly.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
SergeyKopienko: PR created: https://github.com/uxlfoundation/oneDPL/pull/2200 has been merged into main branch, so I resolve this conversation.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: I approved you PR https://github.com/uxlfoundation/oneDPL/pull/2201 and merged it into this branch right now.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: TODO has been removed

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: TODO has been removed.
The PR https://github.com/uxlfoundation/oneDPL/pull/2201 has been merged into this branch.
So I resolve this conversation.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: So I resolve this conversation.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: > Please see #2201. If / when we merge that we can adjust this code accordingly.

I think at first you will merge https://github.com/uxlfoundation/oneDPL/pull/2201 into main branch. then I will merge main into my PR.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
```diff
@@ -478,59 +474,55 @@ struct __histogram_general_private_global_atomics_submitter<__internal::__option
...
 
     // if bins fit into registers, use register private accumulation
     if (__num_bins <= __max_work_item_private_bins)
     {
-        return __future(
-            __histogram_general_registers_local_reduction<__iters_per_work_item, __max_work_item_private_bins>(
-                __backend_tag, ::std::forward<_ExecutionPolicy>(__exec), __init_event, __work_group_size,
-                ::std::forward<_Range1>(__input), ::std::forward<_Range2>(__bins), __binhash_manager));
+        return __future(__histogram_general_registers_local_reduction<_CustomName, __iters_per_work_item,
+                                                                      __max_work_item_private_bins>(
```
MikeDvorskiy, line 514: it looks as wrong formatting

include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -105,20 +105,17 @@ struct __sycl_scan_by_segment_impl
...
     using _SegScanPrefixPhase = __seg_scan_prefix_kernel<__is_inclusive, _Name...>;
 
-    template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3,
-              typename _BinaryPredicate, typename _BinaryOperator, typename _T>
+    template <typename _BackendTag, typename _Range1, typename _Range2, typename _Range3, typename _BinaryPredicate,
+              typename _BinaryOperator, typename _T>
     void
-    operator()(_BackendTag, _ExecutionPolicy&& __exec, _Range1&& __keys, _Range2&& __values, _Range3&& __out_values,
-               _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _T __init, _T __identity)
+    operator()(oneapi::dpl::__internal::__device_backend_tag, sycl::queue& __q, _Range1&& __keys, _Range2&& __values,
```
MikeDvorskiy, line 111: We are introducing an internal concrete tag name as hardcoded name.. It was not before.
BTW, why whether don't use overload by a hetero tag `__hetero_tag<_BackendTag>`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -232,38 +232,35 @@ struct __parallel_scan_submitter;
 template <typename _CustomName, typename... _PropagateScanName>
 struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name<_PropagateScanName...>>
 {
-    template <typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _InitType, typename _LocalScan,
-              typename _GroupScan, typename _GlobalScan>
+    template <typename _Range1, typename _Range2, typename _InitType, typename _LocalScan, typename _GroupScan,
+              typename _GlobalScan>
     __future<sycl::event, __result_and_scratch_storage<typename _InitType::__value_type>>
-    operator()(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2, _InitType __init, _LocalScan __local_scan,
+    operator()(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _InitType __init, _LocalScan __local_scan,
```
MikeDvorskiy, line 238: Why are we passing queue by left reference? not by `&&`?
`&&` would allow to pass directly r-value queue like `exec.__get_queue()`.
The left reference forces us to create a queue instance on the stack first....

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: You point to someone of submitters.
All submitters which works with `sycl::queue` should be always parametrized by  `oneapi::dpl::__internal::__device_backend_tag` tag.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: I think you are not correct: in this case we will have an error like `error : no matching function for call to...` and the message about candidate like `note: candidate function template not viable: expects an rvalue for 2nd argument`

include/oneapi/dpl/internal/scan_by_segment_impl.h
MikeDvorskiy: Does `__device_backend_tag` mean SYCL backend?

SergeyKopienko: Yes, correct.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: That's exactly `sycl::queue&&` makes a more common solution, because it works for both cases:
1) sumitter{}(q_local); //is passed as an lvalue
2) sumitter{}(exec.queue()); ////is passed as an rvalue

And if I have a case (2), the writing `sycl::queue&&` doesn't force me to create an another instance  of queue `q_local` on the stack and after pass it further to a function which accepts  `sycl::queue&`.

SergeyKopienko: https://godbolt.org/z/35sjPdWb1

MikeDvorskiy: Sorry, I meant  such way
```
template<typename Queue>
void fnc(Queue &&)
```

SergeyKopienko: In our de-facto agreement in all internal API for the `sycl-backend` we works with `sycl::queue` and never describe it as through template type. I believe we should align with this style in this PR too.
@akukanov what do you think?

MikeDvorskiy: > we works with `sycl::queue` and never describe it as through template type

I don't argue with that..
I would like just to have two ways will work:
```
sumitter{}(q_local);
sumitter{}(exec.queue());
```
In the proposed code the second way doesnt work. That is my concern.

SergeyKopienko: I don't agreed with this proposal: we should pay more difficult code in submitters for that.
Also, submitters will have the same problem like they had before this PR: you approach will bring ability to have different instantiations of submitters depends on `sycl::queue` value category. Previously we had this problem `ExecutionPolicy` template parameter.

MikeDvorskiy: Actually, my main point - I have a concern that way "submitter{}(exec.queue());" doesn't work.
Before the change I can write both ways
`submitter{}(exe_local); or submitter{}(obj.exec());`

Regarding to `more difficult code` - 
1) it depends how loot at that.
2) There is  a C/CPP rule - to write a function signature such way that is suitable  and convenient for a caller. 

Other words, I don't insist on the code should be changed.
I am just saying about my concern - sumitter{}(exec.queue()); way doesn't work. And I would get a simple answer why we OK(if OK) with that. For example, an answer "sumitter{}(exec.queue()); way is not a value for oneDPL code" will suit me.

MikeDvorskiy: Update: "different instantiations of submitters depends" also suits me

MikeDvorskiy (approved):
I don't have  any more comments on that PR. And I don't have serios arguments against the PR is to be merged.

include/oneapi/dpl/internal/scan_by_segment_impl.h
SergeyKopienko: @mmichel11 please note to remove the first parameter `oneapi::dpl::__internal::__device_backend_tag` from this function and it's calls when you will later move this submitter it into proper namespace.
Also it should be removed from: `__sycl_scan_by_segment_impl::operator()`, `__parallel_find_or_impl_one_wg::operator()`, `__parallel_find_or_impl_multiple_wgs::operator()`, `__parallel_transform_reduce_small_submitter::operator()`, `__parallel_transform_reduce_device_kernel_submitter::operator()`, `__parallel_transform_reduce_work_group_kernel_submitter::operator()`

SergeyKopienko: Issue created: https://github.com/uxlfoundation/oneDPL/issues/2211

danhoeflinger (approved):
LGTM, I think this gets all the locations we want to remove, without bringing sycl specifics into the "hetero" space.

include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -105,20 +105,17 @@ struct __sycl_scan_by_segment_impl
     template <typename... _Name>
     using _SegScanPrefixPhase = __seg_scan_prefix_kernel<__is_inclusive, _Name...>;
 
-    template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3,
-              typename _BinaryPredicate, typename _BinaryOperator, typename _T>
+    template <typename _BackendTag, typename _Range1, typename _Range2, typename _Range3, typename _BinaryPredicate,
```
SergeyKopienko, line 108: Introduced compile error here: fixed in https://github.com/uxlfoundation/oneDPL/pull/2213 by @dmitriy-sobolev

PR #2201, opened 2025-04-18
Fix naming and namespaces for scan_by_segment

danhoeflinger:
As described https://github.com/uxlfoundation/oneDPL/pull/2198#discussion_r2050818264
This fix provides better naming and namespace structure for `*_scan_by_segment`.  

It makes more clear the divide between hetero and device backend, and provides an entry point above the submitter within the device backend to transition from Policy to sycl queue and KernelName.  It also puts things in the correct namespaces.

I'm considering moving stuff between files, and changing the file structure out of the scope of this PR, but this should also be done at some point...

Should be applied before #2198 (as it provides the appropriate place to transition from policy to queue)

### DIFF
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -378,8 +378,8 @@ __parallel_scan_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Execu
         ::std::forward<_Range1>(__keys), ::std::forward<_Range2>(__values), ::std::forward<_Range3>(__out_values),
         __binary_pred, __binary_op, __init, __identity);
 }
-
 } //namespace __par_backend_hetero
+
 namespace internal
 {
 template <typename _BackendTag, typename Policy, typename InputIterator1, typename InputIterator2,
--- a/include/oneapi/dpl/internal/scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/scan_by_segment_impl.h
@@ -52,7 +52,7 @@ namespace oneapi
 {
 namespace dpl
 {
-namespace internal
+namespace __par_backend_hetero
 {
 
 // This function is responsible for performing an exclusive segmented scan between work items in shared local memory.
@@ -146,7 +146,7 @@ struct __sycl_scan_by_segment_impl
              oneapi::dpl::__internal::__kernel_work_group_size(__exec.queue(), __seg_scan_prefix_kernel)});
 #endif
 
-        ::std::size_t __n_groups = __internal::__dpl_ceiling_div(__n, __wgroup_size * __vals_per_item);
+        ::std::size_t __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __wgroup_size * __vals_per_item);
 
         auto __partials = oneapi::dpl::__par_backend_hetero::__buffer<__val_type>(__n_groups).get_buffer();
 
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -366,12 +366,28 @@ struct __sycl_scan_by_segment_impl
...
+{
+    __sycl_scan_by_segment_impl<__is_inclusive>()(
+        oneapi::dpl::__internal::__device_backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec),
+        ::std::forward<_Range1>(__keys), ::std::forward<_Range2>(__values), ::std::forward<_Range3>(__out_values),
+        __binary_pred, __binary_op, __init, __identity);
+}
+
+} //namespace __par_backend_hetero
+namespace internal
+{
```
SergeyKopienko, line 384:
```suggestion
}
} //namespace __par_backend_hetero

namespace internal
{
```

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -366,12 +366,28 @@ struct __sycl_scan_by_segment_impl
...
+                           _BinaryOperator __binary_op, _T __init, _T __identity)
+{
+    __sycl_scan_by_segment_impl<__is_inclusive>()(
+        oneapi::dpl::__internal::__device_backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec),
+        ::std::forward<_Range1>(__keys), ::std::forward<_Range2>(__values), ::std::forward<_Range3>(__out_values),
+        __binary_pred, __binary_op, __init, __identity);
+}
+} //namespace __par_backend_hetero
+
+namespace internal
```
mmichel11, line 383: Is the `internal` or `__internal` namespace better here? This directory in general uses `internal` but oneDPL patterns tend to be placed in `__internal`. Perhaps this is better addressed when reworking the scan-by-segment algorithms to use reduce-then-scan. I expect to move these to the `pstl/hetero` directory similar to what was done for reduce-by-segment.

danhoeflinger: good point.  Definitely the stronger convention in the repo is `__internal`, so lets do it while we are renaming things.

timmiesmith: > good point. Definitely the stronger convention in the repo is `__internal`, so lets do it while we are renaming things.

When the `by_segment` algorithms were added they did not follow all of the name uglification requirements of the rest of the oneDPL algorithms because they were not part of the algorithms that aligned with the C++17 algorithms.  Moving  to `__internal` to be consistent with the other algorithms while renaming is occurring is the correct thing to do in my opinion.

include/oneapi/dpl/internal/scan_by_segment_impl.h
```diff
@@ -392,13 +408,13 @@ __scan_by_segment_impl_common(__internal::__hetero_tag<_BackendTag>, Policy&& po
 
     constexpr iter_value_t identity = unseq_backend::__known_identity<Operator, iter_value_t>;
 
-    __sycl_scan_by_segment_impl<Inclusive::value>()(_BackendTag{}, ::std::forward<Policy>(policy), key_buf.all_view(),
-                                                    value_buf.all_view(), value_output_buf.all_view(), binary_pred,
-                                                    binary_op, init, identity);
+    __bknd::__parallel_scan_by_segment<Inclusive::value>(
+        _BackendTag{}, ::std::forward<Policy>(policy), key_buf.all_view(), value_buf.all_view(),
```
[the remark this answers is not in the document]
danhoeflinger: fixed

mmichel11 (approved):
Thank you for doing this. The changes LGTM assuming passing CI.

Issue #2204, opened 2025-04-22
Remove duplicated `__pattern_min_element_reduce_fn` functor


mmichel11:
Thank you @SergeyKopienko for pointing this out in https://github.com/uxlfoundation/oneDPL/pull/2200#issuecomment-2815873304.

We have two implementations of `__pattern_min_element_reduce_fn`. The first is [here](https://github.com/uxlfoundation/oneDPL/blob/21b83547f4f2405eab72fcb58d7bfbf3c0075ceb/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h#L859) in `algorithm_ranges_impl_hetero.h` and the second [here](https://github.com/uxlfoundation/oneDPL/blob/21b83547f4f2405eab72fcb58d7bfbf3c0075ceb/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h#L466) in `algorithm_impl_hetero.h`

The definitions are different with the second having different codepaths for SPIR-V and non SPIR-V targets. The implementation of the underlying reduce implementation has changed since this was added. We should reevaluate the need for these separate implementations and consolidate `__pattern_min_element_reduce_fn` into a single instance.

mmichel11:
After experimentation, the non-commutative specialization still performs best on Intel GPU Series Max 1550, so it should remain as is. Coalescing is needed to achieve the best memory access pattern on other architectures, so the non SPIR-V path should also remain.

PR #2200, opened 2025-04-18
Remove some duplicated structures

SergeyKopienko:
In this PR we remove some duplicated structures which were introduces in the PR https://github.com/uxlfoundation/oneDPL/pull/2153

Some structures has duplicated implementations and has been moved into include/oneapi/dpl/pstl/hetero/utils_hetero.h :
- `__pattern_minmax_element__reduce_fn` (renamed to `__pattern_minmax_element_reduce_fn`);
- `__pattern_minmax_element_transform_fn`;
- `__pattern_count_transform_fn`;
- `__pattern_min_element_transform_fn`.

Also `struct __pattern_min_element_reduce_fn` has two different implementations...

Just for note: before https://github.com/uxlfoundation/oneDPL/pull/2153 this code has been duplicated in the `lambda`-form

PR #2205, opened 2025-04-22
Remove duplicate `__pattern_min_element_reduce_fn` implementations

mmichel11:
This PR is a follow-up to https://github.com/uxlfoundation/oneDPL/pull/2200 and addresses https://github.com/uxlfoundation/oneDPL/issues/2204. We remove the duplicate definitions of `__pattern_min_element_reduce_fn` and place the implementation in the `utils_hetero.h` header.

The two existing implementations are slightly different. The version of the functor that differentiates between SPIR-V and non SPIR-V target backends is chosen to keep as it continue to show the best performance across assessed platforms.

I do not think this is considered a feature but am okay with holding off on merging until after the milestone's completion if we wish.


SergeyKopienko (approved):
LGTM

PR #2206, opened 2025-04-25
Resolve ambiguous return in `__red_by_seg_op` in SYCL backend

mmichel11:
`__red_by_seg_op` has a potential ambiguous return when the return type of `__binary_op` differs from the input segment's underlying value type. To resolve this issue, we can ensure the second element in the tuple returned by `__red_by_seg_op` matches the return type of `__binary_op` through a conversion.

In practice, this issue was encountered when performing a `reduce_by_segment` operation over zip iterators where a `std::tuple` is explicitly returned by the reduction operator. I have added a test case which catches this issue in `reduce_by_segment_zip.pass`

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -806,6 +806,7 @@ struct __red_by_seg_op
     operator()(const _Tup1& __lhs_tup, const _Tup2& __rhs_tup) const
     {
         using std::get;
+        using _OpOutType = decltype(__binary_op(get<1>(__lhs_tup), get<1>(__rhs_tup)));
         // The left-hand side has processed elements from the same segment, so update the reduction value.
         if (get<0>(__rhs_tup) == 0)
         {
@@ -813,7 +814,7 @@ struct __red_by_seg_op
                                                        __binary_op(get<1>(__lhs_tup), get<1>(__rhs_tup)));
         }
         // We are looking at elements from a previous segment so just update the output index.
-        return oneapi::dpl::__internal::make_tuple(get<0>(__lhs_tup) + get<0>(__rhs_tup), get<1>(__rhs_tup));
+        return oneapi::dpl::__internal::make_tuple(get<0>(__lhs_tup) + get<0>(__rhs_tup), _OpOutType{get<1>(__rhs_tup)});
     }
     _BinaryOp __binary_op;
 };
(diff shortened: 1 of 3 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -806,14 +806,16 @@ struct __red_by_seg_op
     operator()(const _Tup1& __lhs_tup, const _Tup2& __rhs_tup) const
```
adamfidel, line 806: Is it possible to fix the disambiguation with a trailing return type instead of the cast in the second return?
```suggestion
    operator()(const _Tup1& __lhs_tup, const _Tup2& __rhs_tup) const -> decltype(oneapi::dpl::__internal::make_tuple(get<0>(__lhs_tup), __binary_op(get<1>(__lhs_tup), get<1>(__rhs_tup)))
```

I'm not sure if that's better (it's arguably less readable). Just wondering if that would have any difference in terms of copies made as I do not know when the cast to `_OpOutType` would introduce a copy with these changes.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -806,14 +806,16 @@ struct __red_by_seg_op
...
     {
         using std::get;
+        using _OpOutType = decltype(__binary_op(get<1>(__lhs_tup), get<1>(__rhs_tup)));
```
adamfidel, line 809:
```suggestion
        using _OpReturnType = decltype(__binary_op(get<1>(__lhs_tup), get<1>(__rhs_tup)));
```
I think `_OpReturnType` is a slightly clearer name but either works in my opinion.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I think that the number of copies is likely the same for either option, but I may be mistaken. (perhaps this one is worse because it may require copying the whole tuple, I'm not sure...)

danhoeflinger (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I wrote a simple example to test this with the conversion operator from our internal tuple to `std::tuple`. The trailing return type and the cast both result in a single call to this operator so either option works the same.

I think the cast is slightly more readable, so I chose this one with the updated name.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: I agree it's better and made this change.

danhoeflinger (approved):
LGTM

adamfidel (approved):
LGTM now!

PR #2207, opened 2025-04-25
Adding definition of DPLROOT variable to make the library consistent with other oneAPI libraries

timmiesmith:
Installations of oneDPL include an environment activation scripts which define environment variables as needed for DPL to work.

In particular, it defines an environment variable with name 'DPL_ROOT', which points at the path where the DPL installation is to be found.

The name "DPL_ROOT" with an underscore is inconsistent with the rest of oneAPI products, which define similar variables without underscores in their names.

For example:

MKLROOT
TBBROOT
IPPROOT
DALROOT

This PR introduces `DPLROOT` to make the library consistent with other oneAPI libraries to improve the user experience. It mirrors `DPL_ROOT` so no invasive changes to the library are immediately necessary and a plan to deprecate `DPL_ROOT` can be implemented.

### DIFF
--- a/integration/env/vars.bat
+++ b/integration/env/vars.bat
@@ -18,6 +18,7 @@ REM
 set "VARSDIR=%~dp0"
 
 if not defined DPL_ROOT for /f "delims=" %%F in ("%VARSDIR%..") do set "DPL_ROOT=%%~fF"
+if not defined DPLROOT for /f "delims=" %%F in ("%VARSDIR%..") do set "DPLROOT=%%~fF"
 
 set "CPLUS_INCLUDE_PATH=%DPL_ROOT%\include;%CPLUS_INCLUDE_PATH%"
 set "PKG_CONFIG_PATH=%DPL_ROOT%\lib\pkgconfig;%PKG_CONFIG_PATH%"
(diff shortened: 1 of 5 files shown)


SergeyKopienko:
I feel probably it's not all required changes for this task if we take a look at https://github.com/search?q=repo%3Auxlfoundation%2FoneDPL%20DPL_ROOT&type=code

timmiesmith:
> I feel probably it's not all required changes for this task if we take a look at https://github.com/search?q=repo%3Auxlfoundation%2FoneDPL%20DPL_ROOT&type=code

I am not trying to replace `DPL_ROOT` with `DPLROOT` at this point.  I am introducing the new variable now so that developers in other projects that are familiar with the pattern from other projects like oneTBB and oneMath will find the corresponding variable for oneDPL.  I think the update to use it internally and deprecation of `DPL_ROOT` should follow in later milestones.

abagusetty:
Just tagging: DPNP project raised a similar issue with DPLROOT vs DPL_ROOT: https://github.com/IntelPython/dpnp/issues/1319

dmitriy-sobolev (approved):
LGTM

PR #2212, opened 2025-04-30
Remove extra `__device_backend_tag` parameter from sycl-submitters and functions

SergeyKopienko:
In this PR we remove `__device_backend_tag` parameter in the cases when we have
`oneapi::dpl::__internal::__device_backend_tag, sycl::queue& __q` in sycl-submitters::operator() or in the functions of hetero-backend in the namespace `oneapi::dpl::__par_backend_hetero`.

Justification - the `__device_backend_tag` parameter doesn't required as it's not brings to us any semantic: we already have `sycl::queue` in params so implements something for sycl-backend.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
@@ -381,9 +381,8 @@ struct __histogram_general_local_atomics_submitter<__iters_per_work_item,
 template <typename _CustomName, std::uint16_t __iters_per_work_item, typename _Range1, typename _Range2,
           typename _BinHashMgr>
 sycl::event
-__histogram_general_local_atomics(oneapi::dpl::__internal::__device_backend_tag, sycl::queue& __q,
-                                  const sycl::event& __init_event, ::std::uint16_t __work_group_size, _Range1&& __input,
-                                  _Range2&& __bins, const _BinHashMgr& __binhash_manager)
+__histogram_general_local_atomics(sycl::queue& __q, const sycl::event& __init_event, ::std::uint16_t __work_group_size,
+                                  _Range1&& __input, _Range2&& __bins, const _BinHashMgr& __binhash_manager)
 {
     using _iters_per_work_item_t = ::std::integral_constant<::std::uint16_t, __iters_per_work_item>;
 
@@ -523,8 +522,8 @@ __parallel_histogram_select_kernel(oneapi::dpl::__internal::__device_backend_tag
              __local_mem_size)
     {
         return __future(__histogram_general_local_atomics<_CustomName, __iters_per_work_item>(
-            __backend_tag, __q, __init_event, __work_group_size, std::forward<_Range1>(__input),
-            std::forward<_Range2>(__bins), __binhash_manager));
+            __q, __init_event, __work_group_size, std::forward<_Range1>(__input), std::forward<_Range2>(__bins),
+            __binhash_manager));
     }
     else // otherwise, use global atomics (private copies per workgroup)
     {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
@@ -477,10 +477,9 @@ struct __histogram_general_private_global_atomics_submitter<__internal::__option
 };
 template <typename _CustomName, typename _Range1, typename _Range2, typename _BinHashMgr>
 sycl::event
-__histogram_general_private_global_atomics(oneapi::dpl::__internal::__device_backend_tag, sycl::queue& __q,
-                                           const sycl::event& __init_event, ::std::uint16_t __min_iters_per_work_item,
-                                           ::std::uint16_t __work_group_size, _Range1&& __input, _Range2&& __bins,
-                                           const _BinHashMgr& __binhash_manager)
+__histogram_general_private_global_atomics(sycl::queue& __q, const sycl::event& __init_event,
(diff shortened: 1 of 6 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
```diff
@@ -288,9 +288,9 @@ struct __histogram_general_registers_local_reduction_submitter<__iters_per_work_
 template <typename _CustomName, std::uint16_t __iters_per_work_item, std::uint8_t __bins_per_work_item,
           typename _Range1, typename _Range2, typename _BinHashMgr>
 sycl::event
-__histogram_general_registers_local_reduction(oneapi::dpl::__internal::__device_backend_tag, sycl::queue& __q,
-                                              const sycl::event& __init_event, ::std::uint16_t __work_group_size,
-                                              _Range1&& __input, _Range2&& __bins, const _BinHashMgr& __binhash_manager)
+__histogram_general_registers_local_reduction(sycl::queue& __q, const sycl::event& __init_event,
+                                              ::std::uint16_t __work_group_size, _Range1&& __input, _Range2&& __bins,
```
danhoeflinger, line 292:
```suggestion
                                              std::uint16_t __work_group_size, _Range1&& __input, _Range2&& __bins,
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
```diff
@@ -404,11 +403,11 @@ struct __histogram_general_private_global_atomics_submitter;
 template <typename... _KernelName>
 struct __histogram_general_private_global_atomics_submitter<__internal::__optional_kernel_name<_KernelName...>>
 {
-    template <typename _BackendTag, typename _Range1, typename _Range2, typename _BinHashMgr>
+    template <typename _Range1, typename _Range2, typename _BinHashMgr>
     sycl::event
-    operator()(_BackendTag, sycl::queue& __q, const sycl::event& __init_event,
-               ::std::uint16_t __min_iters_per_work_item, ::std::uint16_t __work_group_size, _Range1&& __input,
-               _Range2&& __bins, const _BinHashMgr& __binhash_manager)
+    operator()(sycl::queue& __q, const sycl::event& __init_event, ::std::uint16_t __min_iters_per_work_item,
```
danhoeflinger, line 408:
```suggestion
    operator()(sycl::queue& __q, const sycl::event& __init_event, std::uint16_t __min_iters_per_work_item,
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
```diff
@@ -404,11 +403,11 @@ struct __histogram_general_private_global_atomics_submitter;
...
+               ::std::uint16_t __work_group_size, _Range1&& __input, _Range2&& __bins,
```
danhoeflinger, line 409:
```suggestion
               std::uint16_t __work_group_size, _Range1&& __input, _Range2&& __bins,
```

danhoeflinger: I see at least one more, I think you have a script that can fix all these, I wont call them out anymore.

SergeyKopienko: will fix

danhoeflinger (commented):
In principle this looks good to me pending minor nit picks.  If we have a queue, we have already made the transition to the SYCL backend and no longer need any tag (or execution policy).

I have not searched yet to confirm that these are the only instances to update...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
SergeyKopienko: fixed.

danhoeflinger (approved):
LGTM, 
It may make sense to raise this to the group prior to merging to confirm that there is no objections to removing the tag after we have made the conversion.
I could see an argument that it provides some more clear "documentation" of where the function lies, that being said, I'm in favor of this PR.

akukanov:
The patch seemingly just completes what was already mostly there, that is - the API of the SYCL backend being SYCL-tailored, exposing queues, events etc. The tag plays no role in this case and can/should be removed.

The architectural question I have is therefore not directly for this patch, but maybe for previous iterations of refactoring: do we (want to) assume that there is only one hetero backend based on SYCL? Or, in other words, did not we switch to SYCL specifics too early during the template dispatch? But this question does not preclude acceptance of this PR - if/when architectural changes become needed, I assume those to be broader anyway.

danhoeflinger:
> The architectural question I have is therefore not directly for this patch, but maybe for previous iterations of refactoring: do we (want to) assume that there is only one hetero backend based on SYCL? Or, in other words, did not we switch to SYCL specifics too early during the template dispatch? But this question does not preclude acceptance of this PR - if/when architectural changes become needed, I assume those to be broader anyway.

I think that there is a space left in the "hetero" section of the code, prior to the sycl specific device backend which still allows for other "hetero" backends.  However, our only specific device backend is called `__device_backend_tag` when perhaps it should be named `__sycl_backend_tag`.  I think if we make that change, then there is room to add other device backend tags with implementations to match and dispatch to different "hetero" backends.  This was the major change which required the revert and redo of the main patch.

rarutyun (commented):
I also thing that the PR is good as is. Of course, I haven't checked that all `backend_tag` propagations are removed when we are already on the backend level. I just trust that Sergey does a right thing. No architectural concerns - the PR is doing what we previously agreed on.

rarutyun (approved):

PR #2214, opened 2025-04-30
Fix an error in implementation of `transform-reduce`

SergeyKopienko:
In this PR we fix an error in implementation of `transform-reduce` :
 - the problem has place when the size of the source data is not a multiple of the group size and there are work items for which there is no source data.


SergeyKopienko:
@MikeDvorskiy please check this changes with your PR about ranges & tests

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -66,22 +66,25 @@ __work_group_reduce_kernel(const _NDItemId __item_id, const _Size __n, const _Si
...
-    // 1. Initialization (transform part). Fill local memory
-    __transform_pattern(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0, __is_full,
-                        /*__n_groups*/ (_Size)1, __result, __acc...);
 
-    const _Size __n_items = __transform_pattern.output_size(__n, __group_size, __iters_per_work_item);
-    // 2. Reduce within work group using local memory
-    __result.__v = __reduce_pattern(__item_id, __n_items, __result.__v, __local_mem);
-    if (__local_idx == 0)
+    // 1. Initialization (transform part). Fill local memory
+    if (__transform_pattern(__item_id, __n, __iters_per_work_item, /*global_offset*/ (_Size)0, __is_full,
```
mmichel11, line 72: Is it possible for this to return true for some items in a work-group and false for others? If so, this may result in hangs as we use group barriers within this conditional.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -542,6 +548,8 @@ struct transform_reduce
             const _Size __adjusted_global_id_plus_one = __adjusted_global_id + 1;
             scalar_reduction_remainder(__adjusted_global_id_plus_one, __adjusted_n, __res, __acc...);
         }
```
mmichel11, line 550: If I understand correctly, the implicit `else` under this line should actually return false. This is because our `__adjusted_global_id` never processed any work.

mmichel11 (commented):
Do we have a reproducer of a case the causes this issue?  I'm curious to take a look.

SergeyKopienko:
> Do we have a reproducer of a case that causes this issue? I'm curious to take a look.

@MikeDvorskiy has an example.

SergeyKopienko:
Looks like as incorrect approach for this fix.

PR #2157, opened 2025-04-01
Comment/remove unused parameter `_ExecutionPolicy&& __exec` in some functions

SergeyKopienko:
In this PR we made the next changes:
- Remove `_ExecutionPolicy` patameter from `__parallel_strict_scan_body` function as not required anymore;
- Remove `_ExecutionPolicy` parameter from `__parallel_strict_scan` function as not required anymore;
- Comment unused template parameter `_ExecutionPolicy` in `__pattern_transform_scan` function.

This changes required due the PR https://github.com/uxlfoundation/oneDPL/pull/2146 has been merged into main branch.

PR #2216, opened 2025-05-01
Restore execution policy at mid-level (algorithm_impl)

dmitriy-sobolev:
This PR restores execution policy parameter in `__parallel_strict_scan` used by multiple backends, which was removed in #2157. 
Thanks @danhoeflinger for [noticing it](https://github.com/uxlfoundation/oneDPL/pull/2216#discussion_r2070222326).

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -1317,7 +1317,7 @@ __pattern_copy_if(_Tag, _ExecutionPolicy&&, _ForwardIterator __first, _ForwardIt
 template <class _IsVector, class _ExecutionPolicy, class _RandomAccessIterator1, class _RandomAccessIterator2,
           class _UnaryPredicate>
 _RandomAccessIterator2
-__pattern_copy_if(__parallel_tag<_IsVector>, _ExecutionPolicy&&, _RandomAccessIterator1 __first,
+__pattern_copy_if(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomAccessIterator1 __first,
                   _RandomAccessIterator1 __last, _RandomAccessIterator2 __result, _UnaryPredicate __pred)
 {
     using __backend_tag = typename __parallel_tag<_IsVector>::__backend_tag;
@@ -1596,7 +1596,7 @@ __brick_calc_mask_2(_RandomAccessIterator __first, _RandomAccessIterator __last,
 template <class _IsVector, class _ExecutionPolicy, class _RandomAccessIterator1, class _RandomAccessIterator2,
           class _BinaryPredicate>
 _RandomAccessIterator2
-__pattern_unique_copy(__parallel_tag<_IsVector>, _ExecutionPolicy&&, _RandomAccessIterator1 __first,
+__pattern_unique_copy(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomAccessIterator1 __first,
                       _RandomAccessIterator1 __last, _RandomAccessIterator2 __result, _BinaryPredicate __pred)
 {
     using __backend_tag = typename __parallel_tag<_IsVector>::__backend_tag;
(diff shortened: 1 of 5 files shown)


include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3262,12 +3262,12 @@ __pattern_includes(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __exec, _
...
 
-template <class _IsVector, class _ExecutionPolicy, class _RandomAccessIterator1, class _RandomAccessIterator2,
-          class _OutputIterator, class _Compare, class _SizeFunction, class _SetOP>
+template <class _IsVector, class _RandomAccessIterator1, class _RandomAccessIterator2, class _OutputIterator,
+          class _Compare, class _SizeFunction, class _SetOP>
 _OutputIterator
-__parallel_set_op(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomAccessIterator1 __first1,
-                  _RandomAccessIterator1 __last1, _RandomAccessIterator2 __first2, _RandomAccessIterator2 __last2,
-                  _OutputIterator __result, _Compare __comp, _SizeFunction __size_func, _SetOP __set_op)
+__parallel_set_op(__parallel_tag<_IsVector>, _RandomAccessIterator1 __first1, _RandomAccessIterator1 __last1,
```
danhoeflinger, line 3268: This actually concerns me, and makes me think that #2157 was merged with issues in it.
Our design agreement was to keep the execution policy until we reached a *specific backend*, and then the backend could decide what it wants to do with it (i.e. translate it to whatever backend specifics it needs to extract).

This code is still backend agnostic / generic, so should not be discarding the execution policy.

Right now, no host backend requires anything from the execution policy to be extracted, but that may not always be true.

danhoeflinger: @SergeyKopienko @rarutyun  @akukanov @timmiesmith 
What do we want to do with this?

I don't necessarily think its an incredibly urgent or dangerous thing, but its certainly something we will need to closely consider and patch to fix, before or after the release.

I don't recommend reverting any prior commits at this stage, but I do think we need to audit the current main at some point and closely consider if that specific rule is always followed:

"Do not discard the execution policy until we are in a specific backend."

danhoeflinger: To explain a little further this specific case, I think the appropriate thing instead of removing ExecutionPolicy here as a parameter, we should restore it to `__parallel_strict_scan`, and only discard the policy within the tbb / omp backends.  
I'm not sure how widespread this is otherwise.  

We already addressed this (as best I can determine) for SYCL device backend, but we need to fix this for the host backends also.

danhoeflinger: Sorry for the spam,  reviewing the code a bit further, it seems like `__parallel_strict_scan` may be the only problematic function which doesn't pass the `_ExecutionPolicy` into the specific host backend.  I'd recommend just adapting this PR to fix that problem instead of its current strategy.
I also noticed one tbb definition which has an unused param `__exec`: `__parallel_stable_sort` in `include/oneapi/dpl/pstl/parallel_backend_tbb.h`

@dmitriy-sobolev , let me know if you want me to create a commit to do this, or if it makes sense to you and you have time to address it.

dmitriy-sobolev: Ah, `algorithm_impl.h` is not the backend level yet. Let me fix it.

danhoeflinger (approved):
The code LGTM, sorry for the false alarm earlier, but I think it ended up in with a good result.
Especially with the PR name change, I'm happy with it.

mmichel11 (approved):
LGTM as well.

SergeyKopienko (commented):
LGTM

SergeyKopienko (commented):
LGTM

PR #2220, opened 2025-05-01
WIP: Custom backends for Dynamic Selection

vossmjp:
A very rough draft for custom backends for dynamic selection.

### DIFF
--- a/rfcs/proposed/dynamic_selection_customization/README.md
+++ b/rfcs/proposed/dynamic_selection_customization/README.md
@@ -1,0 +1,495 @@
+# Simplified Customization of Backends and Policies for Dynamic Selection
+
+Dynamic Selection is a Technology Preview feature 
+[documented in the oneDPL Developer Guide](https://www.intel.com/content/www/us/en/docs/onedpl/developer-guide/2022-8/dynamic-selection-api.html)
+and its current design is described by 
+[an experimental RFC](https://github.com/uxlfoundation/oneDPL/tree/main/rfcs/experimental/dynamic_selection).
+When applying Dynamic Selection to use cases, there is often the desire to add a new
+backend or new selection policy. While it is possible to add new policies or backends
+by following the contract described in the current design, the process is non-trivial
+and requires some (unecessarily) verbose code.
+
+## Current backend design
+
+A backend defines resource types and implements reporting that is requireed by
+a policy.
+
+As described in [the current design](https://github.com/uxlfoundation/oneDPL/tree/main/rfcs/experimental/dynamic_selection),
+type `T` satisfies the *Backend* contract if given,
+
+- `b` an arbitrary identifier of type `T`
+- `args` an arbitrary parameter pack of types `typename… Args`
+- `s` is of type `S` and satisfies *Selection* and `is_same_v<resource_t<S>, resource_t<T>>` is `true`
+- `f` a function object with signature `wait_t<T> fun(resource_t<T>, Args…);`
+
+| Required Member Function | Description |
+| --------------------- | ----------- |
+| `submit(s, f, args…)` | Returns an object that satisfies *Submission*. The function invokes `f` but does not wait for the `wait_t<T>` object returned by it. |
+| `get_resources()` | Returns a `std::vector<resource_t<T>>`. |
+| `get_submission_group()` | Returns an object that has a member function `void wait()`. Calling this `wait` function blocks until all previous submissions to this backend are complete. |
+
+## Use Cases to Improve for Backends
+
+| Use Case | Description |
+| --------------------- | -------- |
+| Default Backend | User only provides universe, no custom backend at all. |
+| Uninstrumented Backend | User defines small part of backend: `wait_type`, `submission_group`. Better waiting support. |
+| Instrumented Backend | User provides instrumentation for reporting |
+
+We remind readers of this proposal that the requirements of *Submission* and *Selection* objects are provided in
+[the current design description](https://github.com/uxlfoundation/oneDPL/tree/main/rfcs/experimental/dynamic_selection).
+
+### Default Backend
+
+For policies that require no reporting of `execution_info`, it should be simple to use a new resource type,
+perhaps even without writing a custom backend. For example, what if we want to round-robin through pointers to
+queues instead of queues. 
+
+```cpp
+    sycl::queue q1, q2;
+    ex::round_robin_policy p{ { &q1, &q2 } };
+
+    auto s1 = onedpl::dpl::experimental::submit(p, [](sycl::queue *qp) -> sycl::queue {
+        qp->submit(/*...*/);
+        return *q;
+    }
+
+    auto s2 = onedpl::dpl::experimental::submit(p, [](sycl::queue *qp) -> sycl::queue {
+        qp->submit(/*...*/);
+        return *q;
+    }
+
+    wait(s1);
+    wait(s2);
+```
+
+#### submit
+
+The main role that `submit` plays is to add instrumentation around the call to `f`.
+Experience has shown us that most backends perform four basic steps in their implementation
+of the `submit` function:
+
+1. Do any setup needed for implementing reporting before calling `f`.
+2. The function `f` is called.
+3. Do any setup needed for implementing reporting after calling `f`, perhaps using what was returned by `f`.
+4. The *Submission* object, which typically wraps what is returned by `f`, is constructed and returned.
+
+If a backend writer doesn't care if their backend works with policies that require reporting, then
+they can use a default implementation of submit that simply calls `f` and returns the backends *Submission*
+object. A *Submission* object must support `s.wait()` and `s.unwrap`. A default backend cannot meaningfully
+implement `wait` on an arbitrary type, but could easily wrap a type, such as `sycl::queue` or `tbb::task_group`
+that provides a `wait` member function and call that `wait` member function if it finds it, and in other cases
+provide an empty `wrap` function. `s.unwrap` can return whatever was returned by the user's function. Our simple
+example above returns a `sycl::queue` from its function and so a default backend can easily synthesize a meaningful
+*Submission* type from it, since `q.wait()` is valid.
+
+#### get_resources
+
+The `get_resources` function can default to returning the vector that is passed to the policy, and
+then down to the backend's constructor. A policy that is created with a default constructor will
+have no resources in this case.
+
+#### get_submission_group
+
+The `get_submission_group` is not easily implemented in a meaningful way. It must return a type that
+defines a member function `wait`. But since there is no way to know how to wait on all previous submissions
+for an arbitrary backend resource, this function will likely need to return a dummy type or 
+`get_submission_group` be undefined for the default backend.
+
+### Uninstrumented Backend
+
+It may be possible that a backend writer wants to do a little bit of work, so that `get_submission_group` works
+properly, and/or that there is a meaningful *Submission* type defined, and/or there is a default set of resources
+defined even when called with no explicit universe. But, they may not want to go through the effort of
+implementing the reporting mechanism. In that case, we should define a way to customize the default
+backend for the resource type in a way to opt-in to only those parts that should be customized.
+
+For example, we may round-robin through pointers to queues, have a default set of resources
+and/or wait on the submission group:
+
+```cpp
+    ex::round_robin_policy<sycl::queue*> p;
+
+    auto r = p.get_resources(); // returns some meaningful default set
+
+    auto s1 = onedpl::dpl::experimental::submit(p, [](sycl::queue *qp) -> sycl::queue {
+        qp->submit(/*...*/);
+        return *q;
+    }
+
+    auto s2 = onedpl::dpl::experimental::submit(p, [](sycl::queue *qp) -> sycl::queue {
+        qp->submit(/*...*/);
+        return *q;
+    }
+
+    wait(p.get_submission_group());
+```
+
+#### submit
+
+See [Default Backend](#default-backend).
+
+#### get_resources
+
+Can opt-in to defining a `get_resources` function that will determine a useful universe of resources even if
+the backend was default constructed. For example, a SYCL backend might use `get_devices`.
+
+#### get_submission_group
+
+Can opt-in to defining a `get_submission_group` function that will return a non dummy type.
+
+### Instrumented Backend
+
+As desribed in the [Default backend](#default-backend) section, the `submit` function generally
+performs four steps, two of which are related to instrumentation. We proposed adding two new
+functions `instrument_before` and `instrument_after` that can be used to opt-in to instrumentation
+in a backend. A backend writer will not need to fully define `submit` but instead only write the
+code to set up and reporting needed for the backend.
+
+For example, we may using auto-tune for pointers to queues:
+
+```cpp
+    sycl::queue q1, q2;
+    ex::auto_tune_policy p{ { &q1, &q2 } };
+
+   for (int i = 0; i < 100; ++i) {
+        auto s = onedpl::dpl::experimental::submit(p, [](sycl::queue *qp) -> sycl::queue {
+            qp->submit(/*...*/);
+            return *q;
+        }
+        wait(s);
+   }
+```
+
+#### submit
+
+```cpp
+    template <typename SelectionHandle, typename Function, typename... Args>
+    auto submit(SelectionHandle s, Function&& f, Args&&... args) {
+        instrument_before(s); // do insrumentation before calling `f`
+
+        return instrument_after( // do insrumentation before calling `f`, and create Submission type
+            s,
+            std::forward<Function>(f)(unwrap(s), std::forward<Args>(args)...) // call f
+        );
+    }
+```
+
+#### get_resources and get_submission_group
+
+An instrumented backend, can independently opt-in to customizing other functions to better
+support `get_submission_group`, `get_resources` and to define the *Submission* type.
+See [Uninstrumented Backend](#uninstrumented-backend).
+
+## Backend Customization Approach
+
+We proposed using a CRTP (Curiously Recurring Template Pattern) to define backends since this
+will allow developers to flexibly mix-in functionaility from the base. We decided against free
+functions or customization points, because we found that most backends require some kind of state
+to be maintained (such as the universe) and accessing that state is easiest from within a backend
+object. We also think that static polymorphism is sufficient.
+
+The base will provide common-sense default behaviours but a backend can selectively customization behaviours
+via static polymorphism. The base class will provide default implementations of `get_resources` (returning
+the vector passed to the constructor), `get_submission_group` (returning a dummy type) and `submit`. 
+There will be protected `_impl` functions that can be overridden in the derived classes to change the 
+default behaviors.
+
+There will also be a struct `scratch_t` that can be used to provide space in a *Selection* needed by
+the backend for reporting. For example, space to store a beginning time in `instrument_before` and
+ending event in `instrument_after` and so on. `scratch_t` is a template that will receives
+`execution_info` parameters and so can be specialized for a backend to only include the space needed
+by the policy that uses it. For example, a backend can choose to only include space for a
+beginning time if the policy requires reporting of `execution_info::task_time`.
+
+Show below is a sketch of a `backend_base`:
+
+```cpp
+    template<typename ResourceType, typename Backend>
+    class backend_base
+    {
+    public:
+        using resource_type = ResourceType;
+
+        // space needed by the backend in the Selection type
+        // for implementing reporting, by default an empty struct
+        template<typename ...Info> struct scratch_t;
+
+        // constructors
+        backend_base() {}
+        backend_base(const std::vector<resource_type>& u) : resources_{v} {}
+
+        template <typename SelectionHandle, typename Function, typename... Args>
+        auto submit(SelectionHandle s, Function&& f, Args&&... args) {
+            return static_cast<Backend*>(this)->submit_impl(s, f, std::forward<Args>(args)...);
+        }
+
+        std::vector<resource_type> get_resources() {
+            return static_cast<Backend*>(this)->get_resources_impl();
+        }
+
+        auto get_submission_group() {
+            return static_cast<Backend*>(this)->get_submission_group();
+        }
+
+        template <typename SelectionHandle, typename Function, typename... Args>
+        auto submit(SelectionHandle s, Function&& f, Args&&... args) {
+            return static_cast<Backend*>(this)->submit_impl(s, f, std::forward<Args>(args)...);
+        }
+
+    protected:
+
+        //
+        // The default functionality that can be overriden in derived classes
+        //
+
+        std::vector<resource_type> resources_;
+
+        template <typename SelectionHandle, typename Function, typename... Args>
+        auto submit_impl(SelectionHandle s, Function&& f, Args&&... args);
+
+        template <typename SelectionHandle>
+        void instrument_before_impl(SelectionHandle s);
+
+        template <typename SelectionHandle, typename WaitType>
+        auto instrument_after_impl(SelectionHandle s, WaitType w);
+
+        std::vector<resource_type> get_resources_impl() const noexcept;
+
+        auto get_submission_group_impl();
+
+    };
+```
+
+We expect to provide a catch-all [default backend](#default-backend) for uncustomized resource types:
+
+```cpp
+    // A general default backend for ResourceType
+    template< typename ResourceType >
+    class default_backend : public backend_base<ResourceType, default_backend<ResourceType>> {
+    public:
+        using resource_type = ResourceType;
+    };
+```
+
+#### The Default submit
+
+The default implementation therefore eliminates the need to implement `submit` for backends that
+follow the usual four step pattern. Instead, backends can optionally implement `instrument_before` 
+and `instrument_after`.  The code snippet below shows how this pattern might be implemented in the
+`submit_impl` by calling `instrument_before_impl` and `instrument_after_impl`.  The `async_waiter`
+type in the base class is the *Submission* type and wraps a type that must provide a `wait` function
+(for example `sycl::queue` or `tbb::task_group`).
+
+```cpp
+    template<typename ResourceType, typename Backend>
+    class backend_base
+    {
+        // ...
+
+    protected:
+
+        // default submit implements the usual pattern
+        template <typename SelectionHandle, typename Function, typename... Args>
+        auto submit_impl(SelectionHandle s, Function&& f, Args&&... args)
+        {
+            static_cast<Backend*>(this)->instrument_before_impl(s);
+            return static_cast<Backend*>(this)->instrument_after_impl(
+                s, 
+                std::forward<Function>(f)(oneapi::dpl::experimental::unwrap(s), 
+                                          std::forward<Args>(args)...)
+            );
+        }
+
+        // by default no instrumentation before the call
+        template <typename SelectionHandle>
+        void instrument_before_impl(SelectionHandle /*s*/) { }
+
+        // by default no instrumentation after the call
+        // But it returns the wait_type 
+        template <typename SelectionHandle, typename WaitType>
+        auto instrument_after_impl(SelectionHandle /*s*/, WaitType w)
+        {
+            return async_waiter{w}; // async_waiter is described below
+        }
+    
+    private:
+
+        class async_waiter
+        {
+            wait_type w_;
+        public:
+            async_waiter(wait_type w) : w_{w} {}
+            void wait() { w_.wait(); }
+            wait_type unwrap() { return w_; }
+        };
+
+        // ...
+    };
+```
+
+#### The Default get_resources
+
+For many cases, backends hold the set of resources that are passed to their
+constructors and then return that set when `get_resouces` is called. The
+default implementation in the base class therefore takes that approach:
+
+```cpp
+    template<typename ResourceType, typename Backend>
+    class backend_base
+    {
+        // ...
+
+    protected:
+        // default returns the vector of resources
+        resource_container_t get_resources_impl() const noexcept {
+            return resources_;
+        }
+
+        std::vector<resource_type> resources_;
+    };
+```
+
+If a backend wants to provide a set of resources for a policy/backend that
+uses default construction, they can override this function in their derived
+type to determine that default set.
+
+#### Customizing get_submission_group
+
+As mentioned earlier there is not a generally helpful default implementation of `get_submission_group`,
+but since some backends may want to skip this functionality, we propose a default implementation
+that proves an empty implementation.
+
+```cpp
+    template<typename ResourceType, typename Backend>
+    class backend_base
+    {
+        // ...
+
+    protected:
+        // default returns an empty submission group
+        auto get_submission_group_impl()
+        {
+            return submission_group{};
+        }
+
+    private:
+
+        class submission_group
+        {
+        public:
+            void wait() {
+            }
+        };
+    };
+```
+
+One possible alternative is to iterate over the resources and call `wait` on
+each resource. This might work for some cases such as `sycl::queue` and
+`tbb::task_group`.  Another alternative is to capture all *Submission* objects
+returned by `submit` and iterate over those. 
+
+#### Default scratch space
+
+In the base class, there is a struct `scratch_t`. To implement reporting, a backend
+may need per-selection scratch space to store values. For example, a beginning time,
+a `sycl::event`, etc. These will often be set during `instrument_before` or
+`instrument_after` and then read later, perhaps in `lazy_report`. The `scratch_t`
+defined by a backend is added to the *Selection* type created by the policy for 
+use by the backend.
+
+Policies include the space using a backend trait only for the `execution_info` types
+that they need:
+
+```cpp
+// within oneapi::dpl::experimental
+namespace backend_traits
+{
+    template <typename Backend, typename ...Req>
+    inline constexpr bool scratch_space_v = internal::has_scratch_space<Backend, Req...>::value;
+
+    template <typename Backend, typename ...Req>
+    using selection_scratch_t = typename scratch_trait_t_impl<Backend, backend_traits::scratch_space_v<Backend, Req...>,Req...>::type;
+} //namespace backend_traits
+```
+
+## Backend Examples
+
+### Default Backend for NUMA Nodes
+
+```cpp
+    std::vector<tbb::numa_node_id> numa_nodes = tbb::info::numa_nodes();
+
+    std::vector<tbb::task_arena> arenas(numa_nodes.size());
+    std::vector<tbb::task_group> groups(numa_nodes.size());
+    using pair_t = std::pair<tbb::task_arena*, tbb::task_group*>
+    struct wait_type {
+        tbb::task_group &tg_;
+        void wait() { tg_.wait(); }
+    };
+    std::vector<pair_t> pairs(numa_nodes.size());
+
+    for (int i = 0; i < numa_nodes.size(); i++) {
+        arenas[i].initialize(tbb::task_arena::constraints(numa_nodes[i]), 0);
+        pairs.emplace_back(&arena[i], &group[i]);
+    }
+
+    ex::round_robin_policy<pair_t> rr{ pairs };
+
+    std::vector<wait_type> submissions;
+
+    for (auto i : numa_nodes)
+        w.emplace_back(
+            onedpl::dpl::experimental::submit(rr, [](pair_t p) {
+                p.first->enqueue(
+                    p.second->defer([] { 
+                        tbb::parallel_for( /*... */);
+                    })
+                );
+                return wait_type{ *p.second };
+            });
+        );
+
+    for (auto& s : submissions)
+        s.wait(); // not waiting in correct arena though...
+```
+
+### Uninstrumented Custom Backend for NUMA Nodes
+
+```cpp
+    using pair_t = std::pair<tbb::task_arena*, tbb::task_group*>
+    ex::round_robin_policy<pair_t> rr{ };
+
+    for (int i = 0; i < 100; ++i) {
+        onedpl::dpl::experimental::submit(rr, [](pair_t p) {
+            p.first->enqueue(
+                p.second->defer([] { 
+                    tbb::parallel_for( /*... */);
+                })
+            );
+            return p;
+        });
+    }
+
+    wait(rr.get_submission_group());
+```
+
+### Instrumented Custom Backend for NUMA Nodes
+
+```cpp
+    using pair_t = std::pair<tbb::task_arena*, tbb::task_group*>
+    ex::auto_tune_policy<pair_t> rr{ };
+
+    for (int i = 0; i < 100; ++i) {
+        onedpl::dpl::experimental::submit(rr, [](pair_t p) {
+            p.first->enqueue(
+                p.second->defer([] { 
+                    tbb::parallel_for( /*... */);
+                })
+            );
+            return p;
+        });
+    }
+
+    wait(rr.get_submission_group());
+```
--- a/rfcs/proposed/dynamic_selection_customization/README.md
+++ b/rfcs/proposed/dynamic_selection_customization/README.md
@@ -32,18 +32,18 @@ type `T` satisfies the *Backend* contract if given,
 
 | Use Case | Description |
 | --------------------- | -------- |
-| Default Backend | User only provides universe, no custom backend at all. |
-| Uninstrumented Backend | User defines small part of backend: `wait_type`, `submission_group`. Better waiting support. |
-| Instrumented Backend | User provides instrumentation for reporting |
+| Default Backend | No custom backend at all. |
+| Uninstrumented Backend | Backend overrides a small part of default backend: `wait_type`, `submission_group`, for better waiting support. |
+| Instrumented Backend | Backend includes instrumentation for reporting |
 
-We remind readers of this proposal that the requirements of *Submission* and *Selection* objects are provided in
+The requirements of *Submission* and *Selection* objects are provided in
 [the current design description](https://github.com/uxlfoundation/oneDPL/tree/main/rfcs/experimental/dynamic_selection).
 
 ### Default Backend
 
 For policies that require no reporting of `execution_info`, it should be simple to use a new resource type,
-perhaps even without writing a custom backend. For example, what if we want to round-robin through pointers to
-queues instead of queues. 
+perhaps even without writing a custom backend. Below is an example that uses round-robin to rotate through
+through pointers to queues.
 
 ```cpp
     sycl::queue q1, q2;
@@ -65,7 +65,7 @@ queues instead of queues.
 
 #### submit
 
-The main role that `submit` plays is to add instrumentation around the call to `f`.
+The main role that `submit` plays in a backend is to add instrumentation around the call to `f`.
 Experience has shown us that most backends perform four basic steps in their implementation
 of the `submit` function:
 
@@ -76,12 +76,14 @@ of the `submit` function:
 
 If a backend writer doesn't care if their backend works with policies that require reporting, then
 they can use a default implementation of submit that simply calls `f` and returns the backends *Submission*
-object. A *Submission* object must support `s.wait()` and `s.unwrap`. A default backend cannot meaningfully
+object. 
+
+A *Submission* object must support `s.wait()` and `s.unwrap`. A default backend cannot meaningfully
 implement `wait` on an arbitrary type, but could easily wrap a type, such as `sycl::queue` or `tbb::task_group`
 that provides a `wait` member function and call that `wait` member function if it finds it, and in other cases
-provide an empty `wrap` function. `s.unwrap` can return whatever was returned by the user's function. Our simple
-example above returns a `sycl::queue` from its function and so a default backend can easily synthesize a meaningful
-*Submission* type from it, since `q.wait()` is valid.
+provide an empty `s.wait` function. `s.unwrap` usually returns what was returned by the user's function `f`.
+Our simple example above returns a `sycl::queue` from its function and so a default backend can easily synthesize
+a meaningful *Submission* type from it, with `s.wait` calling `q.wait()` and `s.unwrap` returning the queue.
 
 #### get_resources
 
@@ -101,11 +103,12 @@ for an arbitrary backend resource, this function will likely need to return a du
 It may be possible that a backend writer wants to do a little bit of work, so that `get_submission_group` works
 properly, and/or that there is a meaningful *Submission* type defined, and/or there is a default set of resources
 defined even when called with no explicit universe. But, they may not want to go through the effort of
-implementing the reporting mechanism. In that case, we should define a way to customize the default
-backend for the resource type in a way to opt-in to only those parts that should be customized.
+implementing the reporting mechanisms needed to work with all policies. In that case, we should define a way
+to customize the default backend for the resource type while falling back to default behaviors for functions
+that are unimportant to the backend writer.
 
-For example, we may round-robin through pointers to queues, have a default set of resources
-and/or wait on the submission group:
+For example, a backend that defines `get_resources`, `get_submission_group` and appropriate *Submission* type,
+results in added functionality:
 
 ```cpp
     ex::round_robin_policy<sycl::queue*> p;
@@ -146,7 +149,8 @@ functions `instrument_before` and `instrument_after` that can be used to opt-in 
 in a backend. A backend writer will not need to fully define `submit` but instead only write the
 code to set up and reporting needed for the backend.
 
-For example, we may using auto-tune for pointers to queues:
+For example, we may using auto-tune for pointers to queues if the instrumentation related
+functionality is defined in the backend:
 
 ```cpp
     sycl::queue q1, q2;


rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,499 @@
+# Simplified Customization of Backends and Policies for Dynamic Selection
+
+Dynamic Selection is a Technology Preview feature 
+[documented in the oneDPL Developer Guide](https://www.intel.com/content/www/us/en/docs/onedpl/developer-guide/2022-8/dynamic-selection-api.html)
+and its current design is described by 
+[an experimental RFC](https://github.com/uxlfoundation/oneDPL/tree/main/rfcs/experimental/dynamic_selection).
+When applying Dynamic Selection to use cases, there is often the desire to add a new
+backend or new selection policy. While it is possible to add new policies or backends
+by following the contract described in the current design, the process is non-trivial
+and requires some (unecessarily) verbose code.
```
egfefey, line 10: typo? : unecessarily ->unnecessarily

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,499 @@
...
+of the `submit` function:
+
+1. Do any setup needed for implementing reporting before calling `f`.
+2. The function `f` is called.
+3. Do any setup needed for implementing reporting after calling `f`, perhaps using what was returned by `f`.
+4. The *Submission* object, which typically wraps what is returned by `f`, is constructed and returned.
+
+If a backend writer doesn't care if their backend works with policies that require reporting, then
+they can use a default implementation of submit that simply calls `f` and returns the backends *Submission*
+object.
```
egfefey, line 79: We could add explicitly here that, in this case steps 1 and 3 are not needed. Which is part of the simplification we seek to achieve.

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,499 @@
...
+| Instrumented Backend | Backend includes instrumentation for reporting |
+
+The requirements of *Submission* and *Selection* objects are provided in
+[the current design description](https://github.com/uxlfoundation/oneDPL/tree/main/rfcs/experimental/dynamic_selection).
+
+### Default Backend
+
+For policies that require no reporting of `execution_info`, it should be simple to use a new resource type,
+perhaps even without writing a custom backend. Below is an example that uses round-robin to rotate through
+through pointers to queues.
```
egfefey, line 46: repeated word "through"

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,499 @@
...
+that provides a `wait` member function and call that `wait` member function if it finds it, and in other cases
+provide an empty `s.wait` function. `s.unwrap` usually returns what was returned by the user's function `f`.
+Our simple example above returns a `sycl::queue` from its function and so a default backend can easily synthesize
+a meaningful *Submission* type from it, with `s.wait` calling `q.wait()` and `s.unwrap` returning the queue.
+
+#### get_resources
+
+The `get_resources` function can default to returning the vector that is passed to the policy, and
+then down to the backend's constructor. A policy that is created with a default constructor will
+have no resources in this case.
```
egfefey, line 92: We could maybe add that: it is therefore imperative that a user either passes the list of resources at construction of the policy or call initialize to set the resources.

rfcs/proposed/dynamic_selection_customization/README.md
vossmjp: Fixed in coming re-write.

rfcs/proposed/dynamic_selection_customization/README.md
vossmjp: Will do

rfcs/proposed/dynamic_selection_customization/README.md
vossmjp: Will do.

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,499 @@
...
+
+#### get_submission_group
+
+The `get_submission_group` is not easily implemented in a meaningful way. It must return a type that
+defines a member function `wait`. But since there is no way to know how to wait on all previous submissions
+for an arbitrary backend resource, this function will likely need to return a dummy type or 
+`get_submission_group` be undefined for the default backend.
```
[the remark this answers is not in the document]
vossmjp: Lets consider if there's a meaningful default or not. I'm not sure that there is.

rfcs/proposed/dynamic_selection_customization/README.md
vossmjp: Hopefully addressed in rewrite that I'll post soon.

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+    ex::round_robin_policy<tbb::task_group*> p1{ };
+    ex::round_robin_policy<tbb::task_group*> p2{ deferred_initialization_t };
+```
+#### Default `wait_t<T>`
+
+The `wait_t<T>` can only be deduced at the time a user function is passed to submit since
+it is the type returned by the user's function. There is no way to generally determine a
+`wait_t<T>` from an arbitrary `resource_t<T>` and therefore the default backend cannot
+provide a meaningful `wait_t<T>`.
+
```
danhoeflinger, line 82: Should we propose in this RFC to remove `wait_t<T>` altogether?

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+#### Default `b.submit(s, f, args…)`
+
+The main role that `submit` plays in a backend is to add instrumentation around the call to `f`.
+Experience has shown us that most backends perform four basic steps in their implementation
+of the `submit` function:
+
+1. Do any setup needed for implementing reporting before calling `f`.
+2. The function `f` is called and it return value captured.
+3. Do any setup needed for implementing reporting after calling `f`, perhaps using what was returned by `f`.
+4. The *Submission* object, which typically wraps what is returned by `f`, is constructed and returned.
```
danhoeflinger, line 92: nitpick: 
This seems to go back and forth between present tense and past tense, we should keep the same orientation for all steps

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+
+It is not possible for a default backend to properly instrument execution for an unkown resource type to provide
+reporting of `task_time`, `task_submission` and `task_completion`. And so a default implementation
+cannot provide useful implementations for step 1 and 3. However, it can provide more finer-grained hooks
```
danhoeflinger, line 96:
```suggestion
reporting of `task_time`, `task_submission` and `task_completion`. Therefore a default implementation
cannot provide useful implementations for step 1 and 3. However, it can provide more fine-grained hooks
```

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+that can be overridden to provide steps 1 and 3, without requiring a custom backend to reimplemnted the entire
+four step pattern.
+
+We propose that the default backend provide a `submit` function that implements the four step pattern
+but also calls `instrument_before_impl` and `instrument_after_impl` functions that can be overridden
```
danhoeflinger, line 101:
```suggestion
but also calls `instrument_before_impl` and `instrument_after_impl` functions that can be individually overridden
```

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+            std::forward<Function>(f)(unwrap(s), std::forward<Args>(args)...) // call f
+        );
+    }
+```
+
+#### Default *Submission* type
+
+A *Submission* object must support `s.wait()` and `s.unwrap`. A default backend cannot meaningfully
+implement `wait` on an arbitrary type, but can wrap a type, such as `sycl::queue` or `tbb::task_group`
+that provides a `wait` member function itself and call that `wait` member function if it finds it.
```
danhoeflinger, line 120:
```suggestion
A *Submission* object must support `s.wait()` and `s.unwrap`. A default backend cannot meaningfully
implement `wait` on an arbitrary type, but it can wrap a type, such as `sycl::queue` or `tbb::task_group`
that provides a `wait` member function itself and call that `wait` member function if it exists.
```

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+`s.unwrap` can return what was return by the user's function `f`.
```
danhoeflinger, line 121:
```suggestion
`s.unwrap` can return what was returned by the user's function `f`.
```

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+
+However, a default constructed policy has no information about what set makes sense for
+an arbitrary resource type. And so `get_resources` will return an empty vector.
+
+```cpp
+    ex::round_robin_policy<tbb::task_group*> p1{ };
+
+    // v is empty
+     auto v = p.get_resources();
+```
```
danhoeflinger, line 194: mention deferred initialization.

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+to add instrumentation by backends that need it.
+
+```cpp
+    template <typename SelectionHandle, typename Function, typename... Args>
+    auto submit(SelectionHandle s, Function&& f, Args&&... args) {
+        instrument_before_impl(s); // do insrumentation before calling `f`
+
+        return instrument_after_impl( // do insrumentation before calling `f`, and create Submission type
+            s,
+            std::forward<Function>(f)(unwrap(s), std::forward<Args>(args)...) // call f
```
danhoeflinger, line 111: nitpick:
for explanation, rearrange code to allow function call to appear above `instrument_after_impl`

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+
+#### Default `lazy_report_v<T>`
+
+The default backend, which does not provide any reporting support at all, will
+have `lazy_report_v<T>` as `false`.
+
+## How to customize for specific resource types
+
+We proposed using a CRTP (Curiously Recurring Template Pattern) to define backends since this
+will allow developers to flexibly mix-in functionaility from the base. We decided against free
```
danhoeflinger, line 228:
```suggestion
will allow developers to flexibly mix-in functionality from the base. We decided against free
```

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+{
+  //* override as needed
+};
+```
+
+And the use of this SYCL backend by a policy, such as `round_robin_policy` is implemented by a policy
+as:
+
+```cpp
+template <typename ResourceType = sycl::queue, typename Backend = default_backend<sycl::queue>>
```
danhoeflinger, line 342:
```suggestion
template <typename ResourceType = sycl::queue, typename Backend = default_backend<ResourceType>>
```

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -0,0 +1,436 @@
...
+```
+
+A backend that relies on the defaults but overrides some functionality can then specialize the
+default backend for the type. For example, the existing SYCL backend can be implemented as:
+
+```cpp
+template< >
+class default_backend<sycl::queue> : public backend_base<sycl::queue, default_backend<sycl::queue>>
+{
+  //* override as needed
```
danhoeflinger, line 334: Is this a TODO for the RFC?
Do we want to describe what is actually necessary to define `sycl::queue` backend here?


In the PoC as it stands, it actually its ~30 lines longer implementation than the previous version... can we cut it down more?

Or is there a different simpler example we should use here?

rfcs/proposed/dynamic_selection_customization/README.md
vossmjp: I think so, yes.

rfcs/proposed/dynamic_selection_customization/README.md
vossmjp: I will add different examples.

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -354,83 +397,260 @@ a specialization of teh default backed:
...
+
+
+template <typename ResourceType = sycl::queue, typename Backend = backend_for_resource<ResourceType>::backend_t>
+struct round_robin_policy
+{
+    // policy implementation not shown here
+};
+}
+}
+}
```
danhoeflinger, line 443:
```suggestion
}}} //namespace oneapi::dpl::experimental
```

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -354,83 +397,260 @@ a specialization of teh default backed:
...
+    class default_backend<third_party::custom_resource_2> 
+        : public backend_base<third_party::custom_resource_2, default_backend<third_party::custom_resource_2>> {
+    public:
+        using resource_type = third_party::custom_resource_2;
+        // override _impl functions as desired
     };
-    std::vector<pair_t> pairs(numa_nodes.size());
+}
+}
+}
```
danhoeflinger, line 503:
```suggestion
}}} //namespace oneapi::dpl::experimental
```

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -354,83 +397,260 @@ a specialization of teh default backed:
...
-    std::vector<tbb::task_group> groups(numa_nodes.size());
-    using pair_t = std::pair<tbb::task_arena*, tbb::task_group*>
-    struct wait_type {
-        tbb::task_group &tg_;
-        void wait() { tg_.wait(); }
+    class custom_backend_3 : public oneapi::dpl::experimental::backend_base<custom_resource_3, custom_backend_3> {
+        // override _impl functions as desired
+    };
+
+}
```
danhoeflinger, line 480:
```suggestion
} //namespace third_party
```

rfcs/proposed/dynamic_selection_customization/README.md
```diff
@@ -354,83 +397,260 @@ a specialization of teh default backed:
...
+    }
 
     for (auto& s : submissions)
-        s.wait(); // not waiting in correct arena though...
+        ex::wait(s);
+}
 ```
 
-### Custom Backend for NUMA Nodes but no instrumentation
+In this example, we can see that the user has to manual construct a vector of
```
danhoeflinger, line 567:
```suggestion
In this example, we can see that the user has to manually construct a vector of
```

danhoeflinger (commented):
I think this looks pretty good. Mostly just quite minor changes for now.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -0,0 +1,267 @@
...
+class default_backend : public default_backend_impl<...>
+{
+  public:
+    default_backend(const std::vector<ResourceType>& r, ResourceAdapter adapt = {});
+};
+```
+
+Adapters allow backends to be reused when it is possible to transform a custom resource
+type into a known resource type with an existing backend.
+
```
danhoeflinger: It may be good to describe a brief example here, or point to the section below for the full example.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -0,0 +1,267 @@
...
+    resource_container_t resources_;
+    // Implementation methods that can be overridden
+};
+```
+
+The proposed `default_backend` extends this with adapter support:
+
+```cpp
+template <typename ResourceType, typename ResourceAdapter = oneapi::dpl::identity>
+class default_backend : public default_backend_impl<...>
```
danhoeflinger: I think we need to describe `default_backend_impl` more here.  It is the class that one must *partially* specialize to create a specific backend now (see the sycl backend example).  You fix the `BaseResourceType`, and now partially specialize it for a provided `ResourceType` and `ResourceAdapter` (which result in the `BaseResourceType` when `ResourceType` is passed into `ResourceAdapter`).


`default_backend` is the public interface, but basically just a wrapper to provide creation of the `BaseResourceType`.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -0,0 +1,267 @@
...
+
+```cpp
+template <typename SelectionHandle, typename Function, typename... Args>
+auto submit(SelectionHandle s, Function&& f, Args&&... args) {
+    auto adapted_resource = adapter(unwrap(s));
+    auto result = std::forward<Function>(f)(adapted_resource, std::forward<Args>(args)...);
+    return default_submission{result};
+}
+```
+**Assumptions**: The user function must be callable with the adapted resource type and return a type that can be stored and optionally waited upon.
```
danhoeflinger: "adapted resource type" is a little ambiguous to me whether this means the `ResourceType` (pre adapter call) or `BaseResourceType` (return from adapter call).  In the code right now, the user submitted function gets the `ResourceType` object, and must know what to do with it.  I think this is correct, because they have the option to apply the adapter or not. 
For instance, if using a `queue*` as the resource type and a dereference operator as the adapter with one motivating reason to have the pointer value act as a unique identifier into a memory lookup, the submitted function would need access to the `queue*`, not an instance of a `queue`.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -0,0 +1,267 @@
...
+**Assumptions**: Resources can be copied or moved into a `std::vector` and remain valid throughout the backend's lifetime.
+
+#### `get_submission_group()` Implementation
+Returns a group object that can wait on all resources if they provide a `wait()` method:
+```cpp
+auto get_submission_group() {
+    return default_submission_group{resources_, adapter};
+}
+```
+The `default_submission_group` attempts to call `wait()` on each adapted resource:
```
danhoeflinger: Similar to above, "adapted resource" is could be ambiguous.  I think in this case, as shown in the code, it is calling wait on the `BaseResourceType` object.  
We don't necessarily need to use the explicit type names, but we probably need to define some terminology at the top to distinguish between the base / adapted resource and the provided resource.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -0,0 +1,267 @@
...
+```
+
+### Adapter Usage Patterns
+
+Adapters enable several useful patterns:
+
+1. **Pointer Resources**: Store pointers but work with references/values
+2. **Wrapper Types**: Unwrap resource containers or smart pointers  
+3. **Type Conversion**: Convert between compatible resource types
+4. **Ownership Management**: Separate resource storage from resource usage
```
danhoeflinger: Pair a context (memory space, side information, etc.) with a core resource, but rely on the implementation of the core resource without extra backend implementation.

rfcs/proposed/dynamic_selection_customization/custom_policies.md
```diff
@@ -0,0 +1,330 @@
...
+};
+```
+
+## Conclusion
+
+This proposal presents a simplified approach to policy customization for Dynamic Selection. Key benefits include:
+
+1. **Focused Development**: Policy writers can focus solely on selection strategy implementation
+2. **Automatic Backend Integration**: All resource management and backend operations handled automatically  
+3. **Consistent Interface**: All policies follow the same patterns and provide the same functionality
```
danhoeflinger: "provide the same baseline functionality with their own selection criteria" or something to show that they do have differences.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
vossmjp: Ok, will add more here.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
vossmjp: Yes, the implementation of submit does not match the code. I will fix this and make it clear that the user function still receives the non-adapted resource (of `ResourceType`) and that only the backend-specific logic will need to use `adapter(unwrap(s))` to get to the `BaseResourceType` objects.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
vossmjp: Ok

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -36,8 +36,9 @@ This proposal presents a flexible backend system based on a `backend_base` templ
 ### Key Components
 
 1. **`backend_base<ResourceType, Backend>`**: A proposed base class template that implements the core backend functionality using CRTP.
-2. **`default_backend<ResourceType, ResourceAdapter>`**: A proposed template that provides a complete backend implementation for any resource type, with optional adapter support.
-3. **SYCL specialization**: A proposed specialized implementation for `sycl::queue` resources that handles SYCL-specific event management and profiling.
+2. **`default_backend_impl<typename BaseResourceType, ResourceType, ResourceAdapter>`**: A proposed template that provides a complete backend implementation for any resource type, with optional adapter support.
+3. **`default_backend<ResourceType, ResourceAdapter>`**: A proposed template that determines the `BaseResourceType` from the `ResourceType` and `ResourceAdapter`. A developer creates a partial specialization of `default_backend` to create a specific backend.
```
danhoeflinger: A developer creates a partial specialization of `default_backend_impl` to create a specific backend. (not `default_backend`)

We need the user  to partially specialize the impl, because this is where you can fix the base resource type, and leave free the resource type and adapter type.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -36,8 +36,9 @@ This proposal presents a flexible backend system based on a `backend_base` templ
...
+4. **A SYCL specialization of default_backend**: A specialized implementation for `sycl::queue` resources that handles SYCL-specific event management and profiling. Using an adapter, it is possible to reuse this for other types that can be adapted into a `sycl::queue`, such as a `sycl::queue *` or a struct that contains a `sycl::queue`.
```
danhoeflinger: A SYCL specialization of default_backend_impl

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -73,37 +74,71 @@ class backend_base
...
+    }
+    default_backend(const std::vector<ResourceType>& r, ResourceAdapter adapt = {}) : base_t(r, adapt)
+    {
+    }
 };
 ```
-
-Adapters allow backends to be reused when it is possible to transform a custom resource
-type into a known resource type with an existing backend.
+The `default_backend` class is partially specialized to create a specific backend. Adapters allow backends
```
danhoeflinger: `default_backend_impl`

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -211,14 +253,14 @@ ex::wait(rr.get_submission_group());
 ```
 
 If `ArenaAndGroup` will be used with policies that require instrumentation, then
-a custom backend that overrides `instrument_before` and `instrument_after` will be
-needed.
+a custom backend that provides `instrument_before_impl` and `instrument_after_impl`
+will be needed. This can be done by partially specializing `default_backend`.
```
danhoeflinger: `default_backend_impl`

rfcs/proposed/dynamic_selection_customization/custom_policies.md
vossmjp: Ok, will change.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
vossmjp: Ok, will change.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
vossmjp: Ok, will change.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
vossmjp: Ok, will change.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
vossmjp: Ok, will change.

rfcs/proposed/dynamic_selection_customization/custom_backends.md
```diff
@@ -0,0 +1,309 @@
...
+With sensible defaults, this proposal aims to simplify backend writing to open up Dynamic Selection to more use cases.
+
+## Proposed Design to Enable Easier Customization of Backends
+
+This proposal presents a flexible backend system based on a `backend_base` template class and a `default_backend` template that can be used for most resource types. The design uses CRTP (Curiously Recurring Template Pattern) to allow customization while providing sensible defaults.
+
+### Key Components
+
+1. **`backend_base<ResourceType, Backend>`**: A proposed base class template that implements the core backend functionality using CRTP.
+2. **`default_backend_impl<typename BaseResourceType, ResourceType, ResourceAdapter>`**: A proposed template that provides a complete backend implementation for any resource type, with optional adapter support. A developer creates a partial specialization of `default_backend_impl` to create a specific backend for the `BaseResourceType`. Adapters can be used with `default_backend` to reuse `default_backend_impl` for a `ResourceType` that is adapted to a `BaseResourceType`.
```
danhoeflinger:
```suggestion
2. **`default_backend_impl<BaseResourceType, ResourceType, ResourceAdapter>`**: A proposed template that provides a complete backend implementation for any resource type, with optional adapter support. A developer creates a partial specialization of `default_backend_impl` to create a specific backend for the `BaseResourceType`. Adapters can be used with `default_backend` to reuse `default_backend_impl` for a `ResourceType` that is adapted to a `BaseResourceType`.
```

rfcs/experimental/dynamic_selection/customization/custom_backends.md
```diff
@@ -0,0 +1,309 @@
...
+
+### Adapter Usage Patterns
+
+Adapters enable several useful patterns:
+
+1. **Pointer Resources**: Store pointers but work with references/values
+2. **Wrapper Types**: Unwrap resource containers or smart pointers  
+3. **Type Conversion**: Convert between compatible resource types
+4. **Ownership Management**: Pair a context (memory space, side information, etc.) with a core resource, but rely on the implementation of the core resource without extra backend implementation.
+
```
danhoeflinger, line 300: I think we probably want a testing requirements section. 
Also, a section for open questions if we have them or at least alternatives we have considered, and decided against.  I'm thinking specifically about the alternate solution we explored before arriving at the adapter idea.

rfcs/experimental/dynamic_selection/customization/custom_policies.md
```diff
@@ -0,0 +1,330 @@
...
+            std::uniform_int_distribution<> dis(0, resources_.size() - 1);
+            auto index = dis(gen_);
+            return selection_type{*this, resources_[index]};
+        } else {
+            throw std::logic_error("select called before initialization");
+        }
+    }
+};
+```
+
```
danhoeflinger, line 285: Similar to above, I think we may want a testing section and open questions / alternatives section.

rfcs/experimental/dynamic_selection/README.md
```diff
@@ -106,37 +106,54 @@ The type `T` satisfies *Policy* if given,
 
 - `p` an arbitrary identifier of type `T`
 - `args` an arbitrary parameter pack of types `typename… Args`
-- `s` a selection of a type `selection_t<T>` , which satisfies [Selection](#selection_req_id), and was made by `p`.
-- `f` a function object with signature `wait_t<T> fun(resource_t<T>, Args…);`
+- `f` a function object with signature `wait_t<T> fun(resource_t<T>, Args…);`, where `wait_t<T>` is the wait type for the policy. If the backend defines an explicit `wait_type`, user functions must return that specific type. If the backend does not define an explicit `wait_type`, user functions may return any *waitable-type* (a type with a `wait()` member function that blocks until the submitted work is complete).
```
akukanov: I would move the requirements to the user functions into a note that follows the requirement definition - like a note at the line 123.

rfcs/experimental/dynamic_selection/README.md
```diff
@@ -197,38 +214,81 @@ The type `T` satisfies the *Backend* contract if given,
 - `b` an arbitrary identifier of type `T`
 - `args` an arbitrary parameter pack of types `typename… Args`
 - `s` is of type `S` and satisfies *Selection* and `is_same_v<resource_t<S>, resource_t<T>>` is `true`
-- `f` a function object with signature `wait_t<T> fun(resource_t<T>, Args…);`
+- `f` a function object with signature `backend_traits<T>::wait_type fun(resource_t<T>, Args…)`. If the backend defines an explicit `wait_type`, user functions must return that specific type. If the backend does not define an explicit `wait_type`, user functions may return any *waitable-type* (a type with a `wait()` member function that blocks until the submitted work is complete).
```
akukanov: Similarly here - if you think the note is worth repeating.

rfcs/experimental/dynamic_selection/README.md
```diff
@@ -273,25 +349,27 @@ For example, below is code a function the receives a *Selection* and uses traits
...
 
-A backend may choose not to actively report events and instead lazily report them on demand by defining the
-optional `lazy_report` member function. A backend trait is provided to determine if calls to `lazy_report`
-are required.
+Some execution information (like `task_completion` or `task_time`) cannot be reported immediately when a task is submitted, as the task may still be executing asynchronously. Backends may choose to defer reporting of such events until they are needed by the policy.
+
+A backend that uses lazy reporting should:
+1. Store submission handles/events internally when `submit()` is called
+2. Define a `lazy_report()` member function that checks for completed tasks and reports their execution information
+3. The backend trait `backend_traits<Backend>::lazy_report_v` will be `true` for such backends
```
akukanov: Will the trait be automatically set to `true`? If yes, the last item should probably be moved out of the list.

rfcs/experimental/dynamic_selection/customization/custom_backends.md
```diff
@@ -0,0 +1,316 @@
...
+| `lazy_report_v<T>` | `true` if the backend requires a call to `lazy_report()` to trigger reporting of `execution_info` back to the policy |
+| `b.lazy_report()` | An optional function, only needed if `lazy_report_v<T>` is `true`. Is invoked by the policy before making a selection. |
+
+Currently, these functions and traits (except the `lazy_report` function) must be implemented
+in each backend. The experimental backend for SYCL queues is a bit more than 250 lines of code.
+With sensible defaults, this proposal aims to simplify backend writing to open up Dynamic Selection to more use cases.
+
+## Proposed Design to Enable Easier Customization of Backends
+
+This proposal presents a flexible backend system based on a `backend_base` template class and a `core_resource_backend` template that can be used for most resource types. For resource backends supporting reporting requirements, explicit specialization of the `core_resource_backend` for that resource. This is not possible to be done generically. For simple types serving policies without reporting requirements, use of the generically written `core_resource_backend` may be possible.
```
akukanov:
```suggestion
This proposal presents a flexible backend system based on a `backend_base` template class and a `core_resource_backend` template that can be used for most resource types. For resource backends supporting reporting requirements, explicit specialization of the `core_resource_backend` should exist for that resource. This is not possible to be done generically. For simple types serving policies without reporting requirements, use of the generically written `core_resource_backend` may be possible.
```

rfcs/experimental/dynamic_selection/customization/custom_policies.md
```diff
@@ -0,0 +1,267 @@
...
+The default implementation of these traits depends on types defined in the Policy:
+
+```cpp
+  template <typename Policy>
+  struct policy_traits
+  {
+      using resource_type = typename std::decay_t<Policy>::resource_type;
+  };
+```
```
akukanov, line 90: This default implementation has only one trait and therefore seems incomplete.

rfcs/experimental/dynamic_selection/customization/custom_backends.md
danhoeflinger: applied

rfcs/experimental/dynamic_selection/README.md
danhoeflinger: Good point, relocated below as this is passive not active by the implementer.

rfcs/experimental/dynamic_selection/customization/custom_backends.md
danhoeflinger: applied

rfcs/experimental/dynamic_selection/customization/custom_policies.md
danhoeflinger: good catch, filled in

rfcs/experimental/dynamic_selection/README.md
danhoeflinger: Done, along with a similar note below for scratch space below

rfcs/experimental/dynamic_selection/README.md
danhoeflinger: done

rfcs/experimental/dynamic_selection/customization/custom_backends.md
```diff
@@ -0,0 +1,316 @@
...
+
+## Current Backend Design
+
+As described in [the current design](https://github.com/uxlfoundation/oneDPL/tree/main/rfcs/experimental/dynamic_selection),
+type `T` satisfies the *Backend* contract if given,
+
+- `b` an arbitrary identifier of type `T`
+- `args` an arbitrary parameter pack of types `typename… Args`
+- `s` is of type `S` and satisfies *Selection* and `is_same_v<resource_t<S>, resource_t<T>>` is `true`
+- `f` a function object with signature `/*ret_type*/ fun(resource_t<T>, Args…);`
```
vossmjp: Should "fun" be "f"?

rfcs/experimental/dynamic_selection/customization/custom_backends.md
```diff
@@ -0,0 +1,316 @@
...
+| `lazy_report_v<T>` | `true` if the backend requires a call to `lazy_report()` to trigger reporting of `execution_info` back to the policy |
+| `b.lazy_report()` | An optional function, only needed if `lazy_report_v<T>` is `true`. Is invoked by the policy before making a selection. |
+
+Currently, these functions and traits (except the `lazy_report` function) must be implemented
+in each backend. The experimental backend for SYCL queues is a bit more than 250 lines of code.
+With sensible defaults, this proposal aims to simplify backend writing to open up Dynamic Selection to more use cases.
+
+## Proposed Design to Enable Easier Customization of Backends
+
+This proposal presents a flexible backend system based on a `backend_base` template class and a `core_resource_backend` template that can be used for most resource types. For resource backends supporting reporting requirements, explicit specialization of the `core_resource_backend` should exist for that resource. This is not possible to be done generically. For simple types serving policies without reporting requirements, use of the generically written `core_resource_backend` may be possible.
```
vossmjp:
```suggestion
This proposal presents a flexible backend system that can be used for most resource types. It is based on two template classes: `backend_base` and `core_resource_backend`. For resource backends supporting reporting requirements, explicit specialization of the `core_resource_backend` is necessary since reporting cannot be done generically. For simple types serving policies without reporting requirements, use of the generically written `core_resource_backend` may be possible.
```

rfcs/experimental/dynamic_selection/customization/custom_backends.md
```diff
@@ -0,0 +1,316 @@
...
+
+### Key Components
+
+1. **`backend_base<ResourceType>`**: A proposed base class template that implements core backend functionality to inherit from.
```
vossmjp:
```suggestion
1. **`backend_base<ResourceType>`**: A proposed base class template that implements common backend functionality to inherit from.
```

vossmjp: Since we have a "core_resource_backend" lets not overload the meaning of "core".

rfcs/experimental/dynamic_selection/customization/custom_backends.md
```diff
@@ -0,0 +1,316 @@
...
+
+3. **`default_backend<ResourceType, ResourceAdapter>`**: A wrapper template that **users instantiate** when creating policies. It automatically determines the `CoreResourceType` by applying the `ResourceAdapter` to the `ResourceType`, then inherits from the appropriate `core_resource_backend` specialization.
+4. **A SYCL specialization of core_resource_backend**: A specialized implementation for `sycl::queue` resources that handles SYCL-specific event management and profiling. Using an adapter, it is possible to reuse this for other types that can be adapted into a `sycl::queue`, such as a `sycl::queue *` or a struct that contains a `sycl::queue`.
+
+### Reporting Requirements and Scratch Space Contract
+
+Backends must now explicitly accept a (possibly empty) variadic list of reporting requirements describing the execution information the Policy will need. These reporting requirements are the same `execution_info` tag types used elsewhere in the Dynamic Selection API (for example `execution_info::task_time_t`, `execution_info::task_submission_t`, `execution_info::task_completion_t`).
```
vossmjp:
```suggestion
Backends must accept a (possibly empty) variadic list of reporting requirements describing the execution information the Policy will need. These reporting requirements are the same `execution_info` tag types used elsewhere in the Dynamic Selection API (for example `execution_info::task_time_t`, `execution_info::task_submission_t`, `execution_info::task_completion_t`).
```

vossmjp: "now" and "explicitly" seem unnecessary.

rfcs/experimental/dynamic_selection/customization/custom_backends.md
```diff
@@ -0,0 +1,316 @@
...
+
+- Resource filtering: Some reporting requirements imply properties of the underlying resource or device (for example, timing via `task_time_t` may require device support for profiling tags and queues created with profiling enabled). Backends must examine the provided resources (or query devices when default-initializing resources) and filter out any resources that do not support all requested reporting requirements. Any special resource properties required to implement a reporting requirement must be checked here (for instance, checking `device.has(sycl::aspect::ext_oneapi_queue_profiling_tag)` in addition to creating queues with `sycl::property::queue::enable_profiling()` when `task_time_t` is requested). If after filtering the set of candidate resources there are no resources left that satisfy all requested reporting requirements, the backend must throw a `std::runtime_error` documenting that the requested reporting requirements cannot be satisfied on the available resources.
```
vossmjp: "may require device support for profiling tags and queues created with profiling enabled" this is very SYCL specific. Profiling tags or profiling enabled has no generic meaning. Can you be more generic, for example "may limit to only devices that support profiling".

rfcs/experimental/dynamic_selection/customization/custom_backends.md
```diff
@@ -0,0 +1,316 @@
...
+- Scratch space requirement: Backends need storage space within selection handles to implement instrumentation and fulfill reporting requirements. Backends must provide a nested template struct that allocates whatever per-selection scratch space is necessary for the requested reporting requirements. The required name and form are:
+
+```cpp
+template <typename... ReportingReqs>
+struct scratch_space_t {
+    // members required to implement reporting for ReportingReqs...
+};
+```
+
+When a policy tracks execution information (like task timing or completion), the backend needs to store temporary data with each selection. For example, the SYCL backend's `scratch_space_t<execution_info::task_time_t>` includes an extra `sycl::event` to store the "start" profiling tag needed to measure elapsed time. For policies without reporting requirements, `scratch_space_t<>` should be empty (or inherit from `no_scratch_t<>`), adding no overhead.
```
vossmjp:
```suggestion
When a policy tracks execution information (like task timing or completion), the backend may need to store temporary data with each selection. For example, the SYCL backend's `scratch_space_t<execution_info::task_time_t>` includes an extra `sycl::event` to store the "start" profiling tag needed to measure elapsed time. For policies without reporting requirements, `scratch_space_t<>` should be empty (or inherit from `no_scratch_t<>`), adding no overhead.
```

rfcs/experimental/dynamic_selection/customization/custom_backends.md
```diff
@@ -0,0 +1,316 @@
...
+
+Each resource type has different characteristics for timing, profiling, and performance measurement, making a general instrumentation mechanism impossible.
+
+#### `get_resources()` Implementation
+Returns the vector of resources stored during construction. Resources must be copyable/movable into a `std::vector` and remain valid throughout the backend's lifetime.
+
+#### `get_submission_group()` Implementation
+Returns a group object that waits on all resources. The `default_submission_group` attempts to call `wait()` on the `CoreResourceType` by applying `adapter` to the `ResourceType` object. The `CoreResourceType` must provide a `wait()` method that blocks until all work on that resource is complete. Note that the default implementation waits on each resource, not each submission (works for SYCL queues, oneTBB `task_group` objects, etc.). Adapters can enable types without a `wait()` method; for example, `[](auto pointer){ return *pointer; }` allows `sycl::queue*` to work by adapting to `sycl::queue`.
```
vossmjp, line 214: This is not completely accurate. get_submission_group should wait for all outstanding submissions made via this backend instance. One way to accomplish that is to wait on all resources, such as all queues. But we shouldn't require that it waits on the resources, only that all submissions are done.

rfcs/experimental/dynamic_selection/customization/custom_policies.md
```diff
@@ -0,0 +1,272 @@
...
+}
+```
+Providing `try_select()` results in `try_submit()`, `submit()`, and `submit_and_wait()` functions to be supported
+via generic implementations depending only on the selection logic.
+
+#### Task Reporting Requirements
+
+Policies specify reporting requirements (`task_time_t`, `task_submission_t`, `task_completion_t` from `oneapi::dpl::experimental::execution_info`) via the `policy_base` constructor. These are passed to the backend constructor, which filters devices based on feature availability. Examples: `auto_tune_policy` requires `task_time_t`; `dynamic_load_policy` requires `task_submission_t` and `task_completion_t`.
+
+Selection handles must include a `scratch_space` member of type `backend_traits<Backend>::template selection_scratch_t<reqs...>` to provide backend storage for instrumentation.
```
vossmjp, line 213: Are selection handles described anywhere? For a policy that  needs `task_time_t`, `task_submission_t` or `task_completion_t` reported its not clear how those values are received by the policy via the handle. I suppose the example demonstrates it though.

danhoeflinger: Its a good point, I think its a term that is under described, its been on the edge of "implementation details" where "selection" is actually defined, but with the addition of `selection_scratch_t`, I think we need to define it more clearly.

rfcs/experimental/dynamic_selection/customization/custom_backends.md
vossmjp: Ignore this previous comment, I was reacting to the 1st sentence as if it described the purpose of get_submission_group. But in context, this falls under the default implementation section and given the text at the end of the paragraph, it's ok.

rfcs/experimental/dynamic_selection/customization/custom_backends.md
danhoeflinger: Fixed here and in another place, also fixed ret_type.

rfcs/experimental/dynamic_selection/customization/custom_backends.md
danhoeflinger: taken

rfcs/experimental/dynamic_selection/customization/custom_backends.md
danhoeflinger: fixed this and in other places for backends.

rfcs/experimental/dynamic_selection/customization/custom_backends.md
danhoeflinger: taken

rfcs/experimental/dynamic_selection/customization/custom_backends.md
danhoeflinger: rewritten to be more generic

rfcs/experimental/dynamic_selection/customization/custom_backends.md
danhoeflinger: taken

rfcs/experimental/dynamic_selection/customization/custom_policies.md
danhoeflinger: Added a selection handle section to custom_policies.md and then linked to it from custom_backends.md.

vossmjp (commented):
LGTM. I can't click approve since I'm one of the authors, but I approve!

rfcs/experimental/dynamic_selection/README.md
```diff
@@ -106,37 +106,56 @@ The type `T` satisfies *Policy* if given,
...
+  using backend_t = typename policy_traits<Policy>::backend_type;
+
+  template <typename Policy>
+  using resource_t = typename policy_traits<Policy>::resource_type;
+
+  template <typename Policy>
+  inline constexpr bool has_wait_type_v = policy_traits<Policy>::has_wait_type_v;
+
+  template <typename Policy>
+  using wait_t = typename policy_traits<Policy>::wait_type;
```
akukanov, line 158: At some point we will need to decide on the naming schema for Dynamic Selection API. These names are pretty generic; if we add these to `oneapi::dpl` directly, it might cause confusion. One option will be to use a nested namespace for the API, like we do for kernel templates.

This is just a note, it does not block the RFC - though it could be good to capture it as a feature productization criteria/question.

rfcs/experimental/dynamic_selection/README.md
akukanov, line 1: The overall structure of the page can be improved. Currently topics are somewhat intermixed, and the structure is inconsistent. Take a look:

> Dynamic Device Selection API
> - Introduction
> - Overview of Architecture and Execution Flow
> - Named Requirements
>   - Policy
>     - (unnamed) policy traits
>   - Selection
>   - Submission
> - Provided Concrete Policies
> - Backends
>   - Named Requirements
>   - Resource Adapters
> - Free Functions
>   - (unnamed) backend traits
>   - Policy Member Functions vs Free Functions
>   - Deferred Initialization
> - Execution Info
>   - (unnamed) info traits
>   - Lazy Reporting
> - Exit Criteria

rfcs/experimental/dynamic_selection/README.md
```diff
@@ -106,37 +106,56 @@ The type `T` satisfies *Policy* if given,
...
 
-| *Optional* | Description |
+| *At least one must be well-formed* | Description |
 | --------------------- | ----------- |
-| `p.submit_and_wait(s, f, args…)` | Returns `void`. The function invokes `f` with `s` and `args...` and waits for the `wait_t<T>` it returns to complete. |
+| `p.try_submit(f, args…)` | Returns `std::optional<submission_t<T>>` that satisfies [Submission](#submission_req_id). The function selects a resource and invokes `f` with the selected resource and `args...`. Returns empty `std::optional` if no resource is available for selection. |
 | `p.submit(f, args…)` | Returns `submission_t<T>` that satisfies [Submission](#submission_req_id). The function selects a resource and invokes `f` with the selected resource and `args...`. |
-| `p.submit_and_wait(f, args…)` | Returns `void`. The function selects a resource, invokes `f` and waits for the `wait_t<T>` it returns to complete. |
+| `p.submit_and_wait(f, args…)` | Returns `void`. The function selects a resource, invokes `f` and waits for the job to complete. |
```
akukanov, line 123: The source code for the tables is hard to read. Of course the markdown format is far from perfect for tables, but at least some improvements can be made, such as width alignment for columns at the left, as well as for the heading row and the separator row right under.

```suggestion
| *At least one must be well-formed* | Description |
| ---------------------------------- | ----------- |
| `p.try_submit(f, args…)`           | Returns `std::optional<submission_t<T>>` that satisfies [Submission](#submission_req_id). The function selects a resource and invokes `f` with the selected resource and `args...`. Returns empty `std::optional` if no resource is available for selection. |
| `p.submit(f, args…)`               | Returns `submission_t<T>` that satisfies [Submission](#submission_req_id). The function selects a resource and invokes `f` with the selected resource and `args...`. |
| `p.submit_and_wait(f, args…)`      | Returns `void`. The function selects a resource, invokes `f` and waits for the job to complete. |
```

rfcs/experimental/dynamic_selection/README.md
```diff
@@ -106,37 +106,56 @@ The type `T` satisfies *Policy* if given,
 
 - `p` an arbitrary identifier of type `T`
 - `args` an arbitrary parameter pack of types `typename… Args`
-- `s` a selection of a type `selection_t<T>` , which satisfies [Selection](#selection_req_id), and was made by `p`.
 - `f` a function object with signature `wait_t<T> fun(resource_t<T>, Args…);`
 
+**Note:** If the backend defines an explicit `wait_type`, user functions must return that specific type. If the backend does not define an explicit `wait_type`, user functions may return any *waitable-type* (a type with a `wait()` member function that blocks until the submitted work is complete).
```
akukanov, line 111: Could long lines like this one be avoided, except when necessary (like for the tables)?

rfcs/experimental/dynamic_selection/customization/tests.cpp
akukanov, line 1: Why is this file here, in an RFC directory? If the intent is to test the code samples, how does it run and where are the results? And ideally, the samples would be taken from the file directly (if that's possible with Markdown).

Otherwise, it is probably best to convert this into a regular test or into an example for DS usage.

akukanov (approved):
The comments can be addressed later, so I approve the RFC.

PR #2224, opened 2025-05-05
Convert lambdas in tests into functional objects in tests

SergeyKopienko:
It's required to run tests with the different value categories of execution policy (`l-value`, `r-value`).

We don't touch lambdas in tests if:
- they were declared inside `main()` function;
- they were declared in global scope;
- they were declared inside tests code which doesn't depends from `ExecutionPolicy` template parameter.


SergeyKopienko:
> Actually, I cannot catch a sense of such total changing lambda to functors in the test cases. Yes, in the implementation we have changed the lambdas to the functors for the hetero code, due to it MIGHT lead to an issue in case of "unnamed_lmabda=off" and SYCL compiler cannot treat with lambda instance/type properly. Other words, we are trying don't make situation worse... But a user can try to do it - by passing lambdas. For the algo test cases (a kind of user code) i dont understand why we should change passed lambda by functor. Moreover, by such changing we reduce test coverage for the host policies and valid scenarios with device policy.

In the separate PR we going to extend tests coverage for copied/moved policies.
So we should change all lambdas to functional objects.

test/parallel_api/algorithm/alg.nonmodifying/find_if.pass.cpp
```diff
@@ -54,11 +54,22 @@ struct test_find_if_not
     }
 };
 
+template <typename T, typename Predicate>
+struct NotPred
```
mmichel11: We have a similar utility, `TransformOp`, in `copy_if.pass`. Should we make it a general utility?

test/support/utils.h
```diff
@@ -1061,6 +1061,127 @@ get_pattern_for_test_sizes()
...
+struct IsGreat
+{
+    bool operator()(T x, T y) const
+    {
+        return x < y;
+    }
+};
+
+template <typename T>
+struct IsLess
```
mmichel11, line 1145: Can we just use `std::less<T>` instead of this custom type?

SergeyKopienko: Sometimes in our internal code we checks that predicate is the same like one of standard predicates. And implements some different logic for these cases. So I prefer still continue user predicates in tests as before this PR.

test/parallel_api/algorithm/alg.nonmodifying/find_if.pass.cpp
SergeyKopienko: Fixed, thanks

test/general/sycl_iterator/sycl_iterator_scan.pass.cpp
```diff
@@ -159,6 +177,15 @@ DEFINE_TEST(test_partition)
 {
     DEFINE_TEST_CONSTRUCTOR(test_partition, 2.0f, 0.65f)
 
+    template <typename IteratorValueType>
+    struct IsMultipleOf3And2
```
mmichel11: This struct is defined twice in this test. Should we just define it once outside the test class?

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
```diff
@@ -235,12 +235,22 @@ test_with_buffers(sycl::queue& q, Size n, StabilityTag stability_tag, Compare...
     check_sort(keys.begin(), vals.begin(), origin_keys.begin(), origin_vals.begin(), n, n, StableSortTag{}, compare...);
 }
 
+struct CustomGreat
```
mmichel11, line 238: Can we use `TestUtils::IsGreat` here?

test/support/utils.h
```diff
@@ -1061,6 +1061,139 @@ get_pattern_for_test_sizes()
...
+        return v < value;
+    }
+};
+
+template <typename T>
+struct IsGreat
+{
+    bool operator()(T x, T y) const
+    {
+        return x < y;
```
mmichel11:
```suggestion
        return x > y;
```

test/parallel_api/iterator/permutation_iterator_parallel_find.pass.cpp
```diff
@@ -28,6 +28,38 @@ DEFINE_TEST_PERM_IT(test_find, PermItIndexTag)
         ::std::iota(itBegin, itEnd, initVal);
     }
 
+    template <typename Policy, typename Size>
+    struct TestImlementation
```
mmichel11: nitpick: Should be `TestImplementation` for all of these instances.

test/support/utils.h
```diff
@@ -1061,6 +1061,139 @@ get_pattern_for_test_sizes()
...
+
+    bool
+    operator()(T v) const
+    {
+        return v > value;
+    }
+};
+
+template <typename T>
+struct IsLessThen
```
mmichel11: nitpick: `IsLessThan` and `IsGreatThan` is probably the best name.

SergeyKopienko: Fixed, thanks

test/general/sycl_iterator/sycl_iterator_scan.pass.cpp
SergeyKopienko: Fixed

test/parallel_api/iterator/permutation_iterator_parallel_find.pass.cpp
SergeyKopienko: Fixed

test/parallel_api/algorithm/alg.sorting/sort_by_key_common.h
SergeyKopienko: I think it's not the best option because `StableSortTag` here is just some kind of tag:
```C++
template<typename Policy, typename KeyIt, typename ValIt, typename Size, typename... Compare>
void
call_sort(Policy&& policy, KeyIt keys_begin, ValIt vals_begin, Size n, StableSortTag, Compare... compare)
{
    oneapi::dpl::stable_sort_by_key(policy, keys_begin, keys_begin + n, vals_begin, compare...);
}
```

test/parallel_api/algorithm/alg.nonmodifying/find_if.pass.cpp
SergeyKopienko: Fixed

test/parallel_api/iterator/input_data_sweep.h
```diff
@@ -24,6 +24,29 @@
...
+
+template <typename Size, int __reverses>
+struct get_expected_op
+{
+    Size n;
+
+    template <typename T>
+    auto operator()(T exp) const
+    {
+        (void)n; // avoid unused-lambda-capture, bug https://bugs.llvm.org/show_bug.cgi?id=35450
```
mmichel11: We can probably remove this line. It looks to be related to lambdas.

test/parallel_api/iterator/permutation_iterator_parallel_merge.pass.cpp
```diff
@@ -28,19 +28,88 @@ DEFINE_TEST_PERM_IT(test_merge, PermItIndexTag)
...
     }
 
-    template <typename Policy, typename Iterator1, typename Iterator2, typename Iterator3, typename Size>
-    void
-    operator()(Policy&& exec,
-               Iterator1 first1, [[maybe_unused]] Iterator1 last1,
-               Iterator2 first2, [[maybe_unused]] Iterator2 last2,
-               Iterator3 first3, Iterator3 last3, Size n)
+    template <typename Policy, typename Size, typename Iterator3, typename TPermutationIterator>
+    struct TestImlementationLevel1
```
mmichel11:
```suggestion
    struct TestImplementationLevel1
```
Suggestion to update these cases (there are a few more).

mmichel11 (commented):
I only have minor comments. This PR LGTM in general and tests show no issues. However, I do recommend you get a second opinion due to the breadth of the changes.

test/parallel_api/iterator/input_data_sweep.h
SergeyKopienko: Fixed

test/parallel_api/iterator/permutation_iterator_parallel_merge.pass.cpp
SergeyKopienko: Fixed

mmichel11 (approved):

dmitriy-sobolev (approved):
Looks good to me overall.
I haven’t checked each line in detail, but all the changes appear to align with the purpose of the PR.

PR #2225, opened 2025-05-05
Fix early memory deallocation in algorithm implementations

SergeyKopienko:
In this PR we fix early memory deallocation in algorithm implementations.

# The problem statement.
Typically in oneDPL algorithms we have the code like:
```C++
    __par_backend_hetero::__parallel_partial_sort(
        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
        __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read_write>(__first),
        __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read_write>(__mid),
        __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read_write>(__last), __comp)
        .__deferrable_wait();
```

where `__parallel_partial_sort` return the instance of `__future<sycl::event>`.
So in this case the is no memory management inside and so on.

But in the some cases we have another calls, for example:
```C++
            .__deferrable_wait();
```

where we returns `__future<sycl::event, std::shared_ptr<__result_and_scratch_storage_base>>`
As we may see, some management of temporary objects has place here.

This mean, that after `__deferrable_wait()` call, implemented as
```C++
    void
    __deferrable_wait()
    {
#if !ONEDPL_ALLOW_DEFERRED_WAITING
        wait();
#endif
    }
```
all temporary objects, saved in the `__future` instance, will be destroyed.

But if  `ONEDPL_ALLOW_DEFERRED_WAITING` is defined, the kernel code may still works after exit from `__deferrable_wait()` call and we will work with deallocated memory and so on.

# Solution
Options:
1) change implementation of `__deferrable_wait()` : if `__future` instance contains inside some `std::shared_ptr<__result_and_scratch_storage_base>` data type. then call `wait()` independently from the state if `ONEDPL_ALLOW_DEFERRED_WAITING` macros.
2) call `wait()` in problem cases instead of `__deferrable_wait()` 

I prefer option (1), because we have two implementations of `__parallel_stable_sort` for `oneapi::dpl::__internal::__device_backend_tag` : one of them returns `__future<sycl::event>` and another returns `__future<sycl::event, std::shared_ptr<__result_and_scratch_storage_base>>`. Which implementation will be used depends on the state of `_ONEDPL_USE_RADIX_SORT` macros.

## Proposed implementation of `__deferrable_wait()` :
I renamed `__deferrable_wait()` to `__checked_deferrable_wait()` and implement it by this way:
```C++
void
__checked_deferrable_wait()
{
#if !ONEDPL_ALLOW_DEFERRED_WAITING
    wait();
#else
    if constexpr (sizeof...(_Args) > 0)
    {
        // We should have this wait() call to ensure that the temporary data is not destroyed before the kernel code finished
        wait();
    }
#endif
}
```

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1237,7 +1237,7 @@ __pattern_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _Ite
 
         __par_backend_hetero::__parallel_merge(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
                                                __buf1.all_view(), __buf2.all_view(), __buf3.all_view(), __comp)
-            .__deferrable_wait(); // __future<sycl::event, std::shared_ptr<__result_and_scratch_storage_base>>
+            .wait(); // Calls wait() here because we should extend the life-time of temporary data inside __future instance
     }
     return __d_first + __n;
 }
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1237,7 +1237,7 @@ __pattern_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _Ite
 
         __par_backend_hetero::__parallel_merge(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
                                                __buf1.all_view(), __buf2.all_view(), __buf3.all_view(), __comp)
-            .wait(); // Calls wait() here because we should extend the life-time of temporary data inside __future instance
+            .__deferrable_wait(); // __future<sycl::event, std::shared_ptr<__result_and_scratch_storage_base>>
     }
     return __d_first + __n;
 }
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1237,7 +1237,7 @@ __pattern_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _Ite
 
         __par_backend_hetero::__parallel_merge(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
                                                __buf1.all_view(), __buf2.all_view(), __buf3.all_view(), __comp)
-            .__checked_deferrable_wait(); // __future<sycl::event, std::shared_ptr<__result_and_scratch_storage_base>>
+            .__checked_deferrable_wait();
     }
     return __d_first + __n;
 }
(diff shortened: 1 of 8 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1237,7 +1237,7 @@ __pattern_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _Ite
 
         __par_backend_hetero::__parallel_merge(_BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
                                                __buf1.all_view(), __buf2.all_view(), __buf3.all_view(), __comp)
-            .__deferrable_wait();
+            .__deferrable_wait(); // __future<sycl::event, std::shared_ptr<__result_and_scratch_storage_base>>
```
SergeyKopienko, line 1240: This and the other comments are temporary just for understanding what type in each case we return.
Later I will rollback all temporary comments.

danhoeflinger (commented):
In my opinion, these cases are simply not deferrable because of the keep-alive requirement, so `wait()` should just be called.
I looked through all the deferrable waits and agree that it is this set of 4 that need changes.

If we want to instead have it be "automatic", then I think it would be acceptable to check if there is more than just an `Event` in the future, and simply call `wait()` in that situation.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -799,6 +799,20 @@ class __future : private std::tuple<_Args...>
...
+        if constexpr (__size_of_tuple > 1)
+        {
+            using __temporary_data_t = std::shared_ptr<__result_and_scratch_storage_base>;
+            using __last_element_t = std::decay_t<std::tuple_element_t<__size_of_tuple - 1, decltype(*this)>>;
+            if constexpr (std::is_same_v<__last_element_t, __temporary_data_t>)
+            {
+                // We should have this wait() call to ensure that the temporary data is not destroyed before the kernel code finished
+                wait();
+            }
+        }
```
SergeyKopienko, line 815: Alternative implementation:
```suggestion
        // __result_and_scratch_storage is the last element of the tuple if it exist
        if constexpr (__size_of_tuple > 1)
        {
            using __temporary_data_t = std::shared_ptr<__result_and_scratch_storage_base>;
            using __last_element_t = std::decay_t<std::tuple_element_t<__size_of_tuple - 1, decltype(*this)>>;
            if constexpr (std::is_same_v<__last_element_t, __temporary_data_t>)
            {
                // We should have this wait() call to ensure that the temporary data is not destroyed before the kernel code finished
                wait();
            }
        }
```

SergeyKopienko: One more alternative implementation:

```suggestion
        if constexpr (sizeof...(_Args) > 0)
        {
            // We should have this wait() call to ensure that the temporary data is not destroyed before the kernel code finished
            wait();
        }
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -799,6 +799,12 @@ class __future : private std::tuple<_Args...>
     {
 #if !ONEDPL_ALLOW_DEFERRED_WAITING
         wait();
+#else
```
danhoeflinger, line 802: Optional (but my preference):
If we are going to automatically check, then should we rename this `__checked_deferrable_wait`?

danhoeflinger (commented):
I think its is a reasonable fix.  If we want to apply my suggestions for renaming the function, I feel that is slightly better but not required.
After removing the comments, LGTM.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: Renamed

danhoeflinger (approved):
LGTM, but please wait for green CI and check with @timmiesmith for inclusion into this version before freeze.

timmiesmith:
> LGTM, but please wait for green CI and check with @timmiesmith for inclusion into this version before freeze.

Ok to merge if CI comes back green.

PR #2227, opened 2025-05-06
Remove use of `std::array` in `reduce_by_segment`

mmichel11:
MSVC's `std::array` implementation is not compatible with SYCL kernels in debug builds. Previously, the `icpx` and intel open-source compilers had a workaround for this that enabled `std::array` to work in kernels in this case. However, due to some changes in the MSVC STL implementation, this workaround has been removed, and we cannot safely use `std::array` within kernels.

In `reduce_by_segment`, there is a single use of `std::array` that can be replaced with a standard C++ array.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_by_segment.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_by_segment.h
@@ -48,7 +48,6 @@
 #include <type_traits>
 #include <utility>
 #include <algorithm>
-#include <array>
 
 #include "sycl_defs.h"
 #include "parallel_backend_sycl_utils.h"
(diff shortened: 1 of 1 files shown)


mmichel11:
@timmiesmith Are we okay to commit this for oneDPL 2022.9.0?

SergeyKopienko (approved):
LGTM

SergeyKopienko:
We have one more extra `#include <array>` in `include\oneapi\dpl\pstl\hetero\dpcpp\parallel_backend_sycl_reduce_by_segment.h'
May be better to remove this include too?

SergeyKopienko (approved):
LGTM

danhoeflinger (approved):
LGTM

timmiesmith:
> @timmiesmith Are we okay to commit this for oneDPL 2022.9.0?

Yes, ok to merge.

PR #2228, opened 2025-05-06
[Test] Balanced path tests fix

danhoeflinger:
Fix for unit test of balanced path.
Insufficient temp storage was provided to the set operation, now we have added enough space.  We mistook an issue with windows testing as an issue in MSVC STL, but it really was the result of an OOB write. This removes the broken macro incorrectly made in the previous PR.

### DIFF
--- a/test/general/implementation_details/balanced_path_unit_tests.pass.cpp
+++ b/test/general/implementation_details/balanced_path_unit_tests.pass.cpp
@@ -178,7 +178,7 @@ test_serial_set_op_count_and_write2_large_setA(SetTag set_tag)
     std::vector<int> v2 = {1};
     std::vector<int> v3(v1.size() + v2.size());
 
-    oneapi::dpl::__par_backend_hetero::__temp_data_array<20, int> __temp_data{};
+    oneapi::dpl::__par_backend_hetero::__temp_data_array<15, int> __temp_data{};
     oneapi::dpl::__par_backend_hetero::__get_set_operation<SetTag> __set_op;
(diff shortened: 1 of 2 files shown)


test/general/implementation_details/balanced_path_unit_tests.pass.cpp
```diff
@@ -178,7 +178,7 @@ test_serial_set_op_count_and_write2_large_setA(SetTag set_tag)
     std::vector<int> v2 = {1};
     std::vector<int> v3(v1.size() + v2.size());
 
-    oneapi::dpl::__par_backend_hetero::__temp_data_array<10, int> __temp_data{};
+    oneapi::dpl::__par_backend_hetero::__temp_data_array<20, int> __temp_data{};
```
SergeyKopienko, line 181: Is it exactly '20' ? Why? Can we use something like `v1.size() + v2.size()` ?

test/general/implementation_details/balanced_path_unit_tests.pass.cpp
```diff
@@ -178,7 +178,7 @@ test_serial_set_op_count_and_write2_large_setA(SetTag set_tag)
     std::vector<int> v2 = {1};
     std::vector<int> v3(v1.size() + v2.size());
 
-    oneapi::dpl::__par_backend_hetero::__temp_data_array<10, int> __temp_data{};
+    oneapi::dpl::__par_backend_hetero::__temp_data_array<15, int> __temp_data{};
```
SergeyKopienko, line 181: Probably we can define the amount of items in `v2` array through new compile-time const.

SergeyKopienko (approved):
LGTM

PR #2229, opened 2025-05-06
[Tests] Improve Dynamic Selection tests

danhoeflinger:
Cleanup of some dynamic selection tests.  
Removing `std::array` from tests, (flattening array).
Consolidating redundant code, fixing kernel naming, removing use of single_task in favor of consolidated kernel with 0 iterations.

Seems to fix some issues in windows testing, but looking a little closer.


test/support/test_dynamic_load_utils.h
```diff
@@ -127,9 +127,8 @@ test_submit_and_wait_on_group(UniverseContainer u, ResourceFunction&& f)
 
     constexpr size_t N = 1000; // Number of vectors
     constexpr size_t D = 100;  // Dimension of each vector
-
-    std::array<std::array<int, D>, N> a;
-    std::array<std::array<int, D>, N> b;
+    std::vector<int> a(N * D);
+    std::vector<int> b(N * D);
```
egfefey, line 131: Since we are flattening into a 1D vector, I think there's no need to keep separate N and D variables. We could simply use one variable which would be the product of N and D. (say M = 100000)

test/support/test_dynamic_load_utils.h
```diff
@@ -138,15 +137,15 @@ test_submit_and_wait_on_group(UniverseContainer u, ResourceFunction&& f)
     {
         for (size_t j = 0; j < D; ++j)
         {
-            a[i][j] = distribution(generator);
-            b[i][j] = distribution(generator);
+            a[i * D + j] = distribution(generator);
+            b[i * D + j] = distribution(generator);
```
egfefey, line 141: This initialization would therefore be a single loop over M

test/support/test_dynamic_load_utils.h
```diff
@@ -138,15 +137,15 @@ test_submit_and_wait_on_group(UniverseContainer u, ResourceFunction&& f)
...
     }
 
-    std::array<std::array<int, N>, N> resultMatrix;
-    sycl::buffer<std::array<int, D>, 1> bufferA(a.data(), sycl::range<1>(N));
-    sycl::buffer<std::array<int, D>, 1> bufferB(b.data(), sycl::range<1>(N));
-    sycl::buffer<std::array<int, N>, 1> bufferResultMatrix(resultMatrix.data(), sycl::range<1>(N));
+    std::vector<int> resultMatrix(N * N);
+    sycl::buffer<int, 1> bufferA(a.data(), sycl::range<1>(N * D));
+    sycl::buffer<int, 1> bufferB(b.data(), sycl::range<1>(N * D));
+    sycl::buffer<int, 1> bufferResultMatrix(resultMatrix.data(), sycl::range<1>(N * N));
```
egfefey, line 148: bufferResultMatrix would have length of M as well.

test/support/test_dynamic_load_utils.h
```diff
@@ -174,9 +173,9 @@ test_submit_and_wait_on_group(UniverseContainer u, ResourceFunction&& f)
                                     int dotProduct = 0;
                                     for (size_t i = 0; i < D; ++i)
                                     {
-                                        dotProduct += accessorA[item][i] * accessorB[item][i];
+                                        dotProduct += accessorA[item* D + i] * accessorB[item* D + i];
                                     }
-                                    accessorResultMatrix[item][j] = dotProduct;
+                                    accessorResultMatrix[item * N + j] = dotProduct;
```
egfefey, line 178: the dot product will also be over M

test/support/test_dynamic_load_utils.h
```diff
@@ -220,9 +219,9 @@ test_submit_and_wait_on_group(UniverseContainer u, ResourceFunction&& f)
                                         int dotProduct = 0;
                                         for (size_t i = 0; i < D; ++i)
                                         {
-                                            dotProduct += accessorA[item][i] * accessorB[item][i];
+                                            dotProduct += accessorA[item * D + i] * accessorB[item * D + i];
                                         }
-                                        accessorResultMatrix[item][j] = dotProduct;
+                                        accessorResultMatrix[item * N + j] = dotProduct;
```
egfefey, line 224: Likewise

test/support/test_dynamic_load_utils.h
danhoeflinger: Fair point.  I was trying to preserve the original meaning, but I'm fine with further reducing the complexity. I'll make the changes.

danhoeflinger: On the other hand, its kind of arbitrary.  Effectively all we care about here is having some work to do. We can have a 2D matrix multiply semantically which is just represented with 1D arrays.

My intention was just to flatten the memory representation, not the operation.  Unless there is a compelling reason to also change the work, I'd be in favor of keeping it as is.  I can add some comments describing what the intention is if that is helpful.

danhoeflinger: Alright... I've kept the 1D flattened memory, but changed things in 2 ways:
1) I've adjusted the code to *actually* do a matrix multiplication instead of an odd repeated dot product.  
2) I've used 2D buffer accessor to get a nicer syntax for the matrix multiplication in the kernels

What do you think?  I'm rerunning the internal CI to see if this still resolves the issues on windows, but locally it seems to.

test/support/test_dynamic_load_utils.h
```diff
@@ -168,15 +171,15 @@ test_submit_and_wait_on_group(UniverseContainer u, ResourceFunction&& f)
...
                                 {
                                     int dotProduct = 0;
-                                    for (size_t i = 0; i < D; ++i)
+                                    for (size_t j = 0; j < J; ++j)
                                     {
-                                        dotProduct += accessorA[item][i] * accessorB[item][i];
+                                        dotProduct += accessorA[j][item] * accessorB[k][j];
                                     }
-                                    accessorResultMatrix[item][j] = dotProduct;
+                                    accessorResultMatrix[k][item] = dotProduct;
```
egfefey, line 182: The new format is fine. However with a(I*J), b(J*K) and result(I*K), my expectation was that the dotProduct will be accessorA[i][j] * accessorB[j][k] and the result will be accessorResultMatrix[i][k]. (with i=item). I might be missing something.

test/support/test_dynamic_load_utils.h
egfefey: I like the new format. We just need to check that the indices are in the right order for the matrix multiplication. I added a comment there.

test/support/test_dynamic_load_utils.h
danhoeflinger: Well, they are kind of equivalent as long as we are consistent between allocations, loop bounds, accessor ranges and indices. I believe I've done that, but the spec is not the most clear about multi-dimensional accessors.
With C style arrays / accessors you generally go from least contiguous dimension to most contiguous dimension: 
`Arr[z][y][x]`.  That is the reason I used this axis ordering.

egfefey (approved):

PR #2230, opened 2025-05-06
Bugfix fix naming of ADL overload indirectly device accessible

danhoeflinger:
Fixing bug, ADL name which did not match [specification ](https://github.com/uxlfoundation/oneAPI-spec/pull/620/files).

Also adjusting type of function call to not include lvalue reference.  There is no reason to do that, these are never actually called.


mmichel11 (dismissed):
LGTM. I ensured there are no more instances of `is_onedpl_indirectly_device_accessible_iterator`

timmiesmith (approved):
LGTM

mmichel11 (approved):

PR #2231, opened 2025-05-07
[Bugfix] Forwarding then use bug 

danhoeflinger:
It is not defined the order of evaluation of function arguments, so forwarding the ranges in the same line as calling `size()` on them is a a possible forward then use bug.

This PR also updates the set_a_write function to call the updated signature.

Credit to @mmichel11 for finding the issue / fix


mmichel11 (approved):
LGTM assuming internal CI passes. I double checked that these are all the occurrences.

timmiesmith (approved):
LGTM.

PR #458, opened 2021-12-21
[pstl][async] Future type refactoring  - a kind of generic one type w/o dynamic memory allocation. 

MikeDvorskiy:
1. The main changes - future type refactoring - a kind of generic one type w/o dynamic memory allocation
2. A side change - potential performance optimization for algorithms based on reduction and scan.
Usage of one-element sycl buffer for returned value from a device to host (reducing copy overhead).
3. A side change - applying the last reduction operation with an "init" value was moved to a device.

PR #2232, opened 2025-05-07
Declare return type in the function `__future::__wait_and_get_value(...)`

SergeyKopienko:
In this PR we declare return type in the function `__future::__wait_and_get_value(...)` instead of `auto`.

The goal - to improve readability.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -817,8 +817,8 @@ class __future : private std::tuple<_Args...>
             auto& __val = std::get<0>(*this);
             return __wait_and_get_value(__val);
         }
-        else
-            wait();
+
+        wait();
     }
 
     //The internal API. There are cases where the implementation specifies return value  "higher" than SYCL backend,
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -817,8 +817,8 @@ class __future : private std::tuple<_Args...>
             auto& __val = std::get<0>(*this);
             return __wait_and_get_value(__val);
         }
-
-        wait();
+        else
+            wait();
     }
 
     //The internal API. There are cases where the implementation specifies return value  "higher" than SYCL backend,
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -817,14 +817,14 @@ class __future : private std::tuple<_Args...>
             auto& __val = std::get<0>(*this);
             return __wait_and_get_value(__val);
         }
-        else
-            wait();
+
+        wait();
```
danhoeflinger, line 821: Is there some value in this `else`, since it is attached to a `constexpr` conditional? 

Perhaps its eliminated as dead code either way, but I tend to prefer to keep these `else` blocks anyway when we are not battling very large number of indents anyway for clarity, but here we can help the compiler rather than relying on dead code elimination I think.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -746,7 +746,7 @@ class __future : private std::tuple<_Args...>
     }
 
     template <typename _T, std::size_t _NResults>
-    constexpr auto
+    constexpr _T
```
danhoeflinger, line 749: ... I know this is not new code in this PR, but why is this `constexpr`?
It doesnt seem like `__wait_and_get_value` for `__result_and_scratch_storage` can ever be constexpr.  Can we just remove that?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
SergeyKopienko: this change has been removed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h

danhoeflinger:
It seems the `constexpr` was added here #458 .  My best guess is it was added for the overload which passes through a simple value, rather than something more complicated.  I don't know if there is a real reason why it was added for any other overload.  I think that was then propagated through other changes, etc.  It could be useful to see if @MikeDvorskiy  has any other recollection about this.

danhoeflinger (approved):
Despite my previous comment about the origins of the `constexpr` on some of these functions, I think that none of the ones removed here have any possibility of actually being run at compile time because of what they then call.  
Therefore, I think I'm OK approving this, LGTM.

PR #2234, opened 2025-05-07
Fix nanorange with gcc 15

dmitriy-sobolev:
When compiling with g++ (GCC) 15.1.1 20250425 (available in Fedora 42), there are the following issues:

```c++
nanorange.hpp:8009:80: error: ‘const class __nanorange::nano::ranges::counted_iterator_::counted_iterator<I>’ has no member named ‘current_’ [-Wtemplate-body]
// 3 more locations

nanorange.hpp:17045:42: error: ‘const struct __nanorange::nano::ranges::iota_view<W, Bound>::iterator’ has no member named ‘value’; did you mean ‘value_’?
```

Minimal reproducer: https://godbolt.org/z/zGM6zPbb4
```c++
// error: 'const struct S<T>' has no member named 'x' [-Wtemplate-body]
template <typename T>
struct S
{
    friend void f(const S& s) noexcept(noexcept(s.x)) {}
    T x;
};
int main() {}
```

### DIFF
--- a/include/oneapi/dpl/pstl/ranges/nanorange.hpp
+++ b/include/oneapi/dpl/pstl/ranges/nanorange.hpp
@@ -8006,22 +8006,16 @@ class counted_iterator
     template <typename II = I, ::std::enable_if_t<input_iterator<II>, int> = 0>
 #                    endif
     friend constexpr iter_rvalue_reference_t<I>
-#                   ifndef _MSC_VER
-    // GCC 15 and later cannot find i.current_ outside of immediate context
-    iter_move(const counted_iterator<II>& i)
-#                   else
-    iter_move(const counted_iterator& i)
-#                   endif
-        noexcept(noexcept(ranges::iter_move(i.current_)))
+    iter_move(const counted_iterator& i) noexcept(noexcept(ranges::iter_move(i.current_)))
     {
         return ranges::iter_move(i.current_);
     }
 
-    template <typename I2, typename II = I>
+    template <typename I2>
     friend constexpr auto
-    iter_swap(const counted_iterator<II>& x,
+    iter_swap(const counted_iterator<I>& x,
               const counted_iterator<I2>& y) noexcept(noexcept(ranges::iter_swap(x.current_, y.current_)))
-        -> ::std::enable_if_t<indirectly_swappable<I2, II>>
+        -> ::std::enable_if_t<indirectly_swappable<I2, I>>
     {
         ranges::iter_swap(x.current_, y.current_);
     }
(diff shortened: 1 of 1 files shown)


danhoeflinger (commented):
These changes look OK to me.  I think these wont harm anything while we workaround the issue in g++15.
(and the one issue in nanorange)

include/oneapi/dpl/pstl/ranges/nanorange.hpp
```diff
@@ -8006,16 +8006,22 @@ class counted_iterator
...
 #                    endif
     friend constexpr iter_rvalue_reference_t<I>
-    iter_move(const counted_iterator& i) noexcept(noexcept(ranges::iter_move(i.current_)))
+#                   ifndef _MSC_VER
+    // GCC 15 and later cannot find i.current_ outside of immediate context
+    iter_move(const counted_iterator<II>& i)
+#                   else
+    iter_move(const counted_iterator& i)
+#                   endif
+        noexcept(noexcept(ranges::iter_move(i.current_)))
```
danhoeflinger, line 8015: This is super confusing to read.  
I would recommend redundantly providing the return type and combining the two `#ifndef` blocks.

danhoeflinger:
As mentioned offline, there are a number of other potential issues in nanorange for this same topic (if we ever touch that code, which I'm not sure we do).
The question is if we should preemptively work around such an issue, or leave it lurking...
The ones I've found are:
`filter_view::iterator` @L16629
`join_view::iterator` @L17604
`split_view::inner_iterator` @L18405
`transform_view::iterator` @L19177

Some have multiple functions with issues.

include/oneapi/dpl/pstl/ranges/nanorange.hpp
dmitriy-sobolev: Done

dmitriy-sobolev:
> there are a number of other potential issues in nanorange for this same topic (if we ever touch that code, which I'm not sure we do)....

I've investigated the issue more (see the updated description with the minimal reproducer), and it seems the referenced cases are not prone to it:

1) `filter_view::iterator`
Declaration of the suspected function:
https://github.com/uxlfoundation/oneDPL/blob/59fa2a52d4add44c87dbcb32448c64fbcdcfe67f/include/oneapi/dpl/pstl/ranges/nanorange.hpp#L16636
Accessed data member is declared earlier:
https://github.com/uxlfoundation/oneDPL/blob/59fa2a52d4add44c87dbcb32448c64fbcdcfe67f/include/oneapi/dpl/pstl/ranges/nanorange.hpp#L16538

2) `join_view::iterator`:
Declaration of the suspected function:
https://github.com/uxlfoundation/oneDPL/blob/59fa2a52d4add44c87dbcb32448c64fbcdcfe67f/include/oneapi/dpl/pstl/ranges/nanorange.hpp#L17604
Accessed data member is declared earlier:
https://github.com/uxlfoundation/oneDPL/blob/59fa2a52d4add44c87dbcb32448c64fbcdcfe67f/include/oneapi/dpl/pstl/ranges/nanorange.hpp#L17441

And etc...

danhoeflinger (approved):
LGTM.  
The new reproducer is not quite the same thing, but probably a similar issue, changing it to `noexcept(noexcept(s.x))` rather than a trailing auto definition does not exhibit the problem, unless you make it a hidden friend.  In any case, both seem to be fixed by moving the member field definition above the usage.

timmiesmith:
Should this be pulled into the 2022.9.0 release? It appears to be a low risk change, but the issue is limited to the newer version of GCC 15.

dmitriy-sobolev:
> The new reproducer is not quite the same thing, but probably a similar issue, changing it to `noexcept(noexcept(s.x))` rather than a trailing auto definition does not exhibit the problem, unless you make it a hidden friend. In any case, both seem to be fixed by moving the member field definition above the usage.

You are right. I've updated the reproducer with noexcept construction. Yes, the fix is the same.

dmitriy-sobolev:
> Should this be pulled into the 2022.9.0 release? It appears to be a low risk change, but the issue is limited to the newer version of GCC 15.

@timmiesmith, yes let's include it. It will allow using oneDPL with a default g++ in Fedora 42.

timmiesmith:
> > > Should this be pulled into the 2022.9.0 release? It appears to be a low risk change, but the issue is limited to the newer version of GCC 15.
> > 
> > 
> > @timmiesmith, yes let's include it unless it is too late. It will allow using oneDPL with a default g++ in Fedora 42.
> 
> Ok. There are a couple of commits on main between the merge of this PR and the root of the release branch, so I will cherry-pick this on to the release branch.

#2235 submitted for review.

PR #2235, opened 2025-05-08
Fix nanorange with gcc 15 (#2234)

timmiesmith:
Cherry picking the fix in nanorange for GCC 15 to the release branch.


danhoeflinger (approved):
LGTM

PR #2236, opened 2025-05-08
Update the feature test macro for parallel range algorithms

akukanov:
This change includes documentation, but most importantly it updates the headers to set the new value (`202505L`) to the feature test macro for parallel range algorithms.

It should be cherry-picked to 2022.9.


timmiesmith (approved):
LGTM. Thank you.

dcbenito:
This was already merged, but the changes LGTM.

PR #2239, opened 2025-05-09
Log exceptions from `get_test_queue()` into `std::cerr`

SergeyKopienko:
Sometimes due incorrect environment setting we have throwed exceptions in `oneDPL` tests.
They throws from `get_test_queue()` function, when we get `sycl::queue` for test.

In this PR we added logging of these exceptions from `get_test_queue()` into `std::cerr` and then throw it up.


dmitriy-sobolev (approved):
LGTM

MikeDvorskiy (approved):
LGTM

PR #2240, opened 2025-05-09
[oneDPL][ranges] usage subrange view instead of take and drop views

MikeDvorskiy:
[oneDPL][ranges] usage subrange view instead of take and drop views.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
@@ -533,8 +533,8 @@ __pattern_search_n(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _
 template <typename _BackendTag, typename _ExecutionPolicy, typename _Range, typename _BinaryPredicate,
           typename _OrFirstTag>
 oneapi::dpl::__internal::__difference_t<_Range>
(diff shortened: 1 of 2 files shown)


dmitriy-sobolev (commented):
That change looks good to me.

For the more context, It fixes issues with libstdc++10, which has exceptions inside drop view implementation preventing its use inside kernels. Could you add that motivation into the description of the PR?

There some questions I'd like to ask before approving:
 1) My understanding is that it does not affect the performance. Is that correct?
 2) The PR adds `dpl::experimental::ranges::begin` and `size` into the public interface. Should it be described in the documentation?

MikeDvorskiy:
> * My understanding is that it does not affect the performance. Is that correct?
> * The PR adds `dpl::experimental::ranges::begin` and `size` into the public interface. Should it be described in the documentation?

1. correct
2. A good question. Yes, the change introduces `begin`, `size ` and `subrange` in `dpl::experimental::ranges` namespace...
@akukanov, what do you think what is better - to update the documentation (the range experimental topic) or to use  internal names and don't add these public names to `dpl::experimental::ranges`, because it is just "experimental" and we would not like to develop "experimental" direction for ranges API?

dmitriy-sobolev (approved):
LGTM

dmitriy-sobolev (approved):

danhoeflinger (approved):
LGTM

PR #2241, opened 2025-05-09
Resolving g++ warnings with `is_onedpl_indirectly_device_accessible`

danhoeflinger:
g++ has a warning: "friend declaration declares a non-template function [-Wnon-template-friend]"

This is to flag a common error [(as described here)](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Making_New_Friends) where one declares but does not define a non-template `friend` function in a class with the intention of having a one-to-one mapping of the class instance and the friend.  However, without a forward declaration of a template function, and some other adjustments this is not possible.

The warning in our case is a false one as far as I can understand.  Our intention is to never define a body for this function declaration, since it is only used to communicate a return type, so we have no such problem .  

However, the warning is of course undesirable, and its best to not have what looks like a common error in our code.  To workaround the warning and perception of this error, lets include a minimal function body `{ return {};}` which successfully defines the function only for the instances of the template class which are instantiated.  

This removes the warnings and perceptions of error.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h
@@ -125,7 +125,8 @@ struct sycl_iterator
 
     // While sycl_iterator cannot be "passed directly" because it is not device_copyable or a random access iterator,
     // it does represent indirectly device accessible data.
-    friend std::true_type is_onedpl_indirectly_device_accessible(sycl_iterator)
+    friend std::true_type
+    is_onedpl_indirectly_device_accessible(sycl_iterator)
     {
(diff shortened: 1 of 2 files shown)


dmitriy-sobolev (approved):
We can declare these functions only if a device backend is available (hence rule out g++).
However, any compiler can introduce these warnings, and it does not harm, so it looks good to me.

timmiesmith (approved):
LGTM. Thank you.

danhoeflinger:
> We can declare these functions only if a device backend is available (hence rule out g++). However, any compiler can introduce these warnings following g++, and it does not harm, so it looks good to me.

Right, I think this is a good change both because of the warning, and also because it avoids the (mistaken) perception of an error.

akukanov (approved):

PR #2247, opened 2025-05-12
[oneDPL][hetero] adjacent_find pattern perf potential fix (based on refactoring paralle_find_or hetero pattern)

MikeDvorskiy:
[oneDPL][hetero] adjacent_find pattern perf potential fix (based on refactoring paralle_find_or hetero pattern).

1) Used directly data access (ranges access for several middle layer algorithm patterns) instead of "indirectly access via zip_iterator and _zip_view). 
2) Reduced number of versions of paralle_find_(or ) hetero patterns, because to have just one is absolutely enough.  
3) Extended test coverage for "equal" algorithm
4) At the same time  - a fix for `__pattern_equal` iterator based pattern (trivial sequences case).

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -544,58 +544,37 @@ __pattern_minmax_element(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _
 // adjacent_find
 //------------------------------------------------------------------------
 
-template <typename _BackendTag, typename _ExecutionPolicy, typename _Iterator, typename _BinaryPredicate>
+template <typename _BackendTag, typename _ExecutionPolicy, typename _Iterator, typename _BinaryPredicate, typename _OrFirstTag>
 _Iterator
 __pattern_adjacent_find(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator __first, _Iterator __last,
-                        _BinaryPredicate __predicate, oneapi::dpl::__internal::__or_semantic)
+                        _BinaryPredicate __pred, _OrFirstTag __is_or_semantic)
 {
-    if (__last - __first < 2)
+    const auto __n = __last - __first;
+    if (__n < 2)
         return __last;
 
-    using _Predicate = oneapi::dpl::unseq_backend::single_match_pred<adjacent_find_fn<_BinaryPredicate>>;
+    using _Predicate = oneapi::dpl::unseq_backend::single_match_pred<_BinaryPredicate>;
 
-    auto __keep1 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
-    auto __buf1 = __keep1(__first, __last - 1);
-    auto __keep2 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
-    auto __buf2 = __keep2(__first + 1, __last);
+    auto __keep = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
+    auto __buf = __keep1(__first, __last); //a scope lifetime of this instance should be.
+
+    auto __view = __buf.all_view();
+    auto __view1 = oneapi::dpl::__ranges::take_view_simple(__view, __view.size() - 1);
+    auto __view2 = oneapi::dpl::__ranges::drop_view_simple(__view, 1);
+
+    using __size_calc = oneapi::dpl::__par_backend_hetero::__min_size_calc;
 
     // TODO: in case of conflicting names
     // __par_backend_hetero::make_wrapped_policy<__par_backend_hetero::__or_policy_wrapper>()
-    bool result = __par_backend_hetero::__parallel_find_or(
-        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
-        _Predicate{adjacent_find_fn<_BinaryPredicate>{__predicate}}, __par_backend_hetero::__parallel_or_tag{},
-        oneapi::dpl::__ranges::make_zip_view(__buf1.all_view(), __buf2.all_view()));
+    auto result = __par_backend_hetero::__parallel_find_or(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
+        _Predicate{__pred}, __is_or_semantic, __size_calc{}, __view1, __view2);
 
     // inverted conditional because of
     // reorder_predicate in glue_algorithm_impl.h
-    return result ? __first : __last;
-}
-
-template <typename _BackendTag, typename _ExecutionPolicy, typename _Iterator, typename _BinaryPredicate>
-_Iterator
-__pattern_adjacent_find(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator __first, _Iterator __last,
-                        _BinaryPredicate __predicate, oneapi::dpl::__internal::__first_semantic)
-{
-    if (__last - __first < 2)
-        return __last;
-
-    using _Predicate = oneapi::dpl::unseq_backend::single_match_pred<adjacent_find_fn<_BinaryPredicate>>;
-
-    auto __result = __par_backend_hetero::__parallel_find(
-        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
-        __par_backend_hetero::zip(
-            __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read>(__first),
-            __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read>(__first + 1)),
-        __par_backend_hetero::zip(
-            __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read>(__last - 1),
-            __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read>(__last)),
-        _Predicate{adjacent_find_fn<_BinaryPredicate>{__predicate}}, ::std::true_type{});
-
-    auto __zip_at_first = __par_backend_hetero::zip(
-        __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read>(__first),
-        __par_backend_hetero::make_iter_mode<__par_backend_hetero::access_mode::read>(__first + 1));
-    _Iterator __result_iterator = __first + (__result - __zip_at_first);
-    return (__result_iterator == __last - 1) ? __last : __result_iterator;
+    if constexpr (__is_or_semantic())
+        return result ? __first : __last;
+    else
+        return result == __n - 1 ? __n : result;
 }
 
 //------------------------------------------------------------------------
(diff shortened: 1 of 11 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -544,58 +544,40 @@ __pattern_minmax_element(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _
...
+
+    auto __view = __buf.all_view();
+    auto __view1 = oneapi::dpl::__ranges::take_view_simple(__view, __view.size() - 1);
+    auto __view2 = oneapi::dpl::__ranges::drop_view_simple(__view, 1);
 
-    auto __keep1 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
-    auto __buf1 = __keep1(__first, __last - 1);
-    auto __keep2 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator>();
-    auto __buf2 = __keep2(__first + 1, __last);
+    using __size_calc = oneapi::dpl::__par_backend_hetero::__min_size_calc;
```
dmitriy-sobolev, line 565: `__min_size_calc` is defined at a backend level (`parallel_backend_sycl`). It does not look like a backend specific functionality - let's move it here to the middle-layer, `algorithm_impl_hetero.h`.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -544,58 +544,40 @@ __pattern_minmax_element(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _
...
+    using _TagType = std::conditional_t<__is_or_semantic(), oneapi::dpl::__par_backend_hetero::__parallel_or_tag,
+                                        oneapi::dpl::__par_backend_hetero::__parallel_find_forward_tag<_IndexType>>;
 
     // TODO: in case of conflicting names
     // __par_backend_hetero::make_wrapped_policy<__par_backend_hetero::__or_policy_wrapper>()
-    bool result = __par_backend_hetero::__parallel_find_or(
-        _BackendTag{}, ::std::forward<_ExecutionPolicy>(__exec),
-        _Predicate{adjacent_find_fn<_BinaryPredicate>{__predicate}}, __par_backend_hetero::__parallel_or_tag{},
-        oneapi::dpl::__ranges::make_zip_view(__buf1.all_view(), __buf2.all_view()));
+    auto result = __par_backend_hetero::__parallel_find_or(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
```
dmitriy-sobolev, line 572: Could you apply clang-format? There are dozens of places not aligned with it. Its suggestions look good to me.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -664,19 +645,19 @@ __pattern_equal(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterator1
...
 
-    auto __keep1 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator1>();
-    auto __buf1 = __keep1(__first1, __last1);
-    auto __keep2 = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator2>();
-    auto __buf2 = __keep2(__first2, __last2);
+    auto __keep = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read>();
+    auto __buf1 = __keep(__first1, __last1);
+    auto __buf2 = __keep(__first2, __last2);
+
+    using size_calc = oneapi::dpl::__par_backend_hetero::__min_size_calc;
```
dmitriy-sobolev, line 654: Previously, only `__parallel_find_or_min_size` used `__min_size_calc`, which in turn was used only by `__pattern_mismatch`. So, everything except for mismatch algorithm relied on `__first_size_calc`.

Is that change intentional?

dmitriy-sobolev (commented):
The PR looks good to me overall. 
I've left some non-critical suggestions and question.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
MikeDvorskiy: Yes, it is intentional change.
Everywhere, where we have 1-dimensional iteration space and simultaneous access to more then one range, should be "minimum size logic" used, otherwise we have (might have) got UB.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
dmitriy-sobolev, line 565: Let's use `__first_size_calc` in this particular case: we know that `__view1` and `__view2` have equal sizes, so there no need to call std::min.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -386,11 +384,11 @@ __pattern_any_of(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range&&
         return false;
 
     using _Predicate = oneapi::dpl::unseq_backend::single_match_pred<_Pred>;
-    return oneapi::dpl::__par_backend_hetero::__parallel_find_or(
-        _BackendTag{},
-        __par_backend_hetero::make_wrapped_policy<oneapi::dpl::__par_backend_hetero::__or_policy_wrapper>(
-            ::std::forward<_ExecutionPolicy>(__exec)),
-        _Predicate{__pred}, oneapi::dpl::__par_backend_hetero::__parallel_or_tag{}, ::std::forward<_Range>(__rng));
+    using __or_tag = oneapi::dpl::__par_backend_hetero::__parallel_or_tag;
+    using __size_calc = oneapi::dpl::__par_backend_hetero::__min_size_calc;
```
dmitriy-sobolev, line 388: There is only one range. Let's use `__first_size_calc`

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -705,12 +686,17 @@ __pattern_find_if(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Iterato
         return __last;
 
     using _Predicate = oneapi::dpl::unseq_backend::single_match_pred<_Pred>;
+    using _IndexType = std::make_unsigned_t<typename std::iterator_traits<_Iterator>::difference_type>;
+    using _TagType = __par_backend_hetero::__parallel_find_forward_tag<_IndexType>;
+    using __size_calc = oneapi::dpl::__par_backend_hetero::__min_size_calc;
```
dmitriy-sobolev, line 691: There is only one input range: let's use `__first_size_calc` to avoid calling `std::min` under the hood.

MikeDvorskiy:
> but my understanding is that the predicate should return `false` at least once for the algorithm to produce `false`

In my understanding, I would say "two empty sequences are equal" :-)

danhoeflinger (commented):
A couple minor things, still working on the rest, but it has looked good so far.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2366,122 +2343,9 @@ __parallel_find_or_impl(oneapi::dpl::__internal::__device_backend_tag, _Executio
...
-    auto __buf = __keep(__first, __last);
-    auto __s_keep = oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _Iterator2>();
-    auto __s_buf = __s_keep(__s_first, __s_last);
-
-    using _TagType = ::std::conditional_t<_IsFirst::value, __parallel_find_forward_tag<decltype(__buf.all_view())>,
-                                          __parallel_find_backward_tag<decltype(__buf.all_view())>>;
-    return __first + oneapi::dpl::__par_backend_hetero::__parallel_find_or(
-                         oneapi::dpl::__internal::__device_backend_tag{},
-                         __par_backend_hetero::make_wrapped_policy<__find_policy_wrapper>(
-                             ::std::forward<_ExecutionPolicy>(__exec)),
```
danhoeflinger, line 2463: I'm struggling a bit to understand why we needed this wrapped policy layer previously (and if we would need something similar now).  I don't think it should be required here or in the new code because we are not calling multiple kernels from the same public API call (which is where we should need such wrapped policies).  
Please comment if you agree with this assessment or if there is something else here I'm missing.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -553,14 +559,19 @@ __pattern_adjacent_find(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _R
     auto __rng1 = oneapi::dpl::__ranges::take_view_simple(__rng, __rng.size() - 1);
     auto __rng2 = oneapi::dpl::__ranges::drop_view_simple(__rng, 1);
 
+    using __size_calc = oneapi::dpl::__ranges::__first_size_calc;
+
+    assert(__rng1.size() == __rng2.size());
+
     // TODO: in case of conflicting names
     // __par_backend_hetero::make_wrapped_policy<__par_backend_hetero::__or_policy_wrapper>()
```
danhoeflinger, line 567: I think we should be able to remove this TODO if you agree with my above assessment.  In any case, the comment needs to at least change because `__or_policy_wrapper` no longer exists.  This exists in a number of places.

danhoeflinger (commented):
Other than what I raised about stale comments, and the optional change in the tests, 
this LGTM.
Please fix at least the comments, (if you agree with my assessment about wrapped policies).

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
MikeDvorskiy: In my understanding, in the past, we thought that we have to provide one unique kernel name by each "dispatching patch" of any algorithm call.
We have many cases when several algorithm is dispatching to one middle layer pattern.
So, on each step of dispatching a device policy was wrapped into new one. It  made more or less guarantee that on final SYCL pattern we got (from the  policy) more or less unique kernel name...

But, now, I think such approach is not necessary. SYCL compiler can share kernels if we don't use lambdas..

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
MikeDvorskiy: removed

test/parallel_api/algorithm/alg.nonmodifying/equal.pass.cpp
```diff
@@ -145,18 +145,20 @@ test(size_t bits, Compare comp)
...
         T original = inCopy[0];
         inCopy[0] = !original;
-        invoke_on_all_policies<3>()(test_with_4_iters<T>(), in.begin(), in.begin() + n, inCopy.begin(), inCopy.begin() + n, false);
-        invoke_on_all_policies<4>()(test_with_4_iters<T>(), in.cbegin(), in.cbegin() + n, inCopy.cbegin(), inCopy.cbegin() + n, comp, false);
-        invoke_on_all_policies<5>()(test_with_3_iters<T>(), in.cbegin(), in.cbegin() + n, inCopy.cbegin(), false);
+        const bool ex_res_false = false || n == 0; //In case of the both sequences are empty, result is true.
+        invoke_on_all_policies<3>()(test_with_4_iters<T>(), in.begin(), in.begin() + n, inCopy.begin(), inCopy.begin() + n, ex_res_false);
+        invoke_on_all_policies<4>()(test_with_4_iters<T>(), in.cbegin(), in.cbegin() + n, inCopy.cbegin(), inCopy.cbegin() + n, comp, ex_res_false);
+        invoke_on_all_policies<5>()(test_with_3_iters<T>(), in.cbegin(), in.cbegin() + n, inCopy.cbegin(), ex_res_false);
         inCopy[0] = original;
```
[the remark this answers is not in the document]
MikeDvorskiy: But in this case i should write a additional block of calls outside the loop. I think it looks more bulky...

dmitriy-sobolev (approved):
Reapproving after cleaning up comments for the removed `__or_policy_wrapper`.

PR #2248, opened 2025-05-12
CI: treat a subset of warnings as errors

dmitriy-sobolev:
Let's treat already fixed warnings as errors. This PR excludes warnings which are still present - they should be iteratively removed.

This exclusion logic is possible with g++/clang++/icpx. cl does not support it; its warning-to-error list will expand and, probably, at some point, we would like to replace the whole list with /WX (treat all warnings as errors).


danhoeflinger:
Thanks for working on this.  I was just about to ask if we could do this at least for unused variable.

danhoeflinger (approved):
LGTM

PR #2249, opened 2025-05-12
Fix warnings in library and tests (icpx, device and oneTBB backends)

dmitriy-sobolev:
The PR will be split in multiple PRs

PR #2252, opened 2025-05-12
Fix warnings in esimd KT

dmitriy-sobolev:
The PR fixes `unused-variable`, `unused-parameter` and `unused-local-typedefs` warnings.

It's been cherry-picked from #2249 for easier review.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
+++ b/include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
@@ -211,7 +211,6 @@ __onesweep_impl(sycl::queue __q, _RngPack1&& __input_pack, _RngPack2&& __virt_pa
                 const _MemHolder& __mem_holder, ::std::size_t __n)
 {
     using _KeyT = typename ::std::decay_t<_RngPack1>::_KeyT;
-    using _ValT = typename ::std::decay_t<_RngPack1>::_ValT;
     constexpr bool __has_values = ::std::decay_t<_RngPack1>::__has_values;
 
     using _EsimdRadixSortHistogram = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
```diff
@@ -372,7 +371,7 @@ __onesweep(sycl::queue __q, _RngPack1&& __pack, _RngPack2&& __pack_out, ::std::s
 template <bool __is_ascending, ::std::uint8_t __radix_bits, bool __in_place, typename _RngPack1, typename _RngPack2,
           typename _KernelParam>
 sycl::event
-__radix_sort(sycl::queue __q, _RngPack1&& __pack_in, _RngPack2&& __pack_out, _KernelParam __param)
+__radix_sort(sycl::queue __q, _RngPack1&& __pack_in, _RngPack2&& __pack_out, _KernelParam)
```
SergeyKopienko, line 374: May we stay this type `_KernelParam` only in template parameters?

dmitriy-sobolev: I am inclined towards keeping it as is.
The same approach is in another KT - single pass scan. Also this signature matches the public interface.

SergeyKopienko: Up to you.
But I think that something that required parameter created in run-time but used only type info from it - is a little bit strange pattern.

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/experimental/kt/internal/esimd_radix_sort_dispatchers.h
dmitriy-sobolev: Yes, it is not a straightforward approach.

I'd like to keep it to avoid cascade effect of  `__param` being unused in upper layers. 
Also, this run-time variable should be optimized out, see a minimal example https://godbolt.org/z/rv4azYxf9 (foo and bar have equivalent assembly).

SergeyKopienko: It's not optimization problem. It's design problem: you are forcing the user of the interface to create instances of types that are simply not needed at run time.

dmitriy-sobolev: Thanks, that's a good point to stress in an RFC for this KT, which I am going to create...

PR #2255, opened 2025-05-12
Fix warnings in XPU tests

dmitriy-sobolev:
The PR fixes these types of warnings:

- `unused-parameter`
- `unused-variable`
- `missing-braces` (a weird one, probably a clang bug, but is easy to silence).

It's been cherry-picked from https://github.com/uxlfoundation/oneDPL/pull/2249 for easier review.

### DIFF
--- a/test/xpu_api/containers/sequences/array/array.data/data_const.pass.cpp
+++ b/test/xpu_api/containers/sequences/array/array.data/data_const.pass.cpp
@@ -62,6 +62,7 @@ main()
                     typedef dpl::array<T, 0> C;
                     const C c = {};
                     const T* p = c.data();
+                    (void)p;
                 }
             });
         });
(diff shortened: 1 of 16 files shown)


test/xpu_api/containers/sequences/array/array.data/data_const.pass.cpp
```diff
@@ -62,6 +62,7 @@ main()
                     typedef dpl::array<T, 0> C;
                     const C c = {};
                     const T* p = c.data();
+                    (void)p;
```
SergeyKopienko, line 65: May we use here `[[maybe_unused]]` instead of this new line?

SergeyKopienko (approved):
LGTM

test/xpu_api/containers/sequences/array/array.data/data_const.pass.cpp
dmitriy-sobolev: There is another case of (void)p; above, so I used that for consistency.

SergeyKopienko: OK

PR #2257, opened 2025-05-12
Fix deprecated-copy warning in the library

dmitriy-sobolev:
Examples of the warnings:

> induction_impl.h:53:5: error: definition of implicit copy constructor for '__induction_object<int, unsigned long>' is deprecated because it has a user-provided copy assignment operator

> unseq_backend_simd.h:547:5: error: definition of implicit copy assignment operator for '_Combiner<oneapi::dpl::__internal::tuple<unsigned long, unsigned int>, oneapi::dpl::internal::segmented_scan_fun<unsigned long, unsigned int, std::plus<unsigned long>>>' is deprecated because it has a user-provided copy constructor

---
Note that `_Combiner` has `_Combiner(const _Combiner& __obj) : __value{}, __bin_op(__obj.__bin_op) {}` constructor, which behavior is unusual. However, if a corresponding copy assignment operator also default initializes `__value`, several tests produce wrong results.

It's been cherry-picked from https://github.com/uxlfoundation/oneDPL/pull/2249 for easier review.

### DIFF
--- a/include/oneapi/dpl/pstl/experimental/internal/induction_impl.h
+++ b/include/oneapi/dpl/pstl/experimental/internal/induction_impl.h
@@ -49,6 +49,8 @@ class __induction_object
   public:
     __induction_object(__value_type __var, _Sp __stride) : __var_(__var), __stride_(__stride) {}
 
+    __induction_object(const __induction_object& __other) = default;
+
     __induction_object&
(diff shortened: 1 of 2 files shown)


MikeDvorskiy (approved):
LGTM

PR #2261, opened 2025-05-14
Remove the usages of `__future` on internal layers

SergeyKopienko:


SergeyKopienko:
@danhoeflinger, I think you may continue thinks about re-design of `__future` class and so on.
On internal layers this class doesn't needed anymore.

PR #2262, opened 2025-05-14
CI: treat a subset of warnings as errors in Windows

dmitriy-sobolev:
This PR changes the logic from permissive to restrictive for msvc. Note, there there is a limitation: /wd<dddd> turns off a warning entirely from downgrading it from warning-as-error to plain warning. Still, it should be a better approach

Additionally it fixes the way how these flags are passed. Currently, `warning_flags` variable is not created and thus ignored.


SergeyKopienko (approved):
LGTM

PR #2264, opened 2025-05-14
Fix warning in tests: declaration of <x> hides global declaration

dmitriy-sobolev:
The PR fixes a warning:

> test\support/utils.h(1033): warning C4459: declaration of 'max_n' hides global declaration

It renames local `max_n` to `max_size` to avoid a conflict with `TestUtils::max_n` or removes `using namespace TestUtils;`, whatever is simpler.


mmichel11:
The changes LGTM, but I also see a large number of local `max_n` variables in the `parallel_api/ranges` tests? Do these also need to be updated?

dmitriy-sobolev:
Good catch. I've noticed that issues because only msvc alerts about overshadowing. The rest issues appear to be in device tests only, which is not covered by msvc. I've fixed them.

mmichel11 (approved):
LGTM

PR #2266, opened 2025-05-15
Fix various warnings in tests

dmitriy-sobolev:
The PR fixes warnings like:

> all_of.pass.cpp(59): warning C4805: '^': unsafe mix of type 'size_t' and type 'const bool' in operation

> adjacent_difference.pass.cpp(81): warning C4456: declaration of 'temp' hides previous local declaration

`stateful_functor_no_copy_assign(const stateful_functor_no_copy_assign&) = default;` is added to avoid [-Wdeprecated-copy-with-user-provided-copy](https://clang.llvm.org/docs/DiagnosticsReference.html#id258).

There rest are either `unused-lambda-capture` or `unused-parameter`.

### DIFF
--- a/test/parallel_api/algorithm/alg.modifying.operations/shift_left_right.pass.cpp
+++ b/test/parallel_api/algorithm/alg.modifying.operations/shift_left_right.pass.cpp
@@ -172,7 +172,7 @@ struct shift_right_algo
     //skip the test for non-bidirectional iterator (forward iterator, etc)
     template <typename Policy, typename It>
     ::std::enable_if_t<!TestUtils::is_base_of_iterator_category_v<::std::bidirectional_iterator_tag, It>, It>
-    operator()(Policy&& exec, It first, It last, typename ::std::iterator_traits<It>::difference_type n)
+    operator()(Policy&&, It first, It, typename ::std::iterator_traits<It>::difference_type)
     {
         return first;
     }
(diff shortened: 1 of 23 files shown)


test/parallel_api/iterator/input_data_sweep.h
```diff
@@ -54,6 +54,7 @@ wrap_recurse(Policy&& exec, InputIterator1 first, InputIterator1 last, InputIter
 
     //Run the tests
     auto get_expect = [n](auto exp) {
+        (void)n; // avoid unused-lambda-capture, bug https://bugs.llvm.org/show_bug.cgi?id=35450
```
adamfidel, line 57: It's unfortunate the the `[[maybe_unused]]` attribute doesn't work in lambda capture lists, but it seems like this is the best that can be done here.

test/parallel_api/algorithm/alg.modifying.operations/shift_left_right.pass.cpp
```diff
@@ -172,7 +172,7 @@ struct shift_right_algo
     //skip the test for non-bidirectional iterator (forward iterator, etc)
     template <typename Policy, typename It>
     ::std::enable_if_t<!TestUtils::is_base_of_iterator_category_v<::std::bidirectional_iterator_tag, It>, It>
-    operator()(Policy&& exec, It first, It last, typename ::std::iterator_traits<It>::difference_type n)
+    operator()(Policy&&, It first, It, typename ::std::iterator_traits<It>::difference_type)
```
adamfidel, line 175:
```suggestion
    operator()(Policy&&, It first, It, typename std::iterator_traits<It>::difference_type)
```

adamfidel (approved):
LGTM besides the small suggestion. I can re-approve if you apply it.

adamfidel (approved):

test/parallel_api/iterator/input_data_sweep.h
dmitriy-sobolev: Yes, good note. `[[maybe_unused]]` would fit here it it was possible.

PR #2268, opened 2025-05-16
Fix warning: unary minus operator applied to unsigned type

dmitriy-sobolev:
The PR fixes warning like:

merge test:

> merge.pass.cpp(140): warning C4146: unary minus operator applied to unsigned type, result still unsigned

for_loop implementation:
> for_loop_impl.h(84): warning C4146: unary minus operator applied to unsigned type, result still unsigned

permutation iterator test:
> C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.43.34808\include\functional(62): warning C4146: unary minus operator applied to unsigned type, result still unsigned
...
oneDPL\test\parallel_api\iterator\permutation_iterator_parallel_transform_reduce.pass.cpp(96): note: see reference to...

exlusive_scan_by_segment implementation:
> C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.43.34808\include\functional(62): warning C4146: unary minus operator applied to unsigned type, result still unsigned
internal/exclusive_scan_by_segment_impl.h(84): note: see reference to class template instantiation 'std::negate<oneapi::dpl::__internal::__pattern_exclusive_scan_by_segment::FlagType>' being compiled

The case with the segmented exclusive scan is peculiar. It turned out `std::negate` was applied to `unsigned integer` flags, which should contain only `0` or `1`. It caused unsigned overflow to `1` values, which still resulted in `true` result. Now, flags are checked as they are, but the result is the same.

### DIFF
--- a/include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
+++ b/include/oneapi/dpl/internal/exclusive_scan_by_segment_impl.h
@@ -79,12 +79,8 @@ __pattern_exclusive_scan_by_segment(_Tag, Policy&& policy, InputIterator1 first1
 
     // TODO : add stencil form of replace_copy_if to oneDPL if the
     // transform call here is difficult to understand and maintain.
-#if 1
     transform(policy, first2, last2 - 1, _flags.get() + 1, _temp.get() + 1,
-              internal::replace_if_fun<T, ::std::negate<FlagType>>(::std::negate<FlagType>(), init));
-#else
-    replace_copy_if(policy1, first2, last2 - 1, _flags.get() + 1, _temp.get() + 1, ::std::negate<FlagType>(), init);
-#endif
+              internal::replace_by_flag<T>(init));
 
     // scan key-flag tuples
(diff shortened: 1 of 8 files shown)


dmitriy-sobolev:
I'm investigating wrong result in `HOST,bknd=tbb,cmplr=icx-cl,windows-latest,std=c++17,cfg=release` configuration.

dmitriy-sobolev:
> I'm investigating wrong result in `HOST,bknd=tbb,cmplr=icx-cl,windows-latest,std=c++17,cfg=release` configuration. It occurs only in release mode, with unseq policy, reverse iterator and Matrix type.

Update regarding exclusive_scan_by_segment: there appears to be a compiler bug related to vectorization, which is triggered by the change (probably, the code is not vectorized right now). I've rolled back the changes, and submitted the bug in the internal compiler tracker.

mmichel11 (approved):
LGTM

PR #2271, opened 2025-05-19
CI: remove fixed warnings from -Werror exclusions

dmitriy-sobolev:
Note that some of these warnings still occur when building all the tests with device backend (which is not currently covered in this CI). Anyway, let's have a strict warning policy with what we currently have in the CI.


mmichel11 (approved):
LGTM

PR #2269, opened 2025-05-16
[oneDPL][rfc][ranges][L1] the third portion of the range based API

MikeDvorskiy:
[oneDPL][rfc][ranges][L1] the third portion of the range based API

PR #2272, opened 2025-05-20
Add unique, unique_copy and swap_ranges parallel range algorithms

MikeDvorskiy:
The details can be found in the links below. 

Interface/Specification:
- `unique_copy`: https://github.com/uxlfoundation/oneAPI-spec/pull/635
- `unique`, `swap_ranges`: https://github.com/uxlfoundation/oneAPI-spec/pull/634

[RFC]( https://github.com/uxlfoundation/oneDPL/pull/2269). It denotes the requirements and the rationale.


include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -26,7 +26,7 @@
 #endif
 
 #include "utils_ranges.h" // __difference_t
-#include "utils.h"        // oneapi::dpl::__internal::__swap_ranges_fn
+#include "utils.h"        // oneapi::dpl::__internal::__swap_fn
```
dmitriy-sobolev, line 29: `__swap_fn` is not used in `glue_algorithm_ranges_impl.h` anymore. Is that inclusion still neede?

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -1028,6 +1056,106 @@ struct __remove_fn
 
 inline constexpr __internal::__remove_fn remove;
 
+// [alg.unique]
+
+namespace __internal
+{
+struct __unique_fn
+{
+
```
dmitriy-sobolev, line 1065:
```suggestion
```

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -1028,6 +1056,106 @@ struct __remove_fn
...
+    std::ranges::borrowed_iterator_t<_R>
+    operator()(_ExecutionPolicy&& __exec, _R&& __r) const
+    {
+        const auto __n = std::ranges::size(__r);
+        const auto __n_2 = __n / 2;
+        auto __r1 = std::ranges::take_view(__r, __n_2);
+        auto __r2 = std::ranges::take_view(std::ranges::reverse_view(__r), __n_2);
+
+        oneapi::dpl::ranges::swap_ranges(std::forward<_ExecutionPolicy>(__exec), std::move(__r1),std::move(__r2));
```
dmitriy-sobolev, line 1127: There is `__reverse_functor` called from `__pattern_reverse` in a device backend, which is specially optimized for this algorithm.  Let's reuse it.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -1028,6 +1056,106 @@ struct __remove_fn
...
+    requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<_ExecutionPolicy>>
+        && std::ranges::sized_range<_InRange> && std::ranges::sized_range<_OutRange>
+        && std::indirectly_copyable<std::ranges::iterator_t<_InRange>, std::ranges::iterator_t<_OutRange>>
+
+    std::ranges::reverse_copy_result<std::ranges::borrowed_subrange_t<_InRange>,
+                                     std::ranges::borrowed_subrange_t<_OutRange>>
+    operator()(_ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r) const
+    {
+        return oneapi::dpl::ranges::copy(std::forward<_ExecutionPolicy>(__exec), std::ranges::reverse_view(__in_r),
+            std::forward<_OutRange>(__out_r));
```
dmitriy-sobolev, line 1151: `__pattern_reverse_copy` from the device backend also uses an optimized `__reverse_copy` routine. Let's reuse it.

dmitriy-sobolev (commented):
I've checked the signatures for their compliance with https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2025/p3179r8.html#modify_copy - everything looks good.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
MikeDvorskiy: rather is not. good catch

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
MikeDvorskiy: yeah... it makes sense for hetero policy.

SergeyKopienko (dismissed):
In this changes I see more then one new lambdas like
```C++
template <typename _Tag, typename _ExecutionPolicy, typename _R, typename _Comp, typename _Proj>
std::ranges::borrowed_subrange_t<_R>
__pattern_unique(_Tag __tag, _ExecutionPolicy&& __exec, _R&& __r, _Comp __comp, _Proj __proj)
{
    auto __pred_2 = [__comp, __proj](auto&& __val1, auto&& __val2) { return std::invoke(__comp, std::invoke(__proj,
        std::forward<decltype(__val1)>(__val1)), std::invoke(__proj, std::forward<decltype(__val2)>(__val2)));};

    auto __beg = std::ranges::begin(__r);
    auto __end = __beg + std::ranges::size(__r);
    auto __it = oneapi::dpl::__internal::__pattern_unique(__tag, std::forward<_ExecutionPolicy>(__exec),
                                                          __beg, __end, __pred_2);

    return std::ranges::borrowed_subrange_t<_R>(__it, __end);
}
```
also created in some template code, parametrized by `ExecutionPolicy` as template parameter.

This mean you again introduce the problem with `ExecutionPolicy` passed as `l-value` or `r-value`.

If all lambda's like that created only in host code. please insert
```C++
static_assert(__is_serial_tag_v<_Tag> || __is_parallel_forward_tag_v<_Tag>);
```
in these functions.

UPD: looks as not actual now.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -921,11 +921,11 @@ struct __min_nested_type_size<oneapi::dpl::__internal::tuple<_Ts...>>
...
 };
 
-template <typename _ReferenceType1, typename _ReferenceType2>
-struct __swap_ranges_fn
+struct __swap_fn
 {
+    template <typename _Type1, typename _Type2>
     void
-    operator()(_ReferenceType1 __x, _ReferenceType2 __y) const
+    operator()(_Type1&& __x, _Type2&& __y) const
```
SergeyKopienko, line 928: I believe this change looks not quite correct.
Because now you are using here `universal reference` : this mean you may pass into this function `const reference` too.
But it's invalid use case and I believe better to rewrite this like
```C++
operator()(_Type1& __x, _Type2& __y) const
```

SergeyKopienko:
Minor comment: not all `::` before `std` in changed lines has been removed.

test/parallel_api/ranges/std_ranges_reverse.pass.cpp
```diff
@@ -0,0 +1,34 @@
...
+main()
+{
+#if _ENABLE_STD_RANGES_TESTING
+    using namespace test_std_ranges;
+    namespace dpl_ranges = oneapi::dpl::ranges;
+
+    //A checker below modifies a return type; a range based version with policy has another return type.
+    auto reverse_checker = [](auto&&... args) {
+        return std::ranges::reverse(std::forward<decltype(args)>(args)...);
+    };
```
dmitriy-sobolev, line 28: I do not quite understand why and how the return type is changed.

`std::ranges::reverse`, when fed with a range, returns `ranges::borrowed_iterator_t<R>`. The same type is returned by the algorithm with the policy. 

Is that checker really needed?

test/parallel_api/ranges/std_ranges_unique.pass.cpp
```diff
@@ -0,0 +1,51 @@
...
+#if _ENABLE_STD_RANGES_TESTING
+    using namespace test_std_ranges;
+    namespace dpl_ranges = oneapi::dpl::ranges;
+
+    //A checker below modifies a return type; a range based version with policy has another return type.
+    auto unique_checker = [](auto&&... args) {
+        return std::ranges::unique(std::forward<decltype(args)>(args)...);
+    };
+
+    test_range_algo<0>{big_sz}(dpl_ranges::unique, unique_checker);
```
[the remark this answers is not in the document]
dmitriy-sobolev: I see that `std::not_equal_to{}` is used below, but that's an edge case. Let's have a test case with large size (hence guaranteed parallel execution) with some adjacent copies.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
MikeDvorskiy: Lets wait an update from C++ committee. There is a chance that `ranges::reverse_copy` with policy will be removed from the parallel algo set.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -1028,6 +1055,105 @@ struct __remove_fn
...
+    }
+
+}; //__reverse_fn
+} //__internal
+
+inline constexpr __internal::__reverse_fn reverse;
+
+namespace __internal
+{
+struct __reverse_copy_fn
```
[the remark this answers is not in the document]
MikeDvorskiy: Lets wait an update from C++ committee. There is a chance that ranges::reverse_copy with policy will be removed from the parallel algo set.

MikeDvorskiy:
> This mean you again introduce the problem with `ExecutionPolicy` passed as `l-value` or `r-value`.

Essentially, no. This is a host part of the __pattern_unique implementation.
But yes, we can use `oneapi::dpl::__internal::__compare` functor instead of introducing new lambda.

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: > But it's invalid use case and I believe better to rewrite this like

No.... For example, a proxy type, like `tuple<int&, int&>` doesn't match to `_Type1& __x`.

test/parallel_api/ranges/std_ranges_reverse.pass.cpp
MikeDvorskiy: I guess, the comment is not relevant.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Ok

test/parallel_api/ranges/std_ranges_reverse.pass.cpp
dmitriy-sobolev: I suppose that the checker can be removed as well.

test/parallel_api/ranges/std_ranges_swap.pass.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+        std::ranges::copy(std::ranges::take_view(r1, size), std::ranges::take_view(r2, size).begin());
+
+        using ret_type = std::ranges::swap_ranges_result<std::ranges::borrowed_iterator_t<decltype(r1)>,
+            std::ranges::borrowed_iterator_t<decltype(r2)>>;
+
+        return ret_type{std::ranges::begin(r1) + size, std::ranges::begin(r2) + size};
+    };
+
+    test_range_algo<0, int, data_in_out_lim>{big_sz}(dpl_ranges::swap_ranges,  swap_ranges_checker);
```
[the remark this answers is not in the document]
dmitriy-sobolev: These commands run the test with sequences with different sizes:
https://github.com/uxlfoundation/oneDPL/blob/c57830cf8915a0adfb68701594a145267e5a0df4/test/parallel_api/ranges/std_ranges_test.h#L279-L280

I meant the check of the data inside of the sequences. I've found only one such a check, which is performed over the second sequence:
https://github.com/uxlfoundation/oneDPL/blob/c57830cf8915a0adfb68701594a145267e5a0df4/test/parallel_api/ranges/std_ranges_test.h#L256

The data in the first sequence is not checked.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
MikeDvorskiy: Update: The question was proceed with C++ committee and accepted into the C++ standard.
So,  `ranges::reverse_copy` should return 3 iterators: `the last` in the input range, `a last position` in the input range, `the last` in the output range.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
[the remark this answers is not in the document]
MikeDvorskiy: Update: The question was proceed with C++ committee and accepted into the C++ standard.
So, ranges::reverse_copy should return 3 iterators: the last in the input range, a last position in the input range, the last in the output range.

Yes, a test should be added.

test/parallel_api/ranges/std_ranges_swap.pass.cpp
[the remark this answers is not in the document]
MikeDvorskiy: added

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -1029,6 +1056,118 @@ struct __remove_fn
...
+{
+    template<typename _ExecutionPolicy, std::ranges::random_access_range _InRange,
+             std::ranges::random_access_range _OutRange>
+    requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<_ExecutionPolicy>>
+        && std::ranges::sized_range<_InRange> && std::ranges::sized_range<_OutRange>
+        && std::indirectly_copyable<std::ranges::iterator_t<_InRange>, std::ranges::iterator_t<_OutRange>>
+
+    std::ranges::in_in_out_result<std::ranges::borrowed_subrange_t<_InRange>,
```
dmitriy-sobolev: Should we define `reverse_copy_truncated_result` alias (see https://isocpp.org/files/papers/P3709R2.html) to be aligned with other algorithms?

I do not fully understand the rationale for the standard to have `<algorithm>_result` aliases, but I think we should define our own aliases for consistency if they are not available in C++20.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -706,11 +701,98 @@ __pattern_move(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __r, _OutRange&
...
+template <typename _ExecutionPolicy, typename _R, typename _Comp, typename _Proj>
+std::ranges::borrowed_subrange_t<_R>
+__pattern_unique(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy&&, _R&& __r, _Comp __comp, _Proj __proj)
+{
+    return std::ranges::unique(std::forward<_R>(__r), __comp, __proj);
+}
+
+template <typename _Tag, typename _ExecutionPolicy, typename _R>
+std::ranges::borrowed_subrange_t<_R>
+__pattern_reverse(_Tag __tag, _ExecutionPolicy&& __exec, _R&& __r)
```
dmitriy-sobolev: Let's move it to #2363.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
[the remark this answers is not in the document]
dmitriy-sobolev: Closing as reverse and reverse copy are moved into #2363.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
dmitriy-sobolev: Closing as reverse and reverse copy are moved into https://github.com/uxlfoundation/oneDPL/pull/2363.

dmitriy-sobolev:
@SergeyKopienko and @danhoeflinger, if you have a chance, could you review the PR as well?

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -1010,7 +1039,6 @@ namespace __internal
```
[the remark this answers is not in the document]
dmitriy-sobolev: It's consistent with the rest code in this file.

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -232,13 +232,15 @@ struct test
...
+        Container cont_in_exp(exec, n_in, DataGen1{});
+
         Container cont_out(exec, n_out, data_gen_zero);
         Container cont_exp(exec, n_out, data_gen_zero);
 
         assert(n_in <= max_n);
         assert(n_out <= max_n);
 
-        auto src_view = tr_in(std::views::all(cont_in()));
+        auto src_view = tr_in(std::views::all(cont_in_exp()));
```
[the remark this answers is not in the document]
dmitriy-sobolev: It's a preparation of an expected data. The behavour is correct.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -706,11 +701,75 @@ __pattern_move(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __r, _OutRange&
                                                    oneapi::dpl::__internal::__brick_move<decltype(__tag)>{});
 }
 
-template <typename _Tag, typename _ExecutionPolicy, typename _InRange, typename _OutRange>
+template <typename _ExecutionPolicy, typename _InRange, typename _OutRange>
 void
 __pattern_move(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy&&, _InRange&& __r, _OutRange&& __out_r)
 {
```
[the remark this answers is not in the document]
dmitriy-sobolev: Wrong assumption. _Tag is not needed here as it a serial fallback when _Tag type is known.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -245,8 +245,7 @@ __pattern_adjacent_find_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R&& __r,
...
     static_assert(__is_parallel_tag_v<_Tag> || typename _Tag::__is_vector{});
 
-    auto __pred_2 = [__pred, __proj](auto&& __val, auto&& __next) { return std::invoke(__pred, std::invoke(__proj,
-        std::forward<decltype(__val)>(__val)), std::invoke(__proj, std::forward<decltype(__next)>(__next)));};
+    oneapi::dpl::__internal::__compare __pred_2{__pred, __proj};
```
akukanov: Looks like it is time to return to the discussion here https://github.com/uxlfoundation/oneDPL/pull/1621#discussion_r1634832514 about `__compare` taking parameters by forwarding references vs. const lvalue references.

Back then, we decided to preserve `const T&` in `__compare` (as these were in the lambdas it replaced) even though @MikeDvorskiy wanted to switch to `T&&`. With the change here, we now have the opposite situation: forwarding references in the original lambda are replaced with `const T&` in `__compare`. As stated in the mentioned discussion, it might lead to compilation issues if a provided projection cannot accept a const object.

Tagging @danhoeflinger.

danhoeflinger: After revisiting the conversation from before, and if I understand the context appropriately in this code, we should not restrict the user code to require them use `const T&` projection / comparator arguments if we previously allowed non-const ref arguments.  

We could break user code (even if it is just code where someone neglected to add `const` to something which in practice is const).

danhoeflinger: I believe the remedy to this is to convert `__compare` to accept universal references. I think this would be better than creating a different mostly redundant struct just to handle these cases.

As mentioned in the other conversation, I don't object to this now that there is a good motivation for doing it.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -706,11 +701,75 @@ __pattern_move(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __r, _OutRange&
...
+    std::ranges::move(std::forward<_InRange>(__r), std::ranges::begin(__out_r));
+}
+
+template <typename _Tag, typename _ExecutionPolicy, typename _R1, typename _R2>
+void
+__pattern_swap_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&& __r2)
+{
+    static_assert(__is_parallel_tag_v<_Tag> || typename _Tag::__is_vector{});
+
+    auto __end = std::ranges::begin(__r1) + std::ranges::size(__r1);
```
danhoeflinger: nitpick:
Similar to below, perhaps we should add `auto __beg = std::ranges::begin(__r1);` and then reuse it rather than calling it multiple times.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -1284,13 +1363,9 @@ swap_ranges(_ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2)
...
-    using _ReferenceType1 = oneapi::dpl::__internal::__value_t<_Range1>&;
-    using _ReferenceType2 = oneapi::dpl::__internal::__value_t<_Range2>&;
-
-    return oneapi::dpl::__internal::__ranges::__pattern_swap(
-        __dispatch_tag, ::std::forward<_ExecutionPolicy>(__exec), views::all(::std::forward<_Range1>(__rng1)),
-        views::all(::std::forward<_Range2>(__rng2)),
-        oneapi::dpl::__internal::__swap_ranges_fn<_ReferenceType1, _ReferenceType2>{});
+    return oneapi::dpl::__internal::__ranges::__pattern_swap(__dispatch_tag, ::std::forward<_ExecutionPolicy>(__exec),
+                                                             views::all(::std::forward<_Range1>(__rng1)),
+                                                             views::all(::std::forward<_Range2>(__rng2)));
```
danhoeflinger:
```suggestion
    return oneapi::dpl::__internal::__ranges::__pattern_swap(__dispatch_tag, std::forward<_ExecutionPolicy>(__exec),
                                                             views::all(std::forward<_Range1>(__rng1)),
                                                             views::all(std::forward<_Range2>(__rng2)));
```

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -841,6 +840,36 @@ struct __move_fn
...
+    {
+        using _Size = std::common_type_t<std::ranges::range_size_t<_R1>, std::ranges::range_size_t<_R2>>;
+        const _Size __size = std::ranges::min((_Size)std::ranges::size(__r1), (_Size)std::ranges::size(__r2));
+
+        const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
+        oneapi::dpl::__internal::__ranges::__pattern_swap_ranges(__dispatch_tag, std::forward<_ExecutionPolicy>(__exec),
+                                                                 std::ranges::take_view(__r1, __size),
+                                                                 std::ranges::take_view(__r2, __size));
+
+        return {std::ranges::begin(__r1) + __size, std::ranges::begin(__r2) + __size};
```
danhoeflinger, line 874: I suppose we can technically forward these ranges into the begin call  (or alternatively store the begin iterator for each and forward into the pattern)

test/parallel_api/ranges/std_ranges_unique_copy.pass.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    auto unique_copy_checker = [](std::ranges::random_access_range auto&& r_in,
+                           std::ranges::random_access_range auto&& r_out, auto&&... args)
+    {
+        auto res = std::ranges::unique_copy(std::forward<decltype(r_in)>(r_in), std::ranges::begin(r_out),
+            std::forward<decltype(args)>(args)...);
+
+        using ret_type = std::ranges::unique_copy_result<std::ranges::borrowed_iterator_t<decltype(r_in)>,
+            std::ranges::borrowed_iterator_t<decltype(r_out)>>;
+        return ret_type{res.in, res.out};
```
danhoeflinger, line 33: I'm not following why all this is necessary.  Does `std::ranges::unique_copy` not return exactly this type already?
Why cant we just return its result directly?

test/parallel_api/ranges/std_ranges_swap.pass.cpp
```diff
@@ -0,0 +1,42 @@
...
+                           std::ranges::random_access_range auto&& r2)
+    {
+        const auto size = std::ranges::min(std::ranges::size(r1), std::ranges::size(r2));
+
+        std::ranges::swap_ranges(std::ranges::take_view(r1, size), std::ranges::take_view(r2, size));
+
+        using ret_type = std::ranges::swap_ranges_result<std::ranges::borrowed_iterator_t<decltype(r1)>,
+            std::ranges::borrowed_iterator_t<decltype(r2)>>;
+
+        return ret_type{std::ranges::begin(r1) + size, std::ranges::begin(r2) + size};
```
danhoeflinger: Shouldn't `std::ranges::swap_ranges` do the appropriate sizing and return type for us?  Why do we need to add extra machinery here?

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
SergeyKopienko: Hope it has been replaced now everywhere.

danhoeflinger: I think you may have changed some lines which were not previously modified.

test/parallel_api/ranges/std_ranges_unique_copy.pass.cpp
dmitriy-sobolev: C++20 return type signature:
```c++
ranges::unique_copy_result<borrowed_iterator_t<R>, O>
```

C++26 signature with an execution policy:
```c++
ranges::unique_copy_result<borrowed_iterator_t<R>, borrowed_iterator_t<OutR>>
```

danhoeflinger: Thanks, I forgot this discrepancy.

test/parallel_api/ranges/std_ranges_swap.pass.cpp
danhoeflinger: nevermind, I understand.  The C++20 signature doesnt have the range limited output and then the same buisiness with the borrowed_iterator from the range rather than the iterator.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
akukanov: According to the definition of `std::ranges::begin` https://eel.is/c++draft/range.access#begin, I see no point in forwarding. Also semantically, `begin` returns the first iterator to the range - so how could it ever "consume" that range?

We could compute the first and last iterators for both ranges and pass those to the pattern as a `subrange` instead of using `take_view`. Whether it would be better in any sense, I do not know.

akukanov:
The implementation of `unique_copy` cannot properly handle output ranges of insufficient size.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
danhoeflinger: Makes sense, thanks.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
dmitriy-sobolev: Now `__compare` accepts forwarding references.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -560,8 +559,9 @@ __pattern_copy(__serial_tag</*IsVector*/std::false_type>, _ExecutionPolicy&&, _I
 
 template <typename _Tag, typename _ExecutionPolicy, typename _InRange, typename _OutRange, typename _Pred,
           typename _Proj>
-auto
-__pattern_copy_if_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r, _Pred __pred, _Proj __proj)
+std::ranges::copy_if_result<std::ranges::borrowed_iterator_t<_InRange>, std::ranges::borrowed_iterator_t<_OutRange>>
+__pattern_copy_if_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r, _Pred __pred,
+                         _Proj __proj)
```
akukanov: The changes in this function are not related to the new algorithms.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -635,9 +652,9 @@ __pattern_copy_if(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range1&
 #if _ONEDPL_CPP20_RANGES_PRESENT
 template <typename _BackendTag, typename _ExecutionPolicy, typename _InRange, typename _OutRange, typename _Pred,
           typename _Proj>
-auto
-__pattern_copy_if_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r,
-    _Pred __pred, _Proj __proj)
+std::ranges::copy_if_result<std::ranges::borrowed_iterator_t<_InRange>, std::ranges::borrowed_iterator_t<_OutRange>>
+__pattern_copy_if_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r,
+                         _OutRange&& __out_r, _Pred __pred, _Proj __proj)
```
akukanov: The changes in this function are not related to the new algorithms.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -831,7 +884,8 @@ __pattern_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _Ran
 #if _ONEDPL_CPP20_RANGES_PRESENT
 template <typename _BackendTag, typename _ExecutionPolicy, typename _R1, typename _R2, typename _OutRange,
           typename _Comp, typename _Proj1, typename _Proj2>
-auto
+std::ranges::merge_result<std::ranges::borrowed_iterator_t<_R1>, std::ranges::borrowed_iterator_t<_R2>,
+                          std::ranges::borrowed_iterator_t<_OutRange>>
 __pattern_merge_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&& __r2,
```
akukanov: The changes in this function are not related to the new algorithms.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
dmitriy-sobolev: I've reverted the corresponding changes.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: I've reverted the corresponding changes.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: I've reverted the corresponding changes.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -152,30 +152,47 @@ struct __swap1_wrapper;
...
 
-template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Function>
+template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typename _Range2>
 oneapi::dpl::__internal::__difference_t<_Range1>
-__pattern_swap(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2, _Function __f)
+__pattern_swap(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2)
 {
+    //a trivial pre-check
+    if (__rng1.empty() || __rng2.empty())
+        return 0;
```
akukanov: Please leave a comment why calling the method `empty()` is valid in this place.

test/parallel_api/ranges/std_ranges_swap.pass.cpp
akukanov: Why should `std::ranges::swap_ranges` be wrapped like this, and not called directly?

akukanov (commented):
Overall it looks good to me.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
SergeyKopienko: The usage of `std::ranges::empty()` looks more correct in the common case due it's implementation:
https://www.en.cppreference.com/w/cpp/ranges/empty.html
```
_A call to ranges::empty is expression-equivalent to:

bool(t.empty()), if that expression is valid.
Otherwise, (ranges::size(t) == 0), if that expression is valid.
Otherwise, bool(ranges::begin(t) == ranges::end(t)), if that expression is valid and decltype(ranges::begin(t)) models std::forward_iterator.
In all other cases, a call to ranges::empty is ill-formed, which can result in substitution failure when ranges::empty(t) appears in the immediate context of a template instantiation._
```
Fixed.

test/parallel_api/ranges/std_ranges_swap.pass.cpp
SergeyKopienko: This approach has been introduced by @MikeDvorskiy 
In common case not all API calls from `std::ranges` namespace returns the type which we needed.
So this approach was chosen to explicitly form the result type that wee need.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
SergeyKopienko: This code has been re-implemented based on `size()` calls only:
1) we able to call `empty()` here because practically we pass into this function `all_view` which implements `empty()` method
```C++
class all_view
{
    // ...
    bool
    empty() const
    {
        return size() == 0;
    }
    // ...
};
```
2) But looking at this implementation no reasons anymore to use `empty()` - the sizes of both ranges are needed for us in any case later.

test/parallel_api/ranges/std_ranges_swap.pass.cpp
dmitriy-sobolev: It's called directly now. It's a rare case when the interfaces with/without the policies match.

akukanov (approved):
Looks reasonable. A documentation note is needed about `unique_copy` requiring sufficient output size.

SergeyKopienko (approved):
LGTM

PR #2274, opened 2025-05-20
[oneDPL][ranges][hetero] A hot fix for __pattern_adjacent_find; experimental ranges API use/test case

MikeDvorskiy:
A hot fix for the change from https://github.com/uxlfoundation/oneDPL/pull/2240

It fixes the following tests (experimental range API test cases)

- adjacent_find_ranges_sycl
- is_sorted_until_ranges_sycl.pass
- is_sorted_ranges_sycl.pass

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
@@ -543,18 +543,15 @@ __pattern_adjacent_find(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _R
     using _TagType = ::std::conditional_t<__is__or_semantic(), oneapi::dpl::__par_backend_hetero::__parallel_or_tag,
                                           oneapi::dpl::__par_backend_hetero::__parallel_find_forward_tag<_Range>>;
 
-    //ATTENTION!!! oneDPL supports SYCL buffer via a placeholder accessor; a 'subrange' cannot used here because
+    //ATTENTION!!! oneDPL supports SYCL buffer via a placeholder accessor; a 'subrange' cannot be used here because
     //getting an iterator for the placeholder accessor is incorrect on the host; so, oneDPL uses lazy-access views
     //for range transformations.
+    //For _ONEDPL_CPP20_RANGES_PRESENT, oneDPL may use std::ranges::take_view and std::ranges::drop_view, but there are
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -543,16 +543,17 @@ __pattern_adjacent_find(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _R
...
+    //getting an iterator for the placeholder accessor is incorrect on the host; so, oneDPL uses lazy-access views
+    //for range transformations.
+
 #if _ONEDPL_CPP20_RANGES_PRESENT
-    auto __first = std::ranges::begin(__rng);
     const auto __n = std::ranges::size(__rng);
-    auto __rng1 = std::ranges::subrange(__first, __first + __n - 1);
-    auto __rng2 = std::ranges::subrange(__first + 1, __first + __n);
+    auto __rng1 = std::ranges::take_view(__rng, __n - 1);
+    auto __rng2 = std::ranges::drop_view(__rng, 1);
```
dmitriy-sobolev, line 553: Can we use `take_view_simple` and `drop_view_simple` here to avoid issues with libstdc++10?

MikeDvorskiy: Does libstdc++10 already support C++20?
In that case, it makes sense to try...

dmitriy-sobolev: Yes, it does support the majority of c++20 features. See https://en.cppreference.com/w/cpp/compiler_support/20.

dmitriy-sobolev (approved):
LGTM

danhoeflinger (approved):
LGTM as well. 
It seem though that this issue with exceptions is a larger one, if it forces us not to use significant portions of the standard library. I'm not sure if there is any way around it in practice but we would definitely prefer to be able to use available c++20 functionality.

MikeDvorskiy:
> to be able to use available c++20 functionality.

Sure, we definitely prefer to use the standard views.
But, with a device policy we have to specify restrictions (came from SYCL restrictions)....
 Talking about standard take and drop views, only libcstd++ 10 has such implementation, the newer libcstd++ versions has more robust implementation of those views.

Issue #2210, opened 2025-04-28
Move scan-by-segment implementations into the `pstl/hetero/dpcpp` directory


mmichel11:
We currently have SYCL code in the `oneapi::dpl::__par_backend_hetero` namespace placed in the `internal` directory: https://github.com/uxlfoundation/oneDPL/blob/main/include/oneapi/dpl/internal/scan_by_segment_impl.h. From an architectural point-of-view, this is incorrect as SYCL backend specific code should be located in `pstl/hetero/dpcpp`.

mmichel11:
When we make this movement, we should also perform the following change: https://github.com/uxlfoundation/oneDPL/pull/2198/files#r2066454511

PR #2275, opened 2025-05-21
[Draft] Relocate SYCL scan-by-segment to `pstl/hetero/dpcpp`

mmichel11:
This addresses https://github.com/uxlfoundation/oneDPL/issues/2210.

The current SYCL implementation of the scan-by-segment algorithms is in the `internal` directory in the `oneapi::dpl::__par_backend_hetero` namespace. This does not align with our general architecture, so the implementation has been moved to `pstl/hetero/dpcpp/parallel_backend_sycl_scan_by_segment.h`. Furthermore, `__pattern_scan_by_segment` has been relocated to `pstl/hetero/algorithm_impl_hetero.h`.

Additionally, the following small tweaks are made:

- `::std` -> `std`
- A correction of an access mode in `__pattern_scan_by_segment`
- An added static assert

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_scan_by_segment.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_scan_by_segment.h
@@ -1,3 +1,5 @@
+// -*- C++ -*-
+//===-- parallel_backend_sycl_scan_by_segment.h ---------------------------===//
 /* Copyright (c) 2023 Intel Corporation
  *
  * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
@@ -27,10 +29,8 @@
  * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
  */
 
-#ifndef _ONEDPL_SCAN_BY_SEGMENT_IMPL_H
-#define _ONEDPL_SCAN_BY_SEGMENT_IMPL_H
-
-#if _ONEDPL_BACKEND_SYCL
+#ifndef _ONEDPL_PARALLEL_BACKEND_SYCL_SCAN_BY_SEGMENT_H
+#define _ONEDPL_PARALLEL_BACKEND_SYCL_SCAN_BY_SEGMENT_H
 
 #include <type_traits>
 #include <cstddef>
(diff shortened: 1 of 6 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_scan_by_segment.h
```diff
@@ -1,3 +1,5 @@
+// -*- C++ -*-
+//===-- parallel_backend_sycl_scan_by_segment.h ---------------------------===//
```
MikeDvorskiy: Last couple of years I am sharing a rule "no filename duplication in a file header".
Because, 
1) This information is redundant. A developer can see a filename in any case (in a tab or in a file tree).
2) Writing a filename might lead to "desynchronization" in case of file renaming - a developer rather forget to update a filename in this header... Even if he doesn't forget - it is double effort in any case.
3) A did not get any objects yet regarding that rule.

SergeyKopienko: My opinion - it just doesn't make sense to discuss about the things like this.
At this moment we have ~54 entries in our includes like that.
If author prefer to write it - let's go, no problems.

SergeyKopienko (approved):
LGTM

danhoeflinger (approved):
Yeah, LGTM as well.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_scan_by_segment.h
mmichel11: I do not have an issue if we want this to be the convention going forward, so I made this change. I also made some other small tweaks to the header comment to be consistent with other files.

SergeyKopienko (dismissed):
LGTM

SergeyKopienko (dismissed):
LGTM

MikeDvorskiy (dismissed):
LGTM

mmichel11:
I noticed some header includes in the moved scan-by-segment file are now unnecessary and a few others should be present. I updated these.

SergeyKopienko (approved):
LGTM

SergeyKopienko:
@mmichel11 thank you for this PR!

PR #2277, opened 2025-05-22
[oneDPL][ranges][hetero] usage of functors instead of lambdas

MikeDvorskiy:
[oneDPL][ranges][hetero] usage of functors instead of lambdas

### DIFF
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -150,6 +150,22 @@ struct __predicate
 template <typename _Comp, typename _Proj>
 using __compare = __predicate<_Comp, _Proj>;
 
+template <typename _F, typename _Proj1, typename _Proj2>
+struct __binary_op
+{
+    _F __f;
+    _Proj1 __proj1;
+    _Proj2 __proj2;
+
+    template <typename _TValue1, typename _TValue2>
+    auto
+    operator()(_TValue1&& __val1, _TValue2&& __val2) const
+    {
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/pstl/utils.h
```diff
@@ -261,6 +264,22 @@ class __not_equal_value
...
+
+  public:
+    explicit __set_value(const _Tp& __value) : _M_value(__value) {}
+
+    template <typename _Arg>
+    void
+    operator()(_Arg&& __arg) const
+    {
+        std::forward<_Arg>(__arg) = _M_value;
+    }
```
dmitriy-sobolev, line 280:
```suggestion
    operator()(_Arg& __arg) const
    {
        __arg = _M_value;
    }
```

I do not understand the purpose of a forwarding reference in this context. Moreover, it may lead to assigning to an rvalue.

dmitriy-sobolev: It seems it is for a zip_iterator proxy, which is an rvalue containing references.

MikeDvorskiy: 1. a proxy instance (rvalue) doesn't match to lvalue ref.
2. operator= may be overloaded

dmitriy-sobolev (approved):
LGTM

kboyarinov (dismissed):
LGTM

dmitriy-sobolev (dismissed):

dmitriy-sobolev (approved):
Reapproving after adding `mutable` which were missing even in the original code.
Please, wait for the CI to successfully pass before merging.

PR #2282, opened 2025-05-27
Using the same created `dpcpp` policy instance in tests

SergeyKopienko:
In this PR we replace `TestUtils::default_dpcpp_policy` inline variable by `TestUtils::get_dpcpp_test_policy()` function calls.

The `TestUtils::default_dpcpp_policy` inline variable has been deleted.

The goal - to make code more readable and to highlight that inside `get_dpcpp_test_policy()` function we have some logic which may return different policies for testing.

### DIFF
--- a/test/general/header_inclusion_order_async_1.pass.cpp
+++ b/test/general/header_inclusion_order_async_1.pass.cpp
@@ -23,7 +23,8 @@ main()
 {
 #if TEST_DPCPP_BACKEND_PRESENT
 
-    sycl::queue q = TestUtils::get_test_queue();
+    auto policy = TestUtils::dpcpp_policy();
+    sycl::queue q = policy.queue();
 
     constexpr std::size_t n = 100;
 
@@ -33,7 +34,7 @@ main()
     allocator alloc(q);
     std::vector<T, allocator> data(n, 1, alloc);
 
-    auto f = oneapi::dpl::experimental::reduce_async(TestUtils::make_device_policy(q), data.begin(), data.end());
+    auto f = oneapi::dpl::experimental::reduce_async(policy, data.begin(), data.end());
     f.wait();
 
 #endif // TEST_DPCPP_BACKEND_PRESENT
--- a/test/general/header_inclusion_order_async_1.pass.cpp
+++ b/test/general/header_inclusion_order_async_1.pass.cpp
@@ -23,7 +23,7 @@ main()
 {
 #if TEST_DPCPP_BACKEND_PRESENT
 
-    auto policy = TestUtils::dpcpp_policy();
+    auto policy = TestUtils::get_dpcpp_test_policy();
     sycl::queue q = policy.queue();
 
     constexpr std::size_t n = 100;
--- a/test/general/header_order_ranges_0.pass.cpp
+++ b/test/general/header_order_ranges_0.pass.cpp
@@ -27,7 +27,7 @@ main()
 {
 #if _ENABLE_RANGES_TESTING
     using namespace oneapi::dpl::experimental::ranges;
-    all_of(TestUtils::dpcpp_policy(), views::fill(-1, 10), [](auto i) { return i == -1;});
+    all_of(TestUtils::get_dpcpp_test_policy(), views::fill(-1, 10), [](auto i) { return i == -1;});
 #endif
 
     return TestUtils::done(_ENABLE_RANGES_TESTING);
--- a/test/general/header_order_ranges_0.pass.cpp
+++ b/test/general/header_order_ranges_0.pass.cpp
@@ -27,7 +27,7 @@ main()
 {
 #if _ENABLE_RANGES_TESTING
     using namespace oneapi::dpl::experimental::ranges;
-    all_of(TestUtils::default_dpcpp_policy, views::fill(-1, 10), [](auto i) { return i == -1;});
+    all_of(TestUtils::dpcpp_policy(), views::fill(-1, 10), [](auto i) { return i == -1;});
 #endif
 
     return TestUtils::done(_ENABLE_RANGES_TESTING);
--- a/test/general/lambda_naming.pass.cpp
+++ b/test/general/lambda_naming.pass.cpp
@@ -33,7 +33,7 @@ int main() {
     auto buf_begin = oneapi::dpl::begin(buf);
     auto buf_end = buf_begin + n;
 
-    const auto policy = TestUtils::dpcpp_policy();
+    const auto policy = TestUtils::get_dpcpp_test_policy();
     auto buf_begin_discard_write = oneapi::dpl::begin(buf, sycl::write_only, sycl::property::no_init{});
 
     ::std::fill(policy, buf_begin_discard_write, buf_begin_discard_write + n, 1);
--- a/test/general/lambda_naming.pass.cpp
+++ b/test/general/lambda_naming.pass.cpp
@@ -33,7 +33,7 @@ int main() {
     auto buf_begin = oneapi::dpl::begin(buf);
     auto buf_end = buf_begin + n;
 
-    const auto policy = TestUtils::default_dpcpp_policy;
+    const auto policy = TestUtils::dpcpp_policy();
     auto buf_begin_discard_write = oneapi::dpl::begin(buf, sycl::write_only, sycl::property::no_init{});
 
     ::std::fill(policy, buf_begin_discard_write, buf_begin_discard_write + n, 1);
--- a/test/parallel_api/ranges/std_ranges_test.h
+++ b/test/parallel_api/ranges/std_ranges_test.h
@@ -57,12 +57,10 @@ inline constexpr std::array<int, 2> big_sz = {/*serial*/ small_size, /*par*/ med
 #endif
 
 #if TEST_DPCPP_BACKEND_PRESENT
-template<int call_id = 0>
+template<typename PolicyName = class TestPolicyName, int call_id = 0>
 auto dpcpp_policy()
 {
-    auto exec = TestUtils::default_dpcpp_policy;
-    using Policy = decltype(exec);
-    return TestUtils::make_new_policy<TestUtils::new_kernel_name<Policy, call_id>>(TestUtils::default_dpcpp_policy);
+    return TestUtils::make_new_policy<TestUtils::new_kernel_name<PolicyName, call_id>>(TestUtils::get_test_queue());
 }
 #endif //TEST_DPCPP_BACKEND_PRESENT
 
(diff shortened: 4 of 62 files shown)


test/general/header_order_ranges_0.pass.cpp
```diff
@@ -27,7 +27,7 @@ main()
 {
 #if _ENABLE_RANGES_TESTING
     using namespace oneapi::dpl::experimental::ranges;
-    all_of(TestUtils::default_dpcpp_policy, views::fill(-1, 10), [](auto i) { return i == -1;});
+    all_of(TestUtils::dpcpp_policy(), views::fill(-1, 10), [](auto i) { return i == -1;});
```
dmitriy-sobolev, line 30: There is no `dpcpp_policy()` function in `TestUtils`.


A more general question: why do you consider `TestUtils::default_dpcpp_policy` as an improper way to e.g. create `sycl::queue` for the tests?

SergeyKopienko: > Got it - it makes sense. Could you add that into the description? I do not think it is a bug, though.

it's a bug because it doesn't terminate test process in case of exceptions like we did in `async_handler ` :
```C++
auto async_handler = [](sycl::exception_list ex_list) {
    for (auto& ex : ex_list)
    {
        try
        {
            ::std::rethrow_exception(ex);
        }
        catch (sycl::exception& ex)
        {
            ::std::cerr << ex.what() << ::std::endl;
            ::std::exit(EXIT_FAILURE);
        }
    }
};
```

dmitriy-sobolev: What do you think about replacing the remaining `TestUtils::default_dpcpp_policy` in the file below? 

https://github.com/uxlfoundation/oneDPL/blob/5273d6e3be5adb68e83da9cd14a0f825cc11b7d2/test/parallel_api/ranges/std_ranges_test.h#L65C95-L65C115

And, we still have these default policies, so they should be tested. I vote for doing it in `test_policies.pass`  - that should be enough. We can make `TestUtils::default_dpcpp_policy` local to that file.

SergeyKopienko: Fixed: `TestUtils::default_dpcpp_policy` - removed.

SergeyKopienko: >>> https://github.com/uxlfoundation/oneDPL/blob/5273d6e3be5adb68e83da9cd14a0f825cc11b7d2/test/parallel_api/ranges/std_ranges_test.h#L65C95-L65C115
- this function has been modified and moved into `TestUtils` namespace.

dmitriy-sobolev: > it's a bug because it doesn't terminate test process in case of exceptions

Thanks for the clarification, that's what I missed.

test/general/header_inclusion_order_async_1.pass.cpp
```diff
@@ -23,8 +23,6 @@ main()
 {
 #if TEST_DPCPP_BACKEND_PRESENT
 
-    sycl::queue q = TestUtils::get_test_queue();
-
```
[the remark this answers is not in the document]
SergeyKopienko: Unfortunately I couldn't to fix this right now, at least.
We have a lot of tests which works with `sycl::queue` directly without policy.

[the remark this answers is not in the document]
dmitriy-sobolev: The test does not compile due to the change. It must be fixed.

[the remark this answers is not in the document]
SergeyKopienko: Fixed

test/general/lambda_naming.pass.cpp
```diff
@@ -33,7 +33,7 @@ int main() {
     auto buf_begin = oneapi::dpl::begin(buf);
     auto buf_end = buf_begin + n;
 
-    const auto policy = TestUtils::default_dpcpp_policy;
+    const auto policy = TestUtils::get_dpcpp_test_policy();
```
MikeDvorskiy, line 36: In fact there is not any parameter to pass into `get_dpcpp_test_policy()` calls...
So, we most likely change only `TestUtils::default_dpcpp_policy` definition, instead of changing `TestUtils::default_dpcpp_policy` by `TestUtils::get_dpcpp_test_policy()` in several dozen files....

SergeyKopienko: Previously, in `TestUtils::default_dpcpp_policy` we doesn't used the instance of `sycl::queue` created in `get_test_queue()`

MikeDvorskiy: We can just change `TestUtils::default_dpcpp_policy` definition  - to add usage of the instance of `sycl::queue`, like created in `get_test_queue()`.

SergeyKopienko: good catch, fixed

dmitriy-sobolev: > We can just change TestUtils::default_dpcpp_policy definition - to add usage of the instance of sycl::queue, like created in get_test_queue().

It can make a false impression that `dpcpp_default` is being tested. I do not strongly insist, but I prefer having non-ambiguis names at the expense of more changes in the PR.

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -660,14 +650,14 @@ struct test_range_algo
...
                 test<T, usm_subrange<T>, mode, DataGen1, DataGen2>{}(
-                    n_device, dpcpp_policy<call_id + 30>(), algo, checker, std::identity{}, std::identity{}, args...);
+                    n_device, TestUtils::get_dpcpp_test_policy<class Test, call_id + 30>(), algo, checker, std::identity{}, std::identity{}, args...);
 #if TEST_CPP20_SPAN_PRESENT
                 test<T, usm_vector<T>, mode, DataGen1, DataGen2>{}(
-                    n_device, dpcpp_policy<call_id + 20>(), algo, checker, span_view, subrange_view, args...);
+                    n_device, TestUtils::get_dpcpp_test_policy<class Test, call_id + 20>(), algo, checker, span_view, subrange_view, args...);
                 test<T, usm_span<T>, mode, DataGen1, DataGen2>{}(
-                    n_device, dpcpp_policy<call_id + 40>(), algo, checker, std::identity{}, std::identity{}, args...);
+                    n_device, TestUtils::get_dpcpp_test_policy<class Test, call_id + 40>(), algo, checker, std::identity{}, std::identity{}, args...);
```
danhoeflinger, line 660: Here, if we switch the order, we can omit the repeated `class Test`.

test/support/utils_invoke.h
```diff
@@ -195,17 +216,15 @@ inline void unsupported_types_notifier(const sycl::device& device)
...
-        : queue(_queue)
-    {
-    }
-
     template <typename Op, typename... Args>
     void
     operator()(Op op, Args&&... rest)
     {
+        using kernel_name = unique_kernel_name<Op, CallNumber>;
+        auto my_policy = get_dpcpp_test_policy<kernel_name>();
```
danhoeflinger, line 224: This one just becomes `get_dpcpp_test_policy</*call_id=*/ 0,kernel_name>();` which is not so bad

test/support/utils_invoke.h
```diff
@@ -109,6 +118,21 @@ make_new_policy(sycl::queue _queue)
 #endif
 }
 
+template <typename PolicyName = class TestPolicyName, int call_id = 0>
```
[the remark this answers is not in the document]
danhoeflinger: Ah, I guess that makes sense.  As I mentioned, my initial instinct was to remove these features here entirely, and only return the "default" policy for the respective type.  Then let `new_kernel_name` and `make_new_policy` do their jobs in the situations where they are needed.  Perhaps that is what I would prefer in the end.  

In the end, it is all just preference and I wont hold the PR for that.  I'll leave it as optional.

test/parallel_api/ranges/std_ranges_test.h
SergeyKopienko: Fixed

test/support/utils_invoke.h
SergeyKopienko: Fixed

test/support/utils_invoke.h
```diff
@@ -20,10 +20,19 @@
...
+#else
+#  define TEST_USE_PREDEFINED_POLICIES 1
+#endif
+
 namespace TestUtils
 {
 #if TEST_DPCPP_BACKEND_PRESENT
 
+// Implemented in utils_sycl.h, required to include this file.
+sycl::queue get_test_queue();
```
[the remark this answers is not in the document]
danhoeflinger: I recognize that this is not new code, but rather moved code.  I'd be OK with an issue to be dealt with later.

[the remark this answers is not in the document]
SergeyKopienko: I realize this problem, but let me create some new issue and resolve it later.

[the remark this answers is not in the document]
SergeyKopienko: Issue created: https://github.com/uxlfoundation/oneDPL/issues/2289

test/support/utils_invoke.h
```diff
@@ -109,6 +118,21 @@ make_new_policy(sycl::queue _queue)
 #endif
 }
 
+template <int call_id = 0, typename PolicyName = class TestPolicyName>
+auto
+get_dpcpp_test_policy()
+{
+#    if TEST_USE_PREDEFINED_POLICIES
+#        if ONEDPL_FPGA_DEVICE
+    return make_new_policy<TestUtils::new_kernel_name<PolicyName, call_id>>(oneapi::dpl::execution::dpcpp_fpga);
```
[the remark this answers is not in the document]
SergeyKopienko: Sure: `default_dpcpp_policy` removed.

danhoeflinger (approved):
This LGTM, with a last couple optional nit picks.  
Wait for green CI and I'd recommend seeing if @MikeDvorskiy has any objection.

MikeDvorskiy (approved):
It seems that looks good overall..
But, I would recommend to launch fully testing before merging.

test/support/utils_invoke.h
```diff
@@ -109,6 +118,26 @@ make_new_policy(sycl::queue _queue)
...
+        oneapi::dpl::execution::dpcpp_fpga;
+#        else
+        oneapi::dpl::execution::dpcpp_default;
+#        endif // ONEDPL_FPGA_DEVICE
+#    else
+        get_test_queue();
+#    endif // TEST_USE_PREDEFINED_POLICIES
+
+    return TestUtils::make_new_policy<_NewKernelName>(__arg);
+}
```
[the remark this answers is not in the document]
SergeyKopienko: I think all this logic about `make_new_policy` still have place inside called `make_device_policy` function.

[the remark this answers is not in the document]
dmitriy-sobolev: Right, I see it now.

PR #2284, opened 2025-05-27
Move scan helpers, reorganize, add comments

danhoeflinger:
This PR adds does not intend to make any changes to the code.  

It merely moves the reduce_then_scan helper structures to their own header, `include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_scan_helpers.h`, reorganizes them, and adds some comments.

I'd prefer not to re-litigate contents of individual helper functions in this review, but if we notice bugs we can create issues and/or fix them in their own PR.

The interesting bits to review are probably the includes, and placement.


danhoeflinger:
We could instead move these to the existing reduce_then_scan header.  That header is already quite large, and I like the idea of separating them organizationally, but I'm open to other opinions here.

SergeyKopienko:
> We could instead move these to the existing reduce_then_scan header. That header is already quite large, and I like the idea of separating them organizationally, but I'm open to other opinions here.

For example, we have `parallel_backend_sycl_merge.h` with some functional for `merge`.
Does it make sense to extract `reduce & scan` functional from `parallel_backend_sycl.h` into new separate file too?

PR #2288, opened 2025-05-30
Using the same created `dpcpp` policy instance in tests V2

SergeyKopienko:
In this PR we fix `sycl::queue` instance used in `default_dpcpp_policy`.

The previous definition of `TestUtils::default_dpcpp_policy` in the tests isn't quite correct because it doesn't used `async_handler` [wrapper](https://github.com/uxlfoundation/oneDPL/blob/b061b7a751e1af59439e69c2900001399a699568/test/support/utils_sycl.h#L69) and not process possible exceptions like in `get_test_queue()` [function](https://github.com/uxlfoundation/oneDPL/blob/b061b7a751e1af59439e69c2900001399a699568/test/support/utils_sycl.h#L114), where we catch all  `sycl::exception` and then log it and exit from test process.

This is alternative approach for https://github.com/uxlfoundation/oneDPL/pull/2282 proposed by @MikeDvorskiy

### DIFF
--- a/test/support/utils_sycl.h
+++ b/test/support/utils_sycl.h
@@ -81,6 +81,22 @@ auto async_handler = [](sycl::exception_list ex_list) {
     }
 };
 
+#if ONEDPL_FPGA_DEVICE
+inline auto default_selector =
+#    if ONEDPL_FPGA_EMULATOR
+        sycl::ext::intel::fpga_emulator_selector_v;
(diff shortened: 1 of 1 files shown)


MikeDvorskiy (approved):
LGTM

danhoeflinger (approved):

Issue #1369, opened 2024-01-23
To do the same fix  for output type (by re-using __transform_functor if applicable), to make input type consistently: const T& or T&&


MikeDvorskiy:
//TODO: to do the same fix  for output type (by re-using __transform_functor if applicable) for the other functor below:
// __transform_if_unary_functor, __transform_if_binary_functor, __replace_functor, __replace_copy_functor
//TODO: to make input type consistently: const T& or T&&; to think which way is preferable

_Originally posted by @MikeDvorskiy in https://github.com/oneapi-src/oneDPL/pull/1233#discussion_r1463205525_

Proposed Implementation: 

template <typename _UnaryOper, typename _UnaryPred>
class __transform_if_unary_functor
{
    __transform_functor<_UnaryOper> _M_oper;
    mutable _UnaryPred _M_pred;

  public:
    explicit __transform_if_unary_functor(_UnaryOper&& __op, _UnaryPred&& __pred)
        : _M_oper(::std::forward<_UnaryOper>(__op)), _M_pred(::std::forward<_UnaryPred>(__pred))
    {
    }

    template <typename _Input1Type, typename _OutputType>
    void
    operator()(const _Input1Type& __x, _OutputType&& __output) const
    {
        if (_M_pred(__x))
            _M_oper(__x, ::std::forward<_OutputType>(__output));
    }
};

template <typename _BinaryOper, typename _BinaryPred>
class __transform_if_binary_functor
{
    __transform_functor<_BinaryOper> _M_oper;
    mutable _BinaryPred _M_pred;

  public:
    explicit __transform_if_binary_functor(_BinaryOper&& __op, _BinaryPred&& __pred)
        : _M_oper(::std::forward<_BinaryOper>(__op)), _M_pred(::std::forward<_BinaryPred>(__pred))
    {
    }

    template <typename _Input1Type, typename _Input2Type, typename _OutputType>
    void
    operator()(const _Input1Type& __x, const _Input2Type& __y, _OutputType&& __output) const
    {
        if (_M_pred(__x, __y))
            _M_oper(__x, __y, ::std::forward<_OutputType>(__output));
    }
};

template <typename _Tp, typename _Pred>
class __replace_functor
{
    const _Tp _M_value;
    _Pred _M_pred;

  public:
    __replace_functor(const _Tp& __value, _Pred __pred) : _M_value(__value), _M_pred(__pred) {}

    template <typename _OutputType>
    void
[...]

PR #2291, opened 2025-06-02
[oneDPL] The internal functors refactoring: reducing number of functors, renaming etc

MikeDvorskiy:
[oneDPL] The internal functors refactoring: reducing number of functors, renaming etc


SergeyKopienko (dismissed):
LGTM (please wait CI checks)

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -192,15 +192,21 @@ struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::
...
     : oneapi::dpl::__internal::__are_all_device_copyable<_F, _Proj1, _Proj2>
 {
 };
 
-template <typename _Pred>
-struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::__transform_functor, _Pred)>
-    : oneapi::dpl::__internal::__are_all_device_copyable<_Pred>
+template <typename _Pred, typename _RevTag>
+struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__internal::__transform_functor, _Pred, _RevTag)>
+    : oneapi::dpl::__internal::__are_all_device_copyable<_Pred, _RevTag>
```
danhoeflinger: We don't need to test if `_RevTag` is device copyable, assuming it is a bool_constant it will be device copyable, but it has no impact on if `__transform_functor` is.  Generally these should only be tested when they are types of member fields.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -71,10 +71,13 @@ class __set_value;
 template <typename _Comp, typename _Proj>
 struct __predicate;
 
+template <typename _F, typename _Proj>
```
danhoeflinger, line 74: Generally, when we add these to sycl_traits, we also add positive and negative tests to check them in `device_copyable.pass`. It may be a bit overkill, but that has been the convention so far.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -299,37 +314,41 @@ class __set_value
...
+    mutable _F __f;
 
   public:
-    explicit __transform_functor(_Pred __pred) : _M_pred(::std::move(__pred)) {}
+    explicit __transform_functor(_F __f) : __f(std::move(__f)) {}
 
     template <typename _Input1Type, typename _Input2Type, typename _OutputType>
     void
-    operator()(const _Input1Type& __x, const _Input2Type& __y, _OutputType&& __output) const
+    operator()(_Input1Type&& __x, _Input2Type&& __y, _OutputType&& __output) const
```
danhoeflinger, line 327: I have a memory that using const lvalue ref here for inputs instead of universal references was intentional but I don't remember why.  I think #1233 was the last time we considered this in detail. I'm looking through that PR now to try to remind myself.

danhoeflinger: Ah, now I see the comment and #1369, so it sounds like you have looked into this for this PR...

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
MikeDvorskiy: Yeah, perhaps you are right...

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
MikeDvorskiy: good catch, thanks!

MikeDvorskiy: Added tests for `__unary_op `and `__binary_op`.

danhoeflinger: I wont insist, but we normally check both the positive condition and negative condition. (it is device copyable when the relevant fields are device copyable, AND it is not device copyable when at least one relevant template argument is not device copyable)

MikeDvorskiy:
> LGTM, if you add the negative test, then I'll reapprove.

added.

danhoeflinger (approved):

include/oneapi/dpl/pstl/utils.h
```diff
@@ -150,6 +150,21 @@ struct __predicate
...
 
+template <typename _F, typename _Proj>
+struct __unary_op
+{
+    //'mutable' is to relax the requirements for a user functor or/and projection type operator() may be non-const
+    mutable _F __f;
+    mutable _Proj __proj;
+
+    template <typename _TValue>
+    decltype(auto)
```
SergeyKopienko, line 161: Could you explain here why we should have here exactly `decltype(auto)` instead of `auto` in previous implementation?
Which real gaps we fixing in this change?

include/oneapi/dpl/pstl/utils.h
```diff
@@ -159,7 +174,7 @@ struct __binary_op
     mutable _Proj2 __proj2;
 
     template <typename _TValue1, typename _TValue2>
-    auto
+    decltype(auto)
```
SergeyKopienko, line 177: The same

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: `auto` returns always of copy of type.
But `__unary_op` is kind of general unary operation, can be used for output value as well. For example, `__u_op(val) = x`;
Or, more useful in oneDPL case: `transform` algorithm - we wanna to keep a value category for move semantic: `*it_out = __u_op(*it_in);`

include/oneapi/dpl/pstl/utils.h
MikeDvorskiy: The motivation is the same, please see my explanation above.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: OK, thanks

PR #2294, opened 2025-06-04
Replace `oneapi::dpl::__internal::__no_op` by `oneapi::dpl::identity`

SergeyKopienko:
This PR replaces the internal `__no_op` functor with the public `oneapi::dpl::identity` across tests and library headers, and removes the now-unused `__no_op` definition.

- Swap out all uses of `oneapi::dpl::__internal::__no_op` for `oneapi::dpl::identity`
- Add `#include <functional>` (or the library’s `functional`) where `oneapi::dpl::identity` is used
- Remove the `__no_op` struct from `utils.h`


danhoeflinger:
This LGTM, but I think you need to rebase with main.

include/oneapi/dpl/internal/async_impl/glue_async_impl.h
```diff
@@ -18,6 +18,7 @@
 
 #include "../async_extension_defs.h"
 #include "async_impl_hetero.h"
+#include "oneapi/dpl/functional" // for oneapi::dpl::identity
```
akukanov: I am concerned about including a public header file from internal files. As far as I can tell, there is no precedent of that yet.
Perhaps it's not an immediate problem, because now `oneapi/dpl/functional` and its dependencies do not include any internal headers, Nevertheless, maybe it makes sense to move the actual implementation to an internal header, separate from the public header, and use that for internal purposes.

What do others think?

danhoeflinger: Good point.  I think it makes sense to move these functional definitions to an internal header.

SergeyKopienko: It's not a big deal to move this implementation into internal header.
But as far as this `oneapi::dpl::identity` was public before, my understanding that it should be public after move into internal header too, in `oneapi::dpl` namespace. 
Correct?

SergeyKopienko: > I am concerned about including a public header file from internal files. As far as I can tell, there is no precedent of that yet. Perhaps it's not an immediate problem, because now `oneapi/dpl/functional` and its dependencies do not include any internal headers, Nevertheless, maybe it makes sense to move the actual implementation to an internal header, separate from the public header, and use that for internal purposes.
> 
> What do others think?

Practically we had this include before but in a little bit another form, I found: https://github.com/search?q=repo%3Auxlfoundation%2FoneDPL%20%23include%20%22..%2Ffunctional%22&type=code

Example from https://github.com/uxlfoundation/oneDPL/blob/3f1c9a89141a47571b60f2a90944ef2d209692e7/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h#L19 :
```C++
#include "../../functional"
```
If we allow and prefer this form instead of `#include "oneapi/dpl/functional"` I will fix my changes.
May be it's also one of options for us?

MikeDvorskiy: > maybe it makes sense to move the actual implementation to an internal header

Am I right that `oneapi::dpl::identity` remains in Public API , and we are talking about the implementation moving? And in the functional public header will contain just `using oneapi::dpl::__internal::identity;`? (and for the other function objects - `minimum`, `maximum`) ?

In that case we are breaking the spec a little bit, where it is saying "The oneDPL function objects are defined in the <oneapi/dpl/functional> header."

References: 
[oneDPL spec](https://oneapi-spec.uxlfoundation.org/specifications/oneapi/v1.4-rev-1/elements/onedpl/source/sycl_kernels_api#function-objects)
[oneDPL doc](https://uxlfoundation.github.io/oneDPL/api_for_sycl_kernels/utility_function_object_classes.html?highlight=identity)

MikeDvorskiy: > ```c++
> ```c++
> #include "../../functional"
> ```
>   
> If we allow and prefer this form instead of #include "oneapi/dpl/functional" I will fix my changes.
> May be it's also one of options for us?
> ```

That one "include" case is again for usage of `oneapi::dpl::identity`. So, we will not need `#include "../../functional"` if we decide to use a kind of `oneapi::dpl::__internal::identity`.

SergeyKopienko: My opinion - I would prefer to save changes as is, without move our own `struct identity` into some another places.

danhoeflinger: Its too bad the spec was written in that way... 
I think it would probably be a benefit to adjusting the spec to use language more similar to the guide, where it doesn't specify where they are defined, but rather how you can get access to them. I would be **incredibly** surprised if moving them to an internal namespace and using an alias broke any user code.  
I'm really not sure the best course here though, I'd be interested in @akukanov opinion.

SergeyKopienko: The same approach applied in https://github.com/uxlfoundation/oneDPL/pull/2316 by @MikeDvorskiy :
```C++
#include "../../../functional" //for oneapi::dpl::identity
```
https://github.com/uxlfoundation/oneDPL/pull/2316/files#r2163424204

SergeyKopienko: Also we have the same includes in some other places inside our library like:
```C++
#include "../../functional"
```
,
```C++
#include "../functional"
```

Link for search: https://github.com/search?q=repo%3Auxlfoundation%2FoneDPL+%2Ffunctional%22&type=code

akukanov: I think you read the spec too literally folks.
From the usage standpoint, `oneapi/dpl/functional` is the header where the template is defined, even if in practice there is an internal header included into the public one. It does not break the specification.

SergeyKopienko: Fixed: please see my comment below: https://github.com/uxlfoundation/oneDPL/pull/2294#issuecomment-2999845570

danhoeflinger: I'm happy to hear that this is **too literal** a reading of the specification.  
That is some sort of achievement unlocked I think :).

SergeyKopienko: ...so let me mark this conversation as resolved.

test/parallel_api/iterator/input_data_sweep.h
```diff
@@ -153,6 +154,7 @@ wrap_recurse(Policy&& exec, InputIterator1 first, InputIterator1 last, InputIter
 #    endif // _ONEDPL_DEBUG_SYCL
         oneapi::dpl::discard_iterator discard{};
         // iterate through all wrappers and recurse - 1
+        auto noop = oneapi::dpl::identity{};
```
danhoeflinger, line 155: This needs some further changes... 

We need to remove the definition above `static auto noop = [](auto i) { return i; };`

and also should remove the braces from the comment `//transform_iterator(it, noop{})` L171, L181

SergeyKopienko: Fixed

SergeyKopienko:
> Tentatively this looks good, but it seems there are some CI issues to resolve?

Looks like all problems are in our examples:
```
The following tests FAILED:
	  1 - nstreams_sycl (Subprocess aborted)
	  2 - nstreams_policies (Subprocess aborted)
	  3 - sepia_sycl (Failed)
```

SergeyKopienko:
> > Tentatively this looks good, but it seems there are some CI issues to resolve?
> 
> It seems there are issues with FPGA emulator after automatic updating of icpx to 2025.2. I will take a look at it.

Thanks, @dmitriy-sobolev !

danhoeflinger (approved):
LGTM,
I'm OK with merging it.  Please check in with the nightly CI results to see if there is any affect on that in the coming days.

PR #2295, opened 2025-06-05
Forward execution policy in test environment

SergeyKopienko:
In this PR we forward execution policy in test environment through `std::forward` call.
It's important for some another PR which we going to apply later.

### DIFF
--- a/test/support/utils_sort.h
+++ b/test/support/utils_sort.h
@@ -386,6 +386,8 @@ run_test(SortTestConfig config,
     // Run tests for USM shared memory (external testing for USM shared memory, once already covered in sycl_iterator.pass.cpp)
     if (config.test_usm_shared)
     {
+        // TODO required later to create temporary policy instance here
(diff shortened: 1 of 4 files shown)


test/support/utils_sort.h
```diff
@@ -386,6 +386,8 @@ run_test(SortTestConfig config,
     // Run tests for USM shared memory (external testing for USM shared memory, once already covered in sycl_iterator.pass.cpp)
     if (config.test_usm_shared)
     {
+        // TODO required later to create temporary policy instance here
+        // and pass it to test_usm with value category like exec parameter
```
MikeDvorskiy: I would suggest:
`// TODO: Create a temporary policy instance and pass it to test_usm with the needed value category`

SergeyKopienko: Fixed

MikeDvorskiy (approved):
LGTM

PR #2297, opened 2025-06-06
Fix definition with same mangled name in `parallel_find_or`

SergeyKopienko:
In this PR we fix definition with same mangled name in `parallel_find_or`, which happens in build of `find_end_ranges_sycl.pass` test:
```
cmake -DCMAKE_CXX_COMPILER=icpx -DCMAKE_CXX_STANDARD=20 -DONEDPL_BACKEND=dpcpp -DONEDPL_DEVICE_TYPE=GPU '-DCMAKE_CXX_FLAGS=-DTEST_LONG_RUN=0 -fsycl-device-code-split=per_kernel ' -DCMAKE_BUILD_TYPE=release -D_ENABLE_RANGES_TESTING=1 ..

make find_end_ranges_sycl.pass
```

Build error log:
```
Building CXX object test/CMakeFiles/find_end_ranges_sycl.pass.dir/parallel_api/ranges/find_end_ranges_sycl.pass.cpp.o
In file included from /oneDPL/test/parallel_api/ranges/find_end_ranges_sycl.pass.cpp:18:
In file included from /oneDPL/include/oneapi/dpl/execution:67:
In file included from /oneDPL/include/oneapi/dpl/pstl/algorithm_impl.h:28:
In file included from /oneDPL/include/oneapi/dpl/pstl/execution_impl.h:22:
In file included from /oneDPL/include/oneapi/dpl/pstl/parallel_backend.h:32:
/oneDPL/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h:1443:17: error: definition with same mangled name '_ZTSN6oneapi3dpl20__par_backend_hetero23__find_or_kernel_one_wgIJN9TestUtils18unique_kernel_nameINS0_9execution5__dpl13device_policyINS4_INS3_14TestPolicyNameELm0EEEEELm0EEEEEE' as another definition
1443 |                 [=](sycl::nd_item</*dim=*/1> __item_id) {
      |                 ^
/oneDPL/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h:1443:17: note: previous definition is here
[...]

PR #2286, opened 2025-05-27
find_or fixes for unnamed lambda

danhoeflinger:
This PR  adds the appropriate kernel name wrappers for runtime branches in find_or algorithms.

#2247 removed some wrappers around find_or patterns, and exposed some issues where we are not wrapping kernel names in true runtime branches between patterns.  This results in build errors complaining of same mangled name when using unnamed lambdas.

PR #2298, opened 2025-06-06
Fix definition with same mangled name in `find_end`

danhoeflinger:
Alternative to #2297.  
Fix for `find_end` same mangled name error. 
Wrap kernel name for runtime branch which results in a different kernel.


SergeyKopienko:
@danhoeflinger do you really checked broken test with my `cmake` string?

danhoeflinger:
@SergeyKopienko Yes, I checked it locally with 2025.1, and it is resolved with this change.  Your cmake string was missing the unnamed lambda off argument, but this fixes it after I add it

danhoeflinger:
This is basically the same fix as #2286, just for ranges.

SergeyKopienko (approved):
LGTM

PR #2299, opened 2025-06-06
Ensure the correct type is scanned over in SYCL reduce-then-scan

mmichel11:
Currently, the result of `__unary_op` in `__gen_transform_input` will be scanned over in the reduce-then-scan implementation. This is incorrect, particularly if an initial value is provided by the user. We fix this by converting the resulting of `__unary_op` to the initial value type.

This requires an assumption that the input iterator's value type is convertible to the initial value type which is a known limitation with the TBB backend and the legacy SYCL scan implementation.


timmiesmith:
@mmichel11 would you please also add a note to the Known Limitations section of the introduction as part of this PR?

timmiesmith (approved):
LGTM. Thank you.

PR #2300, opened 2025-06-06
[Draft] RFC future type and keepalives

danhoeflinger:
This is a work in progress draft.

This RFC outlines some issues with our current `__future` and `__result_and_scratch_storage`, and start to lay out a path to toward improvements.

### DIFF
--- a/rfcs/proposed/future_and_keepalive/README.md
+++ b/rfcs/proposed/future_and_keepalive/README.md
@@ -62,11 +62,11 @@ we use to implement them under the hood:
 `set_symmetric_difference` as a combination of scan-like algorithms and merge algorithms because of the need to write 
 from the second input set.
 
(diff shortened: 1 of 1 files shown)


SergeyKopienko:
My proposal:
- remove `__future` usage on all  internal levels at all and use it only in async algorithm implementations (idea of @akukanov) - this approach implemented in the PR https://github.com/uxlfoundation/oneDPL/pull/2261
- extend functional of `__result_and_scratch_storage` to keep different types of result and type of temporary data.
Probably also make sense to implement `void` result type without memory allocation for it in this case.

SergeyKopienko:
I think we may try to isolate temporary data on sycl-submitter's level.

Probably we can implement this approach like in variant (c) in RFC:
 - create special final host task in each sycl-submitter which has temporary data;
 - finalize their temporary data on this special final host task.

PR #2301, opened 2025-06-07
RFC for Kernel Templates

dmitriy-sobolev:
The RFC describes the structure and common design of kernel templates. It also lists design gaps and potential improvements.

RFCs for the individual kernel templates will be added separately.

### DIFF
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -206,6 +206,7 @@ Currently, the specializations of the algorithms belong to different namespaces.
 
 Using a tag as a function or a template parameter should be considered as well.
 This method has both advantages and disadvantages compared to namespace-based separation.
+
 Pros:
 - Avoids potential name clashes caused by `using` directives.
 
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -18,7 +18,7 @@ with many design aspects yet to be addressed.
 See the [Open Questions](#open-questions) list.
 
 Algorithms which are already implemented are described in
-https://uxlfoundation.github.io/oneDPL/kernel_templates_main.html.
+[Kernel Templates API](https://uxlfoundation.github.io/oneDPL/kernel_templates_main.html).
 
 ## Proposal
 
@@ -289,6 +289,20 @@ The `kernel_param` targets a single kernel, and currently, only one instance can
 
 Currently, it is unclear if we should defferentiate algorithms
 which rely on compiler extensions.
+
+### Kernel Templates as a Backend for Algorithms with Standard Interfaces
+
+It should be evaluated whether the proposed algorithms can serve as a backend for oneDPL algorithms
+that use the standard C++ interfaces and device execution policies.
+
+It will affect which data-passing mechanisms,
+as described in [Abstract Algorithm Signature](abstract-algorithm-signature) section, are allowed.
+In particular, it should be investigated if supporting these entities is necessary and
+whether they can be handled by an additional thin layer between
+the algorithms with standard interfaces and Kernel Templates:
+- `oneapi::dpl::begin` and `oneapi::dpl::end`, which mimic iterators to pass `sycl::buffer` to algorithms accepting iterators.
+- Iterators to a host-allocated `std::vector` as described in
+  [Pass Data to Algorithms > Use std::vector](https://github.com/uxlfoundation/oneDPL/blob/release/2022.9.0/documentation/library_guide/parallel_api/pass_data_algorithms.rst#use-stdvector)
 
 ## Exit Criteria
 
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -166,11 +166,15 @@ int main()
 
 The testing harness should cover:
  - Varying numbers of elements to process, from thousands to millions.
- - Different element types.
- - Diverse types of input sequences.
- - Representative data distributions.
+ - Different element types, depending on the algorithm and its implementation details.
+   For example, tests for radix should cover various arithmetic types
+   due to differing ordered bit transformations.
+ - All supported input sequence types.
+   For example, if the algorithm supports
+   `sycl::buffer`, `oneapi::dpl::begin`/`oneapi::dpl::end`, or USM pointers, each should be tested.
  - Various combinations of other parameters, depending on the signature of the algorithm.
- - Edge cases (for example, zero-sized sequences).
+ - Edge cases.
+   For example, zero-sized sequences.
 
 Due to the unlimited number of possible kernel parameter combinations, it is recommended to:
  - Always test the most representative configurations
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -179,7 +179,7 @@ The testing harness should cover:
    due to differing ordered bit transformations.
  - All supported input sequence types.
    For example, if the algorithm supports
-   `sycl::buffer`, `oneapi::dpl::begin`/`oneapi::dpl::end`, or USM pointers, each should be tested.
+   `sycl::buffer` or USM pointers, each should be tested.
  - Various combinations of other parameters, depending on the signature of the algorithm.
  - Edge cases.
    For example, zero-sized sequences.
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -277,7 +277,7 @@ The `kernel_param` targets a single kernel, and currently, only one instance can
 ### Compiler Extension and Differentiation of Algorithms
 
 Currently, it is unclear if we should defferentiate algorithms
-which rely compiler extensions.
+which rely on compiler extensions.
 
 ## Exit Criteria
 
@@ -286,7 +286,7 @@ The proposed set of algorithms should become fully supported if:
   [Kernel Configuration](kernel_configuration/README.md#open-questions) sections are either addressed
   or provided with a justification to postpone or ignore.
 - A significant portion of the algorithms listed in
-  [Algorithms to Implement](#algorithms-to-implement) is implemented.
+  [Algorithms to Implement](#algorithms-to-implement) are implemented.
 - Evidence of sufficiently good performance is provided.
 - There is positive adoption feedback.
 
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -221,12 +221,12 @@ Additionally, it reduces the number of interfaces that need to be supported.
 This should be done in conjunction with the
 [Global and Local Memory Requirements](#reporting-global-and-local-memory-requirements) interface.
 
-### Asynchronous Execution
+### Asynchronous Execution and Dependency Chaining
 
 The algorithms return a `sycl::event` but do not accept any input events.
-As a result, they are not fully asynchronous.
+That limits how they can be embedded in asynchronous dependency chains.
 
-It should be investigated whether full asynchrony is desirable,
+It should be investigated whether passing input events is desirable,
 and if so, how to implement it correctly.
 
 ### Separation of Specializations Based on a Problem Size
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -272,6 +272,11 @@ making default values less relevant.
 It needs to be explored how to configure algorithms with multiple performance-critical kernels.
 The `kernel_param` targets a single kernel, and currently, only one instance can be passed.
 
+### Compiler Extension and Differentiation of Algorithms
+
+Currently, it is unclear if we should defferentiate algorithms
+which rely compiler extensions.
+
 ## Exit Criteria
 
 The proposed set of algorithms should transition to an extension if:
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -104,7 +104,11 @@ for algorithms with device policies.
 Additionally, a plain `sycl::buffer` can also be used.
 Specialized algorithms may impose additional restrictions on how data is passed.
 
-If an algorithm allocates global memory, it must throw `std::bad_alloc` if the allocation is unsuccessful.
+If an algorithm allocates global memory and that allocation is unsuccessful,
+it must throw `std::bad_alloc`.
+That behaviour may change depending on how
+[External Allocation of Global Memory](#external-allocation-of-global-memory)
+is addressed.
 
 ### Example
 
@@ -183,7 +187,7 @@ The renaming requires changing the corresponding namespace.
 ### Specializations and their Differntiation
 
 Currently, the specializations of the algorithms belong to different namespaces.
-Using tags instead tags offers several advantages,
+Using tags instead namespaces offers several advantages,
 for example easier dispatching between specializations and avoiding deeply nested namespaces.
 
 ### Reporting Global and Local Memory Requirements
@@ -237,7 +241,7 @@ for a given workload and hardware.
 
 The proposed set of algorithms should transition to an extension if:
 - All the [Open Questions](#open-questions) are addressed.
-  Note, in the future, some questions may arise,
+  Note that in the future, some questions may arise
   addressing which is unnecessary for the transition.
 - A significant portion of the algorithms listed in
   [Algorithms to Implement](#algorithms-to-implement) is implemented.
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -98,11 +98,10 @@ sycl::event kernel_template (
 ```
 
 `Arg1`, ..., `ArgN` include the sequences to be processed.
-Data can be passed in the same ways as described in the
+Data can be passed using the same mechanisms described in the
 [documentation on passing data](https://uxlfoundation.github.io/oneDPL/parallel_api/pass_data_algorithms.html#pass-data-to-algorithms)
 for algorithms with device policies.
-Additionally, a plain `sycl::buffer` can also be used.
-Specialized algorithms may impose additional restrictions on how data is passed.
+Specialized algorithms may extend or restrict the supported data passing mechanisms.
 
 If an algorithm allocates global memory and that allocation is unsuccessful,
 it must throw `std::bad_alloc`.
@@ -215,7 +214,7 @@ This should be done in conjunction with the
 
 ### Asynchronous Execution
 
-The algorithms return a `sycl::event` but do not accept any input events as inputs.
+The algorithms return a `sycl::event` but do not accept any input events.
 As a result, they are not fully asynchronous.
 
 It should be investigated whether full asynchrony is desirable,
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -314,20 +314,6 @@ Several questions arise:
 It should be evaluated whether the proposed algorithms can serve as a backend for oneDPL algorithms
 that use the standard C++ interfaces and device execution policies.
 
-### Encoding Requirements
-
-Some requirements, for example forward-progress guarantees between work-groups,
-are currently only described in documentation.
-Encoding such requirements programmatically would improve visibility and
-help avoid unintended usage.
-If fallback implementations exist,
-this also enables smooth switching when the target device changes.
-
-One option is to define a macro for each such requirement.
-These macros could be used to select appropriate implementations
-for algorithms with standard interfaces, as described in
-[Kernel Templates as a Backend for Algorithms with Standard Interfaces](kernel-templates-as-a-backend-for-algorithms-with-standard-interfaces).
-
 ## Exit Criteria
 
 The proposed set of algorithms should become fully supported if:
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -354,6 +354,35 @@ and to address concerns related to usability and expressiveness.
 - Range support:
   Should ranges be supported in the interface, or are iterators sufficient?
 
+### Vectorizing/Widenning of Load and Store Operations
+
+Many GPU architectures often support wide load and store operations
+(128B, 256B, or larger depending on a specific architecture) per sub-group
+that can significantly improve global memory and instruction throughput.
+These operations may require each individual work-item to handle multiple contiguous elements,
+effectively performing vectorized memory accesses.
+
+This optimization is already used in several oneDPL algorithms with standard interfaces.
+See [PR #1976](https://github.com/uxlfoundation/oneDPL/pull/1976) as an example.
+
+We need to explore if there is a need for a new control and
+if the optimization can be efficiently and portably done for the most relevant architectures.
+
+If a universal solution is not feasible, we should consider adding a new control,
+which raises questions:
+- How should it interact with `data_per_workitem`?
+- Should it be applied to all or only to a subset of algorithms?
+- What should the interface look like?
+
+If a new control is needed, we may add `vector_width` parameter.
+When combined with `data_per_workitem`, the memory access pattern could be configured as follows:
+1. **Vector mode:** `data_per_workitem` equals `vector_width`.
+  Each work-item loads/stores a contiguous block of `vector_width` elements.
+2. **Strided mode:** `vector_width = 1`.
+  Each work-item loads/stores `data_per_workitem` elements in a strided fashion.
+3. **Combined mode:** `data_per_workitem` is a multiple of `vector_width`.
+  Each work-item loads/stores contiguous `vector_width` elements per stride.
+
 ## Exit Criteria
 
 The proposed set of algorithms should become fully supported if:
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -314,18 +314,6 @@ Several questions arise:
 It should be evaluated whether the proposed algorithms can serve as a backend for oneDPL algorithms
 that use the standard C++ interfaces and device execution policies.
 
-An additional question is how to handle hardware requirements that cannot be queried directly
-through standard SYCL mechanisms. For example, selecting a more performant kernel template
-might depend on forward-progress guarantees between work-groups.
-While oneAPI DPC++ supports querying this via
-the [Root Group](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_root_group.asciidoc)
-experimental extension, other SYCL implementations may lack a corresponding feature.
-Furthermore, even if a device does not explicitly support cooperative kernel launch,
-it may still provide sufficient guarantees in limited cases
-such as chain-like synchronization between work-groups.
-A possible solution is to allow users to define a macro
-indicating that the required hardware capability is present.
-
 ### Encoding Requirements
 
 Some requirements, for example forward-progress guarantees between work-groups,
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -103,7 +103,7 @@ If an algorithm allocates global memory, it must throw `std::bad_alloc` if the a
 ### Example
 
 The example demonstrates the use of a kernel template
-and describes how it is tuned for better performance and what makes it less portable
+and describes how it is tuned for better performance and what makes it less general
 than an alternative algorithm with a standard interface.
 It uses `oneapi::dpl::experimental::kt::gpu::esimd::radix_sort`,
 which can be compared to `oneapi::dpl::stable_sort`.
@@ -119,8 +119,7 @@ global memory access, and the utilization of hardware computational resources.
 They can be easily adjusted for another GPU with different hardware characteristics.
 
 The kernle template relies on ESIMD technology
-and certain forward progress guarantees between work-groups,
-which leads to better performance at the expense of portability.
+and certain forward progress guarantees between work-groups.
 
 ```c++
 // icpx -fsycl radix_sort.cpp -o radix_sort -I /path/to/oneDPL/include && ./radix_sort
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -7,20 +7,26 @@ designed to achieve optimal performance on specific hardware and workloads.
 They prioritize efficiency over generality when compared to
 algorithms that conform to standard C++ interfaces and use device execution policies.
 
-These algorithms are intended for use with SYCL and are expected to evolve into an extension.
+These algorithms are intended for use with SYCL.
 
-This set of algorithms is in an early stage of development, with many questions yet to be addressed.
+The terms "kernel template" and "algorithm" are used interchangeably in this RFC.
+
+## Status
+
+Kernel Templates proposal is expected to evolve into an extension.
+This set of algorithms is in an early stage of development,
+with many design aspects yet to be addressed.
+See the [Open Questions](#open-questions) list.
 
 Algorithms which are already implemented are described in
 https://uxlfoundation.github.io/oneDPL/kernel_templates_main.html.
 
-The terms "kernel template" and "algorithm" are used interchangeably in this RFC.
-
-## Design
+## Proposal
 
 ### High-Level Structure
 
-The algorithms are defined in `<oneapi/dpl/experimental/kernel_templates>`, in `namespace oneapi::dpl::experimental::kt`.
+The algorithms are defined in `<oneapi/dpl/experimental/kernel_templates>`,
+in `namespace oneapi::dpl::experimental::kt`.
 This namespace contains portable algorithms and namespaces with more specialized implementations.
 
 Below is an example of such a structure with possible algorithms:
@@ -167,8 +173,6 @@ for example through cmake arguments.
 
 ## Open Questions
 
-There are several design aspects which should be addressed to make it a fully-supported extension.
-
 ### Name
 
 The name "Kernel Templates" may be misleading because
@@ -228,3 +232,17 @@ A list of algorithms to implement should be defined.
 
 A benchmark suite can help select the best configuration and parameters
 for a given workload and hardware.
+
+## Exit Criteria
+
+The proposed set of algorithms should transition to an extension if:
+- All the [Open Questions](#open-questions) are addressed.
+  Note, in the future, some questions may arise,
+  addressing which is unnecessary for the transition.
+- A significant portion of the algorithms listed in
+  [Algorithms to Implement](#algorithms-to-implement) is implemented.
+- They are performant.
+- There is a positive adoption feedback.
+
+Some individual algorithms may remain experimental
+and have their own exit criteria.
--- a/rfcs/experimental/kernel_templates/README.md
+++ b/rfcs/experimental/kernel_templates/README.md
@@ -104,9 +104,17 @@ sycl::event kernel_template (
 ```
 
 `Arg1`, ..., `ArgN` include sequences to process.
-Data must be passed using the same mechanisms described in the
-[documentation on passing data](https://uxlfoundation.github.io/oneDPL/parallel_api/pass_data_algorithms.html#pass-data-to-algorithms)
-for algorithms with device policies.
+These arguments must reference data which is ready for processing;
+any necessary data transfers must be completed in advance or managed automatically.
+Supported data storage types include:
+- [SYCL buffers](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:buffers).
+- Unified Shared Memory ([USM](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:usm)).
+- `std::vector` with `sycl::usm_allocator`.
+
+Supported data passing mechanisms are listed in
+the [documentation on passing data](https://uxlfoundation.github.io/oneDPL/parallel_api/pass_data_algorithms.html#pass-data-to-algorithms).
+The `oneapi::dpl::begin` and `oneapi::dpl::end` helper functions are not supported;
+however, `sycl::buffer` objects may be passed directly.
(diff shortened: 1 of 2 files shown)


SergeyKopienko (dismissed):
LGTM

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,283 @@
+# Kernel Templates
+
+## Introduction
+
+This RFC describes an experimental set of algorithms with easily tunable controls,
+designed to achieve optimal performance on specific hardware and workloads.
+They prioritize efficiency over generality when compared to
+algorithms that conform to standard C++ interfaces and use device execution policies.
```
akukanov:
```suggestion
algorithms that conform to the standard C++ interfaces.
```

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+```c++
+// defined in <oneapi/dpl/experimental/kernel_templates>
+// in namespace oneapi::dpl::experimental::kt
+
+template <
+    typename AlgorithmParameter1,  // manually specified, may have a default
+    typename AlgorithmParameter2,  // ...
+    typename AlgorithmParameterN,  // manually specified, may have a default
+
+    typename KernelParam,          // manually specified, may have a default in future
```
akukanov: Should not this be marked as deducible instead, and the corresponding function parameter as manually specified?

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+There is no requirement for the sets of algorithms in each namespace to be identical.
+
+### Abstract Algorithm Signature
+
+A kernel template is a C++ function invoked from the host.
+
+Each function takes a `sycl::queue` as its first argument and
+an instance of the [kernel_param](kernel_configuration/README.md) template,
+which applies settings common to all kernel templates, as its last argument.
```
akukanov:
```suggestion
Each function takes a `sycl::queue` as its first argument and
an instance of the [kernel_param](kernel_configuration/README.md) template,
which specializes parameters common to all kernel templates, as its last argument.
```

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+The example demonstrates the use of a kernel template
+and describes how it is tuned for better performance and what makes it less general
+than an alternative algorithm with a standard interface.
+It uses `oneapi::dpl::experimental::kt::gpu::esimd::radix_sort`,
+which can be compared to `oneapi::dpl::stable_sort`.
+
+Performance tuning controls:
+1. Algorithm-specific: the number of bits sorted per radix sort pass (`8`).
+2. Common: the number of elements processed per sub-group (`416`)
+  and the number of sub-groups per work-group (`64`).
```
akukanov: Would not it be better to consistently refer to the WG size parameter as the number of work items per WG, even for ESIMD? Whether the operations on data are done sequentially, by sub-group algorithms, or by ESIMD is actually more of an implementation detail. In the documentation for ESIMD, it is said that the subgroup size must always be 1.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+}
+```
+
+## Testing
+
+The testing harness should cover:
+ - Varying numbers of elements to process, from thousands to millions.
+ - Different element types.
+ - Diverse types of input sequences.
+ - Representative data distributions.
```
akukanov: It would be good to provide examples that clarify the intent. Otherwise, it can be hard to see the difference between these items.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+for example through cmake arguments.
+
+## Open Questions
+
+### Name
+
+The name "Kernel Templates" may be misleading because
+these entities more act like algorithms rather than SYCL kernels.
+Probably, a better name is "Algorithm Templates".
+The renaming requires changing the corresponding namespace.
```
akukanov:
```suggestion
The name "Kernel Templates" may be misleading because
these entities more act like algorithms rather than SYCL kernels.
Renaming would require changing the top-level namespace.
```

I removed the line that contained a subjective assessment.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+### Specializations and their Differentiation
+
+Currently, the specializations of the algorithms belong to different namespaces.
+Using tags instead of namespaces offers several advantages,
+for example easier dispatching between specializations and avoiding deeply nested namespaces.
```
akukanov: There are disadvantages as well, but those are not mentioned. The "deeply nested" namespaces (why is it considered a disadvantage?) give a hierarchical structure that is harder to replicate with tags (unless it is built into the tag names). In case implementations diverge substantially to impact the API (for example, algorithm-specific and even common parameters might differ), different namespaces might be preferred to tags. Namespace aliasing might also provide syntactic advantages.

Do not get me wrong, it is fair to ask if the current API paradigm should be reconsidered. But then let's view things from multiple angles and avoid portraying something as strictly advantageous (perhaps unintendedly, but nevertheless).

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+There are scenarios which benefit from passing pre-allocated temporary memory to an algorithm.
+For example, a user program may manage a reusable memory pool.
+Also, this is especially useful when an algorithm is called in a loop and the allocation overhead
+cannot be amortized by memory pools provided by the device or compiler runtime.
+
+It may be preferable to make this the only option and
+avoid internal global memory allocations altogether.
+This approach eliminates the overhead associated with `host_task`,
+which retains temporary memory until an algorithm completes its asynchronous execution.
```
akukanov:
```suggestion
This approach eliminates the overhead associated with retaining temporary memory
until an algorithm completes its asynchronous execution.
```

Host tasks are not the only way to deal with temporary allocations.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+This should be done in conjunction with the
+[Global and Local Memory Requirements](#reporting-global-and-local-memory-requirements) interface.
+
+### Asynchronous Execution
+
+The algorithms return a `sycl::event` but do not accept any input events.
+As a result, they are not fully asynchronous.
+
+It should be investigated whether full asynchrony is desirable,
+and if so, how to implement it correctly.
```
akukanov: What do you mean by "full asynchrony", and why is it only "partial" without input events?

It might be more appropriate to speak about supporting explicit dependency management.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+
+Different approaches may be more effective when processing different number of elements.
+For example, a small number of elements can be processed by a single work-group.
+
+Separating such cases would reduce compilation time.
+
+### Algorithms to Implement
+
+A list of algorithms to implement should be defined.
+
```
akukanov: Why is this an open question? What purpose would the list of algorithms serve?

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+They prioritize efficiency over generality when compared to
+algorithms that conform to standard C++ interfaces and use device execution policies.
+
+These algorithms are intended for use with SYCL.
+
+The terms "kernel template" and "algorithm" are used interchangeably in this RFC.
+
+## Status
+
+Kernel Templates proposal is expected to evolve into an extension.
```
akukanov: An extension of what?

I guess you mean that there is no intention to add KT into the oneDPL specification. But that might change, and I am not sure if it is important in the context of this RFC.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+The `kernel_param` targets a single kernel, and currently, only one instance can be passed.
+
+### Compiler Extension and Differentiation of Algorithms
+
+Currently, it is unclear if we should defferentiate algorithms
+which rely compiler extensions.
+
+## Exit Criteria
+
+The proposed set of algorithms should transition to an extension if:
```
akukanov: Here, again, let's speak about transition into production, leaving aside whether it's "an extension" or something else.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+- All the questions under [Kernel Templates](#open-questions) and
+  [Kernel Configuration](kernel_configuration/README.md#open-questions) sections are addressed,
+  unless a question is marked as not mandatory.
```
akukanov:
```suggestion
- All the questions under [Kernel Templates](#open-questions) and
  [Kernel Configuration](kernel_configuration/README.md#open-questions) sections are either addressed
  or provided with a justification to postpone or ignore.
```

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,292 @@
...
+- A significant portion of the algorithms listed in
+  [Algorithms to Implement](#algorithms-to-implement) is implemented.
+- They are performant.
```
akukanov:
```suggestion
- Evidence of sufficiently good performance is provided.
```

rfcs/experimental/kernel_templates/kernel_configuration/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+The library guide describes all the signature details in
+https://github.com/uxlfoundation/oneDPL/blob/oneDPL-2022.8.0-release/documentation/library_guide/kernel_templates/kernel_configuration.rst.
+
+## Open Questions
+
+### Runtime Parameters
+
+The `data_per_workitem` and `workgroup_size` parameters are currently provided at compile time,
+but their optimal values depend on the number of elements to process, which is known only at run time.
```
akukanov:
```suggestion
The `data_per_workitem` and `workgroup_size` parameters are currently provided at compile time,
but their optimal values may depend on the number of elements to process, which is known only at run time.
```

rfcs/experimental/kernel_templates/kernel_configuration/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+It should be investigated whether these parameters can be made run-time configurable
+without compromising efficiency.
```
akukanov: It sounds like an "either/or" choice between compile-time and run-time parameters, but it does not have to be such.
```suggestion
It should be investigated whether dynamic, run-time configurable parameters can be supported
without compromising efficiency.
```

rfcs/experimental/kernel_templates/kernel_configuration/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Sub-group Size as a Parameter
+
+Choosing an appropriate sub-group size is important for performance,
```
akukanov:
```suggestion
For some devices configuring a sub-group size might be important for performance,
```

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: Corrected

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: I've removed that note.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: Corrected.

I meant that the algorithms cannot be used in asynchronous chains as a non-first point. That limitation does not make them "partially" asynchronous, so thanks for the comment.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: I've added examples of what I meant under `Different element types` and `Diverse types of input sequences` points and removed `Representative data distributions`.

The later is too obvious, for example: the data should be unsorted before calling the sort algorithm being tested, or if an algorithm is searching something, there should be data to find in one case or nothing in another case and etc.

akukanov: Looks good now, thanks.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,294 @@
...
+    /* any cvref */ ArgN argn,
+    KernelParam param              // manually specified, may have a default in future
+);
+```
+
+`Arg1`, ..., `ArgN` include sequences to process.
+Data must be passed using the same mechanisms described in the
+[documentation on passing data](https://uxlfoundation.github.io/oneDPL/parallel_api/pass_data_algorithms.html#pass-data-to-algorithms)
+for algorithms with device policies.
+Specialized algorithms may extend or restrict the supported data passing mechanisms.
```
danhoeflinger: Do we intend to support host allocated data in Kernel templates?
  
I'd suggest we require data to be indirectly device accessible, given that these are intended to favor performance over productivity and this is sycl specific.  However, if we want to support buffers with `begin/end` (which I think we do), we will still need to process the input, so it may not make much of a technical difference in ease of implementation.  We may want to detect and exclude data which is not indirectly device accessible, and / or process inputs a little differently from mainline oneDPL algorithms.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,294 @@
...
+#include <iostream>
+#include <sycl/sycl.hpp>
+
+#include <oneapi/dpl/experimental/kernel_templates>
+
+namespace kt = oneapi::dpl::experimental::kt;
+
+int main()
+{
+   std::size_t n = 6;
```
SergeyKopienko:
```suggestion
   const std::size_t n = 6;
```

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,294 @@
...
+### Configuring an Algorithm with Multiple Kernels
+
+It needs to be explored how to configure algorithms with multiple performance-critical kernels.
+The `kernel_param` targets a single kernel, and currently, only one instance can be passed.
+
+### Compiler Extension and Differentiation of Algorithms
+
+Currently, it is unclear if we should defferentiate algorithms
+which rely compiler extensions.
+
```
danhoeflinger, line 385: I'd like to propose another open question surrounding using KT as an "enhanced" sycl backend for oneDPL with opt-in macros committing to specific hardware / capabilities at compile time.

For instance, a user could opt in with a macro `ONEDPL_FWD_PROG`, which states that they will run on hardware which has sufficient forward progress guarantees, allowing oneDPL mainline APIs to leverage KTs which require it.
Similarly, one could opt in to specific hardware at compile time with a similar affect.   

This is one way to allow users to benefit from KT development if they know their hardware / capabilities at compile time without changing their code.  

We already need to document all such restrictions for users to properly use KT, so it makes sense to me to encode it programmatically as well.

danhoeflinger: Beyond keeping track of such requirements in an official manner, there is design work to figure out how such algorithms would be utilized in practice by oneDPL mainline.  Since both oneDPL mainline and KT need to "process" input data to handle buffer wrapper inputs, we may want a thin layer between the public KT interface and an internal interface which oneDPL mainline would use.  

So there are a few questions here, whether we want to pursue this, and then how it would work structurally.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,294 @@
...
+### Configuring an Algorithm with Multiple Kernels
+
+It needs to be explored how to configure algorithms with multiple performance-critical kernels.
+The `kernel_param` targets a single kernel, and currently, only one instance can be passed.
+
+### Compiler Extension and Differentiation of Algorithms
+
+Currently, it is unclear if we should defferentiate algorithms
+which rely compiler extensions.
```
mmichel11:
```suggestion
which rely on compiler extensions.
```

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Exit Criteria
+
+The proposed set of algorithms should become fully supported if:
+- All the questions under [Kernel Templates](#open-questions) and
+  [Kernel Configuration](kernel_configuration/README.md#open-questions) sections are either addressed
+  or provided with a justification to postpone or ignore.
+- A significant portion of the algorithms listed in
+  [Algorithms to Implement](#algorithms-to-implement) is implemented.
```
mmichel11:
```suggestion
  [Algorithms to Implement](#algorithms-to-implement) are implemented.
```

rfcs/experimental/kernel_templates/kernel_configuration/README.md
```diff
@@ -0,0 +1,45 @@
+# Kernel Configuration
+
+## Introduction
+
+`kernel_param` is a generic structure for configuring SYCL kernels.
+Each kernel template is invoked with a `kernel_param` type or object.
+Typically, a kernel template supports multiple values for the configuration parameters.
+Optimal values may depend on the invoked kernel, the data size and type(s),
+as well as on the used device.
```
mmichel11: Optional wording adjustment.
```suggestion
Optimal values may depend on the invoked kernel, the data size and type(s),
and underlying hardware characteristics.
```

rfcs/experimental/kernel_templates/kernel_configuration/README.md
```diff
@@ -0,0 +1,45 @@
...
+### Sub-group Size as a Parameter
+
+For some devices configuring a sub-group size might be important for performance,
+as it impacts factors such as register pressure and
+the number of sub-groups to synchronize.
+The optimal value depends on the device architecture and
+the characteristics of the workload, for example, the type of elements being processed.
+
+The performance benefit should be demonstrated in practice,
+given the added complexity to the interface.
```
mmichel11, line 45: I propose adding another open-question to kernel configuration: Indexing type.

We have observed non-trivial performance benefits of using `std::uint32_t` to index as opposed to `std::uint64_t` in certain algorithms (e.g. merge) where input sizes are small enough. Perhaps this could be specified through the user in `kernel_param` to avoid compiling extra kernels for algorithms where this is relevant.

An alternative would be a change to the KT function signature where the user provides the input size as a templated type and only start iterators as opposed to full sequences (something similar to `std::copy_n`). This would enable a compile time decision in our backend.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: Yes, it was unintentional, and thank you giving such feedback - this is what I was looking for. I've updated that section. Now, the approach with namespaces looks much more preferrable, but we can keep that section to make the final decision after adding more algorithms.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: It would serve as a plan which algorithms to implement. It can be useful, e.g. for external contributors as a suggestion what can definitely be helpful for oneDPL.

However, the structure being proposed can have many specializations which cannot be predicted (e.g. specializations for future architectures), it is difficult to define the list. Perhaps, it can be a minimal list of algorithms in the portable namespace only.

dmitriy-sobolev: Also, it can be one of the indicators of maturity of the Kernel Templates feature for it to exit the experimental status.

dmitriy-sobolev: I definitely should make the description more precise or remove the section entirely, depending on what we decide here. I'd like to get your opinion first, e.g. if my answer regarding the purpose of the list is compelling.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: Thanks for bringing up that question. I have not dedicated sufficient time to the data passing mechanisms, essentially saying that kernel templates in the portable namespace should adhere to the rules for the core oneDPL algorithms with an assumption that these kernel templates will be a fallback for the core algorithms. As you have noted below, the whole idea of kernel templates as a backend needs to be investigated, and we can have a thin layer adapting kernel templates for the core algorithms.

There is an idea that kernel templates should be closer to SYCL level and behave as its extension with device-level algorithms. We should allow `sycl::buffer` from that perspective, even if it holds host-allocated data. From the performance standpoint, it may make writing performant code less obvious. I am curious if we can mitigate it somehow, but keep interoperability with SYCL buffers.

I do not think that supporting `begin`/`end` is needed if we can can implement a thin layer between core oneDPL algorithms using `begin`/`end` and kernel templates accepting `sycl::buffer`. That's because these abstractions serve as iterators to allow passing `sycl::buffer` to algorithms with a standard interface.

Would you agree to make it a yet another open question?

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: I'll add a section for KT as a backend for the oneDPL mainline. I like the idea with the macro as both a mechanism to express the requirement on a code level and as a knob for oneDPL mainline.

rfcs/experimental/kernel_templates/README.md
danhoeflinger: Yes, it is fine to make this an open question.  I think it needs to be discussed and planned out.

The real question I have here is not about supporting `sycl::buffer` holding host memory (I think we should support that), but rather my question is about the "automatic" support of `std::vector::iterator` using a host allocator.  For me, I think we should explicitly exclude such inputs from KT (...and really I'd be in favor of dropping support for this in oneDPL main as well, but that's another story). Support for such types does not align with the mindset behind KT.

rfcs/experimental/kernel_templates/kernel_configuration/README.md
dmitriy-sobolev: Thanks, a corresponding section has been added.

I have not listed the alternative approach, though. I am not aware of any practical use-cases or demand to pass a number of elements at compile time, and it is more restrictive than passing an index type.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: I've mentioned it

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: Done. I was a bit hesitant, because we currently need only info about cross work-group forward progress, and DPC++ allows querying it through root group extension. However, it's still an experimental feature and there might be other compilers.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -290,6 +290,20 @@ The `kernel_param` targets a single kernel, and currently, only one instance can
...
+that use the standard C++ interfaces and device execution policies.
+
+It will affect which data-passing mechanisms,
+as described in [Abstract Algorithm Signature](abstract-algorithm-signature) section, are allowed.
+In particular, it should be investigated if supporting these entities is necessary and
+whether they can be handled by an additional thin layer between
+the algorithms with standard interfaces and Kernel Templates:
+- `oneapi::dpl::begin` and `oneapi::dpl::end`, which mimic iterators to pass `sycl::buffer` to algorithms accepting iterators.
+- Iterators to a host-allocated `std::vector` as described in
+  [Pass Data to Algorithms > Use std::vector](https://github.com/uxlfoundation/oneDPL/blob/release/2022.9.0/documentation/library_guide/parallel_api/pass_data_algorithms.rst#use-stdvector)
```
akukanov: At glance, it does not seem necessary for KT, even if it serves as the implementation backend for the "regular" algorithms. The necessary data copying/transferring to the device can and likely should be performed before KT is used.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -309,10 +309,37 @@ as described in [Abstract Algorithm Signature](abstract-algorithm-signature) sec
...
+   which mimic iterators to pass `sycl::buffer` to iterator-based algorithms.
 - Iterators to a host-allocated `std::vector` as described in
   [Pass Data to Algorithms > Use std::vector](https://github.com/uxlfoundation/oneDPL/blob/release/2022.9.0/documentation/library_guide/parallel_api/pass_data_algorithms.rst#use-stdvector)
 
+An additional question is how to handle hardware requirements that cannot be queried directly
+through standard SYCL mechanisms. For example, selecting a more performant kernel template
+might depend on forward-progress guarantees between work-groups.
+While oneAPI DPC++ supports querying this via
+the [Root Group](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_root_group.asciidoc)
+experimental extension, other SYCL implementations may lack a corresponding feature.
```
akukanov: That's not a problem I think - at least not for the use of KT as a backend. Implementations with HW requirements like forward progress are supposed to have a fallback path - either another KT algorithm or a "standard SYCL" implementation. If we cannot prove that the requirements are met for any reason, whether HW or SYCL implementation related, the fallback should be used.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -309,10 +309,37 @@ as described in [Abstract Algorithm Signature](abstract-algorithm-signature) sec
...
+Furthermore, even if a device does not explicitly support cooperative kernel launch,
+it may still provide sufficient guarantees in limited cases
+such as chain-like synchronization between work-groups.
+A possible solution is to allow users to define a macro
+indicating that the required hardware capability is present.
```
akukanov: One of the motivations for KT is to give users choice if the standard API does not meet their needs. If their HW provides "sufficient" guarantees, KT can be used directly.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -309,10 +309,37 @@ as described in [Abstract Algorithm Signature](abstract-algorithm-signature) sec
...
+
+### Encoding Requirements
```
akukanov: Honestly, I do not quite understand the "Encoding Requirements" subsection. It seems as yet another way to mention (and generalize?) the progress guarantee problem with yet another macro-based solution proposed. As written, I find it too abstract to be useful.

Please try to improve it, so that the specifics of the concern and/or the idea is clear and hopefully is different from what has already been discussed. And ideally, avoid this "let's add an opt-in macro" approach to solve - it has too many drawbacks, the main ones being lack of granularity and the risk of multiple function definition at linkage (aka ODR violation).

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,354 @@
...
+Pros:
+- Avoids potential name clashes caused by `using` directives.
+Cons:
+- Does not replicate the hierarchical structure provided by namespaces,
+  unless namespeces are built into the tag name.
+Other aspects:
+- Is an additional parameter, possibly with its own namespaces,
+  preferred over adding namespaces directly to an algorithm?
+- Specializations may not share the same set of arguments,
+  which reduces the value of tag-based dispatching.
```
akukanov: The lists need to be separated with empty lines for proper rendering by GitHub.

rfcs/experimental/kernel_templates/kernel_configuration/README.md
mmichel11: Thanks, I think my initial comment was confusing about the alternative approach.

Consider the case of something like `copy`:
```cpp
template <typename InputIt, typename SizeType, typename OutputIt, typename KernelParam>
sycl:::event copy(InputIt input, SizeType n, OutputIt output, KernelParam param = {});
```
The actual number of elements is only known at runtime, but if the user provides a `uint16_t` as the `SizeType` for instance, then we can safely use that smaller type for indexing in our kernel.

rfcs/experimental/kernel_templates/README.md
danhoeflinger: This came from my suggestion.  Perhaps it is not worth the downsides.  But I will list a couple of advantages:
1) It can allow users to be less conservative if they so choose (For instance, hardware which is does not explicitly have forward progress guarantees but does in practice work without hangs)
2) It can avoid having to do unnecessary JIT compilation for fallbacks which wont be used
3) It allows users to not need to rewrite their code to use KT to take advantage of them

However, with the issues you bring up, perhaps these are not worthwhile, and users who want 1 + 2 can just use the explicit KT API.

akukanov: Looking closer at these sentences:
> Some requirements, for example forward-progress guarantees between work-groups, are currently only described in documentation. Encoding such requirements programmatically would improve visibility and help avoid unintended usage. 

I feel like maybe @dmitriy-sobolev tries to ask the question: For things like forward progress guarantees and ESIMD support expected by a certain function, can we enforce the HW/SW requirements better than just stating those in the docs?

We can discuss possible views on that. Seemingly the further text in the subsection, as @danhoeflinger mentions, is about some opt-in mechanisms to allow/request using these "demanding" implementations in the backend if those would not be chosen otherwise. Though that's not "avoiding unintended usage" but rather enforcing intended usage. I tend to think it's not worth it, both because of the downsides and because of likely niche applicability. Anyway, it is better covered in the "KT as a backend" subsection.

Another possible interpretation is "Can the special requirements be enforced programmatically when KT algorithms are used directly?" I think it depends on how generic and portable we want these algorithms to be. And the current mechanism to control that is the namespace hierarchy. For example, the radix sort implementation for platforms that support ESIMD is placed in the `kt::gpu::esimd` namespace. while a more portable implementation can be added to the `kt::gpu` namespace. 

We can change the API, e.g., use tags or additional options in `kernel_config`, or modify algorithm names (that could be the preference for single work group implementation for example); but in essence it's all the same - there is already compile-time opt-in for users to indicate exactly what they want if choice is available. I do not see how we can do better requirement enforcement / prevention of unintended usage in the compile time; but maybe I am missing something.

rfcs/experimental/kernel_templates/kernel_configuration/README.md
dmitriy-sobolev: I've added this note. However, I renamed `copy` to `copy_n` (I hope you do not mind) and added a comment about range-based inputs.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: I've removed this paragraph in the KT as a backed sub-section, and updated the main description on data passing mechanisms. See https://github.com/uxlfoundation/oneDPL/pull/2301/commits/f968e5817cab09559d3bfd3b92a2aa60b35daa78.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: > If we cannot prove that the requirements are met for any reason, whether HW or SYCL implementation related, the fallback should be used.

This paragraph describes a further step in case we cannot prove - give a user opportunity to explicitly tell what to use under the hood of the "regular" algorithms. However, if the user needs it, he/she can update use KT directly, I think. I am removing this paragraph.

Upd. I see that the you denoted the same idea below.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: I've removed that section as not worth considering further. Thank you both for the input.

rfcs/experimental/kernel_templates/README.md
dmitriy-sobolev: Could you check  "Algorithms to Implement" section added in https://github.com/uxlfoundation/oneDPL/pull/2301/commits/47247afc90b4152162b449c9ee6077ae46bea82e?

akukanov (approved):
Now it is a good description of the KT approach as well as various questions & concerns.

rfcs/experimental/kernel_templates/README.md
```diff
@@ -104,9 +104,17 @@ sycl::event kernel_template (
...
+any necessary data transfers must be completed in advance or managed automatically.
+Supported data storage types include:
+- [SYCL buffers](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:buffers).
+- Unified Shared Memory ([USM](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:usm)).
+- `std::vector` with `sycl::usm_allocator`.
+
+Supported data passing mechanisms are listed in
+the [documentation on passing data](https://uxlfoundation.github.io/oneDPL/parallel_api/pass_data_algorithms.html#pass-data-to-algorithms).
+The `oneapi::dpl::begin` and `oneapi::dpl::end` helper functions are not supported;
+however, `sycl::buffer` objects may be passed directly.
```
danhoeflinger, line 117: To make it clear I think we need one more disclaimer to the info in the link, assuming I understand the intent appropriately:

```suggestion
Supported data passing mechanisms are listed in
the [documentation on passing data](https://uxlfoundation.github.io/oneDPL/parallel_api/pass_data_algorithms.html#pass-data-to-algorithms).
The `oneapi::dpl::begin` and `oneapi::dpl::end` helper functions are not supported;
however, `sycl::buffer` objects may be passed directly. `std::vector` with host-allocated data is not supported.
```

dmitriy-sobolev: You are right, thanks for finding that. I am going to apply a slightly corrected suggestion:

```suggestion
Supported data passing mechanisms are listed in
the [documentation on passing data](https://uxlfoundation.github.io/oneDPL/parallel_api/pass_data_algorithms.html#pass-data-to-algorithms).
The `oneapi::dpl::begin` and `oneapi::dpl::end` helper functions are not supported;
however, `sycl::buffer` objects may be passed directly.
`std::vector` with host-allocated data is not supported.
```

dmitriy-sobolev: I've applied another suggestion after an offline conversation.

danhoeflinger (approved):
LGTM now, we discussed offline another optional potential open question about flavors of API to support (iterator, buffer, range, etc.).  If you choose to add that, I can reapprove.

rfcs/experimental/kernel_templates/kernel_configuration/README.md
```diff
@@ -0,0 +1,69 @@
...
+Another possible solution is to let the user specify the number of elements as a parameter,
+and then deduce the index type from the type of this parameter. For example:
+
+```c++
+template <typename InputIt, typename SizeType, typename OutputIt, typename KernelParam>
+sycl::event copy_n(InputIt input, SizeType n, OutputIt output, KernelParam param);
+```
+
+However, this method is not compatible with range-based inputs,
+where the number of elements is determined implicitly from the range itself.
```
mmichel11, line 69: One more open after thinking about algorithmic parameterization which I am curious to hear your thoughts about:

Most of our algorithms require significant global memory traffic. GPU hardware contains vector load / store instructions for memory operations which can result in:

1. Fewer total memory requests
2. Lower register pressure as we only compute one address per vector request
3. Potentially a more efficient data layout for algorithmic processing (e.g. a reduce operation could load a vec4 of its type and do work-item local reductions before doing sub-group followed by work-group level reductions)

In oneDPL mainline implementations, we have only employed vector memory operations in very limited scenarios as it can be platform specific. Maybe exposing some kernel parameter such as `VectorWidth` could allow users more freedom to improve performance at this low level. Alternatively, it could be an algorithm parameter and decided on a per algorithm basis.

mmichel11 (commented):
After my latest read through, I have just one question about a potential new open. I am curious to hear your thoughts.

Otherwise, everything LGTM!

dmitriy-sobolev:
> LGTM now, we discussed offline another optional potential open question about flavors of API to support (iterator, buffer, range, etc.). If you choose to add that, I can reapprove.

I've added it in [51517f1](https://github.com/uxlfoundation/oneDPL/commit/51517f1a48684b369aa899ad60dc335736bd8d5a).

rfcs/experimental/kernel_templates/kernel_configuration/README.md
dmitriy-sobolev: It's a complicated question, thanks for raising it. Vectorization is needed to reach peak device memory throughput and improve instruction throughput.

I do not have a complete picture how it would work out, though. It requires considering how `data_per_workitem` would coexist with the proposed `vector_width` for each major algorithm type such as transforming, reducing, scanning and sorting.

Anyway, There are 3 possible modes of load/store operations:
1) Vector.
2) Stridden (sub-group, work-group, device-wide).
3) Combined.

(2) mode is simple - it needs only `data_per_workitem`. (1) and (3) are more complicated, but I think `data_per_workitem` must be equal to `vector_width` in (1) mode and multiple of it in (3) mode, therefore `data_per_workitem` still describes the workload volume per work-item as it is.

As for the processing, e.g. sequentially reducing elements by a single work-item, I think it is orthogonal - you can do vectorized loads and then process data in a stridden manner, and then do vectorized stores again.

I will add this question into the RFC.

danhoeflinger (approved):
LGTM, after additions

rfcs/experimental/kernel_templates/kernel_configuration/README.md
dmitriy-sobolev: Added in [d5b12c1](https://github.com/uxlfoundation/oneDPL/commit/d5b12c100b78ff8eabeebfe9cd3de366dc209f6c)

rfcs/experimental/kernel_templates/README.md
```diff
@@ -0,0 +1,397 @@
...
+  helper functions to support `sycl::buffer` with iterators such as `oneapi::dpl::zip_iterator`?
+- Iterator parameter:
+  Should algorithms accept (`first`, `last`) or (`first`, `size`)?
+  The latter avoids passing an explicit index type, see
+  [Kernel Configuration](kernel_configuration/README.md#indexing-type), but
+  it is not consistent with the standard C++ algorithm interfaces.
+- Range support:
+  Should ranges be supported in the interface, or are iterators sufficient?
+
+### Vectorizing/Widenning of Load and Store Operations
```
mmichel11:
```suggestion
### Vectorizing/Widening of Load and Store Operations
```

rfcs/experimental/kernel_templates/kernel_configuration/README.md
mmichel11: Looks good!

mmichel11 (approved):
The latest state LGTM

PR #2303, opened 2025-06-10
Skip example target in codespell

danhoeflinger:
Skip the newly added book for an example in codespell, adding allowed word.

Fixing a typo.


examples/dynamic_selection/sepia-filter-ds/src/1_sepia_sycl.cpp
```diff
@@ -214,7 +214,7 @@ int main(int argc, char **argv) {
           accessor image_exp_acc(out, h, write_only);
   
           // This is the simplest form sycl::handler::parallel_for -
-          // - it specifies "flat" 1D ND range(num_pixels), runtime will select
+          // - it specifies "flat" 1D n-D range(num_pixels), runtime will select
```
SergeyKopienko: Are you sure we should change this?

danhoeflinger: I mean... its just a comment in the code.  Alternatively I can add `ND` to codespell ignore words, it don't think its that important.

SergeyKopienko: In the book about SYCL and oneDPL I saw "Explicit ND-Range Kernels"
So I would prefer to ignore "ND" in comments.

SergeyKopienko (dismissed):
LGTM

examples/dynamic_selection/sepia-filter-ds/src/1_sepia_sycl.cpp
danhoeflinger: Sure, that's fine, done.

danhoeflinger: Heh, it seems it did not work...  one moment.

SergeyKopienko (approved):
LGTM

PR #2191, opened 2025-04-16
Added dynamic_selection, maxloc_reductions and pSTL offload samples

jimmytwei:
Added three code samples from the oneAPI-samples repo.

[dynamic_selection](https://github.com/oneapi-src/oneAPI-samples/tree/main/Libraries/oneDPL/dynamic_selection)
[maxloc_reductions](https://github.com/oneapi-src/oneAPI-samples/tree/main/Libraries/oneDPL/maxloc_reductions)
[pSTL_offload](https://github.com/oneapi-src/oneAPI-samples/tree/main/Libraries/oneDPL/pSTL_offload)

Copied @timmiesmith

PR #2305, opened 2025-06-10
Fixing new examples to be built and tested in CI

danhoeflinger:
This fixes the build of the newly added examples in #2191, and attempts to make them match our convention.

Some minor changes were required to get them working for windows and linux (they are built and run in our per-commit ci with dpcpp backend).

They do not have any assertions checking the results in the tests themselves, but they run without error.  A future PR could address checking correctness if someone finds that a priority.

This PR also includes a change to trigger the tests from changes to the example directory.

We also remove visual studio solution files, and makefiles in favor of cmake files which work on both windows and linux.


examples/dynamic_selection/nstream/CMakeLists.txt
```diff
@@ -1,27 +1,14 @@
...
-    CMAKE_FORCE_CXX_COMPILER (icx-cl IntelDPCPP)
-    include (Platform/Windows-Clang)
-endif()
-
-project(nstreams_ds CXX)
-
-set(CMAKE_CXX_FLAGS "-fsycl -Wall")
+set(CMAKE_CXX_FLAGS "-Wall")
 
 add_executable(1_nstreams_sycl ./src/1_nstreams_sycl.cpp)
```
SergeyKopienko: May be  better to move these "variant no" into postfix in name?

examples/convex_hull/src/main.cpp
```diff
@@ -141,10 +141,10 @@ int main(int argc, char *argv[]) {
 
   using ms = std::chrono::milliseconds;
 
-  auto tm_start = std::chrono::high_resolution_clock::now();
+  auto tm_start = std::chrono::steady_clock::now();
```
SergeyKopienko: Does this change really needed?

examples/dynamic_selection/nstream/README.md
```diff
@@ -87,7 +87,7 @@ Perform the following steps:
    ```
    $ mkdir build
    $ cd build
-   $ cmake ..
+   $ cmake -DCMAKE_CXX_COMPILER=icpx ..
```
SergeyKopienko: We nowhere have this in our tests. So should we have this here?
Is it correct for windows and etc. platforms?

examples/convex_hull/src/main.cpp
danhoeflinger: Yes, with MSVC it encounters build errors without it.  I dont think we gain much of anything with `high_resolution_clock` , anyway, it is generally an alias for `steady_clock` or `system_clock`. For these examples, even if we lose some precision, I don't think it matters.

examples/dynamic_selection/nstream/README.md
danhoeflinger: I guess it isn't right to add this as it does not apply to both windows and linux.  I suppose leaving it out just allows users to select their compiler (though it should support sycl).  I'll remove it.

examples/dynamic_selection/nstream/CMakeLists.txt
danhoeflinger: I suppose we could do that, or remove the number entirely. I just tried to keep these as close as possible to the original intentions, since I am not the author of the examples.

danhoeflinger: I removed the numbers

examples/convex_hull/src/main.cpp
SergeyKopienko: http://www.en.cppreference.com/w/cpp/chrono/high_resolution_clock.html - the usage looks correct.
May I take a look to these build errors?

danhoeflinger: https://github.com/uxlfoundation/oneDPL/actions/runs/15567901779/job/43836420163
This was from the commit prior to the fix.

SergeyKopienko: Let me copy this build error:
```
oneDPL\examples\dynamic_selection\nstream\src\1_nstreams_sycl.cpp(72,29): error: no member named 'high_resolution_clock' in namespace 'std::chrono'
 |   auto begin = std::chrono::high_resolution_clock::now();
 |                ~~~~~~~~~~~~~^
 ```

Comment (danhoeflinger) [reply examples/convex_hull/src/main.cpp]: You raise a good point... I missed that this wasn't one of the new tests... 

I just replaced all instances when I found the build error for nstreams, but there must be something I'm missing unique to that example.  I'll fix it, and revert this.

Comment (danhoeflinger) [reply examples/convex_hull/src/main.cpp]: Good catch, I've added the missing include and reverted that change, it should work now.

**Comment by SergeyKopienko** [comment]
@danhoeflinger, one more overall comment: what about formatting source code of examples?
For me they are looks now very inconvenient...

**Comment by danhoeflinger** [comment]
> @danhoeflinger, one more overall comment: what about formatting source code of examples? For me they are looks now very inconvenient...

Ok, why not... I'll apply clang format to the source files.

File: examples/dynamic_selection/sepia-filter-ds/sepia-filter.sln
Comment (danhoeflinger) [inline examples/dynamic_selection/sepia-filter-ds/sepia-filter.sln:1]: @timmiesmith @jimmytwei 
Are we OK to remove the visual studio solution files, vscode .json files and [GNU]makefiles in favor of a more consistent cmake approach to match the existing samples?
Are they needed to reincorporate as samples for any sort of release?

Comment (jimmytwei) [reply examples/dynamic_selection/sepia-filter-ds/sepia-filter.sln:1]: @danhoeflinger I'm fine with removing all of the VS proj and sln files. I've also removed them when upstreaming oneTBB samples to the oneTBB repo.

**Comment by jimmytwei** [comment]
> @jimmytwei @timmiesmith Similar question regarding formatting. I've applied clang format to the examples to match oneDPL's conventions, but this can easily be reverted if a specific formatting is required.

Thanks for applying clang format. Also did that for oneTBB samples.

File: examples/dynamic_selection/sepia-filter-ds/src/sepia_policies.cpp [hunk 4]
```diff
@@ -0,0 +1,389 @@
...
+                    // the lambda parameter of the parallel_for is the kernel, which
+                    // actually executes on device
+                    h.parallel_for(range<1>(num_pixels), [=](auto i) {
+                        ApplyFilter(image_acc.get_pointer(), image_exp_acc.get_pointer(), i);
+                    });
+                });
+            };
+            ex::submit_and_wait(p, f1, num_pixels);
+        }
+        catch (sycl::exception e)
```
SergeyKopienko:
```suggestion
        catch (const sycl::exception& e)
```

examples/dynamic_selection/sepia-filter-ds/src/sepia_sycl.cpp
```diff
@@ -0,0 +1,265 @@
...
+                 //   local size
+                 // - kernel lambda accepts single sycl::id argument, which has very
+                 //   limited API; see the spec for more complex forms
+                 // the lambda parameter of the parallel_for is the kernel, which
+                 // actually executes on device
+                 h.parallel_for(range<1>(num_pixels),
+                                [=](auto i) { ApplyFilter(image_acc.get_pointer(), image_exp_acc.get_pointer(), i); });
+             }).wait();
+        }
+        catch (sycl::exception e)
```
SergeyKopienko:
```suggestion
        catch (const sycl::exception& e)
```

examples/dynamic_selection/sepia-filter-ds/src/sepia_policies.cpp
danhoeflinger: I would prefer not to make many content changes which are not necessary for build & pass in this PR.  Just because the code was moved / formatted doesn't make this PR responsible for the content.
 
That being said, as long as it is only these 2, I can just apply them.  Beyond that, I'd ask that you make a separate PR (for changes to the content of the code).

examples/dynamic_selection/nstream/CMakeLists.txt
```diff
@@ -1,27 +1,14 @@
...
-    include (Platform/Windows-Clang)
-endif()
+set(CMAKE_CXX_FLAGS "-Wall")
 
-project(nstreams_ds CXX)
+add_executable(nstreams_sycl ./src/nstreams_sycl.cpp)
+add_executable(nstreams_policies ./src/nstreams_policies.cpp)
 
-set(CMAKE_CXX_FLAGS "-fsycl -Wall")
+target_link_libraries(nstreams_sycl tbb)
```
mmichel11: Not a big deal to me but could you explain why we need to link tbb? Could we set the oneDPL backend to `dpcpp_only` to only require SYCL?

danhoeflinger: That is fair, adjusted. Seems like it works, lets see if the CI agrees.

SergeyKopienko (approved):
LGTM

PR #2307, opened 2025-06-10
Example CMake Fixes; Skip example without CMakeLists.txt

danhoeflinger:
Fixes cmake to allow build for PSTL offlead examples (no tests running yet). For the other new examples #2191, skipping them in the CI for now with no CMakeLists.txt in the top level dir.  This allows development to continue for active PRs, using the CI without further interruption. 
We will follow up with another PR to fix for the new examples so they all build, run and pass.

### DIFF
--- a/.github/workflows/ci-testing.yml
+++ b/.github/workflows/ci-testing.yml
@@ -183,15 +183,18 @@ jobs:
 
           if [[ "${{ matrix.backend }}" == "dpcpp" ]]; then
             # build and run the examples
+            example_dirs=()
             for example_dir in ${GITHUB_WORKSPACE}/examples/*; do
-            # Skip the example if it is not a directory
-            if [[ -d "$example_dir" ]]; then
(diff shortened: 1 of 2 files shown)


danhoeflinger:
> I found 4 files which looks like for me as build instructions for some IDE but probably should be removed:
> 
> * `examples/dynamic_selection/nstream/sample.json`
> * `examples/dynamic_selection/sepia-filter-ds/sample.json`
> * `examples/maxloc_reductions/sample.json`
> * `examples/pSTL_offload/sample.json`
> 
> @danhoeflinger, what do you think?

I noticed those too, there may be more stuff like this since these examples did not adhere to our existing conventions. I can remove them here I think.  We can try to do a more thorough examination in the follow up PR.

SergeyKopienko (approved):
LGTM

danhoeflinger:
I'll take care of the json files in the follow up.

PR #2308, opened 2025-06-11
Perfect forwarding of execution policy in tests

SergeyKopienko:
This PR updates test cases to perfectly forward execution policies using `std::forward` and fixes spelling mistakes in comments.

- Introduce `std::forward<Policy>(exec)` (or `ExecutionPolicy`) in various algorithm test calls to preserve value category.
- Corrected typos in comments (“itertors” → “iterators”).


dmitriy-sobolev (approved):
LGTM.

I've checked that introduced `std::forward` are used with forwarding references, and that they are not used multiple times for the same object.

I haven't checked if all possible places are covered, but I trust Sergey here given the scope and coverage of the PR.

PR #2311, opened 2025-06-16
Support `EXPECT_EQ` call for iterators

SergeyKopienko:
This PR enhances error logging in iterator comparisons by introducing streamability detection and structured log output.

- Added a compile-time trait to detect if a type can be streamed (in the namespace `TestUtils`).
- Introduced `TagExpected`/`TagActual` and `log_value` helpers for consistent logging (in the namespace `TestUtils`).
- Updated `expect_equal_val` and `expect_equal` to use the new logging helpers.

So now we can write in the test code something like
```C++
    std::vector<int> data;
    // ...
    auto it = data.begin();
    // ...
    EXPECT_EQ(data.begin(), it, "Error message");
```
Previously we would have had the compile errors in this use-case because iterator variables doesn't supports `operator<<` at all.

### DIFF
--- a/test/support/utils.h
+++ b/test/support/utils.h
@@ -149,25 +149,21 @@ std::string log_value_title(TagActual)
     return " got ";
 }
 
-template <typename Tag, typename TValue>
-std::string log_value(Tag, const TValue& value, bool bCommaNeeded)
+template <typename TStream, typename Tag, typename TValue>
+ void log_value(TStream& os, Tag, const TValue& value, bool bCommaNeeded)
 {
-    std::stringstream outstr;
-
     if (bCommaNeeded)
-        outstr << ",";
-    outstr << log_value_title(Tag{});
+        os << ",";
+    os << log_value_title(Tag{});
 
-    if constexpr (IsSupportStreamOutput<TValue, decltype(outstr)>::value)
+    if constexpr (IsSupportStreamOutput<TValue, decltype(os)>::value)
     {
-        outstr << value;
+        os << value;
     }
     else
     {
-        outstr << "(unable to log value)";
+        os << "(unable to log value)";
     }
-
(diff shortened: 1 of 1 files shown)


test/support/utils.h
```diff
@@ -123,6 +130,52 @@ is_equal_val(const T1& val1, const T2& val2)
     }
 }
 
+template <typename T, typename TOutputStream, typename = void>
+struct IsSupportStreamOutput : std::false_type
```
mmichel11: nitpick: Maybe a name like `IsStreamInsertable` or `IsOutputStreamable` is a bit clearer.

test/support/utils.h
```diff
@@ -123,6 +130,52 @@ is_equal_val(const T1& val1, const T2& val2)
...
+ void log_value(TStream& os, Tag, const TValue& value, bool bCommaNeeded)
+{
+    if (bCommaNeeded)
+        os << ",";
+    os << log_value_title(Tag{});
+
+    if constexpr (IsSupportStreamOutput<TValue, decltype(os)>::value)
+    {
+        if constexpr (std::is_pointer_v<TValue>)
+            os << "0x";
```
mmichel11: When I tried this out with a pointer, I got something like `... 0x0x15061eb0, got 0x0x15061eb4`. I do not know if the output for pointers is standardized or implementation defined, but it seems like libstdc++ at least already has the hexadecimal prefix.

test/support/utils.h
SergeyKopienko: Fixed: renamed to `IsOutputStreamable`

test/support/utils.h
SergeyKopienko: Fixed: "0x" has been removed

PR #2312, opened 2025-06-16
[oneDPL] A fix for oneDPL internal tuple and for oneDPL C++17 zip_view

MikeDvorskiy:
[oneDPL] A fix for oneDPL C++17 zip_view.

The following trivial example doesn't work:

```
    int a[10];
    std::ranges::subrange r1(a, a + 10);
    auto zip_view = make_zip_view(r1);

```

The PR suggests a fix for that.


dmitriy-sobolev (approved):
LGTM

PR #2313, opened 2025-06-16
Log exceptions and device name in `TestUtils::get_dpcpp_test_policy()' function

SergeyKopienko:


test/support/utils_invoke.h
```diff
@@ -135,7 +144,27 @@ get_dpcpp_test_policy()
         get_test_queue();
 #    endif // TEST_USE_PREDEFINED_POLICIES
 
-    return TestUtils::make_new_policy<_NewKernelName>(__arg);
+    try
+    {
+        auto policy = TestUtils::make_new_policy<_NewKernelName>(__arg);
```
dmitriy-sobolev, line 145: I see no benefits in guarding make_new_policy with try-catch block. It's just updates the name of a kernel, which is trivial. Issues with creation of a queue are covered in `get_test_queue`.

test/support/utils_sycl.h
```diff
@@ -92,13 +93,31 @@ inline auto default_selector =
 #    endif
 #endif     // ONEDPL_FPGA_DEVICE
 
+#if _ONEDPL_DEBUG_SYCL
+static std::once_flag device_name_in_get_test_queue_logged;
```
dmitriy-sobolev: Let's use `inline` instead of `static` to have only one counter if a test has multiple TUs.

test/support/utils_sycl.h
```diff
@@ -92,13 +93,31 @@ inline auto default_selector =
...
 sycl::queue get_test_queue()
 {
     try
     {
         // create the queue with custom asynchronous exceptions handler
         static sycl::queue my_queue(default_selector, async_handler);
+
+#if _ONEDPL_DEBUG_SYCL
+        std::call_once(device_name_in_get_test_queue_logged, [&]() {
+            log_device_name(std::cout, my_queue);
```
dmitriy-sobolev, line 116: Let's remove `log_device_name` and output the information right here. That's a trivial one-line function.

test/support/utils_sycl.h
SergeyKopienko: The goal was to log device name here only once for application, but not for translation unit.

test/support/utils_invoke.h
SergeyKopienko: It's not true, I can explain this moment.

test/support/utils_sycl.h
SergeyKopienko: >>> That's a trivial one-line function.

- and it's the reason why I declared this function as `inline`.

test/support/utils_sycl.h
SergeyKopienko: Partially fixed.

test/support/utils_invoke.h
SergeyKopienko: Discussed offline.

dmitriy-sobolev: This line can emit an exception during copying of a policy and the queue under the hood.

PR #2314, opened 2025-06-18
Avoid implicit `class`/`struct` forward declarations in Kernel names

SergeyKopienko:
### Initial problem
During implementation of https://github.com/uxlfoundation/oneDPL/pull/2102 some one more error about not-unique Kernel name has been found.
This example from @AlexeySachkov demonstrate it:
```C++
#include <sycl/sycl.hpp>

template<typename KernelName>
void submitter() {
  sycl::queue q;
  q.single_task<KernelName>([](){});
}

#ifdef WORKAROUND_BUG
class copy;
#endif

template<typename Policy>
void test() {
#ifdef WORKAROUND_BUG
  submitter<copy>();
#else
  submitter<class copy>();
  // Causes:
  //
  // /tmp/n-header-cfa343.h:86:20: error: redefinition of 'KernelInfo<copy>'
  //   86 | template <> struct KernelInfo<copy> {
  //      |                    ^~~~~~~~~~~~~~~~
  // /tmp/n-header-cfa343.h:39:20: note: previous definition is here
  //   39 | template <> struct KernelInfo<copy> {
  //      |                    ^
#endif
}

int main() {
  test<int>();
  test<float>();
}
``` 
This example has no errors if `WORKAROUND_BUG` is defined (we should switch off unnamed lambda in compiler).

Also make sense to note that from C++ point of view in this example (and in our tests too) we make forward declaration of `class copy` inside function `template<typename Policy> void test()` so we will have the separate declaration of `class copy` in each instantiation of function `template<typename Policy> void test()`.

But due special aspects of our `icpx` compiler we will have more then one forward declarations of `class copy` in internal header file in global scope like
```C++
// This is auto-generated SYCL integration header.

#include <sycl/detail/defines_elementary.hpp>
#include <sycl/detail/kernel_desc.hpp>
#include <sycl/ext/oneapi/experimental/free_function_traits.hpp>

#ifndef SYCL_LANGUAGE_VERSION
#define SYCL_LANGUAGE_VERSION 202001
#endif //SYCL_LANGUAGE_VERSION

struct copy;
// ...
struct copy;
struct copy;
struct copy;
// ...
namespace sycl {
inline namespace _V1 {
namespace detail {
// ...
} // namespace detail
} // namespace _V1
} // namespace sycl
```
To avoid this behavior we should make forward declaration of `class copy` in global scope of our tests (like we did in internal headers of our `oneDPL` library).

### Solution (in tests)
So this PR updates kernel name usage to avoid implicit forward declarations by removing the explicit `class` (`struct`) keyword from kernel test and policy instantiations.  So we move all these declarations into global scope.
- Removed redundant "class" keyword from kernel_test instantiation calls in multiple test files.  
- Applied similar changes in policy creation and unique kernel name invocations to improve consistency and clarity.

The changes from this PR later will be needed in https://github.com/uxlfoundation/oneDPL/pull/2102 but I prefer to make them separately in this PR for simplify review process.

### DIFF
--- a/test/general/test_policies.pass.cpp
+++ b/test/general/test_policies.pass.cpp
@@ -23,6 +23,13 @@
 #include <iostream>
 #include <vector>
 
+class Kernel_0;
+class Kernel_11; class Kernel_12; class Kernel_13; class Kernel_14; class Kernel_15; class Kernel_16;
+class Kernel_1;
+class Kernel_21; class Kernel_22; class Kernel_23; class Kernel_24; class Kernel_25;
+class Kernel_31; class Kernel_32; class Kernel_33; class Kernel_34; class Kernel_35;
+class Kernel_41; class Kernel_42; class Kernel_43; class Kernel_44; class Kernel_45;
+
 #if TEST_DPCPP_BACKEND_PRESENT
 
 template<typename Policy>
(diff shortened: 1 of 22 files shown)


test/general/test_policies.pass.cpp
```diff
@@ -23,6 +23,13 @@
 #include <iostream>
 #include <vector>
 
+class Kernel_0;
+class Kernel_11; class Kernel_12; class Kernel_13; class Kernel_14; class Kernel_15; class Kernel_16;
+class Kernel_1;
+class Kernel_21; class Kernel_22; class Kernel_23; class Kernel_24; class Kernel_25;
+class Kernel_31; class Kernel_32; class Kernel_33; class Kernel_34; class Kernel_35;
+class Kernel_41; class Kernel_42; class Kernel_43; class Kernel_44; class Kernel_45;
```
dmitriy-sobolev: That would be more clean:

```
template <int N>
class Kernel; 
```

test/parallel_api/experimental/asynch.pass.cpp
```diff
@@ -30,6 +30,18 @@
...
+class Fill1;
+class ForEach1;
+class Transform1; class Transform2;
+class Reduce1; class Reduce2; class Reduce3;
+class Scan;
+class Sort;
+class Copy2a; class Copy2b;
+class Scan2a; class Scan2b;
+class Fill2a; class Fill2b;
+class async1; class async2; class async3;
```
dmitriy-sobolev: I think this is another place where we can use a single class template due to the large number of names.

test/general/test_policies.pass.cpp
SergeyKopienko: Fixed

test/parallel_api/experimental/asynch.pass.cpp
SergeyKopienko: Fixed

dmitriy-sobolev (approved):
LGTM

PR #2315, opened 2025-06-18
Reduce-then-scan path for SYCL scan-by-segment implementations

mmichel11:
This PR adds a path for `inclusive_scan_by_segment` and `exclusive_scan_by_segment` using our reduce-then-scan infrastructure to improve performance. Additionally, some reorganization of existing "fallback" implementations is required 

Summary of changes:
- `__parallel_scan_by_segment_reduce_then_scan` and associated reduce-then-scan building blocks are added to implement scan-by-segment. The implementation is similar but unique from reduce-by-segment.
- The high level scan-by-segment fallback has been re-implemented at the parallel pattern level to consolidate the location of SYCL backend implementations.
- `__segmented_scan_fun` and `__replace_if_fun` have been moved to our general utilities as they are shared between implementations in `pstl/hetero/dpcpp` and `internal` directories.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -2230,7 +2230,7 @@ __parallel_scan_by_segment_fallback(oneapi::dpl::__internal::__device_backend_ta
             oneapi::dpl::__internal::__device_backend_tag{}, std::forward<_ExecutionPolicy>(__exec),
             oneapi::dpl::__ranges::zip_view(std::forward<_Range2>(__values), __mask_view),
             oneapi::dpl::__ranges::zip_view(std::forward<_Range3>(__out_values), __mask_view), __n,
-            oneapi::dpl::__internal::__no_op{}, oneapi::dpl::unseq_backend::__no_init_value<_ScanInitType>{},
+            oneapi::dpl::identity{}, oneapi::dpl::unseq_backend::__no_init_value<_ScanInitType>{},
             oneapi::dpl::__internal::__segmented_scan_fun<_BinaryOperator, _FlagType, _BinaryOperator>{__binary_op},
             /*_Inclusive*/ std::true_type{})
             .wait();
@@ -2273,7 +2273,7 @@ __parallel_scan_by_segment_fallback(oneapi::dpl::__internal::__device_backend_ta
             oneapi::dpl::__internal::__device_backend_tag{}, std::forward<_ExecutionPolicy>(__exec),
             oneapi::dpl::__ranges::zip_view(__temp_view, __mask_view),
             oneapi::dpl::__ranges::zip_view(std::forward<_Range3>(__out_values), __mask_view), __n,
-            oneapi::dpl::__internal::__no_op{},
+            oneapi::dpl::identity{},
             oneapi::dpl::unseq_backend::__init_value<_ScanInitType>{
                 oneapi::dpl::__internal::make_tuple(__init.__value, _FlagType(1))},
             oneapi::dpl::__internal::__segmented_scan_fun<_BinaryOperator, _FlagType, _BinaryOperator>{__binary_op},
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -2259,7 +2259,7 @@ __parallel_scan_by_segment_fallback(oneapi::dpl::__internal::__device_backend_ta
                 oneapi::dpl::__ranges::all_view<_OutputType, __par_backend_hetero::access_mode::read_write>(
                     __temp.get_buffer(), 1, __n - 1);
             oneapi::dpl::__internal::__replace_if_fun<typename _InitType::__value_type, std::negate<_FlagType>>
-                __replace_fun(std::negate<_FlagType>(), __init.__value);
+                __replace_fun{std::negate<_FlagType>{}, __init.__value};
             using _ReplaceTransform = oneapi::dpl::__internal::__transform_functor<decltype(__replace_fun)>;
             _ReplaceTransform __tf{__replace_fun};
             __parallel_for(
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -2190,8 +2190,6 @@ __parallel_scan_by_segment_fallback(oneapi::dpl::__internal::__device_backend_ta
                                     /*has_known_identity*/ std::false_type)
 {
     using _FlagType = unsigned int;
-    using _ValueType = oneapi::dpl::__internal::__value_t<_Range2>;
-    using _OutputType = oneapi::dpl::__internal::__value_t<_Range3>;
 
     const auto __n = __keys.size();
 
@@ -2239,6 +2237,7 @@ __parallel_scan_by_segment_fallback(oneapi::dpl::__internal::__device_backend_ta
     }
     else
     {
+        using _OutputType = oneapi::dpl::__internal::__value_t<_Range3>;
         // shift input one to the right and initialize segments with init
         oneapi::dpl::__par_backend_hetero::__buffer<_OutputType> __temp(__n);
         {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -2127,7 +2127,8 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
 //------------------------------------------------------------------------
 template <typename _CustomName, bool __is_inclusive, typename _Range1, typename _Range2, typename _Range3,
           typename _BinaryPredicate, typename _BinaryOperator, typename _InitType>
-auto
+__future<sycl::event, __result_and_scratch_storage<
+                          oneapi::dpl::__internal::tuple<std::uint32_t, oneapi::dpl::__internal::__value_t<_Range2>>>>
 __parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, _Range2&& __values,
                                             _Range3&& __out_values, _BinaryPredicate __binary_pred,
                                             _BinaryOperator __binary_op, [[maybe_unused]] _InitType __init)
@@ -2159,15 +2160,17 @@ __parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, 
     }
     else
     {
-        // The init value is manually applied through the write functor in exclusive-scan-by-segment and we always pass __no_init_value.
-        // This is because init handling must occur on a per-segment basis and functions differently than the typical scan init.
+        // The init value is manually applied through the write functor in exclusive-scan-by-segment and we always pass
+        // __no_init_value to the transform scan call. This is because init handling must occur on a per-segment basis
+        // and functions differently than the typical scan init which is only applied once in a single location.
         _WrappedInitType __wrapped_init{{0, __init.__value}};
         using _WriteOp = __write_scan_by_seg<__is_inclusive, _WrappedInitType, _BinaryOperator>;
         return __parallel_transform_reduce_then_scan<sizeof(_PackedFlagValueType), _CustomName>(
             __q, __n,
             oneapi::dpl::__ranges::make_zip_view(std::forward<_Range1>(__keys), std::forward<_Range2>(__values)),
             std::forward<_Range3>(__out_values), _GenReduceInput{__binary_pred}, _ReduceOp{__binary_op},
-            _GenScanInput{}, _ScanInputTransform{}, _WriteOp{__wrapped_init, __binary_op}, __wrapped_init,
+            _GenScanInput{}, _ScanInputTransform{}, _WriteOp{__wrapped_init, __binary_op},
+            oneapi::dpl::unseq_backend::__no_init_value<_PackedFlagValueType>{},
             /*Inclusive*/ std::false_type{}, /*_IsUniquePattern=*/std::false_type{});
     }
 }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -2146,7 +2146,7 @@ __parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, 
     {
         oneapi::dpl::unseq_backend::__no_init_value<oneapi::dpl::__internal::tuple<_FlagType, _ValueType>>
             __wrapped_init;
-        using _WriteOp = __write_scan_by_seg<decltype(__wrapped_init), _BinaryOperator>;
+        using _WriteOp = __write_scan_by_seg<__is_inclusive, decltype(__wrapped_init), _BinaryOperator>;
         return __parallel_transform_reduce_then_scan<sizeof(oneapi::dpl::__internal::tuple<_FlagType, _ValueType>),
                                                      _CustomName>(
             __q, __n,
@@ -2161,7 +2161,7 @@ __parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, 
         // This is because init handling must occur on a per-segment basis and functions differently than the typical scan init.
         oneapi::dpl::unseq_backend::__init_value<oneapi::dpl::__internal::tuple<_FlagType, _ValueType>> __wrapped_init{
             {0, __init.__value}};
-        using _WriteOp = __write_scan_by_seg<decltype(__wrapped_init), _BinaryOperator>;
+        using _WriteOp = __write_scan_by_seg<__is_inclusive, decltype(__wrapped_init), _BinaryOperator>;
         return __parallel_transform_reduce_then_scan<sizeof(oneapi::dpl::__internal::tuple<_FlagType, _ValueType>),
                                                      _CustomName>(
             __q, __n,
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -2126,11 +2126,11 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
 // parallel_scan_by_segment - sync pattern
 //------------------------------------------------------------------------
 template <typename _CustomName, bool __is_inclusive, typename _Range1, typename _Range2, typename _Range3,
-          typename _BinaryPredicate, typename _BinaryOperator, typename _T>
+          typename _BinaryPredicate, typename _BinaryOperator, typename _InitType>
 auto
 __parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, _Range2&& __values,
                                             _Range3&& __out_values, _BinaryPredicate __binary_pred,
-                                            _BinaryOperator __binary_op, [[maybe_unused]] _T __init, _T __identity)
+                                            _BinaryOperator __binary_op, [[maybe_unused]] _InitType __init)
 {
     using _GenReduceInput = __gen_red_by_seg_reduce_input<_BinaryPredicate>;
     using _ReduceOp = __scan_by_seg_op<_BinaryOperator>;
@@ -2138,8 +2138,6 @@ __parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, 
     using _ScanInputTransform = __get_zeroth_element;
     using _ValueType = oneapi::dpl::__internal::__value_t<_Range2>;
     std::size_t __n = __keys.size();
-    // __gen_red_by_seg_scan_input requires that __n > 1
-    assert(__n > 1);
     if constexpr (__is_inclusive)
     {
         oneapi::dpl::unseq_backend::__no_init_value<oneapi::dpl::__internal::tuple<std::size_t, _ValueType>>
@@ -2158,7 +2156,7 @@ __parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, 
         // The init value is manually applied through the write functor in exclusive-scan-by-segment and we always pass __no_init_value.
         // This is because init handling must occur on a per-segment basis and functions differently than the typical scan init.
         oneapi::dpl::unseq_backend::__init_value<oneapi::dpl::__internal::tuple<std::size_t, _ValueType>>
-            __wrapped_init{{0, __init}};
+            __wrapped_init{{0, __init.__value}};
         using _WriteOp = __write_scan_by_seg<decltype(__wrapped_init), _BinaryOperator>;
         return __parallel_transform_reduce_then_scan<sizeof(oneapi::dpl::__internal::tuple<std::size_t, _ValueType>),
                                                      _CustomName>(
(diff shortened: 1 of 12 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_scan_by_segment.h
```diff
@@ -360,20 +361,29 @@ struct __sycl_scan_by_segment_impl
     }
 };
 
-template <bool __is_inclusive, typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3,
-          typename _BinaryPredicate, typename _BinaryOperator, typename _T>
+template <typename _CustomName, bool __is_inclusive, typename _ExecutionPolicy, typename _Range1, typename _Range2,
```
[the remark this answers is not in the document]
mmichel11: I disagree. Given that the `false_type` specialization is already a large function, it is hard to follow if we did this.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -208,6 +208,59 @@ struct __write_red_by_seg
...
+        // through.
+        using _ConvertedTupleType =
+            typename oneapi::dpl::__internal::__get_tuple_type<std::decay_t<decltype(get<1>(get<0>(__v)))>,
+                                                               std::decay_t<decltype(__out_rng[__id])>>::__type;
+        __out_rng[__id] = static_cast<_ConvertedTupleType>(get<1>(get<0>(__v)));
+    }
+};
+
+// Exclusive scan by segment
+template <typename _InitType, typename _BinaryOp>
```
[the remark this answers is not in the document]
mmichel11: This is a good suggestion. I simplified `__write_scan_by_seg` in a similar way.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,200 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
 
+//------------------------------------------------------------------------
+// parallel_scan_by_segment - sync pattern
+//------------------------------------------------------------------------
+template <typename _CustomName, bool __is_inclusive, typename _Range1, typename _Range2, typename _Range3,
+          typename _BinaryPredicate, typename _BinaryOperator, typename _InitType>
+auto
+__parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, _Range2&& __values,
+                                            _Range3&& __out_values, _BinaryPredicate __binary_pred,
+                                            _BinaryOperator __binary_op, [[maybe_unused]] _InitType __init)
```
SergeyKopienko, line 2134: From my point of view this parameter `[[maybe_unused]] _InitType __init` looks a little bit strange.

What about to split this function for two :
1)
```C++
template <typename _CustomName, bool __is_inclusive, typename _Range1, typename _Range2, typename _Range3,
          typename _BinaryPredicate, typename _BinaryOperator, /* some check for __is_inclusive == false*/>
auto
__parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, _Range2&& __values,
                                            _Range3&& __out_values, _BinaryPredicate __binary_pred,
                                            _BinaryOperator __binary_op)
```

2)
```C++
template <typename _CustomName, bool __is_inclusive, typename _Range1, typename _Range2, typename _Range3,
          typename _BinaryPredicate, typename _BinaryOperator, typename _InitType, /* some check for __is_inclusive == true*/>
auto
__parallel_scan_by_segment_reduce_then_scan(sycl::queue& __q, _Range1&& __keys, _Range2&& __values,
                                            _Range3&& __out_values, _BinaryPredicate __binary_pred,
                                            _BinaryOperator __binary_op, _InitType __init)
```

Of course, it's only idea. So up to you.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,200 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
+{
+    using _GenReduceInput = __gen_scan_by_seg_reduce_input<_BinaryPredicate>;
+    using _ReduceOp = __scan_by_seg_op<_BinaryOperator>;
+    using _GenScanInput = __gen_scan_by_seg_scan_input<_BinaryPredicate>;
+    using _ScanInputTransform = __get_zeroth_element;
+    using _ValueType = oneapi::dpl::__internal::__value_t<_Range2>;
+    std::size_t __n = __keys.size();
```
```suggestion
    const std::size_t __n = __keys.size();
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,200 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
+          typename _Range3, typename _BinaryPredicate, typename _BinaryOperator, typename _InitType>
+void
+__parallel_scan_by_segment_fallback(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec,
+                                    _Range1&& __keys, _Range2&& __values, _Range3&& __out_values,
+                                    _BinaryPredicate __binary_pred, _BinaryOperator __binary_op, _InitType __init,
+                                    /*has_known_identity*/ std::false_type)
+{
+    using _FlagType = unsigned int;
+
+    const auto __n = __keys.size();
```
```suggestion
    const std::size_t __n = __keys.size();
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,200 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
+            oneapi::dpl::__ranges::all_view<_FlagType, __par_backend_hetero::access_mode::read_write>(
+                __mask.get_buffer(), 1, __n - 1);
+        using _NegateTransform =
+            oneapi::dpl::__internal::__transform_functor<oneapi::dpl::__internal::__not_pred<_BinaryPredicate>>;
+        _NegateTransform __tf{oneapi::dpl::__internal::__not_pred<_BinaryPredicate>(__binary_pred)};
+        auto __keys_shifted = oneapi::dpl::__ranges::drop_view_simple(__keys, 1);
+        __parallel_for(
+            oneapi::dpl::__internal::__device_backend_tag{},
+            oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__scan_by_seg_transform_wrapper1>(__exec),
+            unseq_backend::walk_n_vectors_or_scalars<_NegateTransform>(__tf, static_cast<std::size_t>(__n - 1)),
```
```suggestion
            unseq_backend::walk_n_vectors_or_scalars<_NegateTransform>(__tf, __n - 1),
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,200 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
+                oneapi::dpl::__ranges::all_view<_OutputType, __par_backend_hetero::access_mode::read_write>(
+                    __temp.get_buffer(), 1, __n - 1);
+            oneapi::dpl::__internal::__replace_if_fun<typename _InitType::__value_type, std::negate<_FlagType>>
+                __replace_fun(std::negate<_FlagType>(), __init.__value);
+            using _ReplaceTransform = oneapi::dpl::__internal::__transform_functor<decltype(__replace_fun)>;
+            _ReplaceTransform __tf{__replace_fun};
+            __parallel_for(
+                oneapi::dpl::__internal::__device_backend_tag{},
+                oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__scan_by_seg_transform_wrapper2>(__exec),
+                unseq_backend::walk_n_vectors_or_scalars<_ReplaceTransform>(__tf, static_cast<std::size_t>(__n - 1)),
```
```suggestion
                unseq_backend::walk_n_vectors_or_scalars<_ReplaceTransform>(__tf, __n - 1),
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,200 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
         oneapi::dpl::unseq_backend::__has_known_identity<_BinaryOperator, __val_type>{});
 }
 
+//------------------------------------------------------------------------
+// parallel_scan_by_segment - sync pattern
+//------------------------------------------------------------------------
+template <typename _CustomName, bool __is_inclusive, typename _Range1, typename _Range2, typename _Range3,
+          typename _BinaryPredicate, typename _BinaryOperator, typename _InitType>
+auto
```
SergeyKopienko: Let's describe return type.
As far as I understood, it's 
```suggestion
__future<sycl::event, __result_and_scratch_storage<typename _InitType::__value_type>>
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,200 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
+    std::size_t __n = __keys.size();
+    // TODO: A bool type may be used here for a smaller footprint in registers / temp storage but results in IGC crashes
+    // during JIT time. The same occurs for uint8_t and uint16_t. uint32_t is used as a workaround until the underlying
+    // issue is resolved.
+    using _FlagType = std::uint32_t;
+    if constexpr (__is_inclusive)
+    {
+        oneapi::dpl::unseq_backend::__no_init_value<oneapi::dpl::__internal::tuple<_FlagType, _ValueType>>
+            __wrapped_init;
+        using _WriteOp = __write_scan_by_seg<__is_inclusive, decltype(__wrapped_init), _BinaryOperator>;
```
SergeyKopienko: This type alias duplicates in two branches.
May be make sense to move it up?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,200 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
+            _GenScanInput{}, _ScanInputTransform{}, _WriteOp{__wrapped_init, __binary_op}, __wrapped_init,
+            /*Inclusive*/ std::true_type{}, /*_IsUniquePattern=*/std::false_type{});
+    }
+    else
+    {
+        // The init value is manually applied through the write functor in exclusive-scan-by-segment and we always pass __no_init_value.
+        // This is because init handling must occur on a per-segment basis and functions differently than the typical scan init.
+        oneapi::dpl::unseq_backend::__init_value<oneapi::dpl::__internal::tuple<_FlagType, _ValueType>> __wrapped_init{
+            {0, __init.__value}};
+        using _WriteOp = __write_scan_by_seg<__is_inclusive, decltype(__wrapped_init), _BinaryOperator>;
```
adamfidel: Just a preference thing, but it might be cleaner to define a type alias for `oneapi::dpl::unseq_backend::__init_value<oneapi::dpl::__internal::tuple<_FlagType, _ValueType>>` and use that both in the declaration above this line and in this template.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: It's a little more complicated as we need a `std::conditional_t`, but I agree and made this change.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Good point, I combined this with Sergey's suggestion to move it outside of the branch and added an alias for the tuple type since it is used elsewhere as well.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Done, the return type is
```cpp
__future<sycl::event, __result_and_scratch_storage<
                          oneapi::dpl::__internal::tuple<std::uint32_t, oneapi::dpl::__internal::__value_t<_Range2>>>>
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Since most of the aliases are shared between the implementation and the transform scan calls are similar, I personally prefer to have only one function here.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -970,6 +971,45 @@ struct __count_fn_pred
 };
 #endif
 
+template <typename _ValueType, typename _FlagType, typename _BinaryOp>
+struct __segmented_scan_fun
+{
+    __segmented_scan_fun(_BinaryOp __input) : __binary_op(__input) {}
```
SergeyKopienko: I think probably it doesn't make sense to have here constructor and private members.
What about remove constructor and private? Like we doing in many other cases...

include/oneapi/dpl/pstl/utils.h
```diff
@@ -970,6 +971,45 @@ struct __count_fn_pred
...
+  private:
+    _BinaryOp __binary_op;
+};
+
+template <typename _T, typename _Predicate>
+struct __replace_if_fun
+{
+    using __result_of = _T;
+
+    __replace_if_fun(_Predicate __pred, _T __new_value) : __pred(__pred), __new_value(__new_value) {}
```
SergeyKopienko: The same

include/oneapi/dpl/pstl/utils.h
```diff
@@ -970,6 +971,45 @@ struct __count_fn_pred
...
+
+    template <typename _T1, typename _T2>
+    _T
+    operator()(_T1&& __a, _T2&& __s) const
```
SergeyKopienko, line 1055:
```suggestion
    operator()(const _T1& __a, _T2&& __s) const
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -970,6 +971,45 @@ struct __count_fn_pred
...
+    {
+        return __pred(__s) ? __new_value : __a;
```
SergeyKopienko:
```suggestion
        return __pred(std::forward<_T2>(__s)) ? __new_value : __a;
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,202 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
         oneapi::dpl::unseq_backend::__has_known_identity<_BinaryOperator, __val_type>{});
 }
 
+//------------------------------------------------------------------------
+// parallel_scan_by_segment - sync pattern
+//------------------------------------------------------------------------
+template <typename _CustomName, bool __is_inclusive, typename _Range1, typename _Range2, typename _Range3,
+          typename _BinaryPredicate, typename _BinaryOperator, typename _InitType>
+__future<sycl::event, __result_and_scratch_storage<
+                          oneapi::dpl::__internal::tuple<std::uint32_t, oneapi::dpl::__internal::__value_t<_Range2>>>>
```
SergeyKopienko, line 2131: Are you sure about this return type?
For example, `__parallel_transform_reduce_then_scan` return type is `__future<sycl::event, __result_and_scratch_storage<typename _InitType::__value_type>>`
So what is it  in this case?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -209,6 +209,46 @@ struct __write_red_by_seg
     std::size_t __n;
 };
 
+template <bool __is_inclusive, typename _InitType, typename _BinaryOp>
+struct __write_scan_by_seg
+{
+    using _TempData = __noop_temp_data;
+    _InitType __init_value;
+    _BinaryOp __binary_op;
+    template <typename _OutRng, typename _ValueType>
```
SergeyKopienko, line 221:
```suggestion

    template <typename _OutRng, typename _ValueType>
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -209,6 +209,46 @@ struct __write_red_by_seg
...
+            static_assert(
+                std::is_same_v<_InitType, oneapi::dpl::unseq_backend::__init_value<typename _InitType::__value_type>>,
+                "exclusive_scan_by_segment must have an initial element");
+            if (get<1>(__v))
+                __out_rng[__id] = static_cast<_ConvertedTupleType>(get<1>(__init_value.__value));
+            else
+            {
+                __out_rng[__id] =
+                    static_cast<_ConvertedTupleType>(__binary_op(get<1>(__init_value.__value), get<1>(get<0>(__v))));
+            }
```
SergeyKopienko: What about ternary operator here?

```suggestion
            __out_rng[__id] = get<1>(__v)
                              ? static_cast<_ConvertedTupleType>(get<1>(__init_value.__value))
                              : static_cast<_ConvertedTupleType>(__binary_op(get<1>(__init_value.__value), get<1>(get<0>(__v))));
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
mmichel11: Yes, because we are using the `_WrappedInitType` defined within this function to call transform reduce-then-scan. It ends up being a tuple of the flag and value type that is passed as init to `__parallel_transform_reduce_then_scan` which is reflected in the return type.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: I agree it's cleaner, done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: Done

include/oneapi/dpl/pstl/utils.h
mmichel11: This is just relocated code, but I don't have a problem changing it.

From my perspective, `operator()(_T1&& __a, const _T2& __s) const` makes the most sense. `__s` is just passed to a unary predicate, so a const ref is all we need here. `__a` may be returned, so I think a forwarding reference makes sense for this as it can be moved in and out of the function object.

include/oneapi/dpl/pstl/utils.h
mmichel11: See the above comment for the change I made here.

include/oneapi/dpl/pstl/utils.h
mmichel11: This is just relocated code, but I don't have a problem changing it. I've made the change.

include/oneapi/dpl/pstl/utils.h
mmichel11: Done

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: ok

SergeyKopienko:
@mmichel11 Now we have:
```C++
template <typename _BackendTag, typename _Policy, typename _InputIterator1, typename _InputIterator2,
          typename _OutputIterator, typename _BinaryPredicate, typename _Operator, typename _Inclusive>
_OutputIterator
__pattern_scan_by_segment(__hetero_tag<_BackendTag> __tag, _Policy&& __policy, _InputIterator1 __first1,
                          _InputIterator1 __last1, _InputIterator2 __first2, _OutputIterator __result,
                          _BinaryPredicate __binary_pred, _Operator __binary_op, _Inclusive __is_inclusive);
```
and
```C++
template <typename _BackendTag, typename _Policy, typename _InputIterator1, typename _InputIterator2,
          typename _OutputIterator, typename _T, typename _BinaryPredicate, typename _Operator, typename _Inclusive>
_OutputIterator
__pattern_scan_by_segment(__hetero_tag<_BackendTag>, _Policy&& __policy, _InputIterator1 __first1,
                          _InputIterator1 __last1, _InputIterator2 __first2, _OutputIterator __result, _T __init,
                          _BinaryPredicate __binary_pred, _Operator __binary_op, _Inclusive)
```
Is it ok that `typename _T` and `_T __init` are placed in the middle of params? What about to move them into the end? How do you think?

In this case we may have only one `__pattern_scan_by_segment` with `_T __init` with default value as `unseq_backend::__no_init_value{}`

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: What we have now:
```C++
template <typename _T, typename _Predicate>
struct __replace_if_fun
{
    using __result_of = _T;

    template <typename _T1, typename _T2>
    _T
    operator()(_T1&& __a, const _T2& __s) const
    {
        return __pred(__s) ? __new_value : std::forward<_T1>(__a);
    }

    _Predicate __pred;
    const _T __new_value;
};
```

In `operator()(_T1&& __a, const _T2& __s) const` you return `std::forward<_T1>(__a)` if `__pred(__s)` return`0', `false` and etc. But if we take a look to the type or return value, it is `_T` - the instance of type `_T`. So I think for now this `std::forward<_T1>(__a)` doesn't make sense and enough simple return `__a`:
```C++
 return __pred(__s) ? __new_value : __a;
```

Now about the `__s` argument of `__pred`. I think better to pass it into `__pred` with source value category: we may describe it as universal reference and pass into `__pred` by `std::forward` call.

mmichel11:
> @mmichel11 Now we have:
> ...
> Is it ok that `typename _T` and `_T __init` are placed in the middle of params? What about to move them into the end? How do you think?
> 
> In this case we may have only one `__pattern_scan_by_segment` with `_T __init` with default value as `unseq_backend::__no_init_value{}`

I looked at this a bit deeper and the inclusive-scan-by-segment case was being handled oddly: an `__init_value` was wrapping a `__no_init_value`. I switched to a similar design to `__pattern_transform_scan` for init and no init handling which I think is the cleanest.

include/oneapi/dpl/pstl/utils.h
mmichel11: Thanks, I agree. Done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,202 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
+    }
+    // Implicit synchronization in this call. We need to wrap the policy as the implementation may still call
+    // reduce-then-scan and needs to avoid duplicate kernel names.
+    __parallel_scan_by_segment_fallback<_CustomName, __is_inclusive>(
+        oneapi::dpl::__internal::__device_backend_tag{},
+        oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__scan_by_seg_fallback>(
+            std::forward<_ExecutionPolicy>(__exec)),
+        std::forward<_Range1>(__keys), std::forward<_Range2>(__values), std::forward<_Range3>(__out_values),
+        __binary_pred, __binary_op, __init,
+        oneapi::dpl::unseq_backend::__has_known_identity<_BinaryOperator, _ValueType>{});
```
danhoeflinger, line 2299: I was thinking about if it is worth it to keep the path where we have a "known identity", but don't qualify for `reduce_then_scan`.

I think that all types with known identities with some binary operator are trivially copyable (arithmetic types and bool).  Therefore, this is effectively a sycl CPU-specific algorithm (or for GPUs which don't support subgroup size of 32).  I think this is still worthwhile, and likely is a large improvement over the no known identity fallback for those cases.

Nothing to do act on here, just wanted to post my thoughts...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -836,6 +921,28 @@ struct __red_by_seg_op
...
+    template <typename _Tup1, typename _Tup2>
+    auto
+    operator()(const _Tup1& __lhs_tup, const _Tup2& __rhs_tup) const
+    {
+        using std::get;
+        using _OpReturnType = decltype(__binary_op(get<1>(__lhs_tup), get<1>(__rhs_tup)));
+        // The left-hand side has processed elements from the same segment, so update the reduction value.
+        if (get<0>(__rhs_tup) == 0)
+        {
+            return oneapi::dpl::__internal::make_tuple(get<0>(__lhs_tup),
```
danhoeflinger, line 940: I like this implementation, but I think its worthwhile to describe the significance of putting `get<0>(__lhs_tup)` here in the result as a comment. Effectively, it updates the current running scan value with the status of whether we have "seen" the start of the current segment, which then "infects" values forward as we interact with larger shifted right values in the subgroup scan, preventing them from incorporating input from the previous segment.

I had to write out a quick example to prove to myself it worked.

danhoeflinger (commented):
Largely this looks good to me.  I want to take another pass in detail at more minor things, but I think this is quite good, and only have the single request for a clarifying comment.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -2055,6 +2055,202 @@ __parallel_reduce_by_segment(oneapi::dpl::__internal::__device_backend_tag, _Exe
...
+            std::forward<_Range3>(__out_values), _GenReduceInput{__binary_pred}, _ReduceOp{__binary_op},
+            _GenScanInput{}, _ScanInputTransform{}, _WriteOp{__wrapped_init, __binary_op}, __wrapped_init,
+            /*Inclusive*/ std::true_type{}, /*_IsUniquePattern=*/std::false_type{});
+    }
+    else
+    {
+        // The init value is manually applied through the write functor in exclusive-scan-by-segment and we always pass
+        // __no_init_value to the transform scan call. This is because init handling must occur on a per-segment basis
+        // and functions differently than the typical scan init which is only applied once in a single location.
+        _WrappedInitType __wrapped_init{{0, __init.__value}};
```
danhoeflinger: Perhaps this makes it more complicated, but do we actually need the wrapped init type to be a `_PackedFlagValueType`?
It seems like in practice in the `WriteOp`, we only ever use the `_ValueType` element, so can we just provide that?
Of course, we can pass `oneapi::dpl::unseq_backend::__no_init_value<_PackedFlagValueType>{}` for the init of the transform scan operation, but it seems like for the segmented part we could have an init type of  `_ValueType` (without the packed tuple and even without the init value wrapper.

I think we could could remove the `conditional_t` part above as well in favor of types / values tailored to the constexpr branches.

danhoeflinger: This will change `_WriteOp`, but actually since inclusive scan never has or uses an init anyway, you can just use `__write_scan_by_seg<__is_inclusive, _ValueType, _BinaryOperator>` for both inclusive and exclusive (and I believe it should be able to move up to join the rest of the using definitions).

mmichel11: Yes, this works and makes things simpler. We can just pass `__init` directly to `_WriteOp` in both cases allowing us to remove the constexpr conditional between inclusive and exclusive scan-by-segment.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
mmichel11: Sure, I updated the comment. The reduce-by-segment equivalent is similar in the approach, so I updated that one as well.

danhoeflinger (approved):
LGTM!

danhoeflinger (approved):
LGTM after rebase & sycl device copyable changes

PR #2316, opened 2025-06-20
[oneDPL][ranges][test] Test coverage improved -  usage of different projections in cases of two sequences.

MikeDvorskiy:
[oneDPL][ranges][test] Test coverage improved -  usage of different projections in cases of two sequences.

### DIFF
--- a/test/parallel_api/ranges/std_ranges_equal.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_equal.pass.cpp
@@ -25,7 +25,7 @@ main()
     auto equal_checker = TEST_PREPARE_CALLABLE(std::ranges::equal);
 
     test_range_algo<0, int, data_in_in>{big_sz}(dpl_ranges::equal, equal_checker, binary_pred);
-    test_range_algo<1, int, data_in_in>{}(dpl_ranges::equal, equal_checker, binary_pred, proj, proj);
+    test_range_algo<1, int, data_in_in>{}(dpl_ranges::equal, equal_checker, binary_pred, proj, std::identity{});
     test_range_algo<2, P2, data_in_in>{}(dpl_ranges::equal, equal_checker, binary_pred, &P2::x, &P2::x);
     test_range_algo<3, P2, data_in_in>{}(dpl_ranges::equal, equal_checker, binary_pred, &P2::proj, &P2::proj);
 #endif //_ENABLE_STD_RANGES_TESTING
--- a/test/parallel_api/ranges/std_ranges_find_end.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_find_end.pass.cpp
@@ -30,9 +30,9 @@ main()
 
     //false result
     test_range_algo<1, int, data_in_in>{big_sz}(dpl_ranges::find_end, find_end_checker, binary_pred);
-    test_range_algo<2, int, data_in_in>{}(dpl_ranges::find_end, find_end_checker, binary_pred_const, proj, proj);
+    test_range_algo<2, int, data_in_in>{}(dpl_ranges::find_end, find_end_checker, binary_pred_const, std::identity{}, proj);
 
-    test_range_algo<3, int, data_in_in>{}(dpl_ranges::find_end, find_end_checker, binary_pred, proj, proj);
+    test_range_algo<3, int, data_in_in>{}(dpl_ranges::find_end, find_end_checker, binary_pred, proj, std::identity{});
     test_range_algo<4, P2, data_in_in>{}(dpl_ranges::find_end, find_end_checker, binary_pred, &P2::x, &P2::x);
     test_range_algo<5, P2, data_in_in>{}(dpl_ranges::find_end, find_end_checker, binary_pred, &P2::proj, &P2::proj);
 #endif //_ENABLE_STD_RANGES_TESTING
--- a/test/parallel_api/ranges/std_ranges_find_first_of.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_find_first_of.pass.cpp
@@ -27,7 +27,7 @@ main()
 
     test_range_algo<0, int, data_in_in>{medium_size}(dpl_ranges::find_first_of, find_first_of_checker, binary_pred);
     test_range_algo<1, int, data_in_in>{}(dpl_ranges::find_first_of, find_first_of_checker, binary_pred_const);
-    test_range_algo<2, int, data_in_in>{}(dpl_ranges::find_first_of, find_first_of_checker, binary_pred, proj, proj);
+    test_range_algo<2, int, data_in_in>{}(dpl_ranges::find_first_of, find_first_of_checker, binary_pred, proj, std::identity{});
     test_range_algo<3, P2, data_in_in>{}(dpl_ranges::find_first_of, find_first_of_checker, binary_pred, &P2::x, &P2::x);
     test_range_algo<4, P2, data_in_in>{}(dpl_ranges::find_first_of, find_first_of_checker, binary_pred, &P2::proj, &P2::proj);
 
--- a/test/parallel_api/ranges/std_ranges_merge.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_merge.pass.cpp
@@ -74,11 +74,11 @@ main()
 
     test_range_algo<0, int, data_in_in_out_lim>{big_sz}(dpl_ranges::merge, merge_checker, std::ranges::less{}, std::identity{}, std::identity{});
 
-    test_range_algo<1, int, data_in_in_out_lim>{}(dpl_ranges::merge, merge_checker, std::ranges::less{}, proj, proj);
+    test_range_algo<1, int, data_in_in_out_lim>{}(dpl_ranges::merge, merge_checker, std::ranges::less{}, proj, std::identity{});
     test_range_algo<2, P2, data_in_in_out_lim>{}(dpl_ranges::merge, merge_checker, std::ranges::less{}, &P2::x, &P2::x);
     test_range_algo<3, P2, data_in_in_out_lim>{}(dpl_ranges::merge, merge_checker, std::ranges::less{}, &P2::proj, &P2::proj);
 
-    test_range_algo<4, int, data_in_in_out_lim>{}(dpl_ranges::merge, merge_checker, std::ranges::greater{}, proj, proj);
+    test_range_algo<4, int, data_in_in_out_lim>{}(dpl_ranges::merge, merge_checker, std::ranges::greater{}, std::identity{}, proj);
     test_range_algo<5, P2, data_in_in_out_lim>{}(dpl_ranges::merge, merge_checker, std::ranges::greater{}, &P2::x, &P2::x);
     test_range_algo<6, P2, data_in_in_out_lim>{}(dpl_ranges::merge, merge_checker, std::ranges::greater{}, &P2::proj, &P2::proj);
 
--- a/test/parallel_api/ranges/std_ranges_mismatch.pass.cpp
+++ b/test/parallel_api/ranges/std_ranges_mismatch.pass.cpp
@@ -26,7 +26,7 @@ main()
 
     test_range_algo<0, int, data_in_in>{big_sz}(dpl_ranges::mismatch, mismatch_checker, binary_pred);
     test_range_algo<1, int, data_in_in>{}(dpl_ranges::mismatch, mismatch_checker, binary_pred_const);
-    test_range_algo<2, int, data_in_in>{}(dpl_ranges::mismatch, mismatch_checker, binary_pred, proj, proj);
+    test_range_algo<2, int, data_in_in>{}(dpl_ranges::mismatch, mismatch_checker, binary_pred, std::identity{}, proj);
     test_range_algo<3, P2, data_in_in>{}(dpl_ranges::mismatch, mismatch_checker, binary_pred, &P2::x, &P2::x);
     test_range_algo<4, P2, data_in_in>{}(dpl_ranges::mismatch, mismatch_checker, binary_pred, &P2::proj, &P2::proj);
 #endif //_ENABLE_STD_RANGES_TESTING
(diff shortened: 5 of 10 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -25,6 +25,7 @@
 
 #include "sycl_defs.h"
 #include "parallel_backend_sycl_utils.h"
+#include "../../../functional" //for oneapi::dpl::identity
```
SergeyKopienko: In the PR https://github.com/uxlfoundation/oneDPL/pull/2294 I have some discussion about are we able to have this include or no: https://github.com/uxlfoundation/oneDPL/pull/2294#discussion_r2127057236

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -603,23 +603,19 @@ __pattern_fill(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy&&, _
 }
 
 //---------------------------------------------------------------------------------------------------------------------
-// pattern_merge
+// __pattern_merge_ranges
```
SergeyKopienko:
```suggestion
// pattern_merge_ranges
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -122,11 +125,12 @@ __find_start_point(const _Rng1& __rng1, const _Index __rng1_from, _Index __rng1_
     __it_t __diag_it_begin(idx1_from);
     __it_t __diag_it_end(idx1_to);
 
-    const __it_t __res =
-        std::lower_bound(__diag_it_begin, __diag_it_end, false,
-                         [&__rng1, &__rng2, __index_sum, __comp](_Index __idx, const bool __value) mutable {
-                             return __value == __comp(__rng2[__index_sum - __idx], __rng1[__idx]);
-                         });
+    const __it_t __res = std::lower_bound(
```
SergeyKopienko, line 128: I am worried about this place for performance: usually `typename _Proj1 = oneapi::dpl::identity, typename _Proj2 = oneapi::dpl::identity` but we call `std::invoke`

test/parallel_api/ranges/std_ranges_equal.pass.cpp
```diff
@@ -25,7 +25,7 @@ main()
     auto equal_checker = TEST_PREPARE_CALLABLE(std::ranges::equal);
 
     test_range_algo<0, int, data_in_in>{big_sz}(dpl_ranges::equal, equal_checker, binary_pred);
-    test_range_algo<1, int, data_in_in>{}(dpl_ranges::equal, equal_checker, binary_pred, proj, proj);
+    test_range_algo<1, int, data_in_in>{}(dpl_ranges::equal, equal_checker, binary_pred, proj, std::identity{});
```
SergeyKopienko: Let's use `oneapi::dpl::identity` in all tests from this PR?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: What about
```C++
    if constexpr (std::is_same_v<std::decay_t<_Proj1>, oneapi::dpl::identity> && std::is_same_v<std::decay_t<_Proj2>, oneapi::dpl::identity>)
    {
        const __it_t __res =
            std::lower_bound(__diag_it_begin, __diag_it_end, false,
                             [&__rng1, &__rng2, __index_sum, __comp](_Index __idx, const bool __value) mutable {
                                 return __value == __comp(__rng2[__index_sum - __idx], __rng1[__idx]);
                             });

        return _split_point_t<_Index>{*__res, __index_sum - *__res + 1};
    }
    else
    {
        const __it_t __res = std::lower_bound(
            __diag_it_begin, __diag_it_end, false,
            [&__rng1, &__rng2, __index_sum, __comp, __proj1, __proj2](_Index __idx, const bool __value) mutable {
                return __value == std::invoke(__comp, std::invoke(__proj2, __rng2[__index_sum - __idx]),
                                              std::invoke(__proj1, __rng1[__idx]));
            });

        return _split_point_t<_Index>{*__res, __index_sum - *__res + 1};
    }
```

test/parallel_api/ranges/std_ranges_equal.pass.cpp
dmitriy-sobolev: It's already used in all the tests in the PR at least once (which should be enough to check the issue with different projection types in my opinion). Could you be more specific, e.g. point other places where you would like to add it?

dmitriy-sobolev: Ah, the suggestion is to use `oneapi::dpl::identity` instead of `std::identity`. Let's use `std::identity` as it is right now - it's a functionality from c++ standard, which is more likely to be employed by a user. It also makes sense to use `oneapi::dpl::identity` if you do no have access to c++20, it it is not a case.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
dmitriy-sobolev: Reverted

SergeyKopienko: It's not reverted, we already have changes in this lines.
But without `__` in the prefix, as it were in the `main` state too.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3021,23 +3020,22 @@ __pattern_merge(_Tag, _ExecutionPolicy&&, _ForwardIterator1 __first1, _ForwardIt
 }
 
 template <typename _Tag, typename _ExecutionPolicy, typename _It1, typename _Index1, typename _It2, typename _Index2,
-          typename _OutIt, typename _Index3, typename _Comp>
+          typename _OutIt, typename _Index3, typename _Comp, typename _Proj1, typename _Proj2>
 std::pair<_It1, _It2>
 ___merge_path_out_lim(_Tag, _ExecutionPolicy&&, _It1 __it_1, _Index1 __n_1, _It2 __it_2, _Index2 __n_2,
-                      _OutIt __it_out, _Index3 __n_out, _Comp __comp)
+                      _OutIt __it_out, _Index3 __n_out, _Comp __comp, _Proj1 __proj1, _Proj2 __proj2)
 {
```
SergeyKopienko, line 3027:
```suggestion
{
    static_assert(__is_serial_tag_v<_Tag> || __is_parallel_forward_tag_v<_Tag>);
```

SergeyKopienko: ...or
```C++
static_assert(__is_serial_tag_v<_Tag>);
```

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -823,23 +819,21 @@ __pattern_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _Ran
...
+          typename _Comp, typename _Proj1, typename _Proj2>
 auto
-__pattern_merge(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&& __r2, _OutRange&& __out_r,
-    _Comp __comp, _Proj1 __proj1, _Proj2 __proj2)
+__pattern_merge_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&& __r2,
+                       _OutRange&& __out_r, _Comp __comp, _Proj1 __proj1, _Proj2 __proj2)
 {
-    oneapi::dpl::__internal::__binary_op<_Comp, _Proj1, _Proj2> __comp_2{__comp, __proj1, __proj2};
-
     using _Index1 = std::ranges::range_difference_t<_R1>;
```
SergeyKopienko, line 838:
```suggestion
    static_assert(__is_parallel_tag_v<_Tag> || typename _Tag::__is_vector{});

    using _Index1 = std::ranges::range_difference_t<_R1>;
```

include/oneapi/dpl/pstl/algorithm_impl.h
dmitriy-sobolev: Let's abstain from making changes unrelated to the goal of the PR. The same applies to the suggestion below.

SergeyKopienko: Could you create sone new issue and add these `static_assert` in the separate PR ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
dmitriy-sobolev: Fixed

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: It's the heterogenous backend. Such asserts are not used here.

include/oneapi/dpl/pstl/algorithm_impl.h
dmitriy-sobolev: A separate PR: https://github.com/uxlfoundation/oneDPL/pull/2324

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
dmitriy-sobolev: I do not see much value in adding such a branch from the performance standpoint. As for readability - it is harmful. 

`std::invoke` and `std::identity` should be optimized away by a compiler. You can play with https://godbolt.org/z/da1Wc7M7n. The example is taken from cppreference.

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: Thanks!

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Yes, this optimizations looking good.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -784,15 +784,11 @@ struct __copy1_wrapper;
 template <typename _Name>
 struct __copy2_wrapper;
 
-struct __out_size_limit : public std::true_type
-{
-};
-
 template <typename _BackendTag, typename _ExecutionPolicy, typename _Range1, typename _Range2, typename _Range3,
-          typename _Compare>
+          typename _Compare, typename _Proj1 = oneapi::dpl::identity, typename _Proj2 = oneapi::dpl::identity>
```
SergeyKopienko, line 789: Required to include `functional_impl.h` header in this file.
And please check another places.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -2953,12 +2953,11 @@ __pattern_remove_if(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __exec,
 //------------------------------------------------------------------------
 // merge
 //------------------------------------------------------------------------
-// Serial version of ___merge_path_out_lim merges the 1st sequence and the 2nd sequence in "reverse order":
-// the identical elements from the 2nd sequence are merged first.
-template <typename _Iterator1, typename _Iterator2, typename _Iterator3, typename _Comp>
+template <typename _Iterator1, typename _Iterator2, typename _Iterator3, typename _Comp,
+          typename _Proj1 = oneapi::dpl::identity, typename _Proj2 = oneapi::dpl::identity>
```
SergeyKopienko:
```suggestion
template <typename _Iterator1, typename _Iterator2, typename _Iterator3, typename _Comp, typename _Proj1, typename _Proj2>
```
- we specify `_Proj1` and `_Proj2` in all calls of `__serial_merge_out_lim` so I think this defaults not required here.

test/parallel_api/ranges/std_ranges_equal.pass.cpp
SergeyKopienko: What is the goal to specify `, std::identity{}` (as last parameter in calls) here and in other places?
I think it will have more sense to without this param: 1) we will check that it's compiled 2) we will check that default state in `identity`.

dmitriy-sobolev: The goal is to check that the interface with two projections passed explicitly works fine. Probably, it is better to pass another, custom projection type.

SergeyKopienko: So in ideal we should test:
- no projections, one projection, two projections;
- each of them - identity + custom.

dmitriy-sobolev: Yes, a case with one projection is skipped and should be added.

include/oneapi/dpl/pstl/algorithm_impl.h
dmitriy-sobolev: I've removed it. Currently `__serial_merge_out_lim` is used only by the interfaces from range algorithms, which provide projections.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: Added. It was the only place in this PR with `oneapi::dpl::identity` and without `functional_impl.h`.

test/parallel_api/ranges/std_ranges_equal.pass.cpp
dmitriy-sobolev: I've added such checks in #2326. Also I've replaced `std::identity` with a custom projection here.

dmitriy-sobolev: After looking at the proposed PR more, I think it is better to remove an extra projection as you suggested to avoid excessive testing. #2326 is irrelevant after that change.

PR #2317, opened 2025-06-20
Set algorithms, revert to set_a_write algorithm for small sizes

danhoeflinger:
This PR makes some infrastructure changes to allow a runtime decision based upon number of elements to select the set algorithm selected, in addition to limitations based upon platform and environment.

1) Relocate algorithm choice from hetero level to sycl backend and make ` __parallel_set_op` blocking. (temporarily required for type alignment, see #2300)
2) Create sycl backend entry points to call "merge" (`__parallel_merge_impl`) and "copy" (`__parallel_copy_impl`).
3) Add a threshold to switch algorithms from set_a_scan to balanced_path based upon set tag in `__set_op_impl` using `__consider_write_a_alg`

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
@@ -1772,13 +1772,8 @@ __pattern_hetero_set_op(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _F
     typedef typename std::iterator_traits<_ForwardIterator1>::difference_type _Size1;
 
     const _Size1 __n1 = std::distance(__first1, __last1);
-    _Size1 __output_size = __n1;
-    if constexpr (_SetTag::__can_write_from_rng2_v)
-    {
-        const _Size1 __n2 = std::distance(__first2, __last2);
-        // one shot algorithm can write from set 1 or set 2, whereas old algorithm can only write from set 1.
-        __output_size = __n1 + __n2;
-    }
+    const _Size1 __n2 = std::distance(__first2, __last2);
+    const _Size1 __output_size = __n1 + __n2;
 
     auto __keep1 =
         oneapi::dpl::__ranges::__get_sycl_range<__par_backend_hetero::access_mode::read, _ForwardIterator1>();
@@ -1792,16 +1787,10 @@ __pattern_hetero_set_op(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _F
 
     auto __result_size = __par_backend_hetero::__parallel_set_op(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
                                                                  __buf1.all_view(), __buf2.all_view(),
-                                                                 __buf3.all_view(), __comp, __set_tag)
-                             .get();
+                                                                 __buf3.all_view(), __comp, __set_tag);
 
     return __result + __result_size;
 }
-
-template <typename Name>
-struct __set_intersection_scan_then_propagate
-{
-};
 
 template <typename _BackendTag, typename _ExecutionPolicy, typename _ForwardIterator1, typename _ForwardIterator2,
           typename _OutputIterator, typename _Compare>
@@ -1813,26 +1802,13 @@ __pattern_set_intersection(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& _
     // intersection is empty
     if (__first1 == __last1 || __first2 == __last2)
         return __result;
-    if (__par_backend_hetero::__can_set_op_write_from_set_b(_BackendTag{}, __exec))
-    {
-        return __pattern_hetero_set_op(__tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2,
-                                       __last2, __result, __comp, unseq_backend::_IntersectionTag<std::true_type>());
-    }
-    return __pattern_hetero_set_op(
-        __tag,
-        oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__set_intersection_scan_then_propagate>(
-            std::forward<_ExecutionPolicy>(__exec)),
-        __first1, __last1, __first2, __last2, __result, __comp, unseq_backend::_IntersectionTag<std::false_type>());
+    return __pattern_hetero_set_op(__tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2,
+                                       __last2, __result, __comp, unseq_backend::_IntersectionTag{});
 }
 
 //Dummy names to avoid kernel problems
 template <typename Name>
 struct __set_difference_copy_case_1
-{
-};
-
-template <typename Name>
-struct __set_difference_scan_then_propagate
 {
 };
 
@@ -1856,19 +1832,8 @@ __pattern_set_difference(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __e
                 ::std::forward<_ExecutionPolicy>(__exec)),
             __first1, __last1, __result, oneapi::dpl::__internal::__brick_copy<__hetero_tag<_BackendTag>>{});
     }
-    if (__par_backend_hetero::__can_set_op_write_from_set_b(_BackendTag{}, __exec))
-    {
-        return __pattern_hetero_set_op(
-            __tag,
-            oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__set_difference_scan_then_propagate>(
-                std::forward<_ExecutionPolicy>(__exec)),
-            __first1, __last1, __first2, __last2, __result, __comp, unseq_backend::_DifferenceTag<std::true_type>());
-    }
-    else
-    {
-        return __pattern_hetero_set_op(__tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2,
-                                       __last2, __result, __comp, unseq_backend::_DifferenceTag<std::false_type>());
-    }
+    return __pattern_hetero_set_op(__tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2,
+                                       __last2, __result, __comp, oneapi::dpl::unseq_backend::_DifferenceTag{});
 }
 
 //Dummy names to avoid kernel problems
@@ -1879,11 +1844,6 @@ struct __set_union_copy_case_1
 
 template <typename Name>
 struct __set_union_copy_case_2
-{
-};
-
-template <typename Name>
-struct __set_union_scan_then_propagate
 {
 };
 
@@ -1917,34 +1877,8 @@ __pattern_set_union(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, 
             __first1, __last1, __result, oneapi::dpl::__internal::__brick_copy<__hetero_tag<_BackendTag>>{});
     }
 
-    if (__par_backend_hetero::__can_set_op_write_from_set_b(_BackendTag{}, __exec))
-    {
-        return __pattern_hetero_set_op(__tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2,
-                                       __last2, __result, __comp, unseq_backend::_UnionTag<std::true_type>());
-    }
-    else
-    {
-        using _ValueType = typename std::iterator_traits<_OutputIterator>::value_type;
-
-        // temporary buffer to store intermediate result
-        const auto __n2 = __last2 - __first2;
-        oneapi::dpl::__par_backend_hetero::__buffer<_ValueType> __diff(__n2);
-        auto __buf = __diff.get();
-
-        //1. Calc difference {2} \ {1}
-        const auto __n_diff =
-            oneapi::dpl::__internal::__pattern_hetero_set_op(
-                __tag, oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__set_union_scan_then_propagate>(__exec),
-                __first2, __last2, __first1, __last1, __buf, __comp, unseq_backend::_DifferenceTag<std::false_type>()) -
-            __buf;
-
-        //2. Merge {1} and the difference
-        return oneapi::dpl::__internal::__pattern_merge(
-            __tag,
-            oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__set_union_copy_case_2>(
-                std::forward<_ExecutionPolicy>(__exec)),
-            __first1, __last1, __buf, __buf + __n_diff, __result, __comp);
-    }
+    return __pattern_hetero_set_op(__tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2,
+                                   __last2, __result, __comp, oneapi::dpl::unseq_backend::_UnionTag{});
 }
 
 //Dummy names to avoid kernel problems
@@ -1955,16 +1889,6 @@ struct __set_symmetric_difference_copy_case_1
 
 template <typename Name>
 struct __set_symmetric_difference_copy_case_2
-{
-};
-
-template <typename Name>
-struct __set_symmetric_difference_phase_1
-{
-};
-
-template <typename Name>
-struct __set_symmetric_difference_phase_2
 {
 };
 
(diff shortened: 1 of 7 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1028,11 +1039,10 @@ __parallel_set_reduce_then_scan_set_a_write(sycl::queue& __q, _Range1&& __rng1,
     using _ScanInputTransform = oneapi::dpl::__par_backend_hetero::__get_zeroth_element;
 
     oneapi::dpl::__par_backend_hetero::__buffer<std::int32_t> __mask_buf(__rng1.size());
-
     const std::size_t __n = __rng1.size();
-    return __parallel_transform_reduce_then_scan<sizeof(_Size), _CustomName>(
+    return __parallel_transform_reduce_then_scan<sizeof(oneapi::dpl::__internal::__value_t<_Range1>), _CustomName>(
```
mmichel11, line 1046: The bytes per work item iter may change here. Is this a better value than in the previous implementation?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1136,34 +1146,248 @@ __parallel_set_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _Range
...
+template <typename _CustomName>
+struct scan_then_propagate_wrapper
+{
+};
+
+template <typename _SetTag>
+struct __consider_write_a_alg
+{
+    // empirically determined threshold for using write_a algorithm
+    static constexpr std::size_t __threshold = 65536;
```
mmichel11: Optional suggestion: Have you tested to see if the size of input / output size types influence this threshold? If it does, it might be useful to have a small heuristic based on that.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1136,34 +1146,248 @@ __parallel_set_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _Range
...
+
+// Selects the right implementation of set based on the size and platform
+template <typename _CustomName, typename _Range1, typename _Range2, typename _Range3, typename _Compare,
+          typename _SetTag>
+std::size_t
+__set_op_impl(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _Range3&& __result, _Compare __comp,
+              _SetTag __set_tag)
+{
+    size_t __n1 = __rng1.size();
+    size_t __n2 = __rng2.size();
```
mmichel11: Formatting nitpick:
```suggestion
    std::size_t __n1 = __rng1.size();
    std::size_t __n2 = __rng2.size();
```

mmichel11 (commented):
Only minor comments from me.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -1276,7 +1271,8 @@ class __brick_set_op
 
         auto __res = __internal::__pstl_lower_bound(__b, _Size2(0), __nb, __val_a, __comp);
 
-        bool bres = _IsOpDifference(); //initialization in true in case of difference operation; false - intersection.
+        constexpr bool __is_difference = std::is_same_v<_IsOpDifference, oneapi::dpl::unseq_backend::_DifferenceTag>;
+        bool bres = __is_difference; //initialization in true in case of difference operation; false - intersection.
```
adamfidel: Small typo suggestion
```suggestion
        bool bres = __is_difference; //initialization is true in case of difference operation; false - intersection.
```

danhoeflinger: fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Yes, this better describes the bytes that an input range will deal with per work-item.  The story is more complicated with the set algorithms than some of the other reduce then scan variants, but we are scanning over the input set A data, and the size type of the input doesn't have much to do with this operation.
I don't think that `__Size` was ever used in earnest, because the bytes per work item template argument was introduced when this set a write code was made "dormant".  This PR essentially revives this code for cases which it is useful.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Thanks for the suggestion. I looked at this and found that due to the temporary register memory requirements of the balanced path, the value type does impact the threshold with larger types favoring set_a write, and smaller favoring balanced path.  

I've also found that the threshold really depends on the size of "set a" in the set operation if we were to use set_a_write algorithm (which is `__rng1` for operations other than union, where set a must be  `__rng2`). 

Therefore, the threshold is now based upon the size of this "set a", and scaled by the value type of that set.

SergeyKopienko:
> I have found a same-mangled name issue with set_union and set_symmetric_difference I am investigating.

@danhoeflinger I can help you with this topic.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1772,13 +1772,8 @@ __pattern_hetero_set_op(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _F
...
     const _Size1 __n1 = std::distance(__first1, __last1);
-    _Size1 __output_size = __n1;
-    if constexpr (_SetTag::__can_write_from_rng2_v)
-    {
-        const _Size1 __n2 = std::distance(__first2, __last2);
-        // one shot algorithm can write from set 1 or set 2, whereas old algorithm can only write from set 1.
-        __output_size = __n1 + __n2;
-    }
+    const _Size1 __n2 = std::distance(__first2, __last2);
+    const _Size1 __output_size = __n1 + __n2;
```
mmichel11: Do we need to use the common type between `std::iterator_traits<_ForwardIterator1>::difference_type` and `std::iterator_traits<_ForwardIterator2>::difference_type` here and add a `_Size2` for `__n2`?

danhoeflinger: After the experience with #2427, yes I will audit this PR again to check for these sorts of issues.

danhoeflinger: This should be fixed.  We still deal in `auto` in some other places but it doesn't seem to add anything to make the types explicit.  Those instances are also on the host side, where we can use `std::size_t` without much thought for performance / footprint.

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
```diff
@@ -1769,16 +1769,12 @@ __pattern_hetero_set_op(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _F
...
 
-    const _Size1 __n1 = std::distance(__first1, __last1);
-    _Size1 __output_size = __n1;
-    if constexpr (_SetTag::__can_write_from_rng2_v)
-    {
-        const _Size1 __n2 = std::distance(__first2, __last2);
-        // one shot algorithm can write from set 1 or set 2, whereas old algorithm can only write from set 1.
-        __output_size = __n1 + __n2;
-    }
+    const _SizeType __n1 = std::distance(__first1, __last1);
```
SergeyKopienko, line 1775: In another cases @akukanov prefer `__last1 - __first1` of course if they are random-access iterators...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -555,22 +557,21 @@ template <typename _SetTag>
 struct __get_set_operation;
 
 template <>
-struct __get_set_operation<oneapi::dpl::unseq_backend::_IntersectionTag<std::true_type>> : public __set_intersection
+struct __get_set_operation<oneapi::dpl::unseq_backend::_IntersectionTag> : public __set_intersection
```
SergeyKopienko: Why `public` needed here?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -225,6 +225,18 @@ class __scan_single_wg_dynamic_kernel;
...
+template <typename _CustomName, typename _Index, typename _Range1, typename _Range2>
+__future<sycl::event>
+__parallel_copy_impl(sycl::queue& __q, _Index __count, _Range1&& __rng1, _Range2&& __rng2)
+{
+    auto __brick = oneapi::dpl::__internal::__pstl_assign{};
+    return oneapi::dpl::__par_backend_hetero::__parallel_for_impl<_CustomName>(
+        __q,
+        unseq_backend::walk_n_vectors_or_scalars<decltype(__brick)>{std::move(__brick),
+                                                                    static_cast<std::size_t>(__count)},
+        __count, std::forward<_Range1>(__rng1), std::forward<_Range2>(__rng2));
```
```suggestion
    return oneapi::dpl::__par_backend_hetero::__parallel_for_impl<_CustomName>(
        __q,
        unseq_backend::walk_n_vectors_or_scalars<decltype(__brick)>{oneapi::dpl::__internal::__pstl_assign{},
                                                                    static_cast<std::size_t>(__count)},
        __count, std::forward<_Range1>(__rng1), std::forward<_Range2>(__rng2));
```

Looks like the `__brick` local variable and their `move` not needed...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1160,34 +1171,257 @@ __parallel_set_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _Range
...
+struct set_a_write_wrapper;
+
+template <typename _SetTag, typename _Rng1, typename _Rng2>
 bool
-__can_set_op_write_from_set_b(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec)
+__use_write_a_alg(_SetTag, _Rng1&& __rng1, _Rng2&&)
 {
-    return oneapi::dpl::__par_backend_hetero::__is_gpu_with_reduce_then_scan_sg_sz(__exec.queue());
+    using __value_t = oneapi::dpl::__internal::__value_t<_Rng1>;
+    return __rng1.size() < 32 * 1024 * sizeof(__value_t);
```
SergeyKopienko: `32 * 1024` - what this mean and why you are using this magic value in this and in some another places?
Probably some comments needed and may be make sense to declare this value as some `constexp` value with required additional comments...

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1160,34 +1171,257 @@ __parallel_set_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _Range
...
+}
+
+template <typename _Rng1, typename _Rng2>
+bool
+__use_write_a_alg(oneapi::dpl::unseq_backend::_UnionTag, _Rng1&&, _Rng2&& __rng2)
```
SergeyKopienko: As far as we calls only some `const` operations on `__rng2` why we declare it as universal ref here? By this way we gives to called side the ability to make an error by moving ranges into this function.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1160,34 +1171,257 @@ __parallel_set_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _Range
...
+    // For union operations, we must are using __n2 as the set a in a difference operation prior to a merge, so the
+    // threshold should be on __n2. This must be in this order because semantically elements must be copied from __rng1
+    // when they are shared (important for algorithms where the key being compared is not the full element).
+    using __value_t = oneapi::dpl::__internal::__value_t<_Rng2>;
+    return __rng2.size() < 32 * 1024 * sizeof(__value_t);
+}
+
+template <typename _Rng1, typename _Rng2>
+bool
+__use_write_a_alg(oneapi::dpl::unseq_backend::_SymmetricDifferenceTag, _Rng1&&, _Rng2&&)
```
SergeyKopienko: The same

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -499,19 +499,14 @@ __get_starting_size_limit_for_large_submitter<int>()
...
 }
 
-template <typename _OutSizeLimit = std::false_type, typename _ExecutionPolicy, typename _Range1, typename _Range2,
+template <typename _CustomName, typename _OutSizeLimit = std::false_type, typename _Range1, typename _Range2,
           typename _Range3, typename _Compare, typename _Proj1 = oneapi::dpl::identity,
           typename _Proj2 = oneapi::dpl::identity>
 __future<sycl::event, std::shared_ptr<__result_and_scratch_storage_base>>
-__parallel_merge(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _Range1&& __rng1,
-                 _Range2&& __rng2, _Range3&& __rng3, _Compare __comp, _Proj1 __proj1 = {}, _Proj2 __proj2 = {})
+__parallel_merge_impl(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _Range3&& __rng3, _Compare __comp,
```
SergeyKopienko, line 506: We still have about 22 places where we have in params `oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec`
So my question - why we made some cleaning only for subset of functions in this PR which is for the new functional? May be better do not touch this in the current PR and make cleanup of all these places in some separate PR with `2022.11` milestone?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -392,9 +392,11 @@ struct __gen_set_mask
         auto __val_a = __set_a[__id];
 
         auto __res = oneapi::dpl::__internal::__pstl_lower_bound(__set_b, std::size_t{0}, __nb, __val_a, __comp);
+        constexpr bool __is_difference = std::is_same_v<_SetTag, oneapi::dpl::unseq_backend::_DifferenceTag>;
```
SergeyKopienko, line 395: The construction `constexpr bool __is_difference = ` used twice in the code.
May be make sense to create some trait?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
danhoeflinger: These are not random cleaning / refactoring.  We have moved the decision about using set_a_write algorithm or partitioned balanced path algorithm down to the sycl backend (from hetero).  Therefore we need to be able to call merge and copy from within the sycl backend (for union and symmetric difference), and must provide this indirection layer at the top of the sycl backend so that it can be called with a queue and CustomName rather than an ExecutionPolicy.

SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/algorithm_impl_hetero.h
danhoeflinger: I believe here we only know they are ForwardIterators according to the typenames.

SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Sure, done with a minor tweak to fix the type provided to template args.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: refactored to combine and added comments.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: I don't really see this being used more than there is right now. I'd prefer not to add a free floating tag for this simple operation, as far as lines of code its the same, and this is more clear to those reading the code I think.

SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: Its not, good catch.  fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1160,34 +1169,264 @@ __parallel_set_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _Range
...
+struct set_a_write_wrapper;
+
+struct __check_use_write_a_alg
+{
+    // Empirically determined threshold for when to switch between algorithms, scaled by the size of the value type.
+    static constexpr std::size_t __threshold_elements = 32768;
+
+    template <typename _SetTag, typename _Rng1, typename _Rng2>
+    bool
+    operator()(_SetTag, const _Rng1& __rng1, const _Rng2&) const
```
mmichel11, line 1362: nitpick: perhaps a comment here for the set algs that take this path may improve readability.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1160,34 +1169,264 @@ __parallel_set_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng2, _Range
...
+        using __value_t = oneapi::dpl::__internal::__value_t<_Rng1>;
+        return __rng1.size() < __threshold_elements * sizeof(__value_t);
+    }
+
+    template <typename _Rng1, typename _Rng2>
+    bool
+    operator()(oneapi::dpl::unseq_backend::_UnionTag, const _Rng1&, const _Rng2& __rng2) const
+    {
+        // For union operations, we must are using __n2 as the set a in a difference operation prior to a merge, so the
+        // threshold should be on __n2. The sets must kepts in this order because semantically elements must be copied
```
mmichel11:
```suggestion
        // For union operations, we must use __n2 as the set a in a difference operation prior to a merge, so the
        // threshold should be on __n2. The sets must be kept in this order because semantically elements must be copied
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: added.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: fixed in a similar way

mmichel11 (approved):
I have done another pass over this PR, and it LGTM assuming green CI.

PR #2318, opened 2025-06-20
Partition Set Algorithm Balanced Path 

danhoeflinger:
Add partitioning kernel to set APIs balanced path algorithm.

Adds a partitioning phase which does a sparse pass over the input data to establish binary search boundaries for the main run.  This allows memory access pattern to fit within L1 cache for the main kernels when performing the binary searches to establish balanced path intersections. 

This improves performance for large sizes of the set algorithms.  (When combined with #2317, it provides a nice combination of performance improvements for both large and small sizes of the set algorithms)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -1105,6 +1282,14 @@ __get_reduce_then_scan_actual_sg_sz_device()
 #endif
 }
 
+inline std::uint32_t
+__get_reduce_then_scan_workgroup_size(sycl::queue& q)
```
SergeyKopienko:
```suggestion
__get_reduce_then_scan_workgroup_size(const sycl::queue& q)
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -1105,6 +1282,14 @@ __get_reduce_then_scan_actual_sg_sz_device()
...
+{
+    const std::size_t __max_wg_size = oneapi::dpl::__internal::__max_work_group_size(q);
+    std::size_t __sg_size = __get_reduce_then_scan_workaround_sg_sz();
```
SergeyKopienko:
```suggestion
    const std::uint8_t __sg_size = __get_reduce_then_scan_workaround_sg_sz();
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -533,85 +534,175 @@ struct __get_set_operation<oneapi::dpl::unseq_backend::_SymmetricDifferenceTag<s
...
 
-    // Calculate the number of matchable "future" repeats in the second set
-    _Index __matchable_forward_ele_rng2 = __balanced_path_intersection_rng2 - __merge_path_rng2;
-    _Index __total_matched_rng2 = __balanced_path_intersection_rng2 - __rng2_repeat_start;
+        // diagonal index of the tile begin
+        _SizeType __wg_begin_idx = (__id / __tile_size) * __tile_size;
+        _SizeType __signed_tile_size = static_cast<_SizeType>(__tile_size);
+        //TODO: ensure partitioning fills in last diagonal
+        _SizeType __wg_end_idx = std::min(
+            ((static_cast<_SizeType>(__id) / __signed_tile_size) + 1) * __signed_tile_size, __rng_tmp_diag.size() - 1);
```
SergeyKopienko:
```suggestion
        const _SizeType __wg_begin_idx = (__id / __tile_size) * __tile_size;
        const _SizeType __signed_tile_size = static_cast<_SizeType>(__tile_size);
        //TODO: ensure partitioning fills in last diagonal
        const _SizeType __wg_end_idx = std::min(
            ((static_cast<_SizeType>(__id) / __signed_tile_size) + 1) * __signed_tile_size, __rng_tmp_diag.size() - 1);
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -623,29 +714,87 @@ struct __gen_set_balanced_path
...
+
+        return std::make_tuple(__rng1_balanced_pos, __rng2_balanced_pos, __star);
+    }
 
-        _SizeType __eles_to_process = std::min(__diagonal_spacing - (__star_offset ? _SizeType{1} : _SizeType{0}),
-                                               __rng1.size() + __rng2.size() - (__i_elem - 1));
+    //Entry point for partitioning phase
+    template <typename _InRng, typename _IndexT>
+    void
+    __calc_partition_bounds(const _InRng& __in_rng, _IndexT __id) const
```
SergeyKopienko, line 781: Does this function really needed? May we call `calc_and_store_balanced_path` directly without this function?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -533,85 +534,175 @@ struct __get_set_operation<oneapi::dpl::unseq_backend::_SymmetricDifferenceTag<s
...
-    // find first element of repeating sequence in the first set of the previous element
-    _Index __rng1_repeat_start = oneapi::dpl::__internal::__biased_lower_bound</*__last_bias=*/true>(
-        __rng1, _Index{0}, __merge_path_rng1, __ele_val, __comp);
-    // find first element of repeating sequence in the second set of the next element
-    _Index __rng2_repeat_start = oneapi::dpl::__internal::__biased_lower_bound</*__last_bias=*/true>(
-        __rng2, _Index{0}, __merge_path_rng2, __ele_val, __comp);
+template <typename _Rng, typename _IdxT>
+auto
+__decode_balanced_path_temp_data_no_star(const _Rng& __rng, const _IdxT __id, const std::uint16_t __diagonal_spacing)
+    -> std::tuple<_IdxT, _IdxT>
```
SergeyKopienko:
```suggestion
std::tuple<_IdxT, _IdxT>
__decode_balanced_path_temp_data_no_star(const _Rng& __rng, const _IdxT __id, const std::uint16_t __diagonal_spacing)
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -533,85 +534,175 @@ struct __get_set_operation<oneapi::dpl::unseq_backend::_SymmetricDifferenceTag<s
...
+{
+    return __decode_balanced_path_temp_data_impl<false>(__rng, __id, __diagonal_spacing);
+}
 
-    _Index __rng1_repeats = __merge_path_rng1 - __rng1_repeat_start;
-    _Index __rng2_repeats_bck = __merge_path_rng2 - __rng2_repeat_start;
+template <typename _Rng, typename _IdxT>
+auto
+__decode_balanced_path_temp_data(const _Rng& __rng, const _IdxT __id, const std::uint16_t __diagonal_spacing)
+    -> std::tuple<_IdxT, _IdxT, decltype(__rng.size())>
```
SergeyKopienko:
```suggestion
std::tuple<_IdxT, _IdxT, decltype(__rng.size())>
__decode_balanced_path_temp_data(const _Rng& __rng, const _IdxT __id, const std::uint16_t __diagonal_spacing)
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -1645,34 +1833,36 @@ __future<sycl::event, __result_and_scratch_storage<typename _InitType::__value_t
...
     using _ReduceKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
         __reduce_then_scan_reduce_kernel<_CustomName>>;
     using _ScanKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
         __reduce_then_scan_scan_kernel<_CustomName>>;
     using _ValueType = typename _InitType::__value_type;
 
+    if (__work_group_size == 0)
+    {
+        __work_group_size = __get_reduce_then_scan_workgroup_size(__q);
+    }
```
SergeyKopienko:
```suggestion
    __work_group_size = std::max(__work_group_size, __get_reduce_then_scan_workgroup_size(__q));
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -670,22 +819,50 @@ struct __gen_set_op_from_known_balanced_path
...
+template <typename _GenInput, typename... _KernelName>
+struct __partition_set_balanced_path_submitter<_GenInput, __internal::__optional_kernel_name<_KernelName...>>
+{
+    template <typename _InRng>
+    sycl::event
+    operator()(sycl::queue& __q, _InRng&& __in_rng, std::size_t __num_diagonals) const
+    {
+        std::size_t __tile_size = __gen_input.__get_bounds.__tile_size;
+        std::size_t __n = oneapi::dpl::__internal::__dpl_ceiling_div(__num_diagonals + __tile_size - 1, __tile_size);
+        return __q.submit([&__in_rng, this, __tile_size, __n, __num_diagonals](sycl::handler& __cgh) {
```
SergeyKopienko: Is it good practice to capture `this` or may be better simple capture `__gen_input` instead?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -1105,6 +1282,14 @@ __get_reduce_then_scan_actual_sg_sz_device()
 #endif
 }
 
+inline std::uint32_t
+__get_reduce_then_scan_workgroup_size(sycl::queue& q)
+{
+    const std::size_t __max_wg_size = oneapi::dpl::__internal::__max_work_group_size(q);
+    std::size_t __sg_size = __get_reduce_then_scan_workaround_sg_sz();
+    return (__max_wg_size / __sg_size) * __sg_size;
```
SergeyKopienko:
```suggestion
    return __max_wg_size - __max_wg_size % __sg_size;
```
 - probably this variant looks more readable.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -1751,7 +1941,21 @@ __parallel_transform_reduce_then_scan(sycl::queue& __q, const std::size_t __n, _
...
         }
     }
-    return __future{std::move(__event), std::move(__result_and_scratch)};
+    return __future{std::move(__prior_event), std::move(__result_and_scratch)};
+}
+
+template <typename _CustomName, typename _InRng, typename _GenReduceInput>
+sycl::event
+__parallel_set_balanced_path_partition(sycl::queue& __q, _InRng&& __in_rng, std::size_t __num_diagonals,
+                                       _GenReduceInput __gen_reduce_input)
```
SergeyKopienko:
```suggestion
__parallel_set_balanced_path_partition(sycl::queue& __q, const _InRng& __in_rng, std::size_t __num_diagonals,
                                       _GenReduceInput __gen_reduce_input)
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -670,22 +819,50 @@ struct __gen_set_op_from_known_balanced_path
...
+struct __partition_set_balanced_path_submitter;
+template <typename _GenInput, typename... _KernelName>
+struct __partition_set_balanced_path_submitter<_GenInput, __internal::__optional_kernel_name<_KernelName...>>
+{
+    template <typename _InRng>
+    sycl::event
+    operator()(sycl::queue& __q, _InRng&& __in_rng, std::size_t __num_diagonals) const
+    {
+        std::size_t __tile_size = __gen_input.__get_bounds.__tile_size;
+        std::size_t __n = oneapi::dpl::__internal::__dpl_ceiling_div(__num_diagonals + __tile_size - 1, __tile_size);
```
SergeyKopienko:
```suggestion
        const std::size_t __tile_size = __gen_input.__get_bounds.__tile_size;
        const std::size_t __n = oneapi::dpl::__internal::__dpl_ceiling_div(__num_diagonals + __tile_size - 1, __tile_size);
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1058,34 +1057,53 @@ __parallel_set_reduce_then_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __
...
+    const std::uint32_t __work_group_size = __get_reduce_then_scan_workgroup_size(__q);
+    const std::size_t __partition_threshold = 2 * 1024 * 1024;
+    const std::size_t __total_size = __rng1.size() + __rng2.size();
     // Should be safe to use the type of the range size as the temporary type. Diagonal index will fit in the positive
     // portion of the range so star flag can use sign bit.
     using _TemporaryType = decltype(__rng1.size());
     //TODO: limit to diagonals per block, and only write to a block based index of temporary data
     oneapi::dpl::__par_backend_hetero::__buffer<_TemporaryType> __temp_diags(__num_diagonals);
 
+    constexpr std::uint32_t __average_input_ele_size = (sizeof(_In1ValueT) + sizeof(_In2ValueT)) / 2;
```
SergeyKopienko, line 1083: I think sometimes this approach may give us not the best performance.
When?
When not only the sizes of the data types but also the sizes of the data differ greatly.
So may be we can improve this by normalizing the data sizes based on the number of data elements somehow.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1058,34 +1057,53 @@ __parallel_set_reduce_then_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __
...
+    // Partition into blocks of half SLM size
```
mmichel11: Is this an empirically determined value and did you experiment with larger partition sizes?

My understanding is that on most architectures SLM is carved out of L1. Since we are trying to cache in L1, it may be possible to actually partition larger than the max SLM size and still have good caching so long as the L1 is large enough. Of course this is device specific and we do not want to overtune for a single device.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1058,34 +1057,53 @@ __parallel_set_reduce_then_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __
...
+        oneapi::dpl::__par_backend_hetero::__gen_set_balanced_path<_SetOperation, _BoundsProvider, _Compare>;
     using _GenScanInput =
         oneapi::dpl::__par_backend_hetero::__gen_set_op_from_known_balanced_path<_SetOperation, _TempData, _Compare>;
     using _ScanInputTransform = oneapi::dpl::__par_backend_hetero::__get_zeroth_element;
     using _WriteOp = oneapi::dpl::__par_backend_hetero::__write_multiple_to_id<oneapi::dpl::__internal::__pstl_assign>;
 
     const std::int32_t __num_diagonals =
         oneapi::dpl::__internal::__dpl_ceiling_div(__rng1.size() + __rng2.size(), __diagonal_spacing);
-
+    const std::uint32_t __work_group_size = __get_reduce_then_scan_workgroup_size(__q);
```
mmichel11: Why do we calculate the work group size here instead of within `__parallel_transform_reduce_then_scan`? It looks like to me we can do this in `__parallel_transform_reduce_then_scan` and remove its `__work_group_size` parameter.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -533,85 +534,175 @@ struct __get_set_operation<oneapi::dpl::unseq_backend::_SymmetricDifferenceTag<s
...
+        // If we needed to step off the diagonal to find the balanced path, mark the diagonal as "starred"
+        __star = __balanced_path_intersection_rng1 + __balanced_path_intersection_rng2 !=
+                 __merge_path_rng1 + __merge_path_rng2;
+
+        return std::make_tuple(__balanced_path_intersection_rng1, __balanced_path_intersection_rng2, __star);
+    }
+
+    template <typename _InRng, typename _IndexT, typename _BoundsProviderLocal>
+    std::tuple<_IndexT, _IndexT, bool>
+    calc_and_store_balanced_path(const _InRng& __in_rng, _IndexT __id, _BoundsProviderLocal __get_bounds_local) const
```
mmichel11: It's not incorrect for `__in_rng` to be const here, but it confused me at first since `__rng1_temp_diag` comes from the third component of the zip view and is modified. Is making `__in_rng` non-const potentially clearer?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: We need this to be `constexpr` to influence `__max_inputs_per_item` in the template arguments. Therefore, I think this is about the best we can do.  I think the most common case is that we have the same type and similar sizes for the two sets.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: I played with a few options here and this seemed to be the best option empirically.  I can revisit and confirm now with a more final version though.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: My intention here is for encapsulation of details of partitioning into this struct.  The external caller doesn't have to know about sending in `__get_bounds_simple{}` here, but rather can just call an interface with the information it has.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: I think its OK to capture `this`. We also do it in reduce_then_scan submitters.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: incorporated.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: Good point, switched to non const lvalue ref

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: incorporated.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: This has to be an auto with trailing return type because of the reliance on the input argument with decltype, unless we can derive it from the range type itself.

danhoeflinger: Eh, nevermind.  I fixed this with a `std::declval<_Rng>()`.  This is probably better than making it a trailing type.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: incorporated.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: Its a good question.  I originally pulled this out to use in the calculation of partitioning size.  
I'm going to explore partition size a bit more.  If it is not useful then I will fold it back in to `__parallel_transform_reduce_then_scan`.
Thanks for calling this out.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: ok, up to you

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: I'm not sure I understand why you want to change this to const lvalue ref rather than forwarding.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: sure, done.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: I've added a comment, because I think your change obfuscates the point of this code.  
I may remove this entirely and fold this function back in, but still need to re-investigate partitioning size first

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
SergeyKopienko: As far as I understand, `__in_rng` describes some source data, which is `const` and can't be modified inside this function. So I think better to describe this semantic by `const ref` instead of universal reference.
As evidence of this point, let's take a look how you call `__parallel_set_balanced_path_partition`:
```C++
    if (__total_size >= __partition_threshold)
    {
        __partition_event =
            __parallel_set_balanced_path_partition<_CustomName>(__q, __in_rng, __num_diagonals, __gen_reduce_input);
    }
```
What happens inside:
```C++
template <typename _CustomName, typename _InRng, typename _GenReduceInput>
sycl::event
__parallel_set_balanced_path_partition(sycl::queue& __q, _InRng&& __in_rng, std::size_t __num_diagonals,
                                       _GenReduceInput __gen_reduce_input)
{
    //...
    return __partition_submitter(__q, std::forward<_InRng>(__in_rng), __num_diagonals);
}
```

```C++
template <typename _GenInput, typename _KernelName>
struct __partition_set_balanced_path_submitter;
template <typename _GenInput, typename... _KernelName>
struct __partition_set_balanced_path_submitter<_GenInput, __internal::__optional_kernel_name<_KernelName...>>
{
    template <typename _InRng>
    sycl::event
    operator()(sycl::queue& __q, _InRng&& __in_rng, std::size_t __num_diagonals) const
    {
        //...
                    __gen_input.__calc_partition_bounds(__in_rng, __id);
        //...
    }
    //...
};
```
And then we are in
```C++
    //Entry point for partitioning phase
    template <typename _InRng, typename _IndexT>
    void
    __calc_partition_bounds(const _InRng& __in_rng, _IndexT __id) const
    {
        calc_and_store_balanced_path(__in_rng, __id, oneapi::dpl::__par_backend_hetero::__get_bounds_simple{});
    }
```
So I think this chain of calls shown us that universal reference for `__in_rng` doesn't make sense and `const _InRng& __in_rng` is quite correct inside all this chain of calls.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
SergeyKopienko: Ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
SergeyKopienko: If you want, you may take a look to the PR https://github.com/uxlfoundation/oneDPL/pull/2346 and merge it into your branch.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: It turns out the performance is not incredibly sensitive to the partitioning size, testing between half and quadruple the size here in this PR (minor changes in performance across those sizes).  At 20x the size here we see some non-trivial slowdowns on our target platform, so there is a limit to this, and we do see significant benefit over no partitioning.  Empirically, this is the best value I've found, and it ties it to a value which is related to the L1 size of the card.  I'll add a comment to the affect.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: ive reverted these changes.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: This range actually is a zip of 2 inputs and a third sequence which is the temporary storage to write the balanced path partition indices as output of this partitioning operation.
I format this into a zip to match the structure of the main kernels, and to be able to easily reuse those methods.  
I don't believe I can mark this as `const`.

I will to rename the range for clarity.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -673,22 +819,51 @@ struct __gen_set_op_from_known_balanced_path
...
+struct __partition_set_balanced_path_submitter<_GenInput, __internal::__optional_kernel_name<_KernelName...>>
+{
+    template <typename _InInOutRng>
+    sycl::event
+    operator()(sycl::queue& __q, _InInOutRng&& __in_in_out_rng, std::size_t __num_diagonals) const
+    {
+        const std::size_t __tile_size = __gen_input.__get_bounds.__tile_size;
+        const std::size_t __n =
+            oneapi::dpl::__internal::__dpl_ceiling_div(__num_diagonals + __tile_size - 1, __tile_size);
+        return __q.submit([&__in_in_out_rng, this, __n, __num_diagonals](sycl::handler& __cgh) {
```
mmichel11, line 897: Since we are capturing `this` and we have a device copyable specialization for `__gen_set_balanced_path` which is a member of this class, I think we may need a device copyable specialization for `__partition_set_balanced_path_submitter` as well.

danhoeflinger: good catch.  fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -536,85 +537,172 @@ struct __get_set_operation<oneapi::dpl::unseq_backend::_SymmetricDifferenceTag<s
...
+{
+    return __decode_balanced_path_temp_data_impl<true>(__rng, __id, __diagonal_spacing);
+}
 
-    if (__rng2_repeats_bck >= __rng1_repeats)
-    {
-        // If we have at least as many repeated elements in rng2, we end up back on merge path
-        return std::make_tuple(__merge_path_rng1, __merge_path_rng2, false);
-    }
+//TODO: check on types here
```
mmichel11: Is there any further action here or can this todo be removed?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -536,85 +537,172 @@ struct __get_set_operation<oneapi::dpl::unseq_backend::_SymmetricDifferenceTag<s
...
+{
+    template <typename _Rng, typename _IndexT>
+    auto // Returns a tuple of the form (start1, end1, start2, end2)
+    operator()(const _Rng& __in_rng, const _IndexT __id) const
+    {
+        auto __rng_tmp_diag = std::get<2>(__in_rng.tuple()); // set a temp storage sequence
 
-    // Calculate the max location to search in the second set for future repeats, limiting to the edge of the range
-    _Index __fwd_search_bound = std::min(__merge_path_rng2 + __fwd_search_count, __rng2.size());
+        using _SizeType = decltype(std::get<0>(__in_rng.tuple()).size());
```
mmichel11, line 632: No necessary action on your side here, but I see we use something along the lines of `decltype(__rng.size()))` frequently in this patch. It looks like we have a `oneapi::dpl::__internal::__range_size_t` but only for C++20 or later. Perhaps it would be worth extending this for C++17 with internal ranges.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -626,29 +714,87 @@ struct __gen_set_balanced_path
...
+            return 0;
+        if (!__is_partitioned)
+        {
+            auto [__idx_rng1, __idx_rng2, __local_star] =
+                calc_and_store_balanced_path(__in_rng, __id, __get_bounds_simple{});
+            __rng1_balanced_pos = __idx_rng1;
+            __rng2_balanced_pos = __idx_rng2;
+            __star = __local_star;
+        }
+        else if (__id % __get_bounds.__tile_size != 0)
```
mmichel11, line 815: I think a comment on how the balanced path calculations for each tile in the partition phase map to the balanced path computations in this kernel would be helpful. To me the details of the indexing can be a bit tricky to understand from just looking at the code.

mmichel11 (commented):
This PR looks near ready to me. The conflicts with main need to be addressed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: thanks for catching this.  I've made some adjustments to turn auto-> explicit types

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: If I'm remembering all this properly, there is no change or mapping of in indexes.  Here we are just checking if are:
1) in a non-partitioned case, and if so using the bounds of the range to limit our search for balanced path intersection
2) in a partitioned case, but not on a boundary which has already been established (use the already calculated boundaries to limit search for balanced path intersection
3) in a partitioned case and on a boundary where the balanced path intersection has already been calculated, if so use it.

I've added some more comments to hopefully clarify.

mmichel11: Thanks, I understand it now. My confusion about the `__id % __get_bounds.__tile_size != 0` line. It's clear after comparing the nd-ranges on the partition and reduce-then-scan kernels.

test/general/implementation_details/device_copyable.pass.cpp
```diff
@@ -204,9 +205,18 @@ test_device_copyable()
     //__gen_set_balanced_path
     static_assert(sycl::is_device_copyable_v<oneapi::dpl::__par_backend_hetero::__gen_set_balanced_path<
                       oneapi::dpl::__par_backend_hetero::__set_intersection,
+                      oneapi::dpl::__par_backend_hetero::__get_bounds_simple,
                       binary_op_device_copyable>>,
                   "__gen_set_balanced_path is not device copyable with device copyable types");
 
+    //__partition_set_balanced_path_submitter
+    static_assert(sycl::is_device_copyable_v<oneapi::dpl::__par_backend_hetero::__partition_set_balanced_path_submitter<
```
mmichel11, line 214: Let's add the guards introduced in https://github.com/uxlfoundation/oneDPL/pull/2414 around both static asserts for the submitters.

danhoeflinger: done, thanks

mmichel11 (approved):
LGTM

mmichel11 (dismissed):
Reapproving

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
SergeyKopienko: ok

SergeyKopienko (approved):
LGTM

SergeyKopienko (approved):
LGTM

mmichel11 (approved):

PR #2319, opened 2025-06-22
Workaround icpx incompatibility issues with VS 17.14

dmitriy-sobolev:
icpx 2025.1 and older have compatibility issues with VS 17.14, which was recently upgraded in the windows-2022 (windows-latest) image.


mmichel11 (approved):
LGTM

PR #2320, opened 2025-06-23
Add set parallel range algorithms

MikeDvorskiy:
This PR adds the following range-based algorithms: `includes`, `set_union`, `set_intersection`, `set_difference`, `set_symmetric_difference`.

Adding these algorithms required passing projections across the whole call chain and adding an extensive testing.
This is because they have two projections, and the patterns they are based on can freely swap the arguments in their comparators (e.g. `comp(a, b)` and `comp(b, a)`).  Extra testing coverage allowed to find (hopefully) all such cases.

Useful links:
- [Specification](https://github.com/uxlfoundation/oneAPI-spec/pull/630): interfaces.
- [RFC](https://github.com/uxlfoundation/oneDPL/pull/2269): rationale and additional requirements.

The algorithms are implemented with a limitation: an output range must be capable of holding all the result elements.

There is one arguable question whether we should reimplement the serial set algorithms with libc++ or not. This library has a bug leading to wrong results with different projections. See:
```c++
// There is a bug in the libc++ at the time of writing this comment with 21 being the latest major release
// 23 is set to avoid frequent bump-ups.
// See: https://github.com/llvm/llvm-project/blob/6096d35ea93c75f648a253a00775b4d74915c819/libcxx/include/__algorithm/ranges_set_union.h#L94
// This line does not take into account that the iterator-based implementation may arbitrary call comp(a, b) or comp(b, a)
```


include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3377,20 +3390,24 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
     const auto __n1 = __last1 - __first1;
     const auto __n2 = __last2 - __first2;
 
+    auto __proj1_deref = [__proj1](_RandomAccessIterator1 __it) { return std::invoke(__proj1, *__it); };
+    auto __proj2_deref = [__proj2](_RandomAccessIterator2 __it) { return std::invoke(__proj2, *__it); };
```
danhoeflinger: Maybe we want to abstract the lambda out to a struct.  Its used at least twice here and once above.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -712,10 +712,11 @@ template <typename _Inclusive, typename _BinaryOperation, typename _UnaryOp, typ
           typename _GlobalAssigner, typename _DataAccessor, typename _InitType>
 struct __scan;
 
-template <typename _Compare, typename _Size1, typename _Size2>
+template <typename _Compare, typename _Size1, typename _Size2, typename _Proj1, typename _Proj2>
```
danhoeflinger: This no longer matches the updated ordering of template arguments, need to fix throughout this file.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -987,54 +987,57 @@ struct __scan
 // __brick_includes
 //------------------------------------------------------------------------
 
-template <typename _Compare, typename _Size1, typename _Size2>
+template <typename _Size1, typename _Size2, typename _Compare, typename _Proj1 = oneapi::dpl::identity,
+          typename _Proj2 = oneapi::dpl::identity>
 struct __brick_includes
 {
+    _Size2 __size1;
+    _Size1 __size2;
```
danhoeflinger: Seems like the type and variable name are flipped here.

include/oneapi/dpl/pstl/parallel_backend_utils.h
```diff
@@ -279,57 +286,66 @@ __set_intersection_construct(_ForwardIterator1 __first1, _ForwardIterator1 __las
...
 _OutputIterator
 __set_difference_construct(_ForwardIterator1 __first1, _ForwardIterator1 __last1, _ForwardIterator2 __first2,
                            _ForwardIterator2 __last2, _OutputIterator __result, _Compare __comp,
-                           _CopyConstructRange __cc_range)
+                           _CopyConstructRange __cc_range, _Proj1 __proj1 = {}, _Proj2 __proj2 = {})
 {
     using _Tp = typename ::std::iterator_traits<_OutputIterator>::value_type;
 
+    auto __proj1_deref = [__proj1](_ForwardIterator1 __it) { return std::invoke(__proj1, *__it); };
+    auto __proj2_deref = [__proj2](_ForwardIterator2 __it) { return std::invoke(__proj2, *__it); };
```
danhoeflinger: more places for the projected deref callable struct here and later in this file.

include/oneapi/dpl/pstl/algorithm_impl.h
dmitriy-sobolev: Mind you, this is only relevant for the device code while this is a host backend.

include/oneapi/dpl/pstl/parallel_backend_utils.h
dmitriy-sobolev: This is used only in the host backends, so using structs is not necessary.

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: yes, I understand its not required for the ExecutionPolicy value category stuff since its host side code.  

My comment is mostly just so we don't have 5+ implementations of the same lambda.  I'm not sure how likely it is that we would want to add something to the implementation, but if that ever comes up it would be better to have one place to do that. 

Its a nitpick, and I wont insist though.

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: I thinks about this earlier and realize that it's not good idea to pack these lambdas into struct.
The reason is in the calling `operator*` from iterator inside lambda: it may return temporary value.
And we have some problem: we should extend their life-time: which type (value category) we will return in this case from potential `struct DerefIT::operator()` ? Already returns it by value?
So I think these lambdas here is good enough solution.

include/oneapi/dpl/pstl/parallel_backend_utils.h
SergeyKopienko: Please see my comment at https://github.com/uxlfoundation/oneDPL/pull/2320#discussion_r2303228180

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: I'm not seeing how this is different between lambda and struct.  Cant we use `decltype(auto)` as return, and it will use whatever type (including value category) the result of the dereference operation is?

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -987,54 +987,57 @@ struct __scan
...
-        auto __a = __a_acc;
-        auto __b = __b_acc;
-
-        auto __a_beg = _Size1(0);
-        auto __a_end = __na;
+        using std::get;
 
-        auto __b_beg = _Size2(0);
-        auto __b_end = __nb;
+        auto __proj_rng_1 = [this](auto&& __val) { return std::invoke(__proj1, std::forward<decltype(__val)>(__val)); };
```
mmichel11: Ares these lambdas problematic for the multiple kernels issue? Should they be made into member functions instead?

danhoeflinger: I believe technically it is only problematic when a template argument may vary without a different kernel being generated (for instance ExecutionPolicy with a different value category).  However, I think we have generally said that to be safe, we should avoid lambdas in sycl backend code.

SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/parallel_backend_utils.h
SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/algorithm_impl.h

include/oneapi/dpl/pstl/utils.h
```diff
@@ -126,6 +128,34 @@ class __pstl_assign
...
+{
+    //'mutable' is to relax the requirements for a user comparator or/and projection type operator() may be non-const
+    mutable _Proj __proj;
+
+    template <typename _Iterator>
+    auto
+    operator()(_Iterator __it) const
+    {
+        return std::invoke(__proj, *__it);
+    }
```
akukanov: Will not we again face the problem of returning a dangling reference with projections like `std::identity`?

SergeyKopienko: No, I don't think so: we return `auto` here, so it's the instance.
@danhoeflinger proposed to avoid multiple lambdas in sycl backend code.
Please see https://github.com/uxlfoundation/oneDPL/pull/2320#discussion_r2304074508 and https://github.com/uxlfoundation/oneDPL/pull/2320#discussion_r2301774878

danhoeflinger: Even if the projection is `std::identity`, or `&Class::member_field`, it will be copied on the way out of this operator by value, before the possibly temporary object (like dereferenced `zip_iterator` ) leaves scope.

As far as I understand, this is no different than a lambda without explicit return type specified. 

If that extra copy of the projected value is a problem here then I think it is also a problem with the lambda version of the same.

SergeyKopienko: What about restore all `std::invoke` instead of these two new structures?
I don't like this option but it's available...

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: To clarify, after offline discussion, `decltype(auto)` is in fact problematic for types like `zip_iterator` which when dereferenced return temporary proxy objects when combined with projections like `&Class::member_field` which depend upon the object before projection staying in scope.  

The temporary would be destroyed before the projected value was used.  Therefore, we came to the conclusion that `auto` is the appropriate return, even though it copies the projected output by value. As far as I understand, in the ways which matter, a struct returning `auto` is the same as a lambda without an explicit return type.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: If I understand correctly, that would allow us to avoid the copy, but with significantly messier code, right?

danhoeflinger: An idea to potentially remove unnecessary copies:

Could we make something which checks the value category of `decltype(*it)`, and if it is a lvalue ref, we return by decltype(auto) or just the explicit lvalue ref type, and returns `auto` otherwise?  

I'm not sure if that causes problems for a `zip_iterator` wrapped inside a `transform_iterator` or some other combination of iterator adapters...

SergeyKopienko: I am already thinks about this valiant before and tried to implement it.
But it was too complicated for this simple structure.
After discussion with @akukanov I declined this idea...

SergeyKopienko: We may remember about this place as about potential point of performance improvement and save as is for now to have more readable code.

akukanov: Besides extra copies, do not we also disallow predicates that take arguments by lvalue references, or just by the iterator reference type?

SergeyKopienko: I think we really disallow predicates that take arguments by l-value references into compare functions, bet we have the same behavior here like in lambdas variants. And in our tests we hadn't any problems with that in our test use-cases. So I think we have two options:
- always use `std::invoke` instead these two predicated and have not clear and bulky code;
- make these two predicates more smarter like proposed @danhoeflinger at https://github.com/uxlfoundation/oneDPL/pull/2320#discussion_r2304916303 and as I thinks before and discussed offline with @akukanov.
@akukanov, @danhoeflinger, @dmitriy-sobolev - what is your opinion?

SergeyKopienko: One of possible solutions implemented in the PR https://github.com/uxlfoundation/oneDPL/pull/2406
@akukanov , @danhoeflinger , @dmitriy-sobolev please take a look.

All CI tests passed for this solution: https://github.com/uxlfoundation/oneDPL/pull/2405/checks

SergeyKopienko: The PR https://github.com/uxlfoundation/oneDPL/pull/2406 has been merged into main branch.

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: The PR #2406 has been merged into this branch.
So I think this comment may be resolved.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3069,8 +3069,10 @@ ___merge_path_out_lim(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _It1
                         auto __r = __get_row(__d);
                         auto __c = __get_column(__d);
 
-                        const auto __res = std::invoke(__comp, std::invoke(__proj2, __it_2[__c]),
-                                                       std::invoke(__proj1, __it_1[__r])) ? 0 : 1;
+                        const auto __res =
```
akukanov: Looks like a formatting-only change. Is it necessary? The new formatting is not objectively better.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -627,9 +655,20 @@ __dpl_signbit(const _T& __x)
     return (__x & __mask) != 0;
 }
 
-template <typename _Acc, typename _Size1, typename _Value, typename _Compare>
+// Adapts __pstl_lower_bound and other related functions to use with host backends
+struct _SubscriptAdapter
+{
+    template <typename _Iterator>
+    decltype(auto)
+    operator[](_Iterator __it) const
```
akukanov: Why do we need a weird operator that pretends to do indexing but takes an iterator as the argument? To me, it screams as an improper workaround for an architecture problem.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -126,6 +128,32 @@ class __pstl_assign
...
+            return std::invoke(__proj, *__it);
+        }
+        else
+        {
+            using __proj_result_t = decltype(std::invoke(__proj, *__it));
+            using __proj_result_decayed_t = std::decay_t<__proj_result_t>;
+
+            // create temporary copy of the projection and return it as prvalue
+            // - for that we should create and return temporary object (but not local variable)
+            return __proj_result_decayed_t{std::invoke(__proj, *__it)};
```
akukanov: Is it guaranteed that the type is copyable?

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -796,6 +796,153 @@ struct __merge_fn
...
+    template <typename _ExecutionPolicy, std::ranges::random_access_range _R1, std::ranges::random_access_range _R2,
+              std::ranges::random_access_range _OutRange, typename _Comp = std::ranges::less,
+              typename _Proj1 = std::identity, typename _Proj2 = std::identity>
+        requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<_ExecutionPolicy>> &&
+                 std::ranges::sized_range<_R1> && std::ranges::sized_range<_R2> &&
+                 std::ranges::sized_range<_OutRange> &&
+                 std::mergeable<std::ranges::iterator_t<_R1>, std::ranges::iterator_t<_R2>,
+                                std::ranges::iterator_t<_OutRange>, _Comp, _Proj1, _Proj2>
+
+    std::ranges::set_difference_result<std::ranges::borrowed_iterator_t<_R1>,
```
danhoeflinger, line 899: I think this is correct and the oneDPL spec needs a bugfix for `set_difference_result`, which includes the `borrowed_iterator_t` for `_R2` as well.

https://github.com/uxlfoundation/oneAPI-spec/pull/630/files#r2308175558

include/oneapi/dpl/pstl/utils.h
danhoeflinger: It also requires that we pass iterator types into template arguments called `_Size[1,2]`.  Since we have that these are random access ranges, can we not just pass them or their beginning random access iterator in to `__pstl_lower_bound` as the accessor and use 0 and the size as the limits?

include/oneapi/dpl/pstl/utils.h
danhoeflinger: Its a good question.  Perhaps it is best just to use `std::invoke(__proj1,*it)` inline everywhere instead of `__proj1_deref(it)` after the declaration of `oneapi::dpl::__internal::__projection_deref<_Proj1> __proj1_deref{__proj1};`.  

All of the complexity of the lambda or struct including dealing with this question of copyability of the projected result is to try to handle different return type value categories and lifetime extension.  If we just do it inline, I think there should be no problems in this regard, and the complexity of the code isn't that much higher. Perhaps any complexity we save in consolidation here is created by its existence in the first place.

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: This is our clang format as far as I understand...

akukanov: We decided a while ago that there is no requirement to follow every clang-format suggestion.

And the change is out of scope for the patch anyway. If you think it should be done, there is certainly more to improve around it, like the abundance of auto variables.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: As far as I understand, it will work in the case if copy disabled too.
Example: https://godbolt.org/z/jsfoq9YsM
Justification: https://en.cppreference.com/w/cpp/language/copy_elision.html :

_When a class object target is copy-initialized with a temporary class object obj that has not been bound to a reference, the copy-initialization can be omitted by constructing obj directly into target. This variant of copy elision is known as unnamed return value optimization (**URVO**). Since **C++17**, **URVO** is mandatory and no longer considered a form of copy elision; see below._

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: Fixed: no changed here anymore in this PR.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: The PR created: https://github.com/uxlfoundation/oneDPL/pull/2409
Please take a look.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Alternative approach prepared in separate PR https://github.com/uxlfoundation/oneDPL/pull/2412

danhoeflinger: Thanks for the investigation Sergey.  I think you are correct that this works with copy disabled.  That being said, I think I prefer inlining `std::invoke` with the explicit dereference.  It lessens our need to consider these complicated topics, and I think the lines with `std::invoke` and explicit dereference are not much more complicated ultimately.

SergeyKopienko: The PR https://github.com/uxlfoundation/oneDPL/pull/2412 has been merged into this branch.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: The PR https://github.com/uxlfoundation/oneDPL/pull/2412 has been merged into this branch.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
dmitriy-sobolev: Thanks. It's being addressed in https://github.com/uxlfoundation/oneAPI-spec/pull/637.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -639,7 +641,7 @@ __pstl_lower_bound(_Acc __acc, _Size1 __first, _Size1 __last, const _Value& __va
         __it = __first;
         __cur = __n / 2;
         __it += __cur;
-        if (__comp(__acc[__it], __value))
+        if (__compareOp(__it))
```
akukanov: This line I would change to `if (std::invoke(__comp, std::invoke(__proj, __acc[__it]), __value))` with `__proj` being the new parameter for a projection.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3315,18 +3317,19 @@ __parallel_set_op(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
...
+                    __e = __internal::__pstl_upper_bound(__e, __last1, std::invoke(__proj1, *__e), __comp, __proj1);
 
                 //check is [__b; __e) empty
                 if (__e - __b < 1)
                 {
                     _RandomAccessIterator2 __bb = __last2;
                     if (__b != __last1)
-                        __bb = ::std::lower_bound(__first2, __last2, *__b, __comp);
+                        __bb = __internal::__pstl_lower_bound(__first2, __last2, std::invoke(__proj1, *__b), __comp,
+                                                              __proj2);
```
akukanov: This call would change to `__pstl_lower_bound(__first2, 0, __last2 - __first2, std::invoke(__proj1, *__b), __comp, __proj2)`.

And etc.

akukanov (commented):
I still do not like the updated implementation of `__pstl_lower/upper_bound`; moreover, I find the new overloads excessive.

I understand the need to replace `std::lower_bound` which has no projections with our own implementation. However the current one that takes an "accessor" and two indexes should be suitable - because random access iterators support `operator[]` which result is equivalent to advancing and dereferencing the iterator. The only change really necessary is to add projections and the use of `std::invoke`.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: I am working on this fix in the PR https://github.com/uxlfoundation/oneDPL/pull/2425

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: I am working on this fix in the PR https://github.com/uxlfoundation/oneDPL/pull/2425

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -1293,11 +1303,15 @@ class __brick_set_op
...
             const _Size1 __count_a_left =
-                __idx_a - __internal::__pstl_left_bound(__a, _Size1(0), _Size1(__idx_a), __val_a, __comp) + 1;
-
-            const _Size2 __count_b = __internal::__pstl_right_bound(__b, _Size2(__res), __nb, __val_b, __comp) - __res +
-                                     __res -
-                                     __internal::__pstl_left_bound(__b, _Size2(0), _Size2(__res), __val_b, __comp);
+                __idx_a -
+                __internal::__pstl_left_bound(__a, _Size1(0), _Size1(__idx_a), std::invoke(__proj1, __val_a), __comp,
+                                              __proj1) +
+                1;
```
akukanov: What a silly and confusing formatting. Why not something like this?
```
            const _Size1 __count_a_left = __idx_a - __internal::__pstl_left_bound(__a,
                _Size1(0), _Size1(__idx_a), std::invoke(__proj1, __val_a), __comp, __proj1) + 1;
```

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -1249,23 +1251,29 @@ struct _SymmetricDifferenceTag : public std::true_type
...
+    _Compare __comp;
+    _Proj1 __proj1;
+    _Proj2 __proj2;
 
   public:
-    __brick_set_op(_Compare __c, _Size1 __n1, _Size2 __n2) : __comp(__c), __na(__n1), __nb(__n2) {}
+    __brick_set_op(_Size1 __na, _Size2 __nb, _Compare __comp, _Proj1 __proj1 = {}, _Proj2 __proj2 = {})
+        : __na(__na), __nb(__nb), __comp(__comp), __proj1(__proj1), __proj2(__proj2)
+    {
+    }
```
akukanov:
```suggestion
    __brick_set_op(_Size1 __na, _Size2 __nb, _Compare __comp, _Proj1 __proj1 = {}, _Proj2 __proj2 = {})
        : __na(__na), __nb(__nb), __comp(__comp), __proj1(__proj1), __proj2(__proj2) {}
```
I know clang-format does not like this but who cares. An empty constructor taking two lines while all the useful stuff takes just as much makes no sense to me.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
dmitriy-sobolev: [35f93bc](https://github.com/uxlfoundation/oneDPL/pull/2320/commits/35f93bcb34e78e2047d3e776746ab137b5f50eaa)

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
dmitriy-sobolev: [35f93bc](https://github.com/uxlfoundation/oneDPL/pull/2320/commits/35f93bcb34e78e2047d3e776746ab137b5f50eaa)

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: The PR https://github.com/uxlfoundation/oneDPL/pull/2425 has been merged into this branch.

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: The PR https://github.com/uxlfoundation/oneDPL/pull/2425 has been merged into this branch.

SergeyKopienko:
> I still do not like the updated implementation of `__pstl_lower/upper_bound`; moreover, I find the new overloads excessive.
> 
> I understand the need to replace `std::lower_bound` which has no projections with our own implementation. However the current one that takes an "accessor" and two indexes should be suitable - because random access iterators support `operator[]` which result is equivalent to advancing and dereferencing the iterator. The only change really necessary is to add projections and the use of `std::invoke`.

The PR https://github.com/uxlfoundation/oneDPL/pull/2425 has been merged into this branch.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -659,15 +665,17 @@ struct __get_bounds_simple
         const auto& __rng1 = std::get<0>(__in_rng.tuple()); // first sequence
         const auto& __rng2 = std::get<1>(__in_rng.tuple()); // second sequence
 
-        using _SizeType = decltype(__rng1.size());
-        return std::make_tuple(_SizeType{0}, __rng1.size(), _SizeType{0}, __rng2.size());
+        using _SizeType = std::common_type_t<std::ranges::range_size_t<decltype(__rng1)>, std::ranges::range_size_t<decltype(__rng2)>>;
```
danhoeflinger, line 668:
```suggestion
        using _SizeType = std::common_type_t<std::make_unsigned_t<decltype(__rng1.size())>, std::make_unsigned_t<decltype(__rng2.size())>>;
```

We probably should make a `__dpl_ranges_size_t<R>`, but I'm not sure if we want to try to do that in this PR.  It probably deserves its own PR after the milestone.  For now, we need to do it like this to avoid c++17 issues.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3526,13 +3543,14 @@ __pattern_set_union(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __exec,
 
     // use serial algorithm
     if (__n1 + __n2 <= __set_algo_cut_off)
-        return ::std::set_union(__first1, __last1, __first2, __last2, __result, __comp);
+        return std::set_union(__first1, __last1, __first2, __last2, __result, __comp);
 
-    typedef typename ::std::iterator_traits<_OutputIterator>::value_type _Tp;
+    using _Tp = typename ::std::iterator_traits<_OutputIterator>::value_type;
```
akukanov:
```suggestion
    using _Tp = typename std::iterator_traits<_OutputIterator>::value_type;
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1107,16 +1108,17 @@ __parallel_set_reduce_then_scan(sycl::queue& __q, _Range1&& __rng1, _Range2&& __
     }
     return __parallel_transform_reduce_then_scan<__bytes_per_work_item_iter, _CustomName>(
         __q, __num_diagonals, std::move(__in_in_tmp_rng), std::forward<_Range3>(__result), __gen_reduce_input,
-        _ReduceOp{}, _GenScanInput{_SetOperation{}, __diagonal_spacing, __comp}, _ScanInputTransform{}, _WriteOp{},
-        oneapi::dpl::unseq_backend::__no_init_value<_Size>{}, /*_Inclusive=*/std::true_type{},
+        _ReduceOp{}, _GenScanInput{_SetOperation{}, __diagonal_spacing, __comp, __proj1, __proj2},
+        _ScanInputTransform{}, _WriteOp{}, oneapi::dpl::unseq_backend::__no_init_value<_Size>{},
+        /*_Inclusive=*/std::true_type{},
         /*__is_unique_pattern=*/std::false_type{}, __partition_event);
```
akukanov:
```suggestion
        /*_Inclusive=*/std::true_type{}, /*__is_unique_pattern=*/std::false_type{}, __partition_event);
```

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -171,6 +176,14 @@ constexpr int trivial_size{0};
...
 constexpr int calc_res_size(int n, int) { return n; }
 
+// If in1 range is empty, then out range is empty
+template<typename>
+int out_size_with_empty_in1(int) { return 0; };
+
+// If in2 range is empty, then out range is empty
+template<typename>
+int out_size_with_empty_in2(int) { return 0; };
+
```
akukanov: Add a comment that these functions can be specialized in the actual tests.

dmitriy-sobolev: Done

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -171,6 +176,16 @@ constexpr int trivial_size{0};
...
+template<typename>
+int out_size_with_empty_in1(int) { return 0; };
+
+// If in2 range is empty, then the out range is always empty
+// Can be specialized with an algorithm type if the behaviour is different, e.g. see set_union test.
+template<typename>
+int out_size_with_empty_in2(int) { return 0; };
+
 auto data_gen2_default = [](auto i) { return i % 5 ? i : 0;};
 auto data_gen_zero = [](auto) { return 0;};
```
danhoeflinger: It would be better to fill with something other than zero.  
The set operations are a great example of this.  Uninitialized memory is often filled with zeros.  
For cases like the set operations, we want to ensure that no data is written past the end of the expected output sequence.  We may miss issues if we fill with zeros.

We should be prepared for this possibly exposing issues with other range algorithms.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3700,35 +3724,38 @@ __pattern_set_difference(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __e
...
+        return __pattern_walk2_brick(__tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __result,
                                      __internal::__brick_copy<__parallel_tag<_IsVector>>{});
 
     // testing  whether the sequences are intersected
-    _RandomAccessIterator1 __left_bound_seq_1 = ::std::lower_bound(__first1, __last1, *__first2, __comp);
+    _RandomAccessIterator1 __left_bound_seq_1 = std::lower_bound(__first1, __last1, *__first2, __comp);
     //{1} < {2}: seq 2 is wholly greater than seq 1, so, parallel copying just first sequence
     if (__left_bound_seq_1 == __last1)
-        return __pattern_walk2_brick(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __result,
+        return __pattern_walk2_brick(__tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __result,
```
danhoeflinger, line 3734: I don't think its a blocker, but there are a lot of lines in this PR where the only change is `::std` -> `std`.

Thus far we have avoided these and formatting only changes.  I don't care as much about the git blame, but it does make review more time consuming.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -643,6 +643,522 @@ __pattern_merge_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&&
...
+    if (__last2 - __first2 == 1)
+        return !std::invoke(__comp, std::invoke(__proj1, *__first1), std::invoke(__proj2, *__first2)) &&
+               !std::invoke(__comp, std::invoke(__proj2, *__first2), std::invoke(__proj1, *__first1));
+
+    return !__internal::__parallel_or(
+        __tag, std::forward<_ExecutionPolicy>(__exec), __first2, __last2,
+        [__first1, __last1, __first2, __last2, __comp, __proj1, __proj2](_RandomAccessIterator2 __i,
+                                                                         _RandomAccessIterator2 __j) {
+            assert(__j > __i);
+            //assert(__j - __i > 1);
```
danhoeflinger: should this line be removed?

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -643,6 +643,522 @@ __pattern_merge_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&&
...
+                __j += oneapi::dpl::__internal::__pstl_upper_bound(__j, _DifferenceType2{0}, __last2 - __j,
+                                                                   std::invoke(__proj2, *__j), __comp, __proj2);
+
+            //2. testing is __a subsequence of the second range included into the first range
+            auto __b = __first1 +
+                       oneapi::dpl::__internal::__pstl_lower_bound(__first1, _DifferenceType1{0}, __last1 - __first1,
+                                                                   std::invoke(__proj2, *__i), __comp, __proj1);
+
+            //assert(!__comp(*(__last1 - 1), *__b));
+            //assert(!__comp(*(__j - 1), *__i));
```
danhoeflinger: should these be removed?

include/oneapi/dpl/pstl/algorithm_impl.h
akukanov: I think it makes sense to clean up a whole function from extra `::` rather than a single line. While I agree it makes reviews more complicated, not sure if reverting these changes selectively makes any sense at this point.

danhoeflinger: That's fair.  I don't mean for someone to go manually doing this.  If there were a commit which applied all of these we could revert, I have a script which selectively applies this change only for the PR's diff I could share.
I'm also OK with just leaving it as is and dealing with it.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -952,6 +952,232 @@ __pattern_merge_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exe
...
+{
+    const auto __first1 = std::ranges::begin(__r1);
+    const auto __first2 = std::ranges::begin(__r2);
+    const auto __result = std::ranges::begin(__out_r);
+
+    const auto __n1 = std::ranges::size(__r1);
+    const auto __n2 = std::ranges::size(__r2);
+
+    // intersection is empty
+    if (__n1 == 0 && __n2 == 0)
```
danhoeflinger:
```suggestion
    if (__n1 == 0 || __n2 == 0)
```

If either list is empty, then the intersection is empty.  This is the cause of the assertion in the per commit CI.

SergeyKopienko: I think it was type some kind of type error. Applied, thanks @danhoeflinger

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -987,54 +987,57 @@ struct __scan
...
+        const _SizeB __b_end = __nb;
 
         // testing __comp(*__first2, *__first1) or __comp(*(__last1 - 1), *(__last2 - 1))
-        if ((__idx == 0 && __comp(__b[__b_beg + 0], __a[__a_beg + 0])) ||
-            (__idx == __nb - 1 && __comp(__a[__a_end - 1], __b[__b_end - 1])))
-            return true; //__a doesn't include __b
+        if ((__idx == 0 && std::invoke(__comp, std::invoke(__projB, __rngB[__b_beg + 0]),
+                                       std::invoke(__projA, __rngA[__a_beg + 0]))) ||
+            (__idx == __nb - 1 &&
+             std::invoke(__comp, std::invoke(__projA, __rngA[__a_end - 1]), std::invoke(__projB, __rngB[__b_end - 1]))))
```
akukanov: Formatting suggestion:
```suggestion
        if ((__idx == 0 && std::invoke(__comp, std::invoke(__projB, __rngB[__b_beg + 0]),
                                               std::invoke(__projA, __rngA[__a_beg + 0]))) ||
            (__idx == __nb - 1 && std::invoke(__comp, std::invoke(__projA, __rngA[__a_end - 1]),
                                                      std::invoke(__projB, __rngB[__b_end - 1]))))
```

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
```diff
@@ -627,17 +627,17 @@ struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__par_backen
...
 {
 };
 
-template <typename _SetOpCount, typename _TempData, typename _Compare>
+template <typename _SetOpCount, typename _TempData, typename _Compare, typename _Proj1, typename _Proj2>
 struct sycl::is_device_copyable<_ONEDPL_SPECIALIZE_FOR(
-    oneapi::dpl::__par_backend_hetero::__gen_set_op_from_known_balanced_path, _SetOpCount, _TempData, _Compare)>
-    : oneapi::dpl::__internal::__are_all_device_copyable<_SetOpCount, _Compare>
+    oneapi::dpl::__par_backend_hetero::__gen_set_op_from_known_balanced_path, _SetOpCount, _TempData, _Compare, _Proj1,
+    _Proj2)> : oneapi::dpl::__internal::__are_all_device_copyable<_SetOpCount, _Compare, _Proj1, _Proj2>
```
akukanov: Formatting suggestion:
```c++
template <typename _SetOpCount, typename _TempData, typename _Compare, typename _Proj1, typename _Proj2>
struct sycl::is_device_copyable<
    _ONEDPL_SPECIALIZE_FOR(oneapi::dpl::__par_backend_hetero::__gen_set_op_from_known_balanced_path,
                           _SetOpCount, _TempData, _Compare, _Proj1, _Proj2)>
    : oneapi::dpl::__internal::__are_all_device_copyable<_SetOpCount, _Compare, _Proj1, _Proj2>
```

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -987,54 +987,57 @@ struct __scan
...
-        const auto __count_a = __internal::__pstl_right_bound(__a, __res, __a_end, __val_a, __comp) -
-                               __internal::__pstl_left_bound(__a, __a_beg, __res, __val_a, __comp);
+        const auto __count_a = __internal::__pstl_right_bound(__rngA, __res, __a_end, __val_a_proj, __comp, __projA) -
+                               __internal::__pstl_left_bound(__rngA, __a_beg, __res, __val_a_proj, __comp, __projA);
 
-        const auto __count_b = __internal::__pstl_right_bound(__b, _Size2(__idx_b), __b_end, __val_b, __comp) -
+        const auto __count_b = __internal::__pstl_right_bound(__rngB, __idx_b, __b_end, __val_b_proj, __comp, __projB) -
                                __idx_b + __idx_b -
-                               __internal::__pstl_left_bound(__b, __b_beg, _Size2(__idx_b), __val_b, __comp);
+                               __internal::__pstl_left_bound(__rngB, __b_beg, __idx_b, __val_b_proj, __comp, __projB);
```
akukanov: Looks like the line in the middle is redundant - it subtracts (see minus at the end of the first line) and then adds the same value of `__idx_b`.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -374,7 +374,7 @@ struct __gen_unique_mask
 
 // A mask generator for set operations (difference or intersection) to determine if an element from Set A should be
 // written to the output sequence based on its presence in Set B and the operation type (difference or intersection).
-template <typename _SetTag, typename _Compare>
+template <typename _SetTag, typename _Compare, typename _Proj1, typename _Proj2>
 struct __gen_set_mask
```
akukanov, line 378: It looks like this class default-constructs projections. Is that a right thing to do? I do not see in the standard that projections are always default-constructible.

SergeyKopienko: This class instantiated only in one place:
```C++
using _GenMaskReduce = oneapi::dpl::__par_backend_hetero::__gen_set_mask<_SetTag, _Compare, _Proj1, _Proj2>;
```
```C++
_GenMaskReduce{__comp, __proj1, __proj2}
```
so we assign to their fields the projections which we have

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -673,7 +683,7 @@ struct __get_bounds_simple
 // Reduce then scan building block for set balanced path which is used in the reduction kernel to calculate the
 // balanced path intersection, store it to temporary data with "star" status, then count the number of elements to write
 // to the output for the reduction operation.
-template <typename _SetOpCount, typename _BoundsProvider, typename _Compare>
+template <typename _SetOpCount, typename _BoundsProvider, typename _Compare, typename _Proj1, typename _Proj2>
 struct __gen_set_balanced_path
```
akukanov, line 687: Same for this class: projections are default-constructed.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
SergeyKopienko: I save this line only to avoid extra changes.
But agreed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -841,19 +851,21 @@ struct __gen_set_balanced_path
...
+    _Proj1 __proj1;
+    _Proj2 __proj2;
 };
 
 // Reduce then scan building block for set balanced path which is used in the scan kernel to decode the stored balanced
 // path intersection, perform the serial set operation for the diagonal, counting the number of elements and writing
 // the output to temporary data in registers to be ready for the scan and write operations to follow.
-template <typename _SetOpCount, typename _TempData, typename _Compare>
+template <typename _SetOpCount, typename _TempData, typename _Compare, typename _Proj1, typename _Proj2>
 struct __gen_set_op_from_known_balanced_path
```
akukanov, line 869: And here as well.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/sycl_traits.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
SergeyKopienko: Fixed

test/parallel_api/ranges/std_ranges_test.h
SergeyKopienko: @dmitriy-sobolev please take a look

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: At @akukanov request, we changed two things in the functions where any corrections were made:
1) `typedef` -> `using`
2) `::std` -> `std`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
SergeyKopienko: We had the variant when we wrote
`, typename _Proj1 = oneapi::dpl::identity, typename _Proj2 = oneapi::dpl::identity`
but this allows us to forgot real used projections so the code was error-prone.

According to the approach suggested by @danhoeflinger in offline discussion I removed these defaults.

SergeyKopienko: Re-phrases my answer, in this PR we applying the same approach for projections like for comparators has been applied before:
1) specify required projection type in instantiation as template argument;
2) create projection instance and passing it into aggregate initialization of structure.

test/parallel_api/ranges/std_ranges_test.h
dmitriy-sobolev: #2446 changes the value to fill unprocessed data. 

@SergeyKopienko, your comment about the projections belongs to another discussion, I suppose.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
SergeyKopienko: So looking to the code we have only 4 structures with these template parameters:
```C++
template <typename _SetTag, typename _Compare, typename _Proj1, typename _Proj2>
struct __gen_set_mask
{
};

template <typename _SetOpCount, typename _BoundsProvider, typename _Compare, typename _Proj1, typename _Proj2>
struct __gen_set_balanced_path
{
};

template <typename _SetOpCount, typename _TempData, typename _Compare, typename _Proj1, typename _Proj2>
struct __gen_set_op_from_known_balanced_path
{
};

template <typename _F, typename _Proj1, typename _Proj2>
struct __binary_op
{
};

template <typename _SizeA, typename _SizeB, typename _Compare, typename _ProjA, typename _ProjB>
struct __brick_includes
{
};

template <typename _SetTag, typename _SizeA, typename _SizeB, typename _Compare, typename _ProjA, typename _ProjB>
class __brick_set_op
{
};
```

And I can confirm by searching the code that in all cases the correct projection types are passed to the template parameters of these structures and the correct projection instances are passed through aggregate initialization of the structure instances.

test/parallel_api/ranges/std_ranges_test.h
dmitriy-sobolev: I've merged #2446 locally into this PR and verified that the set tests pass. `set_intersection` failed since it also fails in the current state of this branch.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: Currently, I believe in all cases we explicitly provide all member fields when constructing these structs (at least for reduce_then_scan helper structs).  However, knowing this requires looking at all instantiations of the structs.  

It is probably preferable to create explicit constructors to remove any doubt within the struct itself about default constructible requirements.  Relying on implicit construction of the struct by specifying the member fields leaves it up to the caller, and if not all member fields are provided then they are attempted to be default constructed.  

This is something we could improve upon throughout the repository.  We rely too much on this style of struct construction in my opinion (I myself have written them this way).

test/parallel_api/ranges/std_ranges_test.h
dmitriy-sobolev: Upd. set_intersection passes now as well.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -643,6 +643,519 @@ __pattern_merge_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&&
...
+    static_assert(__is_serial_tag_v<_Tag> || __is_parallel_forward_tag_v<_Tag>);
+
+    return __brick_includes(std::forward<_R1>(__r1), std::forward<_R2>(__r2), __comp, __proj1, __proj2,
+                            typename _Tag::__is_vector{});
+}
+
+template <class _IsVector, typename _ExecutionPolicy, typename _R1, typename _R2, typename _Comp, typename _Proj1,
+          typename _Proj2>
+bool
+__pattern_includes(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&& __r2, _Comp __comp,
```
danhoeflinger, line 682: It really seems like this implementation should be shared between here and `algorithm_impl.h` as a `__parallel_includes` function.  

The only difference I see is the projections, where `std::identity` could be used to support the iterator version.

SergeyKopienko: Unfortunately we have more differences than that:
- `std::lower_bound` <-> `oneapi::dpl::__internal::__pstl_lower_bound`
- `std::upper_bound` <-> `oneapi::dpl::__internal::__pstl_upper_bound`
- `std::includes` <-> `std::ranges::includes`

So I think it makes sense to remove this duplication but it may be done later due to a potential large number of changes. 

There used to be a similar problem with sort serial algorithm used in the corresponding patterns. It was solved in a follow-up PR.

(commented by @SergeyKopienko and @dmitriy-sobolev)

danhoeflinger: Ok, it at least deserves an issue to be addressed later.   
There are also differences in obtaining the sizes, etc.  I meant semantically it is the same, it is not identical but its unnecessarily redundant code.

SergeyKopienko: Issue created: https://github.com/uxlfoundation/oneDPL/issues/2447

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -643,6 +643,519 @@ __pattern_merge_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&&
...
+    // testing  whether the sequences are intersected
+    auto __left_bound_seq_2 =
+        __first2 + oneapi::dpl::__internal::__pstl_lower_bound(__first2, _DifferenceType2{0}, __last2 - __first2,
+                                                               std::invoke(__proj1, *__first1), __comp, __proj2);
+    //{2} < {1}: seq 1 is wholly greater than seq 2, so, the intersection is empty
+    if (__left_bound_seq_2 == __last2)
+        return __pattern_set_intersection_return_t<_R1, _R2, _OutRange>{__last1, __last2, __result};
+
+    const auto __m1 = __last1 - __left_bound_seq_1 + __n2;
+    if (__m1 > oneapi::dpl::__internal::__set_algo_cut_off)
```
danhoeflinger, line 921: Im not understanding why we have 2 different options for `__m1/2` here.  
should it just be `__m = __last1 - __left_bound_seq_1 + __last2 - __left_bound_seq_2`?  then we run intersection from the left bound to the last of each set?

Can any element between `__first1/2` and `__left_bound_seq_1/2` impact the intersection?

danhoeflinger: I'm seeing that this is also the logic in the non-ranges impl.  I think this further emphasizes the point that non-trivial implementations should not be replicated and we should do whatever we can to unify redundant code.  

Assuming I'm not missing something and this needs a fix, we should also fix the iterator version. It would be better to only need to fix it in one location.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -643,6 +643,519 @@ __pattern_merge_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&&
...
+        return __pattern_set_difference_return_t<_R1, _OutRange>{__last1, __out_last};
+    }
+
+    if (__n1 + __n2 > oneapi::dpl::__internal::__set_algo_cut_off)
+    {
+        auto __out_last = oneapi::dpl::__internal::__parallel_set_union_op(
+            __tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2, __last2, __result,
+            [](_RandomAccessIterator1 __first1, _RandomAccessIterator1 __last1, _RandomAccessIterator2 __first2,
+               _RandomAccessIterator2 __last2, _Tp* __result, _Comp __comp, _Proj1 __proj1, _Proj2 __proj2) {
+                return oneapi::dpl::__utils::__set_difference_construct(
```
danhoeflinger: Similarly, could we replace `__first2` here with `__left_bound_seq_2`?  If elements prior to the left bound don't intersect with sequence 1, then they shouldn't affect the difference, right?

SergeyKopienko: I think both implementations
```C++
template <class _IsVector, class _ExecutionPolicy, class _RandomAccessIterator1, class _RandomAccessIterator2,
          class _RandomAccessIterator3, class _Compare>
_RandomAccessIterator3
__pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __exec, _RandomAccessIterator1 __first1,
                           _RandomAccessIterator1 __last1, _RandomAccessIterator2 __first2,
                           _RandomAccessIterator2 __last2, _RandomAccessIterator3 __result, _Compare __comp)
```
and
```C++
template <class _IsVector, typename _ExecutionPolicy, typename _R1, typename _R2, typename _OutRange, typename _Comp,
          typename _Proj1, typename _Proj2>
__pattern_set_intersection_return_t<_R1, _R2, _OutRange>
__pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&& __r2,
                           _OutRange&& __out_r, _Comp __comp, _Proj1 __proj1, _Proj2 __proj2)
```
looks very similar.
So all these points are not the errors, but some places for the future optimization.
Correct?

danhoeflinger: You are correct in that they are not correctness bugs but rather inefficiencies.

However, I'm just trying to review the code as if it is new code (even if it is basically a translation of the iterator pattern), so I'm pointing out places where it could be better.  I don't see either comment as a blocker really.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -952,6 +952,229 @@ __pattern_merge_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exe
...
+struct __set_symmetric_difference_copy_case_1;
+
+template <typename Name>
+struct __set_symmetric_difference_copy_case_2;
+
+template <typename Name>
+struct __set_symmetric_difference_phase_1;
+
+template <typename Name>
+struct __set_symmetric_difference_phase_2;
```
danhoeflinger:
```suggestion
```

These aren't used.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -952,6 +952,229 @@ __pattern_merge_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exe
...
+
+    const std::size_t __result_size = __par_backend_hetero::__parallel_set_op<unseq_backend::_UnionTag>(
+        _BackendTag{}, unseq_backend::_UnionTag{}, std::forward<_ExecutionPolicy>(__exec), std::forward<_R1>(__r1),
+        std::forward<_R2>(__r2), std::forward<_OutRange>(__out_r), __comp, __proj1, __proj2);
+
+    return {__first1 + __n1, __first2 + __n2, __result + __result_size};
+}
+
+template <typename Name>
+struct __set_intersection_scan_then_propagate;
```
danhoeflinger:
```suggestion
```
This isn't used.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -952,6 +952,229 @@ __pattern_merge_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exe
...
+
+//Dummy names to avoid kernel problems
+template <typename Name>
+struct __set_union_copy_case_1;
+
+template <typename Name>
+struct __set_union_copy_case_2;
+
+template <typename Name>
+struct __set_union_scan_then_propagate;
```
danhoeflinger:
```suggestion
```
unused

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -952,6 +952,229 @@ __pattern_merge_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exe
...
+
+    return {__first1 + __n1, __first2 + __n2, __result + __result_size};
+}
+
+//Dummy names to avoid kernel problems
+template <typename Name>
+struct __set_difference_copy_case_1;
+
+template <typename Name>
+struct __set_difference_scan_then_propagate;
```
danhoeflinger:
```suggestion
```

unused

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
```diff
@@ -258,7 +259,7 @@ struct __parallel_merge_submitter<_OutSizeLimit, _IdType, __internal::__optional
 
                 [[maybe_unused]] const std::pair __ends =
                     __serial_merge(__rng1, __rng2, __rng3, __start.first, __start.second, __i_elem, __n_merge, __n1,
-                                   __n2, __comp, __n, __proj1, __proj2);
+                                   __n2, __comp, __proj1, __proj2);
```
danhoeflinger: Should this have `__n` at the end of the call?
We've moved the param to the end, but it defaults to 0 if we don't provide it.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_merge.h
SergeyKopienko: Good catch. Fixed.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
SergeyKopienko: The issue created: https://github.com/uxlfoundation/oneDPL/issues/2448

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
SergeyKopienko: Issue created: https://github.com/uxlfoundation/oneDPL/issues/2449

test/parallel_api/ranges/std_ranges_set_symmetric_difference.pass.cpp
```diff
@@ -0,0 +1,153 @@
...
+{
+    int a;
+    operator int() const { return a; }
+};
+
+struct B
+{
+    int b;
+    operator int() const { return b; }
+};
```
danhoeflinger: can these be moved to some shared utility header?

test/parallel_api/ranges/std_ranges_set_symmetric_difference.pass.cpp
```diff
@@ -0,0 +1,153 @@
...
+    std::vector<int> out_expected = {0, 1, 2, 3, 5};
+
+    std::vector<int> out_seq(out_expected.size(), 0xCD);
+    std::vector<int> out_par(out_expected.size(), 0xCD);
+    std::vector<int> out_unseq(out_expected.size(), 0xCD);
+    std::vector<int> out_par_unseq(out_expected.size(), 0xCD);
+
+    auto proj_a = [](const A& a) { return a.a; };
+    auto proj_b = [](const B& b) { return b.b; };
+
```
danhoeflinger: Should also move these projections to a shared utility.

test/parallel_api/ranges/std_ranges_set_intersection.pass.cpp
```diff
@@ -0,0 +1,146 @@
...
+    operator int() const { return a; }
+};
+
+struct B
+{
+    int b;
+    operator int() const { return b; }
+};
+
+void test_mixed_types_host()
```
danhoeflinger, line 20: Seems like outside of the actual vector data, both mixed_types functions could be shared infrastructure for the tests as well.

Not a blocker, but it would be better more maintainable code.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -383,27 +383,28 @@ struct __gen_set_mask
     {
         // First we must extract individual sequences from zip iterator because they may not have the same length,
         // dereferencing is dangerous
-        auto __set_a = std::get<0>(__in_rng.tuple());    // first sequence
-        auto __set_b = std::get<1>(__in_rng.tuple());    // second sequence
+        auto __set_a = std::get<0>(__in_rng.tuple());    // first sequence, __proj1 applied
+        auto __set_b = std::get<1>(__in_rng.tuple());    // second sequence, __proj2 applied
```
akukanov: Projections are not yet applied, are they?
```suggestion
        auto __set_a = std::get<0>(__in_rng.tuple());    // first sequence, use with __proj1
        auto __set_b = std::get<1>(__in_rng.tuple());    // second sequence, use with __proj2
```

test/parallel_api/ranges/std_ranges_set_symmetric_difference.pass.cpp
SergeyKopienko: Fixed

test/parallel_api/ranges/std_ranges_set_symmetric_difference.pass.cpp
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/utils.h
```diff
@@ -92,7 +94,7 @@ class __not_pred
     bool
     operator()(_Args&&... __args) const
     {
-        return !_M_pred(::std::forward<_Args>(__args)...);
+        return !std::invoke(_M_pred, std::forward<_Args>(__args)...);
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/utils.h
```diff
@@ -108,7 +110,7 @@ class __reorder_pred
     bool
     operator()(_FTp&& __a, _STp&& __b) const
     {
-        return _M_pred(::std::forward<_STp>(__b), ::std::forward<_FTp>(__a));
+        return std::invoke(_M_pred, std::forward<_STp>(__b), std::forward<_FTp>(__a));
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

test/parallel_api/ranges/std_ranges_set_intersection.pass.cpp
SergeyKopienko: Issue created: https://github.com/uxlfoundation/oneDPL/issues/2453

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -643,6 +643,519 @@ __pattern_merge_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&&
...
+    return std::ranges::includes(std::forward<_R1>(__r1), std::forward<_R2>(__r2), __comp, __proj1, __proj2);
+}
+
+template <typename _Tag, typename _ExecutionPolicy, typename _R1, typename _R2, typename _Comp, typename _Proj1,
+          typename _Proj2>
+bool
+__pattern_includes(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&& __r2, _Comp __comp, _Proj1 __proj1,
+                   _Proj2 __proj2)
+{
+    static_assert(__is_serial_tag_v<_Tag> || __is_parallel_forward_tag_v<_Tag>);
```
akukanov: Is it even possible to get `__is_parallel_forward_tag_v` in the range algorithms? Unlike for iterators, there is the requirement for ranges to have random access and it is independent of the policy, I believe.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -643,6 +643,519 @@ __pattern_merge_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&&
...
+                                                               std::invoke(__proj1, *__first1), __comp, __proj2);
+    //{2} < {1}: seq 1 is wholly greater than seq 2, so, the intersection is empty
+    if (__left_bound_seq_2 == __last2)
+        return __pattern_set_intersection_return_t<_R1, _R2, _OutRange>{__last1, __last2, __result};
+
+    const auto __m1 = __last1 - __left_bound_seq_1 + __n2;
+    if (__m1 > oneapi::dpl::__internal::__set_algo_cut_off)
+    {
+        //we know proper offset due to [first1; left_bound_seq_1) < [first2; last2)
+        auto __out_last = oneapi::dpl::__internal::__parallel_set_union_op(
```
akukanov: Why `__parallel_set_union_op` is used in `__pattern_set_intersection`?

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -643,6 +643,519 @@ __pattern_merge_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&&
...
+    {
+        auto __out_last =
+            __internal::__pattern_walk2_brick(__tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1,
+                                              __result, __brick_copy<__parallel_tag<_IsVector>>{});
+        return __pattern_set_difference_return_t<_R1, _OutRange>{__last1, __out_last};
+    }
+
+    if (__n1 + __n2 > oneapi::dpl::__internal::__set_algo_cut_off)
+    {
+        auto __out_last = oneapi::dpl::__internal::__parallel_set_union_op(
```
akukanov: Why `__parallel_set_union_op` is used in `__pattern_set_difference`?

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
SergeyKopienko: Looks like an error. Will fix.

SergeyKopienko: Fixed

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
SergeyKopienko: Am I correct we should replace this to
```C++
static_assert(__is_serial_tag_v<_Tag>);
```

akukanov: It seems so to me. Or you might change the type of the first parameter to be `__serial_tag`.

SergeyKopienko: Fixed

PR #2321, opened 2025-06-23
Release notes: correct sort issue description

dmitriy-sobolev:
That issue was observed on Windows only, with icpx 2025.0 and an old GPU driver presumably built in December 2024 - a combination of tools which did not fully support Intel® Core Ultra Series architecture. 

It is not observed with icpx 2025.1 (which introduced Intel® Core Ultra Series architecture support) and 32.0.101.6647 (released together with icpx 2025.1), as well as newer versions of these tools.

Let's remove the copy of the note for oneDPL 2022.9, where it is irrelevant, and update it for 2022.8. We can also remove the notes entirely, but I suggest providing an update to the original one to avoid confusion about sudden disappearance of the issue.

### DIFF
--- a/documentation/release_notes.rst
+++ b/documentation/release_notes.rst
@@ -44,7 +44,9 @@ Existing Issues
 See oneDPL Guide for other `restrictions and known limitations`_.
 
 - Incorrect results may be observed when calling ``sort`` with a device policy on Intel® Arc™ graphics 140V with data
-  sizes of 4-8 million elements.
+  sizes of 4-8 million elements on Windows.
+  That issue is resolved in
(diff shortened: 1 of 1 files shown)


documentation/release_notes.rst
```diff
@@ -99,7 +97,9 @@ Known Issues and Limitations
 New in This Release
 ^^^^^^^^^^^^^^^^^^^
 - Incorrect results may be observed when calling ``sort`` with a device policy on Intel® Arc™ graphics 140V with data
-  sizes of 4-8 million elements.
+  sizes of 4-8 million elements on Windows.
+  That issue is resolved in
```
mmichel11:
```suggestion
  This issue is resolved in
```

mmichel11 (approved):
LGTM. I agree it makes the most sense to leave the already existing note and to provide the minimum compiler + driver version we know sort will function correctly on.

PR #2326, opened 2025-06-25
Add checks with a single projection

dmitriy-sobolev:
Some tests of algorithms with two optional projections were missing checks when only one projection is passed.
It should be merged after #2316.


dmitriy-sobolev:
Closing as not relevant (these checks were added in #2316

PR #2328, opened 2025-06-26
CI: Update oneAPI to 2025.2 on Windows

dmitriy-sobolev:
Additionally use windows-latest (Windows Server 2022) instead of windows-2019, which served as a temporary workaround.


SergeyKopienko (approved):
LGTM

PR #2334, opened 2025-07-08
Additional comments for `noop` implementation in tests

SergeyKopienko:
This PR explains in comments why we should implement our own `noop` operator in `inpud_data_sweep*` tests and can't use `oneapi::dpl::identity` instead.

### DIFF
--- a/test/parallel_api/iterator/input_data_sweep.h
+++ b/test/parallel_api/iterator/input_data_sweep.h
@@ -45,6 +45,7 @@ struct get_expected_op
 };
 
 // This noop function is used to create a transform iterator that does not change the value
+//
 // Attention: we can not use oneapi::dpl::identity here because it's return universal reference
 // inside:
 //    template <typename _T>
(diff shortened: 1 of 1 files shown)


test/parallel_api/iterator/input_data_sweep.h
```diff
@@ -44,6 +44,27 @@ struct get_expected_op
...
+//    reference
+//    operator*() const _ONEDPL_CPP20_REQUIRES(std::indirectly_readable<_Iter>)
+//    {
+//        return __my_unary_func_(*__my_it_);
+//    }
+// where reference is defined as:
+//    typedef decltype(__my_unary_func_(::std::declval<typename ::std::iterator_traits<_Iter>::reference>())) reference;
+// because in combination with oneapi::dpl::identity it will return a dangling reference.
+//
+// So we use a lambda that returns the value by copy instead.
```
danhoeflinger: A key piece of this returning a dangling reference is a source iterator which returns a non-reference type when dereferenced (i.e. counting_iterator or zip_iterator). For a iterator returning a reference type, there is no issue.

SergeyKopienko: Fixed. Correct?

danhoeflinger (commented):
I personally think it is overkill to quote code snippets here, and makes it harder to understand.  What about something like this:

```
// Attention:
// We cannot use oneapi::dpl::identity here because it returns a forwarded universal reference.
// Such functors cannot be used within transform_iterator in combination with a source iterator
// which returns a non-reference type when dereferenced (i.e. counting_iterator or zip_iterator). 
// This combination returns a dangling reference and results in undefined behavior.  
// Instead, we use a functor which copies the returned value.
```

PR #2335, opened 2025-07-08
Add documentation for transform_iterator pitfall

danhoeflinger:
Add some documentation to `transform_iterator` to warn of a potential pitfall with a dangling reference when using functors similar to `oneapi::dpl::identity` in combination with source iterators returning a prvalue.

### DIFF
--- a/documentation/library_guide/parallel_api/iterators.rst
+++ b/documentation/library_guide/parallel_api/iterators.rst
@@ -84,17 +84,17 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
   .. note::
      When using ``transform_iterator`` with base iterators that return prvalues (temporary objects) upon dereferencing
      (such as ``counting_iterator`` or ``zip_iterator``), care must be taken with the transform functor. Functors that
-     return references to their input arguments (like ``oneapi::dpl::identity``) will create dangling references when
-     applied to prvalues, resulting in undefined behavior. Instead, in situations like these, use functors that return
-     values by copy:
+     return references they accepted as input reference arguments (like ``oneapi::dpl::identity`` or ``std::identity``)
+     will create dangling references when applied to prvalues, resulting in undefined behavior. Instead, in these
+     situations, use functors that return values by copy:
 
      .. code:: cpp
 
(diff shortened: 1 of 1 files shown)


documentation/library_guide/parallel_api/iterators.rst
```diff
@@ -81,6 +81,21 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
   the reference type of the base iterator as argument. The functor's call operator should not have any side effects and
   should not modify the state of the functor object.
 
+  .. note::
+     When using ``transform_iterator`` with base iterators that return prvalues (temporary objects) upon dereferencing
+     (such as ``counting_iterator`` or ``zip_iterator``), care must be taken with the transform functor. Functors that
+     return references to their input arguments (like ``oneapi::dpl::identity``) will create dangling references when
+     applied to prvalues, resulting in undefined behavior. Instead, in situations like these, use functors that return
+     values by copy:
```
SergeyKopienko:
```suggestion
  .. note::
     When using ``transform_iterator`` with base iterators that return prvalues (temporary objects) upon dereferencing
     (such as ``counting_iterator`` or ``zip_iterator``), care must be taken with the transform functor. Functors that
     return references to their input arguments (like ``oneapi::dpl::identity`` or ``std::identity`` and etc.) will create dangling references when
     applied to prvalues, resulting in undefined behavior. Instead, in situations like these, use functors that return
     values by copy:
```

documentation/library_guide/parallel_api/iterators.rst
```diff
@@ -81,6 +81,21 @@ header.  All iterators are implemented in the ``oneapi::dpl`` namespace.
...
+
+     .. code:: cpp
+
+       //  DANGEROUS: identity returns reference to prvalue (dangling reference)
+       auto bad_transform = dpl::make_transform_iterator(counting_iter, dpl::identity{});
```
SergeyKopienko: May be make sense to describe how we create `counting_iter` in this example

dcbenito (approved):
LGTM.

PR #2338, opened 2025-07-10
Rename the name of execution policy parameter in tests

SergeyKopienko:
This PR renames the execution policy parameter from `policy` to `exec` across test utilities and test cases for consistency.

- Renamed function parameters and forwarded variables from `policy` to `exec` in test helper headers.
- Updated all test implementations to use `exec` when invoking queue-based operations and wrappers.
- Ensured stable and unstable sort helpers use the new parameter name.

## Justification
Previously in tests we had `Policy&& policy` and `Policy&& exec` together:
 - `Policy&& exec` - 373 entries in tests;
 - `Policy&& policy` -  14 (16) entries in tests.

### DIFF
--- a/test/general/test_policies.pass.cpp
+++ b/test/general/test_policies.pass.cpp
@@ -29,9 +29,9 @@ class Kernel;
 #if TEST_DPCPP_BACKEND_PRESENT
 
 template<typename Policy>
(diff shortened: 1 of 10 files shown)


dmitriy-sobolev (approved):
LGTM

PR #2340, opened 2025-07-11
Remove `try` / `catch` block from asynch.pass.cpp

SergeyKopienko:


dmitriy-sobolev (approved):
LGTM.

Addition to the description (after clarifying it with Sergey):
 - try-catch block is not needed here; it should better be added to main to simplify exception handling or be removed to keep the state before #2337.
 - `return EXIT_FAILURE` should be removed if the function returns `void`.

PR #2341, opened 2025-07-11
Fix use after after forward in `__pattern_shift_left` in parallel host backend

mmichel11:
We forward a policy calling parallel-for in an iterative loop which results in undefined behavior.


danhoeflinger (approved):
LGTM, good find.

PR #2344, opened 2025-07-14
Update a known issue for range-based count_if

dmitriy-sobolev:
The issue does not reproduce with other Rolling or LTS versions of the GPU driver.

This addition complements the existing known issue for the 2022.9 release.
Let's update the 2022.9 section of [the release notes](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-dpcpp-library-release-notes.html) if it's expected to be revised in the next release.
If not, it would be more appropriate to add this to the oneDPL 2022.10 section instead.

Upd. Agreed to remove the note entirely as outdated - the mentioned Rolling driver version is no longer supported.

### DIFF
--- a/documentation/release_notes.rst
+++ b/documentation/release_notes.rst
@@ -60,7 +60,7 @@ See oneDPL Guide for other `restrictions and known limitations`_.
 - With libstdc++ version 10, the compilation error *SYCL kernel cannot use exceptions* occurs
   when calling the range-based ``adjacent_find``, ``is_sorted`` or ``is_sorted_until`` algorithms with device policies.
 - The range-based ``count_if`` may produce incorrect results on Intel® Data Center GPU Max Series when the driver version
-  is "Rolling 2507.12" and newer.
+  is *Rolling 2507.12* or *Rolling 2507.17*. Other *Rolling* and *LTS* versions are unaffected.
 
 New in 2022.8.0
 ===============
--- a/documentation/release_notes.rst
+++ b/documentation/release_notes.rst
@@ -59,8 +59,6 @@ See oneDPL Guide for other `restrictions and known limitations`_.
   To avoid the issue, pass ``-fopenmp`` or ``-fopenmp-simd`` option instead.
 - With libstdc++ version 10, the compilation error *SYCL kernel cannot use exceptions* occurs
   when calling the range-based ``adjacent_find``, ``is_sorted`` or ``is_sorted_until`` algorithms with device policies.
-- The range-based ``count_if`` may produce incorrect results on Intel® Data Center GPU Max Series when the driver version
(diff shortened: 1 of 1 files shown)


dmitriy-sobolev:
@dcbenito, could you help with the following question? Is 2022.9 section on [the release notes](https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-dpcpp-library-release-notes.html) page going to be updated when 2022.10 is out?

dcbenito:
Hi, @dmitriy-sobolev I don't document Release Notes and I have been impacted by CPM. I am assuming someone on your team will update the RN as needed for 2022.10.

akukanov:
I think it is more appropriate to update the note for 2022.10. leaving it as is in early revisions.

documentation/release_notes.rst
```diff
@@ -60,7 +60,7 @@ See oneDPL Guide for other `restrictions and known limitations`_.
 - With libstdc++ version 10, the compilation error *SYCL kernel cannot use exceptions* occurs
   when calling the range-based ``adjacent_find``, ``is_sorted`` or ``is_sorted_until`` algorithms with device policies.
 - The range-based ``count_if`` may produce incorrect results on Intel® Data Center GPU Max Series when the driver version
-  is "Rolling 2507.12" and newer.
+  is *Rolling 2507.12* or *Rolling 2507.17*. Other *Rolling* and all *LTS* versions are unaffected.
```
akukanov: Both these rolling releases are not supported by now, see https://dgpu-docs.intel.com/releases/releases.html. Does it still make sense to keep the note?

dmitriy-sobolev: I think we still should provide an update because of the "Rolling [version] **and newer**" wording. We can just recommend updating to the latest (and supported) one.

akukanov: Well, removing an issue that became practically irrelevant is also an update :)

Eventually we will need to remove the note. What would be a trigger for that, general re-evaluation of known issues before the next release?

dmitriy-sobolev: Let's remove it here.

I was concerned about what a user sees in https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-dpcpp-library-release-notes.html. If the note will be updated there for 2022.9, there is nothing to do after deleting it here. If it will not, I still think that we should say something that now count_if works with the relevant drivers for 2022.10 as an update.

On the other hand, it is release-independed and will not be faced by a user who uses relevant drivers, so mentioning it again is not as important...

dmitriy-sobolev: Removed

akukanov: Thanks.
I will keep this conversation open (unresolved) but only for the visibility by other reviewers.

PR #2345, opened 2025-07-14
Remove cpp20 requirement to include <version>

danhoeflinger:
Some compilers (like gcc 10.5)  provide <version>, but do not set `__cplusplus >= 202002L`.  
By not attempting to include <version>, we miss out on these macros, and also decline to attempt to define many features like concepts, etc.  which may exist in an implementation that does not set `__cplusplus >= 202002L`.  

Specifically, this PR fixes a divergence from TBB prior to this change where they are using concepts (checking the version macros explicitly, with no dependence on `__cplusplus`, and therefore requiring `std::input_iterator` concept for inputs to parallel for.  

We have a workaround for `zip_forward_iterator` to help it satisfy that concept which does not get applied because we try to limit the scope of that workaround to only where it is required.
  
https://github.com/uxlfoundation/oneDPL/blob/ef406d231043c7b1df49ebf781ee5cde3624138e/include/oneapi/dpl/pstl/tuple_impl.h#L829-L840

This divergence from TBB for gcc 10 causes a build errors for the tbb backend because we need the workaround and do not apply it.

### DIFF
--- a/include/oneapi/dpl/internal/version_impl.h
+++ b/include/oneapi/dpl/internal/version_impl.h
@@ -18,14 +18,11 @@
 // The oneAPI Specification version this implementation is compliant with
 #define ONEDPL_SPEC_VERSION 104
 
-#if _ONEDPL___cplusplus >= 202002L && __has_include(<version>)
+#if __has_include(<version>)
 #    include <version> // The standard C++20 header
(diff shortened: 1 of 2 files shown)


mmichel11 (approved):
LGTM

PR #2347, opened 2025-07-18
Introduce common function for `std::get<>()` call wrap

SergeyKopienko:
This PR introduces a common function `__get` as a wrapper for `std::get<>()` calls across the `oneDPL` codebase to provide a unified interface for accessing tuple elements. The implementation creates a consistent abstraction layer that can work with both standard library tuples and internal tuple implementations.

- Introduces a new header `get_impl.h` with a common `__get` wrapper function
- Replaces all direct `std::get<>()` calls with the new `__dpl_internal::__get<>()` wrapper
- Updates multiple source files across the codebase to use the new unified interface

### Implementation details:
```C++
namespace oneapi
{
namespace dpl
{
namespace __internal
{
template <std::size_t _Idx, typename _T>
constexpr auto
__get(_T&& __t) -> decltype(std::get<_Idx>(std::forward<_T>(__t)))
{
    using std::get;

    return get<_Idx>(std::forward<_T>(__t));
}

} // namespace __internal
} // namespace dpl
} // namespace oneapi

namespace __dpl_internal = oneapi::dpl::__internal;
```

Now all usages this function in the code may be found by text `__dpl_internal::__get`

### DIFF
--- a/test/general/implementation_details/get_impl.pass.cpp
+++ b/test/general/implementation_details/get_impl.pass.cpp
@@ -36,6 +36,7 @@ template <typename TData, typename T1, typename T2>
 void
 test_get_data(TData&& data, T1 val1, T2 val2)
 {
+    // Check value type and value category
     static_assert(std::is_same_v<decltype(__dpl_internal::__get<0>(std::forward<TData>(data))),
                                  decltype(std::get<0>(std::forward<TData>(data)))>);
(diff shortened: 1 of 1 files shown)


mmichel11:
Could you clarify the issue this PR resolves and the advantages over directly using `std::get`?

SergeyKopienko:
> Could you clarify the issue this PR resolves and the advantages over directly using `std::get`?

The main idea was to have some internal function which can be called simple without writing `using std::get` before.

PR #2349, opened 2025-07-22
Fix performance issue of `parallel_find_or` - 3-Kernels approach

SergeyKopienko:
This PR addresses a performance issue in the `parallel_find_or` implementation by modernizing the synchronization approach and removing manual buffer management. The changes replace the previous synchronization mechanism that relied on SYCL buffer destruction with an explicit event-based waiting pattern.

- Replaces manual SYCL buffer management with `__result_and_scratch_storage` utility class
- Introduces explicit initialization of the result value from the first work-item instead of relying on host initialization
- Changes from implicit synchronization (buffer destruction) to explicit event-based synchronization

Alternative variant implemented in [https://github.com/uxlfoundation/oneDPL/pull/2349 based on 2-kernels approach.](https://github.com/uxlfoundation/oneDPL/pull/2358).

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1450,7 +1450,7 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
             __cgh.parallel_for<KernelName...>(
                 sycl::nd_range</*dim=*/1>(sycl::range</*dim=*/1>(__wgroup_size), sycl::range</*dim=*/1>(__wgroup_size)),
                 [=](sycl::nd_item</*dim=*/1> __item) {
-                    auto __local_idx = __item.get_local_id(0);
+                    const std::size_t __local_idx = __item.get_local_id(0);
 
                     // 1. Set initial value to local found state
                     __FoundStateType __found_local = __init_value;


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1492,62 +1492,73 @@ struct __parallel_find_or_impl_multiple_wgs<__or_tag_check, __internal::__option
...
-                    });
-            });
-            //The end of the scope  -  a point of synchronization (on temporary sycl buffer destruction)
-        }
+            __cgh.parallel_for<KernelName...>(
+                sycl::nd_range</*dim=*/1>(sycl::range</*dim=*/1>(__n_groups * __wgroup_size),
+                                          sycl::range</*dim=*/1>(__wgroup_size)),
+                [=](sycl::nd_item</*dim=*/1> __item_id) {
+                    // Setup initial value into result from the first work-item
+                    const std::size_t __global_id = __item.get_global_id(0);
```
mmichel11:
```suggestion
                    const std::size_t __global_id = __item_id.get_global_id(0);
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1480,74 +1483,92 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
...
+        // Initialization of the result storage
+        auto __event_init = __q.submit([&](sycl::handler& __cgh) {
+            auto __res_acc =
+                __result_storage.template __get_result_acc<sycl::access_mode::write>(__cgh, __dpl_sycl::__no_init{});
 
-                            // Update global (for all groups) atomic state with the found index
-                            _BrickTag::__save_state_to_atomic(__found, __found_local);
-                        }
-                    });
+            __cgh.single_task<KernelNameInit...>([__res_acc, __init_value]() {
```
mmichel11: If __res_ptr is allocated with USM host memory, then we can directly write to it from the host and not incur the overhead of kernel launch / sync on the device which will improve performance for small problem sizes.

From the SYCL spec,
| USM allocation type | Description                                                    |
|---------------------|----------------------------------------------------------------|
| host                | Allocations in host memory that are accessible by a device      |
| device              | Allocations in device memory that are not accessible by the host|
| shared              | Allocations in shared memory that are accessible by both host and device |

We will likely need a specialization for the path where the result is not USM host memory, however.

SergeyKopienko:
```C++
        // Allocate memory for result in USM device memory, if it's available:
        // - we can't use USM host or USM shared memory for atomic operations through __dpl_sycl::__atomic_ref.
        using __result_and_scratch_storage_t = __result_and_scratch_storage_impl<_AtomicType, 1, sycl::usm::alloc::device>;
        __result_and_scratch_storage_t __result_storage{__q, 0};
```

SergeyKopienko: So for works with `__dpl_sycl::__atomic_ref` we may use USM-device memory or `sycl::buffer` only.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1480,74 +1483,92 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
...
+                                            __internal::__optional_kernel_name<KernelName...>>
 {
     template <typename _BrickTag, typename _AtomicType, typename _Predicate, typename... _Ranges>
     _AtomicType
     operator()(sycl::queue& __q, _BrickTag __brick_tag, const std::size_t __rng_n, const std::size_t __n_groups,
                const std::size_t __wgroup_size, const _AtomicType __init_value, _Predicate __pred, _Ranges&&... __rngs)
     {
-        auto __result = __init_value;
+        // Allocate memory for result in USM device memory, if it's available:
+        // - we can't use USM host or USM shared memory for atomic operations through __dpl_sycl::__atomic_ref.
```
SergeyKopienko: https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2024-0/atomic-operations.html
https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2025-2/atomic-operations.html

### Atomic Operations on USM Data
On discrete GPU, Atomic operations on host allocated USM (sycl::malloc_host) are not supported.
Concurrent accesses from host and device to shared USM location (sycl::malloc_shared) are not supported.
We recommend using device allocated USM (sycl::malloc_device) memory for atomics and device algorithms with atomic operations.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Please take a look to https://github.com/uxlfoundation/oneDPL/pull/2349/files#r2225663647 for details.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1480,74 +1486,107 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
...
+                        // Update global (for all groups) atomic state with the found index
+                        _BrickTag::__save_state_to_atomic(__found, __found_local);
+                    }
+                });
+        });
+
+        // Copy data back from scratch part to result part
+        auto __writeback_event = __q.submit([&](sycl::handler& __cgh) {
+            auto __scratch_acc = __result_storage.template __get_scratch_acc<sycl::access_mode::read>(__cgh);
+            auto __res_acc = __result_storage.template __get_result_acc<sycl::access_mode::write>(__cgh);
```
mmichel11:
```suggestion
            auto __res_acc = __result_storage.template __get_result_acc<sycl::access_mode::write>(__cgh, __dpl_sycl::__no_init{});
```

danhoeflinger:
I think using result and scratch storage is likely the wrong choice here.  If only buffers and no USM memory is supported, then we are better off with the original implementation, right?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1480,74 +1486,107 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
...
-                        // Set local found state value value to global atomic
-                        if (__local_idx == 0 && __found_local != __init_value)
-                        {
-                            __dpl_sycl::__atomic_ref<_AtomicType, sycl::access::address_space::global_space> __found(
-                                *__dpl_sycl::__get_accessor_ptr(__result_sycl_buf_acc));
+            __cgh.single_task<KernelNameInit...>([__scratch_acc, __init_value]() {
+                auto __scratch_ptr = __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__scratch_acc);
+                *__scratch_ptr = __init_value;
+            });
+        });
```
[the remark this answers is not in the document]
danhoeflinger: I haven't yet figured it out, but I think that we may be able to reduce this 3 kernel approach down to 2.  
It may require a workgroup-local reduction in the writeback kernel and a bit more temporary memory.  Id guess that on average it will be faster than 3 kernels but I may be mistaken.

I'll think about it.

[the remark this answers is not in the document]
akukanov: Eliminating the writeback requires one extra atomic counter, for counting the groups that completed their work. Once the counter reaches the total number of groups, the last one that incremented it can safely write the result.

SergeyKopienko:
> I think using result and scratch storage is likely the wrong choice here. If only buffers and no USM memory is supported, then we are better off with the original implementation, right?

In this case we will use `sycl::buffer` from result and scratch storage.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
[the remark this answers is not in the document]
SergeyKopienko: Implemented in the PR https://github.com/uxlfoundation/oneDPL/pull/2353

mmichel11:
> Yes, you are correct that `sycl::buffer` will be used. I was more worried about the overhead associated with the single_task launches to do copies vs the implicit copies from `sycl::buffer`. With this new method (if `sycl::buffer` is used) we add an extra device-to-device copy with the "writeback kernel" which is wasted, as we still need the implicit (buffer) copy back.
> 
> Similarly, if host_usm is unavailable, this writeback kernel is again wasted, as it doesn't change anything about the data type.
> 
> Its likely a pretty small issue, especially for nontrivial sizes of `n`. Also, I suppose we are less worried about the case where only sycl::buffer is available for use.

One option I floated here was to skip the third kernel if we have some detection mechanism for if SYCL USM host memory is not being used for the result storage and directly copy back the scratch data to host. It would require some updates to result and scratch storage.

The overhead of the extra single task is likely pretty small and probably only noticeable for small problem sizes, and the buffer path is more of a fallback since USM is usually supported. I think leaving it as-is is appropriate.

PR #2354, opened 2025-07-25
Remove default constructible requirement from result_and_scratch_storage

danhoeflinger:
This PR removes the default constructor requirement from the SYCL backend's `result_and_scratch_storage` utility.

To do this, we require the type to be device copyable (which already was the case), but also provides that a bitwise copy is equivalent to copy construction.

This was split off from #2327.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -667,9 +667,15 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
         }
         else if (__supports_USM_device)
         {
-            _T __tmp;
-            __q.memcpy(&__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
-            return __tmp;
+            static_assert(sycl::is_device_copyable_v<_T>,
+                          "The type _T must be device copyable to use __result_and_scratch_storage.");
+            // Avoid default constructor for _T, we know that _T is device copyable and therefore a copy construction
+            // is equivalent to a bitwise copy
+            _T* __tmp = static_cast<_T*>(malloc(sizeof(_T)));
+            __q.memcpy(__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
+            _T __return_tmp = std::move(*__tmp);
+            free(__tmp);
+            return __return_tmp;
         }
         else
         {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -672,7 +672,7 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
             // Avoid default constructor for _T, we know that _T is device copyable and therefore a copy construction
             // is equivalent to a bitwise copy.  We may treat *__tmp as if it has been constructed.
             oneapi::dpl::__internal::__lazy_ctor_storage<_T> __lazy_ctor_storage;
-            __q.memcpy(&(__lazy_ctor_storage.__v), __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
+            __q.memcpy(&__lazy_ctor_storage.__v, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
 
             // Setting up _T to be destroyed as this function exits. _T being device copyable provides that it has a
             // public non deleted destructor.
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -671,15 +671,13 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
                           "The type _T must be device copyable to use __result_and_scratch_storage.");
             // Avoid default constructor for _T, we know that _T is device copyable and therefore a copy construction
             // is equivalent to a bitwise copy.  We may treat *__tmp as if it has been constructed.
-            _T* __tmp = static_cast<_T*>(malloc(sizeof(_T)));
-            __q.memcpy(__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
-            _T __return_tmp = std::move(*__tmp);
+            oneapi::dpl::__internal::__lazy_ctor_storage<_T> __lazy_ctor_storage;
+            __q.memcpy(&(__lazy_ctor_storage.__v), __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
             
-            // Calling explicit destructor to properly clean up after move and before free.
-            // _T being device copyable provides that it has a public non deleted destructor
-            __tmp->~_T();
-            free(__tmp);
-            return __return_tmp;
+            // Setting up _T to be destroyed as this function exits. _T being device copyable provides that it has a
+            // public non deleted destructor.
+            oneapi::dpl::__internal::__scoped_destroyer<_T> __destroy_when_leaving_scope{__lazy_ctor_storage};
+            return __lazy_ctor_storage.__v;
         }
         else
         {
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -670,10 +670,14 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
             static_assert(sycl::is_device_copyable_v<_T>,
                           "The type _T must be device copyable to use __result_and_scratch_storage.");
             // Avoid default constructor for _T, we know that _T is device copyable and therefore a copy construction
-            // is equivalent to a bitwise copy
+            // is equivalent to a bitwise copy.  We may treat *__tmp as if it has been constructed.
             _T* __tmp = static_cast<_T*>(malloc(sizeof(_T)));
             __q.memcpy(__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
             _T __return_tmp = std::move(*__tmp);
+            
+            // Calling explicit destructor to properly clean up after move and before free.
+            // _T being device copyable provides that it has a public non deleted destructor
+            __tmp->~_T();
             free(__tmp);
             return __return_tmp;
         }
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -518,6 +518,9 @@ struct __result_and_scratch_storage_base
 template <typename _T, std::size_t _NResults = 1>
 struct __result_and_scratch_storage : __result_and_scratch_storage_base
 {
+    static_assert(sycl::is_device_copyable_v<_T>,
+                  "The type _T must be device copyable to use __result_and_scratch_storage.");
+
   private:
     using __sycl_buffer_t = sycl::buffer<_T, 1>;
 
@@ -667,10 +670,8 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
         }
         else if (__supports_USM_device)
         {
-            static_assert(sycl::is_device_copyable_v<_T>,
-                          "The type _T must be device copyable to use __result_and_scratch_storage.");
-            // Avoid default constructor for _T, we know that _T is device copyable and therefore a copy construction
-            // is equivalent to a bitwise copy.  We may treat *__tmp as if it has been constructed.
+            // Avoid default constructor for _T. We know that _T is device copyable and therefore a copy construction
+            // is equivalent to a bitwise copy. We may treat __lazy_ctor_storage.__v as constructed after the memcpy.
             oneapi::dpl::__internal::__lazy_ctor_storage<_T> __lazy_ctor_storage;
             __q.memcpy(&__lazy_ctor_storage.__v, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -673,7 +673,7 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
             // is equivalent to a bitwise copy.  We may treat *__tmp as if it has been constructed.
             oneapi::dpl::__internal::__lazy_ctor_storage<_T> __lazy_ctor_storage;
             __q.memcpy(&(__lazy_ctor_storage.__v), __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
-            
+
             // Setting up _T to be destroyed as this function exits. _T being device copyable provides that it has a
             // public non deleted destructor.
             oneapi::dpl::__internal::__scoped_destroyer<_T> __destroy_when_leaving_scope{__lazy_ctor_storage};
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
@@ -675,8 +675,8 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
             oneapi::dpl::__internal::__lazy_ctor_storage<_T> __lazy_ctor_storage;
             __q.memcpy(&__lazy_ctor_storage.__v, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
 
-            // Setting up _T to be destroyed as this function exits. _T being device copyable provides that it has a
-            // public non deleted destructor.
+            // Setting up _T to be destroyed as this function exits. The __scoped_destroyer calls destroy when it
+            // leaves scope. _T being device copyable provides that it has a public non deleted destructor.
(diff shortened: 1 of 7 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -667,9 +667,15 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
...
         else if (__supports_USM_device)
         {
-            _T __tmp;
-            __q.memcpy(&__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
-            return __tmp;
+            static_assert(sycl::is_device_copyable_v<_T>,
+                          "The type _T must be device copyable to use __result_and_scratch_storage.");
+            // Avoid default constructor for _T, we know that _T is device copyable and therefore a copy construction
+            // is equivalent to a bitwise copy
+            _T* __tmp = static_cast<_T*>(malloc(sizeof(_T)));
```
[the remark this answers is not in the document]
danhoeflinger: We are doing this intentionally to avoid default construction.  Device copyable provides that this is equivalent to copy construction.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -667,9 +667,15 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
...
+            __q.memcpy(__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
+            _T __return_tmp = std::move(*__tmp);
+            free(__tmp);
```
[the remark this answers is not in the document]
danhoeflinger: Fair point, but we cant change it to use default constructor, so we must explicitly call the destructor.  done.

test/support/utils.h
```diff
@@ -1243,6 +1243,47 @@ struct Pow2
     }
 };
 
+template <typename _T>
+struct NoDefaultCtorWrapper {
+    _T value;
+
+    // Default constructor
+    NoDefaultCtorWrapper() = delete;
```
SergeyKopienko, line 1251: I think as far as we have another constructors here, this default deleted constructor doesn't make sense.

danhoeflinger: I understand it is not necessary to write out, but since the point of this class is to delete the default, I think it is worthwhile to write it out explicitly for clarity

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -667,9 +667,19 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
...
         else if (__supports_USM_device)
         {
-            _T __tmp;
-            __q.memcpy(&__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
-            return __tmp;
+            static_assert(sycl::is_device_copyable_v<_T>,
+                          "The type _T must be device copyable to use __result_and_scratch_storage.");
+            // Avoid default constructor for _T, we know that _T is device copyable and therefore a copy construction
+            // is equivalent to a bitwise copy.  We may treat *__tmp as if it has been constructed.
+            _T* __tmp = static_cast<_T*>(malloc(sizeof(_T)));
```
SergeyKopienko: I think the usage of heap memory probably isn't the best solution here.
We may take a look to another possible variants.
For example, we already have `__lazy_ctor_storage` in the code:
```C++
// Storage helper since _Tp may not have a default constructor.
template <typename _Tp>
union __lazy_ctor_storage
{
    using __value_type = _Tp;
    _Tp __v;
    __lazy_ctor_storage() {}

    template <typename _U>
    void
    __setup(_U&& init)
    {
        new (&__v) _Tp(std::forward<_U>(init));
    }
    void
    __destroy()
    {
        __v.~_Tp();
    }
};
```
How about to use it here?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -667,9 +667,19 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
...
+            __q.memcpy(__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
+            _T __return_tmp = std::move(*__tmp);
+            
+            // Calling explicit destructor to properly clean up after move and before free.
+            // _T being device copyable provides that it has a public non deleted destructor
+            __tmp->~_T();
+            free(__tmp);
+            return __return_tmp;
```
SergeyKopienko:
```suggestion
            oneapi::dpl::__internal::__lazy_ctor_storage<_T> __storage;
            __q.memcpy(&__storage.__v, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();

            __call_destroy<_T> __destroyer{__storage};
            return __storage.__v;
```

SergeyKopienko: Also we should place somewhere one helper:
```C++
    template <typename _DataType>
    struct __call_destroy
    {
        oneapi::dpl::__internal::__lazy_ctor_storage<_DataType>& ___lazy_ctor_storage_ref;

        ~__call_destroy()
        {
            // Call destructor for the lazy ctor storage
            ___lazy_ctor_storage_ref.__destroy();
        }
    };
```
Probably make sense to place it near definition of `__lazy_ctor_storage` and then we may use this helper in the 
```C++
    _ValueT
    get_and_destroy(std::uint16_t __idx)
    {
        _ValueT __ele = std::move(__data[__idx].__v);
        __data[__idx].__destroy();
        return __ele;
    }
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: After an offline discussion, I've incorporated your suggestion.  It is better.  Thanks!

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: yes, with a name change, I've added this.  Thanks!

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -667,9 +667,17 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
...
         {
-            _T __tmp;
-            __q.memcpy(&__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
-            return __tmp;
+            static_assert(sycl::is_device_copyable_v<_T>,
+                          "The type _T must be device copyable to use __result_and_scratch_storage.");
+            // Avoid default constructor for _T, we know that _T is device copyable and therefore a copy construction
+            // is equivalent to a bitwise copy.  We may treat *__tmp as if it has been constructed.
+            oneapi::dpl::__internal::__lazy_ctor_storage<_T> __lazy_ctor_storage;
+            __q.memcpy(&(__lazy_ctor_storage.__v), __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
```
SergeyKopienko:
```suggestion
            __q.memcpy(&__lazy_ctor_storage.__v, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -667,9 +667,17 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
...
+
+            // Setting up _T to be destroyed as this function exits. _T being device copyable provides that it has a
+            // public non deleted destructor.
```
SergeyKopienko: Destructor may be not public. But may it be really deleted?

danhoeflinger: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec::device.copyable
"Type T has a public non-deleted destructor;"

You can delete a destructor.  I'm not sure exactly the use case you would want this functionality for, but its possible.  
It stops the object from being stack allocated or deleted, or other situations which would result in a destructor being called.  Anyway, the reason I mention it is because the sycl spec mentions it.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: I prefer the parens, but sure, we can remove them.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -900,6 +900,21 @@ union __lazy_ctor_storage
...
+struct __scoped_destroyer
+{
+    oneapi::dpl::__internal::__lazy_ctor_storage<_DataType>& ___lazy_ctor_storage_ref;
+    ~__scoped_destroyer()
+    {
+        // Explicitly call destructor of __lazy_ctor_storage
+        ___lazy_ctor_storage_ref.__destroy();
+    }
+};
+
```
akukanov: Why are you adding this thing, instead of just defining a destructor for `__lazy_ctor_storage` properly?

Update; OK, you'd need to convert it to be a struct of an anonymous union and a boolean value which would be set in `destroy()`, so that double destruction is prevented.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -667,9 +667,17 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
         }
         else if (__supports_USM_device)
         {
-            _T __tmp;
-            __q.memcpy(&__tmp, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
-            return __tmp;
+            static_assert(sycl::is_device_copyable_v<_T>,
+                          "The type _T must be device copyable to use __result_and_scratch_storage.");
```
akukanov: The assertion does not seem to be specific for `__get_value`, so I would place it somewhere at the class scope instead.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
```diff
@@ -667,9 +667,17 @@ struct __result_and_scratch_storage : __result_and_scratch_storage_base
...
+            // Avoid default constructor for _T, we know that _T is device copyable and therefore a copy construction
+            // is equivalent to a bitwise copy.  We may treat *__tmp as if it has been constructed.
```
akukanov: Where is `__tmp` defined?

include/oneapi/dpl/pstl/utils.h
danhoeflinger: The original thought here was to allow `__lazy_ctor_storage` to have a defaulted template argument which would toggle on automatic destruction based upon scope.  However, @SergeyKopienko  brought up a good point that this can lead to destruction occurring without a type being initialized first, especially if an exception occurs, causing early exit from the scope of a function. 
With this extra struct (and the comment), the intention would be to only use this `__scoped_destroyer` for `__lazy_ctor_storage` instances which have been initialized already, and don't have explicit `destroy()` calls.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: Sorry, this was a stale piece of a comment from an earlier iteration.  fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: moved to the class scope.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: To elaborate further, we'd like to avoid any runtime checks about whether to destroy, as well as any extra storage in `__lazy_ctor_storage`.  

The previous alternative to this was using heap memory for memcpy, then moving the resulting object to a local scope before destroying / freeing.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
mmichel11: It looks like `__tmp` still remains from my view. Could you check if you pushed the commit?

include/oneapi/dpl/pstl/utils.h
akukanov: I do not quite understand why one more word in the class and a true/false runtime check is a big deal; can you elaborate? Is it a structure that is created hundreds of thousands of time and/or used on the hottest path?

akukanov: And even if you do not want any extra checks, for the suggested use case you still do not need this struct. Instead, create a local variable with a move constructor from the storage value, then explicitly call `destroy`, and return the value just created.

akukanov: (Replacing my earlier comments with a more well-thought one :))

It's a weird pattern but I see how it can sometimes be useful. You want to combine on-demand initialization with automatic destruction, and doing both in a single type requires runtime information, while delegating destruction to a different type allows to have near-zero runtime cost. It's even kind of cool once you get it :)

But it's not without cost either: for these runtime improvements you pay with extra code complexity, and so with cognitive burden on readers, reviewers, and maintainers. It took me some time  to understand what the code is doing, why it is designed in this very way, and then to convince myself that it will certainly work as expected. And I am not sure the tradeoff in this case is right. Is the extra word on the thread stack for a flag really of concern for us? Is the code on a very hot path, so that shaving off a single true/false comparison seems important? Honestly, I do not think that it is, and I rather prefer handling this case within the storage class, or perhaps with std::optional if you do not want to modify that class.

We can also think if there is some way to keep the pattern but make it more readable / easier to understand when someone looks at the function.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
akukanov: Have you tested it with both trivial and non-trivial destructors?

include/oneapi/dpl/pstl/utils.h
akukanov: > We can also think if there is some way to keep the pattern but make it more readable / easier to understand when someone looks at the function.

There are two options we can consider: make the new struct nested to `__lazy_ctor_storage`, or make it internal to the function where it is needed. Since for now it is a one-time thing for a very specific case, I recommend the second variant. Below is how I think it could look like:
```c++
        {
            // Uninitialized space to copy data from the device
            oneapi::dpl::__internal::__lazy_ctor_storage<_T> __lazy_ctor_storage;
            // _T is device copyable and therefore a bitwise copy is equivalent to a copy construction
            __q.memcpy(&__lazy_ctor_storage.__v, __scratch_buf.get() + __scratch_n + _Idx, 1 * sizeof(_T)).wait();
  
            // Setting up the local copy to be destroyed as this function exits
            struct __cleaner {
                _T& __ref;
                ~cleaner(){ __ref.~_T(); }
            } __c{__lazy_ctor_storage.__v};
            return __lazy_ctor_storage.__v;
        }
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: Thanks, I had pushed to the wrong branch.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: > There are two options we can consider: make the new struct nested to `__lazy_ctor_storage`, or make it internal to the function where it is needed. Since for now it is a one-time thing for a very specific case, I recommend the second variant. Below is how I think it could look like:

The `__scoped_destroyer` may be applied at least in two places in our code.

danhoeflinger: The `__lazy_ctor_storage` is something used in "hot path" of kernels like reduce and the set algorithms.  Those are the places where I wouldn't want to add any extra storage to already full register files.  

This auto-destruction is not necessarily required in such hot paths or kernels with full register files.  We could have a different variant of the lazy constructed storage which supports cleanup or uses a std optional.  

In response to the local struct, there is one other location which we use this auto-destroy feature, but we don't have to do so.  Therefore, I don't object to the variant you propose here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_utils.h
danhoeflinger: I can add such a test.

danhoeflinger: A non-default destructor test revealed an existing bug in `__lazy_ctor_storage`.  We were using the default destructor, where we should've had a emptied out destructor (relying only on explicit destruction).  I've adjusted the test and also fixed `__lazy_ctor_storage`.

include/oneapi/dpl/pstl/utils.h
akukanov: If the type is used in kernels, I understand the hesitation to add extra overheads to it. Thanks for clarifying.

I will not insist on any specific solution then; use your judgement for what seems best. Thanks.

danhoeflinger: Looking at the code, I'm torn about switching to inline struct as you propose above, or leaving it as is with more comments / and possible naming changes.

As @SergeyKopienko states, the `__scoped_destroyer` *is* useful in 2 places, and saves us a `move` in the other location in a inner loop repeated action of a kernel.  Its unlikely this makes much of a difference for most types, but in theory it could.

SergeyKopienko (approved):
LGTM

PR #2355, opened 2025-07-25
Move only type support for [transform_]reduce for host backends

danhoeflinger:
This PR adds support within [transform_]reduce for types with only move constructor and move assignment operator (no copy assignment or construction) for host backends (tbb and omp).

It adds host policy tests for move only types within [transform_]reduce.pass.  

This is part of the resolution for #1955.

### DIFF
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -371,8 +371,8 @@ __brick_adjacent_difference(_RandomAccessIterator1 __first, _RandomAccessIterato
 
     auto __n = __last - __first;
     *__d_first = *__first;
-    return __unseq_backend::__simd_walk_n(
-        __n - 1, [&__op](_ReferenceType1 __x, _ReferenceType1 __y, _ReferenceType2 __z) { __z = __op(__x, __y); },
+    return __unseq_backend::__simd_walk_n(__n - 1, 
+        [&__op](_ReferenceType1 __x, _ReferenceType1 __y, _ReferenceType2 __z) { __z = __op(__x, __y); },
         __first + 1, __first, __d_first + 1);
 }
 
--- a/include/oneapi/dpl/pstl/numeric_impl.h
+++ b/include/oneapi/dpl/pstl/numeric_impl.h
@@ -44,7 +44,7 @@ __brick_transform_reduce(_ForwardIterator1 __first1, _ForwardIterator1 __last1, 
                          _BinaryOperation1 __binary_op1, _BinaryOperation2 __binary_op2,
                          /*is_vector=*/::std::false_type) noexcept
 {
-    return ::std::inner_product(__first1, __last1, __first2, __init, __binary_op1, __binary_op2);
+    return ::std::inner_product(__first1, __last1, __first2, std::move(__init), __binary_op1, __binary_op2);
 }
 
 template <class _RandomAccessIterator1, class _RandomAccessIterator2, class _Tp, class _BinaryOperation1,
@@ -57,7 +57,7 @@ __brick_transform_reduce(_RandomAccessIterator1 __first1, _RandomAccessIterator1
 {
     typedef typename ::std::iterator_traits<_RandomAccessIterator1>::difference_type _DifferenceType;
     return __unseq_backend::__simd_transform_reduce(
-        __last1 - __first1, __init, __binary_op1,
+        __last1 - __first1, std::move(__init), __binary_op1,
         [=, &__binary_op2](_DifferenceType __i) { return __binary_op2(__first1[__i], __first2[__i]); });
 }
 
@@ -70,7 +70,7 @@ __pattern_transform_reduce(_Tag, _ExecutionPolicy&&, _ForwardIterator1 __first1,
 {
     static_assert(__is_serial_tag_v<_Tag> || __is_parallel_forward_tag_v<_Tag>);
 
-    return __brick_transform_reduce(__first1, __last1, __first2, __init, __binary_op1, __binary_op2,
+    return __brick_transform_reduce(__first1, __last1, __first2, std::move(__init), __binary_op1, __binary_op2,
                                     typename _Tag::__is_vector{});
 }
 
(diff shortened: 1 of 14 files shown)


SergeyKopienko:
Let me ask some overall questions about this change:
1) we still have a lot of code like `, __init,` in some functions calls. I think some of these places may bring for as the same problem as you want to fix in this PR. How do you think?
Example:
```C++
template <class _ExecutionPolicy, class _ForwardIterator, class _Tp, class _BinaryOperation, class... _Events,
          oneapi::dpl::__internal::__enable_if_device_execution_policy_double_no_default<_ExecutionPolicy, int, _Tp,
                                                                                         _BinaryOperation, _Events...>>
auto
reduce_async(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _Tp __init,
             _BinaryOperation __binary_op, _Events&&... __dependencies)
{
    const auto __dispatch_tag = oneapi::dpl::__internal::__select_backend(__exec, __first);

    wait_for_all(std::forward<_Events>(__dependencies)...);
    auto ret_val = oneapi::dpl::__internal::__pattern_transform_reduce_async(
        __dispatch_tag, std::forward<_ExecutionPolicy>(__exec), __first, __last, __init, __binary_op,
        oneapi::dpl::identity{});
    return ret_val;
}
```

2) What it we on all levels of code excepting our external interface level rewrite `_Tp __init` as `_Tp&& __init` and then will not pass it or `move` it into the calls but `forward` it?

include/oneapi/dpl/pstl/parallel_backend_tbb.h
```diff
@@ -131,21 +132,21 @@ __parallel_reduce(oneapi::dpl::__internal::__tbb_backend_tag, _ExecutionPolicy&&
 template <class _Index, class _Up, class _Tp, class _Cp, class _Rp>
 struct __par_trans_red_body
 {
-    alignas(_Tp) char _M_sum_storage[sizeof(_Tp)]; // Holds generalized non-commutative sum when has_sum==true
-    _Rp _M_brick_reduce;                           // Most likely to have non-empty layout
+    oneapi::dpl::__internal::__lazy_ctor_storage<_Tp> __lazy_sum;
```
mmichel11: This usage of `__lazy_ctor_storage` looks correct to me, and I am fine with keeping it.

However, since this is host code is there a reason to not use `std::optional` here? The `_M_has_sum` member variable here seems to reinvent the wheel of functionality already in `std::optional<T>::operator bool, std::optional<T>::has_value`.

test/support/utils.h
```diff
@@ -1243,6 +1243,46 @@ struct Pow2
...
+
+    MoveOnlyWrapper(_T v) : value(v) {}
+
+    operator _T() const { return value; }
+
+    // Move constructor
+    MoveOnlyWrapper(MoveOnlyWrapper&&) = default;
+
+    // Move assignment operator
+    MoveOnlyWrapper& operator=(MoveOnlyWrapper&&) = default;
```
mmichel11, line 1261: Should we add a documentation update of our decision to require move assignment somewhere with the justification of why this is required to implement parallel reduce?

include/oneapi/dpl/pstl/parallel_backend_tbb.h
danhoeflinger: Its a valid question... Let me think about how much it would simplify the code to switch to `std::optional`, and how much overhead in memory it costs us...
I think certainly the performance / memory savings we get on the host are relatively minimal compared to on the GPU.

test/support/utils.h
danhoeflinger: Yes, I can add to the documentation in this PR.

include/oneapi/dpl/pstl/parallel_backend_tbb.h
danhoeflinger: Yeah, switching to std::optional lets us remove the destructor and uses parts of the standard in a way that shouldnt cost much memory or performance overhead.
Thanks, done.

test/support/utils.h
danhoeflinger: I've augmented the documentation.  I did not go into detail about why this is necessary as I do not really see a good location to have this discussion. It is not specific to parallel reduce, it even affects sequential reduce implementations.  There is a way to implement it with only move construction, but it requires recursive handling which creates a stack overflow if the size is too large.

danhoeflinger: @akukanov, Would it be reasonable to link to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0571r2.html from our documentation?

danhoeflinger: I've added the link for now.

include/oneapi/dpl/pstl/numeric_impl.h
```diff
@@ -89,12 +89,12 @@ __pattern_transform_reduce(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec,
...
                 return __binary_op2(*__i, *(__first2 + (__i - __first1)));
             },
-            __init,
+            std::move(__init),
             __binary_op1, // Combine
             [__first1, __first2, __binary_op1, __binary_op2](_RandomAccessIterator1 __i, _RandomAccessIterator1 __j,
                                                              _Tp __init) -> _Tp {
-                return __internal::__brick_transform_reduce(__i, __j, __first2 + (__i - __first1), __init, __binary_op1,
-                                                            __binary_op2, _IsVector{});
+                return __internal::__brick_transform_reduce(__i, __j, __first2 + (__i - __first1), std::move(__init),
```
SergeyKopienko: How we have `__init` inside lambda and how we can move it on this level?

include/oneapi/dpl/pstl/numeric_impl.h
```diff
@@ -151,9 +151,11 @@ __pattern_transform_reduce(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec,
     return __internal::__except_handler([&]() {
         return __par_backend::__parallel_transform_reduce(
             __backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec), __first, __last,
-            [__unary_op](_RandomAccessIterator __i) mutable { return __unary_op(*__i); }, __init, __binary_op,
+            [__unary_op](_RandomAccessIterator __i) mutable { return __unary_op(*__i); }, std::move(__init),
+            __binary_op,
             [__unary_op, __binary_op](_RandomAccessIterator __i, _RandomAccessIterator __j, _Tp __init) {
-                return __internal::__brick_transform_reduce(__i, __j, __init, __binary_op, __unary_op, _IsVector{});
+                return __internal::__brick_transform_reduce(__i, __j, std::move(__init), __binary_op, __unary_op,
```
SergeyKopienko: The same

include/oneapi/dpl/pstl/numeric_impl.h
```diff
@@ -369,8 +371,8 @@ __brick_adjacent_difference(_RandomAccessIterator1 __first, _RandomAccessIterato
 
     auto __n = __last - __first;
     *__d_first = *__first;
-    return __unseq_backend::__simd_walk_n(__n - 1,
-        [&__op](_ReferenceType1 __x, _ReferenceType1 __y, _ReferenceType2 __z) { __z = __op(__x, __y); },
+    return __unseq_backend::__simd_walk_n(
```
SergeyKopienko: May we revert changes at this line at all?

include/oneapi/dpl/pstl/numeric_impl.h
danhoeflinger: its actually a poorly named different variable which is an argument to the lambda.  I'll adjust it.

SergeyKopienko (dismissed):
LGTM

PR #2357, opened 2025-07-27
Optimize scan KT and support user defined binary operators

mmichel11:
This PR reworks our scan kernel template by switching from an SLM driven scan where all data is loaded and operated in SLM to a sub-group driven scan where data is primarily stored and operated on in hardware registers. Additionally, this PR adds the building blocks necessary to implement future scan-like kernel templates.

The following changes are made:
- `internal/cooperative_lookback.h` is introduced to separate the lookback implementation from inclusive scan. We additionally provide an optimized path based on bit packing status and prefix flags for certain types.
- `internal/work_group/work_group_scan.h` and `internal/sub_group/sub_group_scan.h` implement optimized work-group and sub-group level scans, enabling support for user defined operators beyond SYCL function objects with known identities. This implementation and API may be used to propose improvements to SYCL itself as it has been shown to perform significantly better.
- The reworked implementation also serves as a single work-group scan without any extra kernels compiled. At runtime we determine if a single tile is present and skip temporary storage allocation and the init kernel if so. The existing single work-group path is removed as it does not perform better.

The initial implementation only supported SYCL function objects with known identities. User defined operators are now supported so long as the type satisfies the following conditions: 
- Scan input type must be convertible to output type
- Type must be copy constructible / assignable.

### DIFF
--- a/include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
+++ b/include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
@@ -162,7 +162,7 @@ struct __scan_status_flag<__sub_group_size, _T, std::enable_if_t<__can_combine_s
         _PackedStatusPrefixT __tile_status_prefix;
         _FlagStorageType __tile_flag = __initialized_status;
         // Load flag from a previous tile based on my local id.
-        // Spin until every work-item in this subgroup reads a valid status
+        // Spin until every work-item in this subgroup reads a valid (non-initial) status
         do
         {
             if (__tile_flag == __initialized_status)
@@ -291,7 +291,7 @@ struct __scan_status_flag<__sub_group_size, _T, std::enable_if_t<!__can_combine_
     {
         _FlagStorageType __tile_flag = __initialized_status;
         // Load flag from a previous tile based on my local id.
-        // Spin until every work-item in this subgroup reads a valid status
+        // Spin until every work-item in this subgroup reads a valid (non-initial) status
         do
         {
             if (__tile_flag == __initialized_status)
--- a/include/oneapi/dpl/experimental/kt/internal/work_group/work_group_scan.h
+++ b/include/oneapi/dpl/experimental/kt/internal/work_group/work_group_scan.h
@@ -46,11 +46,11 @@ __work_group_scan_impl(const _NdItem& __item, _SlmAcc __local_acc,
         oneapi::dpl::__internal::__dpl_ceiling_div(__items_in_scan, __sub_group_size * __iters_per_item);
 
     const std::uint32_t __items_in_sub_group_scan = __sub_group_size * __iters_per_item;
-    // Perform scan at sub-group level. For non active sub-groups, default constructed type returned.
-    // Due to observed performance regressions when changing this behavior, non-full sub-groups still perform scans over
-    // their full set of values. The returned result is defined as each array element is initialized by the
-    // default constructor, but not part of the actual scan and is ignored. The __sub_group_carry in this case does
-    // not affect the scan as it only occurs with the last sub-group.
+    // Perform scan at sub-group level. For non active sub-groups, we pad these with the last element and ultimately
+    // discard the scan value. Due to observed performance regressions when changing this behavior, non-full sub-groups
+    // still perform scans over their full set of values. The returned result is defined as each array element is
+    // initialized by the last input, but not part of the actual scan and is ignored. The __sub_group_carry in
+    // this case does not affect the scan as it only occurs with the last sub-group.
     //
     // TODO: we should analyze why limiting the sub-group scan causes performance regressions.
     _InputType __sub_group_carry = __sub_group_scan<__sub_group_size, __iters_per_item>(
--- a/include/oneapi/dpl/experimental/kt/single_pass_scan.h
+++ b/include/oneapi/dpl/experimental/kt/single_pass_scan.h
@@ -191,11 +191,11 @@ struct __single_pass_scan_kernel_func
         // Making full / not full case a bool template parameter and compiling two separate functions significantly improves performance
         // over run-time checks immediately before load / store.
         if (__sub_group_next_offset <= __n)
-            impl<true>(__item, __sub_group, __tile_id, __work_group_offset, __sub_group_current_offset,
-                       __sub_group_next_offset);
+            impl</*__is_full=*/true>(__item, __sub_group, __tile_id, __work_group_offset, __sub_group_current_offset,
+                                     __sub_group_next_offset);
         else
-            impl<false>(__item, __sub_group, __tile_id, __work_group_offset, __sub_group_current_offset,
-                        __sub_group_next_offset);
+            impl</*__is_full=*/false>(__item, __sub_group, __tile_id, __work_group_offset, __sub_group_current_offset,
+                                      __sub_group_next_offset);
     }
 };
 
@@ -289,8 +289,8 @@ __single_pass_scan(sycl::queue __queue, _InRange&& __in_rng, _OutRange&& __out_r
     }
     sycl::event __prev_event = __single_pass_scan_submitter<__sub_group_size, __data_per_workitem, __workgroup_size,
                                                             _Type, _FlagType, _SinglePassScanKernel>{}(
-        __queue, __fill_event, __in_rng, __out_rng, __binary_op, __n, __atomic_id_ptr, __lookback_storage,
-        __status_flags_size, __num_wgs);
+        __queue, __fill_event, std::forward<_InRange>(__in_rng), std::forward<_OutRange>(__out_rng), __binary_op, __n,
+        __atomic_id_ptr, __lookback_storage, __status_flags_size, __num_wgs);
     // In the single tile case, we can return the event asynchronously as we do not need to free temporary storage.
     if (__is_single_tile)
         return __prev_event;


include/oneapi/dpl/experimental/kt/internal/sub_group/sub_group_scan.h
```diff
@@ -0,0 +1,171 @@
...
+    else
+    {
+        const std::uint16_t __limited_iters_per_item =
+            oneapi::dpl::__internal::__dpl_ceiling_div(__items_in_scan, __sub_group_size);
+        std::uint16_t __i = 0;
+        if (__limited_iters_per_item == 1)
+        {
+            __sub_group_scan_partial<__sub_group_size, /*__is_inclusive*/ true, /*__init_present*/ false>(
+                __sub_group, __input[__i], __binary_op, __carry,
+                __items_in_scan - __i * __iters_per_item * __sub_group_size);
```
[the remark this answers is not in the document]
mmichel11: Yes, this is a bug. It turns out we were always doing full sub-group scans. This was not triggering any tests to fail as only the last carry out in scan is affected but has been fixed.

Update: Removing the full sub-group scans caused a significant regression. The behavior was not incorrect. I have added a comment explaining.

include/oneapi/dpl/experimental/kt/internal/sub_group/sub_group_scan.h
```diff
@@ -0,0 +1,171 @@
...
+            __sub_group_scan<__sub_group_size, /*__is_inclusive*/ true, /*__init_present*/ false>(
+                __sub_group, __input[__i++], __binary_op, __carry);
+            for (; __i < __limited_iters_per_item - 1; ++__i)
+            {
+                __sub_group_scan<__sub_group_size, /*__is_inclusive*/ true, /*__init_present*/ true>(
+                    __sub_group, __input[__i], __binary_op, __carry);
+            }
+            __sub_group_scan_partial<__sub_group_size, /*__is_inclusive*/ true, /*__init_present*/ true>(
+                __sub_group, __input[__i], __binary_op, __carry,
+                __items_in_scan - __i * __iters_per_item * __sub_group_size);
```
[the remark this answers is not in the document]
mmichel11: Fixed similar to above.

include/oneapi/dpl/experimental/kt/internal/work_group/work_group_scan.h
```diff
@@ -0,0 +1,183 @@
...
+//
+// sub_group 0: work_group_id 0:  0,  4,  8,  12
+//              work_group_id 1:  1,  5,  9,  13
+//              work_group_id 2:  2,  6,  10, 14
+//              work_group_id 3:  3,  7,  11, 15
+//
+// sub_group 1: work_group_id 4:  16, 20, 24, 28
+//              work_group_id 5:  17, 21, 25, 29
+//              work_group_id 6:  18, 22, 26, 30
+//              work_group_id 7:  19, 26, 27, 31
```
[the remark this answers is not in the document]
mmichel11: Fixed this typo.

include/oneapi/dpl/experimental/kt/internal/sub_group/sub_group_scan.h
```diff
@@ -0,0 +1,169 @@
...
+{
+
+namespace gpu
+{
+
+namespace __impl
+{
+
+// This implementation models what is defined in pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h with the
+// default constructibility requirement removed for simplification for the types supported in the KT.
```
danhoeflinger: Do we lose anything performance-wise to just use the existing implementation without default constructibility requirements?

Does this provide some extra functionality that that does not?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -307,96 +233,84 @@ template <bool _Inclusive, typename _InRange, typename _OutRange, typename _Bina
...
+        __status_flags_size = __num_wgs + __status_flag_padding;
+        __mem_bytes = _FlagStorageType::get_reqd_storage(__status_flags_size) + sizeof(std::uint32_t);
+        __device_mem = reinterpret_cast<std::byte*>(sycl::malloc_device(__mem_bytes, __queue));
+        __atomic_id_ptr = reinterpret_cast<std::uint32_t*>(__device_mem + __mem_bytes - sizeof(std::uint32_t));
+        if (!__device_mem)
+            throw std::bad_alloc();
+        _FlagStorageType __lookback_storage(__device_mem, __mem_bytes, __status_flags_size);
+        sycl::event __fill_event =
+            __lookback_init_submitter<__sub_group_size, _FlagType, _Type, _BinaryOp, _LookbackInitKernel>{}(
+                __queue, __atomic_id_ptr, __lookback_storage, __status_flags_size, __status_flag_padding);
```
danhoeflinger: I may prefer to unify the call to `__single_pass_scan_submitter`, and have this section in the if statement (if !__is_single_tile)

However, you would still need to optionally wait / free on the same conditional, I believe.

include/oneapi/dpl/experimental/kt/internal/sub_group/sub_group_scan.h
mmichel11: There are performance reasons to keep the default constructability requirement we had discussed (see how the work group scan always performs a full sub-group scan, relying on default constructability for non full work groups) Thinking about this more it may be mitigated (e.g. always loading the last element in the input as a dummy placeholder for non full items so we can still do full sub-group scans).

I will do some experiments and remove the default constructability requirement via `__lazy_ctor_storage` if there is no performance impact. This is preferable for long-term integration into the mainline scan.

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
```diff
@@ -0,0 +1,404 @@
...
+    set_init()
+    {
+        _PackedStatusPrefixT __packed_flag = __initialized_status;
+        __packed_flag_ref = _PackedStatusPrefixT{__packed_flag};
+    }
+
+    _FlagStorageType
+    get_status(_PackedStatusPrefixT __packed) const
+    {
+        _PackedStatusPrefixT __prefix_mask = ~_PackedStatusPrefixT(0) >> __half_status_prefix_bits;
```
adamfidel: Could this be `constexpr`?

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
```diff
@@ -0,0 +1,404 @@
...
+        return static_cast<_FlagStorageType>(__packed & __prefix_mask);
+    }
+
+    _T
+    get_value(_PackedStatusPrefixT __packed) const
+    {
+        _PackedStatusPrefixT __prefix_mask = ~_PackedStatusPrefixT(0) << __half_status_prefix_bits;
```
adamfidel: Same `constexpr` question

include/oneapi/dpl/experimental/kt/internal/sub_group/sub_group_scan.h
mmichel11: Removed the default constructability requirement without affecting performance. The major changes:

- `__grf_partials` and the work-group / sub-group scans expect the array to be of type `__lazy_ctor_storage<_T>`.
- `__grf_partials` is padded with `__n-1`th element for non-full cases. This lets us still perform full sub-group scans in all cases without hurting performance.
- The lookback step initializes all of the prefix flags to `__in_rng[0]` instead of the default constructed type.

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
mmichel11: Yes, I made it constexpr.

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
mmichel11: Added

include/oneapi/dpl/experimental/kt/single_pass_scan.h
mmichel11: I agree it looks better this way. I made this change.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -38,267 +40,197 @@ namespace __impl
...
-template <typename _FlagType, typename _Type, typename _BinaryOp, typename _KernelName>
-struct __lookback_init_submitter;
+template <std::uint8_t __sub_group_size, std::uint16_t __data_per_workitem, std::uint16_t __workgroup_size,
+          typename _Type, typename _FlagType, typename _KernelName>
+struct __single_pass_scan_submitter;
 
-template <typename _FlagType, typename _Type, typename _BinaryOp, typename... _Name>
-struct __lookback_init_submitter<_FlagType, _Type, _BinaryOp,
-                                 oneapi::dpl::__par_backend_hetero::__internal::__optional_kernel_name<_Name...>>
+template <std::uint8_t __sub_group_size, std::uint16_t __data_per_workitem, std::uint16_t __workgroup_size,
```
adamfidel, line 49: Is `std::uint16_t` large enough for the workgroup size? It appears to be passed in as a `size_t` from the kernel params.

mmichel11: Yes, the `kernel_param` struct defines these as such:
```cpp
    static constexpr std::uint16_t data_per_workitem = __data_per_work_item;
    static constexpr std::uint16_t workgroup_size = __work_group_size;
```
We were storing them as `size_t` types in the implementation. I switched these types to `uint16_t`.

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
```diff
@@ -0,0 +1,423 @@
...
+        {
+            __full_tile_found = __lookback_iter(/*__is_initialized*/ std::true_type{}, __tile);
+        }
+        if (__subgroup.get_local_id() == 0)
+        {
+            __local_flag.set_full(__binary_op(__running.__v, __local_reduction));
+        }
+        const _T __return_val = __running.__v;
+        __running.__destroy();
+        return __return_val;
```
danhoeflinger: Its optional, but you can use `oneapi::dpl::__internal::__scoped_destroyer` to destroy `__running` as this function leaves scope and then you can return `__running.__v` directly without the copy.

include/oneapi/dpl/experimental/kt/internal/sub_group/sub_group_scan.h
```diff
@@ -135,22 +136,24 @@ __sub_group_scan(const _SubGroup& __sub_group, _InputType __input[__iters_per_it
...
             }
             __sub_group_scan_partial<__sub_group_size, /*__is_inclusive*/ true, /*__init_present*/ true>(
-                __sub_group, __input[__i], __binary_op, __carry, __items_in_scan - __i * __sub_group_size);
+                __sub_group, __input[__i].__v, __binary_op, __carry, __items_in_scan - __i * __sub_group_size);
         }
     }
-    return __carry;
+    const _InputType __return_carry = __carry.__v;
+    __carry.__destroy();
+    return __return_carry;
```
danhoeflinger: same comment here with `__scoped_destroyer`.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -38,267 +40,197 @@ namespace __impl
...
+                if (__sub_group_current_offset + __sub_group_local_id + __sub_group_size * __i < __n)
                 {
-                    __tile_vals[__local_id + __workgroup_size * __i] =
-                        __in_rng[__wg_current_offset + __local_id + __workgroup_size * __i];
+                    __out_rng[__sub_group_current_offset + __sub_group_local_id + __sub_group_size * __i] =
+                        __grf_partials[__i].__v;
+                    __grf_partials[__i].__destroy();
                 }
+                else
+                    __grf_partials[__i].__destroy();
```
danhoeflinger: Could remove the `else` and just pull this line out of the `if` block.

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
mmichel11: Rebased and used `__scoped_destroyer`

include/oneapi/dpl/experimental/kt/internal/sub_group/sub_group_scan.h
mmichel11: Done

include/oneapi/dpl/experimental/kt/single_pass_scan.h
mmichel11: Thanks, done.

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
```diff
@@ -0,0 +1,422 @@
...
+    {
+        _PackedStatusPrefixT __packed_flag = __full_status;
+        _PackedStatusPrefixT __integral_bits =
+            static_cast<_PackedStatusPrefixT>(sycl::bit_cast<_TIntegralBitsType, _T>(__val));
+        __packed_flag |= __integral_bits << __half_status_prefix_bits;
+        __atomic_packed_flag.store(__packed_flag);
+    }
+
+    // For initialization routines, we do not need atomicity, so we can write through the
+    // reference directly.
```
adamfidel: Can you move this comment to line 128 (right before the direct write)? I think it would be clearer what it is referring to. Same for the `set_init` function

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -38,267 +40,195 @@ namespace __impl
...
     std::size_t __n;
-    _StatusFlags __status_flags;
+    _TileIdxT* __atomic_id_ptr;
+    typename __scan_status_flag<__sub_group_size, _Type>::storage __lookback_storage;
     std::size_t __status_flags_size;
-    _StatusValues __status_vals_full;
-    _StatusValues __status_vals_partial;
-    std::size_t __current_num_items;
-    _TileVals __tile_vals;
+    std::size_t __num_tiles;
```
adamfidel:
```suggestion
    _TileIdxT __num_tiles;
```
The tile index is 32-bits, so we can probably restrict the size of the number of tiles here.

adamfidel (commented):
I've gone through the changes and I don't see any glaring issues. I just have a few nitpicks but I think it's in a good shape to merge.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
mmichel11: Done, I had to change a few other types as well.

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
mmichel11: Yes, done.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -38,267 +40,196 @@ namespace __impl
...
-        sycl::joint_inclusive_scan(__group, __tile_vals_ptr, __tile_vals_ptr + __wg_local_memory_size, __out_begin,
-                                   __binary_op, __prev_tile_reduction);
+        // Making full / not full case a bool template parameter and compiling two separate functions significantly improves performance
+        // over run-time checks immediately before load / store.
+        if (__sub_group_next_offset <= __n)
+            impl<true>(__item, __sub_group, __tile_id, __work_group_offset, __sub_group_current_offset,
+                       __sub_group_next_offset);
+        else
+            impl<false>(__item, __sub_group, __tile_id, __work_group_offset, __sub_group_current_offset,
+                        __sub_group_next_offset);
```
danhoeflinger:
```suggestion
            impl</*__is_full=*/true>(__item, __sub_group, __tile_id, __work_group_offset, __sub_group_current_offset,
                       __sub_group_next_offset);
        else
            impl</*__is_full=*/false>(__item, __sub_group, __tile_id, __work_group_offset, __sub_group_current_offset,
                        __sub_group_next_offset);
```

[nitpick] personal preference.

include/oneapi/dpl/experimental/kt/internal/work_group/work_group_scan.h
```diff
@@ -0,0 +1,201 @@
...
+    const std::uint8_t __active_sub_groups =
+        oneapi::dpl::__internal::__dpl_ceiling_div(__items_in_scan, __sub_group_size * __iters_per_item);
+
+    const std::uint32_t __items_in_sub_group_scan = __sub_group_size * __iters_per_item;
+    // Perform scan at sub-group level. For non active sub-groups, default constructed type returned.
+    // Due to observed performance regressions when changing this behavior, non-full sub-groups still perform scans over
+    // their full set of values. The returned result is defined as each array element is initialized by the
+    // default constructor, but not part of the actual scan and is ignored. The __sub_group_carry in this case does
+    // not affect the scan as it only occurs with the last sub-group.
+    //
```
danhoeflinger: Does this comment need to be updated?

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -38,267 +40,196 @@ namespace __impl
...
+        {
+            load_global_to_grf<__is_full>(__grf_partials, __sub_group_current_offset, __sub_group_local_id);
+            _Type __local_reduction = __work_group_scan<__sub_group_size, __data_per_workitem>(
+                __item, __slm, __grf_partials, __binary_op, __this_tile_elements);
+            if (__item.get_local_id(0) == 0 && __num_tiles > 1)
             {
-                __flag.set_partial(__local_reduction);
+                _FlagType __flag(__lookback_storage, __tile_id);
+                __flag.set_full(__local_reduction);
             }
```
danhoeflinger: Are we leaving something on the table here to have this local reduction status flag setting take place after the zeroth full work group scan?

Could this not happen in the lookback callback section of the zeroth tile?

Perhaps there should be some separate callback function for the zeroth tile rather than a dummy which merely sets the full status and value.

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
```diff
@@ -0,0 +1,422 @@
...
+    {
+        return __status == __full_status ? __atomic_full_value.load() : __atomic_partial_value.load();
+    }
+
+    std::pair<_FlagStorageType, _T>
+    spin_and_get(const __dpl_sycl::__sub_group& __sub_group) const
+    {
+        _FlagStorageType __tile_flag = __initialized_status;
+        // Load flag from a previous tile based on my local id.
+        // Spin until every work-item in this subgroup reads a valid status
```
danhoeflinger:
```suggestion
        // Spin until every work-item in this subgroup reads a valid (non-initial) status
        // to make sure all previous tiles have completed at least their local reduction
```

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
```diff
@@ -0,0 +1,422 @@
...
+        return sycl::bit_cast<_T, _TIntegralBitsType>(__integral_bits);
+    }
+
+    std::pair<_FlagStorageType, _T>
+    spin_and_get(const __dpl_sycl::__sub_group& __sub_group) const
+    {
+        _PackedStatusPrefixT __tile_status_prefix;
+        _FlagStorageType __tile_flag = __initialized_status;
+        // Load flag from a previous tile based on my local id.
+        // Spin until every work-item in this subgroup reads a valid status
```
danhoeflinger:
```suggestion
        // Spin until every work-item in this subgroup reads a valid (non-initial) status
        // to make sure all previous tiles have completed at least their local reduction
```

include/oneapi/dpl/experimental/kt/single_pass_scan.h
```diff
@@ -307,80 +238,65 @@ template <bool _Inclusive, typename _InRange, typename _OutRange, typename _Bina
...
+        if (!__device_mem)
+            throw std::bad_alloc();
+        __lookback_storage = _FlagStorageType(__device_mem, __mem_bytes, __status_flags_size);
+        __fill_event =
+            __lookback_init_submitter<__sub_group_size, _FlagType, _InRange, _Type, _BinaryOp, _LookbackInitKernel>{}(
+                __queue, __atomic_id_ptr, __in_rng, __lookback_storage, __status_flags_size, __status_flag_padding);
+    }
+    sycl::event __prev_event = __single_pass_scan_submitter<__sub_group_size, __data_per_workitem, __workgroup_size,
+                                                            _Type, _FlagType, _SinglePassScanKernel>{}(
+        __queue, __fill_event, __in_rng, __out_rng, __binary_op, __n, __atomic_id_ptr, __lookback_storage,
```
danhoeflinger: You should be able to forward `__in_rng` and `__out_rng`.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
mmichel11: This is a good point. I made the following changes to resolve it:

- The first tile now uses a callback called `__cooperative_lookback_first_tile` If there is a single tile, then it is essentially a noop. Otherwise, the first item sets the full status and the callback returns.
- The cooperative lookback structs set a member `__apply_prefix` queried by the work-group scan. `__cooperative_lookback` sets this to true and `__cooperative_lookback_first_tile` sets it to false. This is used to determine if we apply the prefix.
- The return value of the callback is returned through a reference and the function is made void. This makes it easy to have a noop case where we do not modify the reference.

include/oneapi/dpl/experimental/kt/single_pass_scan.h
mmichel11: Good point, done.

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
mmichel11: Done

include/oneapi/dpl/experimental/kt/internal/cooperative_lookback.h
mmichel11: Done

include/oneapi/dpl/experimental/kt/single_pass_scan.h
mmichel11: Done

include/oneapi/dpl/experimental/kt/internal/work_group/work_group_scan.h
mmichel11: Yes, it does. I have updated it to mention how we pad with the last element and then discard scan values for non full sub-groups.

PR #2358, opened 2025-07-28
Fix performance issue of `parallel_find_or` - 2-Kernels approach

SergeyKopienko:
This is alternative variant of https://github.com/uxlfoundation/oneDPL/pull/2349 based on 2-kernels approach.

This PR implements a performance optimization for the `parallel_find_or` algorithm using a 2-kernels approach. The change replaces the previous synchronization mechanism that relied on temporary SYCL buffer destruction with an explicit initialization kernel and improved group coordination.

- Introduces a new initialization kernel to set up scratch storage and group counters
- Replaces buffer-based synchronization with atomic group counter tracking
- Adds conditional compilation support for FPGA devices with 32-bit atomic limitations

### Details
Performance can be further improved in `__parallel_find_or`, particularly for discrete GPU devices, by writing single element results in device memory back to the host through USM host memory instead of going through a `memcpy` operation.

This improves performance in two ways:
- Works around an issue where level-zero `memcpy` operations are taking several hundred microseconds for single element transfers.
- We have observed that USM host memory performs better than `memcpy` operations for single element transfers to-and-from the device, so this change is a general performance optimization as opposed to being a temporary workaround.

The one new kernel, initialization, and additional writeback from existing Kernel (from the last work-group), replace implicit `memcpy` operations that occur in the original buffer based version.

Thanks @mmichel11 for preparing this details for description.


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1480,74 +1483,141 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
...
+template <bool __or_tag_check, typename... KernelNameInit, typename... KernelName>
+struct __parallel_find_or_impl_multiple_wgs<__or_tag_check, __internal::__optional_kernel_name<KernelNameInit...>,
+                                            __internal::__optional_kernel_name<KernelName...>>
 {
+#if _ONEDPL_FPGA_DEVICE
+    // FPGA devices don't support 64-bit atomics
+    using _GroupCounterType = uint32_t;
+#else
+    using _GroupCounterType = std::size_t;
+#endif
```
mmichel11: Is it reasonable to assume that the number of work-groups can fit in a 32-bit unsigned integer and just use that everywhere? If not, do we need to check if the device has `sycl::aspect::atomic64`?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1480,74 +1483,141 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
...
+                        __group_counter_acc_w);
+                *__group_counter_ptr = 0;
+            });
+        });
+
+        // main parallel_for
+        auto __event = __q.submit([&](sycl::handler& __cgh) {
+            oneapi::dpl::__ranges::__require_access(__cgh, __rngs...);
+
+            auto __scratch_acc_rw = __result_storage.template __get_scratch_acc<sycl::access_mode::read_write>(
```
mmichel11: Should `__scratch_acc_rw` and `__group_counter_acc_rw` be `__no_init`? We initialized the value in the preceding kernel, so we depend on it, correct?

SergeyKopienko: Fixed: good catch

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: I think we may just use `using _GroupCounterType = uint32_t;` for all cases.
Fixed.

SergeyKopienko:
> Could you please reduce the number of `auto` variables to only those where the actual type cannot be easily determined or would be excessive?

Sure: fixed in the `__parallel_find_or_impl_one_wg::operator()` and in the `__parallel_find_or_impl_multiple_wgs::operator()`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1466,7 +1469,7 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
                         __found_local = __dpl_sycl::__reduce_over_group(__item.get_group(), __found_local,
                                                                         typename _BrickTag::_LocalResultsReduceOp{});
 
-                    // Set local found state value value to global state to have correct result
+                    // Set local found state value to global state
```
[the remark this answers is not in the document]
akukanov:
```suggestion
                    // Copy the locally found state value to the global state
```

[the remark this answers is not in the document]
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1480,74 +1483,135 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
...
+
+                    if (__local_idx == 0)
+                    {
+                        // Set local found state value to global atomic if we found something in the current work-group
+                        if (__found_local != __init_value)
                         {
-                            __dpl_sycl::__atomic_ref<_AtomicType, sycl::access::address_space::global_space> __found(
-                                *__dpl_sycl::__get_accessor_ptr(__result_sycl_buf_acc));
+                            _AtomicType* __scratch_ptr =
+                                __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__scratch_acc_rw);
```
danhoeflinger: Can we pull this up a level, or to the beginning of the kernel to remove the redundancy with 1602?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1480,74 +1483,135 @@ struct __parallel_find_or_impl_one_wg<__or_tag_check, __internal::__optional_ker
...
-        }
 
-        return __result;
+                        _GroupCounterType* __group_counter_ptr =
+                            __result_and_scratch_storage_group_counter_t::__get_usm_or_buffer_accessor_ptr(
+                                __group_counter_acc_rw);
+                        __atomic_ref_t<_GroupCounterType> __group_counter(*__group_counter_ptr);
+
+                        // Copy data back from scratch part to result part when we are in the last work-group
+                        if (++__group_counter == __n_groups)
```
danhoeflinger: I actually don't mind the comment from copilot here to make it more explicit that the `++` operator is done atomically.
I know that this is equivalent, but it may be worthwhile to confirm this in a comment or with a `fetch_add(1)`.

Optional though, I'm ok with whatever you think is best here, with `__atomic_ref_t` the line above, its perhaps clear enough.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: Sure. Fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h

danhoeflinger (approved):
LGTM

danhoeflinger (approved):
LGTM, please wait for green CI

Issue #2296, opened 2025-06-05
`inclusive_scan` and `exclusive_scan` over reverse iterators with different source iterator types fails to compile


mmichel11:
**Describe the Bug:**
When performing an inclusive or exclusive scan with host or device backends, the scan fails to compile if both the source and destination sequences are provided as reverse iterators and the underlying source iterators are different, non equality comparable types. This issue likely affects other iterator adaptors as well.

**To Reproduce:**
The following reproducer may be used:
```cpp
// icpx [-DONEDPL_USE_DPCPP_BACKEND=1] scan_reverse_iterator.cpp
#include <oneapi/dpl/execution>
#include <oneapi/dpl/algorithm>
#include <oneapi/dpl/numeric>
#include <vector>
#include <iostream>

#if ONEDPL_USE_DPCPP_BACKEND
#   define POLICY oneapi::dpl::execution::dpcpp_default
#else
#   define POLICY oneapi::dpl::execution::par_unseq
#endif

int main() {
    // Create a vector of bool values
    std::vector<bool> input = {true, false, true, true, false, true};

    std::cout << "Original vector of booleans:" << std::endl;
    for (const auto& val : input) {
        std::cout << (val ? "true" : "false") << " ";
    }
    std::cout << std::endl;

    // Create a vector to store the result (int type)
    std::vector<int> result(input.size());

    // Create reverse iterators
    auto input_rbegin = std::rbegin(input);
    auto input_rend = std::rend(input);
    auto result_rbegin = std::rbegin(result);

    // Use exclusive_scan with reverse iterators to convert bool to int
    // This will scan from right to left (due to reverse iterators)
    // The initial value (0) will appear at the rightmost position
    oneapi::dpl::exclusive_scan(
        POLICY,                  // Parallel execution policy
        input_rbegin,            // Start of reversed input range
        input_rend,              // End of reversed input range
        result_rbegin,           // Start of reversed output range
        int{0}                   // Initial value
    );

    std::cout << "\nOriginal vector (left to right):" << std::endl;
    for (size_t i = 0; i < input.size(); ++i) {
        std::cout << (input[i] ? 1 : 0) << " ";
    }
    std::cout << std::endl;

    std::cout << "\nResult of exclusive_scan with reverse iterators:" << std::endl;
    for (const auto& val : result) {
        std::cout << val << " ";
    }
    std::cout << std::endl;

    return 0;
}
```
[...]

PR #2359, opened 2025-07-29
Fix compile error in `inclusive_scan` and `exclusive_scan` over reverse iterators with different source iterator types

SergeyKopienko:
This PR fixes an issue with iterator comparison in the oneDPL library (issue #2296) by improving the `__iterators_possibly_equal` function to properly handle iterator comparisons with different value types.

- Adds value type checking before attempting iterator equality comparison
- Restructures the conditional logic to avoid compilation errors when iterators have incompatible types

### Implementation details
```C++
template <typename _Iterator1, typename _Iterator2, typename = void>
struct __is_equality_comparable_with : std::false_type
{
};

// All with implemented operator ==
    requires std::equality_comparable_with<std::decay_t<_Iterator1>, std::decay_t<_Iterator2>>
struct __is_equality_comparable_with<_Iterator1, _Iterator2, std::void_t<int>> : std::true_type
{
};
#else
// All with implemented operator ==
struct __is_equality_comparable_with<
    _Iterator1, _Iterator2,
    std::void_t<decltype(std::declval<std::decay_t<_Iterator1>>() == std::declval<std::decay_t<_Iterator2>>())>>
    : std::true_type
{
};
#endif

constexpr bool
__iterators_possibly_equal_impl(_Iterator1 __it1, _Iterator2 __it2)
{
    if constexpr (__is_equality_comparable_with<_Iterator1, _Iterator2>::value)
    {
        return __it1 == __it2;
    }
    else if constexpr (__is_equality_comparable_with<_Iterator2, _Iterator1>::value)
    {
        return __it2 == __it1;
    }

    return false;
}

#if !_ONEDPL_CPP20_CONCEPTS_PRESENT
template <typename _Iterator, typename = void>
struct __has_value_type : std::false_type
{
};

template <typename _Iterator>
struct __has_value_type<_Iterator, std::void_t<typename std::iterator_traits<_Iterator>::value_type>> : std::true_type
{
};
#endif

// Checks if two iterators are possibly equal, i.e. if they can be compared for equality.
constexpr bool
__iterators_possibly_equal(_Iterator1 __it1, _Iterator2 __it2)
{
    // In C++20 and later the check for equality is done
    // via concepts std:: equality_comparable_with inside of __is_equality_comparable_with
    // so no additional checks here are needed.
    return __iterators_possibly_equal_impl(__it1, __it2);
#else
    // Before C++20 we can compare only the iterators of the same types.
    if constexpr (__has_value_type<_Iterator1>::value && __has_value_type<_Iterator2>::value)
    {
        if constexpr (std::is_same_v<typename std::iterator_traits<_Iterator1>::value_type,
                                     typename std::iterator_traits<_Iterator2>::value_type>)
        {
            return __iterators_possibly_equal_impl(__it1, __it2);
        }
    }
    else if constexpr (!__has_value_type<_Iterator1>::value && !__has_value_type<_Iterator2>::value)
    {
        return __iterators_possibly_equal_impl(__it1, __it2);
    }

    return false;
#endif // _ONEDPL_CPP20_CONCEPTS_PRESENT
}
```

Additional comments about the check that `value_type` exists in both iterators types:
```C++
        if constexpr (std::is_same_v<typename std::iterator_traits<_Iterator1>::value_type,
                                     typename std::iterator_traits<_Iterator2>::value_type>)
```
We have two reasons for check if `value type` is defined inside both iterators:
1. If `value type` not defined in both iterators, we can compare these iterators (of course, if `operator==()` is defined).
2. If `value type` is defined only in one iterator, we can't compare them at all due different iterators types.
3. If `value type` is defined in the both iterators, we should check that the types are the same (and it's not depends on is `operator==()` is defined or no).
Otherwise we have compile error at least in MS STL implementation for example from the issue #2296.

Also we should not break already existed tests in `sycl_iterator_operations.pass.cpp`:
1.
https://github.com/uxlfoundation/oneDPL/blob/d405c47756e006e23d9c061b91ba9fa01985eed8/test/general/sycl_iterator/sycl_iterator_operations.pass.cpp#L55
```C++
         EXPECT_FALSE(__iterators_possibly_equal(oneapi::dpl::begin(buf1), nullptr),
                      "wrong __iterators_possibly_equal result");
         EXPECT_FALSE(__iterators_possibly_equal(nullptr, oneapi::dpl::begin(buf2)),
                      "wrong __iterators_possibly_equal result");
```
2.
https://github.com/uxlfoundation/oneDPL/blob/d405c47756e006e23d9c061b91ba9fa01985eed8/test/general/sycl_iterator/sycl_iterator_operations.pass.cpp#L127
```C++
CustomIterator it1(CustomIterator::Tag{});
CustomIterator it2(CustomIterator::Tag{});
EXPECT_TRUE(__iterators_possibly_equal(it1, it2),
                       "wrong __iterators_possibly_equal result for custom iterator which is not default constructible");
```
where `CustomIterator` is defined as follows:
```C++
class CustomIterator
{
public:
    struct Tag { };
    CustomIterator(Tag) {}
    bool
    operator==(const CustomIterator&) const
    {
        return true;
    }
};
```
so it has not the definition of `value_type` inside.

### References
1) https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3351.pdf
2) https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2022/p2404r3.pdf - 2.1 Overview
3) https://en.cppreference.com/w/cpp/concepts/equality_comparable
4) https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p2404r0.pdf
5) https://stackoverflow.com/questions/66937947/why-is-unique-ptr-not-equality-comparable-with-nullptr-t-in-c20/66938093#66938093


include/oneapi/dpl/pstl/utils.h
```diff
@@ -837,18 +837,22 @@ template <typename _Iterator1, typename _Iterator2>
...
-        return __it2 == __it1;
-    }
-    else
+    using _ValueType1 = typename std::iterator_traits<_Iterator1>::value_type;
+    using _ValueType2 = typename std::iterator_traits<_Iterator2>::value_type;
+
+    if constexpr (std::is_same_v<_ValueType1, _ValueType2>)
     {
-        return false;
+        if constexpr (__is_equality_comparable<_Iterator1, _Iterator2>::value)
```
SergeyKopienko: I have tried to rewrite `__is_equality_comparable` as
```C++
// All with implemented operator ==
struct __is_equality_comparable<
    _Iterator1, _Iterator2,
    std::void_t<
        std::conjunction<
            std::is_same<
                typename std::iterator_traits<_Iterator1>::value_type,
                typename std::iterator_traits<_Iterator2>::value_type
            >,
            std::integral_constant<
                decltype(std::declval<std::decay_t<_Iterator1>>() == std::declval<std::decay_t<_Iterator2>>()),
                true
            >
        >
    >
> : std::true_type
{
};
```
but it doesn't helps.
So I wrote this check here.

SergeyKopienko:
Closed - we have more compact implementation in https://github.com/uxlfoundation/oneDPL/pull/2371

PR #2360, opened 2025-07-29
[Draft] Dynamic Selection proposal for extra resource type

danhoeflinger:
Work in progress draft for adding an extra resource type to pair with a resource, which defaults to empty.

Allows passing in a vector of extra_resources to match the main resources, which can define any of 3 functions:
1) `initialize()` - called once per extra resource during creation of the backend if defined
2) `reset()` - called once per selection to "reset" the extra resource during `instrument_before_impl()` if defined
3) `clean_up()` - called once per extra resource when backend is destroyed if defined

Selection handle contain a function `get_extra_resource()` which provide the `extra_resource_t` object associated with the selected resource.

If a extra resources are provided, it is passed as the second argument to the provided submit function, otherwise it is not.

So far, this only tests the case without extra resources provided.  We need to test the case with extra resources as well.

(I also removed some strange formatting / unicode characters, they were causing warnings and made code difficult to read)


danhoeflinger:
closing in favor of #2367

PR #2366, opened 2025-08-01
Add reverse and reverse_copy parallel range algorithms

dmitriy-sobolev:
The details can be found in:

Interface - specification:
- https://github.com/uxlfoundation/oneAPI-spec/pull/635
- https://github.com/uxlfoundation/oneAPI-spec/pull/634

Rationale - RFC:
- https://github.com/uxlfoundation/oneDPL/pull/2269

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_ranges_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_ranges_impl.h
@@ -696,6 +696,31 @@ __pattern_remove_if(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy
     return std::ranges::remove_if(std::forward<_R>(__r), __pred, __proj);
 }
 
+//----------------------------------------------------------------------------------------------------------------------
+// __pattern_reverse
+//---------------------------------------------------------------------------------------------------------------------
+template <typename _Tag, typename _ExecutionPolicy, typename _R>
+void
+__pattern_reverse(_Tag __tag, _ExecutionPolicy&& __exec, _R&& __r)
+{
+    static_assert(__is_parallel_tag_v<_Tag> || typename _Tag::__is_vector{});
+
+    auto __begin = std::ranges::begin(__r);
+    auto __end = std::ranges::begin(__r) + std::ranges::size(__r);
+    oneapi::dpl::__internal::__pattern_reverse(__tag, std::forward<_ExecutionPolicy>(__exec), __begin, __end);
+}
+
+template <typename _ExecutionPolicy, typename _R>
+void
+__pattern_reverse(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy&&, _R&& __r)
+{
+    std::ranges::reverse(std::forward<_R>(__r));
+}
+
+//----------------------------------------------------------------------------------------------------------------------
+// __pattern_move
+//---------------------------------------------------------------------------------------------------------------------
+
 template <typename _Tag, typename _ExecutionPolicy, typename _InRange, typename _OutRange>
 void
 __pattern_move(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __r, _OutRange&& __out_r)
--- a/include/oneapi/dpl/pstl/algorithm_ranges_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_ranges_impl.h
@@ -696,7 +696,7 @@ __pattern_remove_if(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy
     return std::ranges::remove_if(std::forward<_R>(__r), __pred, __proj);
 }
 
-//----------------------------------------------------------------------------------------------------------------------
+//---------------------------------------------------------------------------------------------------------------------
 // __pattern_reverse
 //---------------------------------------------------------------------------------------------------------------------
 template <typename _Tag, typename _ExecutionPolicy, typename _R>
@@ -717,7 +717,32 @@ __pattern_reverse(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy&&
     std::ranges::reverse(std::forward<_R>(__r));
 }
 
-//----------------------------------------------------------------------------------------------------------------------
+//---------------------------------------------------------------------------------------------------------------------
+// __pattern_reverse_copy
+//---------------------------------------------------------------------------------------------------------------------
+template <typename _Tag, typename _ExecutionPolicy, typename _InRange, typename _OutRange>
+void
+__pattern_reverse_copy(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r)
+{
+    static_assert(__is_parallel_tag_v<_Tag> || typename _Tag::__is_vector{});
+
+    assert(std::ranges::size(__in_r) <= std::ranges::size(__out_r)); // for debug purposes only
+
+    auto __begin = std::ranges::begin(__in_r);
+    auto __end = __begin + std::ranges::size(__in_r);
+    oneapi::dpl::__internal::__pattern_reverse_copy(__tag, std::forward<_ExecutionPolicy>(__exec), __begin, __end,
+                                                    std::ranges::begin(__out_r));
+}
+
+template <typename _ExecutionPolicy, typename _InRange, typename _OutRange>
+void
+__pattern_reverse_copy(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy&&, _InRange&& __in_r,
+                        _OutRange&& __out_r)
+{
+    std::ranges::reverse_copy(std::forward<_InRange>(__in_r), std::ranges::begin(__out_r));
+}
+
+//---------------------------------------------------------------------------------------------------------------------
 // __pattern_move
 //---------------------------------------------------------------------------------------------------------------------
 
--- a/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
+++ b/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
@@ -899,6 +899,31 @@ struct __replace_fn
 
 inline constexpr __internal::__replace_fn replace;
 
+
+namespace __internal
+{
+struct __reverse_fn
+{
+    template<typename _ExecutionPolicy, std::ranges::random_access_range _R>
+    requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<_ExecutionPolicy>>
+        && std::permutable<std::ranges::iterator_t<_R>> && std::ranges::sized_range<_R>
+
+    std::ranges::borrowed_iterator_t<_R>
+    operator()(_ExecutionPolicy&& __exec, _R&& __r) const
+    {
+        const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
+
+        auto __last = std::ranges::begin(__r) + std::ranges::size(__r);
+        oneapi::dpl::__internal::__ranges::__pattern_reverse(
+            __dispatch_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r));
+        return std::ranges::borrowed_iterator_t<_R>{__last};
+    }
+
+}; //__reverse_fn
+} //__internal
+
+inline constexpr __internal::__reverse_fn reverse;
+
 // [alg.is_sorted_until]
 
 namespace __internal
--- a/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
+++ b/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
@@ -900,6 +900,8 @@ struct __replace_fn
 inline constexpr __internal::__replace_fn replace;
 
 
+// [alg.reverse]
+
 namespace __internal
 {
 struct __reverse_fn
@@ -915,7 +917,7 @@ struct __reverse_fn
 
         auto __last = std::ranges::begin(__r) + std::ranges::size(__r);
         oneapi::dpl::__internal::__ranges::__pattern_reverse(
-            __dispatch_tag, std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r));
+            __dispatch_tag, std::forward<_ExecutionPolicy>(__exec), __r);
         return std::ranges::borrowed_iterator_t<_R>{__last};
     }
 
@@ -923,6 +925,44 @@ struct __reverse_fn
 } //__internal
 
 inline constexpr __internal::__reverse_fn reverse;
+
+// [alg.reverse_copy]
+
+namespace __internal
+{
+struct __reverse_copy_fn
+{
+    template<typename _ExecutionPolicy, std::ranges::random_access_range _InRange,
+             std::ranges::random_access_range _OutRange>
+    requires oneapi::dpl::is_execution_policy_v<std::remove_cvref_t<_ExecutionPolicy>>
+        && std::ranges::sized_range<_InRange> && std::ranges::sized_range<_OutRange>
+        && std::indirectly_copyable<std::ranges::iterator_t<_InRange>, std::ranges::iterator_t<_OutRange>>
+
+    std::ranges::in_in_out_result<std::ranges::borrowed_iterator_t<_InRange>,
+                                  std::ranges::borrowed_iterator_t<_InRange>,
+                                  std::ranges::borrowed_iterator_t<_OutRange>>
+    operator()(_ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r) const
+    {
+        const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
+
+        using _Size = std::common_type_t<std::ranges::range_size_t<_InRange>, std::ranges::range_size_t<_OutRange>>;
+        const _Size __sz = std::ranges::min((_Size)std::ranges::size(__in_r), (_Size)std::ranges::size(__out_r));
+
+        auto __last_in = std::ranges::begin(__in_r) + std::ranges::size(__in_r);
+        auto __stop_in = std::ranges::begin(__in_r) + (std::ranges::size(__in_r) - __sz);
+        auto __stop_out = std::ranges::begin(__out_r) + __sz;
+
+        oneapi::dpl::__internal::__ranges::__pattern_reverse_copy(
+            __dispatch_tag, std::forward<_ExecutionPolicy>(__exec),
+            std::ranges::take_view(__in_r, __sz), std::ranges::take_view(__out_r, __sz));
+
+        return {__last_in, __stop_in, __stop_out};
+    }
+
+}; //__reverse_fn
+} //__internal
+
+inline constexpr __internal::__reverse_copy_fn reverse_copy;
 
 // [alg.is_sorted_until]
 
--- a/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
@@ -718,6 +718,21 @@ __pattern_reverse(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R
 // reverse_copy
 //------------------------------------------------------------------------
 
+template <typename _BackendTag, typename _ExecutionPolicy, typename _InRange, typename _OutRange>
+void
+__pattern_reverse_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r,
+                       _OutRange&& __out_r)
+{
+    const auto __n = std::ranges::size(__in_r);
+    if (__n == 0)
+        return;
+
+    using _DiffType = oneapi::dpl::__internal::__difference_t<_InRange>;
+    oneapi::dpl::__par_backend_hetero::__parallel_for(
+        _BackendTag{}, std::forward<_ExecutionPolicy>(__exec), unseq_backend::__reverse_copy<_DiffType>{__n},
+        __n, oneapi::dpl::__ranges::views::all_read(std::forward<_InRange>(__in_r)),
+        oneapi::dpl::__ranges::views::all_write(std::forward<_OutRange>(__out_r))).get(); // is a blocking call
+}
 
 //------------------------------------------------------------------------
 // move
--- a/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
@@ -696,6 +696,33 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, 
     return {std::ranges::begin(__r) + __idx, std::ranges::begin(__r) + __n};
 }
 
+//------------------------------------------------------------------------
+// reverse
+//------------------------------------------------------------------------
+
+template <typename _BackendTag, typename _ExecutionPolicy, typename _R>
+void
+__pattern_reverse(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R&& __r)
+{
+    const auto __n = std::ranges::size(__r);
+    if (__n <= 1)
+        return;
+
+    using _DiffType = oneapi::dpl::__internal::__difference_t<_R>;
+    oneapi::dpl::__par_backend_hetero::__parallel_for(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
+        unseq_backend::__reverse_functor<_DiffType>{__n}, __n / 2,
+        oneapi::dpl::__ranges::views::all(std::forward<_R>(__r))).__checked_deferrable_wait();
+}
+
+//------------------------------------------------------------------------
+// reverse_copy
+//------------------------------------------------------------------------
+
+
+//------------------------------------------------------------------------
+// move
+//------------------------------------------------------------------------
+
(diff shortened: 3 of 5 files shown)


include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -696,6 +696,58 @@ __pattern_remove_if(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy
...
+// __pattern_reverse_copy
+//---------------------------------------------------------------------------------------------------------------------
+
+template <typename _Tag, typename _ExecutionPolicy, typename _InRange, typename _OutRange>
+void
+__pattern_reverse_copy(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r)
+{
+    static_assert(__is_parallel_tag_v<_Tag> || typename _Tag::__is_vector{});
+
+    assert(std::ranges::size(__in_r) <= std::ranges::size(__out_r)); // for debug purposes only
```
SergeyKopienko: Let's remove this line?

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -696,6 +696,49 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec,
...
+__pattern_reverse(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R&& __r)
+{
+    const auto __n = std::ranges::size(__r);
+    if (__n <= 1)
+        return;
+
+    oneapi::dpl::__par_backend_hetero::__parallel_for(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
+                                                      unseq_backend::__reverse_functor<decltype(__n)>{__n}, __n / 2,
+                                                      oneapi::dpl::__ranges::views::all(std::forward<_R>(__r)))
+        .__checked_deferrable_wait();
```
SergeyKopienko, line 714: Here we calls `__checked_deferrable_wait`...

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -696,6 +696,49 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec,
...
+    const auto __n = std::ranges::size(__in_r);
+    if (__n == 0)
+        return;
+
+    using _DiffType = oneapi::dpl::__internal::__difference_t<_InRange>;
+    oneapi::dpl::__par_backend_hetero::__parallel_for(
+        _BackendTag{}, std::forward<_ExecutionPolicy>(__exec), unseq_backend::__reverse_copy<_DiffType>{__n}, __n,
+        oneapi::dpl::__ranges::views::all_read(std::forward<_InRange>(__in_r)),
+        oneapi::dpl::__ranges::views::all_write(std::forward<_OutRange>(__out_r)))
+        .get(); // is a blocking call
```
SergeyKopienko: ...but in this case - we call `get()`.
Can you explain - why?

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
dmitriy-sobolev: Removed

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: Good catch. It is a mistake. We do not use a return value from `__parallel_for`, hence it is enough to use `__checked_deferrable_wait`. It's done.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -696,6 +696,48 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec,
     return {std::ranges::begin(__r) + __idx, std::ranges::begin(__r) + __n};
 }
 
+//------------------------------------------------------------------------
+// reverse
+//------------------------------------------------------------------------
+
+template <typename _BackendTag, typename _ExecutionPolicy, typename _R>
+void
+__pattern_reverse(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R&& __r)
```
mmichel11: `__tag` is unused, so we may want to remove it to prevent a warning.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -696,6 +696,48 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec,
...
+        .__checked_deferrable_wait();
+}
+
+//------------------------------------------------------------------------
+// reverse_copy
+//------------------------------------------------------------------------
+
+template <typename _BackendTag, typename _ExecutionPolicy, typename _InRange, typename _OutRange>
+void
+__pattern_reverse_copy(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r,
```
mmichel11: Same comment about `__tag`.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: Done

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -696,6 +696,47 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec,
...
+void
+__pattern_reverse(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _R&& __r)
+{
+    const auto __n = std::ranges::size(__r);
+    if (__n <= 1)
+        return;
+
+    oneapi::dpl::__par_backend_hetero::__parallel_for(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
+                                                      unseq_backend::__reverse_functor<decltype(__n)>{__n}, __n / 2,
+                                                      oneapi::dpl::__ranges::views::all(std::forward<_R>(__r)))
```
danhoeflinger: do we need this `views::all` call?
My understanding is that this is just to support `sycl::buffer` as a range.  I don't believe we are attempting to accept a `sycl::buffer` as input for the (non-experimental) parallel range API.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -696,6 +696,57 @@ __pattern_remove_if(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy
...
+                                                    __last_in, __first_out);
+}
+
+template <typename _ExecutionPolicy, typename _InRange, typename _OutRange>
+void
+__pattern_reverse_copy(__serial_tag</*IsVector*/ std::false_type>, _ExecutionPolicy&&, _InRange&& __in_r,
+                       _OutRange&& __out_r)
+{
+    std::ranges::reverse_copy(std::forward<_InRange>(__in_r), std::ranges::begin(__out_r));
+}
```
akukanov, line 744: This implementation does not support insufficient-size output ranges. We need to do it even for `__serial_tag` because the semantics is the same for all kinds of policies.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -899,6 +899,69 @@ struct __replace_fn
...
+        using _Size = std::common_type_t<std::ranges::range_size_t<_InRange>, std::ranges::range_size_t<_OutRange>>;
+        const _Size __sz = std::ranges::min((_Size)std::ranges::size(__in_r), (_Size)std::ranges::size(__out_r));
+
+        auto __last_in = std::ranges::begin(__in_r) + std::ranges::size(__in_r);
+        auto __stop_in = std::ranges::begin(__in_r) + (std::ranges::size(__in_r) - __sz);
+        auto __stop_out = std::ranges::begin(__out_r) + __sz;
+
+        oneapi::dpl::__internal::__ranges::__pattern_reverse_copy(
+            __dispatch_tag, std::forward<_ExecutionPolicy>(__exec), std::ranges::take_view(__in_r, __sz),
+            std::ranges::take_view(__out_r, __sz));
```
akukanov: This implementation is incorrect in case if the output range is smaller than the input one, because it copies first elements of the input. A correct implementation should copy the elements "starting from the end" until the output is full,

test/parallel_api/ranges/std_ranges_reverse_copy.pass.cpp
```diff
@@ -0,0 +1,46 @@
...
+#if _ENABLE_STD_RANGES_TESTING
+    using namespace test_std_ranges;
+    namespace dpl_ranges = oneapi::dpl::ranges;
+
+    auto reverse_copy_checker = [](std::ranges::random_access_range auto&& r_in,
+                                   std::ranges::random_access_range auto&& r_out)
+    {
+        const auto size = std::ranges::min(std::ranges::size(r_in), std::ranges::size(r_out));
+
+        std::ranges::reverse_copy(std::ranges::take_view(r_in, size), std::ranges::take_view(r_out, size).begin());
```
danhoeflinger: With @akukanov comment above, I was confused at how the tests pass since we do tests this case, and it is because we encode the same semantics here in the tests.  
I believe it should be `drop_view` dropping `size(r_in) - size` instead.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -899,6 +899,69 @@ struct __replace_fn
...
+                                  std::ranges::borrowed_iterator_t<_OutRange>>
+    operator()(_ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r) const
+    {
+        const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
+
+        using _Size = std::common_type_t<std::ranges::range_size_t<_InRange>, std::ranges::range_size_t<_OutRange>>;
+        const _Size __sz = std::ranges::min((_Size)std::ranges::size(__in_r), (_Size)std::ranges::size(__out_r));
+
+        auto __last_in = std::ranges::begin(__in_r) + std::ranges::size(__in_r);
+        auto __stop_in = std::ranges::begin(__in_r) + (std::ranges::size(__in_r) - __sz);
```
akukanov: `std::ranges::size(__in_r)` is computed three times in three lines. It is not a heavy computation, but nevertheless - why not store it locally?

Also/alternatively, `__stop_in` can be computed as `__last_in - __sz`.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
SergeyKopienko: Fixed.

test/parallel_api/ranges/std_ranges_reverse_copy.pass.cpp
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: You are right, it was excessive here. I've removed it.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
dmitriy-sobolev: The extra calculations have been removed.  As for __stop_in, it is now calculated as `__first_in + __in_r_skipped`, which is also quite short.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -908,6 +908,83 @@ struct __replace_fn
...
+                                  std::ranges::borrowed_iterator_t<_OutRange>>
+    operator()(_ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r) const
+    {
+        const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
+
+        const auto __in_r_size = std::ranges::size(__in_r);
+        const auto __out_r_size = std::ranges::size(__out_r);
+
+        using _Size = std::common_type_t<std::ranges::range_size_t<_InRange>, std::ranges::range_size_t<_OutRange>>;
+        const _Size __sz = std::ranges::min((_Size)__in_r_size, (_Size)__out_r_size);
```
rarutyun: if you have `common_size` it looks weird to me that you first create variable, then cast them to this `common_type`
```suggestion
       using _common_size_type = std::common_type_t<std::ranges::range_size_t<_InRange>, std::ranges::range_size_t<_OutRange>>;
        const _common_size_type __in_r_size = std::ranges::size(__in_r);
        const _common_size_type __out_r_size = std::ranges::size(__out_r);
        const auto__min_size = std::ranges::min(__in_r_size, __out_r_size);
```

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -696,6 +696,46 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec,
...
+//------------------------------------------------------------------------
+// reverse_copy
+//------------------------------------------------------------------------
+
+template <typename _BackendTag, typename _ExecutionPolicy, typename _InRange, typename _OutRange>
+void
+__pattern_reverse_copy(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r)
+{
+    const auto __n = std::ranges::size(__in_r);
+    if (__n == 0)
```
rarutyun, line 728: I think, this is incorrect. What if the size of `_OutRange` is also `0`? I mean it might be handled by `__parallel_for` call, but to me it's super unobvious.

I wonder if we should move all the checks to the API level, not to the pattern level. No strong opinion on that; I just wanted to express the idea

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
SergeyKopienko: Applied with the small changes

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: > I think, this is incorrect. What if the size of _OutRange is also 0? I mean it might be handled by __parallel_for call, but to me it's super unobvious.

The API level (`__reverse_copy_fn`) creates views over the original ranges, trimmed to the common minimal size. If one range is empty, the other view is also empty.

Yes, it is not obvious. We can add asserts or comments - my preferable way. 

> I wonder if we should move all the checks to the API level, not to the pattern level. No strong opinion on that; I just wanted to express the idea

Moving that check to the API level comes with a drawback. For the host path, we simply transfer the processing to the iterator based patterns, which do that check. We do not want to repeat it. We can trace all the checks and move them to the upper level, but it is a lot of work.

dmitriy-sobolev: I've added an assert with a comment.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -908,6 +908,82 @@ struct __replace_fn
...
+        auto __last_in = __first_in + __in_r_size;
+        auto __stop_in = __first_in + __in_r_skipped;
+        auto __stop_out = __first_out + __sz;
+
+        // subrange is used instead of take_view/drop_view because the latter throw exceptions in libstdc++10,
+        // which prevents their use in SYCL kernels.
+        // The internal take_view_simple/drop_view_simple cannot be used either,
+        // as they are not fully implemented as views, which is required for the host backends.
+        // TODO: Complete take_view_simple/drop_view_simple to satisfy the view requirements
+        // and replace subrange/take_view/drop_view with them
```
akukanov: Why do you prefer take/drop views to `subrange`? These views internally need to compute their begin/end positions, which is already done above for return values. I think the use of `subrange` is perfectly fine here and even preferred to take/drop.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -560,8 +560,9 @@ __pattern_copy(__serial_tag</*IsVector*/std::false_type>, _ExecutionPolicy&&, _I
 
 template <typename _Tag, typename _ExecutionPolicy, typename _InRange, typename _OutRange, typename _Pred,
           typename _Proj>
-auto
-__pattern_copy_if_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r, _Pred __pred, _Proj __proj)
+std::ranges::copy_if_result<std::ranges::borrowed_iterator_t<_InRange>, std::ranges::borrowed_iterator_t<_OutRange>>
+__pattern_copy_if_ranges(_Tag __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r, _Pred __pred,
+                         _Proj __proj)
```
akukanov: Would it be better to move unrelated modifications to a separate patch?

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -908,6 +908,82 @@ struct __replace_fn
...
+
+        auto __first_in = std::ranges::begin(__in_r);
+        auto __first_out = std::ranges::begin(__out_r);
+
+        auto __last_in = __first_in + __in_r_size;
+        auto __stop_in = __first_in + __in_r_skipped;
+        auto __stop_out = __first_out + __sz;
+
+        // subrange is used instead of take_view/drop_view because the latter throw exceptions in libstdc++10,
+        // which prevents their use in SYCL kernels.
```
akukanov: Is it because of a bug in libstdc++, or some broader limitation/problem?

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
dmitriy-sobolev: > Why do you prefer take/drop views to subrange?

take_view/drop_view are more specialized for these cases, which intuitively seemed to me more suitable here.

> I think the use of subrange is perfectly fine here and even preferred to take/drop.

I agree after the clarification. I've removed the comment in the code.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
dmitriy-sobolev: I think it is rather a broader limitation. If we cannot use take_view/drop_view here, a user will get the same issues when passing them through the public API. Let me double check it.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
dmitriy-sobolev: Sergey has already prepared: #2416. I've rolled back these changes here.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
akukanov: I mean, if it is only a bug in a reasonably old version of `libstdc++` and not a problem in more recent versions, we can mostly ignore it (aside from not using these views ourselves). If however these views are expected to throw exceptions and therefore cannot be used when algorithms are invoked with device policies, that is a significant limitation we need to document.

dmitriy-sobolev: The problem is present in libstdc++11.1 and older. I am not sure if it was a bug - I see nothing in the standard prohibiting exceptions in `std::ranges::advance`. It was probably a debug related code section.  It was removed in: https://github.com/gcc-mirror/gcc/commit/f2b76257e9a487fd2c7265094f7b4d1bd46f5f03.

Given that, I do not think that it is a significant limitation. However, we have a note:

> * ``std::ranges::drop_view`` from libstdc++ version 10 may throw exceptions.
  This can lead to a "SYCL kernel cannot use exceptions" compilation error
  when it is used to pass data to a range-based algorithm with a device policy.

akukanov: Thanks for exploring this. So we may want to update the note, and there is nothing to do besides that.

test/parallel_api/ranges/std_ranges_reverse_copy.pass.cpp
```diff
@@ -0,0 +1,52 @@
...
+            std::ranges::take_view(r_out, sz).begin());
+
+        using ret_type = std::ranges::in_in_out_result<std::ranges::borrowed_iterator_t<decltype(r_in)>,
+                                                       std::ranges::borrowed_iterator_t<decltype(r_in)>,
+                                                       std::ranges::borrowed_iterator_t<decltype(r_out)>>;
+
+        auto last_in = std::ranges::begin(r_in) + in_r_size;
+        auto stop_in = std::ranges::begin(r_in) + in_r_skipped;
+        auto stop_out = std::ranges::begin(r_out) + sz;
+        return ret_type{last_in, stop_in, stop_out};
```
akukanov: Could we instead reuse iterators returned by `std::ranges::reverse_copy`, and only compute what it does not return? I believe it would be a better check than fully repeating the implementation logic.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -908,6 +908,76 @@ struct __replace_fn
...
+        const _Size __out_r_size = std::ranges::size(__out_r);
+        const _Size __sz = std::ranges::min(__in_r_size, __out_r_size);
+
+        const auto __in_r_skipped = __in_r_size - __sz;
+
+        auto __first_in = std::ranges::begin(__in_r);
+        auto __first_out = std::ranges::begin(__out_r);
+
+        auto __last_in = __first_in + __in_r_size;
+        auto __stop_in = __first_in + __in_r_skipped;
```
akukanov:
```suggestion
        auto __first_in = std::ranges::begin(__in_r);
        auto __first_out = std::ranges::begin(__out_r);

        auto __last_in = __first_in + __in_r_size;
        auto __stop_in = __first_in + (__in_r_size - __sz); // the last element that fits into the output
```
Removed `__in_r_skipped`

test/parallel_api/ranges/std_ranges_reverse_copy.pass.cpp
dmitriy-sobolev: Done, see [dcf0a7a](https://github.com/uxlfoundation/oneDPL/pull/2366/commits/dcf0a7a0986d540066841154d9c027a240d41e47).

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
dmitriy-sobolev: Done, see [dcf0a7a](https://github.com/uxlfoundation/oneDPL/pull/2366/commits/dcf0a7a0986d540066841154d9c027a240d41e47).

test/parallel_api/ranges/std_ranges_reverse.pass.cpp
```diff
@@ -0,0 +1,31 @@
...
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#include "std_ranges_test.h"
+
+std::int32_t
```
rarutyun:
```suggestion
int
```

test/parallel_api/ranges/std_ranges_reverse_copy.pass.cpp
```diff
@@ -0,0 +1,45 @@
...
+// notice:
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+//
+//===----------------------------------------------------------------------===//
+
+#include "std_ranges_test.h"
+
+std::int32_t
```
rarutyun:
```suggestion
int
```

rarutyun (approved):

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
dmitriy-sobolev: An additional note for the future reference and better understanding.

`oneapi::dpl::__ranges::views::all` and alike add some methods on the top of minimal requirements of a range (`begin()`, `end()`, and `size()` for a sized range). One of such methods is `empty()`, which is occasionally used.

I've checked that the new reverse and reverse_copy patterns do not rely on anything using these methods.
Thinking further, `oneapi::dpl::__ranges::views::all` should be removed everywhere for the regular ranges together with the use of these "additional" methods.

Issue #2342, opened 2025-07-13
Iterator increments should be protected against comma operator


jwakely:
There are several places in `pstl/algorithm_impl.h`, `pstl/memory_impl.h`, and `pstl/numeric_impl.h` that break if used with an iterator that has:

```c++

struct EvilIterator
{
  EvilIterator& operator++();
  // etc.
  template<class T> void operator,(const T&) = delete;
};
```

The overloaded `operator,` will be used here if `__first1` is an `EvilIterator` and so will call a deleted function and fail to compile:

https://github.com/uxlfoundation/oneDPL/blob/ef406d231043c7b1df49ebf781ee5cde3624138e/include/oneapi/dpl/pstl/algorithm_impl.h#L276

The fix is to always cast to void when combining multiple expressions with commas:

```diff
-    for (; __first1 != __last1; ++__first1, ++__first2)
+    for (; __first1 != __last1; ++__first1, (void) ++__first2)
```

When a comma separates two expressions and at least one of them has type `void`, it cannot match an overloaded `operator,` and always means the built-in comma.

Some places have three expressions, e.g.

https://github.com/uxlfoundation/oneDPL/blob/ef406d231043c7b1df49ebf781ee5cde3624138e/include/oneapi/dpl/pstl/algorithm_impl.h#L295

Casting the second one to void is sufficient to prevent any problems here:

```diff
-   for (; __n > 0; --__n, ++__first1, ++__first2) 
+    for (; __n > 0; --__n, (void) ++__first1, ++__first2) 
```

danhoeflinger:
Thanks for reporting this.  I can take a look at resolving this in oneDPL.

danhoeflinger:
Following up, this PR fixes this across the repository and adds a compile-only test to prevent regressions in this dimension.
#2369

PR #2369, opened 2025-08-04
Resolve comma deleting iterator

danhoeflinger:
As mentioned in #2342,  an iterator which deletes the comma operator (or otherwise overrides it) can encounter issues in some places in oneDPL, when multiple expressions are combined into a single one via a comma, (usually as a loop increment). This can end up in a build error or unintended consequences.  
```
struct EvilIterator
{
  EvilIterator& operator++();
  // etc.
  template<class T> void operator,(const T&) = delete;
};
```
This PR removes these issues by casting one of the expressions to `(void)`, which avoids the overloaded comma operator, or in some cases, just separating into two lines.


include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -292,7 +292,7 @@ _ForwardIterator2
 __brick_walk2_n(_ForwardIterator1 __first1, _Size __n, _ForwardIterator2 __first2, _Function __f,
                 /*vector=*/::std::false_type) noexcept
 {
-    for (; __n > 0; --__n, ++__first1, ++__first2)
+    for (; __n > 0; --__n, (void)++__first1, ++__first2)
```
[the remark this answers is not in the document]
danhoeflinger: I think we probably want to favor minimal "ugliness" in the code to resolve this somewhat obscure issue, rather than always casting joined comma expressions to `void`, or even ones involving iterators.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -479,7 +479,7 @@ _ForwardIterator3
 __brick_walk3(_ForwardIterator1 __first1, _ForwardIterator1 __last1, _ForwardIterator2 __first2,
               _ForwardIterator3 __first3, _Function __f, /*vector=*/::std::false_type) noexcept
 {
-    for (; __first1 != __last1; ++__first1, ++__first2, ++__first3)
+    for (; __first1 != __last1; ++__first1, (void)++__first2, ++__first3)
```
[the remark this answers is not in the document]
danhoeflinger: same response as above.

timmiesmith, line 482: Here and at line 295 you cast the first iterator operator in the expression.  It most other places here and in other files the second iterator operator is cast.  Is there a reason one is chosen instead of the other?

danhoeflinger: To minimize the number of casts.  If we cast the middle of 3 expressions, it covers both potential commas causing issues.
https://godbolt.org/z/W3vWEWe6P
If we want to, we can add more (void) to prevent future issues, but my goal right now is use the minimal casts to accomplish the issue to not clutter the code.

danhoeflinger:
> Do you think it's worth it to add some simple test for this case to prevent regressions in the future? For example, compile and run a simple case for oneDPL algorithms covering all of our patterns with this "EvilIterator".

Yeah, I was thinking about this but did not initially have a good thought about how to do this simply with large coverage and small impact on test suite time.  While this is a real bug opportunity, its also a pretty exotic rare case, so I'm not sure how much effort it is worth to address. 

My best idea thus far is to expand our "compile only" tests, and make an iterator adaptor which deletes the comma operator, and otherwise just passes operations through.  Then we can add it in one place to our helper code, rather than in all tests. I can explore how feasible it is and add it if its not too bad.

danhoeflinger:
@mmichel11 OK, after a bit of struggling trying to understand where to fit it in, I believe I have a system-wide compile only test now which detects the errors.  I've push *only* the test changes, without the fixes for the code in an effort to show that we do in fact detect the problems with this test addition.

Once the CI completes (very red), I will push the fixes, and we will see if it resolves everything across the board, or if there remain any nagging issues.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3395,14 +3396,10 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
...
-                __internal::__pattern_walk2_brick(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first1, __last1,
-                                                  __result, __copy_range);
-            },
-            [=] {
-                __internal::__pattern_walk2_brick(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first2, __last2,
-                                                  __result + __n1, __copy_range);
+                __internal::__pattern_walk2_brick(__tag, __exec, __first2, __last2, __result + __n1, __copy_range);
             });
         return __result + __n1 + __n2;
     }
```
danhoeflinger: This is a fix for a use-after-move issue (forwarding 3 times, and some in helpers before the call), and also an issue with forwarding a capture-by-value argument to a lambda.  Some changes in the test suite exercised this bug.

There are a few similar cases below as well.

include/oneapi/dpl/pstl/experimental/internal/for_loop_impl.h
```diff
@@ -502,14 +502,14 @@ struct __use_par_vec_helper
...
 auto
 __use_vectorization(_ExecutionPolicy&& __exec)
 {
     return __use_par_vec_helper<_Ip>::__use_vector(::std::forward<_ExecutionPolicy>(__exec));
 }
 
-template <typename _ExecutionPolicy, typename _Ip>
+template <typename _Ip, typename _ExecutionPolicy>
 auto
 __use_parallelization(_ExecutionPolicy&& __exec)
```
danhoeflinger: This was an issue where because of the ordering of template arguments, we were using non-forwarded policy arguments, with a specified type for the arguments which may not match the incoming value category of policy.
Reordering the template arguments allows us to properly use forwarding / universal references for the policy again in these helpers.

Again, this was exercised by some changes to the test suite code.

test/support/utils_invoke.h
```diff
@@ -418,8 +567,9 @@ struct invoke_on_all_hetero_policies
             // performs some checks that fail. As a workaround, define for functors which have this issue
             // __functor_type(see kernel_type definition) type field which doesn't have any pointers in it's name.
             iterator_invoker<std::random_access_iterator_tag, /*IsReverse*/ std::false_type>()(
-                my_policy, op, std::forward<Args>(rest)...);
+                my_policy, op, rest...);
```
danhoeflinger: This was a use-after-move bugfix for whenever `TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY` was true already.  We add our compile check, but this is also a bugfix of an existing bug.

danhoeflinger (commented):
Pointing out and explaining a few changes which are tangential to the main purpose of the PR, but still relevant and changes for for the better.  I'm running a local full build of the dpcpp tests, I think this is good to go pending success of that build.

danhoeflinger:
After some iteration, I believe the testing changes are in a good place, and ready for review.

include/oneapi/dpl/pstl/experimental/internal/for_loop_impl.h
SergeyKopienko: What about to extract this change into separate PR ?

test/support/utils_invoke.h
SergeyKopienko: @danhoeflinger, some comments:
1) this is not an error, because later under `TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY` macro we just check compilation, so only types are interested for us.
2) we should `forward` test params into tests for their run-time execution.
3) at least we can extract this question into separate PR

include/oneapi/dpl/pstl/algorithm_impl.h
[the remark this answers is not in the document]
SergeyKopienko: As far as I understood the `(void)` before `++` is the main idea of this fix.
What about to wrap this construction into some (inline) template function?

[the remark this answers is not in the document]
danhoeflinger: I don't understand the suggestion here.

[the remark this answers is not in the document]
SergeyKopienko: Something like
```C++
template <typename _Iterator>
void
__pre_inc(_Iterator& __it)
{
    ++__it;
}
```

Then we can rewrite
```C++
for (; __n > 0; --__n, (void)++__first1, (void)++__first2)
```
like
```C++
for (; __n > 0; --__n, __pre_inc(__first1), __pre_inc(__first2))
```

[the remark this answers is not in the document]
SergeyKopienko: moreover, probably we may thinks about to pack subset of params into `__pre_inc` and have:
```C++
for (; __n > 0; --__n, __pre_inc(__first1, __first1))
```

include/oneapi/dpl/pstl/experimental/internal/for_loop_impl.h
danhoeflinger: https://github.com/uxlfoundation/oneDPL/pull/2372

include/oneapi/dpl/pstl/algorithm_impl.h
danhoeflinger: Separated to new PR https://github.com/uxlfoundation/oneDPL/pull/2373

test/support/utils_invoke.h
danhoeflinger: reverted.  Its a bit odd, but I see your point.  I still think we are playing with fire a bit by passing these around after moving them, even if we don't actually run the payload, but we can deal with it in another PR / issue, or leave it as is.

include/oneapi/dpl/pstl/algorithm_impl.h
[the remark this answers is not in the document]
danhoeflinger: Is the suggestion to *always* use `__pre_inc(it)` instead of `++it` across all of oneDPL? or just when we have multiple expressions joined by a comma?

I could see possibly doing this for expressions joined by a comma...

I suppose it would allow us to have a single location where we could "explain" the issue we are avoid here.  However, I worry that it obfuscates the code, in that you need to go to a second location to find out what this does.

The nice thing about the `(void`) casts is that all the code remains inline, and its clear what is happening (even if it is not immediately clear why the `(void)` casts are there).

danhoeflinger:
Tests will fail until #2373 and #2372 are merged.

include/oneapi/dpl/pstl/algorithm_impl.h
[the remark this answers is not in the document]
SergeyKopienko: >>>Is the suggestion to always use __pre_inc(it) instead of ++it across all of oneDPL? or just when we have multiple expressions joined by a comma?
- I think this two options make sense.

My opinion that multiple code like `(void)++__first1, (void)++__first2` in any places at first glance will nor bring the understanding why we doing this. As result after some not long time we will forget this requirement.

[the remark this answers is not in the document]
akukanov: I am of the opinion that we should leave the code ugly but clear, not obfuscated "for convenience". If someone does not understand why this strange pattern is used, they could ask.
Alternatively, move all but one increments into the body of the loop.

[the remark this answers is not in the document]
SergeyKopienko: Of course, this option is also on the table.

danhoeflinger:
> @danhoeflinger as alternative approach - may we thinks about wrap source iterators from users and disable the `operator,` from them?

It seems overly complex and also a bit dangerous to effectively disable parts of user defined code.  I'm not instantly coming up with a situation where a user could provide code through a binary operator or function body which would depend on a defined comma operator, but we would have to "wrap" and "unwrap" everything so that types match on the way out.  We would also have to adjust our "input processing" code to prepare sycl kernels.  I strongly prefer the simple solution which just uses void casts.

test/support/compile_only_checks.h
```diff
@@ -0,0 +1,307 @@
...
+        return iter_ >= other.iter_;
+    }
+
+    // Deleted comma operator - this is the key feature
+    template<typename T>
+    void operator,(const T&) = delete;
+    template<typename T>
+    void operator,(T&) = delete;
+    template<typename T>
+    void operator,(T&&) = delete;
```
mmichel11: Is it sufficient to only delete the perfect forwarding version of the operator overload?

danhoeflinger: Yes, I believe you are correct.  Deleted the unnecessary ones.

PR #2371, opened 2025-08-05
Fix compile error in `inclusive_scan` and `exclusive_scan` over reverse iterators with different source iterator types

SergeyKopienko:
This PR fixes a compile error in oneDPL's `inclusive_scan` and `exclusive_scan` operations when used with reverse iterators that have different source iterator types (issue #2296). The solution enhances iterator equality comparison by implementing sophisticated logic to unwrap base iterators (from `std::reverse_iterator` only) and compare them instead of comparing adapter iterators directly.

- Enhanced iterator equality comparison with C++20 concepts support and fallback implementation for complex iterator types
- Added comprehensive test coverage for reverse iterator scenarios with type conversion in `exclusive_scan`
- Removed problematic assertions in USM data transfer utilities that were incompatible with iterator arithmetic


include/oneapi/dpl/pstl/utils.h
```diff
@@ -879,6 +846,100 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
+    struct __iterator_value_type<_Iterator, typename std::iterator_traits<_Iterator>::value_type>
+    {
+        using __type = typename std::iterator_traits<_Iterator>::value_type;
+    };
+
+    // Checks that _Iterator1 and _Iterator2 are iterators or pointers,
+    // their iterated values are the same type and the operation of
+    // equality comparison is implemented for them.
+    template <typename _IteratorParam1, typename _IteratorParam2>
+    using __is_equality_comparable =
```
SergeyKopienko, line 876: I propose to save this formatting because the variant after clang-format looks not so readable like that:
```C++
    template <typename _IteratorParam1, typename _IteratorParam2>
    using __is_equality_comparable =
        std::conjunction<__is_iterator_type<_IteratorParam1>, __is_iterator_type<_IteratorParam2>,
                         std::is_same<typename __iterator_value_type<_IteratorParam1>::__type,
                                      typename __iterator_value_type<_IteratorParam2>::__type>,
                         std::is_invocable<std::equal_to<>, _IteratorParam1, _IteratorParam2>>;
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -879,6 +846,100 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
+        std::conjunction<
+            __is_iterator_type<_IteratorParam1>,
+            __is_iterator_type<_IteratorParam2>,
+            std::is_same<
+                typename __iterator_value_type<_IteratorParam1>::__type,
+                typename __iterator_value_type<_IteratorParam2>::__type
+            >,
+            std::is_invocable<std::equal_to<>, _IteratorParam1, _IteratorParam2>
+        >;
+
```
SergeyKopienko: I propose to save this formatting because the variant after clang-format looks not so readable like that:
```C++
    template <typename _IteratorParam1, typename _IteratorParam2>
    using __is_equality_comparable =
        std::conjunction<__is_iterator_type<_IteratorParam1>, __is_iterator_type<_IteratorParam2>,
                         std::is_same<typename __iterator_value_type<_IteratorParam1>::__type,
                                      typename __iterator_value_type<_IteratorParam2>::__type>,
                         std::is_invocable<std::equal_to<>, _IteratorParam1, _IteratorParam2>>;
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -879,6 +846,81 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
+struct __iterator_value_type
+{
+    template <typename TIterator>
+    struct unknown_value_type { };
+
+    using __type = unknown_value_type<_Iterator>;
+};
+
+template <typename _Iterator>
+struct __iterator_value_type<_Iterator, std::is_same<_Iterator, std::nullptr_t>>
```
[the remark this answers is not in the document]
SergeyKopienko: I believe it's not correct comment: if we rewrite this code like
```C++
    template <typename _Iterator>
    struct __iterator_value_type<_Iterator, std::enable_if_t<std::is_same_v<_Iterator, std::nullptr_t>>>
    {
        static_assert(false);
        using __type = std::nullptr_t;
    };
```
and compile the test `sycl_iterator_operations.pass.cpp` we will see the errors
```
static_assert(false);
              ^~~~~
```
and the stack if this instantiation.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -879,6 +846,81 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
 static constexpr bool __is_iterator_type_v = __is_iterator_type<_T>::value;
 
+template <typename _Iterator, typename = void>
+constexpr _Iterator
+__get_base_iterator_or_the_iterator_itself(_Iterator __it)
+{
+    return __it;
+}
+
+template <typename _Iterator, std::enable_if_t<std::is_invocable_v<decltype(&_Iterator::base), _Iterator&>, int>>
```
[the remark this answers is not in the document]
SergeyKopienko: I believe this comment is not correct.
If we rewrite the second `__get_unwound_iterator` function like
```C++
    // If _Iterator::base() is defined, returns the result of recursive call of __get_unwound_iterator for _Iterator::base()
    template <typename _Iterator, std::enable_if_t<std::is_invocable_v<decltype(&_Iterator::base), _Iterator&>, int> = 0>
    constexpr auto
    __get_unwound_iterator(_Iterator __it) const
    {
        return __get_unwound_iterator(__it.base());
    }
```
then we will have compile error in the `exclusive_scan.pass.cpp` test:
```
  In file included from test\parallel_api\numeric\numeric.ops\exclusive_scan.pass.cpp:18:
  In file included from include\oneapi\dpl\execution:28:
  In file included from include\oneapi\dpl\pstl\execution_defs.h:161:
  In file included from include\oneapi\dpl\pstl\hetero\dpcpp\execution_sycl_defs.h:21:
  In file included from include\oneapi\dpl\pstl\iterator_defs.h:21:
include\oneapi\dpl\pstl\utils.h(916,27): error : call to member function '__get_unwound_iterator' is ambiguous
    916 |         auto __it1_base = __get_unwound_iterator(__it1);
        |                           ^~~~~~~~~~~~~~~~~~~~~~
  include\oneapi\dpl\pstl\utils.h(940,12): note: in instantiation of function template specialization 'oneapi::dpl::__internal::__iterators_possibly_equal_impl::operator()<std::reverse_iterator<bool *>, std::reverse_iterator<int *>>' requested here
    940 |     return __iterators_possibly_equal_impl()(__it1, __it2);
        |            ^
  include\oneapi\dpl\pstl\hetero\numeric_impl_hetero.h(141,74): note: in instantiation of function template specialization 'oneapi::dpl::__internal::__iterators_possibly_equal<std::reverse_iterator<bool *>, std::reverse_iterator<int *>>' requested here
    141 |     const bool __is_scan_inplace_exclusive = __n > 1 && !_Inclusive{} && __iterators_possibly_equal(__first, __result);
        |                                                                          ^
  include\oneapi\dpl\pstl\hetero\numeric_impl_hetero.h(196,12): note: in instantiation of function template specialization 'oneapi::dpl::__internal::__pattern_transform_scan_base<oneapi::dpl::__internal::__device_backend_tag, const oneapi::dpl::execution::device_policy<TestUtils::unique_kernel_name<TestUtils::TestPolicyName, 0>> &, std::reverse_iterator<bool *>, std::reverse_iterator<int *>, oneapi::dpl::identity, oneapi::dpl::unseq_backend::__init_value<int>, std::plus<int>, std::integral_constant<bool, false>>' requested here
    196 |     return __pattern_transform_scan_base(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first, __last, __result,
        |            ^
  include\oneapi\dpl\pstl\glue_numeric_impl.h(222,37): note: in instantiation of function template specialization 'oneapi::dpl::__internal::__pattern_transform_scan<oneapi::dpl::__internal::__device_backend_tag, const oneapi::dpl::execution::device_policy<TestUtils::unique_kernel_name<TestUtils::TestPolicyName, 0>> &, std::reverse_iterator<bool *>, std::reverse_iterator<int *>, oneapi::dpl::identity, int, std::plus<int>, std::integral_constant<bool, false>>' requested here
    222 |     return oneapi::dpl::__internal::__pattern_transform_scan(__dispatch_tag, ::std::forward<_ExecutionPolicy>(__exec),
        |                                     ^
  include\oneapi\dpl\pstl\glue_numeric_impl.h(112,12): note: in instantiation of function template specialization 'oneapi::dpl::transform_exclusive_scan<const oneapi::dpl::execution::device_policy<TestUtils::unique_kernel_name<TestUtils::TestPolicyName, 0>> &, std::reverse_iterator<bool *>, std::reverse_iterator<int *>, int, std::plus<int>, oneapi::dpl::identity>' requested here
    112 |     return transform_exclusive_scan(std::forward<_ExecutionPolicy>(__exec), __first, __last, __result, __init,
        |            ^
  test\parallel_api\numeric\numeric.ops\exclusive_scan.pass.cpp(106,18): note: in instantiation of function template specialization 'oneapi::dpl::exclusive_scan<const oneapi::dpl::execution::device_policy<TestUtils::unique_kernel_name<TestUtils::TestPolicyName, 0>> &, std::reverse_iterator<bool *>, std::reverse_iterator<int *>, int>' requested here
    106 |     oneapi::dpl::exclusive_scan(
        |                  ^
  test\parallel_api\numeric\numeric.ops\exclusive_scan.pass.cpp(137,5): note: in instantiation of function template specialization 'test_diff_iterators<const oneapi::dpl::execution::device_policy<TestUtils::unique_kernel_name<TestUtils::TestPolicyName, 0>> &>' requested here
    137 |     test_diff_iterators(CLONE_TEST_POLICY(exec));
        |     ^
  test\parallel_api\numeric\numeric.ops\exclusive_scan.pass.cpp(147,5): note: in instantiation of function template specialization 'test_impl<oneapi::dpl::execution::device_policy<TestUtils::unique_kernel_name<TestUtils::TestPolicyName, 0>> &>' requested here
    147 |     test_impl(policy);
        |     ^
  include\oneapi\dpl\pstl\utils.h(890,5): note: candidate function [with _Iterator = std::reverse_iterator<bool *>, $1 = void]
    890 |     __get_unwound_iterator(_Iterator __it) const
        |     ^
  include\oneapi\dpl\pstl\utils.h(898,5): note: candidate function [with _Iterator = std::reverse_iterator<bool *>, $1 = 0]
    898 |     __get_unwound_iterator(_Iterator __it) const
        |     ^
include\oneapi\dpl\pstl\utils.h(917,27): error : call to member function '__get_unwound_iterator' is ambiguous
    917 |         auto __it2_base = __get_unwound_iterator(__it2);
        |                           ^~~~~~~~~~~~~~~~~~~~~~
  include\oneapi\dpl\pstl\utils.h(890,5): note: candidate function [with _Iterator = std::reverse_iterator<int *>, $1 = void]
    890 |     __get_unwound_iterator(_Iterator __it) const
        |     ^
  include\oneapi\dpl\pstl\utils.h(898,5): note: candidate function [with _Iterator = std::reverse_iterator<int *>, $1 = 0]
  
    898 |     __get_unwound_iterator(_Iterator __it) const
  
        |     ^
  
  2 errors generated.
```

test/parallel_api/numeric/numeric.ops/exclusive_scan.pass.cpp
```diff
@@ -73,12 +73,68 @@ test_with_usm(Policy&& exec)
     }
 }
 
+template <typename Policy>
+void
+test_diff_iterators(Policy&& exec)
+{
+    constexpr std::size_t N = 6;
+
+    sycl::queue q = exec.queue();
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/utils.h
```diff
@@ -879,6 +846,99 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
+    // If _Iterator::base() is not defined, returns the iterator itself.
+    template <typename _Iterator, typename = void>
+    constexpr _Iterator
+    __unwind_iterator(_Iterator __it) const
+    {
+        return __it;
+    }
+
+    // If _Iterator::base() is defined, returns the result of recursive call of __get_unwound_iterator for _Iterator::base()
+    template <typename _Iterator, std::enable_if_t<std::is_invocable_v<decltype(&_Iterator::base), _Iterator&>, int>>
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/utils.h
```diff
@@ -879,6 +846,99 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
+    constexpr auto
+    __unwind_iterator(_Iterator __it) const
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

test/parallel_api/numeric/numeric.ops/exclusive_scan.pass.cpp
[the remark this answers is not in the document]
SergeyKopienko: Declined

test/parallel_api/numeric/numeric.ops/exclusive_scan.pass.cpp
```diff
@@ -73,12 +73,68 @@ test_with_usm(Policy&& exec)
...
+    
+    // Allocate USM shared memory for input (bool type) and output (int type)
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/utils.h
```diff
@@ -879,6 +846,109 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
+    else if constexpr (__is_equality_comparable<_Iterator2, _Iterator1>::value)
+    {
+        return __it2 == __it1;
+    }
+
+    return false;
+}
+
+template <typename _Iterator1, typename _Iterator2>
+bool
```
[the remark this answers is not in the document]
SergeyKopienko: Applied

include/oneapi/dpl/pstl/utils.h
```diff
@@ -879,6 +846,110 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
+
+// Iterators can be compared for equality if:
+//  - they are both iterators (or pointers);
+//  - the types of their iterated valued are the same;
+//  - the operator==(_Iterator1, _Iterator2) exist.
+template <typename _Iterator1, typename _Iterator2>
+using __is_equality_comparable =
+    std::conjunction<__both_types_are_iterators<_Iterator1, _Iterator2>,
+                     std::is_same<__iterator_value_type_t<_Iterator1>, __iterator_value_type_t<_Iterator2>>,
+                     std::is_invocable<std::equal_to<>, _Iterator1, _Iterator2>>;
```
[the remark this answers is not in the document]
SergeyKopienko: Declined: I don't know the cases when we will have not correctly detection if two iterators can be compared with `==`.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -879,6 +846,117 @@ struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<_T>::dif
...
+};
+
+// Unwind the iterator to its base type, if it has a base() method.
+template <typename _Iterator>
+struct __unwind_iterator<_Iterator, std::enable_if_t<__has_base<_Iterator>::value>>
+{
+    constexpr auto
+    operator()(_Iterator __it) const
+    {
+        return __unwind_iterator<decltype(__it.base())>{}(__it.base());
```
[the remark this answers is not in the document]
SergeyKopienko: It's looks like run-time unwind: declined.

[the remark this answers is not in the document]
SergeyKopienko: I think the current implementation is good enough, so declined.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,25 +818,100 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+template <typename _Iterator1, typename _Iterator2, typename = void>
+struct __is_equality_comparable : std::false_type
+{
+};
+
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable<_Iterator1, _Iterator2,
+                                std::enable_if_t<__is_equality_implemented<_Iterator1, _Iterator2>::value>>
+    : std::conditional_t<std::conjunction_v<__base_iterator<_Iterator1>, __base_iterator<_Iterator2>>,
+                         __is_equality_implemented<_Iterator1, _Iterator2>, // std::true_type,
```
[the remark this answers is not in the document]
SergeyKopienko: Applied

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,25 +818,95 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+};
+
+} // namespace __iterators_possibly_equal_impl
+
+// Checks if two iterators are possibly equal, i.e. if they can be compared for equality.
 template <typename _Iterator1, typename _Iterator2>
 constexpr bool
 __iterators_possibly_equal(_Iterator1 __it1, _Iterator2 __it2)
 {
+    using namespace __iterators_possibly_equal_impl;
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,109 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+                         __is_equality_self_comparable<_Iterator1, _Iterator2>,
+                         __is_equality_comparable<typename __base_iterator_type<_Iterator1>::__type,
+                                                  typename __base_iterator_type<_Iterator2>::__type>>
+{
+};
+#endif // _ONEDPL_CPP20_CONCEPTS_PRESENT
+
+template <typename _Iterator1, typename _Iterator2>
+inline constexpr bool __is_equality_comparable_v =
+    __is_equality_comparable_impl::__is_equality_comparable<_Iterator1, _Iterator2>::value;
```
danhoeflinger: Its a little strange to me to have this within the `__is_equality_compareable_impl` namespace.  
Id recommend moving this final `bool` outside to the `oneapi::dpl::__internal` namespace.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,109 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
 {
 };
 
-// All with implemented operator ==
 template <typename _Iterator1, typename _Iterator2>
-struct __is_equality_comparable<
-    _Iterator1, _Iterator2,
-    std::void_t<decltype(::std::declval<::std::decay_t<_Iterator1>>() == ::std::declval<::std::decay_t<_Iterator2>>())>>
+struct __is_equality_self_comparable<_Iterator1, _Iterator2,
+                                     std::enable_if_t<std::conjunction_v<__has_same_value_types<_Iterator1, _Iterator2>,
```
danhoeflinger: I'm not sure what having the same value type has to do with being equality comparable, should this be a requirement?

danhoeflinger: I guess its probably trying to do something similar to the `std::common_reference_t` part of the c++20 concept.  I *think* we should be OK with just checking the availability of `==` for the self type and the base type recursively, i don't think this changes any cases (but I'm happy to be proven wrong).

Also, this is only doing something for iterators, so if we *are* depending on it, then we need to rename this to be specific to iterators rather than interchangeable with the general purpose `std::is_equality_compareable_with`.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,109 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
 };
 
+template <typename _Iterator1, typename _Iterator2, typename = void>
+struct __is_equality_comparable : std::false_type
+{
+};
+
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable<_Iterator1, _Iterator2,
+                                std::enable_if_t<__is_equality_self_comparable<_Iterator1, _Iterator2>::value>>
```
danhoeflinger: This `enable_if_t` and the base specialization of `false_type` seems redundant with the `__is_equality_self_comparable<_Iterator1, _Iterator2>` inside the conjunction.  I think we can fully define it with the conjunction, and remove the enable if and base definition.

danhoeflinger: Oh, I was misreading it, but this is still strange.  It is required because as written, we need it to combine with the "false" condition for of the "is this a root base iterator type `conditional_t` check as we recurse down.  

I still think its preferable to to have an explicit 
```
std::conjunction<
   __is_equality_self_comparable<_Iterator1, _Iterator2>,
   __is_equality_comparable<typename __base_iterator_type<_Iterator1>::__type,
                            typename __base_iterator_type<_Iterator2>::__type
                           >
                >
``` 
in the false condition, rather than including this as an enable_if.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,109 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+#else
+template <typename _Iterator, typename = void>
+struct __base_iterator_type : std::true_type
+{
+    using __type = std::decay_t<_Iterator>;
+};
+
+template <typename _Iterator>
+struct __base_iterator_type<_Iterator, std::void_t<decltype(std::declval<std::decay_t<_Iterator>>().base())>>
+    : std::false_type
```
danhoeflinger: I tend to agree with copilot that this is a confusing name for the condition.
I thought it was representing:  
value - does this class have a `base()` function?
type - `decltype(base())`

However, the value is flipped from my expectation.  I dont think the code is wrong as written, but it is confusing.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,109 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+#if _ONEDPL_CPP20_CONCEPTS_PRESENT
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable : std::bool_constant<std::equality_comparable_with<_Iterator1, _Iterator2>>
+{
+};
+#else
+template <typename _Iterator, typename = void>
+struct __base_iterator_type : std::true_type
+{
+    using __type = std::decay_t<_Iterator>;
```
danhoeflinger: Id prefer to omit the `__type` from this base definition, I don't believe it is ever used, so why assign it something?

test/general/sycl_iterator/sycl_iterator_operations.pass.cpp
danhoeflinger, line 1: I think we may want to relocate this test to `test/general/implementation_details` since it is really a unit test of internals.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: The idea was that it is one place used thing,
Fixed: moved outside of namespace `__is_equality_comparable_impl`

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Unfortunately type check `__has_same_value_types<_Iterator1, _Iterator2>` required here.

Without this condition I have faced wit `assert` from MS STL library: https://github.com/microsoft/STL/blob/5f8b52546480a01d1d9be6c033e31dfce48d4f13/stl/inc/xutility#L1963
```C++
_EXPORT_STD template <class _BidIt1, class _BidIt2>
_NODISCARD _CONSTEXPR17 bool operator==(const reverse_iterator<_BidIt1>& _Left, const reverse_iterator<_BidIt2>& _Right)
    noexcept(noexcept(_STD _Fake_copy_init<bool>(_Left._Get_current() == _Right._Get_current()))) /* strengthened */
#if _HAS_CXX20
    requires requires {
        { _Left._Get_current() == _Right._Get_current() } -> _Implicitly_convertible_to<bool>;
    }
#endif // _HAS_CXX20
{
    return _Left._Get_current() == _Right._Get_current();
}
```

```C++
1>C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.44.35207\include\xutility(1897,71): : error : comparison of distinct pointer types ('int *' and 'bool *')
1> 1897 |     noexcept(noexcept(_STD _Fake_copy_init<bool>(_Left._Get_current() == _Right._Get_current()))) /* strengthened */
1>      |                                                  ~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~
1>C:\Work\ConsoleApplication6\ConsoleApplication6\ConsoleApplication6.cpp(12,69): note: in instantiation of exception specification for 'operator==<int *, bool *>' requested here
1>   12 | std::enable_if_t<!std::is_same_v<bool, decltype(std::declval<It1>() == std::declval<It2>())>>
1>      |                                                                     ^
1>C:\Work\ConsoleApplication6\ConsoleApplication6\ConsoleApplication6.cpp(25,5): note: while substituting deduced template arguments into function template 'test_eq' [with It1 = std::reverse_iterator<int *>, It2 = std::reverse_iterator<bool *>]
1>   25 |     test_eq(it1, it2);
1>      |     ^
1>1 error generated.

```

danhoeflinger: We are already checking the base class in our code here.  If we can check the self type for `==` operator if and only if {the `base()` types do not exist or they are they have `==` operator}, then we should be able to avoid this check.  Basically recurse down to the root iterator first via `base()` checks, and only check the self type if we are the root, or the base is `==` compare-able.  

If I understand correctly, this is providing a build error, but the base types would not have a valid `==` operator (int* == bool*), so this would allow us to avoid checking the self type which is where we see the build error manifest.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,97 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+using __is_equality_comparable_with = std::equality_comparable_with<_Iterator1, _Iterator2>;
+#else
+template <typename _Iterator, typename = void>
+struct __has_base_iterator : std::false_type
+{
+};
+
+template <typename _Iterator>
+struct __has_base_iterator<_Iterator, std::void_t<decltype(std::declval<std::decay_t<_Iterator>>().base())>>
+    : __is_iterator_type<decltype(std::declval<std::decay_t<_Iterator>>().base())>
```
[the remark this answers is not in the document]
SergeyKopienko: Applied.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,97 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
     : std::true_type
 {
 };
 
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable_with
+    : std::conditional_t<std::disjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
+                         __is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
+                                                       typename __base_iterator_type<_Iterator2>::__type>,
+                         __has_equality_op<_Iterator1, _Iterator2>>
```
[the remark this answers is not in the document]
SergeyKopienko: Declined.

test/parallel_api/numeric/numeric.ops/exclusive_scan.pass.cpp
```diff
@@ -73,12 +73,69 @@ test_with_usm(Policy&& exec)
...
+    // Create reverse iterators to test exclusive_scan's behavior when scanning from right to left.
+    // This verifies that exclusive_scan correctly handles reverse iterator ranges and produces expected results.
+    auto input_rbegin = std::reverse_iterator<bool*>(input + N);
+    auto input_rend = std::reverse_iterator<bool*>(input);
+
+    constexpr int initial_value = 0;
+
+    // Use exclusive_scan with reverse iterators to convert bool to int
+    // This will scan from right to left (due to reverse iterators)
+    // The initial value (0) will appear at the rightmost position
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Refactored.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Refactored.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Refactored.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: I think it's needed here:
```C++
struct __is_equality_comparable_with
    : std::conditional_t<std::disjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
                         __is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
                                                       typename __base_iterator_type<_Iterator2>::__type>,
                         __has_equality_op<_Iterator1, _Iterator2>>
{
};
```
if `__has_base_iterator<_Iterator1>` or  `__has_base_iterator<_Iterator2>` is `TRUE` then we evaluate
```C++
__is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
                              typename __base_iterator_type<_Iterator2>::__type>
```
and should have some type for each of them.

test/general/sycl_iterator/sycl_iterator_operations.pass.cpp
SergeyKopienko: This test has been renamed and moved to  `test/general/implementation_details/test_iterators_possibly_equal.pass.cpp`

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,87 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
-    std::void_t<decltype(::std::declval<::std::decay_t<_Iterator1>>() == ::std::declval<::std::decay_t<_Iterator2>>())>>
+struct __has_equality_op<_Iterator1, _Iterator2,
+                         std::void_t<decltype(std::declval<_Iterator1>() == std::declval<_Iterator2>())>>
     : std::true_type
 {
 };
 
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable_with
+    : std::conditional_t<std::disjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
```
danhoeflinger: Is disjunction(OR) the correct thing here rather than conjunction(AND)?

Lets say only one side has a base iterator, and the other does not.  Won't `__has_equality_op<It1,It2>` always be `false`? (and isn't that the proper response?)  I don't think we need to try to recurse if only one side has a base.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,87 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+                         __is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
+                                                       typename __base_iterator_type<_Iterator2>::__type>,
```
danhoeflinger: Seems like this needs to be:

``` 
struct __is_equality_comparable_with
    : std::conditional_t<std::conjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
                     std::conditional_t<__is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
                                                       typename __base_iterator_type<_Iterator2>::__type>,  
                                       __has_equality_op<Iterator1,Iterator2>,
                                       std::false_type>,
                    __has_equality_op<_Iterator1, _Iterator2>>
                                                       
```
...so that at each level we test the base, then test the self type, but only if the base succeeds (or doesn't exist).

edit:  for clarity I've included the full struct definition rather than just a piece.
edit2: switched from disjunction to conjunction as per my previous comment.

akukanov: No, it should be in the opposite order: we first test the types, then - if the types can be compared - we test the bases. The second step is really only necessary because of that unrestricted implementation of `==` for `reverse_iterator` etc. which compares the base iterators even if those are not comparable - so we have to implement a workaround.

SergeyKopienko: > Seems like this needs to be:
> 
> ```
> template <typename _Iterator1, typename _Iterator2>
> struct __is_equality_comparable_with
>     : std::conditional_t<std::conjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
>                      std::conditional_t<__is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
>                                                        typename __base_iterator_type<_Iterator2>::__type>,  
>                                        __has_equality_op<Iterator1,Iterator2>,
>                                        std::false_type>,
>                     __has_equality_op<_Iterator1, _Iterator2>>
>                                                        
> ```
> 
> ...so that at each level we test the base, then test the self type, but only if the base succeeds (or doesn't exist).
> 
> edit: for clarity I've included the full struct definition rather than just a piece. edit2: switched from disjunction to conjunction as per my previous comment.

This variant
```C++
struct __is_equality_comparable_with
    : std::conditional_t<
          std::conjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
          std::conditional_t<__is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
                                                           typename __base_iterator_type<_Iterator2>::__type>,
                             __has_equality_op<_Iterator1, _Iterator2>, std::false_type>,
          __has_equality_op<_Iterator1, _Iterator2>>
{
};
```
has compile errors:
```C++
  In file included from test\general\implementation_details\test_sycl_iterator_operations.pass.cpp:18:
  In file included from include\oneapi\dpl\iterator:26:
  In file included from include\oneapi\dpl\pstl\iterator_impl.h:25:
include\oneapi\dpl\pstl\utils.h(885,30): error : template argument for non-type template parameter must be an expression
    885 |           std::conditional_t<__is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
        |                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    886 |                                                            typename __base_iterator_type<_Iterator2>::__type>,
        |                                                            ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  C:\Program Files\Microsoft Visual Studio\2022\Professional\VC\Tools\MSVC\14.44.35207\include\xtr1common(68,28): note: template parameter is declared here
     68 | _EXPORT_STD template <bool _Test, class _Ty1, class _Ty2>
        |                            ^
  In file included from test\general\implementation_details\test_sycl_iterator_operations.pass.cpp:18:
  In file included from include\oneapi\dpl\iterator:26:
  In file included from include\oneapi\dpl\pstl\iterator_impl.h:25:
include\oneapi\dpl\pstl\utils.h(896,91): error : no member named 'value' in 'oneapi::dpl::__internal::__is_equality_comparable_impl::__is_equality_comparable_with<int *, int>'
    896 |     __is_equality_comparable_impl::__is_equality_comparable_with<_Iterator1, _Iterator2>::value;
        |     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
  test\general\implementation_details\test_sycl_iterator_operations.pass.cpp(90,20): note: in instantiation of variable template specialization 'oneapi::dpl::__internal::__is_equality_comparable_with_v<int *, int>' requested here
     90 |     static_assert(!__is_equality_comparable_with_v<int*, int       >);
```

include/oneapi/dpl/pstl/utils.h
akukanov: If `__has_equality_op<It1,It2>` is false, we do not need to check the base iterators at all. But what if these types are comparable?

Let's say in tests we want to count how many times an algorithm dereferences its input iterators (e.g. to check complexity) and for that purpose we develop a "dereference counting" adaptor. We would want that to be comparable with the original iterator, would not we?

akukanov: So the interesting scenario is that of `__has_equality_op<It1,It2>` being true, yet one iterator having the base and the other not having it. It's reasonable to expect the implementation of such comparison to compare the second iterator with the base of the first one. And the question is then, again, the necessity to "protect ourselves" against implementations that do not properly restrict the availability of `==`.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: > No, it should be in the opposite order: we first test the types, then - if the types can be compared - we test the bases. The second step is really only necessary because of that unrestricted implementation of `==` for `reverse_iterator` etc. which compares the base iterators even if those are not comparable - so we have to implement a workaround.

Yes, I understand why we have the involvement of the base types in the first place to work around the issue with `move_iterator` and `reverse_iterator`.  However, I was looking thinking about https://github.com/uxlfoundation/oneDPL/pull/2371#discussion_r2265428111 the error mentioned here.  It may be the wrong diagnosis, I don't have the example myself, but it looked to me like it was avoidable by first checking base types then moving to self type (as opposed to checking value_types of iterators).

akukanov: The approach with comparing value types of iterators was really incorrect (because these have little if anything to do with comparing the iterators themselves), and has gone by now - so the discussion you point to seems no more relevant.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,87 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+template <typename _T>
+static constexpr bool __is_iterator_type_v = __is_iterator_type<_T>::value;
+
+namespace __is_equality_comparable_impl
+{
+#if _ONEDPL_CPP20_CONCEPTS_PRESENT
+template <typename _Iterator1, typename _Iterator2>
+using __is_equality_comparable_with = std::bool_constant<std::equality_comparable_with<_Iterator1, _Iterator2>>;
+#else
+template <typename _Iterator, typename = void>
```
akukanov:
```suggestion
#if _ONEDPL_CPP20_CONCEPTS_PRESENT

template <typename _Iterator1, typename _Iterator2>
using __is_equality_comparable_with = std::bool_constant<std::equality_comparable_with<_Iterator1, _Iterator2>>;

#else

template <typename _Iterator, typename = void>
```

Some extra space helps to avoid false impression that `#if` only applies to a small code block till the next empty line.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,87 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+struct __has_base_iterator<
+    _Iterator, std::enable_if_t<__is_iterator_type_v<decltype(std::declval<std::decay_t<_Iterator>>().base())>>>
+    : std::true_type
+{
+};
+
+template <typename _Iterator, typename = void>
+struct __base_iterator_type
+{
+    using __type = std::decay_t<_Iterator>;
```
danhoeflinger: if we must declare a type here for the case which does not have a base, perhaps we should use `void`?

include/oneapi/dpl/pstl/utils.h
danhoeflinger: I understand the example now, and agree that disjunction is OK as long as we have the self check at each level of recursion.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: After offline discussion and from @akukanov example https://github.com/uxlfoundation/oneDPL/pull/2371/files#r2266637457, void is not appropriate, but the iterator type is OK for me.

test/support/sycl_alloc_utils.h
```diff
@@ -157,8 +157,6 @@ class usm_data_transfer
 
             //TODO: support copying data provided by non-contiguous iterator
             auto __src = std::addressof(*__it);
-            assert(std::addressof(*(__it + __objects_count)) - __src == __objects_count);
-
```
akukanov: This change and the one below seem unrelated to the patch. Is it accidental / an artifact of early prototyping?

include/oneapi/dpl/pstl/utils.h

test/support/sycl_alloc_utils.h
SergeyKopienko: This is an error which was found during this PR: we had cases when made dereference of out-of-range iterator:
```C++
*(__it + __objects_count)
```

akukanov: Are these deliberate test cases? Can they be modified not to do out-of-range dereference without value loss?

test/parallel_api/numeric/numeric.ops/exclusive_scan.pass.cpp
```diff
@@ -73,12 +73,70 @@ test_with_usm(Policy&& exec)
...
+    auto input_rbegin = std::reverse_iterator<bool*>(input + N);
+    auto input_rend = std::reverse_iterator<bool*>(input);
+
+    constexpr int initial_value = 0;
+
+    // Use exclusive_scan with reverse iterators to convert bool to int
+    // This will scan from right to left (due to reverse iterators)
+    // The use of reverse iterators causes exclusive_scan to process elements in reverse order,
+    // but the algorithm's semantics remain unchanged. The initial value (0) will appear at the rightmost position.
+    auto result_rbegin = std::reverse_iterator<int*>(result + N);
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,89 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
 };
 
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable_with
+    : std::conditional_t<
+          std::disjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
+          std::conjunction<__is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
+                                                         typename __base_iterator_type<_Iterator2>::__type>,
+                           __has_equality_op<_Iterator1, _Iterator2>>,
+          __has_equality_op<_Iterator1, _Iterator2>>
```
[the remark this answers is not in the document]
SergeyKopienko: Declined: I don't think we should have compile-time recursion for nested iterator adapters

test/parallel_api/numeric/numeric.ops/exclusive_scan.pass.cpp
```diff
@@ -73,12 +73,70 @@ test_with_usm(Policy&& exec)
...
+    input[1] = false;
+    input[2] = true;
+    input[3] = true;
+    input[4] = false;
+    input[5] = true;
+
+    // Create reverse iterators to test exclusive_scan's behavior when scanning from right to left.
+    // This verifies that exclusive_scan correctly handles reverse iterator ranges and produces expected results.
+    auto input_rbegin = std::reverse_iterator<bool*>(input + N);
+    auto input_rend = std::reverse_iterator<bool*>(input);
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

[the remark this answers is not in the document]
SergeyKopienko: Declined

test/support/sycl_alloc_utils.h
SergeyKopienko: Fixed: now we restore `assert` and avoid to out-of-range iterator dereference.

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
```diff
@@ -13,14 +13,118 @@
...
+
+    static_assert(!__is_equality_comparable_with_v<int*, int       >);
+    static_assert( __is_equality_comparable_with_v<int*, int*      >);
+    static_assert(!__is_equality_comparable_with_v<int*, float*    >);
+    static_assert( __is_equality_comparable_with_v<int*, const int*>);
+    static_assert( __is_equality_comparable_with_v<decltype(std::vector<int>().begin()), 
+                                                   decltype(std::vector<int>().cbegin())>);
+    static_assert(!__is_equality_comparable_with_v<decltype(std::vector<int>().begin()), 
+                                                   decltype(std::vector<float>().cbegin())>);
+
```
danhoeflinger: I'm a little surprised not to see `move_iterator` and `reverse_iterator` of  different base iterator types here being checked. 
It would be useful to also check doubly nested versions `move_iterator<reverse_iterator<T*>>`  in both positive and negative way.

danhoeflinger: https://github.com/uxlfoundation/oneDPL/pull/2378

danhoeflinger (dismissed):
LGTM, probably would be good to have someone else approve as well with my two merged commits.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,101 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
-    std::void_t<decltype(::std::declval<::std::decay_t<_Iterator1>>() == ::std::declval<::std::decay_t<_Iterator2>>())>>
+struct __has_equality_op<_Iterator1, _Iterator2,
+                         std::void_t<decltype(std::declval<_Iterator1>() == std::declval<_Iterator2>())>>
     : std::true_type
 {
 };
 
+// Checks equality comparability of two iterators.
+// Pre-C++20 iterator adapters (move_iterator, reverse_iterator) have overly permissive operator== definitions that
+// cause compile-time checks to disagree with runtime availability. This issue requires more than simple
```
akukanov:
```suggestion
// Pre-C++20 iterator adapters (move_iterator, reverse_iterator) have overly permissive operator== definitions that
// cause SFINAE checks to pass for operators that cannot then compile. This issue requires more than simple
```

danhoeflinger: This works for me.

akukanov (approved):

danhoeflinger (approved):
reapproving

include/oneapi/dpl/pstl/utils.h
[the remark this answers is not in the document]
rarutyun: > I don't think we should have compile-time recursion for nested iterator adapters

That's exactly what we have here. However, I don't think that AI is correct in suggesting setting up the depth limit. I also don't see (immediately) how we can flatten the structure

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,101 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+};
+
+template <typename _T>
+struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<std::decay_t<_T>>::difference_type>>
+    : std::true_type
+{
+};
+
+template <typename _T>
+static constexpr bool __is_iterator_type_v = __is_iterator_type<_T>::value;
```
rarutyun:
```suggestion
inline constexpr bool __is_iterator_type_v = __is_iterator_type<_T>::value;
```

Or please explain why is it static?

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,101 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+
+#else
+
+template <typename _Iterator, typename = void>
+struct __has_base_iterator : std::false_type
+{
+};
+
+template <typename _Iterator>
+struct __has_base_iterator<
```
rarutyun: `__has_base_iterator` and `__base_iterator_type` are essentially the same traits (unless I am missing something). Why do we need both of them?

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,101 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+template <typename _T>
+struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<std::decay_t<_T>>::difference_type>>
+    : std::true_type
+{
+};
+
+template <typename _T>
+static constexpr bool __is_iterator_type_v = __is_iterator_type<_T>::value;
+
+namespace __is_equality_comparable_impl
```
rarutyun: I don't see a reason why we need a separate namespace

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,101 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+          __has_equality_op<_Iterator1, _Iterator2>>
+{
+};
+
+#endif // _ONEDPL_CPP20_CONCEPTS_PRESENT
+} // namespace __is_equality_comparable_impl
+
+template <typename _Iterator1, typename _Iterator2>
+inline constexpr bool __is_equality_comparable_with_v =
+    __is_equality_comparable_impl::__is_equality_comparable_with<_Iterator1, _Iterator2>::value;
```
rarutyun: I would suggest to make it differently. I would put this under #ifdef and remove namespace (unless we have reasons to introduce it). Please also see `__is_equality_comparable_with_v` suggestion above

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,101 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+
+template <typename _T>
+static constexpr bool __is_iterator_type_v = __is_iterator_type<_T>::value;
+
+namespace __is_equality_comparable_impl
+{
+#if _ONEDPL_CPP20_CONCEPTS_PRESENT
+
+template <typename _Iterator1, typename _Iterator2>
+using __is_equality_comparable_with = std::bool_constant<std::equality_comparable_with<_Iterator1, _Iterator2>>;
```
rarutyun:
```suggestion
template <typename _Iterator1, typename _Iterator2>
inline constexpr bool __is_equality_comparable_with_v = std::equality_comparable_with<_Iterator1, _Iterator2>;
```

There is no point to create a type out of concept and then read it below as the value. I would rather store it as the value and either keep it like this and don't introduce a type at all (since we don't use it afterwards) or if you must, you can introduce the type as

```cpp
using __is_equality_comparable_with = std::bool_constant<__is_equality_comparable_with_v<_Iterator1, Iterator2>;
```

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
```diff
@@ -0,0 +1,274 @@
...
+    // two sycl_iterators pointing to different elements in the same "root" buffer(expect false)
+    auto it1next = it1 + 1;
+    EXPECT_FALSE(__iterators_possibly_equal(it1, it1next), "wrong __iterators_possibly_equal result");
+
+    {
+        float floatData = .0;
+
+        ::std::vector<int> dataVec{1, 2, 3};
+        const auto intConstData = dataVec.data();
```
rarutyun: why not just use the tools that are already in `std::vector` for decades. `begin` and `cbegin` should do the job

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
```diff
@@ -0,0 +1,274 @@
...
+                "wrong __iterators_possibly_equal result");
+
+    // two sycl_iterators pointing to different elements in the same "root" buffer(expect false)
+    auto it1next = it1 + 1;
+    EXPECT_FALSE(__iterators_possibly_equal(it1, it1next), "wrong __iterators_possibly_equal result");
+
+    {
+        float floatData = .0;
+
+        ::std::vector<int> dataVec{1, 2, 3};
```
rarutyun:
```suggestion
        std::vector<int> dataVec{1, 2, 3};
```

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
```diff
@@ -0,0 +1,274 @@
...
+#endif // _ONEDPL_CPP20_CONCEPTS_PRESENT
+
+// Check the correctness of oneapi::dpl::__internal::__iterators_possibly_equal
+void
+test_iterators_possibly_equal()
+{
+    // Check some internals from oneapi::dpl::__internal
+    using namespace oneapi::dpl::__internal;
+
+    constexpr size_t count = 0;
```
rarutyun:
```suggestion
    constexpr std::size_t count = 0;
```

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
```diff
@@ -0,0 +1,274 @@
...
+    static_assert(!__is_equality_comparable_with_v<__zip_iterator_1, __zip_iterator_2>);
+}
+#endif // _ONEDPL_CPP20_CONCEPTS_PRESENT
+
+// Check the correctness of oneapi::dpl::__internal::__iterators_possibly_equal
+void
+test_iterators_possibly_equal()
+{
+    // Check some internals from oneapi::dpl::__internal
+    using namespace oneapi::dpl::__internal;
```
rarutyun: honestly, it's looks quite dangerous to make everything visible from `__internal` namespace. I would suggest to create a short namespace alias, like `namespace internal = oneapi::dpl::__internal` or something similar

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,101 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
 
+// Checks equality comparability of two iterators.
+// Pre-C++20 iterator adapters (move_iterator, reverse_iterator) have overly permissive operator== definitions that
+// cause SFINAE checks to pass for operators that cannot then compile. This issue requires more than simple
+// __has_equality_op checks to determine if two iterators are equality comparable. We must recursively check base
+// iterator types first to avoid build errors with incompatible base types.
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable_with
+    : std::conditional_t<
+          std::disjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
```
rarutyun: maybe, again, I am missing something but the logic looks weird to me. We check whether any of the types has `base()` and then ask `base()` from both. Do we expect `std::reverse_iterator<int*>` and `int*` to "work"? I know that it will return `false` but this is just an example. A broader question is do we want to even try to support different depth of iterator adapters?

If so, I didn't find any test that would check that. I also doubt that we can come up with a reasonable positive test for this case.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,101 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+    _Iterator, std::enable_if_t<__is_iterator_type_v<decltype(std::declval<std::decay_t<_Iterator>>().base())>>>
+    : std::true_type
+{
+};
+
+template <typename _Iterator, typename = void>
+struct __base_iterator_type
+{
+    // If no base is available, use the iterator type itself. This supports iterator adapters
+    // that can be compared to their base type with a user-defined operator==.
```
rarutyun: I cannot find any iterator adapter that would define `operator==` with there base iterator type. I don't think that user-defined `operator==` is a realistic use-case. I also cannot find such tests. Please see a similar comment below.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: This comes from @akukanov here: https://github.com/uxlfoundation/oneDPL/pull/2371#discussion_r2266637457

It was discussed in an offline conversation.  Its a question of where to draw the line of what we want to support.  We already are excluding some conceivable iterators which overload `operator==` to do non-standard things, like allow equal comparison with non equal comparable `base()` types. 

 Alexey brought up a case where you may want to define an adapter which is equal comparable with its base.  It seems that this does not cost us anything to use disjunction and put the iterator type as the base type for iterators which don't provide a base. Types which don't intend to allow equal comparable will be weeded out by the self-type equal comparable check.  

However, one could argue as I did originally that this should be conjunction and `void` type as the base of iterators without a `base()` function.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: See the comment above https://github.com/uxlfoundation/oneDPL/pull/2371#discussion_r2268083488.

include/oneapi/dpl/pstl/utils.h
rarutyun: Yes. I also agree with `conjunction`. However, my main complain is if this logic intentional (I don't necessarily agree with this choice), let's at least add the test that checks it (preferably both positive and negative). Otherwise, I am not happy with the logic that we implement (likely, don't ever need in my opinion) and don't test it

include/oneapi/dpl/pstl/utils.h
rarutyun: Basically, the same answer as [here](https://github.com/uxlfoundation/oneDPL/pull/2371/files#r2268112744). If we want to keep this logic please either show me the test (if I missed one) or we should add it. I am not a fan of this decision but I can live with that if it's properly tested

include/oneapi/dpl/pstl/utils.h
danhoeflinger: I tried to make a test and I'm not really happy with the result.   #2379

It ends up with a case that only works if the source iterator has no base.  I think any amount of merit this example has is lost in that it is not consistent across a bigger variety of source iterators. I think I'm back to preferring conjunction and a `void` type for a base-less iterator base type.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: This code has been just moved before my namespace.
So this was introduced earlier, not in this PR.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: This code has been just moved before my namespace.
So this was introduced earlier, not in this PR.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Mine intention was to pack implementation details in this PR into some new namespace because we have here a lot of details.
Does this solution impacting anything?

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Issue: https://github.com/uxlfoundation/oneDPL/issues/2380
PR prepared: https://github.com/uxlfoundation/oneDPL/pull/2381

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
SergeyKopienko: I am ready to create some issue and fix this somehow later.

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
SergeyKopienko: Fixed

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
SergeyKopienko: Fixed

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
SergeyKopienko: Fixed

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
SergeyKopienko: Also, this `using namespace oneapi::dpl::__internal;` exists only in one place, so I don't understand how some new alias will help us.

include/oneapi/dpl/pstl/utils.h
akukanov: Dan, your tests break not because of comparing an iterator with base and an iterator without base. Your tests break because two iterators with base are comparable. but their bases are not. In other words, it breaks after the two-side recursion, not the one-side one. The assumption that is broken is that when both operators have base, the comparison compares both bases.

include/oneapi/dpl/pstl/utils.h
akukanov: > I cannot find any iterator adapter that would define `operator==` with there base iterator type. I don't think that user-defined `operator==` is a realistic use-case. I also cannot find such tests. Please see a similar comment below.

Well, I have found one: https://en.cppreference.com/w/cpp/iterator/basic_const_iterator/operator_cmp.html
You can say it's C++23 and will therefore not go into the code in question. Yet, it illustrates the principle.

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: My idea was to have `__is_equality_comparable_with` as a type in both branches: if `_ONEDPL_CPP20_CONCEPTS_PRESENT` is defined or not defined,

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: The trait `__has_base_iterator` describes does this iterator has some `base` iterator through `base()` function:
```C++
template <typename _Iterator, typename = void>
struct __has_base_iterator : std::false_type
{
};

template <typename _Iterator>
struct __has_base_iterator<
    _Iterator, std::enable_if_t<__is_iterator_type_v<decltype(std::declval<std::decay_t<_Iterator>>().base())>>>
    : std::true_type
{
};
```

But the `__base_iterator_type` describes the type of  the `base` iterator:
```C++
template <typename _Iterator, typename = void>
struct __base_iterator_type
{
    // If no base is available, use the iterator type itself. This supports iterator adapters
    // that can be compared to their base type with a user-defined operator==.
    using __type = std::decay_t<_Iterator>;
};

template <typename _Iterator>
struct __base_iterator_type<_Iterator, std::void_t<decltype(std::declval<std::decay_t<_Iterator>>().base())>>
{
    using __type = decltype(std::declval<std::decay_t<_Iterator>>().base());
};
```

As result, we have more readable usage of them in the `__is_equality_comparable_with` :
```C++
struct __is_equality_comparable_with
    : std::conditional_t<
          std::disjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
          // MS STL encounters build errors if we check current iterator before confirming bases are comparable
          std::conjunction<__is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
                                                         typename __base_iterator_type<_Iterator2>::__type>,
                           __has_equality_op<_Iterator1, _Iterator2>>,
          __has_equality_op<_Iterator1, _Iterator2>>
{
};
```

Additionally, this moment has been already discussed at https://github.com/uxlfoundation/oneDPL/pull/2371/files/590ce74cab5276e7a06cfb8c49f819b9a536ce32#r2264012336 : previous implementation were not so readable and had semantics opposite to the reader's expectations

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: The namespace `__is_equality_comparable_impl` has been removed.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Fixed.

include/oneapi/dpl/pstl/utils.h

include/oneapi/dpl/pstl/utils.h
```diff
@@ -882,19 +872,9 @@ struct __has_equality_op<_Iterator1, _Iterator2,
...
-struct __is_equality_comparable_with
-    : std::conditional_t<
-          std::disjunction_v<__has_base_iterator<_Iterator1>, __has_base_iterator<_Iterator2>>,
-          // MS STL encounters build errors if we check current iterator before confirming bases are comparable
-          std::conjunction<__is_equality_comparable_with<typename __base_iterator_type<_Iterator1>::__type,
-                                                         typename __base_iterator_type<_Iterator2>::__type>,
-                           __has_equality_op<_Iterator1, _Iterator2>>,
-          __has_equality_op<_Iterator1, _Iterator2>>
+struct __is_equality_comparable_with : __has_equality_op<typename __base_iterator_type<_Iterator1>::__type,
+                                                         typename __base_iterator_type<_Iterator2>::__type>
```
danhoeflinger: This is not recursive.  
What about `reverse_iterator<reverse_iterator<int*>>`?

SergeyKopienko: Updated.

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
```diff
@@ -0,0 +1,304 @@
...
+        oneapi::dpl::__internal::sycl_iterator<sycl::access::mode::read_write, unsigned int>>;
+
+    using __zip_iterator_2 =
+        oneapi::dpl::zip_iterator<unsigned long long*,
+                                  oneapi::dpl::__internal::sycl_iterator<sycl::access::mode::read_write, unsigned int>>;
+
+    static_assert(!__is_equality_comparable_with_v<__zip_iterator_1, __zip_iterator_2>);
+}
+
+#if _ONEDPL_CPP20_CONCEPTS_PRESENT
```
danhoeflinger: Is this supposed to have tests?

Edit: I'd be fine with just removing this and the line calling it.

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
```diff
@@ -0,0 +1,304 @@
...
+        oneapi::dpl::__internal::sycl_iterator<sycl::access::mode::read_write, unsigned long long>,
+        oneapi::dpl::__internal::sycl_iterator<sycl::access::mode::read_write, unsigned int>>;
+
+    using __zip_iterator_2 =
+        oneapi::dpl::zip_iterator<unsigned long long*,
+                                  oneapi::dpl::__internal::sycl_iterator<sycl::access::mode::read_write, unsigned int>>;
+
+    using __zip_iterator_1_base = decltype(declval<__zip_iterator_1>().base());
+    using __zip_iterator_2_base = decltype(declval<__zip_iterator_2>().base());
+
```
danhoeflinger: can remove these now.

SergeyKopienko: Done

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
SergeyKopienko: Removed

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
SergeyKopienko: Removed

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
```diff
@@ -0,0 +1,287 @@
...
+        EXPECT_TRUE(__internal::__iterators_possibly_equal(intData, intConstData), "wrong __iterators_possibly_equal result");
+        // check pointer + pointer to other type
+        EXPECT_FALSE(__internal::__iterators_possibly_equal(intData, &floatData), "wrong __iterators_possibly_equal result");
+    }
+
+    {
+        std::vector<int> dataVec1{1, 2, 3};
+        std::vector<int> dataVec2{4, 5, 6};
+        EXPECT_FALSE(__internal::__iterators_possibly_equal(dataVec1.begin(), dataVec2.begin()), "wrong __iterators_possibly_equal result");
+    }
```
danhoeflinger, line 224: If I understand our offline discussion correctly, this will throw with MS STL and debug due to comparing iterators from different containers.
Do we want to protect this with a "broken macro" for this case (especially since this is a unit test).  At this point, we don't really have a strategy for resolving this in that situation (other than removing the need to use this function from exclusive_scan). Do we want to pollute our test results with this known failure case, which we don't really plan to fix?

include/oneapi/dpl/pstl/utils.h
```diff
@@ -818,30 +822,57 @@ __shars_upper_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __val
...
+};
+
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable_with<std::reverse_iterator<_Iterator1>, std::reverse_iterator<_Iterator2>>
+    : __is_equality_comparable_with<_Iterator1, _Iterator2>
+{
+};
+
+template <typename _Iterator1, typename _Iterator2>
+struct __is_equality_comparable_with<std::move_iterator<_Iterator1>, std::move_iterator<_Iterator2>>
```
rarutyun: The question is the same as in our private conversation. Do we want to add extra layer with `decay_t`? our previous logic did that for iterators. On the idea level it should be like

```cpp
struct __is_equality_comparable_with : __is_equality_comparable_with_impl<std::decay_t<_Iterator1>, std::decay_t<_Iterator2>>
{
};
```

and then we have partial specializations for reverse and move iterators for `_impl`

test/general/implementation_details/test_sycl_iterator_operations.pass.cpp
SergeyKopienko: Done

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Probably for the comparison with
```C++
template <typename _T>
struct __is_type_with_iterator_traits<
    _T, std::void_t<typename std::iterator_traits<std::remove_reference_t<_T>>::difference_type>> : std::true_type
{
};
```
and taking into account the discussion in the PR https://github.com/uxlfoundation/oneDPL/pull/2376 (https://github.com/uxlfoundation/oneDPL/pull/2376/files/c2ad9512c576d965561d12de72ba5d5a47b43b6c#r2267265672) I think we may have some common `iterator type` cleaner for both these usages like:
```C++
template <typename _Iterator>
using __clean_iterator_t = std::remove_reference_t<_Iterator>;
```
and use it.

SergeyKopienko: Also I think make sense to apply this `__clean_iterator_t` (or may be we would like to rename it somehow) in all other place (at lease in utils) where we using `std::iterator_traits`.
May be not in this PR.

rarutyun: The trait you are talking about it different. Yes, there should not be a difference between reading (some) iterator traits from iterator and `const` iterator (of the same type). However, here you have a partial template specialization. If some how (during our manipulation with the type you end up with `const` iterator the partial specialization would not match and the code will go to a primary template. Thus just having `remove_reference` is not sufficient.

danhoeflinger (dismissed):
This LGTM now

include/oneapi/dpl/pstl/utils.h
danhoeflinger: I already approved, but I realized it would be nice to add a test or two checking that iterator types with `const` and/or different value categories work.  I don't believe that is covered in the tests.

SergeyKopienko: Will do, but a little bit later (in this PR)

SergeyKopienko: @danhoeflinger, done.

danhoeflinger (approved):
thanks for the testing changes. LGTM

PR #2373, opened 2025-08-07
Fix forwarded value captured policy 

danhoeflinger:
In this PR we switch from forwarding after a capture by value to a capture by reference and then pass through.

Forwarding doesn't make sense here as the type may no longer match.
 
branched off from 
https://github.com/uxlfoundation/oneDPL/pull/2369/files#r2254402955

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3396,9 +3396,12 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
         //{1} < {2}: seq2 is wholly greater than seq1, so, do parallel copying seq1 and seq2
         __par_backend::__parallel_invoke(
             __backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec),
-            [=] { __internal::__pattern_walk2_brick(__tag, __exec, __first1, __last1, __result, __copy_range); },
             [=] {
-                __internal::__pattern_walk2_brick(__tag, __exec, __first2, __last2, __result + __n1, __copy_range);
+                __internal::__pattern_walk2_brick(__tag, std::move(__exec), __first1, __last1, __result, __copy_range);
+            },
+            [=] {
+                __internal::__pattern_walk2_brick(__tag, std::move(__exec), __first2, __last2, __result + __n1,
+                                                  __copy_range);
             });
         return __result + __n1 + __n2;
     }
@@ -3411,9 +3414,12 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
         //{2} < {1}: seq2 is wholly greater than seq1, so, do parallel copying seq1 and seq2
         __par_backend::__parallel_invoke(
             __backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec),
-            [=] { __internal::__pattern_walk2_brick(__tag, __exec, __first2, __last2, __result, __copy_range); },
             [=] {
-                __internal::__pattern_walk2_brick(__tag, __exec, __first1, __last1, __result + __n2, __copy_range);
+                __internal::__pattern_walk2_brick(__tag, std::move(__exec), __first2, __last2, __result, __copy_range);
+            },
+            [=] {
+                __internal::__pattern_walk2_brick(__tag, std::move(__exec), __first1, __last1, __result + __n2,
+                                                  __copy_range);
             });
         return __result + __n1 + __n2;
     }
@@ -3427,11 +3433,12 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
             __backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec),
             //do parallel copying of [first1; left_bound_seq_1)
             [=] {
-                __internal::__pattern_walk2_brick(__tag, __exec, __first1, __left_bound_seq_1, __res_or, __copy_range);
+                __internal::__pattern_walk2_brick(__tag, std::move(__exec), __first1, __left_bound_seq_1, __res_or,
+                                                  __copy_range);
             },
             [=, &__result] {
                 __result = __internal::__parallel_set_op(
-                    __tag, __exec, __left_bound_seq_1, __last1, __first2, __last2, __result, __comp,
+                    __tag, std::move(__exec), __left_bound_seq_1, __last1, __first2, __last2, __result, __comp,
                     [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op);
             });
         return __result;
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3396,13 +3396,9 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
         //{1} < {2}: seq2 is wholly greater than seq1, so, do parallel copying seq1 and seq2
         __par_backend::__parallel_invoke(
             __backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec),
+            [=] { __internal::__pattern_walk2_brick(__tag, __exec, __first1, __last1, __result, __copy_range); },
             [=] {
```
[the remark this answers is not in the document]
danhoeflinger: unnecessary in this case.

SergeyKopienko:
@danhoeflinger As far as I understand there are no any technical error, so we just make some preparations to capture something by reference. Correct?

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3396,13 +3396,9 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
...
-                __internal::__pattern_walk2_brick(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first1, __last1,
-                                                  __result, __copy_range);
-            },
-            [=] {
-                __internal::__pattern_walk2_brick(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first2, __last2,
-                                                  __result + __n1, __copy_range);
+                __internal::__pattern_walk2_brick(__tag, __exec, __first2, __last2, __result + __n1, __copy_range);
```
SergeyKopienko, line 3403: I believe  it's not the problem here (exactly on this line) and in all other same places.
Will comment in PR what happens from my point of view.

SergeyKopienko:
@danhoeflinger I have this really have not quite correct code before your changes.
Let's take a look for example to the code:
```C++
template <class _IsVector, class _ExecutionPolicy, class _RandomAccessIterator1, class _RandomAccessIterator2,
          class _OutputIterator, class _Compare, class _SetUnionOp>
_OutputIterator
__parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __exec, _RandomAccessIterator1 __first1,
                        _RandomAccessIterator1 __last1, _RandomAccessIterator2 __first2, _RandomAccessIterator2 __last2,
                        _OutputIterator __result, _Compare __comp, _SetUnionOp __set_union_op)
{
    //...

    if (__left_bound_seq_1 == __last1)
    {
        //{1} < {2}: seq2 is wholly greater than seq1, so, do parallel copying seq1 and seq2
        __par_backend::__parallel_invoke(
            __backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec),
            [=] {
                __internal::__pattern_walk2_brick(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first1, __last1,
                                                  __result, __copy_range);
            },
            [=] {
                __internal::__pattern_walk2_brick(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first2, __last2,
                                                  __result + __n1, __copy_range);
            });
        return __result + __n1 + __n2;
    }

    //...
}
```

akukanov:
Folks, where have you found several forwards in the original code?
We remember that `std::forward` is just a type cast, do not we? No matter in which order the function call parameters are evaluated, both lambdas get **copies** of the original policy object and then forward these copies to the invoked bricks, while the original object is forwarded to `__parallel_invoke`. What do you think is problematic here?

SergeyKopienko:
> Folks, where have you found "multiple forwards" in the original code? We remember that `std::forward` is just a type cast, do not we? No matter in which order the function call parameters are evaluated, both lambdas get **copies** of the original policy object and then forward these copies to the invoked bricks, while the original object is forwarded to `__parallel_invoke`. What do you think is problematic here?

Agreed !

danhoeflinger:
> Folks, where have you found "multiple forwards" in the original code? We remember that `std::forward` is just a type cast, do not we? No matter in which order the function call parameters are evaluated, both lambdas get **copies** of the original policy object and then forward these copies to the invoked bricks, while the original object is forwarded to `__parallel_invoke`. What do you think is problematic here?

I understand that forward is merely a type cast.  I don't think I did a good job of stating the problem.

https://godbolt.org/z/75ejWxY6b

The real issue here (as I understand it) is that when capturing by value into the lambda, the type of `__exec` in the lambda becomes `const _ExecutionPolicy`, even if the incoming "universal" type was non-const rvalue `_ExecutionPolicy&&`.  This means that within the lambda we are using `std::forward<_ExecutionPolicy>(_exec)` which will try to create an rvalue ref out of a const lvalue ref.  The mismatch of constness in the forward causes a build issue (as can be seen in godbolt), but also, the mismatch in value category is a not good. In my opinion, forwarding here really doesnt make sense.  

What I was trying to mention with the "multiple forwards" comment is what is shown in the last option in the godbolt.  *IF* we were to try to "fix" this by capture by reference, the two usages in the lambda would fail with a use-after-move issue.

In this exploration, I think that the original version here in this PR is probably not as good as if we were to `std::move` within the lambdas, so I will adjust to do that.

If I'm missing something here with my explanation and godbolt, please let me know.

SergeyKopienko:
> Folks, where have you found "multiple forwards" in the original code? We remember that `std::forward` is just a type cast, do not we? No matter in which order the function call parameters are evaluated, both lambdas get **copies** of the original policy object and then forward these copies to the invoked bricks, while the original object is forwarded to `__parallel_invoke`. What do you think is problematic here?

So the problem really have place.
And e reproduced it:
- the PR https://github.com/uxlfoundation/oneDPL/pull/2374
- the examples of compile errors: https://github.com/uxlfoundation/oneDPL/actions/runs/16833734286/job/47687785929?pr=2374

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3397,12 +3397,11 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
         __par_backend::__parallel_invoke(
             __backend_tag{}, ::std::forward<_ExecutionPolicy>(__exec),
             [=] {
-                __internal::__pattern_walk2_brick(__tag, ::std::forward<_ExecutionPolicy>(__exec), __first1, __last1,
-                                                  __result, __copy_range);
+                __internal::__pattern_walk2_brick(__tag, std::move(__exec), __first1, __last1, __result, __copy_range);
```
SergeyKopienko: Probably `std::move(__exec)` will give us compile error here in some cases.
I propose simple to pass `__exec` as is.

danhoeflinger: Why would it give problems?  We are capturing a copy when this lambda is created, we cannot move it?

danhoeflinger:
As mentioned in offline discussion with @SergeyKopienko, the reason this did not appear as a build error in our existing test suite is because the nature of our "canned" host policies like `seq`, or `par_unseq` are global `constexpr` instances.  This means that these are passed / around as `const lvalue ref`.  There is no build error for dropping const-ness in the forward.  

If someone copies these policies to a non-const instance before calling an API, we would see the issue (before the fixes in this PR).

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: For experiment I merged this PR into https://github.com/uxlfoundation/oneDPL/pull/2374 - let's check will we have compile errors here or no...
Let's see the results in https://github.com/uxlfoundation/oneDPL/pull/2374/checks

SergeyKopienko: Looks this variant compiled without errors, so I agree with this change.

SergeyKopienko: Let's resolve this conversation.

SergeyKopienko (approved):
LGTM

PR #2374, opened 2025-08-08
Expand tests coverage for not only constexpr policies

SergeyKopienko:
Together with @danhoeflinger we found the way how to detect compile errors in oneDPL code for host policies.

The main problems in our test coverage was that tests always had used `constexpr` host execution policies: `constexpr` has some special aspects of capturing into lambda by value so we can forward them inside lambdas!

So we hope to found some compile errors in oneDPL code by this extended test coverage and then to fix them.

### DIFF
--- a/test/support/utils_invoke.h
+++ b/test/support/utils_invoke.h
@@ -283,10 +283,36 @@ struct invoke_on_all_host_policies
         using namespace oneapi::dpl::execution;
 
 #if !TEST_ONLY_HETERO_POLICIES
+
         // Try static execution policies
         invoke_on_all_iterator_types()(seq,       op, rest...);
         invoke_on_all_iterator_types()(unseq,     op, rest...);
         invoke_on_all_iterator_types()(par,       op, rest...);
+
+        volatile bool always_false = false;
+        if (always_false)
+        {
+            oneapi::dpl::execution::sequenced_policy   user_sequenced_policy;
+            oneapi::dpl::execution::unsequenced_policy user_unsequenced_policy;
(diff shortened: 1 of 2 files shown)


test/support/utils_invoke.h
```diff
@@ -283,18 +283,57 @@ struct invoke_on_all_host_policies
...
+            invoke_on_all_iterator_types()(user_sequenced_policy,   op, rest...);
+            invoke_on_all_iterator_types()(user_unsequenced_policy, op, rest...);
+            invoke_on_all_iterator_types()(user_parallel_policy,    op, rest...);
+
+            const auto& user_sequenced_policy_const_ref   = user_sequenced_policy;
+            const auto& user_unsequenced_policy_const_ref = user_unsequenced_policy;
+            const auto& user_parallel_policy_const_ref    = user_parallel_policy;
+            invoke_on_all_iterator_types()(user_sequenced_policy_const_ref,   op, rest...);
+            invoke_on_all_iterator_types()(user_unsequenced_policy_const_ref, op, rest...);
+            invoke_on_all_iterator_types()(user_parallel_policy_const_ref,    op, rest...);
```
danhoeflinger, line 309: I think this one may be redundant with the existing executed checks (from a type perspective).

SergeyKopienko: It check only compilation, so no additional execution time for run tests, only compilation time will be increased.

danhoeflinger: sure, its not a problem especially if we are not keeping the code but rather just using it to look for issues now.

SergeyKopienko: Agree.

PR #2376, opened 2025-08-11
Refactoring of  `__is_iterator_type`

SergeyKopienko:
This PR updates the `__is_iterator_type` template specialization to replace `std::decay_t` with `std::remove_reference_t` for better compatibility with different C++ standards.

- Replaces `std::decay_t<_T>` with `std::remove_reference_t<_T>` in the iterator type detection logic

### DIFF
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -872,8 +872,8 @@ struct __is_type_with_iterator_traits : std::false_type
 };
 
 template <typename _T>
-struct __is_type_with_iterator_traits<_T, std::void_t<typename std::iterator_traits<std::remove_reference_t<_T>>::difference_type>>
-    : std::true_type
+struct __is_type_with_iterator_traits<
+    _T, std::void_t<typename std::iterator_traits<std::remove_reference_t<_T>>::difference_type>> : std::true_type
 {
 };
 


include/oneapi/dpl/pstl/utils.h
```diff
@@ -871,11 +871,19 @@ struct __is_iterator_type : std::false_type
...
     : std::true_type
 {
 };
+#else
+template <typename _T>
+struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<std::remove_cvref_t<_T>>::difference_type>>
+    : std::true_type
+{
+};
+#endif // _ONEDPL___cplusplus < 202002L
```
akukanov: I think we should have a single definition of the trait for C++17 and later standards. I do not see how cv qualifiers would change the trait value - a `const` iterator is still an iterator, right?

So I suggest to simply replace `std::decay_t` (which does way more than needed) with `std::remove_reference_t`.

dmitriy-sobolev: > a const iterator is still an iterator, right?

It seems it's not due to `std::movable`, see https://godbolt.org/z/vPza71fP5.

SergeyKopienko: So let's introduce
```C++
template <typename _Iterator>
using __remove_cvref_t
#if _ONEDPL___cplusplus < 202002L
    = std::remove_cv_t<std::remove_reference_t<_Iterator>>;
#else
    = std::remove_cvref_t<_Iterator>;
#endif
```
and then we can rewrite:
```C++
template <typename _T, typename = void>
struct __is_iterator_type : std::false_type
{
};

template <typename _T>
struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<__remove_cvref_t<_T>>::difference_type>>
    : std::true_type
{
};
```

akukanov: Let's put it into the context though.

We are not developing a substitution for C++20 concepts, We already have a simple, very rough utility that only checks if `std::iterator_traits<_T>::difference_type` is well defined. In Dmitry's test, this implementation returns `1 0 1` - i.e., it already allows const types to pass, but not reference types. As far as I understand, Sergey wants to simplify usage and for that he added `decay_t` into the definition. With that, the test returns `1 1 1`. If `decay_t` is changed to `remove_reference_t`, the test also returns `1 1 1`.

If we do not want const iterators to pass the test, we need to change the implementation, not just play with removing references and qualifiers. I stay convinced that for Sergey's usability goal `remove_reference_t` is sufficient and should be used for all versions of the standard.

SergeyKopienko: Fixed:
```C++
template <typename _T>
struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<std::remove_reference_t<_T>>::difference_type>>
    : std::true_type
{
};
```

include/oneapi/dpl/pstl/utils.h
```diff
@@ -872,7 +872,7 @@ struct __is_iterator_type : std::false_type
 };
 
 template <typename _T>
-struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<std::decay_t<_T>>::difference_type>>
+struct __is_iterator_type<_T, std::void_t<typename std::iterator_traits<std::remove_reference_t<_T>>::difference_type>>
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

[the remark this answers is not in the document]
SergeyKopienko: declined

include/oneapi/dpl/pstl/utils.h
akukanov: Thanks.
I think we should also rename the thing to avoid confusion. I would name it `__has_iterator_traits` or `__is_type_with_iterator_traits` to better reflect what it actually does.

SergeyKopienko: Done.

akukanov (dismissed):

SergeyKopienko:
@akukanov Clang format has been applied, could you please re-approve this PR ?

Issue #2380, opened 2025-08-12
Required to replace `static constexpr` to `inline constexpr` for global variable declarations


SergeyKopienko:
https://github.com/uxlfoundation/oneDPL/blob/63f87303db791494ab17fd6bcadacc0495a3c13c/include/oneapi/dpl/pstl/utils.h#L881

The source: https://github.com/uxlfoundation/oneDPL/pull/2371#discussion_r2267898442

Additional details: https://stackoverflow.com/questions/71699928/inline-static-constexpr-vs-global-inline-constexpr

And this is not all: https://github.com/search?q=repo%3Auxlfoundation%2FoneDPL+%22static+constexpr+bool+__%22&type=code

I don't know want we fix all together or separately.

SergeyKopienko:
PR prepared: https://github.com/uxlfoundation/oneDPL/pull/2381

SergeyKopienko:
No errors,  this code is correct:
```C++
static constexpr bool __is_iterator_type_v = __is_iterator_type<_T>::value; 
```

PR #2381, opened 2025-08-12
Replace `static constexpr` to `inline constexpr` for global variables and `class` / `struct` fields

SergeyKopienko:


dmitriy-sobolev:
> This PR modernizes C++ constexpr variable declarations by replacing static constexpr with inline constexpr for global variables and class / struct fields. This improves compliance with modern C++ standards where inline is preferred for constexpr variables that may be used across multiple translation units.


I do not understand why we should prefer `inline` to `static` for implementation details. `inline` exposes a symbol in the ABI. I suppose it's better to keep implementation details hidden.

I do not have a good grasp in this question, but I think this PR with https://github.com/uxlfoundation/oneDPL/issues/2380 are good places to clarify it.

danhoeflinger:
As I understand it, `constexpr static` member fields are implicitly `inline` as of c++17.  
It doesn't hurt to explicitly state it for clarity, but I don't know how much it adds in practice.

SergeyKopienko:
> > I do not have a good grasp in this question, but I think this PR with #2380 are good places to clarify it. What is the underlying reason to replace `static` with `inline`?
> 
> My understanding is that this is because of the ODR when compiling for multiple translation units, `static` has issues here whereas `inline` does not (for variables not member fields).

I think so too.

SergeyKopienko:
Not required because we haven't real error here.

PR #2384, opened 2025-08-17
Using existing functional objects instead lambdas in template code

SergeyKopienko:
This PR replaces lambda expressions with functional objects in the algorithm ranges implementation to resolve kernel name duplication errors when execution policies are passed by reference. The changes systematically replace inline lambda definitions with predefined `__unary_op` and `__binary_op` functional objects.

Key changes:
- Replace lambda expressions with `oneapi::dpl::__internal::__unary_op` and `oneapi::dpl::__internal::__binary_op` functional objects
- Apply changes consistently across multiple algorithm implementations


SergeyKopienko:
@dmitriy-sobolev I all tests passed, you may merge this PR into any other existing range PRs if in needed. Or we can merge this PR separatelly.

akukanov:
If these changes are not strongly needed, should we make them or should we better postpone? I prefer to postpone, targeting the .11 milestone.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -46,11 +46,10 @@ __pattern_for_each(_Tag __tag, _ExecutionPolicy&& __exec, _R&& __r, _Fun __f, _P
 {
     static_assert(__is_parallel_tag_v<_Tag> || typename _Tag::__is_vector{});
 
-    auto __f_1 =
-        [__f, __proj](auto&& __val) { std::invoke(__f, std::invoke(__proj, std::forward<decltype(__val)>(__val)));};
+    oneapi::dpl::__internal::__unary_op<_Fun, _Proj> __pred_1{__f, __proj};
```
dmitriy-sobolev: Please, keep the original names. 

Here, for example, `__f_1` suits better - a predicate is something what returns true or false. Also, the c++ standard names these entities. It's `f` for for_each, for example.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -129,8 +128,7 @@ __pattern_find_if(_Tag __tag, _ExecutionPolicy&& __exec, _R&& __r, _Pred __pred,
 {
     static_assert(__is_parallel_tag_v<_Tag> || typename _Tag::__is_vector{});
 
-    auto __pred_1 = [__pred, __proj](auto&& __val) {
-        return std::invoke(__pred, std::invoke(__proj, std::forward<decltype(__val)>(__val)));};
+    oneapi::dpl::__internal::__unary_op<_Pred, _Proj> __pred_1{__pred, __proj};
```
dmitriy-sobolev, line 128: We also have `oneapi::dpl::__internal::__predicate`, which is a slightly better candidate due to its expressiveness . I do not insist on using it - the current `__unary_op` is the direct substitution to the old lambda.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
SergeyKopienko: Disagree: the `__unary_op` and `__predicate` structures doing the different things inside:
```C++
template <typename _F, typename _Proj>
struct __unary_op
{
    //'mutable' is to relax the requirements for a user functor or/and projection type operator() may be non-const
    mutable _F __f;
    mutable _Proj __proj;

    template <typename _TValue>
    decltype(auto)
    operator()(_TValue&& __val) const
    {
        return std::invoke(__f, std::invoke(__proj, std::forward<_TValue>(__val)));
    }
};
```
- it's apply projection only **for single argument**.

```C++
template <typename _Pred, typename _Proj>
struct __predicate
{
    //'mutable' is to relax the requirements for a user comparator or/and projection type operator() may be non-const
    mutable _Pred __pred;
    mutable _Proj __proj;

    template <typename... _Xp>
    bool
    operator()(_Xp&&... __x) const
    {
        return std::invoke(__pred, std::invoke(__proj, std::forward<_Xp>(__x))...);
    }
};
```
- it's apply projection **for each argument**.

So I believe better to use exactly the `__unary_op` structure.

dmitriy-sobolev: Ah, right.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
dmitriy-sobolev: Could you also fix other places like that? For example 74th line in this file.

SergeyKopienko: Fixed: restored old names of predicates

PR #2390, opened 2025-08-22
Improve reporting of issues in the range algorithm tests

dmitriy-sobolev:
Reporting Improvements:
- `EXPECT_EQ` will provide the current and the expected values instead of just true/false with `EXPECT_TRUE`.
- `bool` values will be reported as true/false.

Additional modernization/safety improvement:
- `nullptr` instead of `NULL`.

These changes were added by @SergeyKopienko during the development of some range-based algorithms, and were shown as quite useful while analyzing issues.


SergeyKopienko (approved):
LGTM

PR #2391, opened 2025-08-22
Pass views by value to make_zip_view

dmitriy-sobolev:
This is a draft to investigate the purpose better. This change was initially added in #2320, but it does not seem related, so it was moved into a separate PR.

The assumed purpose is that [zip_view ](https://en.cppreference.com/w/cpp/ranges/zip_view/zip_view.html)'s constructor expects views as input types, but `make_zip_view` may preserve a reference of a view type due to the perfect forwarding, which is not a view anymore. See https://godbolt.org/z/s8zv1c756 as an example of an effect of preserving a reference. Views are lightweight objects, so let's pass them by copy.

### DIFF
--- a/include/oneapi/dpl/pstl/utils_ranges.h
+++ b/include/oneapi/dpl/pstl/utils_ranges.h
@@ -281,9 +281,9 @@ class zip_view
 
 template <typename... _Ranges>
 auto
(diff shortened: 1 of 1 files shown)


danhoeflinger:
Thanks for dealing with this.  I found this issue in the past, but did not fully understand it. closed #1805.

PR #2392, opened 2025-08-22
Match template and argument order in range for_each pattern

dmitriy-sobolev:
Additionally, some clang-format is applied for better readability.


SergeyKopienko (dismissed):
LGTM

SergeyKopienko (approved):
LGTM

Issue #2394, opened 2025-08-25
Incorrect return types in `oneapi::dpl::__internal::__ranges::__pattern_minmax_element` and `oneapi::dpl::__internal::__ranges::__pattern_mismatch`


SergeyKopienko:
I believe we have incorrect return types in two range-patterns:
- `oneapi::dpl::__internal::__ranges::__pattern_minmax_element`
- `oneapi::dpl::__internal::__ranges::__pattern_mismatch`

The definition:
```C++
template <typename _BackendTag, typename _ExecutionPolicy, typename _R, typename _Proj, typename _Comp>
std::pair<std::ranges::iterator_t<_R>, std::ranges::iterator_t<_R>>
__pattern_minmax_element(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R&& __r, _Comp __comp,
                         _Proj __proj)
```

```C++
template <typename _BackendTag, typename _ExecutionPolicy, typename _R1, typename _R2, typename _Pred, typename _Proj1,
          typename _Proj2>
std::pair<std::ranges::iterator_t<_R1>, std::ranges::iterator_t<_R2>>
__pattern_mismatch(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _R1&& __r1, _R2&& __r2, _Pred __pred,
                   _Proj1 __proj1, _Proj2 __proj2)
```

Probably the `std::ranges::borrowed_iterator_t` should be used in return type.

SergeyKopienko:
@akukanov how do you think?

akukanov:
https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst is the source of truth, and also the standard or cppreference - for these algorithms we have the same return types as in C++.

MikeDvorskiy:
Basically, on high level API we have right return type, with `borrowed_iterator_t<_R>`, according to C++ standard.
It defends against dangling range / dangling iterator in case passing a temporary range into oneAPI algo. 

```
std::ranges::minmax_element_result<std::ranges::borrowed_iterator_t<_R>>
    operator()(_ExecutionPolicy&& __exec, _R&& __r, _Comp __comp = {}, _Proj __proj = {}) const
```
As far as I know the implementation (pattern layer and backend) doesn't create and pass further a temporary range, at least for mentioned couple of algorithms...

So, the proposed changed doesn't have significant needs.

[...]

PR #2395, opened 2025-08-25
Fix return types in `oneapi::dpl::__internal::__ranges::__pattern_minmax_element` and `oneapi::dpl::__internal::__ranges::__pattern_mismatch`

SergeyKopienko:
This PR fix the return types of two pattern functions in the oneAPI DPL library to use `borrowed_iterator_t` instead of `iterator_t` for C++20 ranges compliance.

- Updated return types from `std::ranges::iterator_t` to `std::ranges::borrowed_iterator_t` in two function signatures
- Ensures proper iterator borrowing semantics for range-based algorithms

From issue #2394


dmitriy-sobolev (approved):
LGTM.

There is no need for a conversion from iterator_t to borrowed_iterator_t. From the functional perspective, I see no impact after this experiment https://godbolt.org/z/zGv8a87fb.

PR #2397, opened 2025-08-25
Real fix for mangled name comma operator deletion

danhoeflinger:
The previous fix #2393 had issues with where a kernel name could clash between this level and the subsequent call's policy clones.  Since not all calls on this level of the callstack were wrapped, the kernel name could match between this cloned policy and a policy cloned at the next level of the callstack.  

This is a better fix which clones all calls at this level, causing them to be explicitly unique with each other with the same level of name nesting.

This PR also rearranges the order to group together the paired tests which are supposed to check that different value category policy calls don't clash with each other.

The previous fix resolved most issues but a few tests had remaining failures due to this sort of clashing.

### DIFF
--- a/test/support/utils_invoke.h
+++ b/test/support/utils_invoke.h
@@ -371,6 +371,9 @@ struct invoke_on_all_hetero_policies
     {
         auto my_policy = get_dpcpp_test_policy<CallNumber, Op>();
 
+        auto cloned_policy0 = CLONE_TEST_POLICY_IDX(my_policy, 0);
+        auto cloned_policy1 = CLONE_TEST_POLICY_IDX(my_policy, 1);
+
         sycl::queue queue = my_policy.queue();
 
         // Device may not support some types, e.g. double or sycl::half; test if they are supported or skip otherwise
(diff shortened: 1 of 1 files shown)


test/support/utils_invoke.h
```diff
@@ -383,22 +386,21 @@ struct invoke_on_all_hetero_policies
...
-            // Check compilation of the kernel with comma operator deleted iterators. Use policy wrapped with unique
-            // identifier to differentiate from other calls, but, preserve the policy for other two calls.
-            TestUtils::check_compilation_no_comma(CLONE_TEST_POLICY_IDX(my_policy, 0), op, rest...);
-#endif
+                cloned_policy0, op, std::forward<Args>(rest)...);
 
 #if TEST_CHECK_COMPILATION_WITH_DIFF_POLICY_VAL_CATEGORY
             // Check compilation of the kernel with different policy type qualifiers
-            check_compilation(my_policy, [&](auto&& __policy) {
+            check_compilation(cloned_policy0, [&](auto&& __policy) {
```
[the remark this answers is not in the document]
danhoeflinger: This intentionally uses the same cloned policy for this line because it is intended to be the same as the call above it.  This is the point of this test, to check that different value category of policy can be used without issue.

test/support/utils_invoke.h
```diff
@@ -383,22 +386,21 @@ struct invoke_on_all_hetero_policies
             // performs some checks that fail. As a workaround, define for functors which have this issue
             // __functor_type(see kernel_type definition) type field which doesn't have any pointers in it's name.
             iterator_invoker<std::random_access_iterator_tag, /*IsReverse*/ std::false_type>()(
-                my_policy, op, std::forward<Args>(rest)...);
-
-
-#if TEST_CHECK_COMPILATION_WITH_COMMA_OP_DELETED_ITERS
-            // Check compilation of the kernel with comma operator deleted iterators. Use policy wrapped with unique
-            // identifier to differentiate from other calls, but, preserve the policy for other two calls.
-            TestUtils::check_compilation_no_comma(CLONE_TEST_POLICY_IDX(my_policy, 0), op, rest...);
```
mmichel11, line 392: For my understanding, this could cause issues because in line 386 we pass `my_policy` directly and the test may wrap a policy similar to  `CLONE_TEST_POLICY_IDX(my_policy, 0)`?

danhoeflinger: Yes, exactly.  Its sneaky, but we need all calls at the same level of the call stack to be wrapped with the same recursive level of cloning if we are just using numbered policy clones. If we were providing some unique name or something, that would be different.

mmichel11 (approved):
Good catch, LGTM

PR #2398, opened 2025-08-25
Implement P2248R8: Enable list-initialization for algorithms

rarutyun:
The [link](http://wg21.link/p2248) to the proposal

Implementing Iterator-based algorithms only. Ranges-based implementation is going on separately.

There are no breaking changes. All the breaking changes in C++ standard come into `std::ranges` namespace but even they are irrelevant to oneDPL because we implement list-initialization enabling right away, so there is no template parameters swapping

There are two oversights for P2248:
- `find_last` was not included (see [this paper](http://wg21.link/p3217)). However, it's irrelevant for Iterator-based algorithms because `find_last` exists in `std::ranges:` namespace only
- `uninitialized_fill` was not included (see [this paper](http://wg21.link/p3787). However, it's not included for C++26. We are trying to achieve that but there is no guarantee. The question is do we want to implement it proactively?

Binary Search-like algorithms from P2248 (e.g., `lower_bound`, `upper_bound`, etc.) are not relevant either because our API for those does not have `T`-like template parameter.

### DIFF
--- a/include/oneapi/dpl/pstl/glue_algorithm_defs.h
+++ b/include/oneapi/dpl/pstl/glue_algorithm_defs.h
@@ -64,7 +64,8 @@ template <class _ExecutionPolicy, class _ForwardIterator, class _Predicate>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator>
 find_if_not(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _Predicate __pred);
 
-template <class _ExecutionPolicy, class _ForwardIterator, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator,
+          class _Tp = typename std::iterator_traits<_ForwardIterator>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator>
 find(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, const _Tp& __value);
 
@@ -104,7 +105,8 @@ adjacent_find(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardItera
 
 // [alg.count]
 
-template <class _ExecutionPolicy, class _ForwardIterator, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator,
+          class _Tp = typename std::iterator_traits<_ForwardIterator>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<
     _ExecutionPolicy, typename ::std::iterator_traits<_ForwardIterator>::difference_type>
 count(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, const _Tp& __value);
@@ -126,12 +128,14 @@ oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _Forward
 search(_ExecutionPolicy&& __exec, _ForwardIterator1 __first, _ForwardIterator1 __last, _ForwardIterator2 __s_first,
        _ForwardIterator2 __s_last);
 
-template <class _ExecutionPolicy, class _ForwardIterator, class _Size, class _Tp, class _BinaryPredicate>
+template <class _ExecutionPolicy, class _ForwardIterator, class _Size,
+          class _Tp = typename std::iterator_traits<_ForwardIterator>::value_type, class _BinaryPredicate>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator>
 search_n(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _Size __count,
          const _Tp& __value, _BinaryPredicate __pred);
 
-template <class _ExecutionPolicy, class _ForwardIterator, class _Size, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator, class _Size,
+          class _Tp = typename std::iterator_traits<_ForwardIterator>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator>
 search_n(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _Size __count,
          const _Tp& __value);
@@ -189,33 +193,39 @@ transform_if(_ExecutionPolicy&& __exec, _ForwardIterator1 __first1, _ForwardIter
 
 // [alg.replace]
 
-template <class _ExecutionPolicy, class _ForwardIterator, class _UnaryPredicate, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator, class _UnaryPredicate,
+          class _Tp = typename std::iterator_traits<_ForwardIterator>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy>
 replace_if(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _UnaryPredicate __pred,
            const _Tp& __new_value);
 
-template <class _ExecutionPolicy, class _ForwardIterator, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator,
+          class _Tp = typename std::iterator_traits<_ForwardIterator>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy>
 replace(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, const _Tp& __old_value,
         const _Tp& __new_value);
 
-template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2, class _UnaryPredicate, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2, class _UnaryPredicate,
+          class _Tp = typename std::iterator_traits<_ForwardIterator2>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator2>
 replace_copy_if(_ExecutionPolicy&& __exec, _ForwardIterator1 __first, _ForwardIterator1 __last,
                 _ForwardIterator2 __result, _UnaryPredicate __pred, const _Tp& __new_value);
 
-template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2,
+          class _Tp = typename std::iterator_traits<_ForwardIterator2>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator2>
 replace_copy(_ExecutionPolicy&& __exec, _ForwardIterator1 __first, _ForwardIterator1 __last, _ForwardIterator2 __result,
              const _Tp& __old_value, const _Tp& __new_value);
 
 // [alg.fill]
 
-template <class _ExecutionPolicy, class _ForwardIterator, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator,
+          class _Tp = typename std::iterator_traits<_ForwardIterator>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy>
 fill(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, const _Tp& __value);
 
-template <class _ExecutionPolicy, class _ForwardIterator, class _Size, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator, class _Size,
+          class _Tp = typename std::iterator_traits<_ForwardIterator>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator>
 fill_n(_ExecutionPolicy&& __exec, _ForwardIterator __first, _Size __count, const _Tp& __value);
 
@@ -235,7 +245,8 @@ oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _Forward
 remove_copy_if(_ExecutionPolicy&& __exec, _ForwardIterator1 __first, _ForwardIterator1 __last,
                _ForwardIterator2 __result, _Predicate __pred);
 
-template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2,
+          class _Tp = typename std::iterator_traits<_ForwardIterator1>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator2>
 remove_copy(_ExecutionPolicy&& __exec, _ForwardIterator1 __first, _ForwardIterator1 __last, _ForwardIterator2 __result,
             const _Tp& __value);
@@ -244,7 +255,8 @@ template <class _ExecutionPolicy, class _ForwardIterator, class _UnaryPredicate>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator>
 remove_if(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, _UnaryPredicate __pred);
 
-template <class _ExecutionPolicy, class _ForwardIterator, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator,
+          class _Tp = typename std::iterator_traits<_ForwardIterator>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator>
 remove(_ExecutionPolicy&& __exec, _ForwardIterator __first, _ForwardIterator __last, const _Tp& __value);
 
--- a/test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
+++ b/test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
@@ -97,6 +97,85 @@ test_fill_by_type(::std::size_t n)
 #endif
 }
 
+void test_empty_list_initialization_for_fill()
+{
+    {
+        std::vector<int> v{3,6,5,4,3,7,8,0,2,4};
+        oneapi::dpl::fill(oneapi::dpl::execution::seq, v.begin(), v.end(), {});
+        EXPECT_TRUE(std::count(v.begin(), v.end(), 0) == v.size(), "a sequence is not filled properly by oneapi::dpl::fill with `seq` policy");
+    }
+    {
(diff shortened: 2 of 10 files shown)


test/support/utils.h
```diff
@@ -1300,6 +1300,17 @@ struct NoDefaultCtorWrapper {
     }
 };
 
+struct DefaultInitializedToOne
+{
+    int value = 1;
+
+    friend bool
+    operator==(const DefaultInitializedToOne& x, const DefaultInitializedToOne& y)
```
SergeyKopienko, line 1308: Shout this method be declared as `const` ?

danhoeflinger:
Is the intention to merge this for milestone 2022.10.0?

rarutyun:
> Is the intention to merge this for milestone 2022.10.0?

Whenever we think it's ready

include/oneapi/dpl/pstl/glue_algorithm_defs.h
```diff
@@ -189,33 +193,39 @@ transform_if(_ExecutionPolicy&& __exec, _ForwardIterator1 __first1, _ForwardIter
...
 
-template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2, class _UnaryPredicate, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2, class _UnaryPredicate,
+          class _Tp = typename std::iterator_traits<_ForwardIterator2>::value_type>
 oneapi::dpl::__internal::__enable_if_execution_policy<_ExecutionPolicy, _ForwardIterator2>
 replace_copy_if(_ExecutionPolicy&& __exec, _ForwardIterator1 __first, _ForwardIterator1 __last,
                 _ForwardIterator2 __result, _UnaryPredicate __pred, const _Tp& __new_value);
 
-template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2, class _Tp>
+template <class _ExecutionPolicy, class _ForwardIterator1, class _ForwardIterator2,
```
danhoeflinger: Doesn't the paper conclude not to make any changes to replace_copy for non ranges api?

test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
```diff
@@ -97,6 +97,85 @@ test_fill_by_type(::std::size_t n)
...
+        oneapi::dpl::fill(oneapi::dpl::execution::seq, v.begin(), v.end(), {});
+        EXPECT_TRUE(std::count(v.begin(), v.end(), 0) == v.size(), "a sequence is not filled properly by oneapi::dpl::fill with `seq` policy");
+    }
+    {
+        std::vector<int> v{3,6,5,4,3,7,8,0,2,4};
+        oneapi::dpl::fill(oneapi::dpl::execution::unseq, v.begin(), v.end(), {});
+        EXPECT_TRUE(std::count(v.begin(), v.end(), 0) == v.size(), "a sequence is not filled properly by oneapi::dpl::fill with `unseq` policy");
+    }
+
+    {
```
danhoeflinger, line 113: [nitpick] Remove extra unnecessary braces (this goes away if we use `invoke_on_all_policies`)

test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
```diff
@@ -97,6 +97,85 @@ test_fill_by_type(::std::size_t n)
 #endif
 }
 
+void test_empty_list_initialization_for_fill()
```
danhoeflinger, line 100: We have a way to run a test for all policy types as seen in this same test file via `invoke_on_all_policies` and a struct like `test_fill`.  Perhaps we should use that pattern to remove some redundancy.

test/parallel_api/algorithm/alg.modifying.operations/remove.pass.cpp
```diff
@@ -105,6 +105,54 @@ struct test_non_const
     }
 };
 
+void test_empty_list_initialization()
```
danhoeflinger, line 108: Same basic comment on using `invoke_on_all_policies` here (and throughout the rest of the changes).

test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
danhoeflinger: After offline discussion, it sounds like this is purposeful to reduce complexity and increase transparency for finding errors in a very simple test.  I'm OK with that even with some code replication.

test/parallel_api/algorithm/alg.modifying.operations/remove.pass.cpp
danhoeflinger: see above

test/parallel_api/algorithm/alg.modifying.operations/replace_copy.pass.cpp
danhoeflinger, line 1: As discussed offline, these tests (for `replace_copy`, but not `replace_copy_if`) should be removed along with the default template arg for replace_copy in the implementation.

test/parallel_api/algorithm/alg.modifying.operations/replace_copy.pass.cpp
```diff
@@ -109,6 +109,135 @@ struct test_non_const
...
+constexpr void test_default_template_argument_from_output_iterator()
+{
+    static_assert(is_replace_copy_well_formed<std::vector<not_implicitly_convertible>::iterator, std::vector<int>::iterator>::value,
+                  "Positive: The default argument of replace_copy shall be taken from output iterator");
+    static_assert(!is_replace_copy_well_formed<std::vector<int>::iterator, std::vector<not_implicitly_convertible>::iterator>::value,
+                  "Negative: The default argument of replace_copy shall be taken from output iterator");
+    static_assert(is_replace_copy_if_well_formed<std::vector<not_implicitly_convertible>::iterator, std::vector<int>::iterator>::value,
+                  "Positive: The default argument of replace_copy_if shall be taken from output iterator");
+    static_assert(!is_replace_copy_if_well_formed<std::vector<int>::iterator, std::vector<not_implicitly_convertible>::iterator>::value,
+                  "Positive: The default argument of replace_copy_if shall be taken from output iterator");
```
danhoeflinger: The "Positive" and "Negative" here are confusing for me (and the one on like 150 should be "Negative" I think).

I'd prefer something more like:

"replace_copy_if call is *NOT* well-formed for valid list_initialization argument"
and
"replace_copy_if call is well-formed for *INVALID* list_initialization argument" 

Or something similar so that it is clear to the user reading the error message what has gone wrong without looking at the tests.

test/parallel_api/algorithm/alg.nonmodifying/find.pass.cpp
```diff
@@ -71,19 +71,50 @@ class Weird
...
     test<Number>(Weird(42, OddTag()), [](std::int32_t) { return Number(42, OddTag()); }, // hit
-                 [](std::int32_t j) { return Number(j == 42 ? 0 : j, OddTag()); });      // miss
-#endif
+    [](std::int32_t j) { return Number(j == 42 ? 0 : j, OddTag()); });      // miss
+    #endif
 
     // Test with value that is equal to two different bit patterns (-0.0 and 0.0)
     test<float32_t>(-0.0, [](std::int32_t j) { return j & 1 ? 0.0 : -0.0; }, // hit
-                    [](std::int32_t j) { return j == 0 ? ~j : j; });         // miss
+    [](std::int32_t j) { return j == 0 ? ~j : j; });         // miss
```
danhoeflinger: It seems like these are bad formatting changes that should be reverted.

test/parallel_api/algorithm/alg.nonmodifying/search_n.pass.cpp
```diff
@@ -96,6 +96,52 @@ struct test_non_const
...
+        }
+        {
+            auto it = oneapi::dpl::search_n(oneapi::dpl::execution::par_unseq, v_custom.begin(), v_custom.end(), 4, {}, std::equal_to{});
+            EXPECT_TRUE(it == (v_custom.begin() + 5), "an empty list-initialized value is not found by oneapi::dpl::search_n with `par_unseq` policy with predicate");
+        }
+    }
+#if TEST_DPCPP_BACKEND_PRESENT
+    sycl::buffer<int> buf(v);
+    auto it = oneapi::dpl::search_n(oneapi::dpl::execution::dpcpp_default, oneapi::dpl::begin(buf), oneapi::dpl::end(buf), {});
+    EXPECT_TRUE(it.get_idx() == 4, "an empty list-initialized value is not found by oneapi::dpl::search_n with `device_policy` policy");
```
danhoeflinger, line 141: [nitpick] Is there a reason to only test the integer variant on `dpcpp_default` rather that both integer and `DefaultInitializedToOne`?  

I know that from a white box perspective, all these policies are really testing the same part of the code as they share code at the public API level, but for the other policy types we test both.

Same question for the other tests as well.

include/oneapi/dpl/pstl/glue_algorithm_defs.h
rarutyun: Yes, the original proposal, which is P2248, concludes exactly that. Good catch! Thanks.

test/parallel_api/algorithm/alg.modifying.operations/replace_copy.pass.cpp
rarutyun: I've done it in a different way. I hope the messages look good the way I've done it.

include/oneapi/dpl/pstl/glue_algorithm_defs.h
rarutyun: Done

test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
rarutyun: done. I hope I didn't miss anything

test/parallel_api/algorithm/alg.modifying.operations/replace_copy.pass.cpp
rarutyun: I removed runtime tests for `replace_copy` but kept compile-time ones to make sure that if somebody adds those later by mistake, the test will fail.

test/parallel_api/algorithm/alg.nonmodifying/find.pass.cpp
rarutyun: Fully agreed. Also spotted that one.

Reverted.

test/parallel_api/algorithm/alg.nonmodifying/search_n.pass.cpp
rarutyun: We don't use `int` type and a custom type for all policies. It's just a random combination of policies and types.

I do believe that the question is still important, though because we also have two different overloads for `search_n` (with and without comparator). But after offline discussion we decide not to treat `device_policy` specially, so we will keep one call without explicit comparator for `device_policy` policy type.

test/parallel_api/algorithm/alg.modifying.operations/remove_copy.pass.cpp
```diff
@@ -96,6 +96,84 @@ test(T trash, const T& value, Convert convert, bool check_weakness = true)
...
+                                                                                std::declval<It>(),
+                                                                                std::declval<DestIt>(),
+                                                                                {3}))>> : std::true_type {};
+
+constexpr void test_default_template_argument_from_output_iterator()
+{
+    static_assert(is_remove_copy_well_formed<std::vector<int>::iterator, std::vector<not_implicitly_convertible>::iterator>::value,
+                  "The default template argument for list-initialization of remove_copy is NOT a value_type of the input iterator");
+    static_assert(!is_remove_copy_well_formed<std::vector<not_implicitly_convertible>::iterator, std::vector<int>::iterator>::value,
+                  "The default template argument for list-initialization of remove_copy is shall be a value_type of the input iterator");
```
danhoeflinger:
```suggestion
                  "The default template argument for list-initialization of remove_copy must be a value_type of the input iterator");
```

test/parallel_api/algorithm/alg.modifying.operations/replace_copy.pass.cpp
```diff
@@ -109,6 +109,89 @@ struct test_non_const
...
+constexpr void test_default_template_argument()
+{
+    static_assert(!is_replace_copy_well_formed<std::vector<not_implicitly_convertible>::iterator, std::vector<int>::iterator>::value,
+                  "Input iterator value_type test: std::replace_copy shall NOT have any default template argument for list-initialization");
+    static_assert(!is_replace_copy_well_formed<std::vector<int>::iterator, std::vector<not_implicitly_convertible>::iterator>::value,
+                  "Output iterator value_type test: std::replace_copy shall NOT have any default template argument for list-initialization");
+    static_assert(is_replace_copy_if_well_formed<std::vector<not_implicitly_convertible>::iterator, std::vector<int>::iterator>::value,
+                  "The default template argument for list-initialization of replace_copy_if is NOT a value_type of the output iterator");
+    static_assert(!is_replace_copy_if_well_formed<std::vector<int>::iterator, std::vector<not_implicitly_convertible>::iterator>::value,
+                  "The default template argument for list-initialization of replace_copy_if shall be a value_type of the output iterator");
```
danhoeflinger:
```suggestion
    static_assert(!is_replace_copy_well_formed<std::vector<not_implicitly_convertible>::iterator, std::vector<int>::iterator>::value,
                  "Input iterator value_type test: std::replace_copy must NOT have any default template argument for list-initialization");
    static_assert(!is_replace_copy_well_formed<std::vector<int>::iterator, std::vector<not_implicitly_convertible>::iterator>::value,
                  "Output iterator value_type test: std::replace_copy must NOT have any default template argument for list-initialization");
    static_assert(is_replace_copy_if_well_formed<std::vector<not_implicitly_convertible>::iterator, std::vector<int>::iterator>::value,
                  "The default template argument for list-initialization of replace_copy_if is NOT a value_type of the output iterator");
    static_assert(!is_replace_copy_if_well_formed<std::vector<int>::iterator, std::vector<not_implicitly_convertible>::iterator>::value,
                  "The default template argument for list-initialization of replace_copy_if must be a value_type of the output iterator");
```

This would be my choice of word in the form of an error message.  I would expect "must" in an error where this was not satisfied, and "shall" more in the context of an order to someone forcing them to do something in the future.  

They are only very subtly different in a way that's difficult to explain, but I was originally a little confused by "shall" in this context. (I wont insist)

test/parallel_api/algorithm/alg.modifying.operations/remove_copy.pass.cpp
danhoeflinger: Beyond the "shall" vs "must" choice, there is an extra "is" here.

rarutyun: Done.

test/parallel_api/algorithm/alg.modifying.operations/replace_copy.pass.cpp
rarutyun: Done

test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
```diff
@@ -97,6 +97,87 @@ test_fill_by_type(::std::size_t n)
...
+    {
+        std::vector<int> v{3,6,5,4,3,7,8,0,2,4};
+        oneapi::dpl::fill(oneapi::dpl::execution::unseq, v.begin(), v.end(), {});
+        EXPECT_TRUE(std::count(v.begin(), v.end(), 0) == v.size(), "a sequence is not filled properly by oneapi::dpl::fill with `unseq` policy");
+    }
+
+    {
+        std::vector<TestUtils::DefaultInitializedToOne> v_custom{{3},{6},{5},{4},{3},{7},{8},{2},{1},{4}};
+        oneapi::dpl::fill(oneapi::dpl::execution::par, v_custom.begin(), v_custom.end(), {});
+        EXPECT_TRUE(std::count(v_custom.begin(), v_custom.end(), TestUtils::DefaultInitializedToOne{}) == v_custom.size(),
```
SergeyKopienko, line 116: I propose to have `expected` state of vector here and just compare result with it.
The same comment for all other places.

test/parallel_api/algorithm/alg.modifying.operations/remove.pass.cpp
```diff
@@ -105,6 +105,56 @@ struct test_non_const
...
+        v.erase(it, v.end());
+        EXPECT_TRUE(v == expected, "wrong effect from calling oneapi::dpl::remove with empty list-initialized value and with `seq` policy");
+    }
+    {
+        std::vector<int> v{3,6,0,4,0,7,8,0,3,4};
+        std::vector<int> expected{3,6,4,7,8,3,4};
+        auto it = oneapi::dpl::remove(oneapi::dpl::execution::unseq, v.begin(), v.end(), {});
+        EXPECT_TRUE(it == v.begin() + 7, "not all empty list-initialized values are properly removed by oneapi::dpl::remove with `unseq` policy");
+        v.erase(it, v.end());
+        EXPECT_TRUE(v == expected, "wrong effect from calling oneapi::dpl::remove with empty list-initialized value and with `unseq` policy");
```
SergeyKopienko, line 124: We have special test macro `EXPECT_EQ_N` for containers comparisons like that.
The current implementation on the case of any problems doesn't shown us what is wrong inside.
The same comment for all other vector's checks from this PR.

test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
SergeyKopienko: But I don't insist, up to you.

test/parallel_api/algorithm/alg.modifying.operations/remove.pass.cpp
SergeyKopienko: But I don't insist, up to you.

akukanov:
> This mean we allows to user to have something code with {} in the parameters for changed methods which will ill-formed with other standard library implementations for C++17 and C++20.

Correct. Users who prefer to keep their source code base compatible with the current and previous versions of the C++ standard should not use braced-list arguments for the standard algorithms. Even after C++26 is published and implemented, previous (already released) versions of the standard libraries won't support this ability. oneDPL will still support C++17/20/23, meeting (most of) the requirements for parallel algorithms in these standards. The change does not introduce incompatibilities there.

SergeyKopienko (approved):
LGTM

test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
rarutyun: Yeah, rerunning nightly tests take significant amount of time, so I would prefer to withhold from cosmetic changes.

test/parallel_api/algorithm/alg.modifying.operations/remove.pass.cpp
rarutyun: Yeah, rerunning nightly tests take significant amount of time, so I would prefer to withhold from cosmetic changes.

test/parallel_api/algorithm/alg.modifying.operations/fill.pass.cpp
SergeyKopienko: Ok

test/parallel_api/algorithm/alg.modifying.operations/remove.pass.cpp
SergeyKopienko: ok

PR #2399, opened 2025-08-26
Disable No Comma operator tests for libc++

danhoeflinger:
The `std::stable_sort` API called without an execution policy has an issue which our no comma iterator adapter detects.  It results in a build error for the test, but it is a problem with libc++, not oneDPL.  Here we detect the presence of libc++ and disable the no comma tests.  

It is a bit of overkill to disable all tests when it is only `std::stable_sort`.  However, we get plenty of coverage to detect issues in oneDPL with the no comma operator tests with all other cases.  We could even be more restrictive and only enable these tests in limited scenarios as they are preventing regression of a pretty obscure potential bug.

### DIFF
--- a/test/support/test_config.h
+++ b/test/support/test_config.h
@@ -293,8 +293,12 @@
 #    error "TEST_ONLY_HETERO_POLICIES is passed but device backend is not available"
 #endif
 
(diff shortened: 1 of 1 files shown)


mmichel11 (approved):
I think the solution of disabling this test with libc++ makes sense since we have sufficient coverage elsewhere. LGTM.

PR #2403, opened 2025-08-27
Fix `static_assert` in `test_dangling_pointers`

SergeyKopienko:
This PR fixes a compilation issue with an unconditional `static_assert(false, ...)` statement in the `test_dangling_pointers` function. The change replaces the problematic `else static_assert(false, ...)` clause with a conditional `static_assert` that validates the `ArgsSize` parameter at compile time.

- Replaces unconditional `static_assert(false, ...)` with conditional assertion
- Moves the assertion to the beginning of the function for earlier validation

### DIFF
--- a/test/parallel_api/ranges/std_ranges_test.h
+++ b/test/parallel_api/ranges/std_ranges_test.h
@@ -317,7 +317,8 @@ private:
     constexpr void
     test_dangling_pointers(Policy&& exec, Algo&& algo, Args&& ...args)
     {
-        static_assert(ArgsSize == 1 || ArgsSize == 2 || ArgsSize == 3, "Test for amount of args not implemented");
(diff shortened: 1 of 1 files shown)


test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -317,6 +317,8 @@ struct test
     constexpr void
     test_dangling_pointers(Policy&& exec, Algo&& algo, Args&& ...args)
     {
+        static_assert(ArgsSize == 1 || ArgsSize == 2 || ArgsSize == 3, "Test for amount of args not implemented");
```
dmitriy-sobolev:
```suggestion
        static_assert(ArgsSize == 1 || ArgsSize == 2 || ArgsSize == 3,
                      "The test for dangling pointers is not implemented for this number of algorithm arguments");
```

PR #2405, opened 2025-08-28
Clarify new predicate return types

SergeyKopienko:
This PR has been created just for checking per-commit CI for https://github.com/uxlfoundation/oneDPL/pull/2406


SergeyKopienko:
Let's continue in https://github.com/uxlfoundation/oneDPL/pull/2406

PR #2408, opened 2025-08-28
[Test] Restrict no comma operator test builds

danhoeflinger:
No comma operator tests are causing very long build times without additional benefit for debug runs. 
We don't gain any coverage of code by running these compile only tests in DEBUG mode or in no unnamed lambda mode.
Lets turn off the no comma tests for those cases.

### DIFF
--- a/test/support/test_config.h
+++ b/test/support/test_config.h
@@ -293,11 +293,14 @@
 #    error "TEST_ONLY_HETERO_POLICIES is passed but device backend is not available"
 #endif
 
-// There are issues with stable_sort for libcpp with iterators with a deleted comma operator, disable those tests
-#if (defined(_LIBCPP_VERSION))
-#   define _PSTL_LIBCPP_NO_COMMA_TESTS_BROKEN 1
+//There are issues with stable_sort for libcpp with iterators with a deleted comma operator, disable no_comma compile
+// only tests. Also, since this adds significant amount of code compilation to the build, lets be more conservative
+// about when we try to test this. Since debug build mode and unnamed lambda support dont change the code in any way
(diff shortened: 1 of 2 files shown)


test/support/test_config.h
```diff
@@ -293,11 +293,14 @@
...
 #endif
 
-// There are issues with stable_sort for libcpp with iterators with a deleted comma operator, disable those tests
-#if (defined(_LIBCPP_VERSION))
-#   define _PSTL_LIBCPP_NO_COMMA_TESTS_BROKEN 1
+//There are issues with stable_sort for libcpp with iterators with a deleted comma operator, disable no_comma compile
+// only tests. Also, since this adds significant amount of code compilation to the build, lets be more conservative
+// about when we try to test this. Since debug build mode and unnamed lambda support dont change the code in any way
+// which should interact with what we are testing for here, lets disable it for those cases for time / build space.
+#if !defined(_LIBCPP_VERSION) && !defined(_DEBUG) && defined(__SYCL_UNNAMED_LAMBDA__)
```
mmichel11: Does `defined(__SYCL_UNNAMED_LAMBDA__)` turn off this case for all non-SYCL tests since it is not defined? Is this too restrictive?

danhoeflinger: good point... I'll fix it.

mmichel11: With regards to detecting debug configs, I know `_DEBUG` is defined by MSVC. Do other compilers define it as well? In our CMakeLists.txt, we define `PSTL_USE_DEBUG=1` for CMake debug configurations, so perhaps this is a more reliable macro to check.

danhoeflinger: fixed.

PR #2412, opened 2025-08-29
Remove `struct __projection_deref`

SergeyKopienko:
This PR removes the `__projection_deref` struct and replaces its usage with direct `std::invoke` calls throughout the codebase. The purpose is to simplify the projection handling by eliminating an intermediate wrapper that was previously used to handle projection dereferencing.

- Removal of the `__projection_deref` template struct from utils.h
- Direct replacement of `__projection_deref` instances with `std::invoke` calls
- Cleanup of related SYCL device copyability traits


dmitriy-sobolev (approved):
LGTM. I prefer directly calling std::invoke - it short, so there is no need in extra abstractions.

PR #2414, opened 2025-08-29
Disable device copyable kernel submitter tests prior to icpx 2024.1

mmichel11:
Device copyable tests for kernel submitters prior to icpx 2024.1 encounter compilation errors, and I suspect the issue is a now resolved compiler bug.


danhoeflinger (approved):
LGTM

PR #2419, opened 2025-09-01
Fix forward declaration of __pattern_walk1

dmitriy-sobolev:
It's impossible to deduce `_IsVector`. Also, the implementation does not have it.


SergeyKopienko (approved):
LGTM

PR #2422, opened 2025-09-02
Serial range-based set algorithms with limited outputs

dmitriy-sobolev:
These algorithms should replace the ones we use from the standard libraries, because they expect unlimited output sequence, which is not aligned with the API of the algorithms with the execution policies from C++26.

Draft, until all set algorithms are implemented together with the tests


dmitriy-sobolev:
Closing as outdated. It's superseded by #2569.

PR #2427, opened 2025-09-04
Align size types for set alg helpers

danhoeflinger:
In balanced path set operations, we originally assumed that the type returned by size() was aligned between different ranges.  This is not the case, especially as we move to support the ranges set APIs.

This PR aligns types better across ranges so that we do not encounter trouble here.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -614,8 +614,8 @@ __encode_balanced_path_temp_data(const _IdxT __rng1_idx, const bool __star)
 {
     using signed_t = std::make_signed_t<_IdxT>;
 
-    // Convert to signed representation
-    signed_t __signed_idx{__rng1_idx};
+    // Convert to signed representation - we know this is positive and can be represented in the signed portion
+    signed_t __signed_idx{static_cast<signed_t>(__rng1_idx)};
 
     // Branchless negation: (1 - 2 * __star) gives 1 if __star is false, -1 if __star is true
     return __signed_idx * (signed_t{1} - signed_t{2} * signed_t{__star});
@@ -664,8 +664,8 @@ struct __get_bounds_simple
         using _SizeType = std::common_type_t<std::make_unsigned_t<decltype(__rng1.size())>,
                                              std::make_unsigned_t<decltype(__rng2.size())>>;
 
-        return std::make_tuple(_SizeType{0}, _SizeType{static_cast<_SizeType>(__rng1.size())}, _SizeType{0},
-                               _SizeType{static_cast<_SizeType>(__rng2.size())});
+        return std::make_tuple(_SizeType{0}, static_cast<_SizeType>(__rng1.size()), _SizeType{0},
+                               static_cast<_SizeType>(__rng2.size()));
     }
 };
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -629,7 +629,9 @@ struct __get_bounds_partitioned
     {
         auto __rng_tmp_diag = std::get<2>(__in_rng.tuple()); // set a temp storage sequence
 
-        using _SizeType = decltype(std::get<0>(__in_rng.tuple()).size());
+        using _SizeType = std::common_type_t<std::make_unsigned_t<decltype(std::get<0>(__in_rng.tuple()).size())>,
+                                             std::make_unsigned_t<decltype(std::get<1>(__in_rng.tuple()).size())>,
+                                             std::make_unsigned_t<decltype(__rng_tmp_diag.size())>>
 
         // Establish bounds of ranges for the tile from sparse partitioning pass kernel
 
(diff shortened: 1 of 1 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -665,9 +665,11 @@ struct __get_bounds_simple
...
         const auto& __rng2 = std::get<1>(__in_rng.tuple()); // second sequence
 
-        using _SizeType = std::common_type_t<std::ranges::range_size_t<decltype(__rng1)>, std::ranges::range_size_t<decltype(__rng2)>>;
+        using _SizeType = std::common_type_t<std::make_unsigned_t<decltype(__rng1.size())>,
+                                             std::make_unsigned_t<decltype(__rng2.size())>>;
 
-        return std::make_tuple(_SizeType{0}, _SizeType{__rng1.size()}, _SizeType{0}, _SizeType{__rng2.size()});
+        return std::make_tuple(_SizeType{0}, _SizeType{static_cast<std::size_t>(__rng1.size())}, _SizeType{0},
+                               _SizeType{static_cast<std::size_t>(__rng2.size())});
     }
```
SergeyKopienko, line 669: Why we need here `static_cast<std::size_t>` ?

SergeyKopienko: Probably we should have here `static_cast<_SizeType>(...)`

dmitriy-sobolev (commented):
Let's target that PR to main. This problem may probably occur even in the iterator APIs, not only in the range ones (which nevertheless exposed the issue). It will also make the set PR more self-contained.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
danhoeflinger: i think this was a autocomplete mistake.  fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -659,8 +659,11 @@ struct __get_bounds_simple
         const auto& __rng1 = std::get<0>(__in_rng.tuple()); // first sequence
         const auto& __rng2 = std::get<1>(__in_rng.tuple()); // second sequence
 
-        using _SizeType = decltype(__rng1.size());
-        return std::make_tuple(_SizeType{0}, __rng1.size(), _SizeType{0}, __rng2.size());
+        using _SizeType = std::common_type_t<std::make_unsigned_t<decltype(__rng1.size())>,
+                                             std::make_unsigned_t<decltype(__rng2.size())>>;
+
+        return std::make_tuple(_SizeType{0}, _SizeType{static_cast<_SizeType>(__rng1.size())}, _SizeType{0},
```
dmitriy-sobolev: `_SizeType{static_cast<_SizeType>(__rng1.size())}` looks weird. Did you mean just `_SizeType{__rng1.size()}`?

danhoeflinger: fixed.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -629,21 +629,23 @@ struct __get_bounds_partitioned
     {
         auto __rng_tmp_diag = std::get<2>(__in_rng.tuple()); // set a temp storage sequence
 
-        using _SizeType = decltype(std::get<0>(__in_rng.tuple()).size());
+        using _SizeType = std::common_type_t<std::make_unsigned_t<decltype(std::get<0>(__in_rng.tuple()).size())>,
+                                             std::make_unsigned_t<decltype(std::get<1>(__in_rng.tuple()).size())>,
+                                             std::make_unsigned_t<decltype(__rng_tmp_diag.size())>>;
```
dmitriy-sobolev, line 634: I had to remove `make_unsigned_t` in https://github.com/uxlfoundation/oneDPL/pull/2320/commits/84b7d3033715ba7d66adfcb639decd42f9ec7679 to fix these issues: https://github.com/uxlfoundation/oneDPL/actions/runs/17499055119/job/49707265893.

I do not know what exactly wrong here, but a proper fix may be done here.

dmitriy-sobolev: I see that there are 3 ranges here instead of 2. Probably, it is that proper fix.

danhoeflinger: I think that there was previously just some other functions where there was missing type alignment, and now we are in better shape with the current state of the PR.    Merging the current branch with the set ranges PR, I believe it builds now.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -861,16 +867,19 @@ struct __gen_set_op_from_known_balanced_path
...
             return std::make_tuple(std::uint32_t{0}, std::uint16_t{0});
         auto [__rng1_idx, __rng2_idx, __star_offset] =
             oneapi::dpl::__par_backend_hetero::__decode_balanced_path_temp_data(__rng1_temp_diag, __id,
                                                                                 __diagonal_spacing);
 
-        _SizeType __eles_to_process =
-            std::min(_SizeType{__diagonal_spacing - __star_offset}, __rng1.size() + __rng2.size() - __i_elem + 1);
+        std::uint16_t __eles_to_process =
+            static_cast<std::uint16_t>(std::min(static_cast<_SizeType>(__diagonal_spacing - __star_offset),
+                                                static_cast<_SizeType>(__rng1.size() + __rng2.size() - __i_elem + 1)));
```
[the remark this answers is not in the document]
danhoeflinger: we know this will fit within 16 bits due to the context.

PR #2432, opened 2025-09-05
Remove all capture by references to get of tuple temp

danhoeflinger:
We introduced a bug in #2318 which caused a reference to a temporary tuple which had left scope. 

In these cases, `zip_view::tuple()` returns a temporary copy of the tuple member field of the tuple of the views.  We then apply `std::get`, and capture its result by reference. The temporary copy only lives within the expression argument to `std::get` and the reference returned refers to data which is no longer in scope.  

When we try to use it, bad things happen.

This PR removes the capture by reference which were not helping much given that these views should be light-weight, and copies are OK.


SergeyKopienko (approved):
LGTM

Initially it was my suggestion to use reference at these places.

timmiesmith (approved):
LGTM.  Thank you.

PR #2435, opened 2025-09-08
Bump ONEDPL_HAS_RANGE_ALGORITHMS to 202509

dmitriy-sobolev:
#2366 and #2272 added new parallel range algorithms. Let's bump the corresponding version macro.

### DIFF
--- a/documentation/library_guide/macros.rst
+++ b/documentation/library_guide/macros.rst
@@ -47,6 +47,7 @@ Macro                              Macro values and the functionality
 ``ONEDPL_HAS_RANGE_ALGORITHMS``    Parallel range algorithms.
 
                                    * ``202409L`` - see :ref:`available algorithms <range-algorithms-202409L>`.
+                                   * ``202505L`` - see :ref:`available algorithms <range-algorithms-202505L>`.
                                    * ``202509L`` - see :ref:`available algorithms <range-algorithms-202509L>`.
 ================================== ===============================================
 


include/oneapi/dpl/internal/version_impl.h
akukanov, line 1: Should the patch also update the documentation?

SergeyKopienko:
@dmitriy-sobolev I think this change is not enough.
Because I see a lot of checks in our tests like
```C++
#if ONEDPL_HAS_RANGE_ALGORITHMS > 202505L
```
Should we change `202505` in the tests too?

include/oneapi/dpl/internal/version_impl.h
dmitriy-sobolev: Done. I've added algorithms which are in the `main` branch. The rest algorithms targeted for 2022.10 can be added later, when they are also merged.

documentation/library_guide/macros.rst
```diff
@@ -47,7 +47,7 @@ Macro                              Macro values and the functionality
 ``ONEDPL_HAS_RANGE_ALGORITHMS``    Parallel range algorithms.
 
                                    * ``202409L`` - see :ref:`available algorithms <range-algorithms-202409L>`.
-                                   * ``202505L`` - see :ref:`available algorithms <range-algorithms-202505L>`.
+                                   * ``202509L`` - see :ref:`available algorithms <range-algorithms-202509L>`.
```
SergeyKopienko, line 51: Should we have some new link here instead of modify existing line?

dmitriy-sobolev: Fixed

timmiesmith (approved):
LGTM. Thank you.

PR #2436, opened 2025-09-08
RFC for part 2 of range algorithms: supported

dmitriy-sobolev:
Let's move this RFC from the `proposed` into the `supported` folder since it has already been implemented. It should have been done in 2022.9.


dmitriy-sobolev:
Closing this to incorporate all the algorithms in #2439.

PR #2440, opened 2025-09-08
Document an output range size requirement

dmitriy-sobolev:
Some algorithms have an additional requirement for the output range size on top of the specification. Let's document it.

Draft until all the listed algorithms are fully implemented.

### DIFF
--- a/documentation/library_guide/parallel_api/parallel_range_algorithms.rst
+++ b/documentation/library_guide/parallel_api/parallel_range_algorithms.rst
@@ -59,11 +59,11 @@ The following algorithms are not yet fully conformant with the specification
 because they require the output range to have sufficient size to hold all resulting elements:
 
 * ``copy_if``
-* ``unique_copy``
 * ``set_intersection``
 * ``set_union``
 * ``set_difference``
 * ``set_symmetric_difference``
+* ``unique_copy``
 
 The ``ONEDPL_HAS_RANGE_ALGORITHMS`` :ref:`feature macro <feature-macros>` may be used to test for the presence of
 parallel range algorithms.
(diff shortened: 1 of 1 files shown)


documentation/library_guide/parallel_api/parallel_range_algorithms.rst
```diff
@@ -12,6 +12,19 @@ oneDPL only supports random access ranges, because they allow simultaneous const
 at any position in the range. This enables efficient workload distribution among multiple threads or processing units,
 which is essential for achieving high performance in parallel execution.
 
+Their interfaces are described in the
+`oneDPL Specification <https://github.com/uxlfoundation/oneAPI-spec/blob/main/source/elements/oneDPL/source/parallel_api/parallel_range_api.rst#parallel-range-algorithms>_.
+
+The following algorithms are not yet fully conformant with the specification
+because they require the output range to have sufficient size to hold all resulting elements:
```
timmiesmith, line 19:
```suggestion
because they require the output range to have sufficient size to hold all elements in the input ranges:
```
Storage for "all resulting elements" of an algorithm is what should be required.  What is required instead is storage to hold the full input ranges, correct?

dmitriy-sobolev: No, holding the input ranges fully is not required.

These algorithms conditionally copy elements to the resulting buffer. Here is an example. Even though `out` has a smaller capacity than `in`, it is still valid:
```
std::vector<int> in = {0, 1, 2, 3, 4, 5, 6, 7, 8, 9};
std::vector<int> out(5);
auto even = [](auto value){ return value % 2 == 0; };
auto end = std::copy_if(in.begin(), in.end(), out.begin(), even);
```

Our range-based algorithms currently behave the same way.

documentation/library_guide/parallel_api/parallel_range_algorithms.rst
akukanov, line 1: Not sure if documenting the (supposedly temporary) limitation on the feature page is the right way. I would certainly add a release note for that; while on this page I would either do nothing or add an implementation note at the end of the page.

PR #2441, opened 2025-09-08
Update a note about a changed implementation detail: swap/iter_swap

dmitriy-sobolev:
The change is made in #1843

### DIFF
--- a/documentation/library_guide/introduction.rst
+++ b/documentation/library_guide/introduction.rst
@@ -195,7 +195,7 @@ Known Limitations
 * ``std::ranges::drop_view`` from libstdc++ version 10 may throw exceptions.
   This can lead to a "SYCL kernel cannot use exceptions" compilation error
   when it is used to pass data to a range-based algorithm with a device policy.
-* Range-based ``sort`` and ``stable_sort`` called with device execution policies
+* Range-based ``sort`` and ``stable_sort`` algorithms called with device execution policies
   use ``std::swap`` instead of ``std::ranges::iter_swap``.
   As a result, customizations targeting ``std::ranges::iter_swap`` will not be respected.
 


documentation/library_guide/parallel_api/parallel_range_algorithms.rst
```diff
@@ -135,11 +135,6 @@ Usage Example for Parallel Range Algorithms
         oneapi::dpl::ranges::copy(oneapi::dpl::execution::dpcpp_default, view_in, std::span(vec_out));
     }
 
-Implementation Notes
---------------------
-The ``sort`` and ``stable_sort`` algorithms use ``std::swap`` and not ``std::ranges::iter_swap`` for swapping elements.
-As a result, customizations targeting ``std::ranges::iter_swap`` will not be respected.
```
akukanov, line 141: Do we have a test that `iter_swap` is really used?

dmitriy-sobolev:
Sorry for the confusion. #1815 has a note that the issue was solved in #1843. Let me update the description...

documentation/library_guide/parallel_api/parallel_range_algorithms.rst
dmitriy-sobolev: We do not have such a test.

Anyway, looking at the same issue again in the morning makes a difference - I see that we still use std::swap in the device code. #1843 fixed the issue for the host only.

dmitriy-sobolev: Also #1815 has an important detail:

> The fix is not applicable to the device backend since it does not keep the information about the original range; std::ranges::iter_swap is applied to a local accessor rather than the original range iterators.

Given that, I think it is fair to say that the algorithms with the device policies use `std::swap`, while the ones with the host policies rely on the swapping methods used in the corresponding sorting algorithms in the standard libraries.

dmitriy-sobolev: I've updated the note to cite only device policies.

akukanov: I wonder if we should move it to https://uxlfoundation.github.io/oneDPL/introduction.html#known-limitations instead.

dmitriy-sobolev: Done. It makes sense for a permanent detail treated as a limitation.

documentation/library_guide/introduction.rst
```diff
@@ -195,6 +195,9 @@ Known Limitations
 * ``std::ranges::drop_view`` from libstdc++ version 10 may throw exceptions.
   This can lead to a "SYCL kernel cannot use exceptions" compilation error
   when it is used to pass data to a range-based algorithm with a device policy.
+* Range-based ``sort`` and ``stable_sort`` called with device execution policies
```
akukanov:
```suggestion
* Range-based ``sort`` and ``stable_sort`` algorithms called with device execution policies
```

PR #2445, opened 2025-09-12
Do not use size() and empty() methods in ranges

dmitriy-sobolev:
C++20 ranges may not have `size()` or `empty()` methods. A range is guaranteed to have only `begin()` and `end()`.

Let's use `std::ranges::size`, `std::ranges::empty` if its C++20 and continue using `.size()` and `.empty()` if its C++17 - the experimental oneDPL ranges should provide them.

### DIFF
--- a/include/oneapi/dpl/pstl/utils_ranges.h
+++ b/include/oneapi/dpl/pstl/utils_ranges.h
@@ -107,6 +107,12 @@ template <typename _R>
 auto
 __check_size(long) -> decltype(std::declval<_R&>().get_count());
 
+#if _ONEDPL_CPP20_RANGES_PRESENT
+template <typename _R>
+auto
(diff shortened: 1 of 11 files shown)


include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -129,17 +129,24 @@ using projected_value_t = std::remove_cvref_t<std::invoke_result_t<Proj&, std::i
...
 #endif
+}
+
+struct __first_size_calc
+{
+    template <typename _Range, typename... _Ranges>
+    auto
+    operator()(const _Range& __rng, const _Ranges&...) const
+    {
+        return __size(__rng);
```
SergeyKopienko: This line looks like type conversion. What about to rename `__size` function somehow?

SergeyKopienko: The name `__size` is ok

PR #2451, opened 2025-09-15
Initialize a variable before use

dmitriy-sobolev:
It fixes a Coverity scan hit:

> Uninitialized scalar variable (UNINIT): Unclassified, Unspecified, Undecided first detected on 2025-09-12.
var_decl: Declaring variable "__get_data_op" without initializer.
uninit_use: Using uninitialized value "__get_data_op".

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -1145,7 +1145,7 @@ __parallel_set_scan(_SetTag, sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng
     _Assigner __assign_op{};
     _DataAcc __get_data_op{};
     unseq_backend::__copy_by_mask<_ReduceOp, oneapi::dpl::__internal::__pstl_assign, /*inclusive*/ std::true_type, 2>
-        __copy_by_mask_op;
+        __copy_by_mask_op{};
     unseq_backend::__brick_set_op<_SetTag, _Compare, _Size1, _Size2> __create_mask_op{__comp, __n1, __n2};
 
     // temporary buffer to store boolean mask


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -1143,7 +1143,7 @@ __parallel_set_scan(_SetTag, sycl::queue& __q, _Range1&& __rng1, _Range2&& __rng
 
     _ReduceOp __reduce_op;
     _Assigner __assign_op;
-    _DataAcc __get_data_op;
+    _DataAcc __get_data_op{oneapi::dpl::identity{}};
```
```suggestion
    _DataAcc __get_data_op();
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -834,7 +834,7 @@ __parallel_scan_copy(sycl::queue& __q, _InRng&& __in_rng, _OutRng&& __out_rng, _
 
     _Assigner __assign_op;
     _ReduceOp __reduce_op;
-    _DataAcc __get_data_op;
+    _DataAcc __get_data_op{oneapi::dpl::identity{}};
```
```suggestion
    _DataAcc __get_data_op();
```

SergeyKopienko (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -834,7 +834,7 @@ __parallel_scan_copy(sycl::queue& __q, _InRng&& __in_rng, _OutRng&& __out_rng, _
 
     _Assigner __assign_op;
     _ReduceOp __reduce_op;
-    _DataAcc __get_data_op;
+    _DataAcc __get_data_op{};
```
akukanov: Why is initialization needed for `__get_data_op` but not needed for three other variables around it?

dmitriy-sobolev: That's a mystery to me right now why coverity produced such a hit, but adding this kind of initialization in the surrounding code makes sense, at least for consistency. See [e360b7b](https://github.com/uxlfoundation/oneDPL/pull/2451/commits/e360b7ba64151fc7006d442f4ba947583c6169fb).

akukanov: If it is a false positive by coverity, we can as well mark it such.
The question is, as usual - are we sure these types are always default-constructible?

SergeyKopienko (approved):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: I think - yes, `oneapi::dpl::identity` is default-constructible.
But this can mean that we should apply `{}` only for `_DataAcc __get_data_op`

dmitriy-sobolev: To add more details, the modified functors have no members, except for two having `std::plus` and `dpl::identity`. I cannot prove that `std::plus` is default constructible, but this PR does not change its usage.

akukanov: If these classes do not depend on user-provided functions/types and we are sure these are default-constructible, then it's fine (though I wonder then if the variables are even needed).

dmitriy-sobolev: > If these classes do not depend on user-provided functions/types...

No such dependencies.

> we are sure these are default-constructible...

Not sure about `std::plus`, but it's been default constructed everywhere in the library with some cases being 5+ years old. Given that it should be safe, but I've created an issue: #2454.

> though I wonder then if the variables are even needed

They aren't. I'll remove them separately.

PR #2455, opened 2025-09-16
Fix incorrect usage of `_DifferenceType1` / `_DifferenceType2` in the code

SergeyKopienko:
In this PR we fix incorrect usage of `_DifferenceType1` / `_DifferenceType2` in the code.

Introduced in #2320 and earlier

### DIFF
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3452,8 +3452,8 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
             [=, &__exec, &__result] {
                 __result = __internal::__parallel_set_op(
                     __tag, __exec, __left_bound_seq_1, __last1, __first2, __last2, __result,
-                    [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op, __comp,
-                    __proj1, __proj2);
+                    [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op, __comp, __proj1,
+                    __proj2);
             });
         return __result;
     }
@@ -3473,8 +3473,8 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
             [=, &__exec, &__result] {
                 __result = __internal::__parallel_set_op(
                     __tag, __exec, __first1, __last1, __left_bound_seq_2, __last2, __result,
-                    [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op, __comp,
-                    __proj1, __proj2);
+                    [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op, __comp, __proj1,
+                    __proj2);
             });
         return __result;
     }
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3601,7 +3601,10 @@ __pattern_set_intersection(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& _
                            _RandomAccessIterator2 __last2, _RandomAccessIterator3 __result, _Compare __comp)
 {
     using _T = typename std::iterator_traits<_RandomAccessIterator3>::value_type;
-    using _DifferenceType = typename std::iterator_traits<_RandomAccessIterator1>::difference_type;
+
+    using _DifferenceType1 = typename std::iterator_traits<_RandomAccessIterator1>::difference_type;
+    using _DifferenceType2 = typename std::iterator_traits<_RandomAccessIterator2>::difference_type;
+    using _DifferenceType = std::common_type_t<_DifferenceType1, _DifferenceType2>;
 
     const auto __n1 = __last1 - __first1;
     const auto __n2 = __last2 - __first2;
--- a/include/oneapi/dpl/pstl/algorithm_impl.h
+++ b/include/oneapi/dpl/pstl/algorithm_impl.h
@@ -3450,7 +3450,7 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
             [=, &__exec, &__result] {
                 __result = __internal::__parallel_set_op(
                     __tag, __exec, __left_bound_seq_1, __last1, __first2, __last2, __result,
-                    [](_DifferenceType1 __n, _DifferenceType1 __m) { return __n + __m; }, __set_union_op, __comp,
+                    [](_DifferenceType1 __n, _DifferenceType2 __m) { return __n + __m; }, __set_union_op, __comp,
                     __proj1, __proj2);
             });
         return __result;
@@ -3471,7 +3471,7 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
             [=, &__exec, &__result] {
                 __result = __internal::__parallel_set_op(
                     __tag, __exec, __first1, __last1, __left_bound_seq_2, __last2, __result,
-                    [](_DifferenceType1 __n, _DifferenceType1 __m) { return __n + __m; }, __set_union_op, __comp,
+                    [](_DifferenceType1 __n, _DifferenceType2 __m) { return __n + __m; }, __set_union_op, __comp,
                     __proj1, __proj2);
             });
         return __result;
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3450,7 +3452,7 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
             [=, &__exec, &__result] {
                 __result = __internal::__parallel_set_op(
                     __tag, __exec, __left_bound_seq_1, __last1, __first2, __last2, __result,
-                    [](_DifferenceType1 __n, _DifferenceType1 __m) { return __n + __m; }, __set_union_op, __comp,
+                    [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op, __comp,
```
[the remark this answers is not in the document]
SergeyKopienko: I think this new state is correct

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3471,15 +3473,15 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
             [=, &__exec, &__result] {
                 __result = __internal::__parallel_set_op(
                     __tag, __exec, __first1, __last1, __left_bound_seq_2, __last2, __result,
-                    [](_DifferenceType1 __n, _DifferenceType1 __m) { return __n + __m; }, __set_union_op, __comp,
+                    [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op, __comp,
```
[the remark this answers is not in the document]
SergeyKopienko: I think this new state is correct

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3471,15 +3473,15 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
...
                     __proj1, __proj2);
             });
         return __result;
     }
 
     return __internal::__parallel_set_op(
         __tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2, __last2, __result,
-        [](_DifferenceType1 __n, _DifferenceType1 __m) { return __n + __m; }, __set_union_op, __comp, __proj1, __proj2);
+        [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op, __comp, __proj1, __proj2);
```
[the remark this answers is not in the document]
SergeyKopienko: I think this new state is correct

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3471,15 +3473,15 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
             [=, &__exec, &__result] {
                 __result = __internal::__parallel_set_op(
                     __tag, __exec, __first1, __last1, __left_bound_seq_2, __last2, __result,
-                    [](_DifferenceType1 __n, _DifferenceType1 __m) { return __n + __m; }, __set_union_op, __comp,
-                    __proj1, __proj2);
+                    [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op, __comp, __proj1,
+                    __proj2);
```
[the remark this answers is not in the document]
SergeyKopienko: Declined.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3333,7 +3334,7 @@ __parallel_set_op(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
                         __bb = __first2 + __internal::__pstl_lower_bound(__first2, _DifferenceType2{0}, __last2 - __first2,
                                                                          std::invoke(__proj1, *__b), __comp, __proj2);
 
-                    const _DifferenceType1 __buf_pos = __size_func((__b - __first1), (__bb - __first2));
+                    const _DifferenceType __buf_pos = __size_func((__b - __first1), (__bb - __first2));
```
dmitriy-sobolev, line 3337: Nitpick, given that `_DifferenceType` is also used in `_SetRange`.

Why not to use auto? I expect that `__size_func` provides the right common type, and there is no reason to infer it twice.

dmitriy-sobolev (approved):
Having a common type for an index/size type in a resulting sequence looks reasonable.

Even though, it does not lead to issues in the tests, it still may avoid potential issues with unexpected type conversions. The risk is very low, so it may be added to 2022.10 in my opinion.

include/oneapi/dpl/pstl/algorithm_impl.h
SergeyKopienko: We have `_SizeFunction __size_func` as parameter of some template type.
I think make sense explicitly specify the type in this place.

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3450,8 +3452,8 @@ __parallel_set_union_op(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __ex
             [=, &__exec, &__result] {
                 __result = __internal::__parallel_set_op(
                     __tag, __exec, __left_bound_seq_1, __last1, __first2, __last2, __result,
-                    [](_DifferenceType1 __n, _DifferenceType1 __m) { return __n + __m; }, __set_union_op, __comp,
-                    __proj1, __proj2);
+                    [](_DifferenceType __n, _DifferenceType __m) { return __n + __m; }, __set_union_op, __comp, __proj1,
+                    __proj2);
```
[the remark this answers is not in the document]
SergeyKopienko: I think this new state is correct

include/oneapi/dpl/pstl/algorithm_impl.h
```diff
@@ -3309,8 +3310,8 @@ __parallel_set_op(__parallel_tag<_IsVector>, _ExecutionPolicy&& __exec, _RandomA
                                                                   __result + __s.__pos, _IsVector{});
         };
         __par_backend::__parallel_strict_scan(
-            __backend_tag{}, std::forward<_ExecutionPolicy>(__exec), __n1, _SetRange{0, 0, 0}, //-1, 0},
-            [=](_DifferenceType1 __i, _DifferenceType1 __len) {                                // Reduce
+            __backend_tag{}, std::forward<_ExecutionPolicy>(__exec), __n1, _SetRange{0, 0, 0},
+            [=](_DifferenceType1 __i, _DifferenceType1 __len) { // Reduce
```
timmiesmith, line 3314: Should `_DifferenceType1` be replaced with `_DifferenceType` here as well?  This change currently is just removing deadcode and whitespace. What's the expected case where the common `_DifferenceType` should be used instead of `_DifferenceType1` or `_DifferenceType2`?

SergeyKopienko: We should use `_DifferenceType` when we calculate the `sum` or `min` of/from variables of different `_DifferenceType1` and `_DifferenceType2`  types.
I will think and consult with @danhoeflinger about this case.

SergeyKopienko: From my point of view exactly in this place we should use the `decltype(__n1)` inside `Reduce` lambda: it is `_DifferenceType1` type. So this code is correct.

timmiesmith: Thanks for the clarification, Sergey.  This looks ok to me then.

include/oneapi/dpl/pstl/algorithm_ranges_impl.h
```diff
@@ -1065,7 +1067,7 @@ __pattern_set_difference(__parallel_tag<_IsVector> __tag, _ExecutionPolicy&& __e
     {
         auto __out_last = __parallel_set_op(
             __tag, std::forward<_ExecutionPolicy>(__exec), __first1, __last1, __first2, __last2, __result,
-            [](_DifferenceType1 __n, _DifferenceType2) { return __n; },
```
danhoeflinger, line 1068: I'm not sure this one is necessary, but it probably makes sense to keep the change to make it consistent.

SergeyKopienko: Agree that in this line this change is only for align with other lambdas like that.

include/oneapi/dpl/pstl/algorithm_impl.h
akukanov: I really do not want to either guess or look into a function to understand how a variable can be used - the information which the type often gets straight.

PR #2458, opened 2025-09-16
Fix header inclusion memory tests for libstdc++10

dmitriy-sobolev:
`std::make_unique_for_overwrite` is available in libstdc++11 and newer.

### DIFF
--- a/test/general/header_inclusion_order_memory_0.pass.cpp
+++ b/test/general/header_inclusion_order_memory_0.pass.cpp
@@ -21,7 +21,6 @@
 #include "support/utils.h"
 
 #if ONEDPL_HAS_RANGE_ALGORITHMS >= 202509L
-#include <memory>    // std::make_unique_for_overwrite
 #include <ranges>    // std::ranges::subrange
 #include <algorithm> // std::ranges::count
 #endif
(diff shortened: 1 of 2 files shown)


test/general/header_inclusion_order_memory_0.pass.cpp
```diff
@@ -31,7 +30,10 @@ main()
 {
 #if ONEDPL_HAS_RANGE_ALGORITHMS >= 202509L
     constexpr int n = 10;
-    std::unique_ptr<int[]> ptr = std::make_unique_for_overwrite<int[]>(n);
+    std::allocator<int> alloc;
+    int* raw_ptr = alloc.allocate(n);
+    auto deleter = [&](int* p){ alloc.deallocate(p, n); };
+    std::unique_ptr<int, decltype(deleter)> ptr(raw_ptr, deleter);
```
akukanov:
```suggestion
    std::unique_ptr<int[]> ptr(new int[n]);
```

test/general/header_inclusion_order_memory_0.pass.cpp
```diff
@@ -21,7 +21,6 @@
 #include "support/utils.h"
 
 #if ONEDPL_HAS_RANGE_ALGORITHMS >= 202509L
-#include <memory>    // std::make_unique_for_overwrite
```
akukanov, line 24: Technically, `unique_ptr` is defined in `<memory>`

SergeyKopienko (approved):
LGTM

PR #2460, opened 2025-09-16
Fix coverity issues in set parallel range algorithms #1

SergeyKopienko:
This PR fixes Coverity static analysis issues in set parallel range algorithms by addressing potential object lifetime problems. The changes replace direct projection of temporary objects with proper intermediate variable storage to ensure object lifetime extends through the projection operation.

- Introduces intermediate variables to store array element values before applying projections
- Modifies three different set operation implementations to use the safer pattern
- Addresses potential dangling reference issues that could cause undefined behavior

Alternative approach implemented in the PR https://github.com/uxlfoundation/oneDPL/pull/2459

Source issue has been introduced in the PR #2320

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -389,7 +389,8 @@ struct __gen_set_mask
 
         std::size_t __nb = __set_b.size();
 
-        const auto& __val_a_proj = std::invoke(__proj1, __set_a[__id]);
+        auto&& __val_a = __set_a[__id];
+        auto&& __val_a_proj = std::invoke(__proj1, __val_a);
 
         auto __res =
             oneapi::dpl::__internal::__pstl_lower_bound(__set_b, std::size_t{0}, __nb, __val_a_proj, __comp, __proj2);
@@ -404,7 +405,8 @@ struct __gen_set_mask
         }
         else
         {
-            const auto& __val_b_proj = std::invoke(__proj2, __set_b[__res]);
+            auto&& __val_b = __set_b[__res];
+            auto&& __val_b_proj = std::invoke(__proj2, __val_b);
 
             //Difference operation logic: if number of duplication in __set_a on left side from __id > total number of
             //duplication in __set_b then a mask is 1
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -389,7 +389,8 @@ struct __gen_set_mask
 
         std::size_t __nb = __set_b.size();
 
-        // This local variable extends the lifetime of operator[] return in case of temporary object
+        // This reference extends the lifetime of a temporary object returned by operator[]
+        // so that it can be safely used with identity projections
         auto&& __val_a = __set_a[__id];
         auto&& __val_a_proj = std::invoke(__proj1, std::forward<decltype(__val_a)>(__val_a));
 
@@ -406,7 +407,8 @@ struct __gen_set_mask
         }
         else
         {
-            // This local variable extends the lifetime of operator[] return in case of temporary object
+            // This reference extends the lifetime of a temporary object returned by operator[]
+            // so that it can be safely used with identity projections
             auto&& __val_b = __set_b[__res];
             auto&& __val_b_proj = std::invoke(__proj2, std::forward<decltype(__val_b)>(__val_b));
 
@@ -711,7 +713,8 @@ struct __gen_set_balanced_path
             return std::make_tuple(__merge_path_rng1, __merge_path_rng2, false);
         }
 
-        // This local variable extends the lifetime of operator[] return in case of temporary object
+        // This reference extends the lifetime of a temporary object returned by operator[]
+        // so that it can be safely used with identity projections
         auto&& __ele_val = __rng1[__merge_path_rng1 - 1];
         auto&& __ele_val_proj = std::invoke(__proj1, std::forward<decltype(__ele_val)>(__ele_val));
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -389,6 +389,7 @@ struct __gen_set_mask
 
         std::size_t __nb = __set_b.size();
 
+        // This local variable extends the lifetime of operator[] return in case of temporary object
         auto&& __val_a = __set_a[__id];
         auto&& __val_a_proj = std::invoke(__proj1, std::forward<decltype(__val_a)>(__val_a));
 
@@ -405,6 +406,7 @@ struct __gen_set_mask
         }
         else
         {
+            // This local variable extends the lifetime of operator[] return in case of temporary object
             auto&& __val_b = __set_b[__res];
             auto&& __val_b_proj = std::invoke(__proj2, std::forward<decltype(__val_b)>(__val_b));
 
@@ -709,6 +711,7 @@ struct __gen_set_balanced_path
             return std::make_tuple(__merge_path_rng1, __merge_path_rng2, false);
         }
 
+        // This local variable extends the lifetime of operator[] return in case of temporary object
         auto&& __ele_val = __rng1[__merge_path_rng1 - 1];
         auto&& __ele_val_proj = std::invoke(__proj1, std::forward<decltype(__ele_val)>(__ele_val));
 
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
@@ -709,7 +709,8 @@ struct __gen_set_balanced_path
             return std::make_tuple(__merge_path_rng1, __merge_path_rng2, false);
         }
 
-        const auto& __ele_val_proj = std::invoke(__proj1, __rng1[__merge_path_rng1 - 1]);
+        auto&& __ele_val = __rng1[__merge_path_rng1 - 1];
+        auto&& __ele_val_proj = std::invoke(__proj1, __ele_val);
 
         if (std::invoke(__comp, __ele_val_proj, std::invoke(__proj2, __rng2[__merge_path_rng2])))
         {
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -389,7 +389,8 @@ struct __gen_set_mask
 
         std::size_t __nb = __set_b.size();
 
-        const auto& __val_a_proj = std::invoke(__proj1, __set_a[__id]);
+        auto&& __val_a = __set_a[__id];
+        auto&& __val_a_proj = std::invoke(__proj1, __val_a);
```
mmichel11: Should `__val_a` be forwarded here and in similar places?

SergeyKopienko: My understanding - it should not be forwarded: it should just save source data and extend their life-time for temporary data objects which returns some times from `operator[]()`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -707,7 +709,8 @@ struct __gen_set_balanced_path
             return std::make_tuple(__merge_path_rng1, __merge_path_rng2, false);
         }
 
-        const auto& __ele_val_proj = std::invoke(__proj1, __rng1[__merge_path_rng1 - 1]);
+        auto&& __ele_val = __rng1[__merge_path_rng1 - 1];
+        auto&& __ele_val_proj = std::invoke(__proj1, __ele_val);
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -404,7 +405,8 @@ struct __gen_set_mask
         }
         else
         {
-            const auto& __val_b_proj = std::invoke(__proj2, __set_b[__res]);
+            auto&& __val_b = __set_b[__res];
+            auto&& __val_b_proj = std::invoke(__proj2, __val_b);
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -1285,7 +1290,8 @@ class __brick_set_op
         }
         else
         {
-            const auto& __val_b_proj = std::invoke(__projB, __b[__b_beg + __res]);
+            auto&& __val_b = __b[__b_beg + __res];
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -1273,7 +1276,9 @@ class __brick_set_op
 
         auto __idx_c = __idx;
         const _SizeA __idx_a = _SizeA(__idx);
-        const auto& __val_a_proj = std::invoke(__projA, __a[__a_beg + __idx_a]);
+
+        auto&& __val_a = __a[__a_beg + __idx_a];
+        auto&& __val_a_proj = std::invoke(__projA, __val_a);
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -1019,15 +1019,18 @@ struct __brick_includes
...
+        auto&& __val_b_proj = std::invoke(__projB, __val_b);
 
         const _SizeA __res = __internal::__pstl_lower_bound(__rngA, __a_beg, __a_end, __val_b_proj, __comp, __projA);
 
         // {a} < {b} or __val_b != __rngA[__res]
         if (__res == __a_end || std::invoke(__comp, __val_b_proj, std::invoke(__projA, __rngA[__res])))
             return true; //__rngA doesn't include __rngB
 
-        const auto& __val_a_proj = std::invoke(__projA, __rngA[__res]);
+        auto&& __val_a = __rngA[__res];
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -1019,15 +1019,18 @@ struct __brick_includes
             return true; //__rngA doesn't include __rngB
 
         const _SizeB __idx_b = __b_beg + __idx;
-        const auto& __val_b_proj = std::invoke(__projB, __rngB[__idx_b]);
+
+        auto&& __val_b = __rngB[__idx_b];
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
akukanov: Actually, I think it should be forwarded to preserve the value category.
I tested that with `std::identity` and a temporary value bound to `auto&&`. Without forwarding, `identity` returns lvalue reference, with forwarding - rvalue reference.
```suggestion
        auto&& __val_a_proj = std::invoke(__proj1, std::forward<decltype(__val_a)>__val_a);
```

SergeyKopienko: Fixed

danhoeflinger: This is confusing to me though.  What is the point of not just doing the dereference inline if we are going to instantly forward from it.  

For non-temporaries, everything is OK with lifetime and this works ok and should capture by lvalue ref with the correct constness.  For temporaries, we capture into rvalue ref `__val_a`, then forward (move) into the invoke call on the next line.  This forward (move) means that `__val_a` is no longer extending lifetime of the temporary. 

If we instead just dereferenced inline and used `auto&&` for `__val_a_proj` wouldn't we see the exact same with simpler code?  For non-temporaries, everything is OK and this works capturing by lvalue ref of the correct constness.  For temporaries, its passed as a rvalue ref directly into the invoke call, and the second `auto&&` does all the work to then capture the result into `__value_a_proj`.

Definitely possible I'm missing something, but I don't see it.

danhoeflinger: The one difference I see is that with the current formulation which uses two `auto&&`, you could capture the type of the dereferenced value by using `decltype(__val_a)`.  Otherwise, I'm pretty convinced they do the same thing and both preserve the value category.

https://godbolt.org/z/8nz4MfKK9

danhoeflinger: Nevermind, retracted.  What I believe I missed is that with the current formulation (2x `auto&&`), the first one does in fact extend the lifetime of the dereferenced result in the local scope when the function its forwarding to accepts and forwards universal references.  What I was suggesting seems to bring us back to a dangling reference to temporary issue.

https://godbolt.org/z/98j7Pb4d6

danhoeflinger (dismissed):
After some exploring, I think this is good.

We probably can do this similar thing in our recent previous instance to save a copy.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
akukanov: > What is the point of not just doing the dereference inline if we are going to instantly forward from it.

We have to separate dereferencing and the call to projection as expressions if we want to avoid repeated invocations of the projection to the same element,

> This forward (move) means that __val_a is no longer extending lifetime of the temporary.

No, it does not; but the reasons are not obvious from the code. According to the C++20 standard, projections, predicates, etc., must not modify the state of their arguments (see `regular_invocable`, https://en.cppreference.com/w/cpp/concepts/invocable.html). It's a semantical requirement that cannot be enforced by the compiler (at least not without also restricting which value categories a function object can take).

> If we instead just dereferenced inline and used auto&& for __val_a_proj wouldn't we see the exact same with simpler code?

I believe not. I see no difference between `const auto&` (before the patch) and `auto&&`for `__val_a_proj` because the problem is that the lifetime of an rvalue after inline dereferencing into a function argument is apparently not extended, or at least Coverity does not think it is.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -389,7 +389,8 @@ struct __gen_set_mask
 
         std::size_t __nb = __set_b.size();
 
-        const auto& __val_a_proj = std::invoke(__proj1, __set_a[__id]);
+        auto&& __val_a = __set_a[__id];
```
akukanov, line 393:
```suggestion
        // Extend the lifetime of a possibly temporary value before passing it to the projection
        // which might return a reference to its argument (e.g., std::identity)
        auto&& __val_a = __set_a[__id];
```
It makes sense to repeat the comment in other affected functions as well.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce_then_scan.h
```diff
@@ -389,7 +389,9 @@ struct __gen_set_mask
 
         std::size_t __nb = __set_b.size();
 
-        const auto& __val_a_proj = std::invoke(__proj1, __set_a[__id]);
+        // This local variable extends the lifetime of operator[] return in case of temporary object
```
akukanov:
```suggestion
        // This reference extends the lifetime of a temporary object returned by operator[]
        // so that it can be safely used with identity projections
```

SergeyKopienko: fixed

akukanov (approved):

PR #2461, opened 2025-09-17
Refactoring of utility functions to evaluate projected values and compare result  in one place

SergeyKopienko:
This PR modifies binary search utility functions to compare result of both projections in one `std::invoke` call at one place.
So we avoid any life-time extensions for the source/projected values and pass projected values into comparator preserving their value category.

## Justification

Synthetic example **with compile error** based on current state of `main` branch: https://godbolt.org/z/zE7KTKPf9, https://godbolt.org/z/7e5z685cj (https://godbolt.org/z/hj6YMv6rq, https://godbolt.org/z/f9oPGbbeM with additional `static_assert` according to comment https://github.com/uxlfoundation/oneDPL/pull/2461#discussion_r2392307799 from @akukanov)
Synthetic example **without compile errors** based on current state of this PR: https://godbolt.org/z/sbz3ehhK6 (https://godbolt.org/z/aq5GxWhej with additional `static_assert` according to comment https://github.com/uxlfoundation/oneDPL/pull/2461#discussion_r2392307799 from @akukanov)

### Two functions `__pstl_lower_bound_idx` / `__pstl_lower_bound` :
- This function should be used when we using `__rng2[__rng2_idx]` value from the second range :
```C++
template <typename _Rng1, typename _Size1, typename _Size2, typename _Rng2, typename _Compare, typename _Proj1,
          typename _Proj2>
_Size1
__pstl_lower_bound_idx(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2, _Size2 __rng2_idx, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

- This finction should be used when the value from the second range described by iterator `__rng2_it` :
```C++
template <typename _Rng1, typename _Size1, typename _Rng2It, typename _Compare, typename _Proj1, typename _Proj2>
_Size1
__pstl_lower_bound(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2It __rng2_it, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

### Two functions of `__pstl_upper_bound_idx` / `__pstl_upper_bound` :
- This function should be used when we using `__rng2[__rng2_idx]` value from the second range :
```C++
template <typename _Rng1, typename _Size1, typename _Size2, typename _Rng2, typename _Compare, typename _Proj1,
          typename _Proj2>
_Size1
__pstl_upper_bound_idx(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2, _Size2 __rng2_idx, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

- This function should be used when the value from the second range described by iterator `__rng2_it` :
```C++
template <typename _Rng1, typename _Size1, typename _Iterator, typename _Compare, typename _Proj1, typename _Proj2>
_Size1
__pstl_upper_bound(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2It __rng2_it, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

### The function `__pstl_left_bound_idx` :
- This function should be used when we using `__rng2[__rng2_idx]` value from the second range :
```C++
template <typename _Rng1, typename _Size1, typename _Size2, typename _Rng2, typename _Compare, typename _Proj1, typename _Proj2>
_Size1
__pstl_left_bound_idx(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2, _Size2 __rng2_idx, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

### The function `__pstl_right_bound_idx` :
- This function should be used when we using `__rng2[__rng2_idx]` value from the second range :
```C++
template <typename _Rng1, typename _Size1, typename _Size2, typename _Rng2, typename _Compare, typename _Proj1,
          typename _Proj2>
_Size1
__pstl_right_bound_idx(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2, _Size2 __rng2_idx, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

### The function `__biased_lower_bound_idx` :
- This function should be used when we using `__rng2[__rng2_idx]` value from the second range :
```C++
template <bool __bias_last = true, typename _Rng1, typename _Size1, typename _Size2, typename _Rng2, typename _Compare, typename _Proj1, typename _Proj2>
_Size1
__biased_lower_bound_idx(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2, _Size2 __rng2_idx, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

### The function `__biased_upper_bound_idx` :
- This overload should be used when we using `__rng2[__rng2_idx]` value from the second range :
```C++
template <bool __bias_last = true, typename _Rng1, typename _Size1, typename _Size2, typename _Rng2, typename _Compare, typename _Proj1, typename _Proj2>
_Size1
__biased_upper_bound_idx(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2, _Size2 __rng2_idx, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

### DIFF
--- a/include/oneapi/dpl/pstl/histogram_binhash_utils.h
+++ b/include/oneapi/dpl/pstl/histogram_binhash_utils.h
@@ -90,7 +90,7 @@ __custom_boundary_get_bin_helper(_Acc __acc, ::std::int32_t __size, _T2 __value,
     std::int32_t ret = -1;
     if (__value >= __min && __value < __max)
     {
-        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, __value, std::less<_T2>{},
+        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, __value, std::less<_T2>{}, // TODO __pstl_upper_bound - value (maybe processed as pointer dereference)
                                                           oneapi::dpl::identity{}) - 1;
     }
     return ret;
--- a/include/oneapi/dpl/pstl/histogram_binhash_utils.h
+++ b/include/oneapi/dpl/pstl/histogram_binhash_utils.h
@@ -90,7 +90,8 @@ __custom_boundary_get_bin_helper(_Acc __acc, ::std::int32_t __size, _T2 __value,
     std::int32_t ret = -1;
     if (__value >= __min && __value < __max)
     {
-        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, &__value, std::less<_T2>{},
+        _T2* __value_ptr = &__value;
+        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, __value_ptr, std::less<_T2>{},
                                                           oneapi::dpl::identity{}, oneapi::dpl::identity{}) - 1;
     }
     return ret;
--- a/include/oneapi/dpl/pstl/histogram_binhash_utils.h
+++ b/include/oneapi/dpl/pstl/histogram_binhash_utils.h
@@ -92,7 +92,8 @@ __custom_boundary_get_bin_helper(_Acc __acc, ::std::int32_t __size, _T2 __value,
     {
         _T2* __value_ptr = &__value;
         ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, __value_ptr, std::less<_T2>{},
-                                                          oneapi::dpl::identity{}, oneapi::dpl::identity{}) - 1;
+                                                          oneapi::dpl::identity{}, oneapi::dpl::identity{}) -
+              1;
     }
     return ret;
 }
--- a/include/oneapi/dpl/pstl/histogram_binhash_utils.h
+++ b/include/oneapi/dpl/pstl/histogram_binhash_utils.h
@@ -90,8 +90,8 @@ __custom_boundary_get_bin_helper(_Acc __acc, ::std::int32_t __size, _T2 __value,
     std::int32_t ret = -1;
     if (__value >= __min && __value < __max)
     {
-        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, __value, std::less<_T2>{}, // TODO __pstl_upper_bound - value (maybe processed as pointer dereference)
-                                                          oneapi::dpl::identity{}) - 1;
+        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, // TODO __pstl_upper_bound - value (maybe processed as pointer dereference)
+                                                          &__value, std::less<_T2>{}, oneapi::dpl::identity{}, oneapi::dpl::identity{}) - 1;
     }
     return ret;
 }
--- a/include/oneapi/dpl/pstl/histogram_binhash_utils.h
+++ b/include/oneapi/dpl/pstl/histogram_binhash_utils.h
@@ -90,8 +90,8 @@ __custom_boundary_get_bin_helper(_Acc __acc, ::std::int32_t __size, _T2 __value,
     std::int32_t ret = -1;
     if (__value >= __min && __value < __max)
     {
-        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, // TODO __pstl_upper_bound - value (maybe processed as pointer dereference)
-                                                          &__value, std::less<_T2>{}, oneapi::dpl::identity{}, oneapi::dpl::identity{}) - 1;
+        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, &__value, std::less<_T2>{},
+                                                          oneapi::dpl::identity{}, oneapi::dpl::identity{}) - 1;
     }
     return ret;
 }
--- a/include/oneapi/dpl/pstl/histogram_binhash_utils.h
+++ b/include/oneapi/dpl/pstl/histogram_binhash_utils.h
@@ -92,8 +92,7 @@ __custom_boundary_get_bin_helper(_Acc __acc, ::std::int32_t __size, _T2 __value,
     {
         _T2* __value_ptr = &__value;
         ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, __value_ptr, std::less<_T2>{},
-                                                          oneapi::dpl::identity{}, oneapi::dpl::identity{}) -
-              1;
+                                                          oneapi::dpl::identity{}, oneapi::dpl::identity{}) - 1;
     }
     return ret;
 }
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -668,8 +668,7 @@ __pstl_upper_bound(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2, 
     __reorder_pred<_Compare> __reordered_comp{__comp};
     __not_pred<decltype(__reordered_comp)> __negation_reordered_comp{__reordered_comp};
 
-    return __pstl_lower_bound(__rng1, __first1, __last1, __rng2, __rng2_idx, __negation_reordered_comp, __proj1,
-                              __proj2);
+    return __pstl_lower_bound(__rng1, __first1, __last1, __rng2, __rng2_idx, __negation_reordered_comp, __proj1, __proj2);
 }
 
 template <typename _Rng1, typename _Size1, typename _Rng2It, typename _Compare, typename _Proj1, typename _Proj2>
@@ -731,8 +730,7 @@ __biased_lower_bound(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2
     if (__n > 0)
     {
         // End up fully at binary search
-        return oneapi::dpl::__internal::__pstl_lower_bound(__rng1, __first1, __last1, __rng2, __rng2_idx, __comp,
-                                                           __proj1, __proj2);
+        return oneapi::dpl::__internal::__pstl_lower_bound(__rng1, __first1, __last1, __rng2, __rng2_idx, __comp, __proj1, __proj2);
     }
     return __first1;
 }
@@ -746,8 +744,7 @@ __biased_upper_bound(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2
     __reorder_pred<_Compare> __reordered_comp{__comp};
     __not_pred<decltype(__reordered_comp)> __negation_reordered_comp{__reordered_comp};
 
-    return __biased_lower_bound<__bias_last>(__rng1, __first1, __last1, __rng2, __rng2_idx, __negation_reordered_comp,
-                                             __proj1, __proj2);
+    return __biased_lower_bound<__bias_last>(__rng1, __first1, __last1, __rng2, __rng2_idx, __negation_reordered_comp, __proj1, __proj2);
 }
 
 template <typename _IntType, typename _Acc>
--- a/include/oneapi/dpl/pstl/utils.h
+++ b/include/oneapi/dpl/pstl/utils.h
@@ -778,14 +778,15 @@ struct _ReverseCounter
 // Reverse searching for the first element strongly less than a passed value - left bound
 template <typename _Buffer, typename _Index, typename _Value, typename _Compare, typename _Proj>
 _Index
-__pstl_left_bound(_Buffer& __a, _Index __first, _Index __last, const _Value& __val, _Compare __comp, _Proj __proj)
+__pstl_left_bound(_Buffer& __a, _Index __first, _Index __last, _Value&& __value_proj, _Compare __comp, _Proj __proj)
 {
     auto __beg = _ReverseCounter<_Index, _Buffer>{__last - 1};
     auto __end = _ReverseCounter<_Index, _Buffer>{__first - 1};
 
     __not_pred<decltype(__comp)> __negation_comp{__comp};
 
-    return __pstl_lower_bound(__a, __beg, __end, __val, __negation_comp, __proj);
+    // We should forward __value_proj to preserve their value category
+    return __pstl_lower_bound(__a, __beg, __end, std::forward<_Value>(__value_proj), __negation_comp, __proj);
 }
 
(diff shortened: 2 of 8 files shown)


include/oneapi/dpl/pstl/histogram_binhash_utils.h
```diff
@@ -90,8 +90,8 @@ __custom_boundary_get_bin_helper(_Acc __acc, ::std::int32_t __size, _T2 __value,
     std::int32_t ret = -1;
     if (__value >= __min && __value < __max)
     {
-        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, __value, std::less<_T2>{},
-                                                          oneapi::dpl::identity{}) - 1;
+        ret = oneapi::dpl::__internal::__pstl_upper_bound(__acc, std::int32_t{0}, __size, &__value, std::less<_T2>{},
```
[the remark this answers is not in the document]
SergeyKopienko: ### **Declined**

I believe this comment is incorrect.
Justification: now we have two overloads of `__pstl_upper_bound` function:
1) overload with `8` arguments, no any default values ant etc.:
```C++
// __rng1[__first, __last), __rng2(__rng2_idx), __comp, __proj1, __proj2
template <typename _Rng1, typename _Size1, typename _Size2, typename _Rng2, typename _Compare, typename _Proj1,
          typename _Proj2>
_Size1
__pstl_upper_bound(_Rng1 __rng1, _Size1 __first, _Size1 __last, _Rng2 __rng2, _Size2 __rng2_idx, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

2) overload with `7` arguments, no any default values ant etc.:
```C++
// __rng1[__first, __last), __iterator, __comp, __proj1, __proj2
template <typename _Rng1, typename _Size1, typename _Iterator, typename _Compare, typename _Proj1, typename _Proj2>
_Size1
__pstl_upper_bound(_Rng1 __rng1, _Size1 __first, _Size1 __last, _Iterator __iterator, _Compare __comp, _Proj1 __proj1, _Proj2 __proj2);
```

So I believe in this call we calls overload (2) with `7` parameters.

include/oneapi/dpl/pstl/utils.h
```diff
@@ -623,44 +623,93 @@ __pstl_lower_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __valu
...
+// __rng1[__first1, __last1), __rng2(__rng2_idx), __comp, __proj1, __proj2
+template <typename _Rng1, typename _Size1, typename _Size2, typename _Rng2, typename _Compare, typename _Proj1,
+          typename _Proj2>
+_Size1
+__pstl_lower_bound(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2, _Size2 __rng2_idx, _Compare __comp,
+                   _Proj1 __proj1, _Proj2 __proj2)
+{
+    return __pstl_lower_bound_impl(
+        __first1, __last1, [__rng1, __rng2, __rng2_idx, __comp, __proj1, __proj2](_Size1 __rng1_idx) mutable {
+            return std::invoke(__comp, std::invoke(__proj1, __rng1[__rng2_idx]), std::invoke(__proj2, __rng2[__rng2_idx]));
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/utils.h
```diff
@@ -672,36 +721,41 @@ __biased_lower_bound(_Acc __acc, _Size1 __first, _Size1 __last, const _Value& __
...
     }
     if (__n > 0)
     {
-        //end up fully at binary search
-        return oneapi::dpl::__internal::__pstl_lower_bound(__acc, __first, __last, __value, __comp, __proj);
+        // End up fully at binary search
+        return oneapi::dpl::__internal::__pstl_lower_bound(__rng1, __first1, __last1, __rng2, __rng2_idx, __comp, __proj1,
+                                                           __proj2);
     }
     return __first;
```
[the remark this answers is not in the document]

include/oneapi/dpl/pstl/utils.h
```diff
@@ -634,33 +635,73 @@ __pstl_lower_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __valu
...
+                                   [__rng1, __rng2, __rng2_idx, __comp, __proj1, __proj2](_Size1 __rng1_idx) mutable {
+                                       return std::invoke(__comp, std::invoke(__proj1, __rng1[__rng1_idx]),
+                                                          std::invoke(__proj2, __rng2[__rng2_idx]));
+                                   });
+}
+
+template <typename _Rng1, typename _Size1, typename _Rng2It, typename _Compare, typename _Proj1, typename _Proj2>
+_Size1
+__pstl_lower_bound(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2It __rng2_it, _Compare __comp, _Proj1 __proj1,
+                   _Proj2 __proj2)
```
danhoeflinger, line 654: Its a little confusing to me to have these two flavors of this function.  Can't we just use the iterator version for all cases? (create the iterator by getting the beginning of the range and adding the index when we want to call with an index)

include/oneapi/dpl/pstl/utils.h
```diff
@@ -634,33 +635,73 @@ __pstl_lower_bound(_Acc __acc, _Size __first, _Size __last, const _Value& __valu
...
+template <typename _Rng1, typename _Size1, typename _Size2, typename _Rng2, typename _Compare, typename _Proj1,
+          typename _Proj2>
+_Size1
+__pstl_lower_bound(_Rng1 __rng1, _Size1 __first1, _Size1 __last1, _Rng2 __rng2, _Size2 __rng2_idx, _Compare __comp,
+                   _Proj1 __proj1, _Proj2 __proj2)
+{
+    return __pstl_lower_bound_impl(__first1, __last1,
+                                   [__rng1, __rng2, __rng2_idx, __comp, __proj1, __proj2](_Size1 __rng1_idx) mutable {
+                                       return std::invoke(__comp, std::invoke(__proj1, __rng1[__rng1_idx]),
+                                                          std::invoke(__proj2, __rng2[__rng2_idx]));
```
danhoeflinger, line 647: So, is there an advantage to calling project on this fixed element for *every* comparison rather than doing it once and saving / reusing the result?

Even with a lightweight projection, couldn't this add up to be a net performance loss (adding a small action to the innermost loop)?

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Initial idea - to get value from range by index or dereference iterator at one place, together with projection calls and comparator call.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: The current assumption - it doesn't meter to call one or two projections.
I believe if projection is very simple it would be optimized by compiler.
But from the idea of `ranges` / `projections` / `comparator` design this variant looks more correct.
@akukanov am I correct?

include/oneapi/dpl/pstl/utils.h
danhoeflinger: What I mean here is that the two overloads of `__pstl_lower_bound`, one which accepts range and index for the value, and another which accepts iterator pointing to the value itself are confusing. 

I would recommend one of two things:
1) just have a single function which accepts iterator pointing to the value itself, and when you have a range and index you just call `range.begin() + index` to use the single function overload.

-or-

2) at least call the range and index overload something different like `__pstl_lower_bound_by_index`.


I prefer (1), but I think the code as written is confusing and error prone to have the number of arguments to a heavily templated function like this determine the overload used between these two.

include/oneapi/dpl/pstl/utils.h
danhoeflinger: As discussed offline, it seems there are not many requirements for projections coming from the c++ standard (that I have been able to find).  This means that we need to be careful not to add any hidden requirements to the projected value.  Also, we don't have much of any guarantee that these are lightweight operations (though they generally are in practice). We should still try to minimize the number of times they are calculated redundantly.

The specification has requirements similar to the comparator for each projection, but we are not in danger of calculating the individual projections more times than the comparators so there is nothing new to consider there.

All this is to say:  
I would be careful to add redundant calculations of the projection operation unless we are solving some specific concrete issue that we can point to.  Is it possible to come up with an example of a projection / value / comparator combination where the code as written imposes some extra requirement in a way which is problematic? With such an example, this PRs' benefit becomes much more clear to me.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: Renamed: for index-based versions of functions added `_idx` postfix into name:

`__pstl_lower_bound` -> `__pstl_lower_bound_idx`
`__pstl_upper_bound` -> `__pstl_upper_bound_idx`
`__pstl_right_bound` -> `__pstl_right_bound_idx`

`__biased_lower_bound` -> `__biased_lower_bound_idx`
`__biased_upper_bound` -> `__biased_upper_bound_idx`

`__pstl_left_bound` -> `__pstl_left_bound_idx`

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: PR description updated: added examples.

danhoeflinger: The example provided have compiler errors because of the comparator not accepting `const T&` arguments.  According to cppreference, a comparator is required to accept `const T&` arguments (even if this is not exactly the signature).

https://en.cppreference.com/w/cpp/algorithm/merge.html
```
The signature of the comparison function should be equivalent to the following:

bool cmp(const Type1& a, const Type2& b);

While the signature does not need to have const&, the function must not modify the objects passed to it and must be able to accept all values of type (possibly const) Type1 and Type2 regardless of [value category](https://en.cppreference.com/w/cpp/language/value_category.html) (thus, Type1& is not allowed, nor is Type1 unless for Type1 a move is equivalent to a copy(since C++11)).
The types Type1 and Type2 must be such that objects of types InputIt1 and InputIt2 can be dereferenced and then implicitly converted to both Type1 and Type2.​
```

I'm not immediately finding the same language in the C++ draft specification, but I assume it is somewhere.  Its also possible we also may want to be more permissive than the specification.

SergeyKopienko: As I noted, it's synthetic example.

akukanov: > According to cppreference, a comparator is required to accept `const T&` arguments (even if this is not exactly the signature)

For `std::ranges::merge`, I think there is no such requirement. The `mergeable` concept is defined via `std::indirect_strict_weak_order<Comp, std::projected<I1, Proj1>, std::projected<I2, Proj2>>`, and `indirect_strict_weak_order` requires the comparator to accept the result of projections applied to an iterator. See https://en.cppreference.com/w/cpp/iterator/indirect_strict_weak_order.html, where `/*indirect-value-t*/<I1>` and `std::iter_reference_t<I1>` is the same thing if you dig in:

> The exposition-only alias template indirect-value-t denotes the following:
> - `std::invoke_result_t<Proj&, /*indirect-value-t*/<I>>` if T is the same as std::projected<I, Proj> for some types I and Proj.

and 

> `template< /*dereferenceable*/ T > using iter_reference_t = decltype(*std::declval<T&>());`

where for `std::projected` dereferencing is declared as `std::indirect_result_t<Proj&, I> operator*() const` which is the same as `std::invoke_result_t` after dereferencing `I`.

If you further unroll `indirect_strict_weak_order` you will eventually see `regular_invocable` with a semantic requirement not to modify the arguments (https://en.cppreference.com/w/cpp/concepts/invocable.html), but syntactically the comparator is only required to take the result of applying a projection I believe. So Sergey's examples are technically valid, which can be proven by adding e.g. `static_assert(std::indirect_strict_weak_order<UserCmp, std::projected<decltype(v.begin()), UserProj<int>>, std::projected<decltype(v.begin()), UserProj<int>>>);` or/and trying the same predicates & projections with `std::ranges::merge`.

danhoeflinger: OK, I just wanted to make sure we aren't making things potentially slower without something real we are addressing, but I am convinced that there is something to address here.

test/general/implementation_details/balanced_path_unit_tests.pass.cpp
```diff
@@ -239,8 +239,8 @@ bool
 test_right_biased_lower_bound_impl(Rng __rng, std::size_t __location, std::less<typename Rng::value_type> __comp)
 {
     auto expected_res = std::lower_bound(__rng.begin(), __rng.begin() + __location, __rng[__location], __comp);
-    auto res = oneapi::dpl::__internal::__biased_lower_bound</*last_biased=*/true>(
-        __rng.begin(), std::size_t{0}, __location, __rng[__location], __comp, oneapi::dpl::identity{});
+    auto res = oneapi::dpl::__internal::__biased_lower_bound_idx</*last_biased=*/true>(
+        __rng.begin(), std::size_t{0}, __location, __rng.begin(), __location, __comp, oneapi::dpl::identity{}, oneapi::dpl::identity{});
```
danhoeflinger: do you need this `begin()` call?  this is a range already right?

danhoeflinger:
I think I'd prefer some of the clang format suggestions, but I wont insist.  When it is close, I prefer the suggestions from the tool mostly so that it is easier to maintain formatting for future PRs involving these files. However, you can do what you wish with the formatting.

Largely this looks good to me other than the unresolved comment I posted.

test/general/implementation_details/balanced_path_unit_tests.pass.cpp
SergeyKopienko: Good catch, Dan, thanks!
It will works in this form too, but fixed to more correct and readable form without `begin()` call.

include/oneapi/dpl/pstl/utils.h
SergeyKopienko: The `static_assert` check has been added: 
- examples with initial compile error: https://godbolt.org/z/hj6YMv6rq, https://godbolt.org/z/f9oPGbbeM
- example with new implementation from this PR without compile error: https://godbolt.org/z/aq5GxWhej

The code:
```C++
    UserCmp cmp;
    UserProj<int> proj;

    static_assert(
        std::indirect_strict_weak_order<decltype(cmp),
        std::projected<decltype(v.begin()), decltype(proj)>,
        std::projected<decltype(v.begin()), decltype(proj)>>); 
```

And this `static_assert` check really does not triggered.

PR #2462, opened 2025-09-17
Fix <range> inclusion in range memory tests

dmitriy-sobolev:
The initial motivation is to avoid inclusion of <oneapi/dpl/ranges> everywhere, even with c++17. 
Second, its inclusion is not needed, and can be replaced with just <ranges>.


SergeyKopienko (approved):
LGTM

PR #2465, opened 2025-09-18
Histogram: workaround kernel compilation issues with icpx 2025.3

dmitriy-sobolev:
icpx 2025.3 (latest nightly build version) fails to compile `__histogram_general_registers_local_reduction` kernel for CPU devices.

The issue has already been reported, but a potential fix will not likely make it into the release. A quick solution is to fall back to a less optimized kernel. 

The fall back is done on demand by defining `ONEDPL_DISABLE_HISTOGRAM_REGISTER_REDUCTION` to a non-zero value.
The plan is to mention this macro in the release notes as a workaround with no promise of having it for the future releases.

### DIFF
--- a/documentation/release_notes.rst
+++ b/documentation/release_notes.rst
@@ -8,6 +8,17 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
 and provides high-productivity APIs aimed to minimize programming efforts of C++ developers
 creating efficient heterogeneous applications.
 
+New in 2022.10.0
+===============
+
+Known Issues and Limitations
+----------------------------
+New in This Release
+^^^^^^^^^^^^^^^^^^^
+- Calling ``histogram`` algorithm with a device execution policy may cause a segmentation fault in
+  Intel® oneAPI DPC++/C++ Compiler 2025.3 when building a program for CPU devices.
+  To avoid this, define ``ONEDPL_DISABLE_HISTOGRAM_REGISTER_REDUCTION`` macro to a non-zero value.
+
 New in 2022.9.0
 ===============
 
--- a/documentation/release_notes.rst
+++ b/documentation/release_notes.rst
@@ -16,7 +16,7 @@ Known Issues and Limitations
 New in This Release
 ^^^^^^^^^^^^^^^^^^^
 - Calling ``histogram`` algorithm with a device execution policy may cause a segmentation fault in
-  Intel® oneAPI DPC++/C++ Compiler 2025.3 when building a program for CPU devices.
+  Intel® oneAPI DPC++/C++ Compiler 2025.3 when compiling SYCL kernels for CPU devices.
   To avoid this, define ``ONEDPL_DISABLE_HISTOGRAM_REGISTER_REDUCTION`` macro to a non-zero value.
 
 New in 2022.9.0
--- a/documentation/release_notes.rst
+++ b/documentation/release_notes.rst
@@ -9,7 +9,7 @@ and provides high-productivity APIs aimed to minimize programming efforts of C++
 creating efficient heterogeneous applications.
 
 New in 2022.10.0
-===============
+================
 
 Known Issues and Limitations
 ----------------------------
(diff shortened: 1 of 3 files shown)


danhoeflinger (dismissed):
I think this seems like a reasonable workaround. LGTM
Thanks!

include/oneapi/dpl/pstl/onedpl_config.h
```diff
@@ -348,4 +348,14 @@
...
+// DPC++ 2025.3 fails with a segfault when compiling __histogram_general_registers_local_reduction kernel
+// for CPU devices
+// Allow manually enabling this kernel for more performant execution on GPU devices.
+#if !defined(__INTEL_LLVM_COMPILER) || (__INTEL_LLVM_COMPILER < 20250300 || __INTEL_LLVM_COMPILER >= 20250400) || \
+    defined(ONEDPL_ENABLE_HISTOGRAM_REGISTER_REDUCTION)
+#    define _ONEDPL_ENABLE_HISTOGRAM_REGISTER_REDUCTION 1
+#else
+#    define _ONEDPL_ENABLE_HISTOGRAM_REGISTER_REDUCTION 0
+#endif
+
```
akukanov: I think I would prefer to provide an opt-out macro to avoid the observed problem rather than an opt-in macro to enable "partially safe" code.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
```diff
@@ -514,8 +515,13 @@ __parallel_histogram_select_kernel(sycl::queue& __q, const sycl::event& __init_e
             __q, __init_event, __work_group_size, std::forward<_Range1>(__input), std::forward<_Range2>(__bins),
             __binhash_manager));
     }
+    else if
+#else
+    if
+#endif
+        (__num_bins * sizeof(_local_histogram_type) +
```
akukanov:
```suggestion
    else
#endif
    if (__num_bins * sizeof(_local_histogram_type) +
```

SergeyKopienko (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_histogram.h
```diff
@@ -495,7 +495,6 @@ __future<sycl::event>
 __parallel_histogram_select_kernel(sycl::queue& __q, const sycl::event& __init_event, _Range1&& __input,
                                    _Range2&& __bins, const _BinHashMgr& __binhash_manager)
 {
-    using _private_histogram_type = ::std::uint16_t;
     using _local_histogram_type = ::std::uint32_t;
```
akukanov: Drop the leading colons in the next line too? That's the only line in the function still having those.

include/oneapi/dpl/pstl/onedpl_config.h
dmitriy-sobolev: Done. With that opt-out logic there is no need for checking the compiler version, what I've removed.

akukanov (approved):
Looks good to me now.

documentation/release_notes.rst
```diff
@@ -8,6 +8,17 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
 and provides high-productivity APIs aimed to minimize programming efforts of C++ developers
 creating efficient heterogeneous applications.
 
+New in 2022.10.0
+===============
```
akukanov:
```suggestion
New in 2022.10.0
================
```

documentation/release_notes.rst
```diff
@@ -8,6 +8,17 @@ The Intel® oneAPI DPC++ Library (oneDPL) accompanies the Intel® oneAPI DPC++/C
...
+
+Known Issues and Limitations
+----------------------------
+New in This Release
+^^^^^^^^^^^^^^^^^^^
+- Calling ``histogram`` algorithm with a device execution policy may cause a segmentation fault in
+  Intel® oneAPI DPC++/C++ Compiler 2025.3 when compiling SYCL kernels for CPU devices.
+  To avoid this, define ``ONEDPL_DISABLE_HISTOGRAM_REGISTER_REDUCTION`` macro to a non-zero value.
```
danhoeflinger: "prior to including oneDPL header files"?

Also, do we need to put this into CMAKE / cmake readme? (I'm not sure)

dmitriy-sobolev: I'd like to limit it to the release notes as a workaround for a niche problem, which will likely go away with the next compiler update.

> "prior to including oneDPL header files"?

Will mention it.

dmitriy-sobolev: Mentioned

timmiesmith (approved):
LGTM. Thank you.

timmiesmith:
I've checked the clang-format suggestions and think they should not be applied.

PR #2466, opened 2025-09-18
[Tests] Permutation Iterator tests, directly assign to usm_shared for indexes

danhoeflinger:
This PR removes complexity from the permutation iterator tests by directly assigning to usm_shared index data rather than copying from a host `std::vector` and copying via a `queue.copy` call.  This reduces complexity but also works around what seems to be a bug in the sycl runtime, where usm_shared memory assigned on the device is used directly within a parallel section as the first usage on the host resulting in a segfault.


SergeyKopienko (approved):
LGTM (as workaround)

mmichel11 (approved):
LGTM

PR #2470, opened 2025-09-22
Workaround an issue with drop_view and libstdc++10 in views test

dmitriy-sobolev:
One more case in the tests with the issue:

```c++
n file included from /usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/string:54:
/usr/lib/gcc/x86_64-linux-gnu/10/../../../../include/c++/10/bits/range_access.h:981:6: error: SYCL kernel cannot use exceptions
  981 |             throw "attempt to decrement a non-bidirectional iterator";

```


dmitriy-sobolev:
@timmiesmith, can we merge this PR?

PR #2471, opened 2025-09-22
Normalize ranges order in `struct __brick_includes::operator()`

SergeyKopienko:


include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -129,20 +129,41 @@ using projected_value_t = std::remove_cvref_t<std::invoke_result_t<Proj&, std::i
 namespace __ranges
 {
 
-struct __first_size_calc
+template <std::size_t _RngIndex>
+struct __eval_rng_size
```
dmitriy-sobolev: What do you think about it? It will help with #2445, and it's a bit simpler. 

```c++
template <typename _Range>
constexpr auto
__size(const _Range& __range)
{
        return std::ranges::size(__range);
#else
        return __range.size();
#endif
}

template <std::size_t _RangeIndex>
struct __nth_range_size
{
    template<typename... _Ranges>
    auto operator()(const _Ranges&... __ranges)
    {
        static_assert(_RangeIndex < sizeof...(_Ranges));
        return __size(std::get<_RangeIndex>(std::tie(__ranges...)));
    }
};

using __first_size_calc = __nth_range_size<0>;
using __second_size_calc = __nth_range_size<1>;
```

SergeyKopienko: 1) My assumption - this approach look simple but I think the approach from this PR should work more faster in run-time due compiler optimizations. 
2) I think the approach when we get required item from some `pack` by some recursive call is widely-used.
3) The approach from this PR works without any temporary objects at all: it may have sense if we going to use this functional inside Kernel code - why not?
So I propose to save implementation as is.


@danhoeflinger what is your opinion?

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -129,12 +129,31 @@ using projected_value_t = std::remove_cvref_t<std::invoke_result_t<Proj&, std::i
...
-    template <typename _Range, typename... _Ranges>
+    return std::forward<_Type>(__value);
+}
+
+template <std::size_t _Index, typename... _Args>
+decltype(auto)
+__get_pack_item(_Args&&... __args)
+{
+    return __pick_item<_Index>(std::forward<_Args>(__args)...);
+}
```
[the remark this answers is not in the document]
SergeyKopienko: It's depends on it's usage: in fold-expression it's working ok

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -128,21 +128,45 @@ using projected_value_t = std::remove_cvref_t<std::invoke_result_t<Proj&, std::i
...
 #if _ONEDPL_CPP20_RANGES_PRESENT
-        return std::ranges::size(__rng);
+    return std::ranges::size(__rng);
 #else
-        return __rng.size();
+    return __rng.size();
 #endif
+}
+
+
```
dmitriy-sobolev: Extra line

```suggestion
```

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -128,21 +128,45 @@ using projected_value_t = std::remove_cvref_t<std::invoke_result_t<Proj&, std::i
...
+template <std::size_t _RngIndex>
+struct __nth_range_size
+{
+  private:
+    template <std::size_t _RngIndexCurrent, typename _Range, typename... _Ranges>
+    auto
+    __eval_rng_size_impl(const _Range& __rng, const _Ranges&... __rngs) const
```
dmitriy-sobolev:
```suggestion
    __eval_rng_size_impl(const _Range& __rng, const _Ranges&... __rngs) const
```
```suggestion
    __nth_range_size_impl(const _Range& __rng, const _Ranges&... __rngs) const
```

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -128,21 +128,45 @@ using projected_value_t = std::remove_cvref_t<std::invoke_result_t<Proj&, std::i
...
+            return __eval_rng_size_impl<_RngIndexCurrent + 1>(__rngs...);
+    }
+
+  public:
+    template <typename... _Ranges>
+    auto
+    operator()(const _Ranges&... __rngs) const
+    {
+        static_assert(_RngIndex < sizeof...(_Ranges));
+        return __eval_rng_size_impl<0>(__rngs...);
```
dmitriy-sobolev:
```suggestion
        return __nth_range_size_impl<0>(__rngs...);
```

include/oneapi/dpl/pstl/utils_ranges.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/utils_ranges.h
SergeyKopienko: Fixed

PR #2475, opened 2025-09-23
Fix range iterators usage after range perfect forwarding

SergeyKopienko:
This PR fixes an issue where range iterators were being used after the range objects had been moved, which could lead to undefined behavior. It's looked not quite correct. The changes remove unnecessary `std::forward` calls when passing range parameters that need to remain valid for subsequent iterator operations.

- Remove `std::forward` calls for range parameters that are accessed after being passed to functions
- Ensure range objects remain valid when their iterators are used later in the same scope
- Apply consistent parameter passing pattern across multiple algorithm implementations

### DIFF
--- a/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
+++ b/include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
@@ -1102,7 +1102,7 @@ struct __reverse_fn
         auto __last = std::ranges::begin(__r) + std::ranges::size(__r);
 
         oneapi::dpl::__internal::__ranges::__pattern_reverse(__dispatch_tag, std::forward<_ExecutionPolicy>(__exec),
-                                                             std::forward<_R>(__r));
+                                                             __r);
         return std::ranges::borrowed_iterator_t<_R>{__last};
     }
 
@@ -1171,8 +1171,8 @@ struct __is_sorted_until_fn
     {
         auto __last = std::ranges::begin(__r) + std::ranges::size(__r);
 
-        auto __it = oneapi::dpl::ranges::adjacent_find(std::forward<_ExecutionPolicy>(__exec), std::forward<_R>(__r),
-            oneapi::dpl::__internal::__reorder_pred<_Comp>(__comp), __proj);
+        auto __it = oneapi::dpl::ranges::adjacent_find(std::forward<_ExecutionPolicy>(__exec), __r,
+                                                       oneapi::dpl::__internal::__reorder_pred<_Comp>(__comp), __proj);
         return __it == __last ? __last : ++__it;
     }
 }; //__is_sorted_until_fn
--- a/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
@@ -936,7 +936,7 @@ __pattern_merge_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exe
         std::ranges::borrowed_iterator_t<_OutRange>>;
 
     return __return_t{std::ranges::begin(__r1) + __res.first, std::ranges::begin(__r2) + __res.second,
-            std::ranges::begin(__out_r) + __n_out};
+                      std::ranges::begin(__out_r) + __n_out};
 }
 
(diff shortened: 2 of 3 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -1075,15 +1077,14 @@ __pattern_set_difference(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __e
...
+            oneapi::dpl::__internal::__brick_copy<__hetero_tag<_BackendTag>>{}, __r1, __out_r);
 
         return {__first1 + __n1, __result + __idx};
     }
 
     const std::size_t __result_size = __par_backend_hetero::__parallel_set_op<unseq_backend::_DifferenceTag>(
-        _BackendTag{}, unseq_backend::_DifferenceTag{}, std::forward<_ExecutionPolicy>(__exec), std::forward<_R1>(__r1),
-        std::forward<_R2>(__r2), std::forward<_OutRange>(__out_r), __comp, __proj1, __proj2);
+        _BackendTag{}, unseq_backend::_DifferenceTag{}, std::forward<_ExecutionPolicy>(__exec), __r1,
+        std::forward<_R2>(__r2), __out_r, __comp, __proj1, __proj2);
```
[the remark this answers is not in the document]
SergeyKopienko: Declined: the iterator `const auto __first1 = std::ranges::begin(__r1);` is used after this call: the usage of range-iterator not quite correct after moving `__r1` into `__pattern_walk_n` function.

include/oneapi/dpl/pstl/glue_algorithm_ranges_impl.h
```diff
@@ -1099,11 +1099,9 @@ struct __reverse_fn
     {
         const auto __dispatch_tag = oneapi::dpl::__ranges::__select_backend(__exec);
 
-        auto __last = std::ranges::begin(__r) + std::ranges::size(__r);
-
         oneapi::dpl::__internal::__ranges::__pattern_reverse(__dispatch_tag, std::forward<_ExecutionPolicy>(__exec),
-                                                             std::forward<_R>(__r));
-        return std::ranges::borrowed_iterator_t<_R>{__last};
+                                                             __r);
+        return std::ranges::begin(__r) + std::ranges::size(__r);
```
dmitriy-sobolev: Would not the original approach be better? If the range is an r-value, it is moved (not copied), and the iterator type to return would have been `std::ranges::dangling`.

In other words, I see no issues with the original approach with returning an iterator to the forwarded range.

dmitriy-sobolev: The range is not copied and there is no invalid state of the iterator, so the change looks fine.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -1089,15 +1086,14 @@ __pattern_set_difference(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __e
...
```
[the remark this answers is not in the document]
SergeyKopienko: Declined

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -936,7 +936,7 @@ __pattern_merge_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exe
         std::ranges::borrowed_iterator_t<_OutRange>>;
 
     return __return_t{std::ranges::begin(__r1) + __res.first, std::ranges::begin(__r2) + __res.second,
-                      std::ranges::begin(__out_r) + __n_out};
+            std::ranges::begin(__out_r) + __n_out};
```
dmitriy-sobolev: Reverting a space-only change: 

```suggestion
                      std::ranges::begin(__out_r) + __n_out};
```

SergeyKopienko: Done

dmitriy-sobolev (dismissed):
LGTM. Please wait for the CI to succeed before merging.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
SergeyKopienko: Applied

PR #2477, opened 2025-09-23
Updating scripts to correct their location in the oneDPL directory structure

### LINKED ISSUE #2468: Incorrect root and include paths after calling integration/env/vars.sh
**Describe the Bug:**
The `integration/env/vars.sh` script sets the `DPL_ROOT` and other dependent environment variables to the wrong path: `.../oneDPL/integration/...` instead of `.../oneDPL/...`

**To Reproduce:**
```
source .../oneDPL/integration/env/vars.sh && echo ${DPL_ROOT} && echo ${CPLUS_INCLUDE_PATH}
```
which will output:
```
.../oneDPL/integration
.../oneDPL/integration/include
```

**Expected Behavior:**
Folder `integration` should not be present in the path:
```
.../oneDPL/
.../oneDPL/include
```

**Additional Context:**
It looks like root is set to one folder above the script folder, which was correct with the old setup script `env/onedplvars.sh`, but is off one level now that the script was moved to `integration/env/vars.sh`.

timmiesmith:
This PR addresses the issue reported in #2468.  Currently the `DPL_ROOT` and `DPLROOT` directories are set to `oneDPL\integration` with results in the paths added to CMake and pkgconfig paths to also be wrong.  This PR fixes #2468.


integration/env/vars.bat
```diff
@@ -17,8 +17,9 @@ REM
 
 set "VARSDIR=%~dp0"
 
-if not defined DPL_ROOT for /f "delims=" %%F in ("%VARSDIR%..") do set "DPL_ROOT=%%~fF"
-if not defined DPLROOT for /f "delims=" %%F in ("%VARSDIR%..") do set "DPLROOT=%%~fF"
+REM The vars script is in the integration\env directory. The DPL root directory is up two levels
+if not defined DPL_ROOT for /f "delims=" %%F in ("%VARSDIR%..\..") do set "DPL_ROOT=%%~fF"
```
dmitriy-sobolev: We should be careful here. When creating a oneDPL package for the oneAPI installer, we move these files, for example, `integration/env/vars.sh` is moved to `env/vars.sh`.

Possible options on what to do:

1. Move everything from the integration folder to oneDPL root in the repository to align the layout of the installed package and the repository.
2. Detect whether this is an installed package or the repository in this script and set the path accordingly.
3. Advice against using the scripts in the integration folder.

danhoeflinger: Yeah, I think (1) makes the most sense to me.  It is the simplest and least risky in my opinion.

dmitriy-sobolev: One more option complementing (3), which is even less risky than (1):

4. Advice against using the scripts in the integration folder in favor of cmake install, e.g.:
```bash
cmake -DCMAKE_INSTALL_PREFIX=<path/to/install/dir>  <path/to/oneDPL-source>
cmake --install . # Now the installation root has env/vars.sh or env/vars.bat depending on the OS
```

However, I like (1) more due to the less steps on the user side. Also, I still need to read cmake install in detail to make sure nothing is missed.

timmiesmith: I've confirmed @dmitriy-sobolev's issue.  In oneDPL packages the env directory is in the top-level oneDPL directory.  Moving the env directory to the top level would make the repository consistent with the oneAPI packages. I'll update the PR.

timmiesmith: I'm only going to move `env` to the top level.  The `pkgconfig` and `etc` directories in `integration` are explicitly for oneAPI integration and aren't anything we would expect users of the oneDPL repo directly to use.

PR #2479, opened 2025-09-24
Fix duplicate kernel name issue in __parallel_transform_reduce_impl with differing init types

mmichel11:
When `oneapi::dpl::reduce` is called multiple times with a device policy and the only difference in the calls is the type of the initial value, different kernels are generated but we use the same kernel name. This results in a compilation error.

To fix this issue, we need to add the initial value type `_Tp` as a template argument to the call to `__kernel_name_generator` when generating the reduction kernel name.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
@@ -305,7 +305,7 @@ struct __parallel_transform_reduce_impl
     {
         using _NoOpFunctor = unseq_backend::walk_n<oneapi::dpl::identity>;
         using _ReduceKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-            __reduce_kernel, _CustomName, _ReduceOp, _TransformOp, _NoOpFunctor, _Ranges...>;
+            __reduce_kernel, _CustomName, _Tp, _ReduceOp, _TransformOp, _NoOpFunctor, _Ranges...>;
 
         auto __transform_pattern1 =
             unseq_backend::transform_reduce<_ReduceOp, _TransformOp, _Tp, _Commutative, _VecSize>{__reduce_op,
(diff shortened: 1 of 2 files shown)


test/parallel_api/numeric/numeric.ops/reduce.pass.cpp
```diff
@@ -111,6 +111,22 @@ test_short_forms()
     }
 }
 
+void
+test_differing_init_type()
+{
+    using InitType = std::uint64_t;
+    using InputType = std::uint32_t;
+    // Test is designed to check compilation issues, so only a single test run is sufficient.
+    std::size_t n = 10042;
```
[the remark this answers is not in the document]
mmichel11: I don't think we need a named constant here.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl_reduce.h
```diff
@@ -305,7 +305,7 @@ struct __parallel_transform_reduce_impl
     {
         using _NoOpFunctor = unseq_backend::walk_n<oneapi::dpl::identity>;
         using _ReduceKernel = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_generator<
-            __reduce_kernel, _CustomName, _ReduceOp, _TransformOp, _NoOpFunctor, _Ranges...>;
+            __reduce_kernel, _CustomName, _Tp, _ReduceOp, _TransformOp, _NoOpFunctor, _Ranges...>;
```
danhoeflinger, line 308: As mentioned offline, I think that we may want to consider adding `_Commutative` and `_VecSize` to the list, even if it is not explicitly required based upon the current usage of `__parallel_transform_reduce_impl`. 

Perhaps there is some bigger downside to further growing this type list, but it seems that we are laying a trap for future possible usages of this struct as is.

mmichel11: From my perspective, a decision to change `VecSize` would be tightly coupled to the input range type itself (its size in particular to optimize sub-group loads), so I don't imagine we are likely to run into scenarios where `VecSize` must be embedded into the kernel name. Similarly, commutativity is decided on a per algorithm basis, and I don't imagine that will change.

Since the kernel name is already very long, my opinion is that we should not make even longer at this time.

PR #2480, opened 2025-09-25
Revert the workaround for kernel compilation issues in histogram

dmitriy-sobolev:
The compiler devs have already fixed the issue workarounded by #2465. The fix going to be cherry-picked into icpx 2025.3.
The documentation change is better to be done separately to target oneDPL 2022.10.


danhoeflinger (approved):
LGTM

PR #2481, opened 2025-09-25
Ranges stuff refactoring

SergeyKopienko:
This PR refactors the ranges staff implementation by removing explicit `std::ranges::borrowed_iterator_t` and `std::ranges::borrowed_subrange_t` wrapper constructions in favor of direct returns or brace initialization. The changes simplify the code while maintaining the same return types through template argument deduction.

- Removes explicit type wrapper constructions for borrowed iterators and subranges
- Replaces `auto` return types with explicit type annotations for better API clarity
- Simplifies return statements using brace initialization instead of explicit constructor calls
- Avoid multiple `std::ranges::begin`, `std::ranges::end` and `std::ranges::size` calls

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
@@ -497,10 +497,10 @@ __pattern_search_n(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _
     auto __idx = oneapi::dpl::__internal::__ranges::__pattern_search_n(__tag, std::forward<_ExecutionPolicy>(__exec),
         oneapi::dpl::__ranges::views::all_read(__r), __count, __value, __pred_2);
 
-    auto __first = std::ranges::begin(__r);
-    auto __end = (__idx == __r.size() ? __first + __idx : __first + __idx + __count);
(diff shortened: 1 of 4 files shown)


include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -495,8 +497,10 @@ __pattern_search_n(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _
     auto __idx = oneapi::dpl::__internal::__ranges::__pattern_search_n(__tag, std::forward<_ExecutionPolicy>(__exec),
         oneapi::dpl::__ranges::views::all_read(__r), __count, __value, __pred_2);
 
-    auto __end = (__idx == __r.size() ? std::ranges::begin(__r) + __idx : std::ranges::begin(__r) + __idx + __count);
-    return std::ranges::borrowed_subrange_t<_R>(std::ranges::begin(__r) + __idx, __end);
+    auto __first = std::ranges::begin(__r);
+    auto __end = (__idx == __r.size() ? __first + __idx : __first + __idx + __count);
+
+    return {__first + __idx, __end};
```
dmitriy-sobolev: `__first + __idx` appears 3 times. Perhaps, we could store it in a variable

PR #2482, opened 2025-09-25
Fix incorrect return type for `oneapi::dpl::__internal::__ranges::__pattern_minmax_element` + `__hetero_tag`

SergeyKopienko:
This PR fixes the return type for the `__pattern_minmax_element` function template when used with the `__hetero_tag`. The change corrects the return type from `std::ranges::borrowed_iterator_t<_R>` to `std::ranges::iterator_t<_R>` for both elements of the returned pair.

- Changed return type from borrowed iterator to regular iterator for consistency with the function's actual behavior

This error has been introduced in #2395


SergeyKopienko:
Not a bug

PR #2488, opened 2025-10-02
Remove `operator[]`, `size()` and `empty()` requirements from range-based algorithm implementations

SergeyKopienko:
This PR removes the requirement for `operator[]` and `size()` in range-based algorithm implementations by introducing infrastructure to handle ranges that only provide begin() and end() methods. The changes enable algorithms to work with minimalistic ranges while maintaining compatibility with existing ranges that do provide `operator[]`.

- Introduces a new `MinimalisticRange` template for testing ranges with only `begin`/`end` methods (prepared by @dmitriy-sobolev)
- Adds `__get_subscription_view` utility to provide subscription access for ranges without `operator[]`
- Updates internal size calculation and access patterns throughout the codebase

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
@@ -694,7 +694,7 @@ __pattern_remove_if(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, 
 {
     const auto __n = oneapi::dpl::__ranges::__size(__rng);
     if (__n == 0)
-        return __n;
+        return 0;
 
     using _ValueType = oneapi::dpl::__internal::__value_t<_Range>;
 
--- a/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
@@ -61,7 +61,7 @@ __pattern_walk_n(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Function
         oneapi::dpl::__par_backend_hetero::__parallel_for(
             _BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
             unseq_backend::walk_n_vectors_or_scalars<_Function>{__f, static_cast<std::size_t>(__n)}, __n,
-            std::forward<_Ranges>(__rngs)...)
+            oneapi::dpl::__ranges::__get_subscription_view(std::forward<_Ranges>(__rngs))...)
             .__checked_deferrable_wait();
     }
     return __n;
--- a/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
+++ b/include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
@@ -170,7 +170,9 @@ __pattern_swap(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range1&& _
             _BackendTag{},
             oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__swap1_wrapper>(
                 std::forward<_ExecutionPolicy>(__exec)),
-            unseq_backend::__brick_swap<_Function>{_Function{}, __n1}, __n1, __rng1, __rng2)
+            unseq_backend::__brick_swap<_Function>{_Function{}, __n1}, __n1,
+            oneapi::dpl::__ranges::__get_subscription_view(__rng1),
+            oneapi::dpl::__ranges::__get_subscription_view(__rng2))
             .__checked_deferrable_wait();
         return __n1;
     }
@@ -178,7 +180,9 @@ __pattern_swap(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range1&& _
     oneapi::dpl::__par_backend_hetero::__parallel_for(
         _BackendTag{},
         oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__swap2_wrapper>(std::forward<_ExecutionPolicy>(__exec)),
-        unseq_backend::__brick_swap<_Function>{_Function{}, __n2}, __n2, __rng2, __rng1)
+        unseq_backend::__brick_swap<_Function>{_Function{}, __n2}, __n2,
+        oneapi::dpl::__ranges::__get_subscription_view(__rng2),
+        oneapi::dpl::__ranges::__get_subscription_view(__rng1))
         .__checked_deferrable_wait();
     return __n2;
 }
@@ -219,7 +223,8 @@ __pattern_equal(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range1&& 
     return !oneapi::dpl::__par_backend_hetero::__parallel_find_or(
         _BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
         _Predicate{oneapi::dpl::__internal::__not_pred<_Pred>{__pred}}, __or_tag{}, __size_calc{},
-        std::forward<_Range1>(__rng1), std::forward<_Range2>(__rng2));
+        oneapi::dpl::__ranges::__get_subscription_view(std::forward<_Range1>(__rng1)),
+        oneapi::dpl::__ranges::__get_subscription_view(std::forward<_Range2>(__rng2)));
 }
 
 #if _ONEDPL_CPP20_RANGES_PRESENT
@@ -232,8 +237,8 @@ __pattern_equal(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R1&
     oneapi::dpl::__internal::__binary_op<_Pred, _Proj1, _Proj2> __pred_2(__pred, __proj1, __proj2);
 
     return oneapi::dpl::__internal::__ranges::__pattern_equal(__tag, ::std::forward<_ExecutionPolicy>(__exec),
-        oneapi::dpl::__ranges::views::all_read(::std::forward<_R1>(__r1)),
-        oneapi::dpl::__ranges::views::all_read(::std::forward<_R2>(__r2)), __pred_2);
+        oneapi::dpl::__ranges::views::all_read(std::forward<_R1>(__r1)),
+        oneapi::dpl::__ranges::views::all_read(std::forward<_R2>(__r2)), __pred_2);
 }
 #endif //_ONEDPL_CPP20_RANGES_PRESENT
 
@@ -255,9 +260,9 @@ __pattern_find_if(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range&&
     using _TagType = oneapi::dpl::__par_backend_hetero::__parallel_find_forward_tag<_IndexType>;
     using __size_calc = oneapi::dpl::__ranges::__first_size_calc;
 
-    return oneapi::dpl::__par_backend_hetero::__parallel_find_or(_BackendTag{}, std::forward<_ExecutionPolicy>(__exec),
-                                                                 _Predicate{__pred}, _TagType{}, __size_calc{},
-                                                                 std::forward<_Range>(__rng));
+    return oneapi::dpl::__par_backend_hetero::__parallel_find_or(
+        _BackendTag{}, std::forward<_ExecutionPolicy>(__exec), _Predicate{__pred}, _TagType{}, __size_calc{},
+        oneapi::dpl::__ranges::__get_subscription_view(std::forward<_Range>(__rng)));
 }
 
 #if _ONEDPL_CPP20_RANGES_PRESENT
@@ -288,17 +293,18 @@ __pattern_find_end(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _
 {
     const auto __n1 = oneapi::dpl::__ranges::__size(__rng1);
     const auto __n2 = oneapi::dpl::__ranges::__size(__rng2);
+
     //trivial pre-checks
     if (__n1 == 0 || __n2 == 0 || __n1 < __n2)
         return __n1;
 
     if (__n1 == __n2)
     {
-        const bool __res =
-            __ranges::__pattern_equal(__tag,
-                                      oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__equal_wrapper>(
-                                          std::forward<_ExecutionPolicy>(__exec)),
-                                      __rng1, std::forward<_Range2>(__rng2), __pred);
+        const bool __res = __ranges::__pattern_equal(
+            __tag,
+            oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__equal_wrapper>(std::forward<_ExecutionPolicy>(__exec)),
+            oneapi::dpl::__ranges::__get_subscription_view(std::forward<_Range1>(__rng1)),
+            oneapi::dpl::__ranges::__get_subscription_view(std::forward<_Range2>(__rng2)), __pred);
         return __res ? 0 : __n1;
     }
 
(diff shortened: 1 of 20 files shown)


test/support/utils.h
```diff
@@ -1340,6 +1340,29 @@ struct NoDefaultCtorWrapper {
...
+////////////////////////////////////////////////////////////////////////////////
+// A minimalistic range that only provides begin() and end() methods.
+template <typename _FrowardIterator>
+struct MinimalisticRange
+{
+    _FrowardIterator __begin;
+    _FrowardIterator __end;
+
+    _FrowardIterator begin() const { return __begin; };
+    _FrowardIterator end()   const { return __end;   };
```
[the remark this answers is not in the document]
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -129,15 +135,47 @@ using projected_value_t = std::remove_cvref_t<std::invoke_result_t<Proj&, std::i
...
 #else
-    return __rng.size();
-#endif
+template <typename _Range>
+std::enable_if_t<!__is_eval_size_through_size<_Range>::value, decltype( std::distance(std::declval<_Range>().begin(), std::declval<_Range>().end()))>
+__size(_Range&& __rng)
+{
+    return std::distance(__rng.begin(), __rng.end());
 }
+#endif
```
dmitriy-sobolev, line 200: What do you think about using std::ranges::size and constexpr variable trait? For example:

```c++

template <typename _R>
auto
__size(_R&& __r)
{
    return std::ranges::size(__r)
}

#else

template <typename _R, typename = void>
constexpr bool __has_size = false;

template <typename _R>
constexpr bool __has_size<_R, std::void_t<decltype(std::declval<_R>().size())>> = true;

template <typename _R>
auto
__size(_R&& __r)
{
    if constexpr (__has_size<_R>)
        return __r.size();
    else
        return std::distance(__r.begin(), __r.end());
}
#endif
```

test/support/utils.h
```diff
@@ -1340,6 +1340,29 @@ struct NoDefaultCtorWrapper {
...
+struct MinimalisticRange
+{
+    _ForwardIterator __begin;
+    _ForwardIterator __end;
+
+    _ForwardIterator begin() const { return __begin; };
+    _ForwardIterator end()   const { return __end;   };
+};
+
+#if _ENABLE_STD_RANGES_TESTING
```
dmitriy-sobolev: No uglification is needed in the tests.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -91,7 +91,7 @@ void
 __pattern_transform(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r, _OutRange&& __out_r,
                     _F __op, _Proj __proj)
 {
-    assert(std::ranges::size(__in_r) <= std::ranges::size(__out_r)); // for debug purposes only
+    assert(oneapi::dpl::__ranges::__size(__in_r) <= oneapi::dpl::__ranges::__size(__out_r)); // for debug purposes only
```
dmitriy-sobolev: No reason to replace `std::ranges::size` with `oneapi::dpl::__ranges::__size` in c++20 algorithms. It makes sense only in patterns shared by c++20 and experimental c++17 algorithms.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -136,7 +136,7 @@ __pattern_fill(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _R&&
     oneapi::dpl::__internal::__ranges::__pattern_walk_n(__tag, std::forward<_ExecutionPolicy>(__exec), __f,
                                                         oneapi::dpl::__ranges::views::all_write(std::forward<_R>(__r)));
```
dmitriy-sobolev, line 137: Just a note that it was noticed: there is no reason to wrap a range in all/all_write/all_read views. It's better to be done separately. 

These views add a subscription operator as `__get_subscription_view`, but do much more which may introduce overheads.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -529,12 +551,14 @@ __pattern_adjacent_find(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _R
     //C++ standard libraries (f.e libstdc++ 10), where the implementation might throw C++ exceptions, that is an issue,
     //because "SYCL kernel cannot use exceptions".
 
-    auto __rng1 = oneapi::dpl::__ranges::take_view_simple(__rng, __rng.size() - 1);
-    auto __rng2 = oneapi::dpl::__ranges::drop_view_simple(__rng, 1);
+    auto&& __rng_s = oneapi::dpl::__ranges::__get_subscription_view(__rng);
+
+    auto __rng1 = oneapi::dpl::__ranges::take_view_simple(__rng_s, __n - 1);
+    auto __rng2 = oneapi::dpl::__ranges::drop_view_simple(__rng_s, 1);
```
dmitriy-sobolev: take and drop view already have operator[]. Let's roll back this change and use them as they are. This is not different from all/all_read/all_write wrappers which are kept intact.

The same applies to `zip_view`.

include/oneapi/dpl/pstl/utils_ranges.h
SergeyKopienko: Discussed offline, I think we can't apply your variant.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h

test/support/utils.h
SergeyKopienko: Fixed

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h

test/parallel_api/ranges/range_minimal_type_requirements.pass.cpp
```diff
@@ -0,0 +1,84 @@
...
+#if _ENABLE_STD_RANGES_TESTING
+    test_count(oneapi::dpl::execution::seq);
+    test_count(oneapi::dpl::execution::unseq);
+    test_count(oneapi::dpl::execution::par);
+    test_count(oneapi::dpl::execution::par_unseq);
+#if TEST_DPCPP_BACKEND_PRESENT
+    test_count(TestUtils::get_dpcpp_test_policy());
+#endif
+
+    test_merge(oneapi::dpl::execution::seq);
```
dmitriy-sobolev: Let's add more common algorithms for better coverage: `sort`, `copy_if`, `transform`, `all_of`.

SergeyKopienko: I think we can do this in some separate PR

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -662,6 +720,70 @@ struct permutation_discard_view
...
+        !__has_subsctiption_op<_Base>::value,
+        "The usage of __subscription_impl_view_simple prohibited if std::decay_t<_Source>::operator[] implemented");
+
+    using value_type = oneapi::dpl::__internal::__value_t<_Base>;
+    using index_type = oneapi::dpl::__internal::__difference_t<_Base>;
+
+    // Define default constructors
+    __subscription_impl_view_simple() = default;
+    __subscription_impl_view_simple(const __subscription_impl_view_simple&) = default;
+    __subscription_impl_view_simple(__subscription_impl_view_simple&&) = default;
```
dmitriy-sobolev: Should the base constructors and the assignment operators be derived or called?

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -128,17 +134,74 @@ using projected_value_t = std::remove_cvref_t<std::invoke_result_t<Proj&, std::i
...
+{
+    if constexpr (__has_empty<_Range>::value)
+        return __rng.empty();
+    else
+        return __rng.begin() == __rng.end();
+}
 #endif
+
+template <typename _R, typename = void>
+struct __is_eval_size_through_size : std::false_type
```
dmitriy-sobolev: `__is_eval_size_through_size` -> `__has_size`?

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -662,6 +729,70 @@ struct permutation_discard_view
     }
 };
 
+template <typename _R, typename = void>
+struct __has_subsctiption_op : std::false_type
+{
+};
+
+template <typename _R>
+struct __has_subsctiption_op<_R, std::void_t<decltype(std::declval<_R>().operator[](0))>> : std::true_type
```
dmitriy-sobolev: A typo: subsctiption -> subscription

include/oneapi/dpl/pstl/utils_ranges.h
SergeyKopienko: So far we have any custom constructor, we should implement all another forms of default constructors and assignment's operations: "Rule of Five"

SergeyKopienko: UPD: probably we can delete only `__subscription_impl_view_simple() = default;`

include/oneapi/dpl/pstl/utils_ranges.h
SergeyKopienko: Fixed (renamed)

include/oneapi/dpl/pstl/utils_ranges.h
SergeyKopienko: Fixed

test/support/utils.h
dmitriy-sobolev: Could you also remove the leading underscores for `_ForwardIterator`?

test/parallel_api/ranges/range_minimal_type_requirements.pass.cpp
dmitriy-sobolev: I'd recommend doing it here and at least  adding `copy_if` and `transform` - they use "scan" and "for" patterns, which are reused by almost a half of the algorithms in the library.

test/support/utils.h
SergeyKopienko: Done

include/oneapi/dpl/pstl/utils_ranges.h
```diff
@@ -662,6 +729,69 @@ struct permutation_discard_view
...
+    template <typename... _Args>
+    __subscription_impl_view_simple(_Args&&... __args) : _Base(std::forward<_Args>(__args)...)
+    {
+    }
+
+    // Define default operator=
+    __subscription_impl_view_simple&
+    operator=(const __subscription_impl_view_simple&) = default;
+    __subscription_impl_view_simple&
+    operator=(__subscription_impl_view_simple&&) = default;
```
dmitriy-sobolev, line 766: One more question about `__subscription_impl_view_simple`.

There are other views in the library: `take_view_simple`, `drop_view_simple`, `reverse_view_simple` and etc. They all provide size, empty, operator[]. They are simpler, but do not preserve the original methods. Is it better to preserve the methods like it's done here , or just provide size, empty and operator() which is sufficient for our internal implementation?

dmitriy-sobolev: It's probably not worth at this point. It would be useful if we initially had decided to preserve the use of all methods: .size(), .empty() and operator[].

test/parallel_api/ranges/range_minimal_type_requirements.pass.cpp
SergeyKopienko: Done.

include/oneapi/dpl/pstl/utils_ranges.h
SergeyKopienko: My opinion about the overall design of `__subscription_impl_view_simple` - it should be transparent for implemented by `_Base` `class`/`struct` methods. It should only implement one additional method - `operator[]()` (`const` and not `const`). And nothing more.

About `take_view_simple`, `drop_view_simple`, `reverse_view_simple` and etc - may be it is their design/implementation problem?

Also I should highlight that they implements `size()` method as they needed.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -1395,16 +1455,23 @@ __pattern_reduce_by_segment(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&&
         oneapi::dpl::__internal::__ranges::__pattern_walk_n(
             __tag, oneapi::dpl::__par_backend_hetero::make_wrapped_policy<__copy_keys_values_range_wrapper>(__exec),
             __copy_range,
-            oneapi::dpl::__ranges::zip_view(std::forward<_Range1>(__keys), std::forward<_Range2>(__values)),
-            oneapi::dpl::__ranges::zip_view(std::forward<_Range3>(__out_keys), std::forward<_Range4>(__out_values)));
+            oneapi::dpl::__ranges::zip_view(
+                oneapi::dpl::__ranges::__get_subscription_view(std::forward<_Range1>(__keys)),
```
SergeyKopienko, line 1460: @dmitriy-sobolev I believe we need to call `__get_subscription_view` here because inside `zip_view` we have usages of `operator[]` from these ranges: https://github.com/uxlfoundation/oneDPL/blob/73cb69658be365dc82fe4c933d782d8b13444b80/include/oneapi/dpl/pstl/utils_ranges.h#L256

SergeyKopienko: @dmitriy-sobolev, I believe we should remove changes from
```C++
__pattern_walk_n(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Function __f, _Ranges&&... __rngs)
```
- I mean `__get_subscription_view` doesn't needed there

UPD: discussed, let's save as is.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -826,21 +857,26 @@ template <typename _BackendTag, typename _ExecutionPolicy, typename _Range, type
...
-        return __rng.size();
+    const auto __n = oneapi::dpl::__ranges::__size(__rng);
+    if (__n == 0)
+        return __n;
 
     using _ValueType = oneapi::dpl::__internal::__value_t<_Range>;
 
-    oneapi::dpl::__par_backend_hetero::__buffer<_ValueType> __buf(__rng.size());
-    auto res_rng = oneapi::dpl::__ranges::views::all(__buf.get_buffer());
+    auto&& __rng_s = oneapi::dpl::__ranges::__get_subscription_view(__rng);
```
dmitriy-sobolev, line 866: Is not it better to use `oneapi::dpl::__ranges::__get_subscription_view __rng_s(std::forward<_Range>(__rng));`?
I do not quite understand the purpose of `auto&&` in the context of calling a constructor.

dmitriy-sobolev: Should it rather be:
```suggestion
    auto __rng_s = oneapi::dpl::__ranges::__get_subscription_view(std::forward<_Range>(__rng));
```

SergeyKopienko: Why `auto&& __rng_s` still needed: because it's preserve value category of returned value.
This allows us to write
```C++
std::forward<decltype(__rng_s)>(__rng_s)
```
later in the code and it will make sense: if `__get_subscription_view` returns source parameter `__rng` without changes, we will preserve their value category.

PR #2489, opened 2025-10-03
Removal of Selection API from Dynamic Selection and Token Policy

danhoeflinger:
Draft PR for Dynamic Selection and removal of public `select` API, as well as introduction of a new policy, TokenPolicy.

### DIFF
--- a/rfcs/proposed/dynamic_selection_customization/token_policy_and_selection_removal.md
+++ b/rfcs/proposed/dynamic_selection_customization/token_policy_and_selection_removal.md
@@ -1,15 +1,9 @@
 # Token Policy and Changes to Policy Contract
 
 ## Introduction
-The key distinguishing feature between different dynamic selection policies is the method in which they select resources
-each to submit jobs to. Currently, select is a standalone public API which can be called to obtain a resource, but can
-also be invoked by directly calling `submit` or `submit_and_wait` directly. In this RFC, we propose to remove the
-standalone `select` API, as well as introduce a new policy, TokenPolicy. TokenPolicy provides users a way to specify
-hard concurrancy limits on resources.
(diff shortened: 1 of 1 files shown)


rfcs/proposed/dynamic_selection_customization/token_policy_and_selection_removal.md
```diff
@@ -0,0 +1,103 @@
...
+
+```cpp
+  template <typename Policy>
+  struct policy_traits
+  {
+      using resource_type = typename std::decay_t<Policy>::resource_type;
+  };
+```
+
+With this contract, we could also provide a generic `submit_and_wait` implementation that uses `submit` and waits on the result unless overridden.
```
vossmjp:
```suggestion
With this contract, if `p.submit(f, args…)` is well-formed, we will provide a generic `submit_and_wait` implementation that uses `submit` and waits on the result unless overridden.
```

rfcs/proposed/dynamic_selection_customization/token_policy_and_selection_removal.md
```diff
@@ -0,0 +1,103 @@
...
+We can provide an asynchronous `submit` call, but allow that call to fail, with some status flag in the submission object returned that would need to be queried. Currently, `submit` does not have the option of failing for any other policies, so this is extra complexity we would be adding to the API. However, it would allow us to provide both submit calls for TokenPolicy. If we do allow submit to fail, then we would not be able to generically implement `submit_and_wait`, at least without providing a similar fail status for that function as well.
+
+### Benefits of Removing Select for Other Policies
+Beyond simplifying the public interface and requirements, these changes may have some inherent benefits for existing policies in that they force users into a specific usage pattern. Selections and submissions must be paired 1-to-1, and selection should occur very close to submission time. For Dynamic Load Policy and Autotune Policy, which dynamically use statistics about resource load and performance of jobs, when selection occurs close to submission, the selection will be more accurate and up-to-date for the upcoming submission.
```
vossmjp: Selections and submission do not need to be paired 1-1. It is possible to call select, hold the resource and call submit multiple times using the same selection. The change removes this possibility, which is ok.

danhoeflinger: Yes, the intention was to describe that with these changes, the 1-to-1 pairing is now required, but I can make this more clear.

PR #2497, opened 2025-10-10
Fix CMake OpenMP flag checks with CL

dmitriy-sobolev:
cl accepts `-openmp-simd` and produces:
> cl : Command line warning D9035 : option 'o' has been deprecated and will be removed in a future release

This option is then passed to cl with a cmake message:
> oneDPL: OpenMP SIMD is enabled by passing '-openmp-simd' to compiler

Obviously, this option is not supported there and confused with `-o`.


SergeyKopienko:
[Non-blocking suggestion] It may be helpful  in the future to emit an explicit `CMake` status message indicating that `SIMD` probing is intentionally disabled for `MSVC`.

timmiesmith:
> @timmiesmith, I'd like to merge this PR for 2022.11: it is a low risk bug fix. Any objections?

No objection to merging this.

PR #2498, opened 2025-10-20
Extend CI testing

akukanov:
These changes add more C++20 configurations to CI and update the filters for SYCL (dpcpp) configurations to include more tests.

### DIFF
--- a/test/xpu_api/language.support/support.limits/limits/infinity.pass.cpp
+++ b/test/xpu_api/language.support/support.limits/limits/infinity.pass.cpp
@@ -70,7 +70,7 @@ ONEDPL_TEST_NUM_MAIN
     test<__uint128_t>(0);
 #endif
 TEST_DIAGNOSTIC_PUSH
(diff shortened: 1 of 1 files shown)


danhoeflinger (dismissed):
LGTM

.github/workflows/ci-testing.yml
```diff
@@ -204,8 +211,13 @@ jobs:
             done
 
             #set targets to just sycl iterator for dpcpp tests
```
timmiesmith:
```suggestion
```
I think the comment should be removed, or corrected to describe the new set of tests.

timmiesmith (approved):
I have one small nit with a comment being outdated, but that can be postposed to a later PR if you like.

.github/workflows/ci-testing.yml
akukanov: Thanks, the comment is updated.

PR #2499, opened 2025-10-21
Refactor copy_if and unique_copy algorithms

akukanov:
The patch unifies the implementations of `copy_if` and `unique_copy` for CPU policies.
I hope that this will help with adding support for bounded output for `unique_copy` later.

I have tried to organize the change history into logical stages which can be reviewed one by one (per commit), if that helps.

### DIFF
--- a/include/oneapi/dpl/pstl/unseq_backend_simd.h
+++ b/include/oneapi/dpl/pstl/unseq_backend_simd.h
@@ -302,20 +302,18 @@ _DifferenceType
 __simd_copy_by_mask(_InputIterator __first, _DifferenceType __n, _OutputIterator __result, bool* __mask,
                     _Assigner __assigner) noexcept
 {
-    _DifferenceType __cnt = 0;
-    _ONEDPL_PRAGMA_SIMD
+    std::make_signed_t<_DifferenceType> __cnt = -1; // to use inclusive scan of the mask
+    _ONEDPL_PRAGMA_SIMD_SCAN(+ : __cnt)
     for (_DifferenceType __i = 0; __i < __n; ++__i)
     {
+        __cnt += __mask[__i];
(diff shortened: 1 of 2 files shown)


include/oneapi/dpl/pstl/unseq_backend_simd.h
```diff
@@ -282,31 +282,16 @@ __simd_copy_if(_InputIterator __first, _DifferenceType __n, _OutputIterator __re
...
 _DifferenceType
-__simd_calc_mask_2(_InputIterator __first, _DifferenceType __n, bool* __mask, _BinaryPredicate __pred) noexcept
+__simd_compute_mask(bool* __mask, _DifferenceType __n, _Predicate __pred, _Iterators... __it) noexcept
 {
     _DifferenceType __count = 0;
 
     _ONEDPL_PRAGMA_SIMD_REDUCTION(+ : __count)
     for (_DifferenceType __i = 0; __i < __n; ++__i)
     {
-        __mask[__i] = !__pred(__first[__i], __first[__i - 1]);
```
MikeDvorskiy, line 294: Hm..
How did it work in the past (and now) ?
I see a potential a "read violation" issue in case of `i == 0` - `__first[__i - 1]`
More likely a caller processes the first element in different way and does call `__simd_calc_mask_2` starting with the second element. Even it is spot on, it is not good approach to have such a potentially dangerous function.

Yes, the change introduces the  unified `__simd_compute_mask` which looks safer to use.
I'm in favor of such changes.

akukanov: > More likely a caller processes the first element in different way and does call __simd_calc_mask_2 starting with the second element.

Yes, there was that special boundary case in `__pattern_unique_copy` - the first element is always copied, and processing can start with the second one. It is still there, though simplified a lot.

PR #2501, opened 2025-10-27
Add limited output support to copy_if with device policies

akukanov:
This patch adds support for output ranges of potentially insufficient size to `ranges::copy_if`.

Of the three kernel implementations, only two are modified: the one for a single group and the "fallback" `__parallel_scan_copy` one, while `__parallel_reduce_then_scan_copy` is not used if the output size is less than that of the input (it's left as a future work and should be done carefully to avoid performance loss).

The key changes are in `parallel_backend_sycl.h` and `unseq_backend_sycl.h`.

### DIFF
--- a/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
+++ b/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
@@ -534,7 +534,7 @@ struct __parallel_copy_if_single_group_functor<__internal::__optional_kernel_nam
                                 __lacc[2 * __n_uniform] = __idx; // the actual stop position in the input
                         }
                     }
-                    sycl::group_barrier(__group);
+                    __dpl_sycl::__group_barrier(__self_item);
 
                     if (__item_id == 0)
                     {
@@ -911,7 +911,7 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     const std::size_t __max_slm_size =
         __q_local.get_device().template get_info<sycl::info::device::local_mem_size>() / 2;
 
-    // The kernel stores n integers for the predicate and another n integers for the offsets
+    // n predicate evaluations, n offsets, 1 element for the stop position in the input
     const auto __req_slm_size = sizeof(std::uint16_t) * (__n_uniform * 2 + 1);
 
     constexpr std::uint16_t __max_elem_per_item = 2;


include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -474,62 +501,72 @@ template <typename _Size, typename... _ScanKernelName>
 struct __parallel_copy_if_single_group_submitter<_Size, __internal::__optional_kernel_name<_ScanKernelName...>>
 {
     template <typename _InRng, typename _OutRng, typename _UnaryOp, typename _Assign>
-    __future<sycl::event, __result_and_scratch_storage<_Size>>
-    operator()(sycl::queue& __q, _InRng&& __in_rng, _OutRng&& __out_rng, std::size_t __n, _UnaryOp __unary_op,
-               _Assign __assign, std::uint16_t __n_uniform, std::uint16_t __wg_size)
+    std::array<_Size, 2>
```
SergeyKopienko, line 487: I propose to return not `std::array` but `std::tuple` for compatibility with another our code

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -287,16 +289,16 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
         auto __n_groups = oneapi::dpl::__internal::__dpl_ceiling_div(__n, __size_per_wg);
         // Storage for the results of scan for each workgroup
 
-        using __result_and_scratch_storage_t = __result_and_scratch_storage<_Type>;
-        __result_and_scratch_storage_t __result_and_scratch{__q, __n_groups + 1};
+        __combined_storage<_Type> __result_and_scratch{__q, __n_groups + 1, /*result size*/ 2};
```
SergeyKopienko: I propose to rename `__result_and_scratch` into something else to highlight that it is the instance of `__combined_storage` or to highlight the semantic of this variable,

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -702,9 +739,11 @@ __parallel_transform_scan(oneapi::dpl::__internal::__device_backend_tag, _Execut
         // scan between groups
         unseq_backend::__scan</*inclusive=*/std::true_type, _BinaryOperation, _NoOpFunctor, _NoAssign, _Assigner,
                               _NoOpFunctor, unseq_backend::__no_init_value<_Type>>{
-            __binary_op, _NoOpFunctor{}, __no_assign_op, __assign_op, __get_data_op},
+            __binary_op, _NoOpFunctor{}, __ignore_op, __assign_op, __get_data_op},
         // global scan
-        unseq_backend::__global_scan_functor<_Inclusive, _BinaryOperation, _InitType>{__binary_op, __init});
+        unseq_backend::__global_scan_functor<_Inclusive, _BinaryOperation, _InitType>{__binary_op, __init},
+        /*apex*/ __ignore_op);
+    return __future(__event, __result_and_scratch_storage<_Type>(__transfer_state_from(__payload)));
```
```suggestion
    return __future(__event, __result_and_scratch_storage<_Type>(__transfer_state_from(std::move(__payload))));
```
Does it have sense to move `__payload` into `__transfer_state_from` ?
Here and in all other places

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -899,31 +942,45 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
     {
         using _KernelName = oneapi::dpl::__par_backend_hetero::__internal::__kernel_name_provider<
             __scan_copy_single_wg_kernel<_CustomName>>;
-        return __par_backend_hetero::__parallel_copy_if_single_group_submitter<_Size, _KernelName>()(
-            __q_local, std::forward<_InRng>(__in_rng), std::forward<_OutRng>(__out_rng), __n, __pred, __assign,
+        __ret = __par_backend_hetero::__parallel_copy_if_single_group_submitter<_Size, _KernelName>()(
```
SergeyKopienko: I am not sure that there is real reasons in our `hetero` implementations of submitters to change return type from `sycl::event` + `results storage` to exactly `result` and `wait` inside implentation.
I prefer to have the same approach for all submitters....

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -654,16 +655,15 @@ oneapi::dpl::__internal::__difference_t<_Range2>
 __pattern_copy_if(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2,
                   _Predicate __pred, _Assign __assign)
 {
-    oneapi::dpl::__internal::__difference_t<_Range2> __n = oneapi::dpl::__ranges::__size(__rng1);
-    if (__n == 0)
+    oneapi::dpl::__internal::__difference_t<_Range2> __n = oneapi::dpl::__ranges::__size(__rng1),
+                                                     __m = oneapi::dpl::__ranges::__size(__rng2);
```
SergeyKopienko: Variant 1:

```suggestion
    oneapi::dpl::__internal::__difference_t<_Range2> __n = oneapi::dpl::__ranges::__size(__rng1);
    oneapi::dpl::__internal::__difference_t<_Range2> __m = oneapi::dpl::__ranges::__size(__rng2);
```

Variant 2:
```suggestion
    auto __n = oneapi::dpl::__ranges::__size(__rng1);
    auto __m = oneapi::dpl::__ranges::__size(__rng2);
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: The `__copy_return` routine writes the result into a contiguous storage, therefore an array is needed. Moreover, `std::array` supports `std::get` etc. and can sometimes be treated as a tuple.
If there are compatibility/interoperability questions, let's discuss. For mere consistency however I would not want to complicate the design.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: Well, the semantics of this variable is that it stores both the scratch memory to use during the computation and the result to get back :)
I am not opposed to renaming. Do you have a suggestion though?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: `__transfer_state_from` is written to take either rvalues or lvalues, and does the move internally. I can rename it to `__move_state_from` for clarity.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
akukanov: What is wrong with the current variant? :)

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: The reason is that this very kernel is never used by async or even by deferred-wait algorithms, it is specific for `copy_if` which has to be blocking (especially if the output is bounded). The same is true for `__parallel_scan_copy` and `__parallel_reduce_then_scan_copy` BTW, so maybe those should also be changed similarly (but their underlying scan patterns and respective submitters would be still asynchronous).

If others think it is better to keep all submitters consistent, no matter how these are used, I will change the one in question to return a tuple, though I prefer encapsulation over consistency in this case.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: OK

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: What about `__results_container` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: @danhoeflinger, what do you think?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: OK

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
SergeyKopienko: Technically it's ok, but never used in other places. Of course, up to you.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: Changed in #2510

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
danhoeflinger: If I remember correctly, we already don't have a consistent convention for `__parallel_*` level functions in the dpcpp backend to be asynchronous (though most are asynchronous).  
`__parallel_find_or`, `__parallel_scan_by_segment`, `__parallel_reduce_by_segment` are all blocking.

I don't really have a problem with this. Keeping it asynchronous allows more flexibility in the future, but probably not something we would take advantage of.  On the other hand, not overusing asynchrony where its not actually used keeps complexity down, and may allow more flexibility within the altered routines.

akukanov: I am inclining to change the `_submitter` suffix to something else, e.g. `_functor`, in order to indicate the semantic difference. Naming ideas are welcome.

SergeyKopienko: I think `_functor` is ok for me.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
danhoeflinger: I think from these choices, my preference is variant 1, but I understand the benefits of the original.

However, I wonder this is actually better:

```
using common_diff_t = std::common_type_t<oneapi::dpl::__internal::__difference_t<_Range1>,
                                         oneapi::dpl::__internal::__difference_t<_Range2>>;
common_diff_t __n = oneapi::dpl::__ranges::__size(__rng1);
common_diff_t __m = oneapi::dpl::__ranges::__size(__rng2);
```

SergeyKopienko (dismissed):
LGTM

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: OK, I will rename it somehow to avoid confusion with the "old" type.

test/parallel_api/ranges/std_ranges_test.h
```diff
@@ -915,6 +936,12 @@ struct test_range_algo
     void
     operator()(auto algo, auto& checker, auto... args)
     {
+        
+        std::cout << "Run configuration: " <<  n_serial << ", " << n_parallel <<
+#if TEST_DPCPP_BACKEND_PRESENT
+            ", " << n_device <<
+#endif
+            std::endl;
```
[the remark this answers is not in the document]
dmitriy-sobolev: It's a test, hence it's rather optional to remove.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -654,16 +655,17 @@ oneapi::dpl::__internal::__difference_t<_Range2>
 __pattern_copy_if(__hetero_tag<_BackendTag>, _ExecutionPolicy&& __exec, _Range1&& __rng1, _Range2&& __rng2,
                   _Predicate __pred, _Assign __assign)
 {
-    oneapi::dpl::__internal::__difference_t<_Range2> __n = oneapi::dpl::__ranges::__size(__rng1);
-    if (__n == 0)
+    using _Size = std::common_type_t<oneapi::dpl::__internal::__difference_t<_Range1>,
+                                     oneapi::dpl::__internal::__difference_t<_Range2>>;
+    _Size __n = oneapi::dpl::__ranges::__size(__rng1);
+    _Size __m = oneapi::dpl::__ranges::__size(__rng2);
+    if (__n == 0 || __m == 0)
```
dmitriy-sobolev: [nitpick] Let's use __n and __n_out as in the __pattern_bounded_copy_if for consistency.
Also, __n and __m may be confused with N and M from the c++ spec, but their meanings differ.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -465,69 +477,79 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
                     {
-                        if (__lacc[__idx])
-                            __assign(static_cast<__tuple_type>(__in_rng[__idx]),
-                                     __out_rng[__lacc[__idx + __n_uniform]]);
+                        if (__lacc[__idx]) {
+                            _ValueType __out_idx = __lacc[__idx + __n_uniform];
+                            if (__out_idx < __m)
+                                __assign(static_cast<__tuple_type>(__in_rng[__idx]), __out_rng[__out_idx]);
+                            if (__out_idx == __m)
+                                __lacc[2 * __n_uniform] = __idx; // the actual stop position in the input
```
dmitriy-sobolev: `if` -> `else if`?

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -673,15 +675,29 @@ std::ranges::copy_if_result<std::ranges::borrowed_iterator_t<_InRange>, std::ran
 __pattern_copy_if_ranges(__hetero_tag<_BackendTag> __tag, _ExecutionPolicy&& __exec, _InRange&& __in_r,
                          _OutRange&& __out_r, _Pred __pred, _Proj __proj)
 {
+    using _Size = std::common_type_t<oneapi::dpl::__internal::__difference_t<_InRange>,
```
SergeyKopienko: We can use
```C++
using _SizeType = oneapi::dpl::__ranges::__common_size_t<_InRange,_OutRange>;
```

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
```diff
@@ -673,15 +675,29 @@ std::ranges::copy_if_result<std::ranges::borrowed_iterator_t<_InRange>, std::ran
...
+        return {__found_it, std::ranges::begin(__out_r)};
+    }
+
     oneapi::dpl::__internal::__unary_op<_Pred, _Proj> __pred_1{__pred, __proj};
 
-    auto __res_idx = oneapi::dpl::__internal::__ranges::__pattern_copy_if(__tag,
-        std::forward<_ExecutionPolicy>(__exec), oneapi::dpl::__ranges::views::all_read(__in_r),
-        oneapi::dpl::__ranges::views::all_write(__out_r), __pred_1,
-        oneapi::dpl::__internal::__pstl_assign());
+    std::array<_Size, 2> __stops = oneapi::dpl::__par_backend_hetero::__parallel_copy_if(
```
SergeyKopienko, line 693: What about to preserve order in this result of `__parallel_copy_if` according to arguments?
I mean `[0]` - amount of processed in the `__in_r`
'[1]' -  - amount of processed in the '__out_r`

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -465,69 +477,79 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
-                __result.template __get_result_acc<sycl::access_mode::write>(__hdl, __dpl_sycl::__no_init{});
+            auto __lacc = __dpl_sycl::__local_accessor<_ValueType>(sycl::range<1>(std::size_t(__n_uniform) * 2 + 1), __hdl);
+            auto __res_acc = __get_accessor(sycl::write_only, __result, __hdl, __dpl_sycl::__no_init{});
 
-            __hdl.parallel_for<_ScanKernelName...>(
-                sycl::nd_range<1>(__wg_size, __wg_size), [=](sycl::nd_item<1> __self_item) {
-                    auto __res_ptr = __result_and_scratch_storage_t::__get_usm_or_buffer_accessor_ptr(__res_acc);
+            __hdl.parallel_for<_ScanKernelName...>(sycl::nd_range<1>(__wg_size, __wg_size),
+                [=](sycl::nd_item<1> __self_item) {
+                    _Size* __res_ptr = __res_acc.__data();
```
SergeyKopienko: What about to move this local variable under condition check `if (__item_id == 0)` ?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -858,21 +884,23 @@ __parallel_partition_copy(oneapi::dpl::__internal::__device_backend_tag, _Execut
...
     }
 }
 
 template <typename _ExecutionPolicy, typename _InRng, typename _OutRng, typename _Size, typename _Pred,
           typename _Assign = oneapi::dpl::__internal::__pstl_assign>
-__future<sycl::event, __result_and_scratch_storage<_Size>>
+std::array<_Size, 2>
 __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPolicy&& __exec, _InRng&& __in_rng,
-                   _OutRng&& __out_rng, _Size __n, _Pred __pred, _Assign __assign = _Assign{})
+                   _OutRng&& __out_rng, _Size __n, _Size __m, _Pred __pred, _Assign __assign = _Assign{})
```
SergeyKopienko: I think we always have deal with the sized ranges.
This mean `__in_rng` and `__out_rng` knows their own size.
So than why we need extra parameters `_Size __n` and `_Size __m` here?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -465,69 +477,79 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
                     {
+                        _ValueType __stop_in = __lacc[2 * __n_uniform];
+                        __res_ptr[1] = __stop_in;
                         // Add predicate of last element to account for the scan's exclusivity
-                        *__res_ptr = __lacc[__n_uniform + __n - 1] + __lacc[__n - 1];
+                        __res_ptr[0] = (__stop_in == __n) ? __lacc[__n_uniform + __n - 1] + __lacc[__n - 1] : __m;
                     }
                 });
-        });
+        }).wait_and_throw();
```
dmitriy-sobolev, line 548: Why was wait added?
I suppose the original implementation without wait was needed to support asynchronous execution.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -788,11 +812,6 @@ struct __partition_by_mask
                 get<1>(__out_acc[__out_idx]) = static_cast<__tuple_type>(get<0>(__in_acc[__item_idx]));
             }
         }
-        if (__item_idx == 0)
```
SergeyKopienko, line 791: After remove this code I think `__ret_ptr` parameter is not needed anymore?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
SergeyKopienko: I think it was added because `__parallel_copy_if_single_group_functor` is some kind of functor now but not only submitter, so it just return result.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
```diff
@@ -42,21 +38,27 @@ main()
...
+    test_range_algo<4, P2, data_in_out_lim>{}(dpl_ranges::copy_if, copy_if_checker, pred, &P2::proj);
+#endif
+    test_range_algo<5, int, data_in_out_lim>{big_sz}(dpl_ranges::copy_if, copy_if_checker, pred);
+    test_range_algo<6, int, data_in_out_lim>{big_sz}(dpl_ranges::copy_if, copy_if_checker, select_many);
 
-    return TestUtils::done(_ENABLE_STD_RANGES_TESTING);
+    return TestUtils::done();
 }
+#else
+std::int32_t main() { return TestUtils::done(0); } // skipped
```
SergeyKopienko: I propose to have only one `main` function in test and implement all required logic with defines and macros inside `main` function like in all other tests.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
```diff
@@ -42,21 +38,27 @@ main()
...
+    namespace dpl_ranges = oneapi::dpl::ranges;
 
-    test_range_algo<0, int, test_mode>{217}(dpl_ranges::copy_if, copy_if_checker, pred);
-    test_range_algo<1, int, test_mode>{1234}(dpl_ranges::copy_if, copy_if_checker, select_many);
-    test_range_algo<2, int, test_mode>{}(dpl_ranges::copy_if, copy_if_checker, select_many, proj);
-    test_range_algo<3, P2, test_mode>{}(dpl_ranges::copy_if, copy_if_checker, pred, &P2::x);
-    test_range_algo<4, P2, test_mode>{}(dpl_ranges::copy_if, copy_if_checker, pred, &P2::proj);
-    test_range_algo<5, int, test_mode>{big_sz}(dpl_ranges::copy_if, copy_if_checker, pred);
-#endif //_ENABLE_STD_RANGES_TESTING
+#if 0
```
SergeyKopienko: Is this `#if 0` correct?

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
SergeyKopienko: Fixed in the commit https://github.com/uxlfoundation/oneDPL/pull/2501/commits/05e75fc3912b2483f5907f6be6c49832fbce9f3d

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
dmitriy-sobolev: Sorry, I do not understand your point, in particular "not only submitter, so it just return result". It originally returned some data. What is changed is that it no longer returns an event and waits immediately. The question is why was it done?

test/parallel_api/ranges/std_ranges_test.h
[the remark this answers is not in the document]
akukanov: This output is for debugging, and will be cleaned once I am sure there are no failures for which the patch is responsible.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
akukanov: Debugging change, will be reverted before commit.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
```diff
@@ -42,21 +38,27 @@ main()
...
+    test_range_algo<0, int, data_in_out_lim>{217}(dpl_ranges::copy_if, copy_if_checker, pred);
+    test_range_algo<1, int, data_in_out_lim>{1234}(dpl_ranges::copy_if, copy_if_checker, select_many);
+    test_range_algo<2, int, data_in_out_lim>{}(dpl_ranges::copy_if, copy_if_checker, select_many, proj);
+    test_range_algo<3, P2, data_in_out_lim>{}(dpl_ranges::copy_if, copy_if_checker, pred, &P2::x);
+    test_range_algo<4, P2, data_in_out_lim>{}(dpl_ranges::copy_if, copy_if_checker, pred, &P2::proj);
+#endif
```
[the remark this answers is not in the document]
akukanov: Debugging change, will be reverted before commit.

test/parallel_api/ranges/std_ranges_copy_if.pass.cpp
SergeyKopienko: OK

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
akukanov: That will break compatibility with other algorithms that still depend on the pattern and use only the first value from the result.

SergeyKopienko: Probably we can fix these cases. isn't it?

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: While could be done, I was concerned that more complicated conditions could be harder for the compiler to vectorize.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: `__parallel_copy_if` used to return `__future`, but all its callers immediately blocked on the result. The pattern is not used (and never was) in experimental async algorithms, and the oneDPL algorithms it serves cannot pre-compute the return value (iterators at the stop positions) which depends on the actual data - so deferred waiting is not applicable either. Therefore returning the `__future` etc. is just useless overhead.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
akukanov: We can, but it is much easier and less changes, at least for now, to adjust in the only place where it matters.

SergeyKopienko: Up to you.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
akukanov: True, but only because `__partition_by_mask` has not yet been changed to support limited output.
I removed the parameter name.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: To avoid "re-computing" the same values, I guess.

I mostly followed patterns already in the code, to minimize non-critical "refactoring" changes. In this case, the input size parameter was there, so another one for the output was added.

SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
akukanov: Done.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
akukanov: Done.

include/oneapi/dpl/pstl/hetero/algorithm_ranges_impl_hetero.h
akukanov: Done,

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
SergeyKopienko: ok

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -465,69 +477,79 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
+        __q.submit([&](sycl::handler& __hdl) {
             oneapi::dpl::__ranges::__require_access(__hdl, __in_rng, __out_rng);
 
             // Local memory is split into two parts. The first half stores the result of applying the
             // predicate on each element of the input range. The second half stores the index of the output
             // range to copy elements of the input range.
-            auto __lacc = __dpl_sycl::__local_accessor<_ValueType>(sycl::range<1>(std::size_t(__n_uniform) * 2), __hdl);
-            auto __res_acc =
-                __result.template __get_result_acc<sycl::access_mode::write>(__hdl, __dpl_sycl::__no_init{});
+            auto __lacc = __dpl_sycl::__local_accessor<_ValueType>(sycl::range<1>(std::size_t(__n_uniform) * 2 + 1), __hdl);
```
dmitriy-sobolev, line 505: This line also should be updated as more SLM is used:

https://github.com/uxlfoundation/oneDPL/blob/238d76ae8b9dd932c11dde1a587a97a06bfeba61/include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h#L915

dmitriy-sobolev, line 505: Should `__n` and `__n_out` be `_ValueType` like other index-like elements?

akukanov: It does not seem necessary for correctness, as we know these values fit into 16 bytes. Whether it helps performance and how much, we'd need to measure.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -884,7 +912,7 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
         __q_local.get_device().template get_info<sycl::info::device::local_mem_size>() / 2;
 
     // The kernel stores n integers for the predicate and another n integers for the offsets
```
dmitriy-sobolev: [nitpick] Reflect the code change in the comment.

```suggestion
    // n predicate evaluations, n offsets, 1 element for the stop position in the input
```

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -465,69 +477,79 @@ struct __parallel_transform_scan_static_single_group_submitter<_Inclusive, _Elem
...
-                                     __out_rng[__lacc[__idx + __n_uniform]]);
+                        if (__lacc[__idx]) {
+                            _ValueType __out_idx = __lacc[__idx + __n_uniform];
+                            if (__out_idx < __n_out)
+                                __assign(static_cast<__tuple_type>(__in_rng[__idx]), __out_rng[__out_idx]);
+                            if (__out_idx == __n_out)
+                                __lacc[2 * __n_uniform] = __idx; // the actual stop position in the input
+                        }
                     }
+                    sycl::group_barrier(__group);
```
dmitriy-sobolev:
```suggestion
                    __dpl_sycl::__group_barrier(__self_item);
```

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -724,40 +742,46 @@ struct __copy_by_mask
...
-                // Now we have the following assignment ::std::tuple<U..> = ::std::tuple<T..> which works as expected.
-                // NOTE: we only need this explicit conversion when we have internal::tuple and
-                // ::std::tuple as operands, in all the other cases this is not necessary and no conversion
-                // is performed(i.e. __typle_type is the same type as its operand).
-                __assigner(static_cast<__tuple_type>(get<0>(__in_acc[__item_idx])), __out_acc[__out_idx]);
+            {
+                if (__out_idx < __n_out)
+                    __assigner(static_cast<__tuple_type>(get<0>(__in_acc[__item_idx])), __out_acc[__out_idx]);
+                if (__out_idx == __n_out)
+                    __ret_ptr[1] = __item_idx;
```
dmitriy-sobolev, line 749: Is not it done by apex operation (`__copy_by_mask_stops`)?

dmitriy-sobolev: Upd. I see a comment there:

> // The stop position in the input is determined by final scan

So, it looks like that final scan operation.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -896,31 +924,37 @@ __parallel_copy_if(oneapi::dpl::__internal::__device_backend_tag, _ExecutionPoli
...
-                                                 _CopyOp{_ReduceOp{}, __assign});
+        auto&& [__event, __payload] = __parallel_scan_copy<_CustomName>(__q_local, std::forward<_InRng>(__in_rng),
+            std::forward<_OutRng>(__out_rng), __n, _CreateOp{__pred}, _CopyOp{_ReduceOp{}, __assign});
+        __event.wait_and_throw();
+        __payload.__copy_result(__ret.data(), __ret.size());
     }
+
+    assert(__ret[0] >= 0 && __n_out >= __ret[0]);
+    assert(__ret[1] > 0 && __n >= __ret[1]);
+    assert(__ret[0] == __n_out || __ret[1] == __n);
```
dmitriy-sobolev, line 956: Should these asserts be deleted before mering this PR?

akukanov: I would keep these. The reduce-then-scan pattern yet needs to be updated (unless we decide not to), and I see no harm to have these basic post-conditions checked.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -634,10 +634,12 @@ struct __mask_assigner
 {
     template <typename _Acc, typename _OutAcc, typename _OutIdx, typename _InAcc, typename _InIdx>
     void
-    operator()(_Acc& __acc, _OutAcc&, const _OutIdx __out_idx, const _InAcc& __in_acc, const _InIdx __in_idx) const
+    operator()(_Acc& __acc, _OutAcc&, const _OutIdx __out_idx, const _OutIdx __acc_sz, _OutIdx, const _InAcc& __in_acc,
+               const _InIdx __in_idx) const
     {
         using ::std::get;
-        get<N>(__acc[__out_idx]) = __in_acc[__in_idx];
+        if (__out_idx < __acc_sz)
```
danhoeflinger, line 641: I'm wondering about performance impact of this for the cases which do not require it (iterator API).

This is adding a branch to the inner loop operation, though it is likely something which can just become an extra mask operation.  Its something to note if we see any performance impact, we can return to this and make this check optional at compile-time.

include/oneapi/dpl/pstl/hetero/dpcpp/unseq_backend_sycl.h
```diff
@@ -660,17 +662,19 @@ struct __scan_assigner
 
     template <typename _Acc, typename _OutAcc, typename _OutIdx, typename _InAcc, typename _InIdx>
     void
-    operator()(_Acc&, _OutAcc& __out_acc, const _OutIdx __out_idx, const _InAcc& __in_acc, _InIdx __in_idx) const
+    operator()(_Acc&, _OutAcc& __out_acc, const _OutIdx __out_idx, _OutIdx, const _OutIdx __out_sz,
+               const _InAcc& __in_acc, _InIdx __in_idx) const
     {
-        __out_acc[__out_idx] = __in_acc[__in_idx];
+        if (__out_idx < __out_sz)
```
danhoeflinger, line 668: Similar comment about performance for this bounds check.

include/oneapi/dpl/pstl/hetero/dpcpp/parallel_backend_sycl.h
```diff
@@ -305,8 +308,14 @@ struct __parallel_scan_submitter<_CustomName, __internal::__optional_kernel_name
 #endif
                 sycl::nd_range<1>(__n_groups * __wgroup_size, __wgroup_size), [=](sycl::nd_item<1> __item) {
                     auto __temp_ptr = __temp_acc.__data();
-                    __local_scan(__item, __n, __local_acc, __rng1, __rng2, __temp_ptr, __size_per_wg, __wgroup_size,
-                                 __iters_per_witem, __init);
+                    __local_scan(__item, __n, __n_out, __local_acc, __rng1, __rng2, __temp_ptr, __size_per_wg,
+                                 __wgroup_size, __iters_per_witem, __init);
+                    if (__n_groups == 1)
+                    {
+                        __dpl_sycl::__group_barrier(__item);
```
danhoeflinger, line 315: This barrier (and one more) is also something to consider avoiding at compile time for non range API if we find a non-trivial performance detriment.

